PR #2802, opened 2023-02-07
Allow Recipes to influence later Recipe Applicability

JLLeitschuh:
This is a proposal based upon a discussion here:
https://linuxfoundation.slack.com/archives/C04HR6EJ38D/p1674840301966909

This allows any Recipe to influence the applicability test of later Recipes that are executed.

This allows Recipes to provide filter logic without requiring custom filter logic in every later recipe.

This PR obviously needs tests, but it demonstrates the general idea behind the proposal.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/filter/RecipeApplicableTest.java
+++ b/rewrite-core/src/main/java/org/openrewrite/filter/RecipeApplicableTest.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.filter;
+
+import org.openrewrite.*;
+
+/**
+ * A {@link Recipe#getApplicableTests()} test to be applied to {@link Recipe} executions.
+ * </p>
+ * <b>Note: Heavily Incubating</b>
+ * </p>
+ * </p>
+ * To use this class properly, use {@link #addToExecutionContext}.
+ */
+@Incubating(since = "7.36.0")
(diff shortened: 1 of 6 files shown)


rewrite-java/src/main/java/org/openrewrite/java/filter/SourceApplicabilityFilter.java
```diff
@@ -0,0 +1,107 @@
...
+    private static TreeVisitor<?, ExecutionContext> getVisitor(Recipe recipe) {
+        try {
+            Method getVisitor = recipe.getClass().getDeclaredMethod("getVisitor");
+            getVisitor.setAccessible(true);
+            //noinspection unchecked
+            return (TreeVisitor<?, ExecutionContext>) getVisitor.invoke(recipe);
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
JLLeitschuh, line 102: Copied from `DeclarativeRecipe`

rewrite-core/src/main/java/org/openrewrite/filter/RecipeApplicableTest.java
```diff
@@ -0,0 +1,52 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.filter;
```
JLLeitschuh, line 16: New package. Is this fine?

rewrite-core/src/main/java/org/openrewrite/filter/RecipeApplicableTest.java
```diff
@@ -0,0 +1,52 @@
...
+public interface RecipeApplicableTest {
+
+    /**
+     * The test should implement the specification described on {@link Recipe#getApplicableTest()}.
+     *
+     * @param recipe The current {@link Recipe} the {@link Recipe#getApplicableTest()} is being applied to.
+     *               This will be the same as {@link ExecutionContext#getCurrentRecipe()}.
+     * @return A tree visitor that performs an applicability test.
+     */
+    TreeVisitor<?, ExecutionContext> getTest(Recipe recipe);
```
JLLeitschuh, line 39: This argument is the same as `ExecutionContext#getCurrentRecipe()`. I could theoretically drop it, but it may also be useful being explicit. Any thoughts on keeping/not keeping it?

PR #2803, opened 2023-02-07
Added recipe to replace Apache validate.notNull with objects.requireN…

### LINKED ISSUE #2724: Replace `org.apache.commons.lang3.Validate.notNull` with `Objects.requireNonNull`
> Validate.notNull(myObject, "The object must not be null");
> 
>     Type Parameters:
>         T - the object type
>     Parameters:
>         object - the object to check
>         message - the [String.format(String, Object...)](http://download.oracle.com/javase/6/docs/api/java/lang/String.html?is-external=true#format(java.lang.String,%20java.lang.Object...)) exception message if invalid, not null
>         values - the optional values for the formatted exception message 
>     Returns:
>         the validated object (never null for method chaining) 
>     Throws:
>         [NullPointerException](http://download.oracle.com/javase/6/docs/api/java/lang/NullPointerException.html?is-external=true) - if the object is null

See : https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/Validate.html#notNull(T,%20java.lang.String,%20java.lang.Object...)

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-

srmalkan:
…onNull

Closes #2724

### DIFF
--- a/rewrite-java/src/main/resources/META-INF/rewrite/java-best-practices.yml
+++ b/rewrite-java/src/main/resources/META-INF/rewrite/java-best-practices.yml
@@ -35,7 +35,7 @@ recipeList:
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.cleanup.ReplaceApacheCommonsLang3ValidateNotNullWithObjectsRequireNonNull
 displayName: Replace `org.apache.commons.lang3.Validate#notNull` with `Objects#requireNonNull`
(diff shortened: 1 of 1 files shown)


jkschneider:
Thanks @srmalkan !

srmalkan:
Hi Jkschneider,
As I understand, Validate.notNull(..) will also rename methods having more than two arguments but Objects.requireNonNull only supports up to two arguments. As per the issue that will require a different visitor. Let me know if there is a gap in my understanding since I am new to it.

PR #2804, opened 2023-02-07
No sys out print statements

BoykoAlex:
Language Server <-> client comm is corrupted by sys out statements since standard in and out are used for JSON RPC message exchange only.


knutwannheden:
@BoykoAlex In that case this line could also be problematic: https://github.com/openrewrite/rewrite/blob/b00c646dc77df593831681bad09382d7b5b921c4/rewrite-java/src/main/java/org/openrewrite/java/cleanup/ReplaceLambdaWithMethodReference.java#L111

PR #2759, opened 2023-02-02
Fixes issue of Yaml recipes in the list should not be run if single source applicability tests failed

kunli2:
Fixes #2754 

This fixes two tests introduced by PR #2753 

This PR includes and required  https://github.com/openrewrite/rewrite/pull/2765 to Improve `ListUtils.flatMap()` to return equaled reference if no change is made.

> todo: kun, add a test to test source files widened case with single source applicability test.

### Root cause
For recipes in the list that override `visit(List<SourceFile> before, ExecutionContext ctx)` method, the `visit()` methods are invoked directly without the single-source applicability test.

## Notes to reviewer
1. This PR depends on the equality check fix of `listUtil.flatMap()`, there are several PRs to address this, and either one is fine.
   - #2765
   - #2771
3. This PR depends on a fix `RewriteTest.rewriteRun()` that cannot detect unexpected newly generated files #2780

PR #2814, opened 2023-02-08
Fixes issue of Yaml recipes in the list should not be run if single source applicability tests failed [2nd approach]

### LINKED ISSUE #2754: Yaml recipes in the list should not be run if single source applicability tests failed
Thanks @shanman190 for discovering and reporting this as part of https://github.com/openrewrite/rewrite/issues/2746

this issue only happened with recipes that have the `visit(List<SourceFile> before, ExecutionContext ctx)` method overridden.

A recipe has two different ways to run.
- 1. Override `getVisitor()` method, and invoke `getVisitor().visit(@Nullable Tree tree, P p)` to transform a single file.
- 2. Override `visit(List<SourceFile> before, ExecutionContext ctx)` method, and invoke it to transform multiple files.
Typically, for a recipe, only one of the two methods mentioned above is required to be overridden.
The recipe scheduler invokes both methods in different places in the flow, all recipes in the list should not be run only when the applicability test fail.

This yaml recipe can reproduce the problem when applied to a plain text file containing the text "2", expected to have no change.

```yaml
          ---
          name: org.openrewrite.RecipesToTransformMultiFiles
          displayName: Recipes in the list which override 'visit()' method should not be run if applicability test fail
          description: >
            A recipe has two different ways to run.
            1. Override 'getVisitor()' method, and invoke `getVisitor().visit(@Nullable Tree tree, P p)` to transform a single file.
            2. Override `visit(List<SourceFile> before, ExecutionContext ctx)` method, and invoke it to transform multiple files.
            Typically, for a recipe, only one of the two methods mentioned above is required to be overridden.
            The recipe scheduler invokes both methods in different places in the flow, this test is intended to make sure 
            those recipes that overrides 'visit()' method can be run correctly with the applicability test.
            
            All recipes in the list should not be run only when the applicability test fail.
            
            Given a text file containing the number "2", running this recipe should result with no change.
          applicability:
            singleSource:
              - org.openrewrite.text.FindAndReplace:
                  find: "1"
                  replace: "A"
          recipeList:
            - org.openrewrite.text.AppendToTextFile:
                  relativeFileName: "file.txt"
                  content: "->2"
                  preamble: "preamble"
                  appendNewline : false
                  existingFileStrategy: "continue"
            - org.openrewrite.text.AppendToTextFile:
                  relativeFileName: "file.txt"
                  content: "->3"
                  preamble: "preamble"
                  appendNewline : false
                  existingFileStrategy: "continue"
```

kunli2:
Fixes #2754 

This PR is a 2nd approach of PR #2759 (closed) to fix #2754, the 1st approach doesn't work for some cases, the reason is that approach tried to run recipes for each file in parallel and merge the results, but that breaks the nature of `visit(List<SourceFile> before)` is designed to handle bunch files at one time, so it breaks recipes like `AddDependency` or `ActivateStyle`.

### Root cause
For recipes in the list that override `visit(List<SourceFile> before, ExecutionContext ctx)` method, the `visit()` methods are invoked directly without the single-source applicability test.

### Solution
Keep the single source applicability test result for each `scheduleVisit()` run, and uses a list to present the source applicability test result for each source file,  and keep the mapping updated.


rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -308,10 +308,42 @@ default <S extends SourceFile> List<S> scheduleVisit(RecipeRunStats runStats,
...
+            if (hasSingleSourceApplicableTest) {
+                if (afterWidened.size() == after.size()) {
+                    // 'afterWidened' and 'after' are 1-to-1 mapping, reflects single source applicability test results.
+                    //noinspection unchecked
+                    afterWidened = ListUtils.map((List<SourceFile>) after, (i, s) ->
+                            lastSingleSourceApplicableTestResult.get(i) ? afterWidenedRef.get(i) : s);
+                } else if (afterWidened.size() > after.size()) {
+                    newSingleSourceApplicableTestResult.addAll(Collections.nCopies(afterWidened.size() - after.size(), true));
+                }
+            }
```
sambsnyd, line 345: As we have been discussing elsewhere, this does not look quite right since recipes can both add and remove source files. Need something like a `Map<UUID, Boolean>` to keep track of which source files have passed which applicability tests

kunli2: Thanks for the good point! Updated the code.

PR #2815, opened 2023-02-08
Fix `IsEmptyCallOnCollections` in for loop control

### LINKED ISSUE #2813: JavaTemplate error in IsEmptyCallOnCollections in ForLoop$Control.
common-static-analysis on ForgeEssentials/ForgeEssentials

```java
    @Test
    void forLoop() {
        rewriteRun(
          java(
            """
              import java.util.List;

              class Test {
                void method() {
                    List<String> lines = new ArrayList<>();
                    for (int i = lines.size() - 1; lines.size() > 0; i--) {
                        lines.remove(0);
                    }
                }
              }
              """
          )
        );
    }
```

knutwannheden:
Uses `JavaTemplate` to replace the `J.Binary` rather than the `J.MethodInvocation` as that is also what is the intention.

Fixes #2813

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
@@ -250,8 +250,14 @@ public class BlockStatementTemplateGenerator {
             before.insert(0, "assert ");
         } else if (j instanceof J.NewArray) {
             J.NewArray n = (J.NewArray) j;
-            before.insert(0, n.withInitializer(null).printTrimmed(cursor) + "{\n");
-            after.append("\n}");
+            if (n.getInitializer() != null && n.getInitializer().stream().anyMatch(arg -> referToSameElement(prior, arg))) {
+                before.insert(0, n.withInitializer(null).printTrimmed(cursor) + "{\n");
+                after.append("\n}");
+            } else {
+                // same as `Expression` case
+                before.insert(0, "__M__.any(");
+                after.append(");");
+            }
         } else if (j instanceof J.NewClass) {
             J.NewClass n = (J.NewClass) j;
             String newClassString;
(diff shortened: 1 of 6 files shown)


knutwannheden:
@traceyyoshima This fix unfortunately doesn't quite work. In the following test the type attribution for the `isEmpty()` call is missing. Is this a limitation of `JavaTemplate` or what am I doing wrong here? (I haven't tried to debug this myself yet.)

```java
    @Test
    void sizeOnClassImplementationCollection() {
        rewriteRun(
          java(
            """
              import java.util.ArrayList;
              class Test extends ArrayList<String> {
                  public boolean isZeroSize() {
                      return size() == 0;
                  }
              }
              """,
            """
              import java.util.ArrayList;
              class Test extends ArrayList<String> {
                  public boolean isZeroSize() {
                      return isEmpty();
                  }
              }
              """
          )
        );
    }
```

knutwannheden:
@sambsnyd @traceyyoshima I think I was able to solve the problem now. For "generic" expressions (i.e. those without dedicated support) `JavaTemplate` didn't use the `BlockStatementTemplateGenerator` mechanism to create a "skeleton" with all the visible declarations in it. As a result some references from the template could not be resolved and the returned result wasn't fully type-attributed.

I have changed this now by also using `BlockStatementTemplateGenerator` when expressions get substituted.

rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -250,8 +250,14 @@ private void template(Cursor cursor, J prior, StringBuilder before, StringBuilde
             before.insert(0, "assert ");
         } else if (j instanceof J.NewArray) {
```
knutwannheden, line 251: This case was really specific to the handling of expressions in an array's initializer. This is still required, but when replacing the `J.NewArray` as a whole, we use the same logic as for other expressions (i.e. using `__M__.any()`).

rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -406,16 +412,17 @@ private void template(Cursor cursor, J prior, StringBuilder before, StringBuilde
...
         } else if (j instanceof J.EnumValueSet) {
             after.append(";");
+        } else if (j instanceof Expression && j == prior) {
+            before.insert(0, "__M__.any(");
+            after.append(");");
         }
         template(next(cursor), j, before, after, templated);
     }
 
     private void insertControlWithBlock(J body, StringBuilder before, StringBuilder after, Runnable insertion) {
```
knutwannheden, line 422: Not sure why this never caused any issues, but since all the called `insertion` runnables do an `insert(0, ...)` I think the order was wrong here.

PR #2816, opened 2023-02-08
Also validate constructor calls in `RewriteTest` sources

knutwannheden:


rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java
```diff
@@ -872,7 +872,7 @@ class A {
                   Map<String, String> map = new HashMap<>();
                   Set<String> set = new HashSet<>();
                   List<String> test = Collections.singletonList("test");
-                  List<String> test2 = new ArrayList<>();
+                  List<String> test2 = new java.util.ArrayList<>();
```
knutwannheden, line 875: This maybe looks a bit weird, but since the recipe doesn't do any fixups of missing types (nor should it), this reference needs to be valid so that the source type checks.

sambsnyd:
Thanks @knutwannheden , good improvement

PR #2821, opened 2023-02-09
Slight performance improvements to Source Positions recipe

BoykoAlex:
Light performance improvements: ~15%

Was thinking about `includeMarkers` `boolean` option. If `false` won't attempt to visit `Markers` consequently no positions assigned to markers. Yields another 10%. However, compared to overall running time of the recipe felt like it is not worth it.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/UpdateSourcePositions.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/UpdateSourcePositions.java
@@ -97,7 +97,7 @@ public class UpdateSourcePositions extends Recipe {
 
                 Range range = positionMap.get(tree);
                 if (range != null) {
-                    J t = ((J) tree).withMarkers(((J) tree).getMarkers().add(range));
+                    J t = tree.withMarkers(tree.getMarkers().add(range));
                     return super.visit(t, ctx);
                 }
                 return super.visit(tree, ctx);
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/UpdateSourcePositions.java
```diff
@@ -135,7 +135,7 @@ public PrintOutputCapture<ExecutionContext> append(char c) {
             if (c == '\n') {
                 lineBoundary = true;
             }
-            return super.append(c);
+            return this;
```
sambsnyd, line 138: This looks like it would result in nothing being added to the buffer. Have you tested that this change works as intended?

BoykoAlex: This subclass of `PrintOutputCapture` doesn't really care about proper printing. Content of the `out` `StringBuilder` is of no importance as far as I remember. The `PositionPrintOutputCapture` is only to track current offset, line and column of an AST node. I have sanity tested this stuff manually. Markers in the IDE show up and locations seem correct. However, I have just noticed that `UpdateSourcePositionsTest` is disabled. Perhaps, I'll see if can be enabled again.

BoykoAlex:
@sambsnyd I have enabled the unit tests for positions recipe. I understand why it got disabled. Every time parsing of the white space changes or a new node introduce these tests start to fail. I've tried to simplify the tests in hope for them to be less annoying in the future: only test positions of identifier nodes.
I ran tests with previous version of the positions recipe and the current both passed.

sambsnyd:
Thanks @BoykoAlex

Issue #412, opened 2021-04-01
SourceFile implementations should include default SearchResultPrinters


sambsnyd:
Currently we have no straightforward way of printing SearchResult markers when running via our build plugins.
This means that search recipes run but produce no visible output.
As a first step to fixing this, we need to have a default SearchResultPrinter for each type of SourceFile. These default search result printers must print markers inside of a context-appropriate comment so that reifying the markers to code changes doesn't result in broken code. e.g.: `/* ~~> */` for Java and `<!-- ~~> -->` for XML

PR #2822, opened 2023-02-09
Issue #412 Added Provenance Wrapper

reka18:
This wrapper object serves as the Marker and container for GitProvenance and newly defined OsProvenance.


rewrite-core/src/main/java/org/openrewrite/marker/OsProvenance.java
```diff
@@ -13,50 +13,45 @@
...
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 import java.util.regex.Pattern;
 
 import static java.util.Collections.emptyList;
 
 /**
- * Detection logic from <a href="Gradle">https://github.com/gradle/gradle/blob/master/subprojects/base-services/src/main/java/org/gradle/internal/os/OperatingSystem.java</a>
+ * Detection logic from <a href="Gradle">https://github.com/gradle/gradle/blob/master/subprojects/base-services/src/main/java/org/gradle/internal/os/OsProvenance.java</a>
```
shanman190, line 31: @reka18, a little too much find a replace. 😄

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -39,8 +39,7 @@
 
 @Value
 @With
-public class GitProvenance implements Marker {
-    UUID id;
+public class GitProvenance {
```
sambsnyd, line 42: As I mentioned offline, this will cause existing ASTs which have `GitProvenance` markers on them to fail to deserialize. Could largely work around the problem by not removing the `Marker` interface here.

sambsnyd, line 42: As I mentioned offline, this will cause existing ASTs which have GitProvenance markers on them to fail to deserialize. Could largely work around the problem by not removing the Marker interface here.

PR #2825, opened 2023-02-09
First draft of JSON schema for OpenRewrite resources

### LINKED ISSUE #2823: Provide a JSON schema for OpenRewrite resource descriptors
A JSON schema would be useful as it offers code completion and basic validation for editors (including VS Code, JetBrains IDEA, etc.). By registering the schema by submitting a PR to https://github.com/SchemaStore/schemastore it should automatically be picked up by many of the supporting IDEs and editors.

For OpenRewrite the file path pattern would probably be `**/META-INF/rewrite/*.yml`.

knutwannheden:
This is a first draft of the JSON schema for OpenRewrite resource descriptors. Although the resource descriptor types themselves currently include a version "v1beta", this schema currently doesn't have any version number.

Fixes #2823

### DIFF
--- a/rewrite-core/openrewrite.json
+++ b/rewrite-core/openrewrite.json
@@ -1,0 +1,216 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "OpenRewrite Resource",
+  "description": "Contains OpenRewrite resource descriptors like recipes and styles",
+  "type": "object",
+  "properties": {
+    "type": {
+      "description": "Identifies the resource type",
+      "enum": [
+        "specs.openrewrite.org/v1beta/recipe",
+        "specs.openrewrite.org/v1beta/style",
+        "specs.openrewrite.org/v1beta/category",
+        "specs.openrewrite.org/v1beta/example"
+      ],
+      "default": "specs.openrewrite.org/v1beta/recipe"
+    }
+  },
+  "required": [
+    "type"
+  ],
+
+  "allOf": [
+    {
+      "if": {
+        "properties": {
+          "type": {
+            "const": "specs.openrewrite.org/v1beta/recipe"
+          }
+        }
+      },
+      "then": {
+        "properties": {
+          "type": {
+            "const": "specs.openrewrite.org/v1beta/recipe"
+          },
+          "name": {
+            "type": "string"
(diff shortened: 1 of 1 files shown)


knutwannheden:
A few notes:

- I didn't know where else to put the file, so it ended up in `doc` for now. Better ideas?
- I didn't include any version number in the schema file, as I am unsure what that would be and what the strategy is here. Most consistent would probably be something like "v1beta", but somehow I have the feeling that this is here to stay or at least remain supported for a long time (it could become a deprecated alias for a new name down the road).
- To test the schema (until registered with schemastore.org), any recipe can be opened in IDEA and then the schema button in the lower right corner of the IDE can be clicked and then select _New Schema Mapping..._ and enter something like in this dialog 
image

knutwannheden:
@sambsnyd An open question from my side: The `YamlResourceLoader` doesn't care if the YAML file contains any additional properties anywhere, so I didn't lock the schema down in this respect either, as clients may already be adding stuff to the files. But for the sake of schema evolution, I think it would make sense to not allow any additional properties.

sambsnyd:
@knutwannheden additional properties will mostly be ignored. Except for recipe configuration. I don't know much about these schemas. Can everything except recipe configuration be locked down?

knutwannheden:
Yes, we could lock down the other parts of the schema, so that user can't define arbitrary properties in other places.

Moving to rewrite-core seems reasonable so they don't get out of sync. But I can't think of any advantage of having it packaged in the jar.

I will look into these things. I don't have any prior with JSON schema, but having done my share of XSD, RelaxNG, and BNF it seems fairly straightforward.

Maybe we can integrate something bare and then ask @mike-solomon to look it over and augment it with reasonable descriptions, which editors would also display to the user?

knutwannheden:
@sambsnyd We could move the schema into `/rewrite-core/src/main/resources` and then it would be bundled. If we wanted to, we could probably use a JSON schema validator to validate the JSON we load and thus for example enforce that the resources don't declare any unexpected properties or just log warnings. I am unsure about this, however. What is your suggestion?

sambsnyd:
@knutwannheden json validation in a build step sounds like a good idea. For now I'm going to go ahead and merge this and we can iterate later

Issue #2818, opened 2023-02-08
UnnecessaryExplicitTypeArguments incorrectly removes types in `var producerConfig = Map.<String, Object>of(a, b)`


timtebeek:
[UnnecessaryExplicitTypeArguments](https://github.com/openrewrite/rewrite/blob/v7.35.0/rewrite-java/src/main/java/org/openrewrite/java/cleanup/UnnecessaryExplicitTypeArguments.java) incorrectly removes type arguments in [Connor.java](https://github.com/helpermethod/connor/blob/v1.4.1/src/main/java/com/github/helpermethod/connor/Connor.java).

## Steps to reproduce

1. `git clone git@github.com:helpermethod/connor.git`
2. `git checkout v1.4.1`
3. `mvn org.openrewrite.maven:rewrite-maven-plugin:4.39.0:run   -DactiveRecipes=org.openrewrite.java.cleanup.UnnecessaryExplicitTypeArguments`
4. `git diff`
5. `mvn compile`

## Expected

No change & successful compilation.

## Actual

```diff
diff --git a/src/main/java/com/github/helpermethod/connor/Connor.java b/src/main/java/com/github/helpermethod/connor/Connor.java
index 707603b..b193831 100644
--- a/src/main/java/com/github/helpermethod/connor/Connor.java
+++ b/src/main/java/com/github/helpermethod/connor/Connor.java
@@ -57,7 +57,7 @@ public class Connor implements Runnable {
     }
 
     private KafkaConsumer<String, byte[]> createConsumer() {
-        var consumerConfig = Map.<String, Object>of(
+        var consumerConfig = Map.of(
             ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
             bootstrapServers,
             ConsumerConfig.GROUP_ID_CONFIG,
@@ -72,7 +72,7 @@ public class Connor implements Runnable {
     }
 
     private KafkaProducer<String, byte[]> createProducer() {
-        var producerConfig = Map.<String, Object>of(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        var producerConfig = Map.of(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
 
         return new KafkaProducer<>(producerConfig, new StringSerializer(), new ByteArraySerializer());
     }
```
Leading to:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[...]

PR #2827, opened 2023-02-10
`UnnecessaryExplicitTypeArguments`: Don't apply to `var` variables

knutwannheden:
When the initializer of a `var` declared variable calls a method with type arguments, these arguments should not be removed, as otherwise the type cannot be deduced and the Java code doesn't compile.

Fixes: #2818

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryExplicitTypeArgumentsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryExplicitTypeArgumentsTest.java
@@ -129,4 +129,24 @@ class UnnecessaryExplicitTypeArgumentsTest implements RewriteTest {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/1211")
+    @Test
+    void assignedToVar() {
+        rewriteRun(
+          java(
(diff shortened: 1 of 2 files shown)


knutwannheden:
I suggest that we add a utility method to `J.VariableDeclarations` (and possibly also `J.VariableDeclaration.NamedVariable`) like `isVar()`. There is already at least one other recipe which could use this: https://github.com/openrewrite/rewrite/blob/683900a969afeab5d3c83e952193b25c195d0437/rewrite-java/src/main/java/org/openrewrite/java/cleanup/UseCollectionInterfaces.java#L155-L157

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryExplicitTypeArgumentsTest.java
```diff
@@ -129,4 +129,24 @@ void test() {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/1211")
```
timtebeek, line 133:
```suggestion
    @Issue("https://github.com/openrewrite/rewrite/issues/2818')
```

timtebeek (approved):
Great work, thanks for the quick fix!

PR #2828, opened 2023-02-10
Recipe/remove instanceof pattern

AresEkb:
The recipe is the opposite of https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/cleanup/InstanceOfPatternMatch.java

It removes instanceof pattern matching from code. For example, it converts
```
if (obj instanceof String str && str.isEmpty()) {
    System.out.println(str);
}
```
to
```
if (obj instanceof String && ((String) obj).isEmpty()) {
    String str = (String) obj;
    System.out.println(str);
}
```

It works in general, but I have the following problems. Maybe you could help to fix it?

At first, I can run tests in a simple separate project only. When I add them to this project and build it I get the following exception:
```
java.lang.NoClassDefFoundError: Could not initialize class org.openrewrite.java.Assertions
```
How could I fix it?

At second some of the tests are failed:

1) `noExtraParentheses` and `ternary` because of a minor formatting problem. I don't know how to remove extra parentheses and how to add a space.

2) `singleStatement` and `singleElseStatement` because I don't understand how to convert a single statement to a block statement in a `then` part of an `if`. For instance convert:
```
if (obj instanceof String str)
    System.out.println(str);
```
to
```
if (obj instanceof String str) {
    System.out.println(str);
}
```
and after that insert a variable declaration at the beginning of the block:
```
if (obj instanceof String str) {
    String str = (String) obj;
    System.out.println(str);
}
```
I get the following exception:
```
java.lang.IllegalStateException: Expected to find a matching parent for Cursor{Block}
```

3) `negationLocalVariable`. It is the biggest problem. I'm trying to convert
```
if (!(obj instanceof String str)) {
    Integer str = null;
} else {
    System.out.println(str);
}
```
to
```
if (!(obj instanceof String str)) {
    Integer str = null;
} else {
    String str = (String) obj;
    System.out.println(str);
}
```
To achieve it I need associate variable `str` in `System.out.println(str)` with its declaration in `obj instanceof String str`. Is there an utility method to do it? I didn't find any and implemented my own mechanism to find variable definitions. In `VariableUsageAnalyzer::currentScope` field I collect all variable definitions. It works in most of the cases, but doesn't work in that case, because `Integer str = null;` hides `str` variable in my scope. Maybe you could suggest a better mechanism?


knutwannheden (commented):
@AresEkb Looks like you got very far already!

> At first, I can run tests in a simple separate project only. When I add them to this project and build it I get the following exception:

To fix this move the test to `/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup` and it should work.

Here some more notes:
- To remove the extra parentheses you could try to apply the `UnnecessaryParenthesesVisitor` to the replacement that you produce.
- I added a comment to how to maintain the formatting of the space.
- For the negative case there is no utility that I am aware of, but this doesn't mean it doesn't exist, as I am also still learning.
- I also think that there are still a few details which aren't quite right yet in the LST for instanceof pattern variables, which would currently make it difficult to properly implement the negative case, so I suggest to wait with this feature. I will try to identify and fix those issues with the LST.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatch.java
```diff
@@ -156,6 +144,7 @@ public J.If visitIf(J.If iff, ExecutionContext executionContext) {
                                     J.Block.createEmptyBlock().withStatements(
                                             Collections.singletonList(elsePart.getBody())))),
                             executionContext);
+                    elsePart = result.getElsePart();
```
knutwannheden, line 147: The statement above replaces the else part and since the `elsePart` variable is dereferenced on line 153 again, it needs to be updated here.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatch.java
```diff
@@ -180,7 +169,7 @@ public J.If visitIf(J.If iff, ExecutionContext executionContext) {
         private J.Block addVariableDeclaration(J.Block block, J.InstanceOf instanceOf,
                                                ExecutionContext executionContext) {
             JavaTemplate template = JavaTemplate
-                    .builder(this::getCursor, "#{} #{} = (#{}) #{any()}")
+                    .builder(() -> new Cursor(getCursor(), block), "#{} #{} = (#{}) #{any()};")
```
knutwannheden, line 172: This was the problem here: The cursor didn't match the object being modified with the template. Possibly `JavaTemplate` could issue some kind of warning in these cases. But because of this `JavaTemplate` can then not properly derive the surrounding code required to do proper type attribution of the templated code.

knutwannheden (commented):
I made two minor adjustments which should get the tests to pass, except for the last one related to the negated `instanceof` expression.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatch.java
AresEkb: I tried to find such an example in the repository. I didn't know that it can be just created. Thanks for the fix! I've disabled the last test. It seems that everything works now!

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
```diff
@@ -0,0 +1,880 @@
...
+    void sameNamedVariables() {
+        rewriteRun(
+          version(
+            java(
+              """
+              package com.example;
+
+              class Example {
+                  public void test(Object obj) {
+                      if (obj instanceof String v && v.isEmpty() || obj instanceof Integer v && v.equals(1)) {
```
knutwannheden, line 490: This isn't actually legal Java code. We are working on optionally also validating the before sources, but aren't there quite yet. But here you essentially declare two variables with the same name in the same scope. So I think you can probably just remove this test.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
```diff
@@ -0,0 +1,880 @@
...
+                          System.out.println(str);
+                      }
+                  }
+              }
+              """),
+            14));
+    }
+
+    @Test
+    @Disabled("Not supported")
```
knutwannheden, line 443: There is one more case here which at some point may also have to be covered:
```java
    public void test(Object obj) {
        if (!(obj instanceof String str)) {
            throw new IllegalArgumentException();
        }
        System.out.println(str.length());
    }
```
I've never seen code like this in the wild, but it is actually legal.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatch.java
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * The implementation `instanceof` pattern match replacement.
+     */
+    private static class RemoveInstanceOfPatternMatchVisitor extends JavaVisitor<ExecutionContext> {
+
+        private VariableUsage variableUsage;
+
+        @Override
+        public J visitCompilationUnit(J.CompilationUnit cu, ExecutionContext executionContext) {
```
knutwannheden, line 68: The convention is to always name the `ExecutionContext` parameter `ctx`.

knutwannheden:
@AresEkb Thanks for your work on this. I think it looks really good! I added a few last minor details as review comments.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
AresEkb: I'm not 100% sure, but I think it's legal because two declarations are separated by a logical disjunction. So LHS and RHS of the disjunction forms separate scopes. And Eclipse and IDEA doesn't show syntax errors for such a code.

knutwannheden: That seems weird. The Javac 17 compiler doesn't accept it and while I can't see any error highlighted in the block string in IDEA, it does show me an error when I try this outside of a string.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatch.java
knutwannheden: Having to change code in other subtrees than the one being visited is probably also not that common. But there could indeed be more idiomatic ways to solve this problem.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
```diff
@@ -477,6 +477,40 @@ public void test(Object obj) {
             14));
     }
 
+    @Test
+    @Disabled("Not supported")
+    void negationWithoutElse() {
```
knutwannheden, line 482: Detail: It isn't so much the absence of the `else` which is special about this case, it is the `if` throwing an exception. There could also be an `else` which would also be allowed to dereference the `str` pattern variable.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
AresEkb: I added a test case and disabled it. I didn't know that it's correct to use a variable outside if/else. I think It can be implemented, but it requires some rework.

Actually the possibility to navigate from variable usage to its declaration would simplify the implementation. J.Identifier::getFieldType() contains a lot of useful information on a variable. The only missing thing is a reference to a declaring expression.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
AresEkb: I think exception throwing is not a problem for the current implementation. But variable dereference outside if/else is not supported at all. I can try to fix it in this pull request, but not sure that it will be simple with a current approach. Maybe it's better to create another pull-request.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
AresEkb: Yes, you are right. Eclipse doesn't show the error, but IDEA does. I removed the test case.

As Eclipse didn't show any errors I thought it's a correct construction and even found an explanation for myself why it works.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/RemoveInstanceOfPatternMatchTest.java
knutwannheden: I agree. I would also suggest to defer the two disabled tests for a later PR.

knutwannheden (approved):

AresEkb:
Oh, I've tested it on a big code base and found a significant bug. Need some time to fix it.

knutwannheden:
Thanks for testing. What is the bug about?

AresEkb:
The recipe generates multiple variable declarations for each variable usage. But should generate at most one declaration for each `instanceof`. It's easy to fix.

The second bug is that the recipe doesn't handle well complex nested "if/else". So I'm trying to improve variable usage analyzes.

AresEkb:
I fixed the bugs

PR #2831, opened 2023-02-10
Add missing field type to `instanceof` pattern variables

knutwannheden:


sambsnyd:
Ah, good catch. Thanks for fixing this oversight.

PR #2836, opened 2023-02-11
Add first simple ArchUnit tests

knutwannheden:
ArchUnit can be used to enforce certain architectural constraints. Typically, these get checked by the build system by executing them as tests (using JUnit).

This first set of tests only applies to `rewrite-java` and only checks that all Java packages have a `package-info.java` including a `@NonNullApi` annotation and that the JetBrains nullness annotations are not used.

### DIFF
--- a/rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
@@ -17,36 +17,31 @@ package org.openrewrite.java;
 
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.domain.JavaPackage;
-import com.tngtech.archunit.core.importer.ClassFileImporter;
 import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.junit.AnalyzeClasses;
+import com.tngtech.archunit.junit.ArchTest;
 import com.tngtech.archunit.lang.*;
-import org.junit.jupiter.api.Test;
 import org.openrewrite.internal.lang.NonNullApi;
 
 import static com.tngtech.archunit.base.DescribedPredicate.doNot;
 import static com.tngtech.archunit.core.domain.properties.HasName.Predicates.name;
-import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.*;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.all;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noCodeUnits;
 
+@AnalyzeClasses(packages = "org.openrewrite.java", importOptions = ImportOption.DoNotIncludeTests.class)
 public class OpenRewriteArchitectureTest {
-    JavaClasses importedClasses = new ClassFileImporter()
-      .withImportOption(new ImportOption.DoNotIncludeTests())
-      .importPackages("org.openrewrite.java");
 
-    @Test
-    void packageInfo() {
+    @ArchTest
+    public static final ArchRule packageInfo =
         all(packages("org.openrewrite.java"))
           .that(doNot(name("org.openrewrite.java.internal.grammar")))
-          .should(containAnAnnotatedPackageInfo())
-          .check(importedClasses);
-    }
+          .should(containAnAnnotatedPackageInfo());
 
-    @Test
-    void nullness() {
+    @ArchTest
+    public static final ArchRule nullness =
         noCodeUnits()
           .should().beAnnotatedWith("org.jetbrains.annotations.NotNull")
-          .orShould().beAnnotatedWith("org.jetbrains.annotations.Nullable")
-          .check(importedClasses);
-    }
+          .orShould().beAnnotatedWith("org.jetbrains.annotations.Nullable");
 
     private static ClassesTransformer<JavaPackage> packages(String basePackage) {
         return new AbstractClassesTransformer<>("packages") {
--- a/rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
@@ -69,12 +69,10 @@ public class OpenRewriteArchitectureTest {
         }.and(new ArchCondition<>("be annotated with @NonNullApi") {
             @Override
             public void check(JavaPackage javaPackage, ConditionEvents events) {
-                if (javaPackage.tryGetPackageInfo().isPresent()) {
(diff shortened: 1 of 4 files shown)


knutwannheden:
If we decide to go forward with this, I propose that some central rules get moved to `rewrite-test` so that they can be reused by all modules.

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
```diff
@@ -0,0 +1,81 @@
...
+import static com.tngtech.archunit.base.DescribedPredicate.doNot;
+import static com.tngtech.archunit.core.domain.properties.HasName.Predicates.name;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.*;
+
+public class OpenRewriteArchitectureTest {
+    JavaClasses importedClasses = new ClassFileImporter()
+      .withImportOption(new ImportOption.DoNotIncludeTests())
+      .importPackages("org.openrewrite.java");
+
+    @Test
```
timtebeek, line 35: Slight personal preference for `@ArchTest`, as it remove the need to call `check`, which can accidentally be left out otherwise.

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openrewrite.internal.lang.NonNullApi;
+
+import static com.tngtech.archunit.base.DescribedPredicate.doNot;
+import static com.tngtech.archunit.core.domain.properties.HasName.Predicates.name;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.*;
+
+public class OpenRewriteArchitectureTest {
+    JavaClasses importedClasses = new ClassFileImporter()
+      .withImportOption(new ImportOption.DoNotIncludeTests())
+      .importPackages("org.openrewrite.java");
```
timtebeek, line 33: Slight personal preference for [`@AnalyzeClasses`](https://www.archunit.org/userguide/html/000_Index.html#_using_junit_4_or_junit_5), as you can optionally leave out the `importPackages` argument, and it's a little more expressive. Any reason you went for `ClassFileImporter`?

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
```diff
@@ -0,0 +1,81 @@
...
+          .should().beAnnotatedWith("org.jetbrains.annotations.NotNull")
+          .orShould().beAnnotatedWith("org.jetbrains.annotations.Nullable")
+          .check(importedClasses);
+    }
+
+    private static ClassesTransformer<JavaPackage> packages(String basePackage) {
+        return new AbstractClassesTransformer<>("packages") {
+            @Override
+            public Iterable<JavaPackage> doTransform(JavaClasses classes) {
+                return classes.getPackage(basePackage).getSubpackagesInTree();
```
timtebeek, line 55: Didn't know about this one; neat!

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }.and(new ArchCondition<>("be annotated with @NonNullApi") {
+            @Override
+            public void check(JavaPackage javaPackage, ConditionEvents events) {
+                if (javaPackage.tryGetPackageInfo().isPresent()) {
+                    if (!javaPackage.getPackageInfo().isAnnotatedWith(NonNullApi.class)) {
+                        String message = String.format("Package '%s' is not annotated as @NonNullApi", javaPackage.getName());
+                        events.add(SimpleConditionEvent.violated(javaPackage, message));
+                    }
+                }
```
timtebeek, line 77: `isPresent()` within an if statement always looks a bit odd to me, but maybe that's personal.
Had you considered something like
```suggestion
                javaPackage.tryGetPackageInfo()
                        .filter(packageInfo -> !packageInfo.isAnnotatedWith(NonNullApi.class))
                        .ifPresent(packageInfo -> events.add(SimpleConditionEvent.violated(javaPackage,
                                String.format("Package '%s' is not annotated as @NonNullApi", javaPackage.getName()))));
```

timtebeek (approved):
Nice! Looks good to me. Couple questions or suggestions, but very much optional; can leave things as they are as well.

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
knutwannheden: Yes, makes sense. I will rewrite it.

rewrite-java/src/test/java/org/openrewrite/java/OpenRewriteArchitectureTest.java
knutwannheden: Used to be called `getAllSubpackages()` pre 1.0.0.

timtebeek (approved):
Great addition and start to testing with ArchUnit. If it were up to me we indeed pull this into rewrite-test, so we can be sure all modules are tested. But I guess both of us are already fans of ArchUnit, so I'll let the judgement on that & merge to a neutral third.

knutwannheden:
> Great addition and start to testing with ArchUnit. If it were up to me we indeed pull this into rewrite-test, so we can be sure all modules are tested. But I guess both of us are already fans of ArchUnit, so I'll let the judgement on that & merge to a neutral third.

@timtebeek Agreed. I suppose we would first have to find the subset of rules we would like to share via `rewrite-test`. This was mostly meant as a proof-of-concept.

I do have another rule ready which would check that any visitor with the annotation `@CompleteVisitor` (with `CLASS` retention) overrides all LST element visitor methods. The idea would be that one doesn't forget to extend these visitors when new LST element types are introduced (typically when adding support for a new language version). Applying this rule to `SemanticallyEqualVisitor` I found that it lacks an override for `visitYield()`.

It would be interesting to see what other useful rules we can come up with.

sambsnyd:
Thanks @knutwannheden ! I've almost added arch unit a couple times and never quite got around to it.

Another test that would be useful is that no class extending `Recipe` has a public field that isn't a primitive (or enum) type. Recipes which accept complex types as inputs would work fine from yaml, but be a real challenge for the saas UI. Just a wishlist item I've had for a while.

knutwannheden:
@sambsnyd Thanks for the input. I have a few more rules myself which I've implemented, including that complete visitor (where I am a bit unsure about the name) and a few checks for LST types (final flag, all final or transient fields). I will submit another PR for that. And like @timtebeek suggests, I will look into moving some parts into `rewrite-test` so the general rules can be reused by other modules.

Should the package info have more rules? I think many packages have more nullness annotations, for instance.

PR #2839, opened 2023-02-13
Support `instanceof` pattern variables in control and data flow analysis

knutwannheden:
For the data flow analysis the `instanceof` pattern variable must also be checked.


knutwannheden:
Note: This is still work in progress.

JLLeitschuh:
Oh, very interesting. This is new syntax I haven't played with yet. Thanks for the contribution here though! 🎉

knutwannheden:
I still have to learn the ins and outs of the code in this area and I am sure this is very much incomplete, but the flow scoping is rather tricky. It gets really interesting when the condition in the if (or ternary expression) is negated, as then the pattern variable is visible in the else or even to successor statements, if the if body terminates the execution in the current scope (using return, break or continue).

JLLeitschuh:
Wow, yea, this is an annoying operator from a data-flow perspective. 😬 

https://docs.oracle.com/en/java/javase/14/language/pattern-matching-instanceof-operator.html

I wrote 99% of data flow and control flow. Don't hesitate to ping me on slack if you have any issues or need to rubber-duck debug this

PR #2840, opened 2023-02-14
Add recipe to replace records with classes

AresEkb:
Maybe the recipe will be useful for someone else.

Could you please help to fix the following things?

1. I don't understand how to generate non-static fields. If I remove `static` keyword from the template then fields aren't generated at all:
```
        private final JavaTemplate fieldTemplate = JavaTemplate
                .builder(this::getCursor, "private static final #{} #{};").build();
```

2. Don't understand how to embed templates into templates. For instance, there are templates for a constructor and for field assignment:
```
        private final JavaTemplate constructorTemplate = JavaTemplate
                .builder(this::getCursor, "public #{}() { }").build();

        private final JavaTemplate assignmentTemplate = JavaTemplate
                .builder(this::getCursor, "this.#{} = #{};").build();
```
How to generate field assignments inside the constructor?

3. Is it ok to generate fragments of a code as plain strings? For example, I generate a comma-spearated string of field names here:
```
            result = result.withBody(result.getBody()
                    .withTemplate(hashCodeTemplate,
                            result.getBody().getCoordinates().lastStatement(),
                            fields.stream()
                                    .map(field -> field.getVariables().get(0).getSimpleName())
                                    .collect(Collectors.joining(", "))));
```


kunli2:
Seems this is part of Java's downgrading from above 14 to lower. I think we do code modernization by running recipes to upgrade the java version. not sure about this downgrading.

AresEkb:
I see two usage scenarios for the recipe:

1. Yes, it's downgrading code from Java 14. It's important to us because we have to use Java 11 in our projects. But some of the libraries we use are migrated to Java 17. So we have to downgrade those libraries. Java 11 and even Java 8 are still supported. So, it may be useful not only to us, maybe for somebody else as well.
2. Records could be transformed to classes not only to downgrade a code. But also because a record is enough, we need to make it more complicated, add class inheritance and so on. The first step is to convert the record to a class and then add all those complications.

AresEkb:
@timtebeek I improved the recipe. However it seems that similar recipes are moved to https://github.com/openrewrite/rewrite-static-analysis Should I create a new pull-request and close this one?

timtebeek:
Impressive work here @AresEkb ! I like how thorough you are with the tests you've added to rule out issues with various edge cases, some of which I wasn't even aware could be handled in older Java versions with local classes. I presume you're already using this on your own projects, given the level of detail here?

I agree with you and Kun that this recipe would not fit into openrewrite/rewrite as is; the one module that _could_ make sense is [rewrite-migrate-java](https://github.com/openrewrite/rewrite-migrate-java), but that's geared towards Java upgrades, not downgrades at present. I'm hesitant to take on downgrades as well, as this recipe is already quite some code to support & maintain, and would require a further recipes to downgrade text blocks, and pretty much anything else added between 11 and 17 as per https://javaalmanac.io/jdk/17/apidiff/11/

I understand your use case of (presumably) downgrading projects that you depend on through forks, but I'm not sure that's common enough for us to make it available broadly. The second use case you listed of expanding a single existing record into  a class would suffer from the recipe (as-is) not being selective in when to apply, such that it would convert every record that it comes across, which might not be desirable either.

Would it be an option to perhaps start with separate module that you maintain under https://github.com/AresEkb, based off of the [rewrite-recipe-starter](https://github.com/moderneinc/rewrite-recipe-starter) template? That could still help package and distribute your recipe to others, and I could in a personal capacity try to help you with any issues there. 

I do want to say again that we're very appreciative of the work that you have put in; it's not often that we see work of this quality proposed for addition into OpenRewrite. My only reasons to doubt a merge are because by itself it's not yet a full Java 17 to Java 11 downgrade, and we're hesitant to take on the work required to fully support that, given that we tend to recommend and automate upgrades instead. It certainly has nothing to do with the quality, which makes it a tough decision.

AresEkb:
Thanks for the answer!

Yes, we use the recipe on our own project. However, we don't need most of the cases (local records, custom getters in records, ...). I've just implemented them for completeness :)

I published the recipe here https://github.com/AresEkb/rewrite-recipe-record-to-class It's pretty ok for me.

However, the problem is that it doesn't work, tests not passed. The main problems are as follows:

1) The recipe generates word "Template" in constructor declaration after the public modifier:
```java
      public Template Vehicle(String model, int power) {
          this.model = model;
          this.power = power;
      }
```

2) The recipe throws the exception: `java.lang.IllegalStateException: AST contains missing or invalid type information` diring generation of `toString()` method:
```java
     public String toString() {
          return "Vehicle[model=" + model + ", power=" + power + "]";
     }
```
But if I change the template so field names are enclosed in parentheses, then there is no any exceptions:
```java
     public String toString() {
          return "Vehicle[model=" + (model) + ", power=" + (power) + "]";
     }
```

May I ask for suggestion how to fix with these errors?

I marked problematic code with TODO:

1) https://github.com/AresEkb/rewrite-recipe-record-to-class/blob/5fc4e819d8b629626919b6e0b228ff43dadbd1c1/src/main/java/com/advalange/coderewriter/ReplaceRecordWithClass.java#L195

2) https://github.com/AresEkb/rewrite-recipe-record-to-class/blob/5fc4e819d8b629626919b6e0b228ff43dadbd1c1/src/main/java/com/advalange/coderewriter/ReplaceRecordWithClass.java#L433

Is it a suitable place to ask for suggestions on my own recipes or it's better to use another place?

timtebeek:
I'm having a look right now, but not having much luck yet with debugging your recipe. Perhaps best to move the above items into a GitHub issue on your repository and then chime into our Slack for support? @joanvr would then pick up from there; tagged here for visibility already.

Generally speaking, while looking at your implementation, I see that you're generating and inserting multiple smaller JavaTemplate snippets, each time replacing the result with the response of an applied JavaTemplate. I'd have instead maybe expected a single JavaTemplate that fully replaces the record, and contains all the components in a single template with appropriate parameters. Not sure if that would resolve the issue, but it's something to consider.

Another thing that struck me as different from what I've seen before, is that you in the `constructorTemplate` apply method you pass in formatted Strings as arguments, whereas I would have not expected those as parameters, but instead part of the template String, with the apply parameters being LST elements instead. Again: high over feedback mostly pointing out differences rather than a concrete solution, but perhaps it's of some use to you yet.

timtebeek:
Closing the issue here, as the discussion is probably best moved to https://github.com/AresEkb/rewrite-recipe-record-to-class. We don't plan to support downgrades from newer to older Java versions in OpenRewrite, given the complexity to do that perfectly, and instead recommend folks upgrade to the latest versions instead.

You can tag us in an issue on AresEkb/rewrite-recipe-record-to-class, or ask in Slack if there's still outstanding issues that you'd like help with. We much appreciate you using OpenRewrite and writing recipes, not all of which make it in to OpenRewrite itself. The library that you provide from your own account repository can then still prove valuable to others, and we thank you for that!

Issue #2742, opened 2023-01-27
Convert multiline string concatenation to text block


murdos:
Example : 
```java
        @PreAuthorize("hasRole(T(my.company.Constants).ROLE_BASIC_AUTH) "
        + "or @someValidator.validate(#searchCriteria.foo)")
```

to
```java
    @PreAuthorize("""
        hasRole(T(my.company.Constants).ROLE_BASIC_AUTH)
        or @someValidator.validate(#searchCriteria.foo)
    """)
```

knutwannheden:
Fixed by https://github.com/openrewrite/rewrite-migrate-java/commit/b6df90842441e92bbd1c222baa40df9d5c269b9e.

PR #2841, opened 2023-02-14
Add recipe to replace text blocks with strings

AresEkb:
Maybe such a recipe will be useful, if it doesn't already exist.


timtebeek:
Sympathize with you there having to downgrade and appreciate the recipes!

Just to be sure we've covered all cases, would you mind adding a test with a comment just before the text block? Want to make sure we're not accidentally removing those comments with the current implementation, or any changes after this is merged.

timtebeek:
In either case these tests are going to come in handy (in reverse) when we implement #2742, and hopefully for you in the future!

AresEkb:
I agree that comments should not be removed. Added tests and fixed the recipe.

sambsnyd:
Thanks for contributing all of these back, @AresEkb !

PR #2842, opened 2023-02-14
Add recipe to replace Stream.toList() with Stream.collect()

AresEkb:
We use the recipe to downgrade Java 17 code to Java 11. I'm not sure whether it could be useful for somebody else.

Anyway, could you please help to fix `preservesFormatting` test? :) How to preserve spacing within method invocation after template application?

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/ReplaceStreamToListWithCollectTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/ReplaceStreamToListWithCollectTest.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


AresEkb:
Thanks for the idea! It seems that not just formatting was broken, but also comments after `select` were removed.
```
Stream.of()
    // Convert to list
    .toList();
```
transformed to
```
Stream.of().collect(Collectors.toList());
```
instead of
```
Stream.of()
    // Convert to list
    .collect(Collectors.toList());
```

I tried to save `select`, but the problem is that `J.MethodInvocation::getSelect()` is implemented as follows:
```
        @Nullable
        JRightPadded<Expression> select;

        @Nullable
        public Expression getSelect() {
            return select == null ? null : select.getElement();
        }
```
The `JRightPadded<Expression> select` contains comments and spaces. But I guess that I don't have access to them because `select.getElement()` is returned, I can't get `select.getAfter()`. So I changed the implementation as follows:
```
J.MethodInvocation newMethod = result
        .withTemplate(template, result.getCoordinates().replace(), result.getSelect());
result = result
        .withName(newMethod.getName())
        .withMethodType(newMethod.getMethodType())
        .withArguments(newMethod.getArguments());
```
At first generate the method. And then copy its name, type and arguments to the original expression.

All tests are fixed now. The recipe is very useful for us, because it allows one to downgrade a code to Java 11. I'm not sure that it will be useful for others. In that case you are welcome just to close it :)

sambsnyd:
@AresEkb try saving `J.MethodInvocation.getPadding().getSelect()` off to the side... that will be the whole `JRightPadded<Expression>` including any whitespace and comments between the select and the rest of the invocation. And use `J.MethodInvocation.getPadding().withSelect()` to perform the replacement.

Sorry, I should have suggested that from the beginning, but wasn't thinking about exactly where all of the relevant whitespace is stored

AresEkb:
Thanks, it works!

sambsnyd:
Thanks for this contribution @AresEkb !

Issue #2847, opened 2023-02-15
rewriteRun treats all xml as pom if pomXml is not last


kmccarp:
`rewriteRun` tries to parse both of the following xml files as poms if pomXml is not last. I would expect order not to matter since i've defined one as `xml` and the other as `pomXml`.

```    @Test
    void addsSuppressionIfNotExists() {
        rewriteRun(
          pomXml(
            //language=xml
            """
              <project>
                  <groupId>org.openrewrite</groupId>
                  <artifactId>test</artifactId>
                  <version>1.0.0</version>
                  <dependencies>
                      <dependency>
                          <groupId>com.fasterxml.jackson</groupId>
                          <artifactId>jackson-base</artifactId>
                          <version>2.14.2</version>
                      </dependency>
                  </dependencies>
              </project>"""
          ),
          xml(
            """
              <?xml version="1.0" encoding="UTF-8" ?>
              <suppressions xmlns="<https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd>">
              </suppressions>""",
            """
              <?xml version="1.0" encoding="UTF-8" ?>
              <suppressions xmlns="<https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd>">
                  <suppress>
                      <notes><![CDATA[
                          file name: com/fasterxml/jackson/core/JsonFactory.java
                          ]]></notes>
                      <gav regex="true">^com\\\\.fasterxml\\\\.jackson\\\\.core:.*$</gav>
                      <cve>CVE-2016-3720</cve>
                  </suppress>
              </suppressions>""",
            spec -> spec.path("suppressions.xml")
          )
        );
    }```

[Slack Message](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1676422968250399)

kmccarp:
If `pomXml` is first, both are `Pom`. If `pomXml` is last, both are `Xml$Document`.

kmccarp:
[...]

PR #2852, opened 2023-02-16
Group sourceSpecs by parser class

kmccarp:
I believe this fixes https://github.com/openrewrite/rewrite/issues/2847

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -551,7 +551,7 @@ public interface RewriteTest extends SourceSpecs {
 class RewriteTestUtils {
     static boolean groupSourceSpecsByParser(List<Parser.Builder> parserBuilders, Map<Parser.Builder, List<SourceSpec<?>>> sourceSpecsByParser, SourceSpec<?> sourceSpec) {
         for (Map.Entry<Parser.Builder, List<SourceSpec<?>>> entry : sourceSpecsByParser.entrySet()) {
-            if (entry.getKey().getSourceFileType().equals(sourceSpec.sourceFileType)) {
+            if (entry.getKey().getSourceFileType().equals(sourceSpec.sourceFileType) && sourceSpec.getParser().getClass().isAssignableFrom(entry.getKey().getClass())) {
                 entry.getValue().add(sourceSpec);
                 return true;
             }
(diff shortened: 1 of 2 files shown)


rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -551,7 +551,7 @@ default void validateRecipeOptions(Recipe recipe) {
 class RewriteTestUtils {
     static boolean groupSourceSpecsByParser(List<Parser.Builder> parserBuilders, Map<Parser.Builder, List<SourceSpec<?>>> sourceSpecsByParser, SourceSpec<?> sourceSpec) {
         for (Map.Entry<Parser.Builder, List<SourceSpec<?>>> entry : sourceSpecsByParser.entrySet()) {
-            if (entry.getKey().getSourceFileType().equals(sourceSpec.sourceFileType)) {
+            if (entry.getKey().getSourceFileType().equals(sourceSpec.sourceFileType) && sourceSpec.getParser().getClass().isAssignableFrom(entry.getKey().getClass())) {
```
sambsnyd, line 554: The reason the bug can happen in the first place is that both `MavenParser.accept()` and `XmlParser.accept()` will return `true` for a file with an xml extension. And both produce `Xml.Document`... the only difference in the result is that the maven parser also attaches our maven marker with dependency resolution info as metadata.

The test framework will stop trying to match sources to parsers once it finds a parser that will accept the test spec. It isn't clear to me looking at this that it would address that issue.

kmccarp: There is a `Parser$Builder` on the `SourceSpec`, the `Parser$Builder` for the `pomXml` is a `MavenParser$Builder` while the `Parser$Builder` for the `xml` is a `XmlParser$Builder`. I don't know how to test this within RewriteTestTest (because it relies on rewrite-maven which is not on the classpath), but in my recipe in rewrite-java-security, I verified this change will use the correct Parser for both the `xml` and the `pomXml`.

kmccarp: I could put a test in rewrite-maven ... let me try that

kmccarp: https://github.com/openrewrite/rewrite/pull/2852/files#diff-de2491affc07fd204feb64109a63b15e34034a5c40490293d1c712e11bd2b9a5R20

Verified that without my change to `RewriteTest`, I get an error. With my change to `RewriteTest`, the test succeeds.

PR #2853, opened 2023-02-16
Added Test Cases for ReplaceApacheCommonsLang3ValidateNotNullWithObjectsRequireNonNull

srmalkan:
I have added some test cases for ReplaceApacheCommonsLang3ValidateNotNullWithObjectsRequireNonNull Recipe, but unfortunately, all are failing. could someone look at it? or Let me know if there is something missing in my test cases.

Related 

- https://github.com/openrewrite/rewrite/issues/2724
- https://github.com/openrewrite/rewrite/pull/2803


timtebeek:
Hi @srmalkan ! Thanks for adding these tests to highlight the issue with the current implementation. Your tests weren't running because the commons-lang3 dependency was not on the test classpath, which I've fixed with https://github.com/openrewrite/rewrite/pull/2853/commits/0f2dfc2cea952009a946972594caa038212fa2d7 .

Now that the tests run they do indeed highlight the issue: the matching with any amount of arguments `(..)` is too broad, and the type replacement isn't happening I believe.

If you're up for it you might want to try again using a method pattern with an explicit one `(*)` or two `(*,*)` arguments, and potentially use `ChangeMethodTargetToStatic`; there's an [example in rewrite-testing-frameworks](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/resources/META-INF/rewrite/assertj.yml#L39), which I believe to be similar to your use case.

```yaml
  - org.openrewrite.java.ChangeMethodTargetToStatic:
      methodPattern: "org.assertj.core.api.AssertionsForClassTypes assertThat(..)"
      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
```

If you just want to see this fixed I can make that change as well, but it looked like you enjoy learning more and potentially adding a few more such recipes, so again: I don't want to take the fun out of that. Let me know how you do!

srmalkan:
Hi Team,

I have made changes to limit the scope of the recipe to only a single argument method i.e Validate.notnull(object).
I have tried different permutations with wildcards, explicitly mentioning data type, but no luck. Seems there is some issue with either how ChangeMethodTargetToStatic handles static methods especially static methods referenced using class or method pattern.

I have explained the issue here #2859

timtebeek:
hi @srmalkan Thanks for all the effort you've put into this; sorry to hear it's been a struggle to get working so far. I've had another look and it appears to work when you change the order of the recipes, [as seen here as well](https://github.com/openrewrite/rewrite-migrate-java/blob/e9aaf04dd5a90509baee77e9cc215a2b084a09b4/src/main/resources/META-INF/rewrite/no-guava.yml#L104). I've pushed https://github.com/openrewrite/rewrite/pull/2853/commits/b2faf5ec971427e56061d869c9fe20ccad44d1e8 to hopefully see this through successfully, and will try to follow that up with a two argument replacement as well.

I hope this hasn't put you off from building more migration recipes! The first one can be a bit of a struggle, but it gets easier after that, especially once you get to run your own recipes to do replacements in bulk. For what it's worth we're working on an easier way to define recipes such as this one over in [rewrite-templating](https://github.com/openrewrite/rewrite-templating). It's not quite ready to use yet, but once it is we'll let you know!

timtebeek:
Let's pick up the two argument support separately, as I'm failing to pick that up with a method pattern for now. Thanks for this & adding the tests! Should help support the two and three argument variants in the hopefully near future!

PR #2856, opened 2023-02-16
Add `IsLikelyTest` and `IsLikelyNotTest` Recipes

JLLeitschuh:
Enables applicability tests to be applied exclusively to, or excluding
test source.

Resolves https://github.com/openrewrite/rewrite/discussions/2849
Closes #2850

This enables the following which is exactly what is needed for security fixing vulnerabilities at-scale across OSS:
```yaml
---
name: com.example.FilteringTestSources
displayName: Filtering Test Sources Test
description: Only apply a `FindMethods` recipe to all sources when non-test sources are modified.
applicability:
  anySource:
  - org.openrewrite.java.search.IsLikelyNotTest
  - org.openrewrite.java.search.FindMethods: # Stand in for the security fix recipe to be executed
      methodPattern: "java.io.PrintStream println(..)"
recipeList:
  - org.openrewrite.java.search.FindMethods: # Stand in for the security fix recipe to be executed duplicate config from above
      methodPattern: "java.io.PrintStream println(..)"
```

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/IsLikelyTest.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/IsLikelyTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.search;
+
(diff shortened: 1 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/search/IsLikelyTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private static class HasSourceSetIntegrationTestVisitor<P> extends JavaIsoVisitor<P> {
+        private static final Pattern INTEGRATION_TEST_PATTERN =
+                Pattern.compile("integ.*test", Pattern.CASE_INSENSITIVE);
+
+        @Override
+        public JavaSourceFile visitJavaSourceFile(JavaSourceFile cu, P p) {
+            if (cu.getMarkers().findFirst(JavaSourceSet.class)
+                    .filter(s -> INTEGRATION_TEST_PATTERN.matcher(s.getName()).matches())
```
shanman190, line 62: This may be better to check for source sets that **contain** "test". 

In Gradle, they make it really easy to create additional test suites, but the convention is that it should be <*suite*>Test. (ex: compatibilityTest, functionalTest, integrationTest, etc).

JLLeitschuh: Good idea, updated. 👍

PR #2860, opened 2023-02-17
Cleanup `RecipeScheduler` and associated logic

JLLeitschuh:


rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -111,11 +123,29 @@ default RecipeRun scheduleRun(Recipe recipe,
...
+                .withResults(results)
+                .withDataTables(ctx.getMessage(ExecutionContext.DATA_TABLES, emptyMap()));
+    }
+
+    static List<Result> createAndProcessResults(
+            List<? extends SourceFile> before,
+            List<? extends SourceFile> after,
+            ExecutionContext ctx,
+            Map<UUID, Stack<Recipe>> recipeThatAddedOrDeletedSourceFile
+    ) {
```
JLLeitschuh, line 144: This should theoretically be added to the `RecipeSchedulerUtils` but I wanted to simplify the review of this diff

rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -265,7 +302,7 @@ default <S extends SourceFile> List<S> scheduleVisit(RecipeRunStats runStats,
                     //noinspection unchecked
                     afterFile = (S) visitor.visitSourceFile(s, ctx);
 
-                    if (visitor.isAcceptable(s, ctx)) {
+                    if (afterFile != null && visitor.isAcceptable(afterFile, ctx)) {
```
JLLeitschuh, line 305: Bug fix from discussion with @jkschneider 

https://linuxfoundation.slack.com/archives/C04HR6EJ38D/p1676648506919089

kunli2 (approved):

kunli2:
Thanks for cleaning up!

rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -35,41 +35,46 @@
...
 import static org.openrewrite.RecipeSchedulerUtils.handleUncaughtException;
-import static org.openrewrite.Tree.randomId;
 
+/**
+ * The scheduler is responsible for executing a {@link Recipe} full lifecycle and
+ * reporting a {@link RecipeRun} result.
+ */
 public interface RecipeScheduler {
-    default <T> List<T> mapAsync(List<T> input, BiFunction<T, Integer, T> mapFn) {
+    default <T> List<T> mapAsync(List<T> input, UnaryOperator<T> mapFn) {
```
sambsnyd, line 52: Changing this signature is a breaking change for our saas. I would rather not spring it on them without good reason. I am fine with most of the changes in this PR but don't want to change the public method signatures right now

PR #2861, opened 2023-02-17
Actually support `Recipe` in `addApplicableTest` & `addSingleSourceApplicableTest`

JLLeitschuh:
Recipes used in applicability tests will have their own applicability tests invoked now.

This means that expensive recipes with light `singleSourceApplicableTest` can be used without incurring a performance overhead of invoking the recipe visitor every time.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
+++ b/rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
@@ -123,7 +123,7 @@ public interface RecipeScheduler {
             return recipeRun.withDataTables(ctx.getMessage(ExecutionContext.DATA_TABLES, emptyMap()));
         }
 
-        List<Result> results = createAndProcessResults(
+        List<Result> results = RecipeSchedulerUtils.createAndProcessResults(
                 before,
                 after,
                 ctx,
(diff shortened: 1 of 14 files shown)


rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -136,76 +136,6 @@ default RecipeRun scheduleRun(
                 .withDataTables(ctx.getMessage(ExecutionContext.DATA_TABLES, emptyMap()));
     }
 
-    static List<Result> createAndProcessResults(
-            List<? extends SourceFile> before,
-            List<? extends SourceFile> after,
-            ExecutionContext ctx,
-            Map<UUID, Stack<Recipe>> recipeThatAddedOrDeletedSourceFile
-    ) {
```
JLLeitschuh, line 144: Moved into `RecipeSchedulerUtil`

rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -519,20 +527,47 @@ public static <S extends SourceFile> List<S> handleUncaughtException(
...
+            // Re-surface any errors generated applying the applicability tests up to the top level
+            for (S newS : next) {
+                newS.getMarkers().findFirst(Markup.Error.class).ifPresent(m -> {
+                    if (m.getException() instanceof RecipeRunException) {
+                        throw (RecipeRunException) m.getException();
+                    } else {
+                        throw new RuntimeException("Applicable Test Failed", m.getException());
+                    }
+                });
+            }
```
JLLeitschuh, line 570: @jkschneider is there a better/more appropriate way to do this?

Issue #2820, opened 2023-02-09
`JavaTemplate` is unaware of `instanceof` pattern variables


knutwannheden:
When `JavaTemplate` is used to insert a piece of code which references an `instanceof` pattern variable in the code where the template is embedded, the reference in the result is not properly type attributed.

So if in the example:
```java
              class A {
                  void foo(Object o) {
                      int i = 0;
                      if (o instanceof String s) {
                          i = 1;
                      }
                  }
              }
```
the `i = 1` assignment is replaced by `i = s.length()` using a template, the type attribution on `s` in this assignment is missing in the result.

The `BlockStatementTemplateGenerator` would have to include any `instanceof` expressions which are in `if` conditions and ternary operator conditions as placeholders. In the example for example: `if (null instanceof String s)`.

PR #2863, opened 2023-02-18
Support `instanceof` pattern variables in `JavaTemplate`

knutwannheden:
Using `instanceof` pattern matching an `if` statement can declare local variables which can be dereferenced in the condition itself, in the `then` block, the `else` block, or also in successor statements to the `if` statement. Therefore, it is important that `JavaTemplate` keeps these `instanceof` pattern matching expressions when generating the stub to extract the type attributed substitution of a template.

For any `if` statements which contain `instanceof` pattern variables the `if` statement is added in a reduced form to the stub. The flow scoping rules of the pattern variables require the then and else blocks to retain the same "completes normally" characteristics as in the original code, as this influences the scope of the pattern variable.

Additionally this PR fixes a few other minor problems with the stub generation in the `JavaTemplate` engine. Specifically, any variables declared in the same block as the element to be replaced, but preceding it, are now also generated into the stub. This was also the case before, but didn't work in the context of loops or labeled blocks.

Finally, `J.Ternary` expressions with pattern variables in the condition are also supported.

Fixes: #2820

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateInstanceOfTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateInstanceOfTest.java
@@ -1,0 +1,503 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import org.intellij.lang.annotations.Language;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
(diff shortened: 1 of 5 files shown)


knutwannheden:
There is a few different things going on here now. I will factor some parts out and commit them separately.

knutwannheden:
@sambsnyd I think I am getting ready to wrap this PR up. The `JavaTemplate` engine must now check `if` statements for `instanceof` pattern matching expressions in the condition and potentially include these `if` statements (with a "minified" condition) in the stub it produces, as the patter variable could be visible in the then block, the else block, or even after the `if` statement.

Now, I am wondering if the code should check that there is a Java 17 marker and only conditionally do all this extra checking. What do you think? I would also have to check if we can easily access that marker. AFAIK it is attached to the top-level LST element.

sambsnyd:
@knutwannheden sorry it has taken me a long time to review this... why do the tests reference expected missing types? Why are any types missing?

knutwannheden:
> @knutwannheden sorry it has taken me a long time to review this... 

No worries.

> why do the tests reference expected missing types? Why are any types missing?

Right. The tests are very unusual in this respect. It has to do with the fact that when an `if` statement has an `instanceof` pattern variable in the condition, this variable can be dereferenced in any of the four following locations:
- expressions in the `if` condition itself
- statements of the `then` block
- statements of the `else` block
- statements *after* the `if` statements
In which of these locations depends on what the `if` condition exactly looks like, but also what the `then` and `else` blocks look like. Specifically whether these blocks complete normally or not.

Now, since we in `BlockStatementTemplateGenerator` generate a stub, we need to make sure that we generate a correct stub for the `if` statement so that the pattern variable is visible in all the correct locations. So these very specialized tests test this by ensuring that the "abstraction" of the `if` statement generated by `BlockStatementTemplateGenerator` is correct in this respect.

PR #2630, opened 2023-01-11
Improve: Unify Representing J.Binary.Type printing

JLLeitschuh:

PR #2867, opened 2023-02-20
Add `getKeyword()` to `J.Binary.Type`, `J.Unary.Type`, and `J.AssignmentOperation.Type`

knutwannheden:
Similar to #2630 but without adding any Java fields which would increase the size of the LST.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
@@ -204,46 +204,10 @@ public class JavaPrinter<P> extends JavaVisitor<PrintOutputCapture<P>> {
 
     @Override
     public J visitAssignmentOperation(AssignmentOperation assignOp, PrintOutputCapture<P> p) {
-        String keyword = "";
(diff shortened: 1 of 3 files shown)


knutwannheden:
> Why? We are very guarded about adding API surface area to any LST elements.

The only reason would be to get rid of some duplication. For the `J.Binary.Operation` case I have another place, where I would use it. But I understand your reasoning and have no problem with closing the PR again.

knutwannheden:
The new use case I mention is in my other open PR for pattern variable support in `JavaTemplate`: https://github.com/openrewrite/rewrite/pull/2863/files#diff-1642ff64ca8163e1eb1e9f8e8a296966bff9ce05b23eb91dc14466f76231f561R611-R621

sambsnyd:
Hmm, I like this change. The fact that there were a few nigh-identical switch statements kicking around, with another to be added in the other PR @knutwannheden references, suggests to me that this is useful.

jkschneider:
@sambsnyd I asked to close this because getters that return syntax are bound to not work well across languages in the same grammar island. e.g. `or` vs. `||`.

PR #2872, opened 2023-02-21
fix: Fix a bug of recipe FinalizePrivateFields with multiple constructors

### LINKED ISSUE #2865: FinalizePrivateFields: Causes variables to not be initalized correctly
Given the following example:
```
public class Reproducer {
  private String potentiallyFinal;

  public Reproducer(String potentiallyFinal) {
    this.potentiallyFinal = potentiallyFinal;
  }

  public Reproducer() {
  }
}
```
FinalizePrivateFields adds the final modifier to the field `potentiallyFinal` witch does not compile.  The Field only has one mutator with is in the constructor, but the class has different constructors witch do not set the Field. I would expect one of the following:
- rewrite does NOT apply the FinalizePrivateFields at all
- rewrite initializes the Field with its default value (in this case null, but diffrent for primitive types) in all other contructors
- rewrite calls the main contructor with its default value (i believe this would be quite hard to detect: a) main constructor might have other side effects b) other Parameters might influence this)

kunli2:
fixes #2865 
fixes #2869 
Fix a bug in recipe `FinalizePrivateFields` with multiple constructors, if any constructor does not initialize a field, it can not be final.
### Solution:
Skip rewriting if a class has multiple constructors.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/FinalizePrivateFieldsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/FinalizePrivateFieldsTest.java
@@ -449,6 +449,8 @@ class FinalizePrivateFieldsTest implements RewriteTest {
         );
     }
 
+    @Disabled("Multi constructors ignored")
+    @Issue("https://github.com/openrewrite/rewrite/issues/2865")
     @Test
     void fieldAssignedIndirectlyInAllAlternateConstructors() {
         rewriteRun(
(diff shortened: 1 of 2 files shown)


knutwannheden:
@kunli2 Even better would be if we could check that all constructors directly or indirectly initialize a given field.

kunli2:
@knutwannheden , yeah, that's definitely an enhancement, but the change will be big compared to this patch. so having this patch first and then another enhancement next.

PR #2877, opened 2023-02-22
Added Recipe to handle Method Invocation of Validate#NotNull having V…

### LINKED ISSUE #2724: Replace `org.apache.commons.lang3.Validate.notNull` with `Objects.requireNonNull`

srmalkan:
…ariable arguments

Committed Changes handles the remaining part. Fixes #2724


timtebeek:
Great to see how far you've come already with this! I've added some commits to fix the formatting, and added two sample tests to highlight issues with space and comments. Likely we'll want to do something [similar as seen here](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/lang/StringFormatted.java#L79), where we copy the prefix over to the arguments that go into the nested `String.format()`. It's a little more complicated here, as it's on the inner `String.format` rather than the outer call to `Objects.requireNonNull`. 

But before we fix the whitespace & comment handling, we might need to discuss whether we should update the replacements that you're doing now. Previously with `org.apache.commons.lang3.Validate notNull(Object, String, Object[])` the formatted String was not eagerly created, and we might change semantics a bit by feeding a formatted String into `Objects.requireNonNull(Object, String)`. Consider the case where one of the arguments going into `String.format` is an expensive method call; we would not want to execute that method call if the object is not null.

Instead, we might want to replace with `Objects.requireNonNull(Object, Supplier<String>)`, such as the following:
```diff
-Validate.notNull(obj, "Object in %s should not be null", expensiveMethodCall());
-Objects.requireNonNull(obj, String.format("Object in %s should not be null", expensiveMethodCall()));
+Objects.requireNonNull(obj, () -> String.format("Object in %s should not be null", expensiveMethodCall()));
```
That way we're fairly certain not to change the semantics, as the formatted String is only created when actually necessary.

Would you agree we need to make the above two changes? We can of course help with that, as we're already glad you got this far!

srmalkan:
Hi Tim,

Thanks for highlighting it and explaining it in detail. Really appreciate it.
I have incorporated the changes suggested by you. Let me know if there is a better way to deal with whitespaces and comments.
And also I have added extra padding in the test case related to whitespace, I am not sure deal with that.

timtebeek:
hi @srmalkan ; love how you've seen this through from your initial offer to help three weeks ago, to the first implementation, the fix and second single argument implementation, and now this third implementation to handle any amount of arguments. Really appreciated!

timtebeek (approved):

PR #2878, opened 2023-02-22
Update the CodeQL data flow models

knutwannheden:
Updates the `model.csv` and `sinks.csv` files using the most current CodeQL release 2.12.2.

The queries had to be slightly changed, but the output is basically the same (except for the `generated` column which has been renamed to `provenance` with the values `manual` and `generated`). Also, `codeql query run` outputs BQRS files which can then be transformed to CSV using `codeql bqrs decode`. The resulting CSV files look slightly different (with `,` as separator rather than `;` and strings being quoted).

Please note that the new models contain more data and are presumably more accurate.


rewrite-java/src/main/java/org/openrewrite/java/dataflow/internal/csv/CsvLoader.java
```diff
@@ -51,4 +51,32 @@ private static <R extends Mergeable<R>, E> R loadCvs(InputStream input, URI sour
...
+            while (i < line.length() && line.charAt(i) != stopAt) {
+                i++;
+            }
+            result.add(line.substring(valueStart, i));
+            if (stopAt == '"') {
+                i++;
+            }
+        }
+        return result.toArray(new String[0]);
+    }
```
JLLeitschuh, line 80: Ah, okay, interesting. This works I suppose.

rewrite-java/src/main/java/org/openrewrite/java/dataflow/internal/csv/CsvLoader.java
```diff
@@ -51,4 +51,32 @@ private static <R extends Mergeable<R>, E> R loadCvs(InputStream input, URI sour
         }
     }
 
+    private static String[] parseLine(String line) {
+        if (line.isEmpty()) {
+            return new String[0];
+        }
+
+        ArrayList<String> result = new ArrayList<>();
```
JLLeitschuh, line 59: You could theoretically pre-size this as a max size of 10

knutwannheden: Which happens to be the default capacity of an `ArrayList` 😀. If we wanted to optimize this, there would be other means too, such as always reusing the same array (of the correct length). But it's a meager 10K rows, so I didn't bother.

rewrite-java/src/main/java/org/openrewrite/java/dataflow/internal/csv/CsvLoader.java
knutwannheden: The simple `String#split()` doesn't work anymore due to the quoted values and I didn't want to bring in a new Jackson dependency just for this. But like you noticed it isn't pretty and by no means optimized. Feel free to improve :-)

rewrite-java/src/main/java/org/openrewrite/java/dataflow/internal/csv/CsvLoader.java
JLLeitschuh: > Which happens to be the default capacity of an ArrayList.

Oh, wow, didn't know this. Thanks! 😆

PR #2882, opened 2023-02-22
recipe: Replace `StringBuilder.append()` with `String`

### LINKED ISSUE #2862: Recipe wanted: `StringBuilder` can be replaced with `String`
Recipe wanted:
`StringBuilder` can be replaced with `String` .

Example:
Screen Shot 2023-02-17 at 1 25 29 PM

Transform 
```java
String s = new StringBuilder().append("A").append("B").toString();
```

To 
```java
String s = "A" + "B";
```

kunli2:
Introduce recipe `ReplaceStringBuilderWithString`.

**Note to reviewer:**
This PR depends on and includes PR #2844, which needs to be rebased and merge after that one.
Closes #2862.

Replace `StringBuilder.append()` with String if you are only concatenating a small number of strings and the code is simple and easy to read, as the compiler can optimize simple string concatenation expressions into a single String object, which can be more efficient than using StringBuilder.

An example
Before
```java
String s = new StringBuilder().append("A").append("B").toString();
```

After
```java
String s = "A" + "B";
```


rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
```diff
@@ -0,0 +1,126 @@
...
+     * Concat an additive expression in a group and add to groups
+     */
+    private static void addToGroups(List<Expression> group, List<Expression> groups) {
+        if (!group.isEmpty()) {
+            groups.add(ExpressionUtils.additiveExpression(group));
+            group.clear();
+        }
+    }
+
+    private static boolean flatAdditiveExpressions(Expression expression, List<Expression> expressionList) {
```
knutwannheden, line 110: I think here and some other places where you have `instanceof` checks on `Expression` objects you may also want to call `Expression#unwrap()` first to automatically deal with `J.Parentheses`.

But in this specific case, since you already have a recursive method, you could also just add an `if` for `J.Parentheses`.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
```diff
@@ -0,0 +1,126 @@
...
+import org.openrewrite.java.tree.*;
+import org.openrewrite.java.utils.ExpressionUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Collections.singletonList;
+
+public class ChainStringBuilderAppendCalls extends Recipe {
+    private static final MethodMatcher STRING_BUILDER_APPEND = new MethodMatcher("java.lang.StringBuilder append(String)");
```
knutwannheden, line 34: Can't we also match all other `append()` overloads with a single parameter?

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ReplaceStringBuilderWithString.java
```diff
@@ -0,0 +1,123 @@
...
+import org.openrewrite.java.tree.*;
+import org.openrewrite.java.utils.ExpressionUtils;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class ReplaceStringBuilderWithString extends Recipe {
+    private static final MethodMatcher STRING_BUILDER_APPEND = new MethodMatcher("java.lang.StringBuilder append(String)");
```
knutwannheden, line 17: Same question here.

rewrite-java/src/main/java/org/openrewrite/java/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,91 @@
...
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Collections.emptyList;
+
+public final class ExpressionUtils {
```
knutwannheden, line 30: Since this class for now deals with string concatenation specifically, I would suggest to either rename the whole class or the individual utility methods. Because if someone were to use it to add two primitive ints for example, the returned `J.Binary` would be incorrectly type-attributed, which can have hard to track down undesirable consequences, I think.

Also, with this approach, it would for the given reason also be problematic to match the other `StringBuilder#append()` overloads.

Possibly this class should also be in an `internal` package?

rewrite-java-test/src/test/java/org/openrewrite/java/Utils/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.Utils;
```
knutwannheden, line 16: Here an uppercase `Utils` appears to have slipped in.

knutwannheden (commented):
@kunli2 Not sure if this PR was even ready for review yet, but I took the liberty to add some review comments.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
kunli2: Thanks for letting me know the functionality of `Expression#unwrap()`. 

Do you mean to cover a double-parenthesis case like this?
```java
sb.append(("A" + op + "B"));
```

I am also concerned that unnecessary unwrapping could make some problems like 
```
(1 + 2) * 3 
```
unwrapped to
```
1 + 2 * 3
```

rewrite-java-test/src/test/java/org/openrewrite/java/Utils/ExpressionUtilsTest.java
kunli2: this is a typo and I am going to remove this package. and move this class to be package-private.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
kunli2: good point, but for other `append()`  methods with other kinds of parameters, we don't want to transform.
Like this
```java
    void method1() {
        StringBuilder sb = new StringBuilder();
        int x = 1;
        sb.append(1 + x + 2);
    }
```
what do you think?

rewrite-java/src/main/java/org/openrewrite/java/utils/ExpressionUtils.java
kunli2: Moved this Util class to be package-private.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
knutwannheden: Not sure. For this case the result would have to have it parenthesized. But I think the idea of your recipe was to only cover simple cases, so you may want to exclude the case when the argument is a "complex" expression. You currently have an exception for `J.Binary` with the `+` operator which I think is fine, but you may additionally want to check that the type of that `J.Binary` is `java.lang.String`. Then your example would not be rewritten.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
knutwannheden: Also after unwrapping the LST would still have the "correct shape" (i.e. the `*` as the top-level node). But you would still have to be careful to serialize the correct thing back to the source code (I don't know if the JavaPrinter automatically adds "missing" parentheses, but that is an interesting question!).

I haven't looked carefully enough at the code to determine whether unwrapping actually makes sense here or not. Just a thought that struck me.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
kunli2: good point, adding a check that the type of that `J.Binary` is `java.lang.String` makes sense to me, will add it.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
```diff
@@ -0,0 +1,126 @@
...
+                    List<Expression> group = new ArrayList<>();
+                    for (Expression exp : flattenExpressions) {
+                        if (exp instanceof J.Literal) {
+                            group.add(exp);
+                        } else {
+                            addToGroups(group, groups);
+                            groups.add(exp);
+                        }
+                    }
+                    addToGroups(group, groups);
```
knutwannheden, line 83: The grouping also has some tricky edge cases. Rewriting `sb.append(op + 'a' + 'b')` as `sb.append(op).append('a' + `'b')` would not be correct.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/ChainStringBuilderAppendCalls.java
kunli2: will have a separate PR to handle this case of double parentheses since it's a change for another recipe.

Issue #1859, opened 2022-05-31
Maven - attempt anonymous artifact download


nmck257:
Consider a Maven settings.xml which uses property resolution for credentials:
```xml
  <servers>
    <server>
       <id>my-organization-repository</id>
       <username>${MAVEN_USERNAME}</username>
       <password>${MAVEN_PASSWORD}</password>
    </server>
  <servers>
```
And, suppose those properties are not resolved at runtime (eg the environment does not have those variables set, perhaps because most requests to that repository do not actually require authentication).

When rewrite-maven attempts to download artifacts using this config, it will use the placeholders directly as credentials, and inevitably receive a 401 (bad credentials) response. (Note that if the properties *are* set, then rewrite-maven works as expected, and if those properties are explicitly set to blank values, then rewrite-maven proceeds with anonymous requests to the repository)

However, with the same config, Apache Maven succeeds at downloading artifacts. Ostensibly, Apache Maven either attempts anonymous requests (either before or after attempting authenticated requests), or has some logic to discard credentials which are formatted like placeholder properties. It would be nice if rewrite-maven had a similar behavior.

This scenario might be a bit niche, but, solving it would increase the likelihood of rewrite-maven "just working" across diverse environments.

tkvangorder:
Hi @nmck257 , 

It appears that maven will fall back to anonymous access even when you define a user/password with no property place-holders. So this might be a fix where, if we get a 4xx error from the repo (and we have user/pass), try again without.

tkvangorder:
```
    <mirrors>
        <mirror>
            <id>moderne-remote-cache</id>
            <mirrorOf>
                central
            </mirrorOf>
            <url>https://artifactory.moderne.ninja/artifactory/moderne-remote-cache</url>
        </mirror>
    </mirrors>
    <servers>
        <server>
            <id>moderne-remote-cache</id>
            <username>MYNAME</username>
            <password>MYPASS</password>
        </server>
    </servers>
```

Maven still uses anonymous access

nmck257:
Sure, I can prepare an implementation for that.

IIRC, I didn't go that route originally because I was worried about the extreme case of ~doubling round-trips per download -- then I considered logic to remember whether the credentials were good or bad for a given server, but that could get complex if the user is only authorized for certain paths in that repository, and different repos might give different HTTP codes for that case... Eventually I yielded and just went for this "offline" solution.

[...]

PR #2889, opened 2023-02-23
Improve error reporting: do not swallow an error of the first download attempt.

mduerig:
As per https://github.com/openrewrite/rewrite/issues/1859#issuecomment-1440558482

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -793,7 +793,7 @@ public class MavenPomDownloader {
                     "HTTP " + responseCode;
         }
 
-        public boolean isAccessDenid() {
+        public boolean isAccessDenied() {
             return isClientSideException() && responseCode != 404;
         }
     }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -738,7 +738,7 @@ public class MavenPomDownloader {
         try {
             return sendRequest.apply(httpSender.get(uriString).build());
         } catch (HttpSenderResponseException retryException) {
-            if (retryException.isAccessDenid()) {
+            if (retryException.isAccessDenied()) {
                 throw originalException;
             } else {
                 throw retryException;
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -722,19 +722,29 @@ public class MavenPomDownloader {
      */
     private byte[] requestAsAuthenticatedOrAnonymous(MavenRepository repo, String uriString) throws HttpSenderResponseException {
         try {
-            try {
-                return sendRequest.apply(applyAuthenticationToRequest(repo, httpSender.get(uriString)).build());
-            } catch (HttpSenderResponseException e) {
-                if (hasCredentials(repo) && e.isClientSideException()) {
-                    return sendRequest.apply(httpSender.get(uriString).build());
-                } else {
-                    throw e;
-                }
-            }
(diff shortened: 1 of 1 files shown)


timtebeek:
Thank you for this @mduerig ! I've merely organized the imports and applied a formatter just to be consistent and minimize the changes. Still [best viewed with whitespace changes ignored](https://github.com/openrewrite/rewrite/pull/2889/files?diff=unified&w=1). Asked @nmck257 for review, as you two discussed this over on https://github.com/openrewrite/rewrite/issues/1859#issuecomment-1440558482.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -211,33 +211,38 @@ public MavenMetadata downloadMetadata(GroupArtifactVersion gav, @Nullable Resolv
...
-                        result = Optional.of(MavenMetadata.parse(responseBody));
-                    }
-                } catch (HttpSenderResponseException e) {
-                    repositoryResponses.put(repo, e.getMessage());
-                    if (e.isClientSideException()) {
-                        //If we have a 400-404, cache an empty result.
-                        cacheEmptyResult = true;
+                        break;
+                    } catch (HttpSenderResponseException e) {
+                        repositoryResponses.put(repo, e.getMessage());
```
nmck257, line 238: since `repositoryResponses` is a `Map`, and the `MavenRepository` is the same for both the authenticated and anonymous request, wouldn't the first failure response get overwritten in the `Map` at this line in the case where both requests fail?

nmck257 (commented):
`MavenPomDownloaderTest` has a reasonable pattern for testing out different response behaviors from the remote repository -- it would be good to add a test case verifying the newly-added behavior

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
mduerig: That's true, unfortunately. Maybe the PR is taking this too far and a better approach would be to change which exception is thrown from `MavenPomDownloader#requestAsAuthenticatedOrAnonymous`. Instead of throwing the 401 exception from the retry, throw the original one that caused the retry? After all it is kinda expected that the retry would fail with a 401.

nmck257: Hmm -- IIRC, some remote repo implementations return 404 for requests to artifacts which you aren't authorized to see, not 401. But I think something like that could still work, ie, deciding which exception is the "interesting" one based on the status codes from each

mduerig: I'll try to come up with something later. I'm busy with other stuff next week.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -782,6 +792,10 @@ public String getMessage() {
                     requireNonNull(getCause()).getMessage() :
                     "HTTP " + responseCode;
         }
+
+        public boolean isAccessDenid() {
```
timtebeek, line 796:
```suggestion
        public boolean isAccessDenied() {
```

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -722,19 +722,29 @@ public MavenRepository normalizeRepository(MavenRepository originalRepository, @
...
+        } catch (Throwable t) {
+            throw new RuntimeException(t);  // unreachable
+        }
+    }
+
+    private byte[] retryRequestAnonymously(String uriString, HttpSenderResponseException originalException) throws HttpSenderResponseException {
+        try {
+            return sendRequest.apply(httpSender.get(uriString).build());
+        } catch (HttpSenderResponseException retryException) {
+            if (retryException.isAccessDenid()) {
```
timtebeek, line 741:
```suggestion
            if (retryException.isAccessDenied()) {
```

PR #2890, opened 2023-02-24
Remove the gradle-api from being shaded into rewrite-gradle.

shanman190:
This enables better alignment in terms of type attribution by utilizing the Gradle, Groovy, and (future) Kotlin versions as provided by the executing Gradle wrapper.

### DIFF
--- a/rewrite-gradle/build.gradle.kts
+++ b/rewrite-gradle/build.gradle.kts
@@ -1,8 +1,5 @@
-import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
-
 plugins {
     id("org.openrewrite.build.language-library")
-    id("org.openrewrite.build.shadow")
     id("groovy")
 }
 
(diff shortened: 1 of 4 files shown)


shanman190:
@timtebeek and @mike-solomon,

Got a question for you folks. As part of this PR it's going to require a documentation update somewhere to note that for downstream recipe authors that are consuming `rewrite-gradle` that they will need to include the `gradleApi()` dependency in order to properly type attribute for any tests. I was chatting with @sambsnyd in Slack and there didn't sound like there was yet possibly a great place to put such a note. As a result, we figured it'd be good to get more opinions on where would be a good place and we thought that since you all have been in the documentation a lot why not ask you all directly. 😄 Any thoughts?

mike-solomon:
Heyo @shanman190 -- thanks for working on this and thinking of the docs!

Couple of questions to help me better understand what's going on:
1. When we say "downstream recipe authors that are consuming `rewrite-gradle`" is that different than people who write recipes using gradle? Like is there some subsection of gradle authors that won't need this or is it every person who makes a recipe using gradle?
2. If a user _isn't_ writing a recipe but is, instead, using a gradle recipe - do they need to do anything different?
3. If an author of a recipe doesn't include `gradleApi()` will the recipe still work? Is it only the tests that will fail?

My instinct is that we'd want to update the [recipe development environment guide](https://docs.openrewrite.org/authoring-recipes/recipe-development-environment) - but we might want other places to be updated as well depending on the answers to the above.

shanman190:
@mike-solomon,

1. By "downstream recipe author that are consuming `rewrite-gradle`", I mean a user whom is building an OpenRewrite recipes in another place (ie. an external repository) that are specifically wanting to make changes against Gradle **AND** need to have the Gradle type attribution requirement fulfilled for _parsing_. In all of the cases that I can think of this will be for testing recipes where we are giving `rewriteRun(..)` Gradle buildscript text to parse. 
2. If a user is consuming `rewrite-gradle` in the form of execution (ie. running via the `rewrite-gradle-plugin`), then there won't be any issues. Since the `rewrite-gradle-plugin` is parsing from within a Gradle execution, it has proper type attribution from the Gradle distribution itself. Once sources have been parsed, execution operates as normal over the fully type attributed LST.
3. So this one is fun, but I'll try to carefully articulate it... the parsing itself should still succeed just fine. However, if the user is using things such as `MethodMatcher` **AND** are dependent upon the type attribution (such as the `MethodMatcher` needs type attribution to be present), then that will be unavailable and the recipe tests will fail.

PR #2893, opened 2023-02-25
feat: add RemoveCompileScope recipe

### LINKED ISSUE #2888: Maven(Clean up) - Remove explicit `compile` scope
The `compile` scope [is the default](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope) and there is no need to define it explicitly 

This is similar to `[ChangeDependencyScope](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyScope.java)` but that recipe expects a dependency and a scope as the input. We might be able to reuse some of the logic from the `visitor`

yeikel:
Closes #2888


rewrite-maven/src/test/java/org/openrewrite/maven/cleanup/RemoveCompileScopeTest.java
```diff
@@ -0,0 +1,214 @@
...
+                            <version>28.2-jre</version>
+                            <scope>compile</scope>
+                        </dependency>
+                    </dependencies>
+                </dependencyManagement>
+              </project>
+              """
+          )
+        );
+    }
```
timtebeek, line 213: Might want to add a test case here where you override the scope from dependencyManagement, if I [recall correctly](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope). And note that the dependency management might also be in a parent, so that makes this a bit harder to determine perhaps. We would not want to remove an overridden scope if that changes the behaviour.

```suggestion
    }

    @Test
    void overridenDependencyManagementScopeIsUntouched() {
        rewriteRun(
          pomXml(
            """
              <project>
                <modelVersion>4.0.0</modelVersion>

                <groupId>com.mycompany.app</groupId>
                <artifactId>my-app</artifactId>
                <version>1</version>
                <dependencyManagement>
                    <dependencies>
                        <dependency>
                            <groupId>com.google.guava</groupId>
                            <artifactId>guava</artifactId>
                            <version>28.2-jre</version>
                            <scope>test</scope>
                        </dependency>
                    </dependencies>
                </dependencyManagement>
                <dependencies>
                    <dependency>
                        <groupId>com.google.guava</groupId>
                        <artifactId>guava</artifactId>
                        <version>28.2-jre</version>
                        <scope>compile</scope>
                    </dependency>
                </dependencies>
              </project>
              """
          )
        );
    }
```

yeikel: Thank you for the insight. This is another example where to make a decision I'd need to have the entire document in front of me beforehand 

9f77267c93d6ffa5ba27f6f9f81ffb69a1393d97 is a decent attempt but that does not work when the order is different (see the new test added)

Do you have any examples where I can visit the document twice?  I think that using this approach might work, but I'd need to scan the entire document and then run it again with the filled `seen` entries

if there is any better way to do this please let me know

jkschneider: @yeikel You can use `visit(List<SourceFile> before, ExecutionContext ctx)` instead of `getVisitor()` in a recipe. In many cases we take multiple full passes over a file (or a whole project) in the implementation of that method.

yeikel: Thank you for the help @jkschneider 

I pushed d08c82cb35c5c56a2e830bf926cb54aea2bbe133 using that idea

yeikel:
Github is acting up for some reason. I am trying to re-request a review from @jkschneider and @timtebeek 

Thanks!

timtebeek:
Looking at the implementation and current set of tests my worry is if we would correctly handle cases where there is a parent with a `dependencyManagement` section, setting an explicit scope, that is overridden with `compile` in a project `pom.xml`. That parent can also be defined outside the current project, and might need to be resolved & evaluated.

yeikel:
> Looking at the implementation and current set of tests my worry is if we would correctly handle cases where there is a parent with a `dependencyManagement` section, setting an explicit scope, that is overridden with `compile` in a project `pom.xml`. That parent can also be defined outside the current project, and might need to be resolved & evaluated.

Are you aware of any other recipe that is doing this? Or is this a foundational need of rewrite?

If you can share any example that would be helpful

rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/RemoveCompileScope.java
```diff
@@ -0,0 +1,163 @@
...
+public class RemoveCompileScope extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Removes the compile scope";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Removes the compile scope for all the dependencies as it is the default scope,and it is redundant.";
```
sambsnyd, line 38:
```suggestion
        return "Remove the compile scope from dependencies where it is redundant. Compile is the default scope.";
```

rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/RemoveCompileScope.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Override
+    public List<SourceFile> visit(List<SourceFile> before, ExecutionContext ctx) {
+        // First we collect all the dependencies under dependencyManagement
+        // We should not remove the compile scope for dependencies declared with a different scope within dependencyManagement
+
+        final Map<String, String> seen = new HashMap<>();
+
+        for (SourceFile entry : before) {
```
sambsnyd, line 48: The scope of a dependency might be managed by a parent pom not present in the project which will never be discovered by this loop. 

Instead look at the `MavenResolutionResult` marker which will be present on any `Xml.Document` which is a maven pom. This marker contains the complete set of requested and resolved dependencies, parent poms, etc.

I don't have an example of the exact sort of lookup which is required here, but [AddManagedDependency](https://github.com/yeikel/rewrite/blob/f102c64f95f0580dd78e8899d28d38344456f958/rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java#L169) has an example of usage of this marker which you may find helpful

PR #2894, opened 2023-02-25
Added ReplaceStringLiteralWithConstant recipe

### LINKED ISSUE #1622: Recipe: ReplaceLiteralWithConstant
As an inverse to [ReplaceConstant](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstant.java). `ReplaceConstant` allows users to replace a constant with a literal value, for instance to phase out usage of Guava's `Charsets.UTF_8`.

I propose to add an inverse recipe as well to replace String literals with constants. For instance to replace string literals such as `"application/json"` with  Spring's constant `MediaType.APPLICATION_JSON_VALUE`.
I don't think one would typically want to add a dependency merely to use a constant, but perhaps that should be possible to define as well.

This way there's an easy flow for users to add to migration recipes to clean up common string literals throughout multiple projects.

srmalkan:
Closes #1622

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ReplaceStringLiteralWithConstantTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ReplaceStringLiteralWithConstantTest.java
@@ -25,86 +25,107 @@ public class ReplaceStringLiteralWithConstantTest implements RewriteTest {
     @Override
     public void defaults(RecipeSpec spec) {
         spec
-            .parser(JavaParser.fromJavaVersion()
-                .logCompilationWarningsAndErrors(true)
-                .classpath("guava"));
+          .parser(JavaParser.fromJavaVersion()
+            .logCompilationWarningsAndErrors(true)
+            .classpath("guava"));
     }
 
     @Test
     void doNothingIfStringLiteralNotFound() {
         rewriteRun(
-            spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets", "UTF_8", "UTF_8")),
-            java(
-                """
-                  class Test {
-                      void test(Object obj) {
-                          String s = "FooBar";
-                      }
+          spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets.UTF_8", "UTF_8")),
+          java(
+            """
+              class Test {
+                  void test(Object obj) {
+                      String s = "FooBar";
                   }
-                  """
-            )
+              }
+              """
+          )
         );
     }
 
     @Test
     void replaceStringLiteralWithConstant() {
         rewriteRun(
-            spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets", "UTF_8", "UTF_8")),
-            java(
-                """
-                  class Test {
-                      Object o = "UTF_8";
-                  }
-                  """,
-                """
-                  import com.google.common.base.Charsets;
-                  
-                  class Test {
-                      Object o = Charsets.UTF_8;
-                  }
-                  """
-            )
+          spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets.UTF_8", "UTF_8")),
+          java(
+            """
+              class Test {
+                  Object o = "UTF_8";
+              }
+              """,
+            """
+              import com.google.common.base.Charsets;
+
+              class Test {
+                  Object o = Charsets.UTF_8;
+              }
+              """
+          )
         );
     }
 
     @Test
     void replaceLiteralWithUserDefinedConstant() {
         rewriteRun(
-            spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.constant.B", "VAR", "newValue")),
-            java(
-                """
-                  package com.constant;
-                  public class B {
-                      public static final String VAR = "default";
-                      void method() {
-                          String VAR = "";
-                      }
+          spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.constant.B.VAR", "newValue")),
+          java(
+            """
+              package com.constant;
+              public class B {
+                  public static final String VAR = "default";
+                  void method() {
+                      String VAR = "";
                   }
-                  """
-            ),
-            java(
-                """
-                  package com.abc;
-                  class A {
-                      String v = "newValue";
-                      private String method() {
-                          return "newValue";
-                      }
+              }
+              """
+          ),
+          java(
+            """
+              package com.abc;
+              class A {
+                  String v = "newValue";
+                  private String method() {
+                      return "newValue";
                   }
-                  """,
-                """
-                  package com.abc;
-                  
-                  import com.constant.B;
-                  
-                  class A {
-                      String v = B.VAR;
-                      private String method() {
-                          return B.VAR;
-                      }
+              }
+              """,
+            """
+              package com.abc;
+
+              import com.constant.B;
+
+              class A {
+                  String v = B.VAR;
+                  private String method() {
+                      return B.VAR;
                   }
-                  """
-            )
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void replaceStringLiteralWithConstantWithSingleArgument() {
+        rewriteRun(
+          spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets.UTF_8")),
+          java(
+            """
+              class Test {
+                  Object o = "UTF_8";
+              }
+              """,
+            """
+              import com.google.common.base.Charsets;
+
+              class Test {
+                  Object o = Charsets.UTF_8;
+              }
+              """
+          )
         );
     }
 }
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,100 @@
...
+            example = "org.springframework.http.MediaType")
+    String owningType;
+
+    @Option(displayName = "Constant name",
+            example = "APPLICATION_JSON_VALUE")
+    String constantName;
+
+    @Option(displayName = "Literal value",
+            example = "application/json")
+    String literalValue;
```
timtebeek, line 43: We might want to make this value optional, and default the to value of the constant that should replace the literal. That ensures we can configure this recipe with a minimal number of configuration options.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,100 @@
...
+@EqualsAndHashCode(callSuper = false)
+public class ReplaceStringLiteralWithConstant extends Recipe {
+
+    @Option(displayName = "Owning type of the constant",
+            example = "org.springframework.http.MediaType")
+    String owningType;
+
+    @Option(displayName = "Constant name",
+            example = "APPLICATION_JSON_VALUE")
+    String constantName;
```
timtebeek, line 39: Similar to the below; I'd maybe fold `owningType` and `constantName` into one FQN expression; that way there's only minimal parameters when configuring this in yaml.

rewrite-java-test/src/test/java/org/openrewrite/java/ReplaceStringLiteralWithConstantTest.java
```diff
@@ -0,0 +1,110 @@
...
+        rewriteRun(
+            spec -> spec.recipe(new ReplaceStringLiteralWithConstant("com.google.common.base.Charsets", "UTF_8", "UTF_8")),
+            java(
+                """
+                  class Test {
+                      Object o = "UTF_8";
+                  }
+                  """,
+                """
+                  import com.google.common.base.Charsets;
```
timtebeek, line 60: Maybe a static import of the constant would be nice? That way there's typically a minimal change to the line where the literal is used.

knutwannheden:
I was wondering if it would make sense to make the recipe more general, as initially proposed in the issue. I.e. `ReplaceLiteralWithConstant` rather than `ReplaceStringLiteralWithConstant`. I am assuming the reason not to do this was that replacing other literals will be a lot more uncommon. But with corresponding applicability tests, this could still be a valid use case. What are your thoughts on this @sambsnyd ?

rewrite-java-test/src/test/java/org/openrewrite/java/ReplaceStringLiteralWithConstantTest.java
srmalkan: I feel we should make non-static import only , this way we don't have to worry about conflicts and its code is more readable as well. let me know what you think.

timtebeek: Agreed; might complicate the recipe a bit indeed; and we can always tie on another recipe to switch from class imports for constants to direct constant imports, and solve the conflicts there. Thanks again!

rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Option(displayName = "(Optional) String literal value that needs to be replaced. Only String literal accepted.",
+            example = "application/json",
+            required = false )
+    @Nullable
+    String literalValue;
+
+    public ReplaceStringLiteralWithConstant(String fullyQualifiedNameOfConst) {
+        this.fullyQualifiedNameOfConst = fullyQualifiedNameOfConst;
+        this.literalValue = fullyQualifiedNameOfConst.substring(fullyQualifiedNameOfConst.lastIndexOf(".")+1);
```
timtebeek, line 45: This will only work for constants where the name matches the value to be replaced, as seen in the tests for `UTF_8`. But that would not work for the example given on `fullyQualifiedNameOfConst`, where the constant name is `APPLICATION_JSON_VALUE`, but the value to replace is `"application/json"`.

Could this be changed to use the value of the constant, rather than the name of the constant? That way the recipe can be used such as:


```yaml
  - org.openrewrite.java.ReplaceStringLiteralWithConstant:
      fullyQualifiedNameOfConst: org.springframework.http.MediaType.APPLICATION_JSON_VALUE
```

rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class ReplaceStringLiteralWithConstant extends Recipe {
+
+    @Option(displayName = "Fully Qualified Name of the constant",
+            example = "org.springframework.http.MediaType.APPLICATION_JSON_VALUE")
+    String fullyQualifiedNameOfConst;
```
timtebeek, line 35: Elsewhere we use for instance `fullyQualifiedMemberTypeNames`. Propose to rename this to
```suggestion
    String fullyQualifiedConstantName;
```
That way we are a little more consistent, and we get rid of the abbreviation.

sambsnyd:
I agree with @knutwannheden, can we generalize this to be any type of literal? The recipe seems largely agnostic about the specific type of literal being replace.

srmalkan:
I agree that this recipe is limited to a specific type, but at the same time replacing other types of literal is very unlikely and could have side effects as well.

if we want to make it type-agnostic, we should limit the scope to a particular package or maybe to a list of packages to prevent unintended replacements.

Let me know what you guys think.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,114 @@
...
+                                    "}"
+                    );
+
+                    // Retrieving FieldAccess from Compilation Unit
+                    J j = ((J.VariableDeclarations) result.get(0).getClasses().get(0).getBody().getStatements().get(0)).getVariables().get(0).getInitializer();
+                    if (!(j instanceof J.FieldAccess)) {
+                        throw new IllegalArgumentException("Invalid FQN : " + fullyQualifiedNameOfConst);
+                    }
+
+                    J.FieldAccess parsedFieldAccess = (J.FieldAccess) j;
```
timtebeek, line 107: I learned something new yesterday; [we now have a `PartProvider`](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/PartProvider.java), which potentially makes constructions like these a lot easier:
```java
nullableStringMethodTemplate = PartProvider.buildPart(
    "import static org.mockito.ArgumentMatchers" +
    ".nullable;\n" +
    "public class A {\n" +
    "    void method() {\n" +
    "        Object x = nullable(String.class);\n" +
    "    }\n" +
    "}", J.MethodInvocation.class, MOCKITO_CLASS_PATH);
```
Not meant to give you more work to do; If you want I can try to add this here as well. :)

sambsnyd: `PartProvider` was a workaround for a `JavaTemplate` bug... I'd rather we fix the `JavaTemplate` bug and get rid of `PartProvider` than use it more.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java
```diff
@@ -0,0 +1,114 @@
...
+
+            @Nullable
+            J.FieldAccess constant;
+
+            @Override
+            public J visitLiteral(J.Literal literal, ExecutionContext executionContext) {
+                if(isStringLiteral(literal)) {
+                    String owningType = fullyQualifiedNameOfConst.substring(0,fullyQualifiedNameOfConst.lastIndexOf("."));
+                    maybeAddImport(owningType,false);
+                    return buildConstant().withPrefix(literal.getPrefix());
```
sambsnyd, line 75: Have you tried using `JavaTemplate` to construct the replacement for the string literal? It should be able to accomplish the same thing as `buildConstant()` in fewer lines of code.  See [this document](https://docs.openrewrite.org/authoring-recipes/modifying-methods-with-javatemplate) for some examples of how to use it. Give it a try and let me know if you have questions.

sambsnyd:
I think I'm fine with this being specific to string literals... as @srmalkan says it's unlikely that there's any other literal type that we'd want to replace solely by value.

timtebeek:
So we've gonna a bit back and forth on the pull request with various suggestions. I propose we settle on the simplest case here for now and log an issue for possible improvements later. That way we can start to apply this recipe, and make backwards compatible changes later. So:

- [ ] only support String literal replacements
- [ ] use class imports rather than static imports
- [ ] make `literalValue` a required argument, for now
- [ ] do try to use `JavaTemplate` instead of `buildConstant()`

Can we all agree on the above? I don't want to move the goal posts; we're happy with your help @srmalkan! Let's see this through and handle future extensions separately, as all the above could be tackled in a backwards compatible way.

timtebeek:
Thanks again @srmalkan !

PR #2906, opened 2023-02-28
Control Flow: Add support for case switch blocks

JLLeitschuh:


rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1030,6 +1030,15 @@ public Padding getPadding() {
...
         }
 
+        /**
+         * @return true if this case is the default case.
+         */
+        public boolean isDefaultCase() {
+            return getExpressions().size() == 1 &&
+                    getExpressions().get(0) instanceof J.Identifier &&
+                    "default".equals(((J.Identifier) getExpressions().get(0)).getSimpleName());
+        }
```
JLLeitschuh, line 1040: @sambsnyd @jkschneider this fine?

jkschneider: No, this is not language neutral. In some languages default is `else`, etc.

JLLeitschuh: IE. Is there some way we can offer this API on the model in a language agnostic manner?

knutwannheden: I get the impression that there should maybe be an additional `boolean isDefaultCase` field or similar in the LST, which the parsers would have to provide the value for when creating the element.

jkschneider: i'm once again asking that we don't overindex for solving the general problem right now, and keep the "is default" utility constrained to data flow where it is needed for the time being. otherwise these kinds of discussions substantially distract us from other growth areas.

JLLeitschuh: Removed from `J.Case` and moved into control flow

PR #2908, opened 2023-03-01
Cleanup `BlockStatementTemplateGenerator`

knutwannheden:
It looks like over time some hacks have accumulated in `BlockStatementTemplateGenerator` by adding additional `/*TEMPLATE_STOP*/` comments here and there. This is an attempt to clean this up.


knutwannheden:
I still need to check that there are no failing tests. I just wanted to check if it is considered worthwhile to proceed with this or not.

knutwannheden:
If we ever feel a need to come back to this, the PR will still be around. Fixing the final test failure should not be much of an issue.

PR #2914, opened 2023-03-02
DataFlow: Support for data/taint flow from arguments to subject

JLLeitschuh:
Supports cases like `StringBuilder` where methods like `append` taint the `StringBuilder` via the MethodInvocation subject


rewrite-java-test/src/test/java/org/openrewrite/java/dataflow/FindLocalFlowPathsStringTest.java
```diff
@@ -1292,7 +1292,7 @@ void test() {
                           default:
                               break;
                       }
-                      String o = n + "";
+                      String o = /*~~>*/n + "";
```
JLLeitschuh, line 1295: Not quite sure what I did to make this work. But 🤷

PR #2921, opened 2023-03-03
Fix substitution for anonymous class

knutwannheden:
The logic in `Substitutions#substitute()` fails for `J.NewClass` instantiating new anonymous classes because an anonymous class like `A$1` is replaced with `A.1` which isn't a legal class name.


knutwannheden:
@kunli2 I think this should fix the problem you were seeing. I didn't add any test, because I would like you to first verify this for yourself and decide if this is how you want to fix the issue. Let me know if I can help with tests or anything else.

sambsnyd (approved):
Looks reasonable to me. Feel free to merge

PR #2926, opened 2023-03-04
Properly type-attribute `#{any()}` template parameters

knutwannheden:
The type attribution of `#{any()}` template parameters (i.e. without any type attribution on the parameter) now also includes any generic type arguments that the corresponding argument may have.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/Substitutions.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/Substitutions.java
@@ -26,6 +26,7 @@ import org.openrewrite.java.internal.grammar.TemplateParameterParser;
 import org.openrewrite.java.tree.*;
 
 import java.util.List;
+import java.util.StringJoiner;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/internal/template/Substitutions.java
```diff
@@ -138,12 +139,19 @@ public String substitute() {
     }
 
     private String getTypeName(@Nullable JavaType type) {
-        if (type == null) {
-            return "java.lang.Object";
+        if (type instanceof JavaType.Parameterized) {
+            return ((JavaType.Parameterized) type).getFullyQualifiedName()
+                    + '<'
+                    + ((JavaType.Parameterized) type).getTypeParameters().stream().map(this::getTypeName).collect(Collectors.joining(","))
```
jkschneider, line 145: please avoid streams in places like this

knutwannheden: I will refactor it. Possibly this could be a useful method for `TypeUtils`.

PR #2935, opened 2023-03-07
Allow using `JavaTemplate` to match LST subtrees

knutwannheden:
Adds a new method `matcher(J, JavaCoordinates)` to the `JavaTemplate` API, which allows using `JavaTemplate` as a pattern, which can be matched against

TODOs:
 - There is probably no need to assign new IDs to the template tree (see `JavaTemplateParser#cache()`)
 - Rework implementation of `TemplateMatcher` (currently a ripoff of `SemanticallyEqual`)
 - Support matching statements and possibly also other constructs
 - Consider pulling up API to `SourceTemplate` and `JavaTemplateLanguageExtension` once it is solid
 - Tests


knutwannheden:
Scrapping this in favor of #2943.

PR #2943, opened 2023-03-08
Allow matching an LST subtree against a `JavaTemplate`

knutwannheden:
`JavaTemplate` offers a new `matches(J)` method which can be used to check if the given tree matches the receiving `JavaTemplate`. This uses `SemanticallyEqual#matchesTemplate()` and will check if the tree at the cursor is identical to the tree represented by the template.

The template can also contain `#{any(...)}` parameters, which require the input subtree to contain a typed subtree at the given location with a type compatible to the type of the template parameter.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
@@ -17,12 +17,14 @@ package org.openrewrite.java;
 
 import lombok.Value;
 import org.openrewrite.Cursor;
+import org.openrewrite.Incubating;
 import org.openrewrite.Tree;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.internal.template.JavaTemplateJavaExtension;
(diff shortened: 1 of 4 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -107,6 +113,11 @@ public J visit(@Nullable Tree tree, Integer integer) {
                 .visit(changing, 0, parentCursor);
     }
 
+    @Incubating(since = "7.38.0")
+    public boolean matches(Cursor cursor, JavaCoordinates coordinates) {
```
jkschneider, line 117: Would suggest dropping this coordinates argument and always using the "replace" coordinate (weird name for matching, but at least hides the weirdness for now)

jkschneider: If you ever want to match on a subelement of the thing pointed to by `getCursor()` we could provide separate utility on `Cursor` like:

```java
Cursor c = getCursor();
c.findCursor(binary.getLeft());
```

Issue #2932, opened 2023-03-06
Maven - AddDependency - new options `skipIfTransitive`, `unless (gav)`


nmck257:
While testing the Spring Boot 3.0 recipe, we identified two undesirable behaviors with the Java upgrade recipe which handles javax -> jakarta:

* It adds `jakarta.servlet-api` as a direct dependency (when relevant) even if that dependency is already a transitive dependency on the project
* It adds `jakarta.servlet-api` (when relevant) even if `spring-boot-starter-web` is on the project (this starter includes a tomcat container dep which repackages an implementation of the same jakarta servlet-api code)

To fix these, we envision two new options on AddDependency (exact names TBD):

* `skipIfTransitive` - currently, AddDependency checks for an existing matching direct dependency in the project, and skips adding if found. This option would allow existing matching transitive deps to also trigger the skip behavior
* `unless` (list of GAVs) - do not add the dependency if one of these (arbitrary, seemingly-unrelated) deps is already on the classpath

With these options available, we could then adjust the relevant recipe in `rewrite-migrate-java` from this:

```yml
- org.openrewrite.maven.AddDependency:
    groupId: jakarta.servlet
    artifactId: jakarta.servlet-api
    version: 5.x
    onlyIfUsing: javax.servlet.* 
```

to this:

```yml
- org.openrewrite.maven.AddDependency:
    groupId: jakarta.servlet
    artifactId: jakarta.servlet-api
    version: 5.x
    onlyIfUsing: javax.servlet.*
    skipIfTransitive: true
    unless:
     - org.springframework.boot:spring-boot-starter-web 
```

I'm not fully convinced whether the `unless` option is likely to have other generic use cases in the future, though.

jkschneider:
Agreed @nmck257.

jkschneider:
@nmck257 can we make `unless` a general purpose applicable test instead?

nmck257:
so, would that be a standalone visitor class, which can be invoked from yml with the recent declarative applicability test pattern?
is there another general-purpose applicable test I could use as a reference?

I think that makes sense

shanman190:
So what I've been wondering if there should be a general purpose `not` applicability recipe. By having something like that we could then expand applicability tests to something such as:

```
applicability:
  singleSource:
    - org.openrewrite.NotApplicability:
        recipeList:
          - org.openrewrite.maven.search.DependencyInsight:
              groupId: org.springframework.boot
              artifactId: spring-boot-starter-web
[...]

Issue #2476, opened 2022-12-01
Maven AddDependency adds dependencies that are transitive already


BoykoAlex:
While upgrading **spring-petclinic** app (available here: https://github.com/spring-projects/spring-petclinic) we've noticed that a few dependencies are added to the pom but are coming from transitive dependencies anyway.

Lets have a look at **jakarta.persistence-api**. M2E shows dependency hierarchy for this dependency as follows:
![Screenshot 2022-11-30 at 19 18 42](https://user-images.githubusercontent.com/6433390/204935809-7ecda102-c830-48e7-b2cb-e75a1ef4d9a7.png)
In other words spring-data-jpa starter -> hiberbate-core -> persistence-api. All are *compile* scope dependencies. Removing the added persistence-api direct dependency doesn't have any effect. The code still compiles fine because available via *compile* scope transitive dependency I think. The culprit for the dependency to be added is: https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L202 Thus, all transitive dependencies are added.

Lets consider **jakarta.xml.bind-api**. M2E shows dependency hierarchy for this dependency as follows:
![Screenshot 2022-11-30 at 19 26 38](https://user-images.githubusercontent.com/6433390/204936662-6d24df33-87bf-414a-b216-132b18d5a1f6.png)
Besides *compile* scope dependency at the root level it only comes as *runtime* or *test* scope transitive dependency and probably this is the reason for this direct dependency to be necessary.

Now if I lift the restriction of checking direct dependencies only in `AddDependency` recipe class then **jakarta.persistence-api** direct dependency is not added because available via transitive dependency on *compile* scope. However, **jakarta.xml.bind-api** is not added either while it should be. It is not added because `getResolutionResult().getDependencies().get(Scope.Compile)` call includes **jakarta.xml.bind-api** while it seems to me that it should not be in the list of transitive dependencies on `compile` scope as it is a non-direct *runtime* and/or *test* scope dependency.

I have noticed that there is a test https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java#L579 pointing at issue #1443. I'd say that jackson-core dependency isn't needed in this case... seems like this is either similar or the same as **jakarta.persistence-api** case where we would not add the dependency.

Curious what your thoughts are on this.

timtebeek:
[...]

Issue #1443, opened 2022-03-08
org.openrewrite.maven.AddDependency: Inaccurate check of existing dependencies


thomaszub:
Hi,

the `AddDependency`-Recipe checks if the dependency is already existing before adding it to the pom. Imho it should not take transitive dependencies into account.

Take as an example `jackson-databind` as a direct project dependency. This pulls in `jackson-core` as a transitive dependency, effectively disallowing to add `jackson-core` with the `AddDependency` recipe. If the project uses `jackson-core` directly it should be added as a direct dependency. 

For an example, see the project https://github.com/thomaszub/rewrite-add-dep-issue-example (remove the second AddDependency as this triggers an error).

Kind regards
Thomas

jkschneider:
Makes sense. Perhaps we make this another recipe option?

thomaszub:
An optional flag for excluding transitive dependencies sounds good.

jkschneider:
I think you're right that this is the expected behavior in most cases. Can defer a flag until later if such a case arises.

PR #2944, opened 2023-03-08
Added `acceptTransitive` flag to AddDependency, default false. Also f…

nmck257:
…ixed inconsistency with this behavior based on dependency scope. Existing usages of AddDependency may need to be revisited to explicitly select the desired behavior for this flag.

#2932, #1443, #2476

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
@@ -132,33 +132,6 @@ public class AddDependency extends Recipe {
     @Nullable
     Boolean acceptTransitive;
 
-    @JsonCreator
-    public AddDependency(String groupId, String artifactId, String version, @Nullable String versionPattern,
-            @Nullable String scope, @Nullable Boolean releasesOnly, String onlyIfUsing, @Nullable String type,
-            @Nullable String classifier, @Nullable Boolean optional, @Nullable String familyPattern,
-            @Nullable Boolean acceptTransitive) {
-        this.groupId = groupId;
-        this.artifactId = artifactId;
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -124,6 +125,40 @@ public class AddDependency extends Recipe {
...
+            example = "true",
+            required = false)
+    @Nullable
+    Boolean acceptTransitive;
+
+    @JsonCreator
+    public AddDependency(String groupId, String artifactId, String version, @Nullable String versionPattern,
+            @Nullable String scope, @Nullable Boolean releasesOnly, String onlyIfUsing, @Nullable String type,
+            @Nullable String classifier, @Nullable Boolean optional, @Nullable String familyPattern,
+            @Nullable Boolean acceptTransitive) {
```
sambsnyd, line 139: Let's not add multiple constructors. It has caused us problems in the past. It is fine to add a new field without adding a new constructor for backwards compatibility. If there are a lot of recipes that use this recipe directly in code, rather than through yaml, we could use lombok's `@Builder` annotation

nmck257: Sure, I can make that change - I assumed preserving the backwards compatibility would be less disruptive overall, but good to know the preference.

Out of curiosity, what sort of issues? Usual hygiene stuff (eg removing the deprecation is just another task for later and some references will inevitably still break anyway when you do) or maybe some Lombok nuance?

nmck257: done @sambsnyd

PR #2963, opened 2023-03-10
Updated ChangeType to change types in java island grammars.

### LINKED ISSUE #2961: org.openrewrite.java.ChangeType not working on `Groovy` or `Kotlin`
`org.openrewrite.java.ChangeType` should be working on Kotlin sources according to Slack conversation, but this is not the case.

https://rewriteoss.slack.com/archives/C01A843MWG5/p1678437393167559?thread_ts=1678436521.463349&cid=C01A843MWG5

Test by `/gradlew :sample:rewriteDryRun`, reproducible:  https://github.com/nomisRev/rewrite-arrow/pull/4

traceyyoshima:
Changes:
- ChangeType will operate on java island grammars like `G` and `K`.

TODO:
- ~~Revise ChangeType#visitCompilationUnit so that it does not handle imports.~~
- ~~Change imports in postVisit (equivalent to super.visitCompilationUnit).~~
- ~~Update ChangeClassDefinition to work across other J implementations.~~
- ~~Update `RemoveImport` / `AddImport` to work across `J` island grammars.~~
    - Issue: `visitCompilationUnit` will not be executed on other languages like `G` or `K.`
    - `SourceFile` does not have access to `typesInUse.`
    - `visitJavaSourceFile` is used to delegate to the language-specific `visitCompilationUnit.` So, overriding the method in place of `visitCompilationUnit` causes the `K` or `G` visitCompilationUnit not to be visited.
- ~~Implement `withImports` and `withClasses` on `G`.~~

Notes:
- Recipes like AddImport and ChangeType do not add imports from `java.lang`, other languages like Kotlin include additional classes on the classpath.

fixes #2961

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
@@ -159,6 +159,10 @@ public interface G extends J {
                     .collect(Collectors.toList());
         }
 
+        public G.CompilationUnit withImports(List<Import> imports) {
+            return getPadding().withImports(JRightPadded.withElements(this.getPadding().getImports(), imports));
+        }
+
         @Transient
         public List<ClassDeclaration> getClasses() {
             return statements.stream()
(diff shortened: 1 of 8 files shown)


traceyyoshima:
Notes:
- To enable recipes on other language grammars -- do not implement `visitJavaSourceFile` or `visitCompilationUnit` in the recipe. Based on the use case, implement `preVisit` or `postVisit.`
- Any recipe that overrides visitJavaSourceFile or visitCompilationUnit will not visit `K` or `G,` which may be used to limit recipes to their implementation.
- `withImports` on each language implementation may be responsible for filtering out imports that aren't necessary like `java.lang` in `J` or `kotlin` in `K`.

```java
// Add visit to prevent visits on other `J` island grammars.
@Override
J visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
    return super.visitCompilationUnit(cu, ctx);
}
```

PR #2969, opened 2023-03-12
Allow tests to configure default Java version for sources

knutwannheden:
The `Assertions.javaVersion(int)` method is now public. Using this a `RewriteTest`-based test can now use `RecipeSpec.allSources()` to configure a default Java version for the sources:

```java
    @Override
        spec.recipe(new ReplaceTextBlockWithString()).allSources(s -> s.markers(javaVersion(14)));
    }
```


jkschneider:
My initial implementation of `allSources` feels flawed. Would be better to have config be an argument to `allSources` probably.

knutwannheden:
Sounds good. I will look into it.

kunli2 (approved):
nice one!

PR #2983, opened 2023-03-14
Maven ChangePropertyKey

ge0ffrey:
Enhancement to change a maven property key name.

Currently, OpenRewrite only support changing a maven property value, not a key.

- [X] Unit tests

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.marker.SearchResult;
+import org.openrewrite.maven.tree.ResolvedPom;
+import org.openrewrite.xml.ChangeTagNameVisitor;
+import org.openrewrite.xml.ChangeTagValueVisitor;
+import org.openrewrite.xml.tree.Xml;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class ChangePropertyKey extends Recipe {
+
+    @Option(displayName = "Old key",
+            description = "The old name of the property key to be replaced.",
+            example = "junit.version")
+    String oldKey;
+
+    @Option(displayName = "New key",
+            description = "The new property name to use.",
+            example = "version.org.junit")
+    String newKey;
+
+    @Override
+    public String getDisplayName() {
+        return "Change Maven project property key";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Changes the specified Maven project property key leaving the value intact.";
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
+        return new MavenVisitor<ExecutionContext>() {
+            @Override
+            public Xml visitDocument(Xml.Document document, ExecutionContext executionContext) {
+                String oldKeyAsProperty = "${" + oldKey + "}";
+                ResolvedPom pom = getResolutionResult().getPom();
+                if (pom.getProperties().containsKey(oldKey)) {
+                    return SearchResult.found(document);
+                }
+                if (pom.getRequestedDependencies().stream().anyMatch(d -> d.getVersion().contains(oldKeyAsProperty))) {
+                    return SearchResult.found(document);
+                }
+                if (pom.getDependencyManagement().stream().anyMatch(d -> d.getVersion().contains(oldKeyAsProperty))) {
+                    return SearchResult.found(document);
+                }
+                return document;
+            }
+        };
+    }
+
+    @Override
+    public TreeVisitor<Xml, ExecutionContext> getVisitor() {
+        return new MavenIsoVisitor<ExecutionContext>() {
+            final String oldKeyAsProperty = "${" + oldKey + "}";
+            final String newKeyAsProperty = "${" + newKey + "}";
+
+            @Override
+            public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                if (isPropertyTag() && oldKey.equals(tag.getName())) {
+                    doAfterVisit(new ChangeTagNameVisitor<>(tag, newKey));
+                }
+                if (oldKeyAsProperty.equals(tag.getValue().orElse(null))) {
+                    doAfterVisit(new ChangeTagValueVisitor<>(tag, newKeyAsProperty));
+                }
+                return super.visitTag(tag, ctx);
+            }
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,93 @@
...
+                ResolvedPom pom = getResolutionResult().getPom();
+                if (pom.getProperties().containsKey(oldKey)) {
+                    return SearchResult.found(document);
+                }
+                if (pom.getRequestedDependencies().stream().anyMatch(d -> d.getVersion().contains(oldKeyAsProperty))) {
+                    return SearchResult.found(document);
+                }
+                if (pom.getDependencyManagement().stream().anyMatch(d -> d.getVersion().contains(oldKeyAsProperty))) {
+                    return SearchResult.found(document);
+                }
```
timtebeek, line 69: Generally we try to avoid using the Streams API, especially on hot paths like applicability tests, as the object creation involved can lead to poor recipe performance.

ge0ffrey: Should I replace them with for loops?

ge0ffrey: Done

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,99 @@
...
+            final String oldKeyAsProperty = "${" + oldKey + "}";
+            final String newKeyAsProperty = "${" + newKey + "}";
+
+            @Override
+            public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                if (isPropertyTag() && oldKey.equals(tag.getName())) {
+                    doAfterVisit(new ChangeTagNameVisitor<>(tag, newKey));
+                }
+                if (oldKeyAsProperty.equals(tag.getValue().orElse(null))) {
+                    doAfterVisit(new ChangeTagValueVisitor<>(tag, newKeyAsProperty));
```
timtebeek, line 93: Do you think we should support cases where the property name is only part of the tag value?
So for instance `<value>http://some/doc/url/${guava.version}</value>`

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,99 @@
...
+                for (Dependency dependency : pom.getRequestedDependencies()) {
+                    if (dependency.getVersion().contains(oldKeyAsProperty)) {
+                        return SearchResult.found(document);
+                    }
+                }
+                for (ResolvedManagedDependency dependency : pom.getDependencyManagement()) {
+                    if (dependency.getVersion().contains(oldKeyAsProperty)) {
+                        return SearchResult.found(document);
+                    }
+                }
```
timtebeek, line 75: Should we for instance also support fairly normal common cases like plugin (management) dependency versions?

ge0ffrey: I think we should just select all pom files. Wdyt?

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
ge0ffrey: Good idea.
I would ignore elements that have elements, as `<value>${guava.version}<foo></foo></value>` seems impossible to me in typical maven land. But we can do a contains() check and replace it accordingly.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,99 @@
...
+    String oldKey;
+
+    @Option(displayName = "New key",
+            description = "The new property name to use.",
+            example = "version.org.junit")
+    String newKey;
+
+    @Override
+    public String getDisplayName() {
+        return "Change Maven project property key";
```
sambsnyd, line 48:
```suggestion
        return "Rename Maven property key";
```

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Changes the specified Maven project property key leaving the value intact.";
```
sambsnyd, line 53:
```suggestion
        return "Rename the specified Maven project property key leaving the value unchanged.";
```

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
knutwannheden: Or should we completely leave out the applicability test? We would have to traverse the entire tree to check for references, which in the end then basically does the same thing as the recipe's visitor. Or maybe there is some other good reason to use the applicability test here, which I missed.

ge0ffrey: @knutwannheden If I completely leave out the applicability test, it doesn't know to only look at maven files.

So I propose replacing it by this:

```
    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
        return new MavenVisitor<ExecutionContext>() {
            @Override
            public Xml visitDocument(Xml.Document document, ExecutionContext executionContext) {
                // Scanning every tag's value not efficient as an applicable test
                return SearchResult.found(document);
            }
        };
    }
```

Wdyt?

ge0ffrey: Change above pushed, happy to adjust if it's not the best solution.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
ge0ffrey: Done.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,99 @@
...
+                    if (dependency.getVersion().contains(oldKeyAsProperty)) {
+                        return SearchResult.found(document);
+                    }
+                }
+                for (ResolvedManagedDependency dependency : pom.getDependencyManagement()) {
+                    if (dependency.getVersion().contains(oldKeyAsProperty)) {
+                        return SearchResult.found(document);
+                    }
+                }
+                return document;
```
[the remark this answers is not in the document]
ge0ffrey: Not longer applicable.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePropertyKey.java
```diff
@@ -0,0 +1,91 @@
...
+                    doAfterVisit(new ChangeTagNameVisitor<>(tag, newKey));
+                }
+                if (tag.getChildren().isEmpty()) {
+                    Optional<String> value = tag.getValue();
+                    if (value.isPresent() && value.get().contains(oldKeyAsProperty)) {
+                        String newValue = value.get().replace(oldKeyAsProperty, newKeyAsProperty);
+                        doAfterVisit(new ChangeTagValueVisitor<>(tag, newValue));
+                    }
+                }
+                return super.visitTag(tag, ctx);
```
knutwannheden, line 87: If you instead to do the `super.visitTag(tag, ctx)` call at the beginning of the method and thus visit the tree bottom-up, I think there would be no need for `doAfterVisit()` and you could modify the `Xml.Tag` instances (returned by the super call) directly here (and return) them. Or is there something else I am missing here?

knutwannheden: Sam already updated it accordingly: https://github.com/openrewrite/rewrite/commit/4619f4c96be36fcc76db71fc779f59be9561f815#diff-0e24c5cfdd297374ffa9106a3bab282a31bcc6c2e42c73c70b55b9ca8abb9ac8L77-L79

knutwannheden:
Thanks a lot for the contribution @ge0ffrey !

PR #2989, opened 2023-03-14
recipe to find implements specific interfaces

kunli2:
Recipe to find `implements` specific interfaces

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/FindImplements.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/FindImplements.java
@@ -1,0 +1,48 @@
+package org.openrewrite.java.search;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.TypeTree;
+import org.openrewrite.marker.SearchResult;
+
+import java.util.Objects;
+
+public class FindImplements extends Recipe {
+    private final String interfaceFullyQualifiedName;
+
+    public FindImplements(String interfaceFullyQualifiedName) {
+        this.interfaceFullyQualifiedName = interfaceFullyQualifiedName;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Find class declaration with implements";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Find source files that contain a class declaration implementing a specific interface.";
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+            @Override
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/search/FindImplements.java
```diff
@@ -0,0 +1,48 @@
...
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl,
+                                                            ExecutionContext ctx) {
+                classDecl = super.visitClassDeclaration(classDecl, ctx);
+                if (classDecl.getImplements() != null) {
+                    for (TypeTree impl : classDecl.getImplements()) {
+                        if (interfaceFullyQualifiedName.equals(Objects.requireNonNull(impl.getType()).toString())) {
+                            return SearchResult.found(classDecl);
+                        }
+                    }
+                }
```
sambsnyd, line 58: If the class in question implements the interface transitively through a superclass or other interface this wont find it. You should look at the type information on the class for this. Also see the logic in `TypeUtils.findOverriddenMethod()`, it isn't exactly what you need here but it is similar logic

sambsnyd (changes requested):

rewrite-java/src/main/java/org/openrewrite/java/search/FindImplements.java
kunli2: addressed, it supports finding transitive interfaces now.

rewrite-java/src/main/java/org/openrewrite/java/search/FindImplementsOrExtends.java
```diff
@@ -0,0 +1,99 @@
...
+public class FindImplementsOrExtends extends Recipe {
+    private final String interfaceFullyQualifiedName;
+
+    public FindImplementsOrExtends(String interfaceFullyQualifiedName) {
+        this.interfaceFullyQualifiedName = interfaceFullyQualifiedName;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Find class declaration with implements";
```
sambsnyd, line 37:
```suggestion
        return "Find class declarations implementing an interface";
```

rewrite-java/src/main/java/org/openrewrite/java/search/FindImplementsOrExtends.java
```diff
@@ -0,0 +1,99 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeTree;
+import org.openrewrite.marker.SearchResult;
+
+public class FindImplementsOrExtends extends Recipe {
```
murdos, line 28: It might be clearer that way:
```suggestion
public class FindImplementations extends Recipe {
```

kunli2: thanks!

PR #2990, opened 2023-03-15
POC: Create a Trait-Based Model for the AST

JLLeitschuh:
This AST allows API users to ask higher-level questions of the AST
without requiring custom cursor traversal and visitors.

Questions like:
 - What parameter index is this `J.VariableDeclarations.NamedVariable`?
 - What callable is this `Parameter` a part of?

Future questions that could be asked are things like:
 - Is this parameter effectively final?
 - Where is this parameter read from?

The model for the traits matches, as closely as possible, CodeQL's

From the documentation:

```java
/**
 * Traits represent higher-level abstractions of Java syntax that are useful for writing recipes.
 * </p>
 * They offer APIs that allow users to write recipes that are more concise and easier to understand by offering
 * APIs that hide the details of the underlying {@link org.openrewrite.Cursor} navigation.
 * </p>
 * The traits are modeled after CodeQL's representation of the Java syntax and the queries and predicates it provides.
 */
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marker/SearchResult.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/SearchResult.java
@@ -18,9 +18,11 @@ package org.openrewrite.marker;
 import lombok.*;
 import lombok.experimental.FieldDefaults;
 import org.openrewrite.Cursor;
+import org.openrewrite.Incubating;
 import org.openrewrite.Tree;
 import org.openrewrite.internal.lang.Nullable;
 
+import java.util.Objects;
 import java.util.UUID;
 import java.util.function.UnaryOperator;
 
@@ -49,6 +51,43 @@ public class SearchResult implements Marker {
         }
         return t.withMarkers(t.getMarkers().computeByType(new SearchResult(randomId(), description),
                 (s1, s2) -> s1 == null ? s2 : s1));
+    }
+
+    /**
+     * Merge the description of two search results into a single search result with a unified description.
+     */
+    @Incubating(since ="7.41.0")
+    public static <T extends Tree> T foundMerging(@Nullable T t,String description) {
+        return foundMerging(t, description, ", ");
+    }
+
+    /**
+     * Merge the description of two search results into a single search result with a unified description.
+     * @param delimiter The delimiter to use when merging descriptions.
+     */
+    @Incubating(since ="7.41.0")
+    public static <T extends Tree> T foundMerging(@Nullable T t, String description, String delimiter) {
+        Objects.requireNonNull(delimiter, "delimiter must not be null");
+        if (t == null) {
+            //noinspection ConstantConditions
+            return null;
+        }
+        return t.withMarkers(t.getMarkers().computeByType(new SearchResult(randomId(), description),
+                (s1, s2) -> {
+                    if (s1 == null) {
+                        return s2;
+                    }
+                    if (s2 == null) {
+                        return s1;
+                    }
+                    if (s1.getDescription() == null) {
+                        return s2;
+                    }
+                    if (s2.getDescription() == null) {
+                        return s1;
+                    }
+                    return s1.withDescription(s1.getDescription() + delimiter + s2.getDescription());
+                }));
     }
 
     /**
--- a/rewrite-java/src/main/java/org/openrewrite/java/trait/util/Either.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/trait/util/Either.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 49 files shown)


rewrite-java/src/main/java/org/openrewrite/java/trait/variable/Parameter.java
```diff
@@ -0,0 +1,95 @@
...
+    J.VariableDeclarations.NamedVariable getNamedVariable();
+
+    /**
+     * The zero indexed position of the parameter in the method declaration.
+     */
+    int getPosition();
+
+    boolean isVarArgs();
+
+    static Validated of(Cursor c) {
```
knutwannheden, line 44: One problem I see with these `of(Cursor)` methods is that typically a visitor method starts by calling `xyz = super.visitXyz(xyz, ctx);` and assigns the result to some local variable. Now, if that ended up returning a different object, because it was transformed by some visitor method, then the cursor's object will be different. So either you would have to produce a new cursor like `new Cursor(cursor.getParent(), xyz)` or you would have to pass in the object into the `of()` method (something like `of(Cursor, J.VariableDeclarations.NamedVariable)`.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3397,6 +3400,14 @@ public MethodDeclaration withParameters(List<Statement> parameters) {
             return getPadding().withParameters(JContainer.withElements(this.parameters, parameters));
         }
 
+        public List<VariableDeclarations> getParametersAsVariableDeclarations() {
+            return getParameters()
+                    .stream()
+                    .filter(VariableDeclarations.class::isInstance)
+                    .map(VariableDeclarations.class::cast)
+                    .collect(toList());
```
knutwannheden, line 3408: This is probably one of those streams which we should try to avoid.

rewrite-java/src/main/java/org/openrewrite/java/trait/variable/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@Incubating(since = "7.37.4")
```
knutwannheden, line 16: I think it would have to be `7.37.5` or `7.38.0` now.

knutwannheden (commented):
I can't tell how we should go forward with this, but I added a comment which I think should be considered before proceeding with this.

rewrite-java/src/main/java/org/openrewrite/java/trait/variable/Parameter.java
JLLeitschuh: > One problem I see with these `of(Cursor)` methods is that typically a visitor method starts by calling `xyz = super.visitXyz(xyz, ctx);` and assigns the result to some local variable.

I agree, but I've gotten feedback from others in the past that taking both arguments is redundant. As an example, see `InvocationMatcher.Advanced`. That API formerly took two arguments, the cursor and the `Expression`. But I was talked into only accepting the `Cursor`.

knutwannheden: I just quickly browsed through your PR again. Did I see it correctly that you changed to requiring a `(Cursor, Tree)` tuple?

JLLeitschuh: Not everywhere. One place in a protected method.

rewrite-core/src/main/java/org/openrewrite/marker/SearchResult.java
```diff
@@ -51,6 +52,41 @@ public static <T extends Tree> T found(@Nullable T t, @Nullable String descripti
...
+                    }
+                    if (s1.getDescription() == null) {
+                        return s2;
+                    }
+                    if (s2.getDescription() == null) {
+                        return s1;
+                    }
+                    return s1.withDescription(s1.getDescription() + delimiter + s2.getDescription());
+                }));
+    }
```
JLLeitschuh, line 88: @sambsnyd @jkschneider this API addition fine? I can mark these as `@Inclubating` if you wish

rewrite-java/src/main/java/org/openrewrite/java/trait/util/Either.java
JLLeitschuh, line 1: Much of this file is directly copied from Functional Java. Question about licensing here.

rewrite-java/src/main/java/org/openrewrite/java/trait/util/Semigroup.java
JLLeitschuh, line 1: Much of this file is directly copied from Functional Java. Question about licensing here.

rewrite-java/src/main/java/org/openrewrite/java/trait/util/Validation.java
JLLeitschuh, line 1: Much of this file is directly copied from Functional Java. Question about licensing here.

JLLeitschuh:
Closing this as all of this code has been moved to `rewrite-analysis`

PR #2992, opened 2023-03-15
fix: merge new entry into existing mapping (#2881)

### LINKED ISSUE #2881: ChangePropertyKey - Remove node, sub structure is not embedded into existing node
Hi,

I' trying to write a recipe which removes a node `d` from a YAML keeping the sub structure (`e` in the test case, in my real use case this can be more complex and is unknown). I tried the `ChangePropertyKey` recipe:

```java
@Test
void indentedPropertyEmbedIntoExisting() {
    rewriteRun(
      spec -> spec.recipe(new ChangePropertyKey("a.b.d", "a.b", null, null, null)),
      yaml(
        """
          a:
            b:
              c: false
              d:
                e: true
          """,
        """
          a:
            b:
              c: false
              e: true
          """
      )
    );
}
```

The test result is
```
expected:
a:
  b:
    c: false
    e: true
but was: 
a:
  b:
    c: false
  b:
    e: true
```

It essentially works but the a second `b` node is created leading to an invalid YAML.

Thanks and kind regards
Thomas

thomaszub:
Fixes #2881

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
@@ -778,4 +778,28 @@ class ChangePropertyKeyTest implements RewriteTest {
         );
     }
 
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite/issues/2881")
+    void embedIndentedPropertyIntoExisting() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangePropertyKey("a.b.d", "a.b", null, null, null)),
+          yaml(
+            """
+              a:
+                b:
+                  c: false
+                  d:
+                    c: true
+                    e: true
+              """,
+            """
+              a:
+                b:
+                  c: false
+                  e: true
+              """
+          )
+        );
+    }
+
 }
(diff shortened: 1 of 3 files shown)


thomaszub:
One remark: In `MergeYamlVisitor` I added `@AllArgsConstructor` with `shouldAutoFormat = true` to not break rewrite-spring. This leads to two constructors (one with and one without `shouldAutoFormat`)  instead of one with `shouldAutoFormat`.

rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
```diff
@@ -778,4 +778,28 @@ void doesNotBreakOnKeysWhichIncludeRegexSpecialCharacters() {
...
+    void embedIndentedPropertyIntoExisting() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangePropertyKey("a.b.d", "a.b", null, null, null)),
+          yaml(
+            """
+              a:
+                b:
+                  c: false
+                  d:
+                    c: true
```
timtebeek, line 792: Was it a conscious choice to deviate slightly from the sample given in #2881 to show the merge behavior with clashing fields such as `c: true` here? Like how you leave no doubt about the impact of your implementation in those cases.

thomaszub: Ah, yes, I changed the test. Clashes can possibly occur with the current behavior of keeping the "old" values. This could be changed if `acceptTheirs` of the `MergeYamlVisitor` is set to `false`.

sambsnyd:
Thanks @thomaszub !

PR #2997, opened 2023-03-16
Handle default case for switch statements in G island grammar

shanman190:


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/SwitchTest.java
```diff
@@ -59,7 +61,12 @@ void defaultSwitch() {
               switch(0) {
                   default: System.out.println("default!");
               }
-              """
+              """,
+            spec -> spec.afterRecipe(cu -> {
```
shanman190, line 65: This test was falsely passing due to the default case source being shoved into the `sourceBefore("}")`. So printing out the statement was successfully recreating the semantic code, but incorrectly representing the rewrite grammar.

PR #3004, opened 2023-03-17
Add new `contributors` resource type

knutwannheden:
Introduces a new YAML resource type `specs.openrewrite.org/v1beta/contributors` which contains information regarding recipe contributors. These resources get loaded through the `ClasspathScanningLoader` and are used to augment the `RecipeDescriptor`s which get created for declarative and imperative recipes.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Recipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Recipe.java
@@ -35,7 +35,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Field;
-import java.net.URI;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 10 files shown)


knutwannheden:
@sambsnyd Some notes regarding the implementation:

- At the moment the `ClasspathScanningLoader` must be used to get recipe contributor information. The `YamlResourceLoader` also implements the `ResourceLoader` interface, but since it only processes a single resource at the time and we will typically have a separate resource for the contributors data (containing nothing else), I don't see that it makes sense to implement this for `YamlResourceLoader`. I.e. `YamlResourceLoader#listRecipeDescriptors()` will always return recipe descriptors *without* contributor data.
- When `Recipe#getDescriptor()` is used, the returned descriptor currently also doesn't contain any contributor data.
- The `ClasspathScanningLoader` implementation was changed to first scan YAML files, so that the contributor data (loaded from YAML) is available when processing the `Recipe` classes.
- The resource type is now `specs.openrewrite.org/v1beta/contributors`. But please let me know if we should switch to `authors` as you documented it in openrewrite/rewrite-build-gradle-plugin#5.

sambsnyd:
Thanks for this, Knut. I have added this functionality in [this commit](https://github.com/openrewrite/rewrite/commit/9ac0b7ed5afaedaf1cb13922cf0449f8cb3a70f8#diff-0467d97b36e1eea9fdc8fe3f9e644a4704d64f744aec9453923ec863442cb46c).

PR #3010, opened 2023-03-18
Do not rely on Path.toUri().getPath()

marschall:
Change `*ParserInputFileObject` classes to use `Path.getFileName()` instead of `Path.toUri().getPath()`.

The `*ParserInputFileObject` classes use `Path.toUri().getPath()` to check the last element of a path. This is both awkward and not guaranteed to work on non-default file systems. [Path#toUri](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#toUri()) makes no guarantees about the return value of `URI#getPath()` of the returned object. This breaks on the JAR file system and others. `Path.getFileName()` is a better option.

See also https://github.com/marschall/memoryfilesystem/issues/141


sambsnyd:
Thanks @marschall !

marschall:
Thank you. Good news for the weekend :-)

PR #3011, opened 2023-03-20
Try out 20-ea for now, 21-ea soon

timtebeek:
For potential inclusion in https://wiki.openjdk.org/display/quality/Quality+Outreach

### DIFF
--- a/.github/workflows/ci-early-access.yml
+++ b/.github/workflows/ci-early-access.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   build:
-    uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@java_version_input
+    uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
     with:
       java_version: 20-ea
     secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-early-access.yml
```diff
@@ -0,0 +1,23 @@
...
+on:
+  workflow_dispatch: {}
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@java_version_input
```
timtebeek, line 13:
```suggestion
    uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
```

knutwannheden (commented):
Let's try it out 😄

shanman190:
@timtebeek, just FYI, this is a bit more complicated than just switching the GitHub Actions Java versions due to us using Java toolchains.

https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/47369d2da1d62bc0a63a7f0a680129eb168aece4/src/main/java/org/openrewrite/gradle/RewriteJavaPlugin.java#L52

So basically with this current setup, all that's being tested is that Gradle can run as 20-ea/21-ea. In order to actually test rewrite against the newer version, we would have to instead switch the toolchains in use.

PR #3013, opened 2023-03-21
Fix a template error with Try in a For loop

kunli2:
This PR is to fix this error from SAAS
Screen Shot 2023-03-21 at 12 43 35 AM


rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -475,6 +475,9 @@ private void template(Cursor cursor, J prior, StringBuilder before, StringBuilde
             before.insert(0, ev.getName());
         } else if (j instanceof J.EnumValueSet) {
             after.append(";");
+        } else if (j instanceof J.Try.Resource) {
+            before.insert(0, "{");
+            after.append("}");
```
knutwannheden, line 480: I am unsure if this fix is entirely correct, because the test also succeeds if these three lines are removed again. I will take a look and see if I can identify the problem.

PR #3014, opened 2023-03-21
Generify `Validated` to `Validated<T>`

JLLeitschuh:
Funnily enough, this is NOT an API breaking change.

Any downstream users not using validated with generics will, mostly continue to have their code compile and run fine. But if you want to use generics, you can get added type protection from the compiler! 🎉 

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
@@ -138,7 +138,7 @@ public class LatestRelease implements VersionComparator {
         return normalized1.compareTo(normalized2);
     }
 
-    public static Validated build(String toVersion, @Nullable String metadataPattern) {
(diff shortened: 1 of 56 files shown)


rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
```diff
@@ -138,7 +138,7 @@ public int compare(@Nullable String currentVersion, String v1, String v2) {
         return normalized1.compareTo(normalized2);
     }
 
-    public static Validated build(String toVersion, @Nullable String metadataPattern) {
+    public static Validated<LatestRelease> buildLatestRelease(String toVersion, @Nullable String metadataPattern) {
```
JLLeitschuh, line 141: Needed to rename this to avoid name conflicts with supertypes

sambsnyd (approved):
Seems reasonable to me. I've thought about doing this before and never quite got around to it

PR #3015, opened 2023-03-22
Detect the required Java version from the failure log

timtebeek:


rewrite-core/src/main/java/org/openrewrite/FindBuildToolFailures.java
```diff
@@ -58,3 +75,45 @@ public Tree visitSourceFile(SourceFile sourceFile, ExecutionContext ctx) {
         };
     }
 }
+
+class FailureLogAnalyzer {
```
timtebeek, line 79: Figured keep this in a separate class, in case we want to pull this out to somewhere else, such as when running with successive Java versions from ingestion. Not something I expect we'll use soon, but can't hurt to keep things modular.

rpau (approved):
LGTM

PR #3024, opened 2023-03-23
Recipe instances now store their `RecipeDescriptor` to avoid cost of …

nmck257:
…repeated instantiation (especially during OSS plugin recipe stack printout)

https://github.com/openrewrite/rewrite-maven-plugin/issues/486

Significant time savings with this change, running a large recipe (based on the Spring Boot upgrade recipes) via Maven plugin on a (small!) private project:
- 1.17 recipe bom + plugin: 8m50s
- 1.18 snapshot recipe bom + plugin (with these changes): 4m48s

And those times *include* time spent on parsing+visits; all of the time savings were outside of that "critical path".

Execution time from generating RecipeDescriptors seems nearly eliminated.
The recipe "stack trace" view at the end of OSS plugin output now seems to move at basically the speed of IO.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Recipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Recipe.java
@@ -23,9 +23,9 @@ import lombok.Setter;
 import lombok.Value;
 import org.intellij.lang.annotations.Language;
 import org.openrewrite.config.DataTableDescriptor;
+import org.openrewrite.config.OptionDescriptor;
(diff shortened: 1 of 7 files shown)


nmck257:
caching the RecipeDescriptors relies on the assumption that none of the inputs to the RecipeDescriptor on a Recipe instance can change either 1) between the first call to `getDescriptor` and subsequent calls or 2) at any time post-instantiaton, which I think is accurate, but open to notes

knutwannheden:
Sounds very promising! 🚀

PR #3037, opened 2023-03-26
Added replace constant with another constant recipe.

### LINKED ISSUE #2755: Replace constant with another constant, possibly from a different class
As [asked on StackOverflow](https://stackoverflow.com/questions/75204013/how-to-replace-static-imports-of-class-constants-in-openrewrite) by @bottemav (?) , but also [posted as a comment previously](https://github.com/openrewrite/rewrite/issues/1622#issuecomment-1330273588).

**Before**
```java
import javax.ws.rs.*;
import static javax.ws.rs.core.MediaType.APPLICATION_JSON;

class ControllerClass {
    @Produces(APPLICATION_JSON)
    public String getHelloWorldJSON(String name) {
        return "{}";
    }
}
```

**After**
```java
import javax.ws.rs.*;
import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;

class ControllerClass {
    @Produces(APPLICATION_JSON_VALUE)
    public String getHelloWorldJSON(String name) {
        return "{}";
    }
}
```

**Diff**
```diff
import javax.ws.rs.*;
-import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;

class ControllerClass {
-    @Produces(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON_VALUE)
    public String getHelloWorldJSON(String name) {
        return "{}";
    }
}
```

**Related**
- Existing [ReplaceConstant](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstant.java) recipe
- https://github.com/openrewrite/rewrite/issues/1622

srmalkan:
Closes #2755

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java
@@ -17,11 +17,10 @@ package org.openrewrite.java;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.openrewrite.Cursor;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Option;
-import org.openrewrite.Recipe;
+import org.openrewrite.*;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.TypeUtils;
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public String getDisplayName() {
+        return "Replace constant with another constant";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace constant with another constant, adding/removing import on class if needed.";
+    }
+
```
timtebeek, line 48: Maybe it's good to add a applicability test that checks for the use of the owning type, such that the visitor isn't called for every field access and identifier.

timtebeek (commented):
Looking good already. Thanks for adding this one as well!

PR #3046, opened 2023-03-28
Allow `JavaTemplate` to use shared `JavaTypeCache`

knutwannheden:
As currently every application of `JavaTemplate` leads to a new `JavaParser` getting created with its own `JavaTypeCache` the result is that every time a `JavaTemplate` get applied, the LST will end up containing type attribution linked to identical yet distinct copies of `JavaType` objects, even if the LST elements refer to the same Java elements on the classpath. This in turn can cause a lot of memory overhead.

This commit aims at mitigating this issue by allowing `JavaTemplate` (specifically `JavaTemplateParser`) to use a shared `JavaTypeCache`. The life cycle of the `JavaTypeCache` is the same as that of the `ExecutionContext` and it will get used by every `JavaTemplate` with the same classpath. At the end of each compilation performed by the `JavaTemplateParser` the `JavaType` objects which were created for the parsed Java sources get purged again.

Since the `JavaTypeCache` structure itself isn't thread-safe, each `JavaTemplateParser` will create a clone of the shared `JavaTypeCache` and work with that.

The feature can be disabled by setting the `org.openrewrite.java.typecache.sharing` system property to `false`.

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/Java17Parser.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/Java17Parser.java
@@ -27,10 +27,10 @@ import java.nio.charset.Charset;
 import java.nio.file.*;
 import java.util.*;
 
-public class Java17Parser implements JavaParser {
-    private final JavaParser delegate;
+public class Java17Parser implements JavaParser, JavaParser.Internal {
+    private final JavaParser.Internal delegate;
 
-    Java17Parser(JavaParser delegate) {
+    Java17Parser(JavaParser.Internal delegate) {
         this.delegate = delegate;
     }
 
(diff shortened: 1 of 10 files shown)


knutwannheden:
> **Warning**
> This is just a proof-of-concept and there is at least one issue which would first have to be addressed: The types in the `JavaTemplate` itself "pollute" the `JavaTypeCache` with types which mustn't be shared between applications of the `JavaTemplate`. This could potentially be addressed by "partitioning" the `JavaTypeCache` along the classpath / sourcepath axis and thus throwing the `JavaTemplate`-induced sourcepath partition away at the end of the run.

jkschneider:
image

PR #3050, opened 2023-03-29
Add code examples to recipe descriptor

kunli2:
Currently, we don't have code example(s) on the recipe doc page. 
Code speaks more than words, so adding this helps people visually understand what changes the recipe will going to make.
This is the corresponding UI of this change.

Screen Shot 2023-03-29 at 8 54 58 AM

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/CodeExample.java
+++ b/rewrite-core/src/main/java/org/openrewrite/CodeExample.java
@@ -1,0 +1,21 @@
+package org.openrewrite;
+
+import lombok.Value;
+
+@Value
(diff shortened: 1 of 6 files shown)


timtebeek:
Great to see work on this; I would have thought we'd try to reuse the unit tests we have to pick a first/representative unit test to add a before/after to the recipe meta data; that way it could also work for yaml recipes perhaps. Bit of a wild idea, but thought to mention it here to get your thoughts on that.

kunli2:
Talked with Sam and we are going to implement this very similarly to the contributor feature, so closing this PR here and will work on a different approach.

PR #3056, opened 2023-03-30
Introduced UsesType includeImplicit parameter, and updated all usages…

nmck257:
… to make an appropriate choice for that param

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/java/cleanup/ChangeTypeTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/java/cleanup/ChangeTypeTest.java
@@ -15,8 +15,10 @@
  */
 package org.openrewrite.java.cleanup;
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.java.ChangeType;
+import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.TypeUtils;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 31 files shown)


rewrite-groovy/src/test/java/org/openrewrite/java/cleanup/ChangeTypeTest.java
```diff
@@ -103,4 +105,31 @@ class newFile {
           )
         );
     }
+
+    @SuppressWarnings("DataFlowIssue")
+    @Disabled("fails because there's a reference change but no content diff but that's the point; would need to adjust RewriteTest")
```
nmck257, line 110: open to suggestions if there's an easy workaround for this or if we'd prefer to just delete the test

timtebeek: While we wait for a real answer to your question you could already swap `@Disabled` for `@ExpectedToFail`. Then we at least know if this gets fixed accidentally. 
```suggestion
    @ExpectedToFail("fails because there's a reference change but no content diff but that's the point; would need to adjust RewriteTest")
```

nmck257: I didn't know about that annotation and I love it; will make that change in a few

timtebeek: Sure! If you combine it with an `@Issue` link we have a feedback loop on where to post the update if fixed.

nmck257: done, with new issue ref https://github.com/openrewrite/rewrite/issues/3058

rewrite-java/src/main/java/org/openrewrite/java/search/UsesType.java
```diff
@@ -25,21 +25,34 @@
 
 public class UsesType<P> extends JavaIsoVisitor<P> {
     private final Pattern typePattern;
+    private final Boolean includeImplicit;
 
-    public UsesType(String fullyQualifiedType) {
+    public UsesType(String fullyQualifiedType, Boolean includeImplicit) {
```
knutwannheden, line 30: Since this visitor is so very commonly used by recipes (as witnessed by the breaking builds for the three downstream projects we test this change against), I wonder if it in this case wouldn't make sense to overload the constructor and remain backward compatible. We could declare the single argument.constructor as `@Deprecated`. Thoughts?

knutwannheden (commented):
I think this looks like a very good change. It looks like you for every recipe individually checked whether it needs to know about the additional type uses or not. That must have been quite a lot of work!

rewrite-java/src/main/java/org/openrewrite/java/search/UsesType.java
knutwannheden: Thanks for clarifying. Let's move with this then and update `rewrite-spring` accordingly.

nmck257: sounds good; I'll probably merge this in an hour or two and start on corresponding changes for some of the downstream repos

PR #1436, opened 2022-03-07
Add managed dependency. closes: #1286

pway99:

PR #3064, opened 2023-03-31
Improve onlyIfUsing validation regex

jonaslindmark:
Support dashes in group and artifact IDs

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/AddManagedDependencyTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/AddManagedDependencyTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.Validated;
 import org.openrewrite.test.RewriteTest;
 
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/AddManagedDependencyTest.java
```diff
@@ -19,6 +19,7 @@
 import org.openrewrite.Validated;
 import org.openrewrite.test.RewriteTest;
 
+import static org.assertj.core.api.Assertions.as;
```
timtebeek, line 22:
```suggestion
```

timtebeek (approved):

timtebeek:
Looking back there was an earlier change to the regex in https://github.com/openrewrite/rewrite/pull/1447 ; the only change before that was the original introduction https://github.com/openrewrite/rewrite/pull/1436. Linking them here in case we need to revisit this again in the future.

Much improved, thanks!

Issue #2473, opened 2022-11-30
Add support for continuations in properties.


traceyyoshima:
```kotlin
    @Suppress("WrongPropertyKeyValueDelimiter", "TrailingSpacesInProperty")
    @Issue("https://github.com/openrewrite/rewrite/issues/2473")
    @Test
    fun escapedNewLines() {
        val source = """
                \
                
                k\
                 e\
                
                  y \
                   = \
                
                    v\
                
                     al\
                      ue\
                       
            """.trimIndent()
        val props = PropertiesParser().parse(source)[0] as Properties.File

        assertThat(props.content.map { it as Properties.Entry }.map { it.key })
            .hasSize(1)
            .containsExactly("key")
        assertThat(props.content.map { it as Properties.Entry }.map { it.value.text })
            .hasSize(1)
            .containsExactly("value")

        val outputCapture = PrintOutputCapture(InMemoryExecutionContext())
        val printer = PropertiesPrinter<InMemoryExecutionContext>()
        printer.visitFile(props, outputCapture)
        assertThat(outputCapture.out.toString()).isEqualTo(source)
    }
```

traceyyoshima:
Unassigning myself until this is a higher priority.
I've opened a branch with most of the necessary changes [here](https://github.com/openrewrite/rewrite/tree/preserve_source_code_of_properties_with_continuations) and plan to come back to this work later.

The remaining changes are fairly simple to add, but this isn't high-priority work yet.

TODOs:
- Read through the Properties documentation and apply all the parsing rules.
    - I.E. treat all spaces as a delimiter between a key and value.
- Add support for comments in the whitespace of continued lines.
- Add support for comments delimited by `!` character.
- Add with(X) in Entry to update continuations automatically.
- Update the continuation models to support key, `beforeEquals`, and value.
- Add printing for search markers for continuations for `beforeSyntax`.

tkvangorder:
![image](https://user-images.githubusercontent.com/1878529/205137906-74bd7010-9ccf-46d0-89ac-660f2fb96838.png)

gsmet:
@traceyyoshima hey. I just stumbled upon this issue after having opened https://github.com/openrewrite/rewrite/issues/2960 .

I see you have done some work related to that in a branch and from what I can see it looks like excellent progress. I wonder if we could at least get the first bits in, even if there is some later polishing remaining for corner cases after that.

[...]

PR #3065, opened 2023-03-31
Support continuation lines in properties files

knutwannheden:
Java `.properties` files can have continuation lines. The parser already supported this but the continuation lines were lost after parsing, meaning that after applying a recipe and serializing the LST again, the continuation lines got lost.

Adds a new `Properties.Entry#keySource` property and a new `Properties.Value#source` property to keep the original parsed source text, so that it can be serialized again.

Fixes: #2473

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ActivateStyle.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ActivateStyle.java
@@ -122,7 +122,7 @@ public class ActivateStyle extends Recipe {
                     keyExists.set(true);
                     if(overwriteExistingStyles) {
                         if(!fullyQualifiedStyleName.equals(entry.getValue().getText())) {
-                            entry = entry.withValue(entry.getValue().withText(fullyQualifiedStyleName));
+                            entry = entry.withValue(entry.getValue().withText(fullyQualifiedStyleName).withSource(fullyQualifiedStyleName));
                         }
                     } else {
                         List<String> activeStyles = Arrays.stream(entry.getValue().getText().split(",")).collect(Collectors.toList());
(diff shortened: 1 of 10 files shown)


knutwannheden:
The `rewrite-spring` recipe `UseTlsJdbcConnectionString` would also require updating. Should the original constructor be kept for compatibility?

traceyyoshima (changes requested):
The PropertiesParserTest#escapedEndOfLine fails.

There are some examples of tests that check the continuation rules here:
https://github.com/openrewrite/rewrite/compare/properies_parser_with_continued_lines

I don't think we need to cover all the cases immediately, but we should at least cover continued keys. I recall that occurring in a code base.

knutwannheden:
@traceyyoshima I will try to check those tests for keys with line continuations and if easy to fix also fix it. Otherwise, if this PR doesn't represent a regression in this area, I would be inclined to integrate it anyway as an enhancement and fix the remaining work in follow-ups.

knutwannheden:
I implemented support for continuation lines in the property keys, but I still need to pull in all tests from the original PR. Also, I suspect there could still be an uncovered corner case, when there are continuation lines in the prefixes (key, equals, or value).

The current parser is terse, but quite hard to follow. Thinking about replacing it with a proper ANTLR parser or rewriting it to be more readable.

traceyyoshima:
Optionally, we can store the raw text, including the continuations, and then revise the get method to return the trimmed value.
For printing, we'd return the raw text with an additional getRawText or something along those lines.

That will allow recipes to check and set the value, replacing the continued text. As well as allow users to add continuations in the raw text if desired.

knutwannheden:
I am thinking it may make sense to only allow to set the raw value and thus have the unescaped value be a "view" of that (i.e. the getter would unescape it on the fly).

traceyyoshima:
> I am thinking it may make sense to only allow to set the raw value and thus have the unescaped value be a "view" of that (i.e. the getter would unescape it on the fly).

Agreed, that is essentially what I had in mind as well. Rather than having two fields, have one field where the get method will return the unescaped value. That will allow downstream recipes to use the same get and with methods without any changes. Then an additional method will return the continued text.

knutwannheden:
This PR was superseded by #3770 which now finally fixes this issue.

PR #3067, opened 2023-03-31
Correctly handle the many ways to describe a root source file

shanman190:

PR #3068, opened 2023-04-01
Optimize `StringUtils#matchesGlob()`

knutwannheden:
This Glob matching utility was using Ant's `SelectorUtils#match()` method, which uses `String#toCharArray()`. Often it is however faster to work with `String#charAt()`. Performance measurements indicate that the implementation is now more than twice as fast and since it is used very intensively in `rewrite-maven` (to match GAVs), the boost will be visible in some recipes.

### DIFF
--- a/rewrite-core/build.gradle.kts
+++ b/rewrite-core/build.gradle.kts
@@ -9,7 +9,6 @@ dependencies {
     compileOnly("org.eclipse.jgit:org.eclipse.jgit:5.13.+")
 
     implementation("org.openrewrite.tools:java-object-diff:latest.release")
-    implementation("org.apache.ant:ant:latest.release")
 
     implementation("io.quarkus.gizmo:gizmo:1.0.+")
 
--- a/rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite.internal;
 
-import org.apache.tools.ant.types.selectors.SelectorUtils;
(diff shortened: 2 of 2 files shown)


knutwannheden:
I did a search across the codebase and couldn't find any references to the packages provided by the Ant module, so I now also removed this 2 MiB dependency from `rewrite-core`.

knutwannheden:
I think we should probably split `matchesGlob()` into two versions: One for file globs (where we replace "wrong" path separators) and one for general purpose matching. The latter could be made a whole lot faster.

shanman190:
@knutwannheden, I think I'm going to implement the directory matching separately now. I'm running into a case where the current matching has some issues for #3067.

I'm seeing just a single case of path matching from `FindSourceFiles`, so at the moment in considering adding a `StringUtils#matchesGlob(Path, String)` to handle directory globs, then update `StringUtils#matchesGlob(String, String)` to only handle string globs as described here. WDYT?

knutwannheden:
Yes, that sounds about right. I haven't checked all the callers of this utility yet, but as I remember it, there are quite many. So we will carefully have to check that the other callers indeed don't match file paths.

If we change the first parameter to `Path` we should then probably also move the implementation to `org.openrewrite.PathUtils` which already has some methods along these lines.

shanman190:
Yeah, I checked over at least the `openrewrite/*` org and it appeared that all of the uses for `StringUtils#matchesGlob(String, String)` outside of `FindSourceFiles` are using actual string glob matches, so I think we're safe at least that far. Like you said, a vast majority of them are within `rewrite-maven` which is doing things like matching on `groupId` and `artifactId` by and large. Since I'm fixing up `FindSourceFiles`, that would be addresses as part of this.

PR #3070, opened 2023-04-01
Shared cursor root on all `TreeVisitor` returned by `Recipe#getVisitor`

jkschneider:
This is just a start. More changes would be required to ensure that `singleSourceApplicableTest` visitors also shared that same cursor root, but not sure these changes are worthwhile at this point.

cc / @knutwannheden

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
@@ -54,7 +54,7 @@ import java.util.function.Function;
  * @param <P> An input object that is passed to every visit method.
  */
 public abstract class TreeVisitor<T extends Tree, P> {
-    private Cursor cursor = new Cursor(null, Cursor.ROOT_VALUE);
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
```diff
@@ -54,7 +54,7 @@
  * @param <P> An input object that is passed to every visit method.
  */
 public abstract class TreeVisitor<T extends Tree, P> {
-    private Cursor cursor = new Cursor(null, Cursor.ROOT_VALUE);
+    Cursor cursor = new Cursor(null, Cursor.ROOT_VALUE);
```
knutwannheden, line 57: @jkschneider I suggest that we add the following few lines to the `Cursor(Cursor, Object)` constructor, so that we don't run into concurrency issues:
```java
        if (parent == null && value == ROOT_VALUE) {
            messages = new ConcurrentHashMap<>();
        }
```
What do you think?

PR #3074, opened 2023-04-02
Added recipe to replace Optional#isPresent to Optional#ifPresent

### LINKED ISSUE #3043: Replace `isPresent-> get` with `ifPresent`
```java

final JsonObject retryResponse = new JsonObject();

if (originalServiceResponseMessage.isPresent()) {

retryResponse.mergeIn((JsonObject) originalServiceResponseMessage.get().body());

}

```

```java

final JsonObject retryResponse = new JsonObject();
originalServiceResponseMessage.ifPresent(objectMessage -> retryResponse.mergeIn((JsonObject) objectMessage.body()));


```

srmalkan:
Closes #3043


rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/ReplaceIsPresentWithIfPresentTest.java
```diff
@@ -0,0 +1,211 @@
...
+                    import java.util.List;
+                    public class A {
+                        void method() {
+                            boolean x = true;
+                            List<Integer> list = new ArrayList<>();
+                            Optional<Integer> o = Optional.of(2);
+                            if(x){
+                                o.ifPresent((obj) -> {
+                                    list.add(obj);
+                                });
```
timtebeek, line 203: Could we chain [LambdaBlockToExpression](https://docs.openrewrite.org/reference/recipes/java/cleanup/lambdablocktoexpression) to get rid of the lambda block?

timtebeek:
Thanks a lot so far @srmalkan ! I've pushed a few small improvements (formatting & conventions) but also found two test cases that were not correctly handled yet. Would be great if you could look into those!

@yeikel originally opened #3043 , so perhaps he also has some insights to share here to influence the implementation.

srmalkan:
Hi Tim,

Thanks for highlighting edge cases and adding new test cases, what are your thoughts on nested Optional#IsPresent, should we skip it or should we handle it ?

srmalkan:
> 

Hi Tim, 

Yes we do have utility named VariableNameUtils, but I am facing problem to generate unique name since it asks for a cursor hence do not take into account the changes made by recipe while evaluating names.

knutwannheden:
I didn't review the code, but unless we check the accessed parameters and arguments for final or effectively final, we cannot safely apply this refactoring. Take this code for example:
```java
                int x = 0;
                if (Optional.of("foo").isPresent()) {
                    System.out.println(x);
                }
                x = 1;
```
Converting to `Optional#ifPresent()` would yield a compiler error, unless we also introduce a new effectively final variable declaration in a preceding statement. This would be kind of ugly, so I think instead we need to have effectively final calculation in place, so that we don't produce uncompilable code.

srmalkan:
Hi,

I have added a visitor to validate if the given block is lambda convertible or not. please have a look at it.
And also added a fix to handle nested optional#IfPresent but the unit test case is failing because, after each execution cycle, inner J.if becomes eligible for the transformation. I am not able to figure out what should I do in this case?

rewrite-java/src/main/java/org/openrewrite/java/cleanup/IsBlockLambdaConvertibleVisitor.java
```diff
@@ -0,0 +1,115 @@
...
+        return _return;
+    }
+
+    private void populateLocalVariables(Cursor cursor) {
+        J.MethodDeclaration methodDeclaration = cursor.firstEnclosing(J.MethodDeclaration.class);
+        if (methodDeclaration == null) {
+            throw new IllegalStateException("A method declaration is required in the cursor path.");
+        }
+
+        new JavaIsoVisitor<Map<Pair<String, JavaType>, Boolean>>() {
```
knutwannheden, line 73: I think this visitor needs to stop the recursion at "class boundaries". So when a class declaration, anonymous class or a lambda is encountered, the visitor should not descend into these, as they are not allowed to assign to the variables from the lexical scope in which they are declared, so no need to check them. Additionally, checking them is problematic, because they could introduce a new variable which shades the one you want to check.

In addition to the above comment, the check for reassigned variables should apply to method input parameters. So the following should not be convertible:
```java
class A {
    void method(int x) {
        x = 10;
        Optional<Integer> o = Optional.of(2);
        o.ifPresent((obj) -> {
            System.out.println(x);
        });
    }
}
```

Also, the visitor should follow all other scoping rules. So for example the following code should be convertible, because the x being assigned in the second block is a different variable:
```java
class A {
    void method() {
        {
            int x = 10;
            Optional<Integer> o = Optional.of(2);
            if (o.isPresent()) {
                System.out.println(x);
            }
        }
        {
            int x = 10;
            x = 20;
        }
    }
}
```
or a slightly less esoteric example, but with same principle idea:
```java
class A {
    void method() {
        if (true) {
            int x = 10;
            Optional<Integer> o = Optional.of(2);
            if (o.isPresent()) {
                System.out.println(x);
            }
        } else {
            int x = 10;
            x = 20;
        }
    }
}
```

rewrite-java/src/main/java/org/openrewrite/java/cleanup/IsBlockLambdaConvertibleVisitor.java
```diff
@@ -0,0 +1,115 @@
+package org.openrewrite.java.cleanup;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
```
knutwannheden, line 6: We would actually like to get rid of the Apache Commons dependencies. So I suggest you instead introduce a custom Lombok `@Value` class for this purpose.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/IsBlockLambdaConvertibleVisitor.java
```diff
@@ -0,0 +1,115 @@
+package org.openrewrite.java.cleanup;
```
knutwannheden, line 1: Please note that the copyright header is missing here.

knutwannheden (commented):
@srmalkan Thanks for looking into this! I have added a few review comments with a few examples where the recipe would currently either
1. produce code that doesn't compile (due to the discussed effectively final problem) or
2. not apply the recipe, even though it is applicable

The latter is not so much of a problem, as that can be handled as future improvements. But the former would be good if we could solve up-front.

knutwannheden:
> I have added a visitor to validate if the given block is lambda convertible or not. please have a look at it.
> And also added a fix to handle nested optional#IfPresent but the unit test case is failing because, after each execution cycle, inner J.if becomes eligible for the transformation. I am not able to figure out what should I do in this case?

I think you should not use `SearchResult#found()` to markup the `Optional#isPresent()` call in the if-condition. This could end up leaking out of the recipe. E.g. for the following example the marker would remain after applying the recipe:
```java
if (o.isPresent() && false) {
    list.add(o.get());
}
```
Instead you could here use cursor messaging, but in this case I would probably just inline the check of the if condition into `visitIf()` like so:
```java
            if (!OPTIONAL_IS_PRESENT.matches(__if.getIfCondition().getTree())) {
                return super.visitIf(_if, context);
            }
```
If you now remove the `isParentWithOptionalIsPresentMi()` check, I think it should basically work. One problem is that the `VariableNameUtils#generateVariableName()` utility won't quite work here, as it won't be able to find the newly created optionals, as they haven't been added to the parent LST yet. I think this is a limitation we should address in `VariableNameUtils`. Until then you could for example maintain your own set of variable names you have already assigned.

srmalkan:
Hi @knutwannheden,
Thanks for reviewing the code and suggesting improvements, the task looks challenging, will try to make it work.

knutwannheden:
@srmalkan Yes, it is rather challenging at the moment. We plan to add features which will make recipes like this one simpler, by allowing the recipe to query whether a variable is final or not.

FWIW, `VariableNameUtils#generateVariableName()` will now check the LST element represented by the cursor, so that a recipe like yours works properly. See this commit for details and an example: https://github.com/openrewrite/rewrite/commit/1afee4c5097c94bc4e8c62ba6cc14d76ce121458

timtebeek:
Quick note that this is pull request a likely candidate to move to the new https://github.com/openrewrite/rewrite-static-analysis repository after the upcoming 8.0 release, and might best be recreated there, referring back here for some of the discussion items.

kunli2:
@srmalkan, thanks a lot for your contribution! I believe this recipe will be useful,   I made some upgrades based on this PR and moved this to the repo `rewrite-static-analysis` in this [PR](https://github.com/openrewrite/rewrite-static-analysis/pull/118). All the comments are addressed there so we can close this here.

PR #3083, opened 2023-04-03
Add considering largest frequency group to calculate tabSize

kunli2:
Add considering the largest frequency group to calculate tabSize to improve the confidence of tabSize calculation.

Before this PR, we use a way based on statistical means to calculate tab size.
This PR introduced a way to calculate the `tabSize` based on the biggest group, which has the benefit of being less susceptible to outliers.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/style/Autodetect.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/style/Autodetect.java
@@ -257,45 +257,20 @@ public class Autodetect extends NamedStyles {
              *      PT = (NT / D), which means the percentage of tabs.
              * So
              *      useTabs = PT > 0.5
-             *
-             * (2) Details to solve X via formula (5)
-             * There are three scenarios here
-             * #1. If the code contains tabs only, D ~= NT, PT ~= 100%, and NW ~= 0, so X = 0/0, it's an unknown value,
-             *      then we use the default value 4. Because small errors exist in reality, it is impossible to strictly
-             *      satisfy the condition of (D - NT) being zero, so we use a threshold here to determine this case.
-             *      let's say PT > 80%.
-             * #2. If the code contains spaces only. NT ~= 0, X ~= NW / D.
-             * #3. Mixed tabs and spaces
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/style/Autodetect.java
```diff
@@ -229,6 +230,7 @@ public void decrementContinuationDepth() {
 
         public TabsAndIndentsStyle getTabsAndIndentsStyle() {
             /**
+             * Calculate tabSize based on the means
```
sambsnyd, line 233:
```suggestion
```

rewrite-java/src/main/java/org/openrewrite/java/style/Autodetect.java
```diff
@@ -288,9 +288,15 @@ public TabsAndIndentsStyle getTabsAndIndentsStyle() {
...
-            if (tabSize < 2) {
-                tabSize = 2;
+            // Calculate tabSize based on the frequency, pick up the biggest frequency group.
+            // Using frequency are less susceptible to outliers than means.
+            int moreFrequentTabSize = getBiggestGroupOfTabSize(deltaSpaceIndentFrequencies);
+
+            // If two statistic results fight with each other, it means there are mess indentations in the code, then we use the default.
+            if (moreFrequentTabSize > 0 && tabSize != moreFrequentTabSize) {
+                tabSize = 4;
             }
```
sambsnyd, line 298: So is this doing two different calculations of tab size, checking that they agree, and if they do not defaulting to 4?
Seems like we should be able to pick a single method to determine tab size that is good enough

kunli2: Yes, it should be OK to use the max frequency group to determine, let me verify if it works, going to remove the previous methods.

kunli2: Yeah, removed the previous method of tabSize calculating, and kept the new method only.

Issue #3088, opened 2023-04-04
[rewrite-maven] Dependency scope is Compile and Runtime rather than just Runtime


BoykoAlex:
The issue uncovered while upgrading **spring-petclinic** from Spring Boot 2.7 to Spring Boot 3.0

Consider **spring-petclinic** from https://github.com/BoykoAlex/spring-petclinic/tree/2.7.x (i.e. branch **2.7.x**

Upgrade it to spring boot 3.0. Note compiler errors for unresolved `jakarta.xml` in `Vet` and `Vets` classes. This is due to the missing `jakarta.xml.bind-api` dependency in the pom. The recipe `org.openrewrite.java.migrate.jakarta.JavaxXmlBindMigrationToJakartaXmlBind` from **rewrite-migrate-java** was supposed to add the dependency.

The issue seems to be coming from the fact that `MavenResolutionResult` for petclinic pom resolution has `jakarta.xml.bind-api` not only under **Runtime** scope (correct) but also under the **Compile** scope (wrong).

Therefore I have raised the issue against **rewrite-maven**

Unit test to paste into `AddDependencyTest` class in **rewrite-maven**:
```java
    @Test
    void noCompileScopeDependency() {
        rewriteRun(
          spec -> spec.recipe(addDependency("jakarta.xml.bind:jakarta.xml.bind-api:latest.release", "com.google.common.math.IntMath", true)),
          mavenProject(
            "project",
            srcMainJava(
              java(usingGuavaIntMath)
            ),
            pomXml(
              """
                    <project>
                      <modelVersion>4.0.0</modelVersion>
                      <groupId>org.springframework.samples</groupId>
                      <artifactId>spring-petclinic</artifactId>
                      <version>2.7.3</version>
                    
                      <parent>
                        <groupId>org.springframework.boot</groupId>
                        <artifactId>spring-boot-starter-parent</artifactId>
                        <version>3.0.5</version>
                      </parent>
                      <name>petclinic</name>
                    
                      <properties>
                        <jakarta-servlet.version>5.0.0</jakarta-servlet.version>
                    
                        <java.version>17</java.version>
                        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
                        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
                    
                        <webjars-bootstrap.version>5.1.3</webjars-bootstrap.version>
                        <webjars-font-awesome.version>4.7.0</webjars-font-awesome.version>
                    
[...]

PR #3097, opened 2023-04-06
No scope BOM dependency support

BoykoAlex:
Candidate fix for #3088 

Requires thorough review!


BoykoAlex:
I recall there were checks laso applying this PR to **rewrite-spring** and **rewrite-migrate-java**... don't see them now :-\

sambsnyd:
@BoykoAlex those checks weren't set up properly, so they would just run the latest snapshots, not necessarily with the changes in the pull request, on the downstream recipe jars. Gave a false sense of security

BoykoAlex:
I ran **rewrite-migrate-java** and **rewrite-spring** builds with this change. Tests seem to pass now.

PR #3103, opened 2023-04-11
added regex support to Properties\ChangePropertyValue, and added Yaml…

nmck257:
…\ChangePropertyValue with feature parity

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


nmck257:
PR mostly for visibility since I added an arg to an existing recipe

rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java
```diff
@@ -0,0 +1,145 @@
...
+    String newValue;
+
+    @Option(displayName = "Old value",
+            required = false,
+            description = "Only change the property value if it matches the configured `oldValue`.")
+    @Nullable
+    String oldValue;
+
+    @Option(displayName = "Regex",
+            description = "Default false. If enabled, `oldValue` will be interepreted as a Regular Expression, and capture group contents will be available in `newValue`",
```
mike-solomon, line 52: @nmck257 If you have some time at some point, I'm not following what this parameter does. Specifically, I don't understand what `capture group contents will be available in newValue` means.

nmck257: Sure, quick example:
Input text: "hi I'm Nick"
oldValue: "hi I'm (.+)"
newValue: "hello $1"
regex: true
Output text: "hello Nick"

mike-solomon: Thanks -- appreciate it!

Issue #3053, opened 2023-03-30
Gradle AddDependency `onlyIfUsing` fails to detect usage in Groovy


timtebeek:
As [reported by @dauer via Slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1679569790556479), with a minimal reproduction sample: https://github.com/dauer/openrewrite-gradle-adddependency

In short he as a [conditional `onlyIfUsing` before adding a dependency](https://github.com/dauer/openrewrite-gradle-adddependency/blob/69b5b772aae211b3baf695c08d175d7c46cd89ac/rewrite.yml#L6C9-L11), which should be satisfied in [MyClass.groovy#L6](https://github.com/dauer/openrewrite-gradle-adddependency/blob/main/src/main/groovy/MyClass.groovy#L6), possibly because we only evaluate JavaProjects;
https://github.com/openrewrite/rewrite/blob/2dbfdac6acd3bb09f0b60fbf2da2cebf6f2466ad/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java#L136-L144

We might be able to replicate his sample in a unit test for AddDependency & onlyIfUsing, similar to this existing test:
https://github.com/openrewrite/rewrite/blob/2dbfdac6acd3bb09f0b60fbf2da2cebf6f2466ad/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java#L70-L110

That currently uses https://github.com/openrewrite/rewrite/blob/2dbfdac6acd3bb09f0b60fbf2da2cebf6f2466ad/rewrite-java/src/main/java/org/openrewrite/java/Assertions.java#L142-L144

For which [there's not yet an equivalent in the Groovy Assertions](https://github.com/openrewrite/rewrite/blob/main/rewrite-groovy/src/main/java/org/openrewrite/groovy/Assertions.java#L142).

shanman190:
Ok, I've found the issue in particular that is causing this to fail...

In particular, there is no `JavaSourceSet` marker on the groovy source files. As a result, this prevents identification of which source set the class belongs to leads to an early exit for the `AddDependency` recipe.

The reason for there not being any `JavaSourceSet` marker is because:
1) Unlike the `JavaParser`, the `GroovyParser` doesn't add a default `SourceSet` to any files that it parses. 
2) The `rewrite-gradle-plugin` doesn't add a `SourceSet` marker while it's doing the parsing of the Gradle Groovy source set.

[...]

PR #3105, opened 2023-04-11
Remove automatic source set addition to compilation units

shanman190:
The purpose of this chunk of work is to fully align the way that `SourceSet` handling is performed. Presently, there are many ways that `SourceSet` instances are being handled. Below is a list of the places where this is occurring:

* `Java*Parser` implementations
* `KotlinParser` implementation
* Gradle build plugin
* Maven build plugin

Additionally, in the `GroovyParser` there is no `SourceSet` handling.

This change is important, so that we can properly add `SourceSet` markers when they are truly necessary and leave them missing when they should not be present. The _only_ place where this determination can be performed presently is by the build plugins themselves. A few examples of places where a `SourceSet` marker should *NOT* be present are:

* Gradle Groovy DSL (`*.gradle`)
* Gradle Kotlin DSL (`*.gradle.kts`; NOTE: not yet parsing these types)
* Maven Java Wrapper file (`.mvn/wrapper/MavenWrapperDownloader.java`)

While the above is not meant to by any means be an exhaustive list, it does illustrate that for the parsers that presently *ARE* doing `SourceSet` manipulation within their various `parseInputs(..)` implementations, there are cases with which we would desire for there to be no `SourceSet` marker added in that language parser implementation.

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
@@ -26,6 +26,8 @@ import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.gradle.Assertions.buildGradle;
 import static org.openrewrite.gradle.Assertions.settingsGradle;
+import static org.openrewrite.groovy.Assertions.groovy;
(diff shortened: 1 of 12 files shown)


shanman190:
Related to #3053. The `rewrite-gradle-plugin` and `rewrite-maven-plugin` PRs to follow will actually fix the issue itself.

Issue #3102, opened 2023-04-10
`/*__TEMPLATE_STOP__*/` shows when going from rewrite-recipe-bom 1.17.0 -> 1.18.1


timtebeek:
This is an issue on behalf of cucumber-jvm-migration; when [upgrading the rewrite-recipe-bom there to 1.18.1](https://github.com/cucumber/cucumber-jvm-migration/pull/6) we now see failures in a number of different tests.
![image](https://user-images.githubusercontent.com/1027334/230971235-828b0db6-755a-42f1-92ed-1251a9ba3db5.png)

As an example, in the bottom test we assert that the following is correctly converted:
```java
package com.example.app;

import io.cucumber.java8.En;

import static org.junit.jupiter.api.Assertions.assertEquals;

public class CalculatorStepDefinitions implements En {
    private RpnCalculator calc;

    public CalculatorStepDefinitions() {
        Given("a calculator I just turned on", () -> {
            calc = new RpnCalculator();
        });

        When("I add {int} and {int}", (Integer arg1, Integer arg2) -> {
            calc.push(arg1);
            calc.push(arg2);
            calc.push("+");
        });

        Then("the result is {double}", (Double expected) -> assertEquals(expected, calc.value()));
    }

    static class RpnCalculator {
        void push(Object string) {
        }

        public Double value() {
            return Double.NaN;
        }
    }
}
```

Which still mostly works, with the one exception.
![image](https://user-images.githubusercontent.com/1027334/230971826-beeea91c-057b-41b1-9a1b-e045b835c124.png)

The corresponding replacement is constructed here:
https://github.com/cucumber/cucumber-jvm-migration/blob/c7ecb188a335c9d2de30f6629cfb4fa9cd92763c/src/main/java/io/cucumber/migration/CucumberJava8StepDefinitionToCucumberJava.java#L116-L124

This used to work previously, and up to recently was included in rewrite-testing-frameworks. Unfortunately it's since been split off, which caused this delay in finding the issue.

kunli2:
@timtebeek is the `cucumber-jvm-migration` our owned project? I tried to pull up a PR to add some tests but it says I don't have permission to create PR.

timtebeek:
The missing semicolons were just added in https://github.com/cucumber/cucumber-jvm-migration/pull/6/commits/a254bd972793d2614636d554218ef40e1996ece3, which indeed seems to have resolved the template stop issues, thanks!

A new issue though is that the double block braces are no longer removed from:
[...]

PR #3106, opened 2023-04-12
Try to fix a template issue on `cucumber-jvm-migration`

kunli2:
This PR attempts to fix the issue https://github.com/openrewrite/rewrite/issues/3102
I think the issue was introduced by [this commit](https://github.com/openrewrite/rewrite/commit/626d8c645bb7b6b9d81e69a7985759701eb9b133#r108541429), 
This PR just patched against that specified problem, it can fix the template problem, but it's a patch fix. I don't know whether it's a correct change.  I think the issue needs @knutwannheden 's help to also take a look.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
@@ -264,7 +264,8 @@ public class BlockStatementTemplateGenerator {
                 before.insert(0, "{\n");
             }
 
-            if (prior instanceof Statement) {
+            if (prior == insertionPoint && prior instanceof Expression) {
+                // the template represents an expression, so we need to wrap it in a statement
                 after.append(';');
             }
             after.append('}');
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest.java
@@ -19,10 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Issue;
 import org.openrewrite.java.cleanup.IndexOfReplaceableByContains;
-import org.openrewrite.java.tree.Expression;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JavaType;
-import org.openrewrite.java.tree.TypeUtils;
+import org.openrewrite.java.tree.*;
 import org.openrewrite.test.RewriteTest;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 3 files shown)


sambsnyd (approved):

rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -264,7 +264,7 @@ private void template(Cursor cursor, J prior, StringBuilder before, StringBuilde
                 before.insert(0, "{\n");
             }
 
-            if (prior instanceof Statement) {
+            if (prior instanceof Statement && !(prior instanceof J.MethodDeclaration)) {
                 after.append(';');
             }
```
sambsnyd, line 269: `J.MethodDeclaration` and `J.ClassDeclaration` should not have a `;` appended, although it is generally not a syntax error to have one. (Does result in an extra empty statement being generated). 

There may be other exceptions to the rule, but those are the two that occur to me thinking about it.

shanman190: A number of control flow statements also come to mind as well (if, for, while, and switch to name a few)

knutwannheden: I was the one to add a semicolon after statements here. The problem as I remember it has to do with the fact that we have some expressions which can also be used as statements (e.g. method invocations) and unlike the Javac AST we don't wrap them in some kind of `ExpressionStatement` element.

I will take a look at this, add some more tests and possibly add a cleaner fix.

knutwannheden:
There is many aspects to this problem and all of them put together produce this behavior:
- Cucumber uses a JavaTemplate which basically looks like this: `void m() { #{any()} }`. Since `#{any()}` basically represents an expression the template really should include a terminating semicolon, as it is used as a statement.
- The `javac` parser is really quite permissive when constructing the AST and is often able to ignore missing semicolons and other minor syntax errors
- So even though the `javac` compiler successfully produces a "correct" AST for the template above (which for example could end up being translated to `/*__TEMPLATE__*/void m() { __P__./*__p0__*/intp() }/*__TEMPLATE_STOP__*/;` our Java parser then associates the `/*__TEMPLATE_STOP__*/` comment with the templated `p0` statement, because it scans forward until it finds the terminating semicolon.
- Whereas when we don't insert a semicolon after the method declaration, then our Java parser associates the comment with the method declaration (using `J.RightPadded`).

I agree that we should be removing the semicolon in this context (I will make a proposal). At the same time I think it would be good for Cucumber to also add a terminating semicolon to their templates, so that the Java source file being parsed is syntactically correct.

knutwannheden:
I slightly changed the fix, added a test case, and also corrected two similar templates we had in our code.

kunli2:
nice, Thanks @knutwannheden for looking into this!

PR #3119, opened 2023-04-16
Unwrap when if condition is a single parenthesized expression

timtebeek:
Needed for https://github.com/openrewrite/rewrite-migrate-java/blob/713e52ae29069ee187ff8abf7d292de17448b31b/src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java#L78


knutwannheden (approved):
LGTM.

PR #3128, opened 2023-04-17
Support for Lombok's @NoArgsConstructor in org.openrewrite.java.clean…

### LINKED ISSUE #3114: org.openrewrite.java.cleanup.HideUtilityClassConstructor is missing ignore annotation for @NoArgsConstructor(access = AccessLevel.PRIVATE)
(https://github.com/openrewrite/rewrite/blob/aaa06ff9f94b45f35263795dabf5d5dd665a9abf/rewrite-java/src/main/java/org/openrewrite/java/style/Checkstyle.java#L96) has 

```
    public static HideUtilityClassConstructorStyle hideUtilityClassConstructorStyle() {
        return new HideUtilityClassConstructorStyle(Arrays.asList(
                "@lombok.experimental.UtilityClass",
                "@lombok.Data"
        ));
    }
```

but doesn't ahve `@NoArgsConstructor(access = AccessLevel.PRIVATE)` so the compilation after transformation will fail with 

constructor Foo() is already defined in class Foo
@NoArgsConstructor(access = AccessLevel.PRIVATE)

blipper:
…up.HideUtilityClassConstructor

https://projectlombok.org/features/constructor

Fixes #3114

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/style/Checkstyle.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/style/Checkstyle.java
@@ -96,6 +96,7 @@ public class Checkstyle extends NamedStyles {
     public static HideUtilityClassConstructorStyle hideUtilityClassConstructorStyle() {
         return new HideUtilityClassConstructorStyle(Arrays.asList(
                 "@lombok.experimental.UtilityClass",
+                "@lombok.NoArgsConstructor",
                 "@lombok.Data"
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks a lot! This will be in the next release; we typically release every two to three weeks. Until then you can [use our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions) to get the fix.

PR #3137, opened 2023-04-18
New recipe to utilise ChangeTagValueVisitor capabilities for general XML files

m-brophy:
The desired endpoint is to be able to change the value of a specific XML tag, defined by an XPath elementName parameter, and optional oldValue and fileMatcher parameters

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagValue.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagValue.java
@@ -30,16 +30,16 @@ import org.openrewrite.xml.tree.Xml;
 public class ChangeTagValue extends Recipe {
     @Override
     public String getDisplayName() {
-        return "Change XML Tag Name";
+        return "Change XML Tag Value";
     }
 
     @Override
     public String getDescription() {
-        return "Alters the name of XML tags matching the provided expression.";
+        return "Alters the value of XML tags matching the provided expression.";
     }
 
     @Option(displayName = "Element name",
-            description = "The name of the element whose attribute's value is to be changed. Interpreted as an XPath Expression.",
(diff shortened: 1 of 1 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagValue.java
```diff
@@ -0,0 +1,91 @@
...
+@EqualsAndHashCode(callSuper = true)
+public class ChangeTagValue extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Change XML Tag Name";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Alters the name of XML tags matching the provided expression.";
```
timtebeek, line 38:
```suggestion
        return "Alters the value of XML tags matching the provided expression.";
```

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagValue.java
```diff
@@ -0,0 +1,91 @@
...
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.xml.tree.Xml;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class ChangeTagValue extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Change XML Tag Name";
```
timtebeek:
```suggestion
        return "Change XML tag value";
```
We might want to rename the display name for its sibling as well to use lowercase where possible.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagValue.java
m-brophy: ah - I just changed that stuff in commit no 2 :)

PR #3148, opened 2023-04-21
Add `pre-commit` checks to repo

JLLeitschuh:
This will automatically catch quick things like the license header missing.

This change intentionally avoids removing all trailing whitespace from the multi-line string tests, however, it will remove trailing whitespace from all other files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,16 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-commit-check:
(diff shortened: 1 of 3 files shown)


sambsnyd (commented):
I do not want to introduce any more tools other than OpenRewrite or the Moderne CLI for making code changes to our repositories. The license format plugin, dependabot, etc., are things we want to get rid of and replace with recipes.

I'm fine with adding newlines at eof and license headers, etc, it just has to be OpenRewrite/Moderne based mechanisms which accomplish it.

PR #3149, opened 2023-04-21
Rewrite rewrites Rewrite  🥘 🐕 

JLLeitschuh:
This adds an EndOfLineAtEndOfFile recipe, this recipe is applied
to this recpie using the rewrite plugin.

Dogfood :tada:


shanman190:
This also probably could have been handled via editorconfig to automatical handle it in the IDE permanently. On save, if the new line is absent any IDE that uses editorconfig will automatically add the new line on save or source format actions.

```
[*]
insert_final_newline = true
```

PR #3150, opened 2023-04-22
ClasspathScanningLoader should catch Throwable not Exception to catch Error

blipper:
Specifically for `java.lang.NoClassDefFoundError`

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
@@ -138,7 +138,7 @@ public class ClasspathScanningLoader implements ResourceLoader {
                     Recipe recipe = constructRecipe(recipeClass);
                     recipeDescriptors.add(recipe.getDescriptor());
                     recipes.add(recipe);
-                } catch (Exception e) {
+                } catch (Throwable e) {
                     logger.warn("Unable to configure {}", recipeClass.getName(), e);
                 }
             }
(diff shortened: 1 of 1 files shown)


knutwannheden:
@blipper Thanks for thr contribution. Of this is an actual issue you have encountered, it would be great if we could add a test case for it. Can you provide some more details regarding the use case?

blipper:
This is a bit niche but for reasons I needed to use custom dependency management with standard versions of maven central deps.  Antlr 4.10 has a backward incompatible change that means earlier versions will fail reading the 4.11 generated ones.  

But that doesn't really matter for the general case which is that if you are testing classloading any failure at class init will likely be an Error not Exception

knutwannheden:
@blipper Thanks for clarifying. I just wonder if we can't make the exception handling a bit more specific to the kind of errors (like the `NoClassDefFoundError` you mention), so that we don't unintentionally make it harder to debug other errors (e.g. an `OutOfMemoryError`). Possibly we could start with also adding `java.lang.LinkageError` here, which is a supertype of `NoClassDefFoundError` and e.g. `VerifyError`. What do you think?

blipper:
Trying to predict seems hard to be honest.  We have no control over the errors that will be thrown and is very implementation dependent.  If it is an oom it might be due to messed up static init and is really the problem of the classs (same with stack overflow).  If it is a general oom then the exception still gets logged and if we are really at the threshold then oom will likely occur again (or even during the exception/logging creation.

blipper:
We are basically implementing a plugin verification here with dynamic class loading.  Trying to guess and rethrow some higher class of error seems fragile and complex.  Stack overflow is recoverable and is by definition due to the code scanned.  Oom will either be due to the code or not.  If not then it will almost immediately get thrown again during next gc cycle in something else.  If there is a bad native library linked that corrupts it isn't going to throw an error anyways.  If it is some serious internal jvm error or whatever either it will be confined to the throw or it is so serious that the jvm will panic and exit.  

My .02 CAD

blipper:
99% of the time stack overflow is due to infinite recursion and user tuning doesn't solve anything

sambsnyd:
@blipper this makes sense to me. Thanks for the contribution!

Issue #1753, opened 2022-05-03
Rewrite-maven: ChangeParentPom recipe does not work correctly with multi-module maven projects


ferblaca:
I have problems applying the `ChangeParentPom` recipe in multi-module maven projects where the goal is to modify the parent version of the root module of type `pom`.

It seems that the problem is related to the number of modules the maven project contains...

For example, I have created a Recipe in the following [project ](https://github.com/ferblaca/OpenRewriteRecipe) with the goal of modifying the Spring Boot parent to version 2.6.7:

```
---
type: specs.openrewrite.org/v1beta/recipe
name: com.yourorg.ChangeParentPomExample
displayName: Change Maven Parent Pom example
recipeList:
  - org.openrewrite.maven.ChangeParentPom:
      oldGroupId: org.springframework.boot
      newGroupId: org.springframework.boot
      oldArtifactId: spring-boot-starter-parent
      newArtifactId: spring-boot-starter-parent
      newVersion: 2.6.7
      versionPattern:
      allowVersionDowngrades: false
```

And on the other hand, I have a [demo project](https://github.com/ferblaca/OpenRewriteDemo) with the rewrite plugin configured for the above active recipe which contains 6 modules, whose [pom.xml](https://github.com/ferblaca/OpenRewriteDemo/blob/main/pom.xml) of type `pom` has Spring Boot parent version 2.6.5 configured. When running `mvn rewrite:run` the expected result is to update the Spring-boot parent version to `2.6.7`, but this is not the case.

Only when I run the command with -N (`mvn rewrite:run -N`) or comment out all or most of the [modules](https://github.com/ferblaca/OpenRewriteDemo/blob/main/pom.xml#L21-L26), the recipe correctly updates the Spring Boot parent version to `2.6.7`.

Can you check it? thank you very much.

tkvangorder:
Thank you for the detailed example, we will look into this.

dmt:
Just a note if this is relevant… I've had some similar issues that got fixed with version 4.24.0 of the maven plugin

ferblaca:
> Just a note if this is relevant… I've had some similar issues that got fixed with version 4.24.0 of the maven plugin

thanks @dmt ! I'll try it and see if it has been corrected ;)

ferblaca:
@tkvangorder, 

as @dmt says, when doing the same test now with version `4.24.0` of the rewrite plugin it works correctly!!

Can you close the issue thank you!

[...]

PR #3155, opened 2023-04-24
Add sporadically failing multi-module ChangeParentPom tests

timtebeek:
For https://github.com/openrewrite/rewrite/issues/1753

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
@@ -17,6 +17,7 @@ package org.openrewrite.maven;
 
 import com.google.common.collect.Lists;
 import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


sambsnyd:
Are the failures problems downloading things from external services? Or because our treatment of what we download is in some way inconsistent?

timtebeek:
No hints yet in the current output, suspect it's caught somewhere. Hadn't yet made a plan how to tackle this one; welcome to have as look if you have the time (but guessing you don't :)).

timtebeek:
Ok this change in `RewriteTest`(!) https://github.com/openrewrite/rewrite/pull/3155/commits/8d7eff280c016aa7c47238898062c29457f81ce4 frustratingly only fixes the unit test, after noticing that tests failed when the order of processed source files was anything but parent-first. That means the test does not use the same mechanism that might lead to the errors seen in #1753 ; I was so far unable to find a direct parallel in `src/main`.

@sambsnyd Any chance you know of, or could help find a similar pattern in `src/main` that might help resolve #1753 ? Or do you want to let this rest until after the large iterable changes have been applied?

timtebeek:
Merged as we track the issue separately already, although some of the findings above might still apply. Best to have another look after 8.0

PR #3162, opened 2023-04-24
NoStaticImport recipe updates inner calls #3161

### LINKED ISSUE #3161: NoStaticImport recipe updates inner calls
From Slack: https://rewriteoss.slack.com/archives/C01A843MWG5/p1682347693901599

What about adding a configuration option to the NoStaticImport recipe so that it does not update method calls inside a class with a static method, like calls to "staticMethod" in class "C" updated to "C.staticMethod".

that sounds more like a bug to me than a need for a configuration option, if I'm understanding you correctly

Cimballi:
Fixes #3161

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/NoStaticImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/NoStaticImportTest.java
@@ -49,6 +49,43 @@ class NoStaticImportTest implements RewriteTest {
         );
     }
 
+    @Test
+    void removeLocalImports() {
+        rewriteRun(
+          spec -> spec.recipe(new NoStaticImport("*..* *(..)")),
+          java(
+            """
+              package org.openrewrite.java;
+              
+              import static org.openrewrite.java.TestNoStaticImport.method0;
+
+              public class TestNoStaticImport {
+
+                  public static void method0() {
+                  }
+
+                  public static void method1() {
+                      method0();
+                  }
+              }
+              """,
+            """
+              package org.openrewrite.java;
+
+              public class TestNoStaticImport {
+
+                  public static void method0() {
+                  }
+
+                  public static void method1() {
+                      method0();
+                  }
+              }
+              """
+          )
+        );
+    }
+
     @Nested
     class Retain {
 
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/NoStaticImportTest.java
```diff
@@ -47,4 +47,33 @@ void test() {
...
+
+        public class TestNoStaticImport{
+
+            public static void method0() {
+            }
+
+            public static void method1() {
+                method0();
+            }
+        }
```
timtebeek, line 76: If you want to assert there are no code changes, then you do not have to pass the second parameter.

In this case I only saw a change in whitespace (`TestNoStaticImport {` versus `TestNoStaticImport{`) which I don't think was intentional, correct?

timtebeek:
```suggestion
```

Cimballi: Ok, yes, correct, the before and after should be the same in that case.

timtebeek:
@knutwannheden is there anything you'd want to salvage from https://github.com/openrewrite/rewrite/commit/ac8e11184112d663c770c840c77d630a5556480a ?

knutwannheden:
@timtebeek No need. I started looking into allowing zo also removing statically imported fields, but that doesn't really work because the API requires a `MethodMatcher`.

rewrite-java/src/main/java/org/openrewrite/java/NoStaticImport.java
```diff
@@ -66,8 +66,17 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                     if (m.getMethodType() != null) {
                         JavaType.FullyQualified receiverType = m.getMethodType().getDeclaringType();
 
-                        RemoveImport<ExecutionContext> op = new RemoveImport<>(receiverType.getFullyQualifiedName() + "." + method.getSimpleName(),
-                                true);
+                        // Do not replace if receiverType is the same as surrounding class
+                        if (getCursor().firstEnclosing(J.ClassDeclaration.class).getType().equals(receiverType)) {
```
knutwannheden, line 70: I suggest to also add a test with a nested type accessing a static member from the outer type and vice-versa.

knutwannheden (approved):

PR #3177, opened 2023-04-29
`ReplaceConstantWithAnotherConstant` should keep static imports

knutwannheden:
To minimize the diff and maintain the original coding style the `ReplaceConstantWithAnotherConstant` recipe should replace static imports of the constant with another static import for the new constant and update `J.Identifier` references accordingly.


timtebeek (approved):
Nice!

Issue #3005, opened 2023-03-17
Add Gradle Enterprise Maven Extension recipe


alextu:
There's at the moment no recipe available to add Maven extensions to Maven projects.
Extensions are defined in a `.mvn/extensions.xml` file. The Gradle Maven extension requires as well a configuration file `.mvn/gradle-enterprise.xml`. Here's a sample repository setting the Gradle Enterprise extension: https://github.com/gradle/gradle-enterprise-build-config-samples/tree/main/common-gradle-enterprise-maven-configuration/.mvn

The recipe would allow the following attributes to be set:
- `version` (optional) for the Gradle Enterprise extension. If not set, try to get the latest
- `server` (mandatory)
- `allowUntrusted` (optional) 

The recipe should set some default build scan configuration
```
     <buildScan>
       <backgroundBuildScanUpload>false</backgroundBuildScanUpload>
       <publish>ALWAYS</publish>
     </buildScan>
```

So here's an example of using such a recipe, given:

```
- org.openrewrite.maven.AddGradleEnterprise:
   version: 1.16.5
   server: https://foo
   allowUntrusted: true
```

It would set in `.mvn/extensions.xml`:
```
<extensions>
    <extension>
        <groupId>com.gradle</groupId>
        <artifactId>gradle-enterprise-maven-extension</artifactId>
        <version>1.16.5</version>
    </extension>
</extensions>
```
and in `.mvn/gradle-enterprise.xml`

```
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<gradleEnterprise
    xmlns="https://www.gradle.com/gradle-enterprise-maven" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="https://www.gradle.com/gradle-enterprise-maven https://www.gradle.com/schema/gradle-enterprise-maven.xsd">
  <server>
    <url>https://foo</url>
    <allowUntrusted>true</allowUntrusted>
  </server>
  <buildScan>
    <backgroundBuildScanUpload>false</backgroundBuildScanUpload>
    <publish>ALWAYS</publish>
  </buildScan>
</gradleEnterprise>
```

As for the implementation, maybe a generic recipe to add Maven extensions would be valuable that this recipe would leverage.

PR #3193, opened 2023-05-03
Add recipe to add gradle enterprise maven extension to maven projects

kunli2:
closes https://github.com/openrewrite/rewrite/issues/3005

To implement the required recipe [here](https://github.com/openrewrite/rewrite/issues/3005) 

To integrate Gradle Enterprise Maven Extension into Maven projects, ensure that the `gradle-enterprise-maven-extension` is added to the `.mvn/extensions.xml` file if not already present. 
Additionally, configure the extension by adding the `.mvn/gradle-enterprise.xml` configuration file.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
@@ -1,0 +1,187 @@
+package org.openrewrite.maven;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.intellij.lang.annotations.Language;
+import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.xml.*;
+import org.openrewrite.xml.tree.Xml;
+
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class AddGradleEnterpriseMavenExtension extends Recipe {
+    private static final String GRADLE_ENTERPRISE_MAVEN_EXTENSION_ARTIFACT_ID = "gradle-enterprise-maven-extension";
+    private static final String EXTENSIONS_XML_PATH = ".mvn/extensions.xml";
+    private static final String GRADLE_ENTERPRISE_XML_PATH = ".mvn/gradle-enterprise.xml";
+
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Option(displayName = "Plugin version",
+        description = "An exact version number or node-style semver selector used to select the gradle-enterprise-maven-extension version.",
+        example = "3.x")
+    @Nullable
+    String version;
+
+    @Option(displayName = "Server URL",
+        description = "The URL of the Gradle Enterprise server.",
+        example = "https://ge.openrewrite.org/")
```
sambsnyd, line 69:
```suggestion
        example = "https://scans.gradle.com/")
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
```diff
@@ -0,0 +1,187 @@
...
+                                         "    xmlns=\"https://www.gradle.com/gradle-enterprise-maven\" " +
+                                         "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\n" +
+                                         "    xsi:schemaLocation=\"https://www.gradle.com/gradle-enterprise-maven" +
+                                         " https://www.gradle.com/schema/gradle-enterprise-maven.xsd\">\n" +
+                                         "  <server>\n" +
+                                         "    <url>%s</url>\n" +
+                                         "    <allowUntrusted>%b</allowUntrusted>\n" +
+                                         "  </server>\n" +
+                                         "  <buildScan>\n" +
+                                         "    <backgroundBuildScanUpload>false</backgroundBuildScanUpload>\n" +
```
sambsnyd, line 56: Why is this hardcoded to `false`? Is this parameter required? If we're not making it configurable would rather not set a value for it unless required

sambsnyd (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
kunli2: Yes, good question, I thought it was false by default, just asked the question [here](https://github.com/openrewrite/rewrite/issues/3005#issuecomment-1533593785)

kunli2: I saw @alextu is OOO until May 8, so maybe let us keep it as is (not sure if it's a required field and safe to remove it), and we can make another code update if we want to make it configurable.

PR #3194, opened 2023-05-04
Support git origin URLs that start with org-

timtebeek:
Fixes https://github.com/moderneinc/moderne-cli/issues/319

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
@@ -103,12 +103,10 @@ public class GitProvenance implements Marker {
             if (path.startsWith("/")) {
                 path = path.substring(1);
             }
-            int firstIndexOfSlash = path.indexOf('/');
-            int lastIndexOfSlash = path.lastIndexOf('/');
-            if (firstIndexOfSlash != lastIndexOfSlash) {
-                return path.substring(firstIndexOfSlash + 1, lastIndexOfSlash);
-            }
-            return path.substring(0, lastIndexOfSlash);
+            // Strip off any trailing repository name
+            path = path.substring(0, path.lastIndexOf('/'));
+            // Strip off any leading sub organization names
+            return path.substring(path.lastIndexOf('/') + 1);
         } catch (URISyntaxException e) {
             return null;
         }
--- a/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
@@ -96,7 +96,7 @@ public class GitProvenance implements Marker {
             return null;
         }
         String path;
-        if (origin.startsWith("git")) {
+        if (origin.startsWith("git") || origin.startsWith("org-")) {
             path = origin.substring(origin.indexOf(':') + 1);
         } else {
             path = URI.create(origin).getPath().substring(1);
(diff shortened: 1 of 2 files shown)


rpau (changes requested):
`org-` is not a URL pattern, so this solution is not robust. The requirement to solve the issue is to support user@host.xz:/path/to/repo.git/ urls

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -95,21 +97,20 @@ public String getOrganizationName() {
...
+                path = path.substring(1);
+            }
+            int firstIndexOfSlash = path.indexOf('/');
+            int lastIndexOfSlash = path.lastIndexOf('/');
+            if (firstIndexOfSlash != lastIndexOfSlash) {
+                return path.substring(firstIndexOfSlash + 1, lastIndexOfSlash);
+            }
+            return path.substring(0, lastIndexOfSlash);
+        } catch (URISyntaxException e) {
+            return null;
```
timtebeek, line 113: Wasn't sure about returning null here & in getRepositoryName

rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
```diff
@@ -57,7 +57,9 @@ private static Stream<String> remotes() {
           "https://github.com/openrewrite/rewrite.git",
           "file:///openrewrite/rewrite.git",
           "http://localhost:7990/scm/openrewrite/rewrite.git",
-          "git@github.com:openrewrite/rewrite.git"
+          //TODO "http://localhost:7990/scm/some/openrewrite/rewrite.git",
```
timtebeek, line 60: This is a case where we actually need a baseUrl to determine which is which.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -95,21 +97,20 @@ public String getOrganizationName() {
...
-            return "";
+        try {
+            URIish gituri = new URIish(origin);
+            String path = gituri.getPath();
+            if (path.startsWith("/")) {
+                path = path.substring(1);
+            }
+            int firstIndexOfSlash = path.indexOf('/');
+            int lastIndexOfSlash = path.lastIndexOf('/');
+            if (firstIndexOfSlash != lastIndexOfSlash) {
```
timtebeek, line 108: This isn't the best, as it fails on additional slashes such as:
`http://localhost:7990/scm/some/openrewrite/rewrite.git`
Those cases now also already requite a baseUrl, and are common in GitLab.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -87,29 +89,21 @@ public String getOrganizationName(String baseUrl) {
...
-        int secondSlashPos = path.lastIndexOf('/', firstSlashPos - 1);
-
-        if (secondSlashPos > -1) {
-            return path.substring(secondSlashPos + 1, firstSlashPos);
-        } else if (firstSlashPos > -1) {
-            return path.substring(0, firstSlashPos);
-        } else {
-            return "";
+        try {
+            String path = new URIish(origin).getPath();
```
rpau, line 100: nice!

rpau (approved):
LGTM

PR #3199, opened 2023-05-05
Fixes circular references serializing markers

rpau:
This error appears when we build the rewrite repository with the Moderne CLI

```
Caused by: java.lang.RuntimeException: java.io.UncheckedIOException: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Direct self-reference leading to cycle (through reference chain: java.util.ArrayList[272]->org.openrewrite.text.PlainText["markers"]->org.openrewrite.marker.Markers["markers"]->java.util.ArrayList[5]->org.openrewrite.marker.Markup$Warn["exception"]->java.lang.RuntimeException["cause"]->org.gradle.api.artifacts.ResolveException["cause"])
	at io.moderne.gradle.DelegatingSerializingProjectParser.unwrapInvocationException(DelegatingSerializingProjectParser.java:172)
	at io.moderne.gradle.DelegatingSerializingProjectParser.writeAstFile(DelegatingSerializingProjectParser.java:64)
	at io.moderne.gradle.ModerneAstTask.run(ModerneAstTask.java:91)
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
@@ -15,10 +15,7 @@
  */
 package org.openrewrite.internal;
 
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Optional;
-import java.util.StringJoiner;
+import java.util.*;
 
 public class ExceptionUtils {
     /**
@@ -48,18 +45,17 @@ public class ExceptionUtils {
     }
 
     public static boolean containsCircularReferences(Throwable exception) {
-        List<Throwable> causes = new LinkedList<Throwable>();
+        Set<Throwable> causes = Collections.newSetFromMap(new IdentityHashMap<>());
         causes.add(exception);
         boolean containsACircularReference = false;
         while (exception != null && exception.getCause() != null) {
             Throwable exceptionToFind = exception.getCause();
             if (exceptionToFind != null) {
-                Optional<Throwable> duplicate = causes.stream().filter(e -> e == exceptionToFind).findFirst();
-                if (duplicate.isPresent()) {
+
+                if (!causes.add(exceptionToFind)) {
                     containsACircularReference = true;
-                    exception = null;
+                    break;
                 } else {
-                    causes.add(exceptionToFind);
                     exception = exceptionToFind;
                 }
             } else {
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
@@ -15,6 +15,9 @@
  */
 package org.openrewrite.internal;
 
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
 import java.util.StringJoiner;
 
 public class ExceptionUtils {
@@ -42,5 +45,27 @@ public class ExceptionUtils {
             sanitized.add("  " + stackTraceElement);
         }
         return sanitized.toString();
+    }
+
+    public static boolean containsCircularReferences(Throwable exception) {
+        List<Throwable> causes = new LinkedList<Throwable>();
+        causes.add(exception);
+        boolean containsACircularReference = false;
+        while (exception != null && exception.getCause() != null) {
+            Throwable exceptionToFind = exception.getCause();
+            if (exceptionToFind != null) {
+                Optional<Throwable> duplicate = causes.stream().filter(e -> e == exceptionToFind).findFirst();
+                if (duplicate.isPresent()) {
+                    containsACircularReference = true;
+                    exception = null;
+                } else {
+                    causes.add(exceptionToFind);
+                    exception = exceptionToFind;
+                }
+            } else {
+                exception = null;
+            }
+        }
+        return containsACircularReference;
     }
 }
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
```diff
@@ -43,4 +46,26 @@ public static String sanitizeStackTrace(Throwable t, Class<?> until) {
         }
         return sanitized.toString();
     }
+
+    public static boolean containsCircularReferences(Throwable exception) {
+        List<Throwable> causes = new LinkedList<Throwable>();
```
knutwannheden, line 51: Detail: It would be more efficient with a `Set`. Specifically, I suggest we use `Collections.newSetFromMap(new IdentityHashMap<>())`.

rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
```diff
@@ -43,4 +46,26 @@ public static String sanitizeStackTrace(Throwable t, Class<?> until) {
...
+        causes.add(exception);
+        boolean containsACircularReference = false;
+        while (exception != null && exception.getCause() != null) {
+            Throwable exceptionToFind = exception.getCause();
+            if (exceptionToFind != null) {
+                Optional<Throwable> duplicate = causes.stream().filter(e -> e == exceptionToFind).findFirst();
```
knutwannheden, line 57: With a `Set` as described above we can here use `causes.contains(exceptionToFind)`. Even better:
```java
if (!causes.add(exceptionToFind)) {
    containsACircularReference = true;
    break;
} else {
    exception = exceptionToFind;
}
```

rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
```diff
@@ -43,4 +46,26 @@ public static String sanitizeStackTrace(Throwable t, Class<?> until) {
...
+                if (duplicate.isPresent()) {
+                    containsACircularReference = true;
+                    exception = null;
```
knutwannheden, line 60: Here we should instead just add a `break;` or even a `return true;`.

rewrite-java-test/src/test/java/org/openrewrite/java/RecipeMarkupDemonstrationTest.java
```diff
@@ -40,4 +50,46 @@ class Test {
...
+        @Override
+        public String getDisplayName() {
+            return "test recipe that produces an error";
+        }
+        @Override
+        public String getDescription() {
+            return "test recipe that produces an error using a circular exception cause.";
+        }
+        @Override
+        protected List<SourceFile> visit(List<SourceFile> before, ExecutionContext ctx) {
```
knutwannheden, line 81: For the OpenRewrite 8.0 upgrade it would be better if we here could implement a "normal" recipe implementing `getVisitor()` rather than `visit()`, as the latter requires modifications to compile. This is even easier by using the static `org.openrewrite.test.RewriteTest#toRecipe()` utility, as you then only need to supply the visitor rather than a full-blown recipe.

rewrite-core/src/main/java/org/openrewrite/internal/ExceptionUtils.java
rpau: Thanks! I did not know about the `IdentityHashMap`

knutwannheden (approved):
LGTM 👍

knutwannheden:
It appears like we still have this problem and I think it is related to this Jackson issue: https://github.com/FasterXML/jackson-databind/issues/1986.

I would like to suggest that we attempt to solve it like this instead:
```java
    static <T extends Tree> T error(T t, Throwable throwable) {
        Throwable temp = throwable;
        do {
            if (temp.getCause() == null) {
                temp.initCause(null);
                break;
            }
        } while ((temp = temp.getCause()) != null);

        return markup(t, new Markup.Error(randomId(), throwable));
    }
```

What isn't quite clear to me is why this apparently worked before. Or have we in the past never serialized the exceptions as part of the marker?

/cc @sambsnyd

rpau:
@knutwannheden How do you know that we still have this problem? 

I tested this with the CLI, but the last (Moderne) gradle build plugin was 29th April. To properly use this PR from the CLI, we also need to release the build plugin and test it with the rewrite repository. 

Regarding the solution, I already explored that path, and I thought that is not valid because:

1. If Jackson is using the getters instead of the properties to serialize the objects, then `getCause` will return null.
2. It does not prevents that there are cross references between Exceptions that are at different levels.

knutwannheden:
I am not 100% sure, but I think that the `Direct self-reference leading to cycle` message means that there is an object referencing itself here. While we could also try to detect cycles, I suspect that cycles in exceptions is mostly theoretical and am not convinced we need to check for that.

A Java exception which hasn't had its cause initialized will always reference itself and the `getCause()` method will then return `null`. This is what makes me believe that the problem you encountered was with an exception with an uninitialized cause. And for that case I don't think the current solution provides a fix.

But we should indeed build a new version of the Moderne Gradle plugin to verify if the problem is fixed or not.

I debugged this for a bit and the cause for all of this seems to be that the when the build plugin attempts to resolve the dependencies (using `rewrite-gradle-tooling-model`) then that fails and it is these exceptions which are being attached as markers to the source file it later attempts to serialize.

knutwannheden:
@rpau So I think what I proposed should fix the issue. If you think we should also be covering cycles involving multiple dependencies already now, we can also add that. But I really think this would be very unusual.

I would still like to understand what the root cause is of all this. It seems like an exception is getting added as markup to the LST which wasn't there before. I suspect this is a completely separate issue, which we also need to investigate at some point.

PR #3202, opened 2023-05-07
Add a failing test in ReplaceStringBuilderWithStringTest

pramud:


rpau:
Thanks for reporting this failing test!

timtebeek:
Thanks! We can indeed try to replicate the issue here, but any fix should go into the new rewrite-static-analysis, such that it's not lost with the upcoming 8.0

kunli2:
hi @pramud , thanks for looking into this, however, this case doesn't work because it doesn't compile.

I changed the code a bit to pass compile and it works.
```java

    @Test
    void withMethods() {
        rewriteRun(
          java(
            """
              class A {
                  void method() {
                      String key0 = new StringBuilder().append("test".toUpperCase()).append(3).append(count()).toString();
                  }
                  int count() {
                      return 10;
                  }
              }
              """
            ,
            """
              class A {
                  void method() {
                      String key0 = "test".toUpperCase() + 3 + count();
                  }
                  int count() {
                      return 10;
                  }
              }
              """
          )
        );
    }
```

kunli2:
closing out this, please let me know if you still have any concerns or new findings. Thanks in advance.

PR #3207, opened 2023-05-07
Show diff on RewriteTest comparison failures

timtebeek:
I've spent a bit too much time looking for whitespace changes on long before/after blocks. This ought to help, even if it's crude. Figured get feedback on if we want something like this first, and see from there how best to phrase this in the message.

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -423,7 +423,11 @@ public interface RewriteTest extends SourceSpecs {
                             String expected = sourceSpec.noTrim ?
                                     expectedAfter :
                                     trimIndentPreserveCRLF(expectedAfter);
-                            assertThat(actual).as("Unexpected result in \"" + result.getAfter().getSourcePath() + "\"").isEqualTo(expected);
+                            assertThat(actual)
+                                    .as(() -> String.format("Unexpected result in \"%s\"%s",
+                                            result.getAfter().getSourcePath(),
+                                            result.diff().isEmpty() ? "" : "\n" + result.diff()))
+                                    .isEqualTo(expected);
                             sourceSpec.eachResult.accept(result.getAfter(), testMethodSpec, testClassSpec);
                         } else {
                             if (result.diff().isEmpty() && !(result.getAfter() instanceof Remote)) {
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -15,6 +15,8 @@
  */
 package org.openrewrite.test;
 
+import lombok.RequiredArgsConstructor;
+import lombok.experimental.Delegate;
 import org.assertj.core.api.SoftAssertions;
 import org.openrewrite.*;
 import org.openrewrite.config.CompositeRecipe;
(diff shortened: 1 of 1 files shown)


rpau (approved):
Nice! 👍

knutwannheden:
I've also thought about this before. Can you show a screenshot of what it actually looks like? I suppose this means that the hyperlink in IDEA for the diff view no longer shows up?

knutwannheden:
@timtebeek To answer my own question: The link is still there, so I think this is great. Only, I wonder if we are diffing the "right" thing? The diff is between the before and after rather than between the expected and actual afters.

image

timtebeek:
Well spotted there that the diff is not the diff between expected/actual that I had thought it'd be.. slightly complicates the logic, but since it's in a supplier we can defer that until only when it's needed.

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -423,7 +423,11 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
                             String expected = sourceSpec.noTrim ?
                                     expectedAfter :
                                     trimIndentPreserveCRLF(expectedAfter);
-                            assertThat(actual).as("Unexpected result in \"" + result.getAfter().getSourcePath() + "\"").isEqualTo(expected);
+                            assertThat(actual)
+                                    .as(() -> String.format("Unexpected result in \"%s\"%s",
+                                            result.getAfter().getSourcePath(),
+                                            result.diff().isEmpty() ? "" : "\n" + result.diff()))
```
timtebeek, line 429: This is not the diff between expected after & actual ; and it seems a tad more difficult the create that using the components going into `Result.diff()` than anticipated. Looking at alternatives now.

timtebeek: `org.assertj.core.util.diff.DiffUtils.diff(List, List)` is a bit clunky, but might work; not a fan of the output yet:
```
org.opentest4j.AssertionFailedError: [Unexpected result in "a/b/goodbye.txt"
Changed content at line 1:
expecting:
  ["hello world2"]
but was:
  ["hello world"]
] 
expected: "hello world2"
 but was: "hello world"
Expected :"hello world2"
Actual   :"hello world"
<Click to see difference>
```

knutwannheden: Can we create a new `Result` object based on the expected and actual afters and use that to create the diff?

timtebeek: It's a bit awkward of an API, since it's so specific to diffing SourceFiles in memory using JGit, and the SourceFile could come from any parser. There's also no easy way to do "result.after().withText(actual)` for instance.

timtebeek: I think at most you could do something like
```java
assertThat(actual)
        .as(() -> {
            SourceFile after = result.getAfter();
            PlainText actualSourceFile = PlainText.builder()
                    .id(after.getId())
                    .charsetBomMarked(after.isCharsetBomMarked())
                    .fileAttributes(after.getFileAttributes())
                    .charsetName(after.getCharset().name())
                    .checksum(after.getChecksum())
                    .markers(after.getMarkers())
                    .sourcePath(after.getSourcePath())
                    .text(actual)
                    .build();
            String diff = new Result(after, actualSourceFile, emptyList()).diff();
            return String.format("Unexpected result in \"%s\"%s",
                    after.getSourcePath(),
                    diff.isEmpty() ? "" : "\n" + diff);
        })
        .isEqualTo(expected);
```
but that's not really what I'd hope to do here.

timtebeek:
Managed to still use the regular `Result#diff()` through a Lombok `@Delegate` in 3743b5e, which now results in:
```
[Unexpected result in "file.yaml"
index bfe8fe2..287bbfa 100644
--- a/file.yaml
+++ b/file.yaml
@@ -4,5 +4,4 @@ 
     - name: us-foo-1
     - name: us-foo-2
   other:
-    - name: outerspace-1
-    - foo
\ No newline at end of file
+    - name: outerspace-1
\ No newline at end of file
] 
expected: 
  "prod:
    regions:
      - name: us-bar-1
      - name: us-foo-1
      - name: us-foo-2
    other:
      - name: outerspace-1
      - foo"
 but was: 
  "prod:
    regions:
      - name: us-bar-1
      - name: us-foo-1
      - name: us-foo-2
    other:
      - name: outerspace-1"
```

PR #3211, opened 2023-05-08
fix: parse environment properties in maven repositories

### LINKED ISSUE #3142: `MavenPomDownloader` fails to understand environment properties in `settings.xml`
I have the following configuration in my `settings.xml` file

```xml

<servers>
        <server>
            <id>corporate</id>
            <username>${env.ARTIFACTORY_USERNAME}</username>
            <password>${env.ARTIFACTORY_PASSWORD}</password>
        </server>
        <server>
            <id>snapshots</id>
            <username>${env.ARTIFACTORY_USERNAME}</username>
            <password>${env.ARTIFACTORY_PASSWORD}</password>
        </server>
    </servers>
```

Where `${env.ARTIFACTORY_USERNAME}` and `${env.ARTIFACTORY_PASSWORD}` are resolved at runtime by maven **if present**

My Artifactory instance allows anonymous reads, but expects authentication for writes. Due to that, these environment properties are only defined in the CI environment and are not needed locally

When rewrite runs, it interprets these values **literally** and tries to authenticate using `${env.ARTIFACTORY_USERNAME}:${env.ARTIFACTORY_PASSWORD}` (resolves to `JHtlbnYuQVJUSUZBQ1RPUllfVVNFUk5BTUV9OiR7ZW52LkFSVElGQUNUT1JZX1BBU1NXT1JEfQ==` in base64)

That leads to `401` as my server does not allow that combination

image

I'd expect `MavenPomDownloader` to ignore these values when they are not defined just like Maven does

This was discovered in this issue : https://github.com/openrewrite/rewrite-maven-plugin/issues/443

yeikel:
Fixes #3142


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
```diff
@@ -116,6 +118,15 @@ public static class Builder {
...
         }
 
+        @VisibleForTesting
+        @Nullable
+        public static String getEnvironmentPropertyName(@Nullable String rawProperty) {
+            if (rawProperty == null){
+                return null;
+            }
+            return rawProperty.replace("${env.", "").replace("}", "");
+        }
```
yeikel, line 128: This is the best I was able to come up with to be able to test this method for the positive scenario. 

I did not find any test in this source code that works with environment properties and it is currently enabled

yeikel:
The test failures seem unrelated to my changes

timtebeek (approved):

PR #3212, opened 2023-05-08
fix: skip blocked repository

### LINKED ISSUE #3141: MavenPomDownloader seems to retry indefinitely for unresolvable urls
I have the following configuration in my `settings.xml`

```xml

<mirror>
            <!-- Blocks and mirrors for third-party repositories -->
            <id>blocked-third-party</id>
            <url>http://0.0.0.0</url>
            <mirrorOf>apache.snapshots,sonatype-nexus-snapshots,snapshots-repo,vertx-snapshots-repository</mirrorOf>
        </mirror>
```

This is a common pattern implemented in many different ways (localhost is also used). See https://stackoverflow.com/questions/67833372/getting-blocked-mirror-for-repositories-maven-error-even-after-adding-mirrors

The challenge is that `MavenPomDownloader` seems to retry indefinitely never resolving that url as it always returns `connection refused`

I support that we should retry because many repositories are slow, but we should probably have a limit

image

yeikel:
Closes #3141

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -665,6 +665,11 @@ public class MavenPomDownloader {
             if (repository.getUri().contains("${")) {
                 return null;
             }
+            // Skip blocked repositories
+            // See https://github.com/openrewrite/rewrite/issues/3141
+            else if (repository.getUri().contains("0.0.0.0")) {
+                return null;
+            }
 
             String originalUrl = repository.getUri();
             if ("file".equals(URI.create(originalUrl).getScheme())) {
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
@@ -87,6 +87,18 @@ class MavenPomDownloaderTest {
     }
 
     @ParameterizedTest
+    @Issue("https://github.com/openrewrite/rewrite/issues/3141")
+    @ValueSource(strings = {"http://0.0.0.0","https://0.0.0.0","0.0.0.0:443"})
+    void skipBlockedRepository(String url) {
+        var downloader = new MavenPomDownloader(emptyMap(), ctx);
+        MavenRepository oss = downloader.normalizeRepository(
+          MavenRepository.builder().id("myRepo").uri(url).build(),
+          null);
+
+        assertThat(oss).isNull();
+    }
+
+    @ParameterizedTest
     @ValueSource(ints = {500, 400})
     void normalizeAcceptErrorStatuses(Integer status) {
         var downloader = new MavenPomDownloader(emptyMap(), ctx);


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -665,6 +665,11 @@ public MavenRepository normalizeRepository(MavenRepository originalRepository, @
             if (repository.getUri().contains("${")) {
                 return null;
             }
+            // Skip blocked repositories
+            // See https://github.com/openrewrite/rewrite/issues/3141
+            else if (repository.getUri().contains("0.0.0.0")) {
+                return null;
+            }
```
yeikel, line 672: This solves the issue of blocked repositories, but does not fix that our retry configuration does not have a limit

```java

private static final RetryConfig retryConfig = RetryConfig.custom()
            .retryOnException(throwable -> throwable instanceof SocketTimeoutException ||
                                           throwable instanceof TimeoutException)
            .build();


```

We can configure a limit using `maxAttempts`, but I am just not sure what value I should set and that's why I did not include it with this change



```java

private static final RetryConfig retryConfig = RetryConfig.custom()
            .retryOnException(throwable -> throwable instanceof SocketTimeoutException ||
                                           throwable instanceof TimeoutException)
            .maxAttempts(<int>)
            .build();


```

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -60,6 +60,7 @@ public class MavenPomDownloader {
     private static final RetryConfig retryConfig = RetryConfig.custom()
             .retryOnException(throwable -> throwable instanceof SocketTimeoutException ||
                                            throwable instanceof TimeoutException)
+            .maxAttempts(5) // what should we set?
```
yeikel, line 63: What value should I set? I picked `5` arbitrarily, Maybe we can make this configurable with a system property?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -665,6 +666,11 @@ public MavenRepository normalizeRepository(MavenRepository originalRepository, @
             if (repository.getUri().contains("${")) {
                 return null;
             }
+            // Skip blocked repositories
+            // See https://github.com/openrewrite/rewrite/issues/3141
+            else if (repository.getUri().contains("0.0.0.0")) {
+                return null;
+            }
```
yeikel, line 673: Trying to resolve these is a waste of time so it should be safe to skip them. Unless you think that we should just use the retry configuration?

timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
timtebeek: I think it's fine to short circuit here indeed; might be a slightly specific, but clearer than a retry failure.

timtebeek:
Thanks again @yeikel !

Issue #3217, opened 2023-05-08
FR: Support constructor rewrite for DeleteMethodArgument recipe


thomaszub:
Hi,

the recipe `DeleteMethodArgument` does not support the rewriting of constructors despite the method pattern allows constructor rules. This would be a useful feature to add.

I make a PR.

Kind regards
Thomas

PR #3218, opened 2023-05-08
feat: add visit for constructor calls (#3217)

thomaszub:


rewrite-java/src/main/java/org/openrewrite/java/DeleteMethodArgument.java
```diff
@@ -87,11 +88,22 @@ public DeleteMethodArgumentVisitor(MethodMatcher methodMatcher) {
...
+        }
+
+        @Override
+        public J.NewClass visitNewClass(J.NewClass newClass, ExecutionContext ctx) {
+            J.NewClass n = super.visitNewClass(newClass, ctx);
+            return (J.NewClass) visitMethodCall(n);
+        }
+
+        private MethodCall visitMethodCall(MethodCall methodCall) {
+            MethodCall m = methodCall;
```
timtebeek, line 101:
```suggestion
        private MethodCall visitMethodCall(J.MethodCall methodCall) {
            J.MethodCall m = methodCall;
```

rewrite-java/src/main/java/org/openrewrite/java/DeleteMethodArgument.java
```diff
@@ -23,6 +23,7 @@
 import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.MethodCall;
```
timtebeek, line 26:
```suggestion
```

timtebeek (approved):
Looking good! Thanks for immediately providing a fix.

rpau (approved):
LGTM after the suggestions from Tim are addressed

rewrite-java/src/main/java/org/openrewrite/java/DeleteMethodArgument.java
```diff
@@ -87,11 +87,22 @@ public DeleteMethodArgumentVisitor(MethodMatcher methodMatcher) {
...
+            return (J.MethodInvocation) visitMethodCall(m);
+        }
+
+        @Override
+        public J.NewClass visitNewClass(J.NewClass newClass, ExecutionContext ctx) {
+            J.NewClass n = super.visitNewClass(newClass, ctx);
+            return (J.NewClass) visitMethodCall(n);
+        }
+
+        private MethodCall visitMethodCall(J.MethodCall methodCall) {
```
timtebeek, line 99:
```suggestion
        private J.MethodCall visitMethodCall(J.MethodCall methodCall) {
```

thomaszub:
@timtebeek The `MethodCall` is a separate interface not located in the `J` class. I think the suggested change would not work. This is also the reason why I added the overrides to `J.NewClass` and `J. MethodInvocation`.

PR #3224, opened 2023-05-09
Add guard for loading yaml source, gracefully handling the loading error.

kunli2:
Add guard for loading yaml source, gracefully handling the loading error.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
@@ -29,6 +29,8 @@ import org.openrewrite.internal.PropertyPlaceholderHelper;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.style.Style;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -167,17 +169,19 @@ public YamlResourceLoader(InputStream yamlInput, URI source, Properties properti
...
+            for (Object resource : yaml.loadAll(yamlSource)) {
+                if (resource instanceof Map) {
+                    @SuppressWarnings("unchecked") Map<String, Object> resourceMap = (Map<String, Object>) resource;
+                    if (resourceType.equals(ResourceType.fromSpec((String) resourceMap.get("type")))) {
+                        resources.add(resourceMap);
+                    }
                 }
             }
+        } catch (Exception e) {
+            logger.error("Loading yaml {} type failed, yaml source: {}", resourceType, yamlSource);
```
rpau, line 183: should we print the error message as well to understand what is the parsing error?

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -167,17 +169,19 @@ public YamlResourceLoader(InputStream yamlInput, URI source, Properties properti
 
     private Collection<Map<String, Object>> loadResources(ResourceType resourceType) {
```
rpau, line 170: Can we add a test to validate that this do not happen again? It is fine if we want to include it using a different PR

rpau (commented):
Minor comments. Otherwise 👍

Issue #3227, opened 2023-05-10
Gradle change dependency groupId/artifactId ignores platform dependencies


ge0ffrey:
A recipe to change the gradle groupId (or the artifactId) does not alter dependencies in `platform(...)`.

For example, change the gradle groupId `org.optaplanner` to `ai.timefold.solver`:

```
...
dependencies {
  implementation platform("org.optaplanner:optaplanner-bom:9.37.0.Final")
  implementation "org.optaplanner:optaplanner-core"
   ...
}
```
changes to, actual:
```
...

dependencies {
  implementation platform("org.optaplanner:optaplanner-bom:...")
  implementation "ai.timefold.solver:timefold-solver-core"
   ...
}
```
but it should be, expected:
```
...

dependencies {
  implementation platform("ai.timefold.solver:timefold-solver-bom:...")
  implementation "ai.timefold.solver:timefold-solver-core"
   ...
}
```

ge0ffrey:
Reproducer (not a fix): https://github.com/openrewrite/rewrite/pull/3228

PR #3228, opened 2023-05-10
Reproducer for #3227 Gradle change dependency ignores platform

ge0ffrey:
This is a reproducer for #3227 Gradle change dependency groupId/artifactId ignores platform dependencies

It does not fix the issue.

@sambsnyd Tim said I should tag you as you might be working in this corner and might want to be aware of it.


shanman190:
@ge0ffrey, would you mind trying this out in the latest snapshots/main? I believe that I just fixed it for a different, but similar, use case.

https://github.com/openrewrite/rewrite/commit/70f0bed112f343742717521924b08235900ec301

shanman190:
So this is only partially fixed. If a configuration is set, then it would fail for that case. Both test cases here do not have a configuration specified, so they should succeed now.

ge0ffrey:
Thank you @shanman190 !

PR #3229, opened 2023-05-10
Fixes serialization issues with Exception#cause [rework]

rpau:
Context: https://github.com/openrewrite/rewrite/pull/3199

I merged assuming that GH Actions run in the previous commit, but there were tests errors. So I reverted the commit and this PR fixes the tests errors.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
@@ -58,6 +58,7 @@ public interface Markup extends Marker {
     }
 
     static Throwable exceptionWithoutSelfReferences(Throwable throwable) {
+        // remove self-references from the cause chain to avoid serialization problems from the Moderne plugins
         Throwable temp = throwable;
         do {
             if (temp.getCause() == null) {
--- a/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
@@ -50,19 +50,23 @@ public interface Markup extends Marker {
     }
 
     static <T extends Tree> T error(T t, Throwable throwable) {
-        if (ExceptionUtils.containsCircularReferences(throwable)) {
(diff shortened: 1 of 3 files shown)


knutwannheden (approved):

rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
```diff
@@ -50,19 +50,28 @@ default String print(Cursor cursor, UnaryOperator<String> commentWrapper, boolea
...
+                    temp.initCause(null);
+                } catch (IllegalStateException ignored) {
+                    //ignore, the cause was already initialized to null
+                }
+                break;
+            }
+        } while ((temp = temp.getCause()) != null);
+
+        return throwable;
     }
```
JLLeitschuh, line 75: Isn't this a rendering issue, not an issue in the `Markup.Warn` creation problem? This seems like something that should be fixed in rendering or serialization, not by messing with the exception that gets returned by the AST.

Basically, this looks like you're fixing the wrong problem

knutwannheden: AFAICT the serialization problem we've seen is due to https://github.com/FasterXML/jackson-databind/issues/1986. I don't know if I understand https://github.com/FasterXML/jackson-databind/issues/1986#issuecomment-378692915 correctly, but also looking at https://github.com/FasterXML/jackson-databind/issues/877 it seems like it should be possible to define and register a mixin similar to this:
```java
public abstract class ThrowableMixin {
    @JsonIgnore private Throwable cause;
    @JsonIgnore public abstract Throwable getCause();
}
```
I don't know if that will work or if it will have the effect of not serializing any of the exception causes.

This being said, I agree with you that it would be better if we could get Jackson to properly serialize exceptions without having to muck with them like this.

rpau: I think that you are right. We can create instead a jackson module to serialize exceptions :) 
Closing the PR for now.

PR #3233, opened 2023-05-11
Add options to the Add Gradle Enterprise Maven recipe

alextu:
Follow up on https://github.com/openrewrite/rewrite/pull/3193#discussion_r1184183668
This PR adds some optional settings:
- `captureGoalInputFiles`
- `uploadInBackground`
- `publishCriteria`

An optional setting not set will not result in a xml element in the `gradle-enterprise.xml` configuration, this is implemented by using Jackson xml instead of plain string formatting.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
@@ -15,12 +15,18 @@
  */
 package org.openrewrite.maven;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.intellij.lang.annotations.Language;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.maven.internal.MavenXmlMapper;
 import org.openrewrite.xml.*;
 import org.openrewrite.xml.tree.Xml;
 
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
```diff
@@ -158,6 +183,59 @@ protected List<SourceFile> visit(List<SourceFile> before, ExecutionContext ctx)
...
+    private String gradleEnterpriseConfiguration() {
+        BuildScanConfiguration buildScanConfiguration = buildScanConfiguration();
+        ServerConfiguration serverConfiguration = new ServerConfiguration(server, allowUntrustedServer);
+        try {
+            ObjectMapper objectMapper = MavenXmlMapper.writeMapper();
+            objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
+            objectMapper.setSerializationInclusion(JsonInclude.Include.NON_ABSENT);
+            return objectMapper.writeValueAsString(new GradleEnterpriseConfiguration(serverConfiguration, buildScanConfiguration));
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
```
alextu, line 225: Not sure how those kind of exceptions are usually handled in recipes ?

rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
```diff
@@ -85,13 +74,50 @@ public class AddGradleEnterpriseMavenExtension extends Recipe {
...
+            description = "When set to `always` the extension will publish build scans of every single build. " +
+                    "This is the default behavior when omitted." +
+                    "When set to `failure` the extension will only publish build scans when the build fails. " +
+                    "When set to `demand` the extension will only publish build scans when explicitly requested.",
+            required = false,
+            valid = {"always", "failure", "demand"},
+            example = "true")
+    @Nullable
+    PublishCriteria publishCriteria;
+    public enum PublishCriteria {
```
timtebeek, line 110:
```suggestion
    PublishCriteria publishCriteria;

    enum PublishCriteria {
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddGradleEnterpriseMavenExtension.java
timtebeek: I'm seeing a mix of [`ctx.getOnError().accept(e)`](https://github.com/search?q=org%3Aopenrewrite%20ctx.getOnError&type=code), [`Markup.error(e)`](https://github.com/search?q=org%3Aopenrewrite+Markup.error&type=code) and [`throw new RuntimeException`](https://github.com/search?q=org%3Aopenrewrite+%22throw+new+RuntimeException%22&type=code), and possibly others.

I think it will mostly depend on if your recipe will be executed as part of a larger chain, in which case you might want to go for "elegant" handing over just throwing a runtime exception. I'm guessing in this case you're fine with a RuntimeException, since we're also not expecting a lot of different failure modes here.

timtebeek:
And thank you for this improvement; I'm going to merge this one as it is; if there's any feedback from Kun we can pick that up separately, as we did after the last one.

Issue #3234, opened 2023-05-11
Rewrite 8 migration requirements 


kunli2:
Background of rewrite-8
Background for the impetus for rewrite 8. It is to support large repositories. Before 8, all LST for a repository needed to be loaded into memory and available for LST to operate.  For large repos like Shelter’s e-commerce or Block’s franklin, this would not fit into memory. So we needed to redesign everything because the constraint now is that only part of LST is available in memory at a time and you don’t have random access - therefore LST needs to be scanned first to collect all info and then edit LST.

---
Rewrite 7 -> 8 change log highlights
1. All applicable test methods `org.openrewrite.Recipe getSingleSourceApplicableTest()`  and `org.openrewrite.Recipe getApplicableTest() `methods are removed, introduced new `Preconditions` APIs as a replacement.
2. The method  `Recipe visit(List<SourceFile> before, ExecutionContext ctx)` that is used to visit multiple sources has been removed in rewrite 8.  This method is used to change files, collect information, or generate new files. 
    There are 3 phases in the new rewrite 8 API design. 
        - (1) Scanning phase 
        - (2) generating phase 
        - (3) Editing phase
    basically, the Scanning phase doesn't make any code change, but just does information collection, rewrite8 introduces a new `ScanningRecipe` class to support this, For the generating phase, there is a new method .
a `ScanningRecipe` may first scan a repository and study it in one pass over the repository's source files before, in another pass, it decides how to transform the code. New source file generation is part of this type as well, since in almost every case we check that a file doesn't yet exist (and perhaps other conditions) before deciding to generate a file.
3. applicable tests will not be supported for yaml recipes anymore.
4. JavaTemplate upgrade, introduced context free JavaTemplate, @knutwannheden may provide more precise information. 
5. All clean up recipes are moved to repository `rewrite-static-analysis` and the package name is changed from `org.openrewrite.java.cleanup` to `org.openrewrite.staticanalysis`
6. Method `JavaSourceFile org.openrewrite.java.JavaVisitor.visitJavaSourceFile(JavaSourceFile cu, P p)` is removed, we need to use `J org.openrewrite.java.TreeVisitor.visit(@Nullable Tree tree, P p)` instead.
7. Some other minor changes
    - Methods `org.openrewrite.marker.Markers#SearchResult()` and `org.openrewrite.marker.Markers.searchResult(@Nullable String description` are deprecated and removed, use `SearchResult.found()` instead.
[...]

PR #3241, opened 2023-05-13
Kunli/rewrite8 migrate

kunli2:
Recipe to migrate rewrite 8.
ticket: https://github.com/openrewrite/rewrite/issues/3234

This PR covers below features
- [ ] 1. getVisitor() to public
- [ ] 2. single source applicability test to Preconditions
     * Includes potentially changing return type of `getVisitor()` to `TreeVisitor<?, ExecutionContext>`, because that is what Preconditions returns
- [ ] 3. Add a comment with a link to our migration documentation (to be written) to classes which use non-single source applicability tests
- [ ] 4. Add a comment with a link to our migration documentation to classes which override Recipe.visit(List<SourceFile>)
- [ ] 5. replace removed `JavaSourceFile visitJavaSourceFile(JavaSourceFile cu, ExecutionContext ctx)` to `J visit(@Nullable Tree tree, ExecutionContext ctx)`
- [ ] 6. `Applicability.of()` -> `Preconditions.of()`, and `Applicability.and()` -> `Preconditions.and()`
- [ ] 7. For YAML recipes using applicability tests I think we should comment the lines out and add a marker / comment which makes it clear that this is no longer supported and requires migrating the recipe to Java code
- [ ] 8. Migrate both imperative and declarative recipes referencing the recipes which were moved to `rewrite-static-analysis`
- [ ] 9. The `JavaTemplate.Builder#javaParser(Supplier<JavaParser>)` method no longer exists. Jonathan created a recipe to migrate it to `JavaTemplate.Builder#javaParser(JavaParser.Builder)`, but I think that recipe needs to be extended a bit, in case the `Supplier<JavaParser>` is represented by a variable or field
- [ ] 10. `TreeVisitor.doNext(..)` is removed, leave a comment to replace with `TreeVisitor.doAfterVisit(..)` or follow the instruction.
- [ ] 11.  For unit test, change `spec.recipe(X.doNext(Y)`) to `spec.recipes(X, Y)`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,6 @@
 plugins {
-    id("org.openrewrite.build.root") version("latest.release")
-    id("org.openrewrite.build.java-base") version("latest.release")
+    id("org.openrewrite.build.root") version("1.11.6")
+    id("org.openrewrite.build.java-base") version("1.11.6")
     id("org.openrewrite.rewrite") version("latest.release")
 }
 
--- a/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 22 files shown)


jsoref:
`org.openrewrite.java.spring.boot2.MigrateHibernateContraintsToJavax` to `org.openrewrite.java.spring.boot2.MigrateHibernateConstraintsToJavax`

Issue #3198, opened 2023-05-05
Javadoc link to array not preserved 


mduerig:
OpenRewrite does not properly preserve Javadoc links to arrays:

```
/**
 * Foo {@link javax.jcr.Value[] values} bar.
 */
```
is incorrectly converted to
```
/**
 * Foo {@link javax.jcr.Valuevaluesbar. bar.
 */
```
when running _mvn_ rewrite:run -Drewrite.activeRecipes=org.openrewrite.java.AddLicenseHeader_

timtebeek:
Sorry to hear! That's a bug indeed, and annoying that it won't let you run recipes without producing such unwanted changes. Feel free to open a pull request to replicate the issue in a test, such that we can more easily pick this up when we have time.

mduerig:
https://github.com/openrewrite/rewrite/pull/3213

timtebeek:
Thank you for adding a test! As also noted in https://github.com/openrewrite/rewrite/pull/3213#issuecomment-1538033667 I've merge the test already, and briefly explored what might cause this issue. It's either a problem printing the results back out, or a problem with how we parse the `int[]` in the code snippet (or both). Either way having the test helps to pick this up once we have time, or notify you here if it's resolved as part of other work.

PR #3247, opened 2023-05-16
Invalid Javadoc should remain unchanged

mduerig:
Related to https://github.com/openrewrite/rewrite/issues/3198

In this case invalid Javadoc gets changed in an unexpected way. I would expect OpenRewrite to either leave such JavaDoc unmodified or to bail out with a failure (like it would on regular Java code I suppose).


knutwannheden:
Thanks for the test case!

PR #3249, opened 2023-05-16
Support default package in `JavaParser.Builder.classpath(byte[][])`

### LINKED ISSUE #3222: When specifing a JavaParser's classpath from byte[][], usage of classes in the default package cannot be resolved
## The setup

There are 3 classes in the default package:

```java
public interface InterfaceA {
    void methodA();
}

public interface InterfaceB {
    void methodA();

    void methodB();
}

public class User implements InterfaceA, InterfaceB {
    @Override
    public void methodA() {}

    @Override
    public void methodB() {}
}
```

The `JavaParser` instance can be specified in two ways:
```java
// Classpath from compiled class files, fails to acknowledge classes in the default package
public static JavaParser prepareBinaryCP(byte[][] classpath) {
    JavaParser parser = JavaParser.fromJavaVersion()
            .classpath(classpath)
            .build();
    return wrap(parser);
}

// Classpath from directories (in our case, a single path pointing to the default package containing our sources)
public static JavaParser prepareSourceDirCP(Collection<Path> classpath) {
    JavaParser parser = JavaParser.fromJavaVersion()
            .classpath(classpath)
            .build();
    return wrap(parser);
}
````

Using `org.openrewrite:rewrite-java-17` attempting to resolve the `J.Identifier` type of the classes referened in `implements InterfaceA, InterfaceB` yields different results based on the type of classpath.

The parser created from `prepareBinaryCP` fails, and tells you the types are `Unknown`. The parser from `prepareSourceDirCP` succeeds and reports `InterfaceA` and `InterfaceB` respectively.

Attached you can find a gradle project with the full setup demonstrating this behavior.

[demo_project.zip](https://github.com/openrewrite/rewrite/files/11427326/demo_project.zip)

### LINKED ISSUE #3332: Datatable to collect recipe errors
terrible experience collecting/analyzing errors on recipe results. Before we undertake the next round of error fixing we should create/fix a data table that collects errors and allows us to group them similarly to what we did with parse failures. Once we get a data table it will already make it easier.

We have source files that errored on a recipe data table, but migrate to Java 20 recipe run on a Default group only contains 1 entry in that table.

Need to consider various error types like loading failures, parse failures that are on repositories, not on individual files, so these get collected as well.

[Slack Message](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1687022974321479)

timtebeek:
Fixes #3222

Sets package to "" for classes in default package, not identical to class name.

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17Parser.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17Parser.java
@@ -20,7 +20,6 @@ import com.sun.tools.javac.file.JavacFileManager;
 import com.sun.tools.javac.main.JavaCompiler;
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.util.Context;
-import com.sun.tools.javac.util.DiagnosticSource;
 import com.sun.tools.javac.util.Log;
 import com.sun.tools.javac.util.Options;
 import io.micrometer.core.instrument.Metrics;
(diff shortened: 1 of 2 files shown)


knutwannheden (approved):
LGTM.

PR #3251, opened 2023-05-16
Compare XML tag names for equality

### LINKED ISSUE #3250: SemanticallyEqualVisitor does not detect changed tags in child elements
Hi - working on a recipe to change contents of a pom, and I came across a potential issue in the `org.openrewrite.xml.SemanticallyEqual$SemanticallyEqualVisitor` class.

Specifically the `visitTag` method of this class, when considering child elements, doesn't appear to cater for the case where the existing and proposed lists of child elements are the same length, but the child element tag names are different.

e.g. it thinks the following

```&lt;configuration&gt;
  &lt;tasks&gt;
    &lt;delete file="erewhon" /&gt;
  &lt;/tasks&gt;
&lt;/configuration&gt;```
is the same as this:

```&lt;configuration&gt;
  &lt;target&gt;
    &lt;delete file="erewhon" /&gt;
  &lt;/target&gt;
&lt;/configuration&gt;```
because the `configuration` element in each case has just one child which is identical in all respects except that the child element name is different.

[Slack Message](https://rewriteoss.slack.com/archives/C01A843MWG5/p1684226699078279?thread_ts=1684226699.078279&cid=C01A843MWG5)

timtebeek:
Fixes #3250

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/SemanticallyEqual.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/SemanticallyEqual.java
@@ -70,6 +70,10 @@ public class SemanticallyEqual {
                 return null;
             }
             Xml.Tag otherTag = (Xml.Tag)other;
+            if (!tag.getName().equals(otherTag.getName())) {
+                areEqual = false;
+                return null;
+            }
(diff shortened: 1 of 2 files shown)


knutwannheden (approved):
LGTM.

PR #3263, opened 2023-05-22
Update Gradle model on removal of dependency

shanman190:


rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveDependency.java
```diff
@@ -78,6 +84,39 @@ protected TreeVisitor<?, ExecutionContext> getVisitor() {
             final MethodMatcher dependencyDsl = new MethodMatcher("DependencyHandlerSpec *(..)");
             final DependencyMatcher dependencyMatcher = requireNonNull(DependencyMatcher.build(groupId + ":" + artifactId).getValue());
 
+            @Override
+            public J visitJavaSourceFile(JavaSourceFile cu, ExecutionContext ctx) {
```
sambsnyd, line 88: We've gotten rid of `visitJavaSourceFile()` in rewrite 8.0.
This recipe will be easier to port if we go ahead and stop using it here, too. 

Can use `visitComilationUnit()`

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveDependency.java
```diff
@@ -78,6 +84,39 @@ protected TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                JavaSourceFile sourceFile = (JavaSourceFile) super.visitJavaSourceFile(cu, ctx);
+                if (sourceFile != cu) {
+                    Optional<GradleProject> maybeGp = sourceFile.getMarkers().findFirst(GradleProject.class);
+                    if (maybeGp.isPresent()) {
```
sambsnyd, line 92: Recipes are often written very defensively, with a long list of things to be checked, casted, null-safe'd before any change is made. To avoid ever-increasing indentation I find it more readable to return early.
```java
if(sourceFile != cu) {
     doStuff();
     if(otherCondition()) {
       doOtherStuff();
     }
}
return cu;
```
becomes the less-nested:
```java
if(sourceFile == cu) {
    return cu;
}
doStuff();
if(!otherCondition()) {
    return cu;
}
doOtherStuff();
return cu;
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveDependency.java
shanman190: Done via 807efaca908ce525a97466f52e91f7206a4d1039.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveDependency.java
shanman190: Done via 807efaca908ce525a97466f52e91f7206a4d1039.

PR #3265, opened 2023-05-22
Support using different configurations for different test source sets when adding gradle dependency

kzwang:
We are using the Nebula test facets plugin (https://github.com/nebula-plugins/nebula-project-plugin#test-facets) to create `smokeTest` source set, if a class is only used in `smokeTest`, we'd like to add that as `smokeTestImplementation` dependency instead of `implementation` or `testImplementation` dependency. 

Currently, the openrewrite logic only handles `implementation` and `testImplementation` dependencies. If `onlyIfUsing` matches code in `smokeTest` source set, it would add that as `implementation`. 

With this change, it adds an additional optional config `testSourceSet`. If that's not specified, it will keep the existing behavior (any source sets other than `test` will add as `implementation` dependency). If that's specified, `main` source set will be `implementation`, and other matching source sets will add `<source set name>Implementation` (e.g. `smokeTest` source set will add `smokeTestImplementation`)

Slack thread: https://rewriteoss.slack.com/archives/C01A843MWG5/p1684359544460089

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -20,19 +20,13 @@ import lombok.Value;
 import org.openrewrite.*;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
 import org.openrewrite.gradle.marker.GradleProject;
-import org.openrewrite.gradle.util.Dependency;
-import org.openrewrite.gradle.util.DependencyStringNotationConverter;
-import org.openrewrite.groovy.GroovyIsoVisitor;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.StringUtils;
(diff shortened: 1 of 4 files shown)


shanman190:
@kzwang, thanks for the PR! I've added a couple of test cases that weren't covered, but also uncover a bug in the current implementation.

kzwang:
@shanman190 thanks for adding that test! I pushed another commit that would check if the configuration is already `implementation` or not. If yes, it would skip to prefer `implementation`, if not, it will continue, to allow it to change to `implementation` if it's also used in the main source set.

shanman190:
So just to leave a quick status update here...

I've come up with another test, but for this one it pretty much broke the base implementation upon which this is built. I've taken it up on myself to perform that part of the refactor as it's quite a bit more intricate in nature, but I should have the updates out here soon (just have one more existing test case to get back to passing).

shanman190:
Latest updates depend upon https://github.com/openrewrite/rewrite-gradle-tooling-model/pull/5

sambsnyd:
Thanks @kzwang and @shanman190 !

PR #3272, opened 2023-05-27
Unsafe set requested and resolved dependencies to ensure a consistent instance identity

shanman190:


jkschneider:
Generally `unsafeSetXXX` methods are parse time only considerations, since modifications to an instance can't be observed as a change.

shanman190:
@jkschneider, right, I looked at other examples as well as having a short conversation with @sambsnyd. The issue that resulted in the need for this is specifically in later migration where we have to add, change, upgrade, or remove dependencies. Those each take actions against the `GradleDependencyConfiguration` model objects of the `GradleProject`. By default these are immutable objects, but are transitively referenced in the `extendsFrom` via what are supposed to be the same object instances (there's a bug with the tooling API model form, which I'm going to get a fix in for). As a result of these two things colliding, it means that either of the following two things have to happen:

1. Update the internal state using `unsafeSet*`
2. For _any_ dependency update, we would additionally have to traverse all configurations and update their `extendsFrom` to retain the instance identity and consistency of dependencies in the `requested` and `resolved` lists.

In my chat with Sam, we felt that it was best to go ahead and use the `unsafeSet*` mechanic here, so as to keep things minimal in terms of the above issues. 

If you've got another idea on how to deal with this, I'm all for that.

jkschneider:
I believe the intention was to fully resolve the model again after updates, similar to what we do with the maven plugin.

shanman190:
Alternate fix via https://github.com/openrewrite/rewrite-gradle-tooling-model/commit/c28cdb29c86e8444f879f61654a4470db0c33247.

PR #3276, opened 2023-05-29
Feature/maven wrapper 3.2.0+ add checksum verification

### LINKED ISSUE #2996: Maven wrapper 3.2.0+  : Add checksum verification
Starting with the Maven wrapper 3.2.0, they added checksum verification

This recipe should add this verification. 

See

https://issues.apache.org/jira/browse/MWRAPPER-75
https://github.com/apache/maven-wrapper/blob/master/src/site/markdown/index.md#checksum-verification-of-downloaded-binaries

coiouhkc:
## What's your motivation?
Closes #2996 

Please advice, whether chosen approach is OK.
In particularly
* initially started similar to GradleWrapper + AddGradleWrapper/UpdateGradleWrapper, then realized, that Maven doesn't offer a similar services endpoint (or just couldn't find it) - also since lacking `AddMavenWrapper` it would be more effort.
* fallback to `AddProperty`-like implementation - if this is the way to go, is there a possibility to chain the Recipies, e.g. replace custom implementation with AddProperty once appropriate (sha256 computed)?

### Checklist


joanvr:
Hi @coiouhkc, thank you very much for your contribution! 
Do you have plans to further work on it or Is it already ready for review?

rewrite-maven/src/test/java/org/openrewrite/maven/AddMavenWrapperChecksumValidationTest.java
```diff
@@ -0,0 +1,24 @@
+package org.openrewrite.maven;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.properties.Assertions.properties;
+
+public class AddMavenWrapperChecksumValidationTest implements RewriteTest {
```
joanvr, line 8: Could you add another test to verify that we do not make any changes if there is already a checksum?

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
```diff
@@ -0,0 +1,108 @@
...
+              FindProperties.find(mavenWrapperProperties, DISTRIBUTION_SHA_256_SUM, false);
+          if (distributionUrls.size() == 1 && distributionSha256Hexs.isEmpty()) {
+            Properties.Entry distributionUrl = distributionUrls.stream().findFirst().get();
+
+            try {
+              File distribution = File.createTempFile("rewrite", "download");
+              FileUtils.copyURLToFile(new URL(distributionUrl.getValue().getText()), distribution, 10_000, 10_000);
+              distribution.deleteOnExit();
+
+              String distributionSha256Hex = new DigestUtils("SHA-256").digestAsHex(distribution);
```
joanvr, line 84: Do you think it would be possible to do the checksum without external dependencies?

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
```diff
@@ -0,0 +1,108 @@
...
+          Set<Properties.Entry> distributionUrls =
+              FindProperties.find(mavenWrapperProperties, DISTRIBUTION_URL, false);
+          Set<Properties.Entry> distributionSha256Hexs =
+              FindProperties.find(mavenWrapperProperties, DISTRIBUTION_SHA_256_SUM, false);
+          if (distributionUrls.size() == 1 && distributionSha256Hexs.isEmpty()) {
+            Properties.Entry distributionUrl = distributionUrls.stream().findFirst().get();
+
+            try {
+              File distribution = File.createTempFile("rewrite", "download");
+              FileUtils.copyURLToFile(new URL(distributionUrl.getValue().getText()), distribution, 10_000, 10_000);
```
joanvr, line 81: Do you think it would be possible to download the file without external dependencies?

joanvr (commented):
Thank you very much for your contribution! Just a couple of questions regarding new external dependencies and requesting new unit tests.

rewrite-maven/src/test/java/org/openrewrite/maven/AddMavenWrapperChecksumValidationTest.java
coiouhkc: Sure, there is more to do, also a maven-wrapper.jar sha256 verification.
Just wanted to use the opportunity of EAP review to determine whether the rigth approach was chosen.

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
coiouhkc: Sure, that's possible, e.g. https://www.baeldung.com/sha-256-hashing-java though somewhat inconvenient/boilerplaty.

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
coiouhkc: That's a good one. Generally I was thinking about utilizing `MavenArtifactDownloader` o.s., but i wasn't quite sure, whether it's the best option.

coiouhkc:
> Hi @coiouhkc, thank you very much for your contribution! Do you have plans to further work on it or Is it already ready for review?

There is still some work to be done, I just wanted to get early feedback regarding chosen implementation (`AddProperty` vs. `AddGradleWrapper` style)

rewrite-maven/src/test/java/org/openrewrite/maven/AddMavenWrapperChecksumValidationTest.java
coiouhkc: Done

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
coiouhkc: Using `MavenArtifactDownloader`

rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
```diff
@@ -59,7 +59,7 @@ public String getDescription() {
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor() {
+    public PropertiesVisitor<ExecutionContext> getVisitor() {
```
timtebeek, line 62: This change is now also in #3384 such that it's not hidden away here.

rewrite-maven/src/main/java/org/openrewrite/maven/AddMavenWrapperChecksumValidation.java
```diff
@@ -0,0 +1,217 @@
...
+        this.distributionVersion = distributionVersion;
+        this.wrapperVersion = wrapperVersion;
+    }
+
+    @Override
+    public Properties visitFile(Properties.File file, ExecutionContext context) {
+        Properties.File mavenWrapperProperties = (Properties.File) super.visitFile(file, context);
+
+        if (equalIgnoringSeparators(file.getSourcePath(), Paths.get(MVN_WRAPPER_MAVEN_WRAPPER_PROPERTIES))) {
+            // TODO I feel like we should instead use the URL defined in the properties for download & sha calculation
```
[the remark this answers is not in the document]
coiouhkc: Exactly my thoughts as well, and also the reason for downloading the distributions from the official source (depending on the local network setup might be compromised as well, but slightly more difficult, then an arbitrary URL).
Setting the checksum to the checksum of the official distro will fail the build if the distro in the URL doesn't match to the provided/expected version, which is good and in the spirit of the checksum imo.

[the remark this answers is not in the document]
timtebeek: I agree that the official download URLs are probably safest, but my worry is that we fail in a number of common cases:

1. companies using custom internal URLs, or even distributions
2. recipe runs against projects with a mismatched wrapper or distribution version

Both would then end up with a project in a broken state, as the validation fails on the checksums (now) stored.

Perhaps it's then better to instead download from any URL provided (only), and calculate & store the hash for that. That at least guards against further tampering with the binaries, after an initial run of the recipe.

Would you agree that's perhaps better than the alternatives for the reasons outlined above?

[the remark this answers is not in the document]
coiouhkc: That depends on the scenario/ use-case/ environment, but I agree you have a strong point (btw, it was the initial implemenation as well until I started to second-guess/ doubt it).

Would you like me to revert implemenation to your suggestion?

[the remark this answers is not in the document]
timtebeek: If you'd be open to pick up where I left off, that'd be much appreciated! So downloading the URL that's there, possibly also with caching, and using that checksum

shanman190:
Just throwing a comment here real quick, I had started on an `UpdateMavenWrapper` locally on a branch for fun. I'll see about getting that aligned with current state and pushed out here shortly.

shanman190:
Here's the PR of mine: https://github.com/openrewrite/rewrite/pull/3392.

I think it supports all of this PRs original goals, so may be enough to fully supersede this one. I'll let the group decide that definitely though.

coiouhkc:
> Here's the PR of mine: #3392.
> 
> I think it supports all of this PRs original goals, so may be enough to fully supersede this one. I'll let the group decide that definitely though.

Looks great to me

shanman190:
So I personally think that `UpdateMavenWrapper` is able to completely satisfy the feature set here. Based on what I'm seeing, this recipe only sets the checksums when they're missing. With `UpdateMavenWrapper` it handles the same and more, including for cases where the version selected is the same as the current and the checksums aren't there presently. Do you see additional value coming from this one in some way that I may have missed?

shanman190:
@timtebeek / @joanvr, what's your opinions?

knutwannheden:
@coiouhkc 🙏  Thank you for all the effort you put into this first openrewrite PR of yours! I hope that you enjoyed working on it even if it in the end didn't get merged 😿

PR #3277, opened 2023-05-29
Fix unexpected varargs parsing

kunli2:
To fix unexpected varargs parsing in the Java17 parser (The parsing implementation of Java7 and Java11 parsers for varargs are different, so not included in this PR)

Found by @traceyyoshima in this Slack [thread](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1685155406306999)

To parse the code 
```java
class Test {
    void foo(String  ...  args) {
    }
}
```

Before this change, the `...` is stored in the variable's whitespace prefix.
Screen Shot 2023-05-29 at 11 43 54 AM

After this change, the `...` is stored in the `varargs` field of the `J.VariableDeclarations`.
Screen Shot 2023-05-29 at 11 42 33 AM

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -1472,10 +1472,14 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
 
         Space varargs = null;
         if (typeExpr != null && typeExpr.getMarkers().findFirst(JavaVarKeyword.class).isEmpty()) {
-            int varargStart = indexOfNextNonWhitespace(vartype.getStartPosition(), source);
+            int varargStart = indexOfNextNonWhitespace(vartype.pos, source);
             if (source.startsWith("...", varargStart)) {
-                varargs = format(source.substring(cursor, varargStart));
                 cursor = varargStart + 3;
+                if (vartype instanceof JCArrayTypeTree) {
+                    JCExpression elementType = ((JCArrayTypeTree) vartype).elemtype;
+                    varargStart = vartype.getStartPosition() + elementType.toString().length();
+                }
+                varargs = format(source.substring(varargStart, cursor));
             }
         }
 
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodDeclarationTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodDeclarationTest.java
@@ -64,6 +64,20 @@ class MethodDeclarationTest implements RewriteTest {
     }
 
     @Test
+    void typeVarargs() {
+        rewriteRun(
+          java(
+            """
+              class Test {
+                  void foo(String  ...  args) {
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
     void interfaceMethodDecl() {
         rewriteRun(
           java(
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
@@ -794,7 +794,6 @@ public class JavaPrinter<P> extends JavaVisitor<PrintOutputCapture<P>> {
         }
         if (multiVariable.getVarargs() != null) {
             visitSpace(multiVariable.getVarargs(), Space.Location.VARARGS, p);
-            p.append("...");
         }
         visitRightPadded(multiVariable.getPadding().getVariables(), JRightPadded.Location.NAMED_VARIABLE, ",", p);
         afterSyntax(multiVariable, p);


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1472,10 +1472,14 @@ private J.VariableDeclarations visitVariables(List<VariableTree> nodes, Space fm
 
         Space varargs = null;
         if (typeExpr != null && typeExpr.getMarkers().findFirst(JavaVarKeyword.class).isEmpty()) {
-            int varargStart = indexOfNextNonWhitespace(vartype.getStartPosition(), source);
+            int varargStart = indexOfNextNonWhitespace(vartype.pos, source);
```
sambsnyd, line 1475: `DiagnosticPosition.getPreferredPosition()` returns the `pos` field. Use that since the interface is less likely to change in future java versions than the field name.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1472,10 +1472,14 @@ private J.VariableDeclarations visitVariables(List<VariableTree> nodes, Space fm
 
         Space varargs = null;
         if (typeExpr != null && typeExpr.getMarkers().findFirst(JavaVarKeyword.class).isEmpty()) {
```
sambsnyd, line 1474: We probably need a corresponding change in java 8 and 11 parsers

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodDeclarationTest.java
```diff
@@ -63,6 +63,20 @@ public <P, R> R foo(P p, String s, String... args) {
...
+          java(
+            """
+              class Test {
+                  void foo(String  ...  args) {
+                  }
+              }
+              """
+          )
+        );
+    }
```
sambsnyd, line 78: `MethodDeclarationTest.typeArguments()` already has a method with varargs signifier. Of course neither that test or this one failed before this fix because the correct result is printed out despite the LST being wrong. So neither test case catches this bug.

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
```diff
@@ -794,7 +794,6 @@ public J visitVariableDeclarations(VariableDeclarations multiVariable, PrintOutp
         }
         if (multiVariable.getVarargs() != null) {
             visitSpace(multiVariable.getVarargs(), Space.Location.VARARGS, p);
-            p.append("...");
```
sambsnyd, line 797: This doesn't look right. It seems like without this the test would ONLY pass if the "..." were inside the whitespace of `multiVariable.getVarargs()`, which isn't correct either. No "..." should exist within the whitespace of _any_ element, except within a comment.

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodDeclarationTest.java
kunli2: Right, neither test case catches this bug because the printed results are all correct, so seems the only way to capture it is add some verification in `spec.afterRecipe()`

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
kunli2: making the change here because the  "..." has been included in `multiVariable.getVarargs()` and don't want to print it twice.
I thought `Space VariableDeclarations.varargs`  ([code](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L5464)) is intentionally to store "...", it's a Space type, so maybe we should put "..." into `Space.comments`? what do you think?

kunli2: I thought about this again, and it seems like putting "..." to the `Space.comments ` does not look correct because it's not a comment.

kunli2: Changed this line back, and instead, remove "..." from `multiVariable.getVarargs()`

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1472,10 +1472,14 @@ private J.VariableDeclarations visitVariables(List<VariableTree> nodes, Space fm
...
-            int varargStart = indexOfNextNonWhitespace(vartype.getStartPosition(), source);
+            int varargStart = indexOfNextNonWhitespace(vartype.getPreferredPosition(), source);
             if (source.startsWith("...", varargStart)) {
-                varargs = format(source.substring(cursor, varargStart));
                 cursor = varargStart + 3;
+                if (vartype instanceof JCArrayTypeTree) {
+                    JCExpression elementType = ((JCArrayTypeTree) vartype).elemtype;
+                    varargStart = vartype.getStartPosition() + elementType.toString().length();
```
knutwannheden, line 1480:
```suggestion
                    varargStart = elementType.getEndPosition(endPosTable);
```

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodDeclarationTest.java
knutwannheden: Yes, that should do it.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
kunli2: This looks good, thanks

PR #3281, opened 2023-05-30
Fix for type cast error in Gradle AddDependency recipe.

jeremyg484:
## What's changed?

An instanceof check for JavaSourceFile is added to AddDependency.java.

## What's your motivation?

Without this type check, the invocation of `UsesType` in `AddDependency`'s scanning visitor fails when encountering non-Java files such as a properties file.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -160,14 +160,16 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
             @Override
             public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                 SourceFile sourceFile = (SourceFile) requireNonNull(tree);
-                sourceFile.getMarkers().findFirst(JavaProject.class).ifPresent(javaProject ->
-                        sourceFile.getMarkers().findFirst(JavaSourceSet.class).ifPresent(sourceSet -> {
-                            if (sourceFile != new UsesType<>(onlyIfUsing, true).visit(sourceFile, ctx)) {
-                                acc.usingType = true;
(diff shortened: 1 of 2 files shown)


shanman190:
Hi @jeremyg484, I'm not sure if I missed something here, but it appears that `UsesType`'s first check is an instanceof `JavaSourceFile` in it's visit method. So unless I've totally missed something, then removing the if check in `AddDependency` should still result in a passing test case.

jeremyg484:
> Hi @jeremyg484, I'm not sure if I missed something here, but it appears that `UsesType`'s first check is an instanceof `JavaSourceFile` in it's visit method. So unless I've totally missed something, then removing the if check in `AddDependency` should still result in a passing test case.

@shanman190 The problem comes here: https://github.com/openrewrite/rewrite/blob/011685861c1cc49a3b8e8ad77c5d45314ec60f5a/rewrite-java/src/main/java/org/openrewrite/java/search/UsesType.java#L82

The `tree` is cast to `org.openrewrite.java.tree.J`. Running the test case I included without the change to `AddDependency` will result in a `ClassCastException` at that line.

Note that the Maven version of `AddDependency` already has the type check:
https://github.com/openrewrite/rewrite/blob/011685861c1cc49a3b8e8ad77c5d45314ec60f5a/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L173

jeremyg484:
@shanman190 It looks like https://github.com/jeremyg484/rewrite/commit/011685861c1cc49a3b8e8ad77c5d45314ec60f5a has taken an alternate (better) approach to fixing this same problem. My test case now passes after pulling in this change.

That said, the above commit does not include a test case, so it might still be worth merging the added test case from this PR. I've updated the PR branch accordingly.

shanman190:
I agree. This test case can help to ensure that specific check is made which would be good going forward.

PR #3282, opened 2023-05-30
Support `Recipe.getRecipeList()` for `doAfterVisit` method

kunli2:
just found a failing test that `TreeVisitor#doAfterVisit()` doesn’t work if the parameter is a recipe which overridden  `Recipe#getRecipeList()`, [here](https://github.com/openrewrite/rewrite-spring/blob/main/src/testWithSpringBoot_2_4/java/org/openrewrite/maven/spring/UpgradeExplicitSpringBootDependenciesTest.java#L812) is the failing test.

Root cause: Method  `doAfterVisit` ([Code](https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java#L158)) takes recipe.getVisitor() but ignored `getRecipeList`.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
@@ -158,6 +158,7 @@ public abstract class TreeVisitor<T extends Tree, P> {
         List<Recipe> subRecipes = recipe.getRecipeList();
         if (!subRecipes.isEmpty()) {
             for (Recipe r : subRecipes) {
+                //noinspection unchecked
                 afterVisit.add((TreeVisitor<T, P>) r.getVisitor());
             }
         }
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
```diff
@@ -154,6 +154,14 @@ protected void doAfterVisit(Recipe recipe) {
         if (afterVisit == null) {
             afterVisit = new ArrayList<>(2);
         }
+
+        List<Recipe> subRecipes = recipe.getRecipeList();
+        if (!subRecipes.isEmpty()) {
+            for (Recipe r : subRecipes) {
+                //noinspection unchecked
+                afterVisit.add((TreeVisitor<T, P>) r.getVisitor());
```
knutwannheden, line 162: Should this algorithm be recursive? And what if we see a ScanningRecipe? Throw an IllegalArgumentException?

knutwannheden:
doAfterVisit(Recipe) was now removed.

kunli2:
what do you do instead of `doAfterVisit(Recipe) `?

PR #3284, opened 2023-05-31
Recursively init Declarative Recipes

BoykoAlex:
`Enviroment#listRecipes()` has empty `recipesList` fields starting from the second level. Recipes are there but not initialized.


joanvr (commented):
Could you provide some test that was previously failing and after the fix is passing?

BoykoAlex:
@joanvr sure. Please see this PR: https://github.com/openrewrite/rewrite-spring/pull/322. This test went in but then i guess it was removed from **rewrite-spring**. Anyway, paste the test in the **rewrite-spring** code and run it. It'd fail.

kunli2:
LGTM, not sure why the test `RecipeDescriptorTest` introduced in this PR https://github.com/openrewrite/rewrite-spring/pull/322 was removed, but seems we should have this test.

kunli2 (approved):

BoykoAlex:
Could someone merge this one? Don't think i have the push rights...

PR #3285, opened 2023-05-31
JavaTemplate#apply

jkschneider:
@sambsnyd FYI

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Cursor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Cursor.java
@@ -51,6 +51,15 @@ public class Cursor {
         return c;
     }
 
+    /**
+     * @param value The value to attach to this cursor.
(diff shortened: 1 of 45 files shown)


rewrite-core/src/main/java/org/openrewrite/Cursor.java
```diff
@@ -51,6 +51,15 @@ public Cursor getRoot() {
         return c;
     }
 
+    /**
+     * @param value The value to attach to this cursor.
+     * @return A new cursor whose {@link #getValue()} returns the supplied value and whose
+     * ancestry is the current cursor.
+     */
+    public Cursor attach(Object value) {
+        return new Cursor(this, value);
```
knutwannheden, line 60: Would it make sense to add something like an `assert` here to check that the given value doesn't match this cursor's value?

PR #3288, opened 2023-06-02
Skip scanning phase when no recipes in the recursive recipe list are …

jkschneider:
@knutwannheden FYI

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
+++ b/rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
@@ -68,7 +68,9 @@ public class RecipeScheduler {
 
             // pre-transformation scanning phase where there can only be modifications to capture exceptions
             // occurring during the scanning phase
-            after = cycle.scanSources(after, i);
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -92,6 +94,18 @@ public RecipeRun scheduleRun(Recipe recipe,
         );
     }
 
+    private boolean hasScanningRecipe(Recipe recipe) {
+        if (recipe instanceof ScanningRecipe) {
+            return true;
+        }
+        for (Recipe r : recipe.getRecipeList()) {
```
knutwannheden, line 101: The `getRecipeList()` method can now get called multiple times, but I guess that is unproblematic for this use case. Or can we iterate over a "resolved" recipe tree?

jkschneider: That's just a listing of recipes, not an execution of them.

Issue #3200, opened 2023-05-05
Rewrite fully qualified package name used as argument parameter inside annotation when using Change Package Name rewrite rule


apupier:
I think that Change package name rule https://docs.openrewrite.org/recipes/java/changepackage#source is ignoring the packages which are inside annotations.

I see no reference of annotation in the implementation class: https://github.com/openrewrite/rewrite/blob/v7.40.5/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java

I think it is what it is used by the Quarkus OpenRewrite recipe. I reported intially in Quarkus repository: https://github.com/quarkusio/quarkus/issues/33157

timtebeek:
Could you provide a recipe configuration, along with a code sample that illustrates the issue? I'm having a bit of difficulty visualizing the issue, and a clear example could help reproduce, and rule out configuration issues.

We have [unit tests that look to cover what you're reporting here](https://github.com/openrewrite/rewrite/blob/b9a0475eb72a5cdde5003a64f57bae0a9bf08cb4/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java#L514-L553), so a similar test that fails would be a ideal.

apupier:
I'm not familiar with OpenRewrite neither Gradle and it was only as a side tasks that I noticed that so it will take some time to provide a simpler reproducible.

In the pointed test, it seems that there is no package name migration for one of the value inside the annotation.

The use case, i spotted is inside a package-info.java file with this content:
```
@XmlSchema(namespace = "https://www.lfenergy.org/compas/example", elementFormDefault = javax.xml.bind.annotation.XmlNsForm.QUALIFIED)
package org.lfenergy.compas.core.commons.model;

import javax.xml.bind.annotation.XmlSchema;
```

that the migration resulted in:
```
@XmlSchema(namespace = "https://www.lfenergy.org/compas/example", elementFormDefault = javax.xml.bind.annotation.XmlNsForm.QUALIFIED)
package org.lfenergy.compas.core.commons.model;

import jakarta.xml.bind.annotation.XmlSchema;
```

although I think it will be better if it is rewriting like:
```
@XmlSchema(namespace = "https://www.lfenergy.org/compas/example", elementFormDefault = jakarta.xml.bind.annotation.XmlNsForm.QUALIFIED)
package org.lfenergy.compas.core.commons.model;

import jakarta.xml.bind.annotation.XmlSchema;
```

[...]

PR #3294, opened 2023-06-05
ChangePackage on annotation with FQCN argument

timtebeek:
## What's changed?
Fixed issues with tests that attempted to replicate 3200, but in reality failed because of incorrect visibility on `Argument.class`.

## What's your motivation?
https://github.com/openrewrite/rewrite/issues/3200

### Checklist


rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
```diff
@@ -600,14 +597,13 @@ void method() {}
               """,
             spec -> spec.afterRecipe(cu -> {
                 assertThat(cu.findType("org.openrewrite.Argument")).isEmpty();
-                assertThat(cu.findType("org.openrewrite.test.Argument")).isNotEmpty();
+                // assertThat(cu.findType("org.openrewrite.test.Argument")).isNotEmpty();
```
timtebeek, line 600: Weirdly this test has no such issues, so we might have a problem with the type attribution here:
https://github.com/openrewrite/rewrite/blob/b21c0a46588cffb894fe938d4b22330611c82446/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java#L1259-L1283

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -88,6 +88,15 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration c, Integer p)
             return c;
         }
 
+        @Override
+        public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, Integer integer) {
+            if ("class".equals(fieldAccess.getSimpleName()) && fieldAccess.getTarget() instanceof J.FieldAccess) {
+                J.FieldAccess target = (J.FieldAccess) fieldAccess.getTarget();
+                types.add(JavaType.buildType(target.toString()));
+            }
```
timtebeek, line 96: This might not be necessary anyone if we manage to resolve the underlying missing type issue. Best to check that before a merge if we make that part of this PR.

PR #3295, opened 2023-06-05
Improve the context-free template caching

knutwannheden:
`JavaTemplateParser` now caches based on the input template code rather than on the stub produced by `BlockStatementTemplateGenerator`. This requires including the imports in the cache key.


knutwannheden:
@jkschneider We also need to include the parser's classpath as a component of the cache key. My proposal would be to add a package protected accessor to `JavaParser.Builder` and then some kind of public `Internals` class to the `org.openrewrite.java` package with a static method to call this accessor. What do you think?

knutwannheden:
In a further step I would for the `JavaTemplate#matches()` use case to add an instance field at the `JavaTemplate` level, so we can short-circuit even more parts of the code and as a result get a considerable performance boost for Refaster-style usages.

PR #3300, opened 2023-06-07
Fix NPE caused by org.openrewrite.staticanalysis.CombineSemanticallyEqualCatchBlocks

Bananeweizen:
## What's changed?
getMethodType() is nullable and must be checked at every invocation.

## What's your motivation?
```
java.lang.NullPointerException: Cannot invoke
"org.openrewrite.java.tree.JavaType$Method.hasFlags(org.openrewrite.java.tree.Flag[])" because the return value of
"org.openrewrite.java.tree.J$MethodInvocation.getMethodType()" is null at org.openrewrite.java.search.SemanticallyEqual$SemanticallyEqualVisitor.visitMethodInvocation (SemanticallyEqual.java:874)
```


timtebeek:
Thanks for the quick suggested fix @Bananeweizen ! Would you mind sharing how you first came across this issue?
Just so I know why we might not have encountered this before. Are you working on your own recipe development perhaps?

rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java
```diff
@@ -871,7 +871,7 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, J j)
                     return method;
                 }
 
-                boolean static_ = method.getMethodType().hasFlags(Flag.Static);
+                boolean static_ = method.getMethodType() != null && method.getMethodType().hasFlags(Flag.Static);
                 J.MethodInvocation compareTo = (J.MethodInvocation) j;
                 if (!method.getSimpleName().equals(compareTo.getSimpleName()) ||
                     !TypeUtils.isOfType(method.getMethodType(), compareTo.getMethodType()) ||
```
timtebeek, line 877: There's another instance of `compareTo.getMethodType().hasFlags(Flag.Static)` just below this line; would that not be similarly affected?

Bananeweizen: I have been wondering about that one myself too. I'm still fighting with the gradle build (company machine behind proxy), so I cannot easily test it. I can use the patched version at home to see if there is more to fix, but that might take a day or two.

Bananeweizen:
> Are you working on your own recipe development perhaps?

As mentioned in the title, that's one of the default recipes. I would surely like to provide a reproducer, but I haven't yet found a way to easily identify the problematic part of the source file. Are there some switches to be enabled or good objects to look at in a debugger to more easily find the line and position of the currently investigated part of the abstract syntax tree?

PR #3303, opened 2023-06-08
added QualifyThisVisitor

nmck257:
## What's changed?
Added a new visitor (not a full recipe)

## What's your motivation?
Wrote this for a private project, but it has generic usefulness (eg when wrapping existing code into a new inner/anonymous class)

Is there a better alternative to instantiating these Tree objects directly?

### Checklist
- [X] I've used the IntelliJ auto-formatter on affected files
- ~~[ ] I've updated the documentation (if applicable)~~


timtebeek:
Already looks good to me; perhaps @knutwannheden can answer your question, given that he's more up to date on any potential template support.
> Is there a better alternative to instantiating these Tree objects directly?

knutwannheden:
Typically we use `JavaTemplate` rather than manually instantiating the LST elements, in order to get the type attribution right. However, sometimes this is either not possible or the edits are trivial and a `JavaTemplate` does carry quite a bit of overhead with it.

knutwannheden (approved):

PR #3304, opened 2023-06-08
Rename `InvocationMatcher#fromInvocationMatchers()` to `or()`

knutwannheden:
This is consistent with `Preconditions` and will in the future allow us to add `and()` and `not()` methods, so that they can be logically combined, if ever.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/InvocationMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/InvocationMatcher.java
@@ -26,8 +26,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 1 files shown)


jkschneider:
I'm highly suspicious of adding ever more API surface area for things that the language has binary operators for...

`MM1 || MM2` vs. `InnovcationMatchers.or(mm1, mm2)` etc.

JLLeitschuh:
> and will in the future allow us to add and() and not() methods, so that they can be logically combined

Erm..... What would `and` even mean for `MethodMatcher`? And when would `not` be relevant?

Fundamentally, these methods are creating a composite `InvocationMatcher` from the passed arguments.

Also, perhaps `or` makes more sense as a non-static method.

`methodMatcherA.or(methodMatcherB).matches(method)`

Thoughts?

knutwannheden:
With the wildcards one could use `and` and `not` to match against all methods of a class except one or similar. But I am also OK with dropping this PR. To me the InvocationMatcher.fromInvocationMatchers() just looked a bit weird.

Issue #3156, opened 2023-04-24
Auto Lombok @UtilityClass


blipper:
For a class that is a utility class https://projectlombok.org/features/experimental/UtilityClass automatically transform the class into a @UtilityClass

coiouhkc:
Could you please assign it to me? I would attempt it in coming weeks...

PR #3306, opened 2023-06-11
Feature/3156 lombok utility class

coiouhkc:
## What's your motivation?
Closes https://github.com/openrewrite/rewrite/issues/3156

### Checklist


coiouhkc:
WIP (sommer camp) together with @lukas-poos-openvalue

rewrite-java/src/test/java/org/openrewrite/java/LombokUtilityClassTest.java
```diff
@@ -0,0 +1,314 @@
...
+ */
+class LombokUtilityClassTest implements RewriteTest {
+
+    @Nested
+    class ShouldApplyLombokUtility implements RewriteTest {
+
+        @Test
+        void givenOneStaticMethod() {
+            rewriteRun(
+                    recipeSpec -> recipeSpec.recipe(new LombokUtilityClass()),
```
timtebeek, line 27: You might want to override the `defaults` method so define the recipe to run once, instead of in each individual test.

rewrite-java/src/main/java/org/openrewrite/java/LombokUtilityClass.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public String getDescription() {
+        return "This recipe will check if any class is transformable (only static methods in class)" +
+                " into the Lombok UtilityClass and will perform the change if applicable.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new ChangeVisitor();
```
timtebeek, line 46: You could look at using `Precondition.check` here to first verify that a class is a utility class candidate, before applying the visitor that adds the annotation.

timtebeek:
Looking good already; thanks for getting started on this! Since you still have TODO items in the code I only posted some quick suggestions. Let me know when you'd like a more thorough review.

kunli2:
nice work! looks good to me besides a few comments from Tim and todo items.

rewrite-java/src/test/java/org/openrewrite/java/LombokUtilityClassTest.java
```diff
@@ -0,0 +1,385 @@
...
+                    java(
```
timtebeek, line 39: Suggest to 
1. override defaults to define the recipe only once; `spec.recipe(Recipe)` calls can be removed from all other tests too
2. nested classes don't have a to also implement `RewriteTest`

```suggestion
    @Override
    public void defaults(RecipeSpec spec) {
        spec.recipe(new LombokUtilityClass());
    }

    @Nested
    class ShouldApplyLombokUtility {

        @Test
        void givenOneStaticMethod() {
            rewriteRun(
                    java(
```

rewrite-java/src/main/java/org/openrewrite/java/LombokUtilityClass.java
```diff
@@ -0,0 +1,185 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
```
JLLeitschuh, line 16:
```suggestion
package org.openrewrite.java.lombok;
```

Put this in a lombok sub package?

timtebeek: It might fit in best with https://github.com/openrewrite/rewrite-migrate-java/tree/main/src/main/java/org/openrewrite/java/migrate/lombok , since we've rearchitectured openrewrite/rewrite a bit to have less recipes in there, and more in specific modules

timtebeek:
As indicated I suggest we move this over to [the lombok package on rewrite-migrate-java](https://github.com/openrewrite/rewrite-migrate-java/tree/main/src/main/java/org/openrewrite/java/migrate/lombok). I didn't want to just copy over the content here, as we use Git history to give credit to our contributors. Would you mind opening a PR there @coiouhkc ?

blipper:
I am a big confused on the state here?

timtebeek:
> I am a big confused on the state here?

This PR was opened against openrewrite/rewrite, but is a better fit for openrewrite/rewrite-migrate-java, given that we already have a few other Lombok recipes there.

PR #3307, opened 2023-06-11
Replace `String.replaceAll(..)` with `Pattern.compile(..)` use

### LINKED ISSUE #1659: Replace `String#replaceAll` with a compiled Regular expression
String `replaceAll` uses a compiled regular expression. In "hot" methods this often leads to performance problems

```java

public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }

```

```java

class A {
   public void replace(){
           String init = "Bob is a Bird... Bob is a Plane... Bob is Superman!";
            String changed = init.replaceAll("/[@]/g,", "_");
}
}
``` 

```java

class A {

    private static final java.util.regex.Pattern p = Pattern.compile("/[@]/g,");
 
   public void replace(){
           String init = "Bob is a Bird... Bob is a Plane... Bob is Superman!";
           String changed = p.matcher(init).replaceAll("_");

}
}
```

sebastiankonieczek:
## What's your motivation?

- Closes #1659

### Checklist


sebastiankonieczek:
WIP (sommer camp) together with @dietmarweb

timtebeek (commented):
Great start to this recipe, thanks! I've posted some comments, but also have a concern about the proposed recipe itself, where that relates to:

> This recipe could [do "harm"](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) if the method is not "hot" as the compiled expression will take unnecessary memory space. We should warn users about this in the recipe description and let them review the diff manually

Not sure if or how we could detect hot code paths, but this makes it a challenge if the recipe can only be run with manual inspection of the results.

knutwannheden:
> We also want to at least add a warning for the user about the possible affection on memory footprint. Is there a good way to add a prominent warning with the rewrite tool chain?
> 

You could look into using `Markup` as you can see in this example: https://github.com/openrewrite/rewrite/blob/03c458404073082275a752063887c188d213afbe/rewrite-java/src/main/java/org/openrewrite/java/RecipeMarkupDemonstration.java#L53. Here is the corresponding test case: https://github.com/openrewrite/rewrite/blob/ca982c70e4a7c5113c80e466b8ebc49e47d132d4/rewrite-java-test/src/test/java/org/openrewrite/java/RecipeMarkupDemonstrationTest.java#L31-L40

> We already thought about how to give the user more control over the recipe through configuration but did not yet find a feasible solution.

Can you tell us a bit more about what you are referring to here? Would you for example want to include something like a class name or method pattern, so the user can specify to which classes / methods this recipe should be applied?

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,117 @@
...
+                Object regExp = parentOrThrow.pollNearestMessage(REG_EXP_KEY);
+                if (regExp == null) {
+                    return processedBlock;
+                }
+                patternProcessingInProgress.set(false);
+
+                final String regExpName = parentOrThrow.pollNearestMessage(REG_EXP_KEY + "_NAME");
+                return super.visitBlock(
+                        JavaTemplate
+                                .builder("private static final java.util.regex.Pattern " + regExpName + " = Pattern.compile(#{});")
```
sebastiankonieczek, line 76: Can you give me a hint on how to use substitution here for the regExpName? I was not able to get it working. It ended up with a weird content containing "private static final java.util.regex.Pattern __P__;/*__TEMPLATE_STOP__*/" and I couldn't get my head around it.

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,107 @@
...
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.Markers;
+
+import java.util.regex.Pattern;
+
+import static org.openrewrite.Tree.randomId;
+
+public class UsePrecompiledRegExpForReplaceAll extends Recipe {
+
+    public static final String REG_EXP_KEY = "regExp";
```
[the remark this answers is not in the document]
sebastiankonieczek: How can VariableNameUtils help me with the key for the message I use to transport the actual regular expression up to the class level?

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,119 @@
...
+            final String regexPatternVariableName = VariableNameUtils.generateVariableName(PATTERN_VAR, getCursor(), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+            final Object regexPatternIdentifier = new J.Identifier(
+                    randomId(),
+                    Space.SINGLE_SPACE,
+                    Markers.EMPTY,
+                    regexPatternVariableName,
+                    JavaType.buildType(Pattern.class.getName()),
+                    null);
+
+            final Cursor cursor = getCursor();
```
sebastiankonieczek, line 108: Is it safe to declare a variable for the cursor and reuse it or should I always call the getCursor() function?

sebastiankonieczek:
> Can you tell us a bit more about what you are referring to here? Would you for example want to include something like a class name or method pattern, so the user can specify to which classes / methods this recipe should be applied?

I had the idea that one must define the regular expressions to pre compile in the configuration. But that seems to me like almost doing the refactoring manually. So I would refrain from doing so.

It is hard to realize which parts of the code are hot. But one could argue, that a pre compiled regular expression does not take up that much space and furthermore typically there are not a lot of these expressions present. Like in most cases there are maybe up to 10 or 20 of these calls and I already consider this a lot. I know, one cannot foresee what people are doing. But it seems atypical to me.I think, this will barely do real harm to an application. So a hint or a warning should suffice for this case.

Any thoughts about this?

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
knutwannheden: Declaring a variable is safe. But in case you were to modify the input object (in this case `method`) then you have to create a new `Cursor` to point to that updated instance, if you later want to modify the instance again using a `JavaTemplate`. To this end you can use `updateCursor(Object)`.

Here you do in fact do this on line 98 with the `super.visitMethodInvocation()` call. So I suggest the following:
```suggestion
            final Cursor cursor = updateCursor(invocation);
```

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,119 @@
...
+            cursor.putMessageOnFirstEnclosing(J.ClassDeclaration.class, REG_EXP_KEY, invocation.getArguments().get(0));
+            cursor.putMessageOnFirstEnclosing(J.ClassDeclaration.class, REG_EXP_KEY + "_NAME", regexPatternVariableName);
+
+            return super.visitMethodInvocation(MATCHER_TEMPLATE.apply(
```
knutwannheden, line 112: When you make the change above, you may no longer need the `super.visitMethodInvocation()` here anymore. Or possibly you don't need the call on 98?

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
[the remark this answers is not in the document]
knutwannheden: I see you are already using `VariableNameUtils` on line 99, which is good. It would be nice if we at some point can figure out a strategy to get better variable names, as for example IDEA does it. But that is out of scope for this PR.

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static org.openrewrite.Tree.randomId;
+
+public class UsePrecompiledRegExpForReplaceAll extends Recipe {
+
+    public static final String REG_EXP_KEY = "regExp";
+
+    @Override
+    public String getDisplayName() {
+        return "Replace `String.replaceAll` with a compiled regular expression";
```
knutwannheden, line 38: Should we expand this recipe to also include `String#replaceFirst()`?

rewrite-java-test/src/test/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAllTest.java
```diff
@@ -0,0 +1,152 @@
...
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class UsePrecompiledRegExpForReplaceAllTest implements RewriteTest {
```
knutwannheden, line 23: I think it would be great to also have a test where a replacement occurs in a nested class as well as one for an anonymous subclass. In both cases we should then have a second `replaceAll()` call in the scope of the outer class. I just think it would be good to make sure we don't get any conflicts on the introduced constants in these cases.

knutwannheden (commented):
@sebastiankonieczek I am very sorry about the long delay here! I've provided a few review comments.

rewrite-java/src/main/java/org/openrewrite/java/UsePrecompiledRegExpForReplaceAll.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new ReplaceVisitor();
+    }
+
+    public static class ReplaceVisitor extends JavaIsoVisitor<ExecutionContext> {
+        // use boolean flag to only process one pattern at a time
+        private final AtomicBoolean patternProcessingInProgress = new AtomicBoolean(false);
+        private static final String PATTERN_VAR = "openRewriteReplaceAllPattern";
```
JLLeitschuh, line 54: To get a guaranteed unique name for a given scope, you can use this:
```java
VariableNameUtils.generateVariableName(
                                newVariableBaseName,
                                getCursor(),
                                VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER
                        )
```

knutwannheden:
@sebastiankonieczek I just wanted to check with you if this PR is still something you would like to continue working on. It is not quite a while ago you created it 😄

sebastiankonieczek:
@knutwannheden sorry for the long break, I will continue on this PR right now :)

knutwannheden:
Note: Merging the main branch into your PR should resolve the CI build failures.

timtebeek:
Closing this PR as it's likely a better fit for rewrite-static-analysis, and has some comments left unaddressed above. We welcome the addition of the recipe, but probably best picked up anew taking in what was learned above and applying that to rewrite-static-analysis.

Issue #2413, opened 2022-11-11
Change mutable public collections to immutable versions


yeikel:
```java

public static final List<String> entries = Arrays.asList("A","B");
```

```java

public static final List<String> entries = Collections.unmodifiableList(Arrays.asList("A","B"));
```

While List.of is also an option, it is important to note that List.of throws NPE for null access patterns (like contains) and that can be problematic. 

```java

var a = Collections.unmodifiableList(Arrays.asList("A","B"));
var b =Arrays.asList("A","B");
var c = List.of("A","B");

a.contains(null) // false
b.contains(null) // false
c.contains(null) // NPE

````

**Comment by coiouhkc** [comment]
Could you please assign it to me? I would attempt it in coming weeks...

**Comment by timtebeek** [comment]
I have some doubts if this recipe could [do harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) in certain situations. Consider something like:
```java
@Test
void doNotWrapIfSet() {
    rewriteRun(
      java("""
        import java.util.Arrays;
        import java.util.List;
        
        class A {
            public static void main(String[] args) {
                List<String> entries = Arrays.asList("A", "B");
                entries.set(0, "C");
                System.out.println(entries);
            }
        }
        """
      )
    );
}
```
This is allowed currently, as changing an element of the backing array does not change the size of the list.

But this recipe could convert that into `Collections.unmodifiableList(Arrays.asList("A", "B"))`, which no longer allows `entries.set(0, "C")`.

JLLeitschuh:
You could use data flow to determine if `entries` is ever used with `set`

PR #3308, opened 2023-06-11
Change mutable public collections to immutable versions: ref issue #2413

MoritzCSchmidt:
## What's changed?

Added new recipe to wrap modifiable lists to be unmodifiable.

## What's your motivation?

Closes openrewrite/rewrite-static-analysis#125 

## Any additional context

Thanks for your work! I hope we can support with our small improvements.

### Checklist


timtebeek:
Thank you for starting this implementation! It's looking quite good already.
I've added two small polishing commits, and will add some review comments.

rewrite-java/src/main/java/org/openrewrite/java/recipes/ArraysAsListToImmutableRecipe.java
```diff
@@ -0,0 +1,86 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+
+/**
+ * This recipe modifies the invocation of a method so that it returns an unmodifiable list, if the
+ * method returns a modifiable one.
```
timtebeek, line 31: Perhaps this is more of an issue with the original feature request rather than the recipe implementation, but I have some doubts if this recipe could [do harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) in certain situations. More details here: https://github.com/openrewrite/rewrite/issues/2413#issuecomment-1586767731

rewrite-java/src/main/java/org/openrewrite/java/recipes/ArraysAsListToImmutableRecipe.java
```diff
@@ -0,0 +1,86 @@
...
+                    return super.visitMethodInvocation(method, executionContext);
+                }
+
+                J.MethodInvocation result =
+                        JavaTemplate.builder("Collections.unmodifiableList(#{any()})")
+                                .imports("java.util.Collections", "java.util.Arrays")
+                                .contextSensitive()
+                                .build()
+                                .apply(getCursor(), method.getCoordinates().replace(), method);
+                maybeAddImport("java.util.Arrays");
```
timtebeek, line 78: I don't think this import is needed, since we already have a method matcher on `Arrays.asList`.
```suggestion
```
I could be wrong though; did you add this in response to seeing issues that I might have missed?

rewrite-java/src/main/java/org/openrewrite/java/recipes/ArraysAsListToImmutableRecipe.java
```diff
@@ -0,0 +1,86 @@
...
+                new MethodMatcher("java.util.Collections unmodifiableList(..)");
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(
+                final J.MethodInvocation method, final ExecutionContext executionContext) {
+            if (asListMatcher.matches(method, true)) {
+                final J.MethodInvocation parentInvocation =
+                        getCursor().getParentOrThrow().firstEnclosing(J.MethodInvocation.class);
+                if (unmodifiableListMatcher.matches(parentInvocation, true)) {
+                    return super.visitMethodInvocation(method, executionContext);
```
timtebeek, line 69: Typically we start our visitor methods with a single call to `super.visitMethodInvocation`, instead of the two code paths that call that method currently. Not saying this is wrong, but could be good to standardize if there was no particular reason to diverge from that standard.

rewrite-java/src/main/java/org/openrewrite/java/recipes/ArraysAsListToImmutableRecipe.java
marynasavchenko: removed

rewrite-java/src/test/java/org/openrewrite/java/recipes/ArraysAsListToImmutableRecipeTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Test
+    void wrapArrayListToBeUnmodifiableWhenStrings() {
+        rewriteRun(
+          java("""
+              import java.util.Arrays;
+              
+              class A {
+                 public static final List<String> entries = Arrays.asList("A", "B");
```
JLLeitschuh, line 38: IIRC, `Arrays.asList` already returns a list that is unmodifiable.

JLLeitschuh: Ah, I'm wrong, you can't append to it, but you can replace existing values.

timtebeek: Indeed; more details discussed in [this thread](https://github.com/openrewrite/rewrite/issues/2413#issuecomment-1586767731); I like your suggestion of using data flow to make this recipe safe to execute. I've not used data flow myself before, but would be good to explore here.

JLLeitschuh: To use data flow, this recipe would need to be moved to the `cleanup` package in `rewrite-static-analysis`

timtebeek: That's indeed perhaps a better place for this recipe; sorry about that @marynasavchenko . Would you mind targeting https://github.com/openrewrite/rewrite-static-analysis/tree/main/src/main/java/org/openrewrite/staticanalysis for these files?
As seen above that's necessary to use data flow, which we need to be sure we can safely apply this recipe.

timtebeek:
Closed pending a move to rewrite-static-analysis, which we need for the data flow analysis to determine if it's safe to do this replacement: we can't replace when list elements or order is modified. Great start, looking forward to the full implementation.

PR #3311, opened 2023-06-13
Tweak style autodetect

sambsnyd:
Rearrange style autodetection so that it can come before LST serialization


shanman190 (approved):
LGTM!

shanman190 (commented):
The current code looks good, but upon looking at the attached PR, the issue remains in play I believe.

knutwannheden:
@shanman190 I believe the other PR is only failing because this one hasn't been integrated yet.

Issue #3312, opened 2023-06-13
YAML encoding errors with Windows 


thomaszub:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.0.0
- Maven plugin v5.0.0

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project.

## What is the smallest, simplest way to reproduce the problem?
```yaml
some: # german umlaute äöü
    key: value
```

## What did you expect to see/What did you see instead?
After applying a YAML recipe, the encoding is broken.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Yes, I make a PR.

timtebeek:
Appreciate you're looking to provide a fix!
 -- Timotheüs

knutwannheden:
@thomaszub Just for clarification: Is this issue new with OpenRewrite 8.0? Did this work as expected with older versions?

thomaszub:
@knutwannheden: I tested it with the maven-plugin 4.42.0 (rewrite 7.38.0) and the bug is present.

PR #3313, opened 2023-06-13
Fix/yaml encoding

thomaszub:
## What's changed?
- The InputStreamReader use UTF-8 instead of the default encoding (ISO-8859-1 for windows)
- Close all readers and streams with try-resource

## What's your motivation?
Issue #3312

I added the streams and readers to the try-resource.
This bug is only testable under Windows, so no test is added. Tested it locally and it works.

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -34,11 +34,9 @@ import org.yaml.snakeyaml.reader.StreamReader;
 import org.yaml.snakeyaml.scanner.Scanner;
 import org.yaml.snakeyaml.scanner.ScannerImpl;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStreamReader;
+import java.io.StringReader;
 import java.io.UncheckedIOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.*;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -106,9 +106,12 @@ private Yaml.Documents parseFromInput(Path sourceFile, EncodingDetectingInputStr
             yamlSourceWithVariablePlaceholders.append(yamlSource, pos, yamlSource.length());
         }
 
-        try (FormatPreservingReader reader = new FormatPreservingReader(
-                new InputStreamReader(new ByteArrayInputStream(yamlSourceWithVariablePlaceholders.toString()
-                        .getBytes(StandardCharsets.UTF_8))))) {
+        try (
```
knutwannheden, line 109: I suggest we skip the entire encoding-decoding and just use a `StringReader`. Something like:
```java
       try (FormatPreservingReader reader = new FormatPreservingReader(
                new StringReader(yamlSourceWithVariablePlaceholders.toString()))) {
```

knutwannheden (commented):
I left a small comment. Thanks a lot for the fix!

rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
thomaszub: Good point, I change it.

knutwannheden:
@thomaszub Thanks again for the fix!

PR #3324, opened 2023-06-15
Refactor `MethodMatcher` and extract Interface

JLLeitschuh:
Extracts a majority of the logic of `MethodMatcher` into an interface `SimpleMethodMatcher`.

Simplifies the `SimpleMethodMatcher` into 3 basic predicates:
 - `matchesTargetType`
 - `matchesMethodName`
 - `matchesParameterTypes`

A custom implementation of `SimpleMethodMatcher` can implement those predicates however they choose.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.StringJoiner;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static java.util.stream.Collectors.joining;
 import static org.openrewrite.java.tree.TypeUtils.fullyQualifiedNamesAreEqual;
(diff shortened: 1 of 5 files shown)


JLLeitschuh:
If we do this, then `GenericExternalModel` can implement `SimpleMethodMatcher` in `rewrite-analysis`, which will let us create the custom `SimpleMethodMatcher` matching logic we need.

JLLeitschuh:
I'm also not wedded to the name `SimpleMethodMatcher` if you want a different name. But I thought `IMethodMatcher` wouldn't fly 😆

rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
```diff
@@ -141,28 +142,6 @@ public MethodMatcher(JavaType.Method method) {
...
-            return false;
-        }
-
-        StringJoiner joiner = new StringJoiner(",");
-        for (JavaType javaType : type.getParameterTypes()) {
-            String s = typePattern(javaType);
-            if (s != null) {
-                joiner.add(s);
-            }
-        }
```
JLLeitschuh, line 161: This was duplicated twice, or three times if you include this:
```java
        String arguments = method.getParameters().stream()
                .map(this::variableDeclarationsType)
                .filter(Objects::nonNull)
                .map(MethodMatcher::typePattern)
                .filter(Objects::nonNull)
                .collect(joining(","));
```

rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
```diff
@@ -210,41 +186,31 @@ private JavaType variableDeclarationsType(Statement v) {
...
-    public boolean matches(@Nullable J.MethodInvocation method) {
-        return matches(method, false);
-    }
-
     /**
      * Prefer {@link #matches(J.MethodInvocation)}, which uses the default `false` behavior for matchUnknownTypes.
      * Using matchUnknownTypes can improve Visitor resiliency for an AST with missing type information, but
      * also increases the risk of false-positive matches on unrelated MethodInvocation instances.
      */
     public boolean matches(@Nullable J.MethodInvocation method, boolean matchUnknownTypes) {
```
JLLeitschuh, line 194: `matchUnknownTypes` should really be an argument to `MethodMatcher` instead of an argument to this specific method IMHO.

PR #3326, opened 2023-06-15
Add `FindMethodsVisitor` and refactor `FindMethods`

JLLeitschuh:

Also adds a new incubating `JavaTypeMethodMatcher` interface that can be seen as `MethodMatcher` "light". A significantly smaller interface that consumers can rely upon over depending upon the entire `MethodMatcher` API. Thus decoupling the visitors away from the implementation details of how methods are actually matched against, and leaving them with the sole responsibility of performing the appropriate actions when matching occurs.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaTypeMethodMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaTypeMethodMatcher.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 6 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaTypeMethodMatcher.java
```diff
@@ -0,0 +1,60 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
```
JLLeitschuh, line 16: I'm also happy to move this to an `internal` package if that's a better/safer home for it in the short term.

JLLeitschuh:
CI passes locally. Happy to move `JavaTypeMethodMatcher` to the `internal` package if needed in a later change.

PR #3331, opened 2023-06-16
Drop BuildToolFailure marker, recipe and table

timtebeek:
## What's changed?
Drop Marker, Recipe and Table as uploaded ASTs are no longer desired.

## What's your motivation?
For https://github.com/moderneinc/moderne-cli/issues/459


sambsnyd (approved):
If you guys don't use these then I don't believe there is any reason to keep them

timtebeek:
Probably best to drop this _after_ https://github.com/moderneinc/moderne-cli/pull/462
I'd hoped to use this to increase the number of successful CI builds, but it looks like we're aiming for rudderstack there now.

timtebeek:
We might want keep this around and publish the failures under different coordinates, so we can still analyze these at scale to improve java detection, ingestion issues and/or pin or disable projects in repos.csv.

rpau (approved):

PR #3337, opened 2023-06-19
Add scope to injected class

m-brophy:
## What's changed?
This is a recipe to detect injected classes and add a `Dependent` class anotation to that class if there isn't one already present

## What's your motivation?
The need was prompted while trying to migrate jave EE projects to quarkus. quarkus  doesn't use beans.xml and needs bean scope to create a bean

The unit test fails, because when the test classes are instantiated the parser doesn't create the member variable's `@Inject` annotation meaning that the recipe doesn't fire. I'm creating this PR to raise an issue as to why this happens

Apologies in advance if it's a coding error on my part!

### Checklist


kunli2:
Hi @m-brophy, seems the unit test is failing, can you double-check? 
and also, should this recipe be part of `jakarta-ee` and move to repo `rewrite-migrate-java`?

like this https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml

Also, it's missing license headers, you can run `./gradlew lF` to generate license headers automatically.

m-brophy:
Hi @kunli2 , I raised the PR so I could raise an issue as to why the unit test was failing, here it is - https://github.com/openrewrite/rewrite/issues/3338

The test is failing because the test framework doesn't include the member variable annotation in the class it creates from the test spec, and I think that's a bug

rewrite-java/src/test/java/org/openrewrite/java/AddScopeToInjectedClassTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new AddScopeToInjectedClass())
+                .cycles(1);
+             //   .expectedCyclesThatMakeChanges(1);
+    }*/
+
+    @DocumentExample
+    @Test
+    void scopeRequired() {
+        rewriteRun(spec -> spec.recipe(new AddScopeToInjectedClass())
```
timtebeek, line 22: You might need to add an explicit JavaParser builder here, with a classpath entry containing the annotations you use in your code samples below (depending perhaps on the version of Java used). So something like `spec.parser(JavaParser.fromJavaVersion().classpath("jar-containing-javax-inject"))`.

Right now I suspect the annotations in your code samples below are not found and added to the lossless semantic tree, which indeed if I step through with the debugger appear to be missing on the `Foo.service` field.

Instead of adding a classpath entry you can also add those annotations directly as code blocks with something like:
```java
java("""
  package javax.enterprise.context;
  @Target({ElementType.Type})
  public @interface Dependent {
  }
  """),
```

Once the code snippets you pass into the unit tests are valid, as in not referring to any missing types, then you should be able to step through your recipe and see what next to change.

rewrite-java/src/main/java/org/openrewrite/java/AddScopeToInjectedClass.java
```diff
@@ -0,0 +1,111 @@
...
+                    for (J.ClassDeclaration classDeclaration : cu.getClasses()) {
+                        JavaType.FullyQualified type =(JavaType.FullyQualified) classDeclaration.getType();
+                        for (JavaType.Variable variable : type.getMembers())
+                        if (variableTypeRequiresScope(variable)) {
+                            acc.getScopeTypes().add((JavaType.FullyQualified)variable.getType());
+                        }
+                    }
+                }
+                return tree;
+            }
```
timtebeek, line 61: You can potentially simplify this a bit by returning a JavaIsoVisitor. Here's a quick sample.
```suggestion
        return new JavaIsoVisitor<ExecutionContext>() {
            @Override
            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext executionContext) {
                J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, executionContext);
                JavaType.FullyQualified type = cd.getType();
                for (JavaType.Variable variable : type.getMembers()) {
                    if (variableTypeRequiresScope(variable)) {
                        acc.getScopeTypes().add((JavaType.FullyQualified)variable.getType());
                    }
                }
                return cd;
            }
        };
```

knutwannheden:
Should this PR possibly be moved to rewrite-migrate-java?

m-brophy:
moving this to rewrite-migrate-java so closing this PR

Issue #3346, opened 2023-06-21
`AutoFormat` add extraneous space before annotated return type expression


kunli2:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.1.2
Run recipe `org.openrewrite.java.format.AutoFormat`

## What is the smallest, simplest way to reproduce the problem?

```java
class A {
    public @Deprecated String method() {
        return "name";
    }
}
```

## What did you expect to see?

no change is expected

## What did you see instead?

```diff
--- public @Deprecated String method() {
+++ public  @Deprecated String method() {
```

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

PR #3347, opened 2023-06-21
Fix AutoFormat issue of extraneous spaces added for annotated return type

kunli2:
Fixes https://github.com/openrewrite/rewrite/issues/3346

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
@@ -21,10 +21,7 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.marker.ImplicitReturn;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JContainer;
-import org.openrewrite.java.tree.JavaSourceFile;
-import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.*;
 
 @RequiredArgsConstructor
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
```diff
@@ -130,8 +127,13 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, P
 
         if (m.getReturnTypeExpression() != null && m.getReturnTypeExpression().getPrefix().getWhitespace().isEmpty()) {
             if (!first) {
-                m = m.withReturnTypeExpression(m.getReturnTypeExpression()
-                        .withPrefix(m.getReturnTypeExpression().getPrefix().withWhitespace(" ")));
+                TypeTree returnTypeExpression = m.getReturnTypeExpression();
+                // If it's a J.AnnotatedType, because the first annotation has its prefix, so don't need to set the
+                // prefix again to avoid two spaces.
+                if (!(returnTypeExpression instanceof J.AnnotatedType)) {
+                    m = m.withReturnTypeExpression(returnTypeExpression
```
traceyyoshima, line 134: Hi Kun, nice work -- I'm curious whether IntelliJ will reduce the spaces on an inline annotated return type.
Does the prefix of the `@Deprecated` annotation get reduced to 1 space if more than 1 space exists in the prefix?

kunli2: Hi Tracey, good call out, I verified IntelliJ does reduce the space to 1 space before the annotation. 
I think we probably want the same change here.
let me add the logic for that one.

kunli2: Added the logic to reduce the 1st annotation prefix to be 1 single space.

PR #3355, opened 2023-06-25
More accurate type checking

knutwannheden:
The `FindMissingTypes` doesn't quite detect all missing types. This is an attempt at cleaning some of these cases up.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/Assertions.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/Assertions.java
@@ -70,18 +70,20 @@ public class Assertions {
             List<FindMissingTypes.MissingTypeResult> missingTypeResults = FindMissingTypes.findMissingTypes(sf);
             missingTypeResults = missingTypeResults.stream()
                     .filter(missingType -> {
-                        if (typeValidation.identifiers() && missingType.getJ() instanceof J.Identifier) {
-                            return true;
-                        } else if (typeValidation.classDeclarations() && missingType.getJ() instanceof J.ClassDeclaration) {
-                            return true;
-                        } else if (typeValidation.methodInvocations() && missingType.getJ() instanceof J.MethodInvocation) {
-                            return true;
-                        } else if (typeValidation.constructorInvocations() && missingType.getJ() instanceof J.NewClass) {
-                            return true;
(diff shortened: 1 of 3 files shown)


knutwannheden:
@sambsnyd This may cause some test failures in recipe repos, that we would have to clean up afterwards.

I am also a bit concerned about limitations in `JavaTemplate` which can lead to missing type attribution, which I don't quite know what to do about. I am referring to this case in rewrite-static-analysis, where we currently stub out a type to make sure it is accessible to the JavaParser: https://github.com/openrewrite/rewrite-static-analysis/blob/29cc68915f2d136e15b7d8e882fbd0d5dacf61fb/src/main/java/org/openrewrite/staticanalysis/ReplaceLambdaWithMethodReference.java#L103-L105 I think this is quite a hack and will also result in incomplete type attribution (the type won't necessarily have the correct kind, visibility, supertype, etc.). Also the member reference check I am adding here will fail for this recipe unless the stub also includes method declarations (which is a slippery slope as the method could be inherited...).

I am thinking that for the mentioned example we should just create the correct type attribution manually instead of using `JavaTemplate`.

rewrite-java/src/main/java/org/openrewrite/java/Assertions.java
```diff
@@ -70,18 +70,20 @@ public static void assertValidTypes(TypeValidation typeValidation, J sf) {
...
-                            return true;
-                        } else if (typeValidation.constructorInvocations() && missingType.getJ() instanceof J.NewClass) {
-                            return true;
-                        } else if (typeValidation.methodDeclarations() && missingType.getJ() instanceof J.MethodDeclaration) {
-                            return true;
+                        if (missingType.getJ() instanceof J.Identifier) {
+                            return typeValidation.identifiers();
+                        } else if (missingType.getJ() instanceof J.ClassDeclaration) {
+                            return typeValidation.classDeclarations();
+                        } else if (missingType.getJ() instanceof J.MethodInvocation || missingType.getJ() instanceof J.MemberReference) {
```
knutwannheden, line 77: We could also introduce a separate option for this, as we have it for `J.NewClass` (constructor invocations) as well. I am indecisive about this one.

sambsnyd (approved):
Seems reasonable to me

PR #3363, opened 2023-06-26
Issue 1698: adding additional checks for internal class imports

tclayton-newr:
## What's changed?

This PR is attempting to address imports being removed by the RemoveUnusedImports recipe. It attempts to do this by creating a regex matcher to match against class paths that include both `.` and `$` class delimiters. Since internal classes use the `$`, it causes them to not be recognized when checking if a variable is still referencing the class path. 

## What's your motivation?

My motivation is to fill this gap where RemoveUnusedImports causes an error in the resulting file by removing imports still in use. 

I would like reviewing to scrutinize the regex approach I took. I think I have convinced myself that this approach will not be too generous as to result in false negatives.

I currently cannot think of any. 

### Checklist
- ~[ ] I've updated the documentation (if applicable)~


jkschneider:
Thank you @tclayton-newr !

PR #3375, opened 2023-06-29
Optimize `LatestVersion#countVersionParts()`

knutwannheden:
Using `Scanner` has the downsides of using regexp and also allocating a 1KiB `HeapCharBuffer` every time.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
@@ -19,7 +19,6 @@ import org.openrewrite.Validated;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 
-import java.util.Scanner;
 import java.util.regex.Matcher;
 
 import static java.lang.Integer.parseInt;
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
```diff
@@ -72,14 +71,21 @@ static String normalizeVersion(String version) {
...
-        scanner.useDelimiter("[.\\-$]");
-        while (scanner.hasNext()) {
-            String part = scanner.next();
-            if (part.isEmpty() || !Character.isDigit(part.charAt(0))) {
-                break;
+        int len = version.length();
+        int lastSepIdx = -1;
+        for (int i = 0; i < len; i++) {
+            char c = version.charAt(i);
+            if (c == '.' || c == '-' || c == '$') {
```
sambsnyd, line 78: This is slightly different from the delimiter regex which originally also included "\". But maybe that doesn't matter? Haven't ever seen a backslash in a version number

sambsnyd (approved):

rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java
knutwannheden: I think the backslash in `[.\\-$]` is so that the regex doesn't interpret it as a range, as it is in square brackets.

PR #3378, opened 2023-06-30
Cache the GradleWrapper validation and any initial remote archive download by uri

shanman190:


rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java
```diff
@@ -77,16 +80,34 @@ public class RemoteArchive implements Remote {
      */
     List<String> paths;
 
+    private static Map<URI, Path> downloadCache = new HashMap<>();
```
jkschneider, line 83: Can we instead pass this through `ExecutionContext` much like we do with `MavenPomCache` interface?

jkschneider: What I'm imagining is providing a `RemoteArchiveCache` interface whose default implementation is similar to `LocalMavenArtifactCache`, that would be able to find previously downloaded archives at a predictable path on disk and that wouldn't require static state.

shanman190: I don't see any issue with that at the moment. I'll get it adjusted to match that implementation.

shanman190: Added the first iteration via https://github.com/openrewrite/rewrite/pull/3378/commits/215d3fae04b14c994953e21b6e65e3db57abd92d.

jkschneider:
Superseded by #3391

PR #3383, opened 2023-06-30
Only validate recipes once

knutwannheden:
The recipes should only be validated once per recipe run and not once per cycle, phase, and source file.


knutwannheden:
This still needs fixing, as currently the root recipe is getting validated. I am thinking it would probably be enough if we validate the root recipe and count on it validating all its recipes in its recipe list (recursively). AFAIK that is what recipes do by default.

PR #3384, opened 2023-06-30
Allow AddProperty reuse

timtebeek:
## What's changed?
- Use PropertiesVisitor for the returned visitor
- Have PropertiesVisitor.visitFile return Properties.File
- Remove now excess casts

## What's your motivation?
As discovered while going over #3276

Might require a code change in return type in the rare(?) case that people themselves override `PropertiesVisitor.visitFile`.
Or was there a specific reason to return `Properties` instead of `Properties.File`, which implementes `Properties`?

Not really; quick proposal to validate the approach, open to other viewpoints or solutions.
Without this change there would have to be two unsafe casts to reuse AddProperty:
1. for the return of `AddProperty.getVisitor()` from `TreeVisitor` to `PropertiesVisitor`
2. another for return of `PropertiesVisitor visitFile(..)` from `Properties` to `Properties.File`

## Any additional context
In #3276 it was requested if it's possible to reuse `AddProperty`. That's awkward right now as `getVisitor()` returns a generic `TreeVisitor`, and the `visitFile()` within that returns a `Properties` instead of `Properties.File`. 

### Checklist

### DIFF
--- a/rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
+++ b/rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
@@ -59,11 +59,11 @@ public class AddProperty extends Recipe {
     }
 
     @Override
-    public PropertiesVisitor<ExecutionContext> getVisitor() {
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new PropertiesVisitor<ExecutionContext>() {
             @Override
-            public Properties.File visitFile(Properties.File file, ExecutionContext executionContext) {
-                Properties.File p = super.visitFile(file, executionContext);
+            public Properties visitFile(Properties.File file, ExecutionContext executionContext) {
+                Properties.File p = (Properties.File) super.visitFile(file, executionContext);
                 if (!StringUtils.isBlank(property) && !StringUtils.isBlank(value)) {
                     Set<Properties.Entry> properties = FindProperties.find(p, property, false);
                     if (properties.isEmpty()) {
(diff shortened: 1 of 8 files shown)


rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
```diff
@@ -59,11 +59,11 @@ public String getDescription() {
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor() {
+    public PropertiesVisitor<ExecutionContext> getVisitor() {
```
jkschneider, line 62: Why make this more specific?

timtebeek: In #3276 it was requested if it's possible to reuse `AddProperty`. That's awkward right now as `getVisitor()` returns a generic `TreeVisitor`, and the `visitFile()` within that returns a `Properties` instead of `Properties.File`.

_(taken from above, in case you'd missed it; if my description is not clear do please tell!)_

jkschneider: (2) is the purpose of `PropertiesIsoVisitor` vs. `PropertiesVisitor`.

timtebeek: Ah didn't realize we had an iso visitor for properties as well; that's indeed better. I'll clean this up when I find a moment.

timtebeek: Now we only switch to `PropertiesIsoVisitor` and change the `getVisitor()` return type. Much better.

knutwannheden (approved):

Issue #3377, opened 2023-06-30
[Rewrite8 migration][ERROR] java.lang.NoSuchMethodError: org.openrewrite.java.MethodMatcher.matches(...)


jdelobel:
<!--
Thank you for reporting an issue with OpenRewrite!
We appreciate you taking the time to help us improve.
Please fill out the template below to help us understand and reproduce the issue.
Feel free to delete any sections that don't apply to your issue.
-->

## What version of OpenRewrite are you using?
<!--
Whenever possible please try to replicate your issue with the latest versions of OpenRewrite.
The latest major and minor versions of OpenRewrite projects are usually listed here:
https://docs.openrewrite.org/reference/latest-versions-of-every-openrewrite-module
For patch releases check the GitHub Releases page for the respective project.

We release every few weeks, so it's possible that your issue has already been fixed.

If you want to try the most recent changes that haven't been fully released yet, you can check out our snapshot releases
https://docs.openrewrite.org/reference/snapshot-instructions
-->
I am using

- OpenRewrite v2.0.2 bom
- Maven plugin v5.2.4
- rewrite-_module_ v2.0.2

## How are you running OpenRewrite?
<!--
Are you using the Maven plugin, Gradle plugin, Moderne CLI, Moderne.io or something else?
Is your project a single module or a multi-module project?

Can you share your configuration so that we can rule out any configuration issues?

Is your project public? If so, can you share a link to it?
Code snippets can also be shared privately via [our public Slack](https://join.slack.com/t/rewriteoss/shared_invite/zt-nj42n3ea-b~62rIHzb3Vo0E1APKCXEA).
-->
I am using the Maven plugin, and my project is a single module project.
```xml
  <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>5.2.4</version>
                <configuration>
                    <activeRecipes>
                        [all my custom recipe]
                    </activeRecipes>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-migrate-java</artifactId>
                        <version>2.0.2</version>
                    </dependency>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-testing-frameworks</artifactId>
                        <version>2.0.2</version>
                    </dependency>
                    <dependency>
                        <groupId>com.my.org</groupId>
[...]

PR #3387, opened 2023-06-30
Replicate `FindAndReplace` `No group with name {name}`

timtebeek:
## What's changed?
Added a test to replicate a bug; not yet a fix.

## What's your motivation?
Replicate a bug reported with FindAndReplace around `find` strings containing `{..}`.

Suggestions for a fix welcome.

@jdelobel

Not yet.

## Any additional context
Reported in https://github.com/openrewrite/rewrite/issues/3377#issuecomment-1614637746

### Checklist


philippe-granet:
I have the same problem and it works if I escape dollar sign (`\\${...}`)

timtebeek:
That works indeed; thanks! This one had slipped my mind after capturing it from a Slack thread. Just today we saw the same fix in;
- https://github.com/openrewrite/rewrite-logging-frameworks/pull/120

rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
```diff
@@ -134,7 +134,8 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                 if (!matcher.find()) {
                     return sourceFile;
                 }
-                String newText = matcher.replaceAll(replace);
+                String replacement = Boolean.TRUE.equals(regex) ? replace : replace.replaceAll("\\$", "\\\\\\$");
```
joanvr, line 137:
```suggestion
                String replacement = replace;
                if (!Boolean.TRUE.equals(regex)) {
                    replacement = replacement.replace("$", "\\$");
                }```
WDYT to have it more similar to the similar issue on line 117? Also, use replace instead of replaceAll since we do not need a regex?

timtebeek: Good suggestion, thanks! Let's merge as soon as the tests pass.

joanvr (approved):

PR #3395, opened 2023-07-04
Improved newVersion description

joanvr:
## What's changed?
Make it explicit the possibility to use `latest.release` and `latest.patch` as newVersion to upgrade to the latest available version or the latest patch version of the current version (only if already a valid semantic version).

## What's your motivation?
Right now it's not clear for users how to get this behavior since it's not properly documented anywhere.

Maybe trying to find a better text to explain the functionality.

@mike-solomon

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -70,7 +70,8 @@ public class UpgradeDependencyVersion extends Recipe {
     @Option(displayName = "New version",
             description = "An exact version number or node-style semver selector used to select the version number. " +
                           "You can also use `latest.release` for the latest available version and `latest.patch` if " +
-                          "the current version is a valid semantic version.",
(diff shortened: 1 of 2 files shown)


rpau (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -68,13 +68,15 @@ public class UpgradeDependencyVersion extends Recipe {
     String artifactId;
 
     @Option(displayName = "New version",
-            description = "An exact version number or node-style semver selector used to select the version number.",
+            description = "An exact version number or node-style semver selector used to select the version number. " +
+                          "You can also use `latest.release` for the latest available version and `latest.patch` if " +
+                          "the current version is a valid semantic version.",
```
knutwannheden, line 73: I like this, but I think it would be even better if we could also add a Markdown link to the documentation page on [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors). On that page I notice that the documentation for `latest.patch` is missing.

mike-solomon (approved):
Changes look good to me. Thanks for doing this :)

Issue #3275, opened 2023-05-27
`RemoveUnusedImports` removes used imports and adds duplicated ones


pzygielo:
## What version of OpenRewrite are you using?
I am using
- Maven plugin v4.46.0

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project

## What is the smallest, simplest way to reproduce the problem?
https://github.com/pzrep/openrewrite-unused-import-record

## What did you expect to see?
I expected to see no changes.

## What did you see instead?
https://github.com/pzrep/openrewrite-unused-import-record/actions/runs/5100022041/jobs/9168054354

Depends on number of imports done with * I guess:

1. star-import is replaced with explicit imports, including duplicated [`Record`] one:
```diff
diff --git a/src/main/java/pzrep/p2/Client1.java b/src/main/java/pzrep/p2/Client1.java
index 2533533..edfecd5 100644
--- a/src/main/java/pzrep/p2/Client1.java
+++ b/src/main/java/pzrep/p2/Client1.java
@@ -1,7 +1,10 @@
 package pzrep.p2;
 
 import pzrep.p1.Record;
-import pzrep.p1.*;
+import pzrep.p1.A;
+import pzrep.p1.B0;
+import pzrep.p1.B1;
+import pzrep.p1.Record;
 
 class Client1 {
     void f(Record r) {
```
This is unfortunate but allows to compile modified class.

2. required import gets removed
```diff
diff --git a/src/main/java/pzrep/p2/Client2.java b/src/main/java/pzrep/p2/Client2.java
index 2bb52f5..50aea66 100644
--- a/src/main/java/pzrep/p2/Client2.java
+++ b/src/main/java/pzrep/p2/Client2.java
@@ -1,6 +1,5 @@
 package pzrep.p2;
 
-import pzrep.p1.Record;
 import pzrep.p1.*;
 
 class Client2 {
```
This makes modified source to be uncompilable. (That's the original issue I observed with https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/Record.html)

pzygielo:
I see that duplicated `Record` import (p. 1 above) is no longer inserted (looks that something in maven plugin [5.0.0->5.2.1](https://github.com/openrewrite/rewrite-maven-plugin/compare/v5.0.0...v5.2.1) helps :smile:)

joanvr:
[...]

PR #3397, opened 2023-07-05
Issue 3275/fix remove used param

tclayton-newr:
## What's changed?
https://github.com/openrewrite/rewrite/issues/3275 pointed out an issue with incorrectly removing an import that was being used as a function parameter. This change updates the class path matching logic to check for either a `.` or a `$` class path separator. 

## What's your motivation?
My motivation is to resolve https://github.com/openrewrite/rewrite/issues/3275 and make the RemoveUnusedImports recipe more reliable for wider adoption. 

I co-opted the predicate used in https://github.com/openrewrite/rewrite/pull/3363 to more generally match against class paths that have a `$` as a separator. 

Any maintainer 😄 

n/a

## Any additional context
n/a

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
@@ -1072,6 +1072,68 @@ class RemoveUnusedImportsTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/3275")
+    @Test
+    void doesNotRemoveReferencedClassesBeingUsedAsParameters() {
+        rewriteRun(
+          java(
+            """
+              package com.Source.$;
+
+              public class A {
+                  public static final short SHORT1 = (short)1;
+
+                  public short getShort1() {
+                    return SHORT1;
+                  }
+              }
+              """
+          ),
+          java(
+            """
+              package com.test;
+
+              import com.Source.$.A;
+
+              class Test {
+                  void f(A classA) {
+                    classA.getShort1();
+                  }
+              }
+              """
+          )
+        );
+
+        rewriteRun(
+          java(
+            """
+              package com.Source.mine;
+
+              public class A {
+                  public static final short SHORT1 = (short)1;
+                  
+                  public short getShort1() {
+                    return SHORT1;
+                  } 
+              }
+              """
+          ),
+          java(
+            """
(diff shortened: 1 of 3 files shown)


knutwannheden:
@tclayton-newr Thanks for contributing this PR! I will take a look at it later today or tomorrow.

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
```diff
@@ -1072,6 +1072,68 @@ public static void main(String[] args) {
...
+
+              class Test {
+                  void f(A classA) {
+                    classA.getShort1();
+                  }
+              }
+              """
+          )
+        );
+
```
timtebeek, line 1106: I'm wondering if this should be two tests; we don't usually see two invocations of `rewriteRun` in a single test method.

rewrite-java/src/main/java/org/openrewrite/java/RemoveUnusedImports.java
```diff
@@ -285,6 +286,10 @@ public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionCon
         }
     }
 
+    private static Predicate<String> classPathMatcher(final String target){
+        return (s) -> s.matches(target.replaceAll("[\\.\\$]", "(\\\\\\.|\\\\\\$)"));
```
knutwannheden, line 290: I was going to suggest:
```suggestion
        return (s) -> s.matches(target.replaceAll("[.$]", "[.\\$]"));
```
but I now wonder if we can't simply do:
```suggestion
            return s -> TypeUtils.fullyQualifiedNamesAreEqual(s, target);
```

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
tclayton-newr: I am happy to split them out

rewrite-java/src/main/java/org/openrewrite/java/RemoveUnusedImports.java
tclayton-newr: Interesting, I'll have a look at the second approach

joanvr (approved):

joanvr:
Thank you very much for your contribution and addressing all the comments :)

PR #3399, opened 2023-07-07
Use default constructor as `@JsonCreator` for `JavaType` implementations

knutwannheden:
In order for Jackson to be able to correctly deserialize an object graph containing cycles, the `@JsonCreator` annotated method must not contain the properties, which correspond to the edges participating in the cycles.

This commit adds a new (package private) default constructor to all `JavaType` implementations and moves the `@JsonCreator` annotation to this new constructor. This then allows Jackson to properly serialize and deserialize the object cycles. Eventually this will then allow removing the `managedReference` fields from the implementation classes to save some more memory.


knutwannheden:
I can investigate if we can again remove some of the `@NonFinal` annotations by instead declaring these properties in the `@JsonCreator` annotated constructor.

PR #3402, opened 2023-07-07
JavaTemplate.Builder.imports - throw on blank import

nmck257:
Small change; including a blank `String` in the `imports` when building a `JavaTemplate` now throws an exception, instead of silently ignoring the blank entry.

I think failing fast here is more helpful and matches the spirit of a non-null API

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
@@ -153,22 +153,20 @@ public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
 
         public Builder staticImports(String... fullyQualifiedMemberTypeNames) {
             for (String typeName : fullyQualifiedMemberTypeNames) {
-                if (shouldAddImport(typeName)) {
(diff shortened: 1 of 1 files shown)


knutwannheden (approved):

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -153,22 +153,20 @@ public Builder imports(String... fullyQualifiedTypeNames) {
 
         public Builder staticImports(String... fullyQualifiedMemberTypeNames) {
             for (String typeName : fullyQualifiedMemberTypeNames) {
-                if (shouldAddImport(typeName)) {
```
knutwannheden, line 156: There is apparently one more reference on line 147, when needs refactoring.

nmck257: that's what I get for using the online editor :)

knutwannheden: You are very brave! :-)

PR #3404, opened 2023-07-07
String fields are null by default.

alexkir28:
## What's changed?
When creating a Recipe instance String constructor arguments will be null

## What's your motivation?
Recipes with required parameters with type String and that don't override validate method will be considered valid since those parameters would be initialized with an empty string. As a result, default implementation won't catch it.

I'd like to know whether there is a negative implication of this change.

@sambsnyd @jkschneider 

## Any additional context
This change is part of the scope of the issue that allows run recipe without required parameters been passed on


alexkir28:
Should we make enum parameters also null by default instead of choosing the first value?

sambsnyd:
> Should we make enum parameters also null by default instead of choosing the first value?

That sounds reasonable, although I'm not sure exactly what the effects might be

sjungling:
> > Should we make enum parameters also null by default instead of choosing the first value?
> 
> That sounds reasonable, although I'm not sure exactly what the effects might be

@sambsnyd I think that might resolve the issue you found the other day with the _Change slf4j log level_ recipe using the value `TRACE` for the `from` and `to` fields by default

sambsnyd:
Yes, agree that sounds related @sjungling

PR #3407, opened 2023-07-10
Feature/composite scanning recipe

nmck257:
## What's changed?
Simplifies the syntax for invoking a ScanningRecipe from another ScanningRecipe, by implementing the boilerplate code to call the base ScanningRecipe methods and providing new abstract methods for the new recipe's ScanningRecipe lifecycle.

## What's your motivation?
Having to implement this pattern a few times while migrating to Rewrite 8, and seeing an opportunity to simplify and genericize.

* Is it worthwhile to codify this pattern in the first place?
* I had to give the new methods distinct names (eg `getScannerz` vs `getScanner`), because with type erasure, the two versions of the method have the same signature. I don't claim that appending `z` is a good solution, but, I didn't have a better one in mind :)
* I'd love to remove the `RT` type param, as it leaks implementation knowledge about `R` and would grow linearly if you tried to chain together multiple `CompositeScanningRecipe`s. I tried just replacing it with `?`, but I couldn't find a trick to get past the typechecks whenever I pass `acc.getValue()` to a `baseRecipe` method (ie "cannot convert `Object` to `capture##`"). Not sure if some other clever hackery could make it work

Client recipes can do this legwork themselves, but I think it's a bit tricky to figure out and implement if you're less proficient with OpenRewrite

### Checklist
- [X] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/CompositeScanningRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/CompositeScanningRecipe.java
@@ -15,11 +15,6 @@
  */
 package org.openrewrite;
 
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.ScanningRecipe;
-import org.openrewrite.SourceFile;
-import org.openrewrite.Tree;
-import org.openrewrite.TreeVisitor;
 import org.openrewrite.internal.lang.Nullable;
 
 import java.util.AbstractMap;
@@ -81,6 +76,7 @@ public abstract class CompositeScanningRecipe<T, R extends ScanningRecipe<RT>, R
         }
     }
 
+    @SuppressWarnings("unused")
     protected Collection<? extends SourceFile> generatez(T acc, Collection<SourceFile> generatedInThisCycle, ExecutionContext ctx) {
         return generatez(acc, ctx);
     }
@@ -99,6 +95,7 @@ public abstract class CompositeScanningRecipe<T, R extends ScanningRecipe<RT>, R
         }
     }
 
+    @SuppressWarnings("unused")
     protected Collection<? extends SourceFile> generatez(T acc, ExecutionContext ctx) {
         return Collections.emptyList();
     }
@@ -118,6 +115,7 @@ public abstract class CompositeScanningRecipe<T, R extends ScanningRecipe<RT>, R
         };
     }
 
+    @SuppressWarnings("unused")
     protected TreeVisitor<?, ExecutionContext> getVisitorz(T acc) {
         return TreeVisitor.noop();
     }
--- a/rewrite-core/src/main/java/org/openrewrite/CompositeScanningRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/CompositeScanningRecipe.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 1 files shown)


jkschneider:
What do you mean by this?

> invoking a ScanningRecipe from another ScanningRecipe

Can you give us a quick sketch of a recipe that uses this?

nmck257:
re @jkschneider 

For a usage example, start with this: https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/jakarta/MaybeAddJakartaServletApi.java

This is a case where we want to call a `ScanningRecipe`, with some ("single-source") `Precondition`. To make that work, the new recipe also needs to be a `ScanningRecipe`, which calls the base recipe (`AddDependency`)'s various `ScanningRecipe` methods. `doNext(Recipe)` would've simplified this, but that's gone.

Now suppose the `Precondition` in this new recipe were more complex, and itself warranted a scan phase. You might land on something like this:
```java
public class MyNewRecipe extends ScanningRecipe<Pair<NewRecipeAcc, BaseRecipeAcc>> {

private final ScanningRecipe<BaseRecipeAcc> baseRecipe = new SomeScanningRecipe();

@Override
public Pair<NewRecipeAcc, BaseRecipeAcc> getInitialValue(ExecutionContext ctx) {
    return new Pair(
        new NewRecipeAcc(),
        baseRecipe.getInitialValue(ctx)
    );
}

@Override
public TreeVisitor<Tree, ExecutionContext> getScanner(Pair<NewRecipeAcc, BaseRecipeAcc> accPair) {
        public Tree visit(Tree tree, ExecutionContext ctx) {
            baseRecipe.getScanner(accPair.second()).visit(tree, ctx);
            // new logic to set accPair.first
        }
    };
}

// override generate if the base recipe or the new recipe needs it
// or, override generate and call the base even if it appears unnecessary, in case a future change to baseRecipe starts using generate

@Override
public TreeVisitor<Tree, ExecutionContext> getVisitor(Pair<NewRecipeAcc, BaseRecipeAcc> accPair) {
        public Tree visit(Tree tree, ExecutionContext ctx) {
            Tree t = tree;
            if (accPair.first().shouldDoStuff()) t = baseRecipe.getVisitor(accPair.second());
            // maybe do some new visit logic here
            return t;
        }
    };
}
```

With `CompositeScanningRecipe`, that could instead be:
```java
public class MyNewRecipe extends CompositeScanningRecipe<NewRecipeAcc, SomeScanningRecipe, BaseRecipeAcc> {

@Override
protected SomeScanningRecipe getBaseRecipe() {
    return new SomeScanningRecipe();
}

@Override
public NewRecipeAcc getInitialValuez(ExecutionContext ctx) {
    return new NewRecipeAcc();
}

@Override
public TreeVisitor<Tree, ExecutionContext> getScannerz(NewRecipeAcc acc) {
        public Tree visit(Tree tree, ExecutionContext ctx) {
            // new logic to set accPair.first
        }
    };
}

// override generatez only if your new recipe needs it
// same with getVisitorz; only override if your new recipe is making changes beyond what's in baseRecipe

@Override
public boolean baseShouldMakeChanges(NewRecipeAcc acc) {
    return acc.shouldDoStuff();
}
```

The first example requires the implementer to know a fair amount of the internals of the base recipe, and write a bunch of error-prone boilerplate to do both the base recipe's scanning and their new logic. With the second example, you only need to know the relevant types for the class signature, and you only implement the logic specific to your recipe.

I have a private example where the base recipe is `AppendToTextFile`. And I believe I discussed this pattern in Slack a few weeks back with @shanman190, so they might also have some opinions/examples.

nmck257:
Withdrawing this PR; it'll stay in the history if we ever need it again

PR #3408, opened 2023-07-11
`UsesMethod` should rely on `TypesInUse`

knutwannheden:
If `UsesMethod` does not find any reference to the desired method in `TypesInUse` then it should return rather than traversing the entire LST in an attempt to find a match.

This change is being made for performance reasons, because this visitor is used very frequently as a precondition by recipes.


joanvr (approved):
Does this new behavior break any existing recipe?

knutwannheden:
> Does this new behavior break any existing recipe?

That is currently not even possible. The code uses `MethodMatcher#matches(MethodCall)` which will always return `false` when the method invocation doesn't have any type attribution.

If we in the future want `UsesMethod` to support a use case similar to `MethodMatcher#matchesAllowingUnknownTypes()` we can still do that.

PR #3412, opened 2023-07-12
Improve OrderImports description

mike-solomon:
When I originally looked at this recipe, I was a bit confused on what it does as the description was very generic. I think that providing more information would help people who view this recipe in our docs or in the saas who won't necessarily take the time to dig into the code.

Would appreciate a second pair of eyes to make sure I've summarized this recipe correctly. I largely borrowed from the information [here](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/OrderImports.java#L39-L41).

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/OrderImports.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/OrderImports.java
@@ -60,7 +60,9 @@ public class OrderImports extends Recipe {
 
     @Override
     public String getDescription() {
-        return "Group and order imports.";
+        return "Groups and orders import statements. If a style has been defined, this recipe will order the imports " +
+                "according to that style. If no style is detected, this recipe will default to ordering imports in " +
+                "the same way that IntelliJ does.";
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/OrderImports.java
```diff
@@ -60,7 +60,9 @@ public String getDisplayName() {
 
     @Override
     public String getDescription() {
-        return "Group and order imports.";
+        return "Groups and orders import statements. If a style has been defined, this recipe will order the imports " +
```
knutwannheden, line 63: How about a Markdown hyperlink to the docs? https://docs.openrewrite.org/concepts-explanations/styles
```suggestion
        return "Groups and orders import statements. If a [style has been defined](https://docs.openrewrite.org/concepts-explanations/styles), this recipe will order the imports " +
```

mike-solomon: Great suggestion!

timtebeek (approved):

PR #3420, opened 2023-07-17
Checking RecipeIntrospectionUtils in rewrite-test

joanvr:
## What's changed?
Added checking if a recipe can be instantiated via RecipeIntrospectionUtils, because we are using it to get RecipeDescriptor, and if they fail because of null parameters it will make the recipe execution to fail in SaaS, CLI and build plugins.

## What's your motivation?
Recent changes in #3404

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -157,13 +157,23 @@ public interface RewriteTest extends SourceSpecs {
                 .isNotNull();
 
         if (!(recipe instanceof AdHocRecipe) && !(recipe instanceof CompositeRecipe) &&
+            !(recipe.equals(Recipe.noop())) &&
             testClassSpec.serializationValidation &&
             testMethodSpec.serializationValidation) {
             RecipeSerializer recipeSerializer = new RecipeSerializer();
             assertThat(recipeSerializer.read(recipeSerializer.write(recipe)))
                     .as("Recipe must be serializable/deserializable")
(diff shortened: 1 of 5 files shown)


timtebeek:
Looks good to me, but I've only had passing exposure there so far, so I'd like to leave judgement up to @alexkir28

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -157,13 +157,23 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
...
-            assertThatCode(() -> RecipeIntrospectionUtils.constructRecipe(recipe.getClass()))
+            assertThatCode(() -> {
+                Recipe r = RecipeIntrospectionUtils.constructRecipe(recipe.getClass());
+                // getRecipeList should not fail with default parameters from RecipeIntrospectionUtils.
+                r.getRecipeList();
+                // We add recipes to HashSet in some places, we need to validate that hashCode and equals does not fail.
+                //noinspection ResultOfMethodCallIgnored
+                r.hashCode();
+                //noinspection EqualsWithItself
+                assert r.equals(r);
```
timtebeek, line 175: Maybe this won't hold up in test environments, but whenever I see an assert statement I tend to think they [might not be enabled at runtime](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html#enable-disable):
> By default, assertions are disabled at runtime.

Would it then not be better to not use assert here?

timtebeek (approved):

PR #3421, opened 2023-07-17
Further Yaml unicode issues

timtebeek:
Continue on from #3419

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/FormatPreservingReader.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/FormatPreservingReader.java
@@ -32,7 +32,6 @@ class FormatPreservingReader extends Reader {
 
     private ArrayList<Character> buffer = new ArrayList<>();
     private int bufferIndex = 0;
-    private int unicodeOffset = 0;
 
     FormatPreservingReader(Reader delegate) {
         this.delegate = delegate;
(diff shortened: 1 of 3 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/FormatPreservingReader.java
```diff
@@ -79,7 +83,7 @@ public int read(@NonNull char[] cbuf, int off, int len) throws IOException {
             for (int i = 0; i < read; i++) {
                 char e = cbuf[i];
                 if (Character.UnicodeBlock.of(e) != Character.UnicodeBlock.BASIC_LATIN && i % 2 == 0) {
-                    bufferIndex--;
+                    unicodeOffset++;
```
timtebeek, line 86: Wondering if we'd need to reset this counter when we read the next "chunk". But let's first see if we even keep this in the solution we end up with.

timtebeek:
Starting to suspect an issue with the Start/End index markers that SnakeYaml reports to us when unicode is involved; as shown in the output following https://github.com/openrewrite/rewrite/pull/3421/commits/0a26f95127f4eec8dc4b0c6503e82b39fc7bd26c.

By their own admission in [the `Mark` class JavaDoc](https://github.com/henryyan/snakeyaml/blob/79c129da303bf655241429e78ab91157d5656821/src/main/java/org/yaml/snakeyaml/error/Mark.java#L21-L24), they don't use the Mark much themselves:
> It's just a record and its only use is producing nice error messages. Parser does not use it for any other purposes.

We're relying on those Mark indices to set `lastEnd`, with some accommodations here and there for various elements.

Not sure enough yet to report an issue with those to SnakeYaml, but something to explore / verify / reject / counteract.

timtebeek:
Fixed in https://github.com/openrewrite/rewrite/pull/3507

PR #3424, opened 2023-07-18
Improve recipe descriptions

mike-solomon:
I'm working through the recipes trying to make sure that the descriptions and params make sense grammatically and things are spelled correctly. This PR is mostly dedicated to YAML recipes with 1 minor JSON recipe included.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
@@ -28,8 +28,7 @@ public class CoalesceProperties extends Recipe {
 
     @Override
     public String getDescription() {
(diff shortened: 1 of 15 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
```diff
@@ -28,8 +28,7 @@ public String getDisplayName() {
 
     @Override
     public String getDescription() {
-        return "Simplify nested map hierarchies into their simplest dot separated property form, i.e. as Spring Boot interprets " +
-                "application.yml files.";
+        return "Simplify nested map hierarchies into their simplest dot separated property form, similar to how Spring Boot interprets `application.yml` files";
```
timtebeek, line 31:
```suggestion
        return "Simplify nested map hierarchies into their simplest dot separated property form, similar to how Spring Boot interprets `application.yml` files.";
```

PR #3430, opened 2023-07-21
Update `AutoFormat` to be more generic to support J-extended language like Kotlin

kunli2:
As we know, currently, `AutoFormat` doesn't support Kotlin well.
This PR is intended to update `AutoFormat` to be more generic to support J-extended language like Kotlin.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Tree.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Tree.java
@@ -19,8 +19,10 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.openrewrite.formatting.IndentType;
 import org.openrewrite.internal.MetricsHelper;
 import org.openrewrite.internal.StringUtils;
+import org.openrewrite.internal.lang.NonNull;
 import org.openrewrite.internal.lang.Nullable;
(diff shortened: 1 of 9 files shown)


rewrite-core/src/main/java/org/openrewrite/Tree.java
```diff
@@ -36,6 +38,10 @@ default String getJacksonPolymorphicTypeTag() {
         return getClass().getName();
     }
 
+    default IndentType getIndentType() {
+        return IndentType.NONE;
+    }
+
```
sambsnyd, line 44: As explained, this is not how continuations work. Depending on where it appears a method invocation's prefix may be a continuation or it might not be. So it cannot be a property of the tree itself.

sambsnyd (changes requested):

rewrite-core/src/main/java/org/openrewrite/Tree.java
kunli2: Yes, I understand that, this is a default indent type, and it can be overridden by code depending on where it appears.
because we are using a way of putting messages on a cursor to let the `autoFormattor` know the indent type like this 
```java
getCursor().putMessage("indentType", IndentType.CONTINUATION_INDENT);
```
currently, we have default indentType for each element as well (in `preVisit` as a central place), and it will be overridden by the algorithm of formatters if it depends on the appearance place, and context. 

the benefit of this is it will allow elements like `K.Return` to tell what is the default indent type of it. otherwise from AutoFormat perspective, it has no idea about the correct indent type.

kunli2:
Since we are going to separate autoFormat by language, this becomes unnecessary. Closing out.

PR #3434, opened 2023-07-22
RenameVariable incorrectly renames javadoc parameter

haumacher:
## What's changed?
Added test case for problem with invalid JavaDoc rename:

If a field is renamed, the JavaDoc of a method parameter with the same name is also renamed, if the parameter and the field had the same name before. The rename of the JavaDoc is invalid, because it does not refer to the field.

## What's your motivation?
Get the bug fixed.

@timtebeek As discussed on Discord I opened a pull-request with a test case reproducing the problem. 

### Checklist


timtebeek:
Thanks a lot for reporting this here! Sure helps to be able to step through with the debugger.

So if I understood correctly we should not rename the method parameter JavaDoc in this snippet, since it's the field that's renamed, not the method parameter. Correct?
```diff
-                    * @param name My parameter.
+                    * @param _name My parameter.
                     */
                    int fooA(String name) {
                        return name.length() + this._name.length();
```

haumacher:
Yes, you are correct. The name after `@param` is the name of the parameter, not the name of the renamed field and should not be touched.

timtebeek:
Had a brief look but I'm getting the sense that you're further along in analyzing this than I am at this stage. 😅 
Indeed sounds like we're off in detecting whether the variable we're renaming the JavaDoc for is still the same.
I'd need to find time for a more thorough look at a fix, but that's proving troublesome these days unfortunately.
Keep us updated if you learn any more!

traceyyoshima:
@timtebeek @haumacher I took a look and we can prevent changes on JavaDoc parameters by checking the cursor ahead of time. I'll have a fix for this shortly.

timtebeek:
Thanks a lot for the report @haumacher ; we'll continue this in 
- #3773

PR #3437, opened 2023-07-24
Support IntelliJ's formatting of binary align when multiline

kunli2:
Currently, we don't support this IntelliJ's code style setting in our `autoFormat` to support element alignment.

Screen Shot 2023-07-23 at 3 12 34 PM

This PR implements a visitor to support binary alignment (there are more we probably can make a more generic one) and we can add this into `AutoFormatVisitor` next (an according detector maybe required).

An example with binary alignment.

image


rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -620,6 +620,14 @@ final class Binary implements J, Expression, TypedTree {
...
         JLeftPadded<Type> operator;
 
+        public JLeftPadded<Type> getLeftPaddedType() {
+            return operator;
+        }
+
+        public Binary withOperator(JLeftPadded<Type> leftPadded) {
+            return getPadding().withOperator(leftPadded);
+        }
+
```
sambsnyd, line 630: These methods are redundant. `J.Binary.getPadding().getType()` and `J.Binary.getPadding.getOperator()` already exist and serve these purposes, they do not need to be duplicated.

rewrite-java/src/main/java/org/openrewrite/java/format/BinaryAlignWhenMultiline.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.format;
+
+import org.openrewrite.*;
+
+public class BinaryAlignWhenMultiline extends Recipe {
```
sambsnyd, line 20: There should not be a separate recipe for this. Its visitor should be part of `AutoFormat`, responding appropriately to the relevant `Style`

sambsnyd (changes requested):

rewrite-java/src/main/java/org/openrewrite/java/format/BinaryAlignWhenMultiline.java
kunli2: but we will need a style detector to detect this

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
kunli2: resolved

PR #3438, opened 2023-07-25
Added unpinned version test

joanvr:
## What's changed?
Added a new test to validate the behavior of ChangeDependency on unpinned dependencies.
Although we cannot check managed versions of Gradle right now, if the previous dependency is managed, we are going to assume that the new one will be managed too, and do not add a version tag. That was the previous behavior of the recipe before recent changes.
However, with the the new parameter `overrideManagedVersion` it will always add the tag version to the new dependency. 
This way we can force to add the newVersion, but we keep the default previous behavior. 

## What's your motivation?
This recipe is also used in rewrite-java-dependencies and rewrite-spring, and it was not clear the behavior of the individual recipe.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -15,8 +15,8 @@
  */
 package org.openrewrite.gradle;
 
-import lombok.EqualsAndHashCode;
-import lombok.Value;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.*;
 import org.openrewrite.*;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
 import org.openrewrite.gradle.marker.GradleProject;
@@ -46,6 +46,7 @@ import java.util.Optional;
 import static java.util.Objects.requireNonNull;
 
 @Value
+@AllArgsConstructor(onConstructor_ = {@JsonCreator})
 @EqualsAndHashCode(callSuper = true)
 public class ChangeDependency extends Recipe {
     @Option(displayName = "Old groupId",
@@ -89,6 +90,20 @@ public class ChangeDependency extends Recipe {
             required = false)
     @Nullable
     String versionPattern;
+
+    @Option(displayName = "Override managed version",
+            description = "If the old dependency has a managed version, this flag can be used to explicitly set the version on the new dependency. " +
+                    "WARNING: No check is done on the NEW dependency to verify if it is managed, it relies on whether the OLD dependency had a managed version. " +
+                    "The default for this flag is `false`.",
+            required = false)
+    @Nullable
+    Boolean overrideManagedVersion;
+
+
+    // Added constructor for compatibility with previous API
+    public ChangeDependency(String oldGroupId, String oldArtifactId, @Nullable String newGroupId, @Nullable String newArtifactId, @Nullable String newVersion, @Nullable String versionPattern) {
+        this(oldGroupId, oldArtifactId, newGroupId, newArtifactId, newVersion, versionPattern, null);
+    }
 
     @Override
     public String getDisplayName() {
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -90,6 +91,20 @@ public class ChangeDependency extends Recipe {
...
+            description = "If the old dependency has a managed version, this flag can be used to explicitly set the version on the new dependency. " +
+                    "WARNING: No check is done on the NEW dependency to verify if it is managed, it relies on whether the OLD dependency had a managed version. " +
+                    "The default for this flag is `false`.",
+            required = false)
+    @Nullable
+    Boolean overrideManagedVersion;
+
+
+    // Added constructor for compatibility with previous API
+    public ChangeDependency(String oldGroupId, String oldArtifactId, @Nullable String newGroupId, @Nullable String newArtifactId, @Nullable String newVersion, @Nullable String versionPattern) {
```
timtebeek, line 104: Maybe best to immediately deprecate this old constructor, and note the replacement in javadoc such that we don't have to maintain this for too long.

timtebeek (approved):

shanman190 (approved):

joanvr:
> @joanvr, there are actually 2 more checks on the version that need to be addressed.
> 
> With Gradle versioning we have to contend with 3 different rewrite types: `J.Literal`, `G.GString`, and `G.MapEntry`. Each of these should have this same check.

does it make sense to add it for GString? For what I see in the code it's expecting to be somewhat like "groupId:artifactId:${version}", so, we will always have a version there, right?

joanvr:
Fixed and allowing Map dependencies without a version (it was checking that version != null to apply the changes).

Fixed and added tests for GString dependencies, previously the GString was maintained and something like "newGroupId:newArtifactId:newVersion${variable}" was produced, keeping the variable at the end of the already full GAV coordinates.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -208,10 +235,9 @@ private J.MethodInvocation updateDependency(J.MethodInvocation m, ExecutionConte
                             }
                             if (original != updated) {
                                 String replacement = updated.toStringNotation();
-                                m = m.withArguments(ListUtils.mapFirst(depArgs, arg -> {
-                                    G.GString gString = (G.GString) arg;
-                                    return gString.withStrings(ListUtils.mapFirst(gString.getStrings(), l -> ((J.Literal) l).withValue(replacement).withValueSource(replacement)));
-                                }));
+                                J.Literal newLiteral = literal.withValue(replacement)
```
shanman190, line 238: We could employ the `ChangeStringLiteral` helper class here.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -176,10 +176,47 @@ implementation platform("org.apache.commons:commons-lang3:3.11")
...
+            """
+              plugins {
+                  id "java-library"
+              }
+              
+              repositories {
+                  mavenCentral()
+              }
+
+              def version = '2.6'
```
shanman190, line 207: Should we schedule an after visit to delete this if it's unused?

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
joanvr: I considered it, but ChangeStringLiteral was not working fine with GString, since it was taking the delimiters of the literal inside the GString, which is none

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
joanvr: I would prefer to close this PR as it is right now, and add extra features later in another one, because right now other repos builds are failing right now.

shanman190 (approved):

PR #3447, opened 2023-07-29
Ensure that system-installed artifacts can be resolved while constructing JavaParser classpath

haumacher:
When running Maven on Linux from a system maintainer-installed package and adding a dependency to a recipe that is also a dependency of Maven itself (e.g. `slf4j-api`), then the system-installed version is used (e.g. `file:/usr/share/java/slf4j-api.jar`). Unfortunately, this path is not matched be the regular expression, since no version is added to the file name. This results in an incomprehensible error "Unable to find runtime dependencies beginning with: slf4j-api" that gives no clue what's wrong. Therefore, its better to also output the current class path being scanned.

In the situation described above, the following template cannot be built, because the dependency `slf4j-api` cannot be resolved because the artifact cannot be found on the class path:
```
private static JavaTemplate createTemplate() {
    return JavaTemplate.builder("#{} static final Logger LOG = LoggerFactory.getLogger(#{any(java.lang.Class)});")
        .javaParser(JavaParser.fromJavaVersion().classpath(dependenciesFromClasspath("slf4j-api")))
        .imports("org.slf4j.LoggerFactory", "org.slf4j.Logger")
        .build();
}
```

## What's changed?
Bug in class path scanning code while constructing a JavaParser fixed. Allow to find the artifact `slf4j-api` from a class path entry like `file:/usr/share/java/slf4j-api.jar`. Made the version optional in the artifact pattern.

## What's your motivation?
Create a recipe that uses SLF4J classes (that are in the dependencies of Maven itself and therefore system-installed).

One can copy the classpath scanning code to the recipes code and fix it there, but makes no sense at all.

## Any additional context
Had a hard time debugging this - more context information in error messages is always a win.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
@@ -83,7 +83,7 @@ public interface JavaParser extends Parser {
         List<Path> artifacts = new ArrayList<>(artifactNames.length);
         List<String> missingArtifactNames = new ArrayList<>(artifactNames.length);
         for (String artifactName : artifactNames) {
-            Pattern jarPattern = Pattern.compile(artifactName + "-.*?\\.jar$");
+            Pattern jarPattern = Pattern.compile(artifactName + "(?:" + "-.*?" + ")?" + "\\.jar$");
             // In a multi-project IDE classpath, some classpath entries aren't jars
             Pattern explodedPattern = Pattern.compile("/" + artifactName + "/");
             boolean lacking = true;
@@ -104,7 +104,8 @@ public interface JavaParser extends Parser {
 
         if (!missingArtifactNames.isEmpty()) {
             throw new IllegalArgumentException("Unable to find runtime dependencies beginning with: " +
-                                               missingArtifactNames.stream().map(a -> "'" + a + "'").sorted().collect(joining(", ")));
+                                               missingArtifactNames.stream().map(a -> "'" + a + "'").sorted().collect(joining(", ")) +
+                                               ", classpath: " + runtimeClasspath);
         }
 
         return artifacts;


timtebeek:
Sorry to hear you had a hard time debugging this one! Curious to hear what kind of recipe you're developing; at first glance it seems similar to [our AddLogger recipe](https://github.com/openrewrite/rewrite-logging-frameworks/blob/ae017388d8ea15aa47399e19e2337e9fe8a18e3a/src/main/java/org/openrewrite/java/logging/AddLogger.java#L65-L74), but I can't see past the snippet you've provided above of course. There we don't use the `dependenciesFromClasspath` but rather `.javaParser(JavaParser.fromJavaVersion().classpath("slf4j-api"))`; perhaps that could be of some use to you still.

Agree with the changes you've made here; thank you for contributing!

timtebeek (approved):

haumacher:
> Curious to hear what kind of recipe you're developing; at first glance it seems similar to [our AddLogger recipe](https://github.com/openrewrite/rewrite-logging-frameworks/blob/ae017388d8ea15aa47399e19e2337e9fe8a18e3a/src/main/java/org/openrewrite/java/logging/AddLogger.java#L65-L74).

Thanks for the pointer - I'll try this one as part of my recipe, too. You are right, I tried to migrate an outdated custom logging abstraction layer to SLF4J. Besides adding the SLF logger, the recipe requires to transform logging statement of our own logging abstraction to SLF.

Thanks for accepting the fix!

timtebeek:
You're welcome; let us know if you have any further questions and be sure to checkout what other [logging recipes we already cover](https://docs.openrewrite.org/recipes/java/logging). You might be able to delegate to some of our existing recipes, or after an initial migration [adopt logging best practices](https://docs.openrewrite.org/recipes/java/logging/slf4j/slf4jbestpractices). We often see recipes implemented in such a composite fashion, to make best use of what's available already.

PR #3455, opened 2023-08-04
Fix CRLF newlines in AddImport

joanvr:
## What's changed?
AddImport does not respect new line styling when adding new imports. It is related to `ImportLayoutStyle::addImport` works, because it can override whatever prefix we add to the new import.

Maybe we should fix addImport and instead of always adding \n to consider CRLF based on a parameter? 

Also changed some ListUtils.map to ListUtils.mapFirst equivalent when only changing the first element, and withPrefix(getPrefix().withComments) for withComments directly in import. Probably those "helpers" weren't available at the time of building this recipe.

I thought about adding an extra parameter to addImport (or to ImportLayoutStyle builder?) so it adds \n or \r\n according to the provided preference. But this may require adding a "default" value with the same signature to not affect the public signature of the class (don't know if it's widely used in other recipes).

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
@@ -171,8 +171,7 @@ public class AddImport<P> extends JavaIsoVisitor<P> {
         if (generalFormatStyle.isUseCRLFNewLines()) {
             return ListUtils.map(newImports, rp -> rp.map(
                     i -> i.withPrefix(i.getPrefix().withWhitespace(i.getPrefix().getWhitespace()
-                            .replace("\n", "\r\n")
-                            .replaceAll("\r+", "\r")))
+                            .replaceAll("(?<!\r)\n", "\r\n")))
             ));
         }
         return newImports;
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
@@ -170,7 +170,9 @@ public class AddImport<P> extends JavaIsoVisitor<P> {
                 .orElse(autodetectGeneralFormatStyle(cu));
         if (generalFormatStyle.isUseCRLFNewLines()) {
             return ListUtils.map(newImports, rp -> rp.map(
-                    i -> i.withPrefix(i.getPrefix().withWhitespace(i.getPrefix().getWhitespace().replace("\n", "\r\n")))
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot for diving in! Looks good to me, although I lack a bit of background here. The adoption of `mapFirst` does help readability, even if it took me a while to spot in the diff.

rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
```diff
@@ -151,23 +147,37 @@ public AddImport(String type, @Nullable String statik, boolean onlyIfReferenced)
...
 
             j = cu;
         }
         return j;
     }
 
+    private List<JRightPadded<J.Import>> checkCRLF(JavaSourceFile cu, List<JRightPadded<J.Import>> newImports) {
+        GeneralFormatStyle generalFormatStyle = Optional.ofNullable(((SourceFile) cu).getStyle(GeneralFormatStyle.class))
+                .orElse(autodetectGeneralFormatStyle(cu));
+        if (generalFormatStyle.isUseCRLFNewLines()) {
```
timtebeek, line 171: I'm wondering if we should even call `autodetectGeneralFormatStyle` here, or just skip on detecting, as I'd assume a different part of the code base should already add this style to the `JavaSourceFile`.

rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
```diff
@@ -151,23 +147,37 @@ public AddImport(String type, @Nullable String statik, boolean onlyIfReferenced)
...
+            return ListUtils.map(newImports, rp -> rp.map(
+                    i -> i.withPrefix(i.getPrefix().withWhitespace(i.getPrefix().getWhitespace()
+                            .replace("\n", "\r\n")
+                            .replaceAll("\r+", "\r")))
```
shanman190, line 175: You could do this in one shot by using a negative look behind.

```suggestion
                            .replaceAll("(?<!\r)\n", "\r\n")))
```

joanvr: thanks! :)

rewrite-java/src/main/java/org/openrewrite/java/AddImport.java
joanvr: it's extracted from the auto format visitor (called by autoformat). In the tests at least, there was no markers there, and autodetectGeneralFormatStyle was used.

PR #3456, opened 2023-08-04
Add JavaTemplate.Builder Collection overloads

JLLeitschuh:
## What's changed?

Overload the following methods with `Collection` accepting variants:
 - `imports`
 - `staticImports`

## What's your motivation?

Make the API more friendly to work with when you need to add multiple imports.

The prior alternative was to use the following syntax, assuming a `imports` is a collection of strings: `builder.imports(imports.toArray(new String[0]);`

Not the most friendly user-experience.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java;
 
+import lombok.Getter;
 import lombok.Value;
 import lombok.experimental.NonFinal;
 import org.openrewrite.Cursor;
@@ -29,12 +30,15 @@ import org.openrewrite.template.SourceTemplate;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Consumer;
 
-@SuppressWarnings("unused")
+@SuppressWarnings({"unused", "LombokGetterMayBeUsed"})
 public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
+    @Getter
     private final String code;
     private final int parameterCount;
     private final Consumer<String> onAfterVariableSubstitution;
@@ -46,10 +50,6 @@ public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
         this.onAfterVariableSubstitution = onAfterVariableSubstitution;
         this.parameterCount = StringUtils.countOccurrences(code, "#{");
         this.templateParser = new JavaTemplateParser(contextSensitive, javaParser, onAfterVariableSubstitution, onBeforeParseTemplate, imports);
-    }
-
-    public String getCode() {
-        return code;
     }
 
     @Override
@@ -143,6 +143,10 @@ public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
         }
 
         public Builder imports(String... fullyQualifiedTypeNames) {
+            return imports(Arrays.asList(fullyQualifiedTypeNames));
+        }
+
+        public Builder imports(Collection<String> fullyQualifiedTypeNames) {
             for (String typeName : fullyQualifiedTypeNames) {
                 validateImport(typeName);
                 this.imports.add("import " + typeName + ";\n");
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -29,12 +30,15 @@
...
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Consumer;
 
-@SuppressWarnings("unused")
+@SuppressWarnings({"unused", "LombokGetterMayBeUsed"})
 public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
+    @Getter
```
jkschneider, line 41: Why?

jkschneider:
I'm trying to imagine when you would dynamically compute the import list... Can we see that use site before adding API surface area?

jkschneider:
There is a _very_ high bar to bloating API surface area, as you know.

JLLeitschuh:
https://github.com/openrewrite/rewrite-java-security/blob/147b8668f52c597457d5d17da77be30fa8499ea1/src/main/java/org/openrewrite/java/security/xml/XmlFactoryInsertPropertyStatementVisitor.java#L151

We're encountering this when we need to dynamically generate templates based upon variable conditions.

We're likely going to encounter this in future scenarios moving forward as well.

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
JLLeitschuh: IJ was suggesting this change to me. It was flagging the method as a warning. I don't care either way, I was just trying to clean up the warning.

JLLeitschuh: If you want to just add a suppression on the `getCode` method, I'm also more than happy to do that instead

JLLeitschuh:
We likely have 3-4 additional XML parsers that we will be looking to support, and need dynamic imports at those points as well. We may be able to, eventually, refactor the XML parser fix logic to use some generic base classes, but this seemed like a simple enough fix on the `rewrite` side so we didn't need to do this moving forward

jkschneider:
It should be harmless to add imports that aren't used. I don't see the necessity of filtering this list. You should be able to pass in a static array of potential imports.

image

JLLeitschuh:
> It should be harmless to add imports that aren't used.

The same set of imports is then also passed to `maybeAddImport` a few lines below. I agree with your assertion, but that's not true for `maybeAddImport`.

https://github.com/openrewrite/rewrite-java-security/blob/147b8668f52c597457d5d17da77be30fa8499ea1/src/main/java/org/openrewrite/java/security/xml/XmlFactoryInsertPropertyStatementVisitor.java#L156

jkschneider:
The same is true of `maybeAddImport`. It won't be added unless it is demonstrably used.

jkschneider:
This is why the bar is so high... I only see 15 lines of suggested API in my IDE. In the hundreds of use cases of JavaTemplate, this is the first to suggest dynamic imports. All I ask is that we provably repeat this pattern before jumping to add surface area on the presumption that there will be further cases.

image

JLLeitschuh:
> The same is true of maybeAddImport. It won't be added unless it is demonstrably used.

Sure... but isn't it more expensive as then `AddImport` has to do the inspection of the class to determine if the import is needed for cases we know for a fact won't be needed?

jkschneider:
We're talking about 4 imports max here. This is an optimization I wouldn't complicate your code for.

jkschneider:
All of those 4 checks use `TypesInUse` transient cache, so only the first one triggers a visit.

JLLeitschuh:
> We're talking about 4 imports max here. This is an optimization I wouldn't complicate your code for.

Fair. I also didn't realize this was an optimization, and not making the behaviour correct. I didn't know, before this conversation, that `maybeAddImports` was a no-op if the import wasn't needed. I thought `maybeAddImports` always added imports, even if they weren't needed. I hadn't seen the `onlyIfReferenced` flag before just now.

All of that being said, if I'm going to need to create a large list of imports, and use them multiple times, perhaps for multiple templates, and multiple recipes I need to import, I'm likely going to want to use an immutable `Collection` to hold that, maybe in a static variable.

I'm feeling confused by why this is contentious, as, in my experience with most APIs across both the JDK and most other libraries, any method accepting varargs almost always has a variant that also accepts `Collection`. In my experience, varargs are a convenience for API consumers, while `Collection` is almost always there as passing around arrays is generally frowned upon.

JLLeitschuh:
> All I ask is that we provably repeat this pattern before jumping to add surface area on the presumption that there will be further cases.

I was at two, and likely going on a third with the third XML parser, which is what got us here. But I do see what you're saying

jkschneider:
I don't want to be abrupt, but I don't see how further conversation here is productive. Let's not beat the dead horse.

jkschneider:
Appreciate the suggestion here, regardless.

PR #3459, opened 2023-08-04
Require parse-to-print idempotence

jkschneider:
## What's changed?

Parsers should prove that their printed representation matches the original source code to prevent patch apply failures on commit. The lack of parse-to-print idempotence can otherwise go unnoticed until a commit.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -19,24 +19,23 @@ import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.style.Style;
-import org.openrewrite.tree.ParseError;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
-import java.util.stream.Stream;
 
 public interface SourceFile extends Tree {
 
     /**
-     * Does this source file, when printed produce a byte-for-byte identical result to the input?
+     * Does this source file represented as an LST, when printed, produce a byte-for-byte identical
+     * result to the original input source file?
      *
      * @param input The input source.
      * @return <code>true</code> if the parse-to-print loop is idempotent, <code>false</code> otherwise.
      */
-    default boolean isPrintIdempotent(Parser.Input input, ExecutionContext ctx) {
+    default boolean printEqualsInput(Parser.Input input, ExecutionContext ctx) {
         return printAll().equals(StringUtils.readFully(input.getSource(ctx)));
     }
 
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -15,17 +15,31 @@
  */
 package org.openrewrite;
 
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.style.Style;
+import org.openrewrite.tree.ParseError;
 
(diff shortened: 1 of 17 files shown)


timtebeek (approved):

rewrite-core/src/main/java/org/openrewrite/SourceFile.java
```diff
@@ -26,6 +27,18 @@
...
+
+    /**
+     * Does this source file represented as an LST, when printed, produce a byte-for-byte identical
+     * result to the original input source file?
+     *
+     * @param input The input source.
+     * @return <code>true</code> if the parse-to-print loop is idempotent, <code>false</code> otherwise.
+     */
+    default boolean printEqualsInput(Parser.Input input, ExecutionContext ctx) {
+        return printAll().equals(StringUtils.readFully(input.getSource(ctx)));
```
timtebeek, line 39: Had thought to see if I could help out with the test failures in https://ge.openrewrite.org/s/crdmk2pb27noq/tests/overview?outcome=failed .
I'd locally expanded this to the following, such that I could set a conditional breakpoint on `!printed.equals(read)`.
```suggestion
        String printed = printAll();
        String read = StringUtils.readFully(input.getSource(ctx));
        return printed.equals(read);
```

timtebeek: That then fails in 
https://github.com/openrewrite/rewrite/blob/dc9f04ae6dd44ac6c981fa63c5ee165d85c8f3ed/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateInstanceOfTest.java#L208-L220
with an excess `}}returnnull;` in the printed result.
```
tim@tim-xps-15-9520:~/.config/JetBrains/IntelliJIdea2023.1/scratches$ cat read 
class __P__ {  static native <T> T p();  static native <T> T[] arrp();  static native boolean booleanp();  static native byte bytep();  static native char charp();  static native double doublep();  static native int intp();  static native long longp();  static native short shortp();  static native float floatp();}class __M__ {  static native Object any(Object o);  static native Object any(java.util.function.Predicate<Boolean> o);  static native <T> Object anyT();}class T{Object m(Object o){if(true) {if (((Object)null) instanceof String s) {} else {__M__.any(
/*__TEMPLATE__*/s.length()/*__TEMPLATE_STOP__*/
)}}
return null;
}}
```
```
tim@tim-xps-15-9520:~/.config/JetBrains/IntelliJIdea2023.1/scratches$ cat printed 
class __P__ {  static native <T> T p();  static native <T> T[] arrp();  static native boolean booleanp();  static native byte bytep();  static native char charp();  static native double doublep();  static native int intp();  static native long longp();  static native short shortp();  static native float floatp();}class __M__ {  static native Object any(Object o);  static native Object any(java.util.function.Predicate<Boolean> o);  static native <T> Object anyT();}class T{Object m(Object o){if(true) {if (((Object)null) instanceof String s) {} else {__M__.any(
/*__TEMPLATE__*/s.length()/*__TEMPLATE_STOP__*/
)}}
return null;
}}returnnull;
}}
```

timtebeek: Workaround for this preexisting issue in https://github.com/openrewrite/rewrite/pull/3459/commits/788040479ad0a28a628fab8bc405943078cea36f ; we can revisit at a more convenient time if needed.

rewrite-maven/src/main/java/org/openrewrite/maven/table/DependencyGraph.java
```diff
@@ -0,0 +1,49 @@
...
+        @Column(displayName = "Source set",
+                description = "The source set that contains the dependency.")
+        String sourceSet;
+
+        @Column(displayName = "From dependency",
+                description = "A dependency that depends on the 'to' dependency.")
+        String from;
+
+        @Column(displayName = "From dependency",
+                description = "A dependency that depends on the 'to' dependency.")
```
timtebeek, line 46:
```suggestion
        @Column(displayName = "To dependency",
                description = "A dependency depended on by the 'from' dependency.")
```

PR #3462, opened 2023-08-07
Add ChangeNamespace recipe to rewrite-xml

ammachado:
## What's changed?

Add `ChangeNamespace` recipe to `recipe-xml`. 

## What's your motivation?

This recipe allows arbitrary XML namespace changes

## Have you considered any alternatives or workarounds?

The current `ChangeTagAttribute` recipe requires previous knowledge of namespace prefixes.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ auto-formatter on affected files
- [x] I've updated the documentation (if applicable)

PR #3465, opened 2023-08-07
Added ChangeNamespaceValue recipe and tests

cjobinabo:
## What's changed?
Add ChangeNamespaceValue recipe to recipe-xml.

## What's your motivation?
This recipe was moved from the `rewrite-liberty` recipe on Tim request.

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
@@ -71,16 +71,18 @@ public class ChangeNamespaceValue extends Recipe {
             required = false)
     String versionMatcher;
 
+    @Nullable
     @Option(displayName = "Search All Namespaces",
-            description = "Specify whether evaluate all namespaces",
-            example = "true")
-    boolean searchAllNamespaces;
+            description = "Specify whether evaluate all namespaces. Defaults to true",
+            example = "true",
+            required = false)
+    Boolean searchAllNamespaces;
 
 
     @JsonCreator
     public ChangeNamespaceValue(@Nullable @JsonProperty("elementName") String elementName, @Nullable @JsonProperty("oldValue") String oldValue,
                                 @NonNull @JsonProperty("newValue") String newValue, @Nullable @JsonProperty("versionMatcher") String versionMatcher,
-                                @NonNull @JsonProperty("searchAllNamespaces") boolean searchAllNamespaces) {
+                                @Nullable @JsonProperty("searchAllNamespaces") Boolean searchAllNamespaces) {
         this.elementName = elementName;
         this.oldValue = oldValue;
         this.newValue = newValue;
@@ -90,7 +92,11 @@ public class ChangeNamespaceValue extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
+
         return new XmlIsoVisitor<ExecutionContext>() {
+
+            @Nullable
+            private final XPathMatcher elementNameMatcher = elementName != null ? new XPathMatcher(elementName) : null;
 
             @Override
             public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
@@ -104,14 +110,17 @@ public class ChangeNamespaceValue extends Recipe {
             }
 
             private boolean matchesElementName(Cursor cursor) {
-                return elementName == null || new XPathMatcher(elementName).matches(cursor);
+                return elementNameMatcher == null || elementNameMatcher.matches(cursor);
             }
 
             private boolean matchesVersion(Xml.Tag tag) {
                 if (versionMatcher != null) {
-                    return tag.getAttributes().stream()
-                            .filter(this::isVersionAttribute)
-                            .anyMatch(this::isVersionMatch);
+                    for (Xml.Attribute attribute : tag.getAttributes()) {
+                        if (isVersionAttribute(attribute) && isVersionMatch(attribute)) {
+                            return true;
+                        }
+                    }
+                    return false;
                 }
                 return true;
             }
@@ -129,8 +138,9 @@ public class ChangeNamespaceValue extends Recipe {
             }
 
             private boolean isXmlnsAttribute(Xml.Attribute attribute) {
-                return searchAllNamespaces && attribute.getKeyAsString().startsWith(XMLNS_PREFIX) ||
-                        !searchAllNamespaces && attribute.getKeyAsString().equals(XMLNS_PREFIX);
+                boolean searchAll = searchAllNamespaces == null || Boolean.TRUE.equals(searchAllNamespaces);
+                return searchAll && attribute.getKeyAsString().startsWith(XMLNS_PREFIX) ||
+                        !searchAll && attribute.getKeyAsString().equals(XMLNS_PREFIX);
             }
 
             private boolean isVersionAttribute(Xml.Attribute attribute) {
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


timtebeek:
Likely supersedes #3462

ammachado:
Can you add the unit tests from #3462? My version just changes the namespace and does not rely on any version attribute. In my scenarios (migrating Spring / Apache Camel XMLs), the version attribute is not present.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,134 @@
...
+                return false;
+            }
+
+            public Xml.Attribute visitChosenElementAttribute(Xml.Attribute attribute) {
+                if (this.version == null && !attribute.getKeyAsString().equals("version") ||
+                    this.version != null && !attribute.getKeyAsString().equals("xmlns")) {
+                    return attribute;
+                }
+
+                if (this.version != null && matchesVersion()) {
```
timtebeek, line 119:
```suggestion
                if (this.version != null && Semver.validate(this.version, versionMatcher).isValid()) {
```

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,134 @@
...
+                    } catch (NumberFormatException e) {
+                        return false;
+                    }
+                    if (!checkGreaterThan && dVersionExpected == dVersion || checkGreaterThan && dVersionExpected <= dVersion) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+
```
timtebeek, line 112:
```suggestion
```

timtebeek (commented):
I think we can delegate to Semver for the version comparison; What do you think about doing that?

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,134 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.NonNull;
+import org.openrewrite.xml.tree.Xml;
```
timtebeek, line 28:
```suggestion
import org.openrewrite.semver.Semver;
import org.openrewrite.xml.tree.Xml;
```

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,134 @@
...
+                                @NonNull @JsonProperty("newValue") String newValue,
+                                @NonNull @JsonProperty("versionMatcher") String versionMatcher) {
+        this.elementName = elementName;
+        this.newValue = newValue;
+        this.versionMatcher = versionMatcher;
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new XmlVisitor<ExecutionContext>() {
```
timtebeek, line 71: We can do away with the casts if you return the same type of tag.
```suggestion
        return new XmIsoVisitor<ExecutionContext>() {
```

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,134 @@
...
+
+            String version = null;
+
+            @Override
+            public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                Xml.Tag t = (Xml.Tag) super.visitTag(tag, ctx);
```
timtebeek, line 77:
```suggestion
            public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                Xml.Tag t = super.visitTag(tag, ctx);
```

timtebeek (commented):
Another quick few suggestions.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -41,94 +46,93 @@ public String getDescription() {
         return "Alters XML Attribute value within specified element of a specific resource versions.";
     }
 
+    @Nullable
     @Option(displayName = "Element name",
             description = "The name of the element whose attribute's value is to be changed. Interpreted as an XPath Expression.",
             example = "property")
```
timtebeek, line 52: `required` defaults to `true`, so if this is truly nullable then we likely need to add that flag to the annotation here and elsewhere.
```suggestion
            example = "property",
            required = false)
```

izozapzazp (approved):
kjkljkljkljkl

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,141 @@
...
+                Xml.Tag t = super.visitTag(tag, ctx);
+                this.version = null;
+                if (elementName == null || new XPathMatcher(elementName).matches(getCursor())) {
+                    // find versions
+                    t = t.withAttributes(ListUtils.map(t.getAttributes(), this::visitChosenElementAttribute));
+                }
+
+                if (this.version != null) {
+                    //change namespace
+                    t = t.withAttributes(ListUtils.map(t.getAttributes(), this::visitChosenElementAttribute));
```
timtebeek, line 109: What surprised me here is that we do two passes through the same method depending on whether a version is found and assigned to a field; that had me scratching my head a bit wondering if there's a different way. From the tests I can see it works, but for maintainability I'd like to briefly explore alternatives.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -93,47 +92,57 @@ public ChangeNamespaceValue(@Nullable @JsonProperty("elementName") String elemen
...
 
-            public Xml.Attribute visitChosenElementAttribute(Xml.Attribute attribute) {
-                boolean isXmlnsAttr = (searchAllNamespaces && attribute.getKeyAsString().startsWith(XMLNS_PREFIX) ||
-                        !searchAllNamespaces && attribute.getKeyAsString().equals(XMLNS_PREFIX));
-                boolean isVersionAttr = attribute.getKeyAsString().startsWith(VERSION_PREFIX);
-                if (oldValue != null && (!isXmlnsAttr || !attribute.getValueAsString().equals(oldValue))) {
-                    return attribute;
-                }
+            private boolean matchesElementName(Cursor cursor) {
+                return elementName == null || new XPathMatcher(elementName).matches(cursor);
```
timtebeek, line 107: Can we extract this `XPathMatcher` to a fields and reuse it? Now it's created for every tag we visit.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
joanvr: I've made some changes that make the code more explicit about what we are doing and add more semantics and readability.
This way we are still doing two passes, one to validate the version and another one to maybe change the namespace attribute if it matches the oldValue, but all the boolean expressions are in methods that explain what they are about.
If you don't like the stream API, you can change the matchesVersion with:
```
if (versionMatcher != null) {
    for (Xml.Attribute attribute : tag.getAttributes()) {
        if (isVersionAttribute(attribute) && isVersionMatch(attribute)) {
            return true;
        }
    }
    return false;
}
return true;
```

joanvr:
I set the default value of searchAllNamespaces to true, because it seems to be the most common scenario based on the tests, but I'm not sure if that's true, maybe the original author can make a more informed decision about this :)

joanvr (approved):

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeNamespaceValue.java
```diff
@@ -0,0 +1,160 @@
...
+            private boolean matchesVersion(Xml.Tag tag) {
+                if (versionMatcher != null) {
+                    for (Xml.Attribute attribute : tag.getAttributes()) {
+                        if (isVersionAttribute(attribute) && isVersionMatch(attribute)) {
+                            return true;
+                        }
+                    }
+                    return false;
+                }
+                return true;
```
timtebeek, line 125:
```suggestion
                if (versionMatcher == null) {
                    return true;
                }
                for (Xml.Attribute attribute : tag.getAttributes()) {
                    if (isVersionAttribute(attribute) && isVersionMatch(attribute)) {
                        return true;
                    }
                }
                return false;
```

timtebeek: Figured have the method read true -> true -> false, instead of true -> false -> true.

timtebeek:
Thanks again!

PR #3467, opened 2023-08-10
Change `J.Modifier` to be visitable by JavaVisitor

kunli2:
Visitors like Kotlin auto-format need to visit `J.Modifiers` and internal a list of annotations `List<Annotation> J.Modifiers.annotations`, and spaces as well.
Currently, it's not visitable in JavaVisitor, this PR updates it to be visitable.


sambsnyd:
As we discussed in standup I don't think we need this. There isn't much you can do with a modifier aside from adding it or removing it, both of which are usually more convenient to do when you're visiting the thing the modifier applies to, rather than the modifier itself.

PR #3468, opened 2023-08-10
Supporting yaml "collapsed" keys

joanvr:
## What's changed?
When checking if the incomingEntry matches an existingEntry, we need to break the incoming intro prefixes splitted by `.` to check if we have any match instead of just checking equality of the whole key string. Otherwise, we cannot detect if we need to "split" the full collapsed key and insert it as a child.

## What's your motivation?
We are using this recipe in ChangePropertyKey that is used in Spring Migrations, and it happens quite a lot that we have and incomingEntry that needs to be partially matched inside and existing tree of entries. Otherwise, it's being inserted in the root, and causing an invalid merged yaml due to duplicated key.

@timtebeek 

## Any additional context
Some scenario not covered yet is that the incomingEntry key partially matches an existing collapsed key partially, needing to split the existing key into a mapping and adding the two entries:
existing:
```
management:
    endpoints.jmx.other: /admin
```
incoming: `management.endpoints.jmx.unique-names: true`,

expected:
```
management:
    endpoints.jmx:
        other: /admin
        unique-names: true
```
actual:
```
management:
    endpoints.jmx.other: /admin
    endpoints.jmx.unique-names: true
```

I've created a disabled test for this one, but not fixed yet, since it does not seem to have any impact yet.

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
@@ -841,4 +841,26 @@ class MergeYamlTest implements RewriteTest {
         );
     }
 
+    @Test
+    void collapsedKey() {
+        rewriteRun(
+          spec -> spec.recipe(new MergeYaml(
+            "$",
+            //language=yaml
+            "management.endpoints.jmx.unique-names: true",
+            true,
+            null
+          )),
+          yaml(
+            """
+              management:
+                context-path: /admin
+              """, """
+              management:
+                context-path: /admin
+                endpoints.jmx.unique-names: true
+              """
+          )
+        );
+    }
 }
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
@@ -842,7 +842,7 @@ class MergeYamlTest implements RewriteTest {
     }
 
     @Test
-    void collapsedKey() {
+    void incomingCollapsedKey() {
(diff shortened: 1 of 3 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
```diff
@@ -841,4 +841,51 @@ void mergeSequenceMapChangeComplexMapping() {
...
+              """, """
+              management:
+                context-path: /admin
+                endpoints.jmx.unique-names: true
+              """
+          )
+        );
+    }
+
+    @Disabled("fix not implemented yet, but potentially an issue")
```
shanman190, line 867: This one is definitely an issue to resolve. SnakeYAML will expect for the subkey `endpoints.jmx` not to be duplicated within it's subkeys, but to instead for that subpath to be broken out as described by this test.

shanman190 (changes requested):
Looking good so far. I've left a comment on the disabled test that we'll definitely need to fix.

rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
knutwannheden: @joanvr In case you don't have time for this today, can you create an issue? Then I will look into it.

shanman190: I think we may also need another test case to assist with breaking an existing flatten property on the longest common subpath. 

```
management.context-path: /admin
```

Which would then produce the same output as the first test case contained in this PR.

PR #3469, opened 2023-08-10
Add data table to EffectiveMavenRepositories and its test

kmccarp:
## What's changed?
Add data table to the recipe EffectiveMavenRepositories.

## What's your motivation?
This recipe was created before data tables existed and benefits from providing structured data as such. Also, change markers are now optional.

No

@timtebeek 
@joanvr 

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/EffectiveMavenRepositories.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/EffectiveMavenRepositories.java
@@ -15,9 +15,13 @@
  */
 package org.openrewrite.maven.search;
 
+import lombok.EqualsAndHashCode;
+import lombok.Value;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.marker.SearchResult;
 import org.openrewrite.maven.MavenExecutionContextView;
 import org.openrewrite.maven.MavenIsoVisitor;
@@ -27,6 +31,8 @@ import org.openrewrite.xml.tree.Xml;
 
 import java.util.StringJoiner;
 
+@Value
+@EqualsAndHashCode(callSuper = false)
 public class EffectiveMavenRepositories extends Recipe {
 
     @Override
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/EffectiveMavenRepositories.java
```diff
@@ -41,6 +46,13 @@ public String getDescription() {
                "determined when the LST was produced.";
     }
 
+    @Option(displayName = "Use markers",
+            description = "Whether to add markers for each effective Maven repository to the POM. Default `false`.",
+            required = false)
+    Boolean useMarkers;
```
jkschneider, line 52: `@Nullable` if it isn't required

rewrite-maven/src/test/java/org/openrewrite/maven/search/EffectiveMavenRepositoriesTest.java
```diff
@@ -168,4 +168,45 @@ void fromMavenSettingsOnAst() {
...
+            """
+              <project>
+                  <groupId>org.openrewrite.example</groupId>
+                  <artifactId>my-app</artifactId>
+                  <version>1</version>
+                  <modules>
+                      <module>module</module>
+                  </modules>
+              </project>
+                              """,
```
jkschneider, line 197: formatting off here

rewrite-maven/src/main/java/org/openrewrite/maven/search/EffectiveMavenRepositories.java
```diff
@@ -41,6 +47,14 @@ public String getDescription() {
                "determined when the LST was produced.";
     }
 
+    @Option(displayName = "Use markers",
+            description = "Whether to add markers for each effective Maven repository to the POM. Default `false`.",
+            required = false)
+    @Nullable
+    Boolean useMarkers;
+
+    transient EffectiveMavenRepositoriesTable table = new EffectiveMavenRepositoriesTable(this);
```
joanvr, line 56: Maybe I would move this as a variable inside the getVisitor method, just to not have transient (because of recipe serde), or at least make it private so it's not exposed.

joanvr (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/search/EffectiveMavenRepositories.java
kmccarp: `transient` table seems to be the pattern we follow:

* https://github.com/openrewrite/rewrite/blob/f629bcc3c3f6b9cb6fa696e2a4bed99764c1ebcf/rewrite-core/src/main/java/org/openrewrite/FindCollidingSourceFiles.java#L33
* https://github.com/openrewrite/rewrite/blob/b5f0d7e46441cdd91e82b8600d112366544e4e38/rewrite-core/src/main/java/org/openrewrite/FindParseFailures.java#L47
* https://github.com/openrewrite/rewrite/blob/361b7ba095abaa0381da124049de20355a3a398a/rewrite-core/src/main/java/org/openrewrite/search/FindParseToPrintInequality.java#L32

Issue #3426, opened 2023-07-19
ChangePropertyValue automatically coalesces properties


wabrit:
Given a source YAML that looks like this:

```yaml
quarkus:
  opentelemetry:
    enabled: true
```
and a recipe `ChangePropertyKey("quarkus.opentelemetry.enabled", "quarkus.otel.enabled", ..)`

the result is

```yaml
quarkus:
  otel.enabled: true
```

This automatic coalescing of property names confuses Quarkus, which treats the new property as `quarkus."otel.enabled"` and doesn't match it to `quarkus.otel.enabled`.

The correct output in this case would be:

```yaml
quarkus:
  otel:
    enabled: true
```

wabrit:
The following simple failing test should demonstrate the issue:

```java
class ChangePropertyKeyTest implements RewriteTest {

    @Test
    void simpleTest() {
        rewriteRun(
            spec -> spec.recipe(
                new ChangePropertyKey("quarkus.opentelemetry.enabled", "quarkus.otel.enabled",
                        null, null, "application.yml")
            ),
            yaml(
                """
                     quarkus:
                       foo:
                         bar: 1
                       opentelemetry:
                         enabled: true
                         
                    """,
                """
                    quarkus:
                       foo:
                         bar: 1
                       otel:
                         enabled: true
                         
                   """,
                spec -> spec.path("application.yml")
            )
        );
    }
```

timtebeek:
Hmm; what version of OpenRewrite are you using? Since that `"application.yml"` causes a compiler error on the main branch.

timtebeek:
Either way confirmed to be an issue also without that argument ; not clear yet what the fix should be unfortunately, so this goes onto the backlog.

wabrit:
> Hmm; what version of OpenRewrite are you using? Since that `"application.yml"` causes a compiler error on the main branch.

This is not latest major release; will check tomorrow when i have access to the code and record it here.

timtebeek:
Previously seen on https://github.com/openrewrite/rewrite/issues/1841

timtebeek:
Posting the suggested replacement specifically for Quarkus here:
https://docs.openrewrite.org/recipes/quarkus/changequarkuspropertykey

You'll find this recipe is used as part of the following composite recipes already:

[...]

Issue #1873, opened 2022-06-01
Yaml - ChangePropertyKey does not adjust child prefixes if new key is different length


nmck257:
See test cases below:

```kotlin
@Test
fun `shorter new key with indented config`() = assertChanged(
    recipe = ChangePropertyKey("a.b.c.d.e", "x.y", null, null),
    before = """
    a:
      b:
        c:
          d:
            e:
              child: true
    """,
    after = """
    x.y:
      child: true
    """
    /* actual:
    x.y:
              child: true
    */
)

@Test
fun `longer new key with indented config`() = assertChanged(
    recipe = ChangePropertyKey("x.y", "a.b.c.d.e",  null, null),
    before = """
    x:
      y:
        child: true
    """,
    after = """
    a.b.c.d.e:
      child: true
    """
    /* actual:
    a.b.c.d.e:
        child: true
    */
)
```

traceyyoshima:
Fixed by ffea5120b95d136a32b9b39b63655721370d2f88

Issue #1841, opened 2022-05-24
Yaml - coalesce less


nmck257:
As discussed w/ @jkschneider on slack, consider a test case like this:
```kotlin
@Test
fun `does not reformat properties outside the affected key`() = rewriteRun({ spec ->
        spec.recipe(ChangePropertyKey("a.b.c", "x.y.z", true, null))
    },
    yaml(
        """
        a:
          b:
            c: abc
        something:
          else: qwe
        """, """
        x:
          y:
            z: abc
        something:
          else: qwe
        """
    )
)
```
This case fails, as the actual value is this:
```yml
# file leads with an empty line
something.else: qwe
x.y.z: abc
```
Few observations, ordered subjectively from most-to-least severe:
* The key `something.else` gets coalesced, though it was unrelated to the change
* `x.y.z` gets coalesced, though the key it replaced was styled as non-coalesced
* `x.y.z` gets relocated in the file
* The resultant file has an extraneous newline at the beginning

These tend to conflict w/ this design intent: "OpenRewrite recipes make minimally invasive changes to your source code that honor the original formatting" (from the main page of the docs site).

This issue targets that first observation, and we'd like to try simply removing nested usages of `CoalesceProperties` from other recipes. This should get us closer to a target state, with low risk of functional changes.

This could include:
* `ChangePropertyKey`
* `DeleteProperty`
* `MergeBootstrapYamlWithApplicationYaml`
* `MigrateDatabaseCredentials`

Future enhancements could include defining Styles for Yaml, to better-protect indent-vs-dot preferences.

nmck257:
Closing; this issue was broad, and progress was indeed made, and at this point the issue has been quiet for a few years

Issue #1249, opened 2021-11-29
ChangePropertyKey overwrites an existing key


ruben-garciapariente:
Hello,

I am using the ChangePropertyKey recipe and I don't know if there is a bug or I am using it incorrectly.

My source yaml is
```yaml
root:
  stuff0:
    field0: value0

other:
  field0: value0
  field1: value1
```

My recipe is

```yaml
  - org.openrewrite.yaml.ChangePropertyKey:
      oldPropertyKey: other
      newPropertyKey: root.stuff1
```

I expect something like
```yaml
root:
  stuff0.field0: value0
  stuff1:
    field0: value0
    field1: value1
```

but the real result is 
```yaml
root.stuff0.field0: value0

root.stuff1:
  field0: value0
  field1: value1
```

Is the behavior of ChangePropertyKey correct? 
If it is, is there any way to get the result I expect?

We have a complete example in https://github.com/ruben-garciapariente/test-openwrite/tree/ChangePropertyKey-yaml

Thank you very much

pway99:
Hello @ruben-garciapariente,
Thanks for identifying this issue `ChangeKey` should not move entries to non-scalar nodes.
In the example you provided, the action would be to move the contents of `other` to `root.stuff1`, which is more of a merge/move operation than changing the key.

PR #3470, opened 2023-08-10
no coalesced keys + key separator

joanvr:
## What's changed?
Do not coalesce keys by default. The current behavior was to always coalesce the keys into a single key separated by dots when possible, but it's not actually true that for yaml they are equivalent. So instead of:
```
a.b.c: true
```
we will have:
```
a:
  b:
    c: true
```
For scenarios when keys with dots are required, added an extra configurable parameter to change the key separator. So we can support having
```
a:
  b.c: true
```
if that's the intended behavior.

## What's your motivation?
This recipe has a lot of tests and involved in a lot of issues (some of them still open) regarding coalescing keys (or not coalescing them), and this has implications when later on we also use the MergeYaml visitor to merge results, since MergeYaml does not consider to be the same
```
a: 
  b: true
```
than
```
a.b: true
```
because in yaml specification, it is not... Just in some scenarios and frameworks that uses yaml for configuration/properties files, this happens to be the same.

I'm worried that changing the recipe expectations/api, this could cause unwanted behavior in people who is using this recipe, although, based on the tests and open issues, the new behavior is the expected/wanted by most users. However, we are only using this recipe in Spring [here](https://github.com/openrewrite/rewrite-spring/blob/f2e9faffdabec06550c1eeacc4da185dde30f1c4/src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java#L67) and for people that is using this recipe in a pinned version, it might not even be a problem

@timtebeek @shanman190 

## Any additional context
I think that this requires more discussion/agreement, maybe we can use this PR to come to an agreement.
Also, "fixed" all the existing tests to pass, but I have the feeling that a lot of them were added due to "issues" in coalesced keys that don't have sense anymore, that could be remove because they are redundant...

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
@@ -807,6 +807,24 @@ class ChangePropertyKeyTest implements RewriteTest {
               """
           )
         );
-
     }
+
+    @Test
+    void rootSingleKeyPrefixNewline() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangePropertyKey("spring.profiles", "spring.config.activate.on-profile", null, null)),
+          yaml(
+            """
+              spring.profiles: prod
+              """,
+            """
+              spring:
+                config:
+                  activate:
+                    on-profile: prod
+              """
+          )
+        );
+    }
+
 }
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
@@ -33,11 +33,11 @@ class ChangePropertyKeyTest implements RewriteTest {
     @Override
     public void defaults(RecipeSpec spec) {
         spec.recipe(new ChangePropertyKey(
-          "management.metrics#binders#*.enabled",
-          "management.metrics#enable.process.files",
+          "management.metrics:binders:*.enabled",
+          "management.metrics:enable.process.files",
           null,
           null,
-          "#"
+          ":"
         ));
     }
 
@@ -45,7 +45,7 @@ class ChangePropertyKeyTest implements RewriteTest {
     @Test
     void shorterNewKeyWithIndentedConfig() {
         rewriteRun(
-          spec -> spec.recipe(new ChangePropertyKey("a#b#c#d#e", "x.y", null, null, "#")),
+          spec -> spec.recipe(new ChangePropertyKey("a:b:c:d:e", "x.y", null, null, ":")),
           yaml(
             """
               a:
@@ -67,7 +67,7 @@ class ChangePropertyKeyTest implements RewriteTest {
     @Test
     void longerNewKeyWithIndentedConfig() {
         rewriteRun(
-          spec -> spec.recipe(new ChangePropertyKey("x#y", "a.b.c.d.e", null, null, "#")),
+          spec -> spec.recipe(new ChangePropertyKey("x:y", "a.b.c.d.e", null, null, ":")),
           yaml(
             """
               x:
@@ -103,11 +103,11 @@ class ChangePropertyKeyTest implements RewriteTest {
     void nestedEntry() {
         rewriteRun(
           spec -> spec.recipe(new ChangePropertyKey(
-            "management.metrics#binders#files.enabled",
-            "management.metrics#enable.process.files",
+            "management.metrics:binders:files.enabled",
+            "management.metrics:enable.process.files",
             null,
             null,
-            "#"
+            ":"
           )),
           yaml(
             """
@@ -184,7 +184,7 @@ class ChangePropertyKeyTest implements RewriteTest {
         @Issue("https://github.com/openrewrite/rewrite/issues/1114")
         void changePathToOnePathLonger() {
             rewriteRun(
-              spec -> spec.recipe(new ChangePropertyKey("a.b.c", "a.b.c.d", null, null, "#")),
+              spec -> spec.recipe(new ChangePropertyKey("a.b.c", "a.b.c.d", null, null, ":")),
               yaml(
                 "a.b.c: true",
                 "a.b.c.d: true"
@@ -196,7 +196,7 @@ class ChangePropertyKeyTest implements RewriteTest {
         @Test
         void changePathToOnePathShorter() {
             rewriteRun(
-              spec -> spec.recipe(new ChangePropertyKey("a.b.c.d", "a.b.c", null, null, "#")),
+              spec -> spec.recipe(new ChangePropertyKey("a.b.c.d", "a.b.c", null, null, ":")),
               yaml(
                 "a.b.c.d: true",
                 "a.b.c: true"
@@ -245,7 +245,7 @@ class ChangePropertyKeyTest implements RewriteTest {
             "acme.my-project.person.changed-first-name-key",
             false,
             null,
-            "#"
+            ":"
           )),
           yaml(
             """
@@ -271,7 +271,7 @@ class ChangePropertyKeyTest implements RewriteTest {
             "a.b.c",
             false,
             null,
-            "#"
+            ":"
           )),
           yaml(
             """
@@ -305,7 +305,7 @@ class ChangePropertyKeyTest implements RewriteTest {
             "new-property.sub-property.super-sub",
             true,
             null,
-            "#"
+            ":"
           )),
           yaml(
             """
@@ -382,7 +382,7 @@ class ChangePropertyKeyTest implements RewriteTest {
     @Test
     void updatePropertyWithMapping() {
         rewriteRun(
-          spec -> spec.recipe(new ChangePropertyKey("app#foo.change.from", "app#bar.change.to", null, null, "#")),
+          spec -> spec.recipe(new ChangePropertyKey("app:foo.change.from", "app:bar.change.to", null, null, ":")),
           yaml(
             """
               app:
@@ -432,7 +432,7 @@ class ChangePropertyKeyTest implements RewriteTest {
     @Issue("https://github.com/openrewrite/rewrite/issues/1841")
     void relocatesPropertyIfNothingElseInFamily() {
         rewriteRun(
-          spec -> spec.recipe(new ChangePropertyKey("a#b#c", "x.y.z", true, null, "#")),
+          spec -> spec.recipe(new ChangePropertyKey("a:b:c", "x.y.z", true, null, ":")),
           yaml(
             """
               a:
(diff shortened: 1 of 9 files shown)


shanman190:
I think overall the change looks fine. One alternative to note is the possible use of the jsonpath style. With the Json path style properties, they opt to quote properties with dots in them that should not be separated (ex: `a.b."c.d".e.f`). Either the current implementation or a jsonpath style seems like they would get the job done accurately from my perspective.

joanvr:
> I think overall the change looks fine. One alternative to note is the possible use of the jsonpath style. With the Json path style properties, they opt to quote properties with dots in them that should not be separated (ex: `a.b."c.d".e.f`). Either the current implementation or a jsonpath style seems like they would get the job done accurately from my perspective.

I also thought of this approach, to use quotes or ticks to specify keys that cannot be separated, the problem is that (I believe) even all those characters are valid characters for a key according to the spec, although it's probably super uncommon...

joanvr:
Some related issues that would fix with this:
https://github.com/openrewrite/rewrite/issues/3426
https://github.com/openrewrite/rewrite/issues/1249
https://github.com/openrewrite/rewrite/issues/1841
https://github.com/openrewrite/rewrite/issues/1873
https://github.com/openrewrite/rewrite/issues/1114
https://github.com/openrewrite/rewrite-spring/issues/189

rewrite-yaml/src/test/java/org/openrewrite/yaml/ChangePropertyKeyTest.java
```diff
@@ -468,13 +471,16 @@ void relocatesPropertyWithSamePrefix() {
...
+                        interval: 1
               """
           )
         );
     }
 
     @Nested
+    @Disabled
     @Issue("https://github.com/openrewrite/rewrite-spring/issues/189")
     class WhenOldPropertyKeyIsPrefixOfDotSeparatedKeyTest implements RewriteTest {
```
timtebeek, line 485: What should we do about these test cases? Looked to be added for https://github.com/openrewrite/rewrite-spring/issues/189 by @nmck257 
Only when I add an explicit non null, non default separate does the first one pass again.

Does that mean there's a risk we break things for people that were using the recipe as it was before a merge?

nmck257: Thanks for the tag -- hadn't seen this PR.
Conceptually, I'm always going to be in favor of these yml utilities better-preserving existing style in the source code when it comes to dot separation vs indents. Will review these changes more specifically soon

nmck257: I believe `WhenOldPropertyKeyIsPrefixOfDotSeparatedKeyTest` is mostly here to contrast with the `ExceptTest` below it. So in that sense, the exact output of these particular cases is not super-important, because `rewrite-spring` is now doing what's shown in `ExceptTest` for this specific property migration, not what's shown in `WhenOldPropertyKeyIsPrefixOfDotSeparatedKeyTest`.

nmck257:
have we tested a snapshot of this with `rewrite-spring`, and `ChangeSpringPropertyKeyTest`?

It looks like the strategy for the tests in this PR has been to adjust the recipe arguments to recreate the exact existing `before/after`. That feels like the wrong choice.
We know in the short term that existing usages of this recipe will bring their existing arguments, and we expect in the long term that most usages of this recipe would still choose to use the default value for this new argument. Explicitly overriding this arg is the least common pattern, but that's now the majority of what our test cases cover.

Instead, I'd recommend:
* revert the test cases to use their old/existing recipe args (aside from some new test cases to prove the new arg)
* adjust the `before/after` of each test case accordingly
* in this PR, have a healthy review/discussion of whether the new `before/after` scenarios are acceptable for existing usage (mainly `ChangeSpringPropertyKey`)

My local clone/IntelliJ is misbehaving today when running failing rewrite tests (`java.lang.SecurityException: class "org.eclipse.jgit.util.FS"'s signer information does not match signer information of other classes in the same package` from the classloader at `org.openrewrite.internal.InMemoryDiffEntry.<init>(InMemoryDiffEntry.java:62)`- something amiss between rewrite's shaded jgit and regular jgit?...), but when I try these (new) test cases:

```java
        @Test
        void foo() {
            rewriteRun(
              spec -> spec.recipe(new ChangePropertyKey("spring.profiles", "spring.config.activate.on-profile", null, null)),
              yaml(
                """
                  spring.profiles: prod
                  """,
                """
                  spring:
                    config:
                      activate:
                        on-profile: prod
                  """
              )
            );
        }

        @Test
        void bar() {
            rewriteRun(
              spec -> spec.recipe(new ChangePropertyKey("spring.profiles", "spring.config.activate.on-profile", null, null)),
              yaml(
                """
                  spring:
                    profiles: prod
                  """,
                """
                  spring:
                    config:
                      activate:
                        on-profile: prod
                  """
              )
            );
        }
```

...the former case fails, having produced:
`spring:  config:    activate:      on-profile: prod` (no newlines).

`rewrite-spring` uses `ChangeSpringPropertyKey` to make a *lot* of changes just like that one, so that would be a considerable regression.

joanvr:
Hi @nmck257, thank you very much for the feedback!
The problem that I see with this recipe, is that it relies on an invalid assumption, that `a.b: c` equals `a: b: c` (with new lines). That's no true for yaml, and when it relies on MergeYaml to do the merges, it usually fails to properly merge the keys, and adding a new "coalesced" key at the end. For most tests that just checks the moving of one single key, this might not be an issue, but when it comes to real cases, we have "new coalesced keys" that do not merge with previously existing properties....
Most of the tests, uses keys with dots (which I do not think are the same as coalesced keys). Because having a key called `files.enabled`, `enable.process.file` or `management.metrics` makes sense semantically, the dot is just as an "space", and does not make sense to always split them. Probably it would be better to use slashes (`-`) instead of dots in a yaml ecosystem, but thats completely out of our scope. Here comes the issue, the usage of the dot (`.`) both as a character in a key and as a key separator in the path to reference a key in the parameters. How we can support outputting some key with dots, when we try to split by dots to merge the keys?

Previously, the behavior was to always "coalesce" the new key, so it will maintain keys with dots joined, but also creates a new "coalesced" key that does not match and cannot be merged properly with existing keys... And as said, for single scenarios checked in a test it seems to work, but fails in a more general context.

I agree that changing all the tests parameters to make them pass is probably not the best idea, because probably we are even removing the original idea around this test, so it might be not needed anymore or we are missing some corner case... But how we can "fix" those tests were there are keys with dots (the default of these tests) without adding this extra parameter? Since this only affects output, because we can sort out the "input" part, maybe the idea of @shanman190 of Json path style with quotes is a good idea, but I'm worried because then we are removing the possibility of having those quotes as a characters in the key (which are valid yaml keys, but probably not used that much?).

And finally, some of the yaml tests (ans usages) of this recipe on rewrite-spring would. need to be changed too... Right now the first test even changes the key `servlet-path` for `servlet.path`, which without any change will result in an "split"... But this will happen no matter the solution we use... on how to keep keys with dots.

joanvr:
> My local clone/IntelliJ is misbehaving today when running failing rewrite tests (`java.lang.SecurityException: class "org.eclipse.jgit.util.FS"'s signer information does not match signer information of other classes in the same package` from the classloader at `org.openrewrite.internal.InMemoryDiffEntry.<init>(InMemoryDiffEntry.java:62)`- something amiss between rewrite's shaded jgit and regular jgit?...), but when I try these (new) test cases:
> 
> ```java
>         @Test
>         void foo() {
>             rewriteRun(
>               spec -> spec.recipe(new ChangePropertyKey("spring.profiles", "spring.config.activate.on-profile", null, null)),
>               yaml(
>                 """
>                   spring.profiles: prod
>                   """,
>                 """
>                   spring:
>                     config:
>                       activate:
>                         on-profile: prod
>                   """
>               )
>             );
>         }
> 
>         @Test
>         void bar() {
>             rewriteRun(
>               spec -> spec.recipe(new ChangePropertyKey("spring.profiles", "spring.config.activate.on-profile", null, null)),
>               yaml(
>                 """
>                   spring:
>                     profiles: prod
>                   """,
>                 """
>                   spring:
>                     config:
>                       activate:
>                         on-profile: prod
>                   """
>               )
>             );
>         }
> ```
> 
> ...the former case fails, having produced: `spring: config: activate: on-profile: prod` (no newlines).
> 
> `rewrite-spring` uses `ChangeSpringPropertyKey` to make a _lot_ of changes just like that one, so that would be a considerable regression.

Fixed this issue with the newlines when the matching key is on the root! Thanks for the feedback! :)

nmck257:
Hey @joanvr,
I think we're approaching this recipe with different design goals. Re:
> The problem that I see with this recipe, is that it relies on an invalid assumption, that a.b: c equals a: b: c (with new lines). That's no true for yaml,

Agreed that this isn't true for general Yaml. But this recipe isn't targeting general Yaml; that's what [`ChangeKey`](https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangeKey.java) does. This recipe is specifically targeting the Spring Boot config scenario, as described in the comment at the top (this is also implied by the presence of the `relaxedBinding` option, which is explicitly a Spring concern):

https://github.com/openrewrite/rewrite/blob/eda18b5582148608b5dd0d91933ff1519747f260/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyKey.java#L40-L47

With that assumption in place, dot-separating vs nesting are both correct and equivalent. Currently, the recipe prefers dot-separating the output, but I think that was an arbitrary choice. If we see benefit in preferring nesting, then that's a fine change.

And it sounds like (based on your comments) that we do see benefit in preferring nesting, for example, in the case where we run `ChangePropertyKey` / `MergeYaml` to produce a bunch similar/overlapping keys.

In that example, if I understand correctly, a preference for dot-separation might produce something like this:
```yml
a:
  b.c.d: qwe
  b.c.e: qwe
  q.z: qwe
  b.g: qwe
```
...whereas a preference for nesting might produce something like this:
```yml
a:
  b:
    c:
      d: qwe
      e: qwe
    g: qwe
  q:
    z: qwe
```
...and we expect that the latter is more-desirable for most users. That's reasonable to me. But my key point is that in the scope of this recipe, both of those outputs are logically correct; just stylistically different.

shanman190:
@joanr, I was thinking about this some more and the quote version in source that you're mentioning, could appear in two different shapes for a single-level property key (unsplit): `a."b.c".d` (omitting the yaml spec's optional key quotes) or `"a.\"b.c\".d"` (using the yaml spec's optional key quotes). Given this, I don't think it would come off confusing to a user to have to escape the inner quotes for keys which appeared in the raw source themselves as they would have to be escaped had the optional, full-key quotes been present.

To me, this seems to match more closely with the heart of the yaml spec as any character can appear in the yaml key, but if it contains a YAML special character them it must be quoted and potentially escaped. Even for the current test cases using `:` as the delimiter, those could appear in raw source keys without any issue if the original source key was quoted (ie. `"a:b"` is a valid yaml key).

shanman190:
I think where the problem comes in is that frameworks like Spring have chosen to interpret `"a.b.c.d": value`, `a: b: c: d: value` (using newlines), and all patterns in between as having equality to each other. This attribute is kinda counter to the yaml spec itself where each of those is truly different and could all appear even in the same document without a collision.

With this said, it makes me wonder if the `rewrite-spring`, etc frameworks really need a specialized yaml visitor that turns it's head to the yaml spec just like their internal configuration parsers do and treats every key path, including those in source, as being their fully separated with new lines equivalent. This would probably simplify some of the issues that we end up seeing here in terms of yaml merging not behaving as a user would expect.

PR #3475, opened 2023-08-14
Make `JavaTemplate` engine extensible

knutwannheden:
To allow the `JavaTemplate` mechanism to be used for other languages like Kotlin, the `JavaTemplateParser` must accept a generic `Parser.Builder` (i.e. not require `JavaParser.Builder`). This parser is expected to be preconfigured with a classpath entry containing the internal `__M__` and `__P__` types used by the template engine, as these are required for the parameter substitution.

Some of the classes and members have been declared as `protected`, but this may change again in the future.

Issue: openrewrite/rewrite-kotlin#218


knutwannheden:
Converted to draft as Kotlin support most likely will require a dedicated `KotlinTemplate` type.

rewrite-java/src/main/java/org/openrewrite/java/Internals.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import java.nio.file.Path;
+import java.util.Collection;
+
+public class Internals {
+    public static Collection<Path> getClasspath(JvmParser.Builder<?, ?> parserBuilder) {
+        return parserBuilder.classpath;
```
knutwannheden, line 23: I realize we could get rid of this if we move the code which updates the classpath from `JavaTemplateParser` to `JavaTemplate`, as this class is in the same package as `JvmParser`.

jkschneider:
i worry about the impact on classloading of `JvmParser`. for now could we avoid the generalization and just create an identical `classpath` method on `KotlinParser.Builder`?

rewrite-java/src/main/java/org/openrewrite/java/Internals.java
jkschneider: There is already an `internal` package in `rewrite-java` for things like this: 
image

knutwannheden: Yes, but JavaParser is not in that package. But like I pointed out, we can solve this without any additional accessor, since JavaTemplate is in the same package as JavaParser.

knutwannheden:
> i worry about the impact on classloading of `JvmParser`. for now could we avoid the generalization and just create an identical `classpath` method on `KotlinParser.Builder`?

Yes, now that we have a KotlinTemplate class that should be possible. I will work that out later today.

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -168,8 +200,8 @@ private void validateImport(String typeName) {
             }
         }
 
-        public Builder javaParser(JavaParser.Builder<?, ?> javaParser) {
-            this.javaParser = javaParser;
+        public Builder javaParser(JavaParser.Builder<?, ?> parser) {
+            this.parser = parser.clone();
```
sambsnyd, line 204: Why is this `clone()` now when it wasn't before?

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -27,25 +27,53 @@
...
 import java.util.function.Consumer;
 
 @SuppressWarnings("unused")
 public class JavaTemplate implements SourceTemplate<J, JavaCoordinates> {
+    protected static final Path TEMPLATE_CLASSPATH_DIR;
+
+    static {
+        try {
+            TEMPLATE_CLASSPATH_DIR = Files.createTempDirectory("java-template");
+            Path templateDir = Files.createDirectories(TEMPLATE_CLASSPATH_DIR.resolve("org/openrewrite/java/internal/template"));
```
sambsnyd, line 46: File IO in a static initializer feels unfortunate. But I do like using precomplied `__M__` and `__P__` to get around language differences. Smart

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
knutwannheden: I want to call clone before adding the classpath entry for the `__M__` and `__P__` classes. But it could also be done slightly later. Also, we call clone() later anyway before invoking the parser.

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
knutwannheden: I could do a lazy initialization as well at runtime.

traceyyoshima (approved):

rewrite-java/src/main/java/org/openrewrite/java/internal/template/__P__.java
```diff
@@ -0,0 +1,38 @@
...
+    public static native double doublep();
+
+    public static native int intp();
+
+    public static native long longp();
+
+    public static native short shortp();
+
+    public static native float floatp();
+}
```
JLLeitschuh, line 38: EOL @ EOF

rewrite-java/src/main/java/org/openrewrite/java/internal/template/__M__.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package org.openrewrite.java.internal.template;
+
+public class __M__ {
+    public static native Object any(Object o);
+
+    public static native Object any(java.util.function.Predicate<Boolean> o);
+
+    public static native <T> Object anyT();
+}
```
knutwannheden, line 24:
```suggestion
}

```

rewrite-java/src/main/java/org/openrewrite/java/internal/template/__P__.java
knutwannheden, line 38:
```suggestion
}

```

PR #3477, opened 2023-08-15
Change AutoFormatter to be dynamically provided by the tree root SourceFile

kunli2:
To support language-specific autoFormat like Kotlin, we need to change AutoFormatter to be dynamically provided using the same way as the language-specific printer.


knutwannheden:
Instead of adding another method to `Tree` we could consider using the `service()` method. What do you think?

kunli2:
yes, let's use `service()` method to support this.

PR #3478, opened 2023-08-15
refactor language specify service and add support to autoFormatter

kunli2:
1. Refactor and simplify language specify service, and get rid of potential class loading conflict errors.
2. Add support to `autoFormatter`

The according PR for Kotlin is (https://github.com/openrewrite/rewrite-kotlin/pull/224)


rewrite-java/src/main/java/org/openrewrite/java/tree/Service.java
```diff
@@ -13,17 +13,16 @@
...
-import org.openrewrite.Incubating;
+import org.openrewrite.Tree;
 import org.openrewrite.internal.lang.Nullable;
-import org.openrewrite.java.AddImport;
 import org.openrewrite.java.JavaVisitor;
 
-@Incubating(since = "8.2.0")
-public class ImportService {
+// Language specific service provider
+public interface Service {
```
knutwannheden, line 23: I see how you refactored this to circumvent the class loading issue we had on the service lookup, but I think we should revert this change and change it to do the service lookup by class name instead. Later on we can still discuss if the lookup can be done using referential equality on the class.

The reason I think we should revert this is that I think we may want to use this approach for other grammar islands as well (if it proves to be useful) and then we would probably end up moving the `service()` method up into `SourceFile` (and probably also `Tree`). Then we could no longer have a single `Service` interface, as the services will be specific to the grammar island. Also, the idea was that a language like Kotlin should be able to add its own services or also allow lookup using its own service extension classes to provide access to specific methods.  `KotlinImportService` will at some point probably provide a method to create a visitor to add rename imports and Kotlin visitors may then want to call `service(KotlinImportService.class)` to access that.

kunli2: Thanks for the inputs, do you think changing to service lookup by a fixed name (pre-defined like `IMPORT_SERVICE` , `AUTOFORMAT_SERVICE`) name makes more sense?

PR #3481, opened 2023-08-16
Upgrade Maven plugin dependencies

timtebeek:
## What's changed?
First attempt at upgrading plugin dependency versions.

## What's your motivation?
Maven plugins can have dependencies themselves, as the rewrite-maven-plugin does.
We want to support upgrading those dependencies too.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
@@ -161,7 +161,7 @@ public class UpgradeDependencyVersion extends ScanningRecipe<Set<GroupArtifact>>
                             }
                         }
                     } else if (isPluginDependencyTag(groupId, artifactId)) {
-                        // TODO handle plugin dependencies
+                        t = upgradePluginDependency(ctx, t);
                     }
                 } catch (MavenDownloadingException e) {
                     return e.warn(t);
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -271,6 +275,34 @@ private TreeVisitor<Xml, ExecutionContext> upgradeManagedDependency(Xml.Tag tag,
...
+                String version = t.getChildValue("version").orElse(null);
+                if (groupId != null && artifactId != null && version != null) {
+                    String newerVersion = findNewerVersion(groupId, artifactId, resolveVersion(version), ctx);
+                    if (newerVersion != null) {
+                        if (version.startsWith("${") && !implicitlyDefinedVersionProperties.contains(version)) {
+                            doAfterVisit(new ChangePropertyValue(version.substring(2, version.length() - 1), newerVersion, overrideManagedVersion, false).getVisitor());
+                        } else {
+                            Optional<Xml.Tag> versionTag = t.getChild("version");
+                            assert versionTag.isPresent();
+                            t = (Xml.Tag) new ChangeTagValueVisitor<>(versionTag.get(), newerVersion).visitNonNull(t, 0, getCursor().getParentOrThrow());
```
timtebeek, line 290: I don't think Maven plugin dependencies allow for missing version tags, as they don't partake in dependency management. And we already retrieve the version tag value above, so we don't have to assert anything here.
```suggestion
                            t = (Xml.Tag) new ChangeTagValueVisitor<>(version, newerVersion).visitNonNull(t, 0, getCursor().getParentOrThrow());
```
Or am I missing something?

timtebeek (commented):
Approved aside from that one suggestion. Thanks for taking this on!

rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
joanvr: The only reason I was adding the assert, is to remove the warnings on IntelliJ, because as you said, we already got the value before and we know it's there for sure, but here we need the tag, instead of the value... 
I agree that we can also change the assignment for a return, but I just do it this way to be more similar to how the other code is in the recipe.

joanvr (approved):

PR #3484, opened 2023-08-17
Fix a bug of last comment suffix

kunli2:
We found a maybe long-existing bug in the method `Space#format` that the suffix of the last comment could be wrong.
For example:
```java
class A
// comment
```
The last comment here should have an empty suffix, however, it has a `\n` suffix.


traceyyoshima:
https://github.com/openrewrite/rewrite-kotlin/issues/208

knutwannheden (approved):

PR #3486, opened 2023-08-18
Changing method type in ReorderMethodArguments

joanvr:
## What's changed?
This recipe was not changing the method type, so even when changing the order of the arguments, it was matching again the next cycle and "reordering" the arguments again. If we update the methodType of the methodInvocation it's not needed to "hack" the tests forcing one cycle only, and thus we can check the idempotency of the recipe.

## What's your motivation?
I was using this recipe in another recipe, and not getting the expected results, because the recipe was applied two times (one in each cycle).

@knutwannheden @timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
@@ -16,13 +16,9 @@
 package org.openrewrite.java;
 
 import org.junit.jupiter.api.Test;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
-import static org.openrewrite.test.RewriteTest.toRecipe;
 
 class ReorderMethodArgumentsTest implements RewriteTest {
 
@@ -76,7 +72,7 @@ class ReorderMethodArgumentsTest implements RewriteTest {
     void reorderArgumentsWithNoSourceAttachment() {
         rewriteRun(
           spec -> spec.recipe(new ReorderMethodArguments("a.A foo(String,..)",
-              new String[]{"s", "n"}, new String[]{"n", "s"}, null, null)),
+            new String[]{"s", "n"}, new String[]{"n", "s"}, null, null)),
           java(
             """
               package a;
@@ -113,7 +109,7 @@ class ReorderMethodArgumentsTest implements RewriteTest {
     void reorderArgumentsWhereOneOfTheOriginalArgumentsIsVararg() {
         rewriteRun(
           spec -> spec.recipe(new ReorderMethodArguments("a.A foo(String,Integer,..)",
-              new String[]{"s", "o", "n"}, null, null, null)),
+            new String[]{"s", "o", "n"}, null, null, null)),
           java(
             """
               package a;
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
```diff
@@ -31,17 +27,7 @@ void reorderArguments() {
...
-              new String[]{"n", "m", "s"}, null, null, null),
-            toRecipe(() -> new JavaVisitor<>() {
-                @Override
-                public J visitLiteral(J.Literal literal, ExecutionContext p) {
-                    if (literal.getType() == JavaType.Primitive.String) {
-                        doAfterVisit(new ChangeLiteral<>(literal, l -> "anotherstring"));
-                    }
-                    return super.visitLiteral(literal, p);
-                }
-            })
```
joanvr, line 43: I'm not sure what was the purpose of this? It seems not related to the actual test of the recipe, so I decided to remove it

rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
```diff
@@ -85,9 +71,8 @@ public void test() {
     @Test
     void reorderArgumentsWithNoSourceAttachment() {
         rewriteRun(
-          spec -> spec.recipe(new ReorderMethodArguments("a.A foo(..)",
-              new String[]{"s", "n"}, new String[]{"n", "s"}, null, null))
-            .cycles(1).expectedCyclesThatMakeChanges(1),
+          spec -> spec.recipe(new ReorderMethodArguments("a.A foo(String,..)",
```
joanvr, line 74: had to make the matcher a bit more specific, otherwise it was always matching no matter if we changed the method Type

rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
```diff
@@ -123,15 +108,14 @@ public void test() {
     @Test
     void reorderArgumentsWhereOneOfTheOriginalArgumentsIsVararg() {
         rewriteRun(
-          spec -> spec.recipe(new ReorderMethodArguments("a.A foo(..)",
-              new String[]{"s", "o", "n"}, null, null, null))
-            .cycles(1).expectedCyclesThatMakeChanges(1),
+          spec -> spec.recipe(new ReorderMethodArguments("a.A foo(String,Integer,..)",
```
joanvr, line 111: had to make the matcher a bit more specific, otherwise it was always matching no matter if we changed the method Type

rewrite-java-test/src/test/java/org/openrewrite/java/ReorderMethodArgumentsTest.java
```diff
@@ -123,15 +108,14 @@ public void test() {
...
+            new String[]{"s", "o", "n"}, null, null, null)),
           java(
             """
               package a;
               public class A {
                  public void foo(String s, Integer n, Object... o) {}
-                 public void bar(String s, Object... o) {}
+                 public void foo(String s, Object... o) {}
```
joanvr, line 118: changed this to foo, otherwise we were changing to a non existing method, and this bar was useless.

rewrite-java/src/main/java/org/openrewrite/java/ReorderMethodArguments.java
```diff
@@ -144,12 +146,19 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                         // this is a varargs argument
                         List<JRightPadded<Expression>> varargs = originalArgs.subList(fromPos, originalArgs.size());
                         reordered.addAll(varargs);
+                        for (int j = 0; j < varargs.size(); j++) {
+                            reorderedNames.add(name + j);
```
joanvr, line 150: Not sure what to do here, since this is the varargs argument, and can have N arguments, I've decided to put in the names the name of the original vararg + index.

knutwannheden (approved):
LGTM. Is there any issue we should link it to?

timtebeek:
> LGTM. Is there any issue we should link it to?

This was discovered in
- https://github.com/openrewrite/rewrite-okhttp/pull/5

PR #3490, opened 2023-08-19
Allow rewrite-static-analysis to reuse `SimplifyBooleanExpressionVisitor`

timtebeek:
## What's changed?
Add `shouldSimplifyEqualsOn` to `SimplifyBooleanExpressionVisitor`, which a subclass in rewrite-static-analysis can override to accommodate Kotlin.

## What's your motivation?
Such that we don't have to maintain two separate copies while we wait for rewrite-kotlin to merge into OpenRewrite/rewrite.

Do we need additional such calls for comparison to `null` in `maybeReplaceCompareWithNull`?

## Any additional context
Method modeled after what we already have in
https://github.com/openrewrite/rewrite-static-analysis/blob/772c82062d277e3cab1a1006904817a84b484c8c/src/main/java/org/openrewrite/staticanalysis/SimplifyBooleanExpression.java#L200-L213


knutwannheden (approved):
LGTM. I will verify that a Kotlin-specific recipe subclass also works as intended in other respects.

In the future we should introduce some kind of `NullabilityService` using which recipes can query the nullability of expressions and declarations.

Issue #2062, opened 2022-07-19
YamlParsing failure if various unicode characters exist in the source file.


traceyyoshima:
Example:
```yaml
root:
  - value1: 🛠
    value2: check
```

Exception:
```
com.fasterxml.jackson.databind.JsonMappingException: Invalid surrogate pair, starts with invalid high surrogate (0xDEE0), not in valid range [0xD800, 0xDBFF] (through reference chain: java.util.ArrayList[6]->org.openrewrite.yaml.tree.Yaml$Documents["documents"]->java.util.ArrayList[0]->org.openrewrite.yaml.tree.Yaml$Document["block"]->org.openrewrite.yaml.tree.Yaml$Mapping["entries"]->java.util.ArrayList[0]->org.openrewrite.yaml.tree.Yaml$Mapping$Entry["value"]->org.openrewrite.yaml.tree.Yaml$Mapping["entries"]->java.util.ArrayList[1]->org.openrewrite.yaml.tree.Yaml$Mapping$Entry["value"]->org.openrewrite.yaml.tree.Yaml$Sequence["entries"]->java.util.ArrayList[0]->org.openrewrite.yaml.tree.Yaml$Sequence$Entry["block"]->org.openrewrite.yaml.tree.Yaml$Mapping["entries"]->java.util.ArrayList[1]->org.openrewrite.yaml.tree.Yaml$Mapping$Entry["prefix"])
[...]

PR #3507, opened 2023-08-28
yaml parser support unicode

kunli2:
Update the YAML parser to support multiple bytes (> 2) unicode characters.

Root cause: Snake-yaml parser is based on the characters index and the reader (prefix parsing is based on the reader index) is based on the source index.
If there are any >2 bytes of Unicode characters in the source code, it will make the index mismatch and cause problem.

fixes https://github.com/openrewrite/rewrite/issues/2062

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/FormatPreservingReader.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/FormatPreservingReader.java
@@ -24,8 +24,6 @@ import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
 
 /**
  * Maintains a sliding buffer of characters used to determine format prefixes of
@@ -39,7 +37,7 @@ class FormatPreservingReader extends Reader {
     // Characters index to source index mapping, valid only when `hasMultiBytesUnicode` is true.
     // Snake yaml parser is based on characters index and reader is based on source index. If there are any >2 bytes
     // unicode characters in source code, it will make the index mismatch.
-    private final List<Range> indexRanges;
+    private final int[] indexes;
 
     @Data
     public static class Range {
@@ -56,8 +54,12 @@ class FormatPreservingReader extends Reader {
         this.delegate = new StringReader(source);
 
         boolean hasUnicodes = false;
-        List<Range> ranges = new ArrayList<>(source.length());
+        int[] pos = new int[source.length() + 1];
+
         int cursor = 0;
+        int i = 1;
+        pos[0] = 0;
+
         while (cursor < source.length()) {
             int newCursor = source.offsetByCodePoints(cursor, 1);
             int start = cursor;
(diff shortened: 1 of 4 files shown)


kunli2:
> what
I think for the big Yaml file, if there are no >2 bytes Unicode characters contained, there is no change so the performance is the same as before.
For the big yaml file which contains >2 bytes Unicode characters, this algorithm is O(1) time complexity and O(N) space complexity, which I think it's fine.

kunli2:
> Can we do a performance benchmark to check what the effect is of this change when parsing big YAML files? Possibly also on the memory overhead.


I think for the big Yaml file, if there are no >2 bytes Unicode characters contained, there is no change so the performance is the same as before.
For the big yaml file which contains >2 bytes Unicode characters, this algorithm is O(1) time complexity and O(N) space complexity, which I think it's fine.

knutwannheden:
I am mostly worried about this index map. While complexity of parsing doesn't increase in terms of Big-O, the parse time could still be considerably increased due to the frequent hashing and resizing. Also, the additional space required is not entirely negligible.

kunli2:
> Can we do a performance benchmark to check what the effect is of this change when parsing big YAML files? Possibly also on the memory overhead.

How to do a performance benchmark? by JMH? I think I need to learn about this

knutwannheden:
Is it correct that while parsing we create one index map entry for each parsed character? So for a 1MB file that would be at least 80MB of memory alone for this map (16 bytes for the map key, 24 bytes for the value, 40 bytes for overhead). Or do we only index some of the characters?

knutwannheden:
Yes, JMH would be the way to go. There are some examples in our repo, but I recommend you read the documentation.

kunli2:
I just realized that the algorithm can be optimized to use a List instead of a map, so we will have no concern about frequent hashing and resizing. updated the code.

knutwannheden:
I see you switched to List. That will be much better. But do we even need the Range? Can't you just get the value for the next element in the list and use that (minus one) as end? Haven't looked at it in detail. And if you then switch to `int[]` it will be a lot more compact and faster.

kunli2:
> I see you switched to List. That will be much better. But do we even need the Range? Can't you just get the value for the next element in the list and use that (minus one) as end? Haven't looked at it in detail. And if you then switch to `int[]` it will be a lot more compact and faster.

I think ranges are required, it needs a start and end, otherwise, the algorithm will not work.
Using int[] instead might be a good idea.

kunli2:
> I see you switched to List. That will be much better. But do we even need the Range? Can't you just get the value for the next element in the list and use that (minus one) as end? Haven't looked at it in detail. And if you then switch to `int[]` it will be a lot more compact and faster.

I see your point, I think after switching to a list instead of a map, we can get rid of range by an int array `int[]` should work.

knutwannheden:
I think you probably don't need the Range object. But I haven't looked at it in detail. If you like I can take a look tomorrow. That would save 20 bytes per character.

knutwannheden:
With only `int[]` you are now down to 4 bytes per character, which is 20 times better than what you first had 🙂

kunli2:
yeah, thanks for the insight! 
polished the algorithm, and I think it has better memory consumption now.

knutwannheden (approved):
LGTM.

knutwannheden:
An additional overhead of up to 4 bytes per byte should be acceptable. If not, we can still bring it down more.

PR #3514, opened 2023-08-29
Fix the Javadoc Cursor path not being restored

JLLeitschuh:

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavadocVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavadocVisitor.java
@@ -15,16 +15,30 @@
  */
 package org.openrewrite.java;
 
+import org.openrewrite.Cursor;
+import org.openrewrite.Tree;
 import org.openrewrite.TreeVisitor;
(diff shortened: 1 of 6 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavadocVisitor.java
```diff
@@ -27,6 +30,13 @@ public JavadocVisitor(JavaVisitor<P> javaVisitor) {
         this.javaVisitor = javaVisitor;
     }
 
+    public J visitJavaVisitor(@Nullable Tree tree, P p) {
```
JLLeitschuh, line 33: This should be private

JLLeitschuh: Or... likely `protected`

sambsnyd (approved):

rewrite-java/src/main/java/org/openrewrite/java/JavadocVisitor.java
JLLeitschuh: Fixed

PR #3515, opened 2023-08-30
Added new recipe to change project version of a maven project.

joanvr:
## What's changed?
Added a new recipe to allow changing the project version of a maven project.

## What's your motivation?
Customer request

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeProjectVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeProjectVersion.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2023 the original author or authors.
  * <p>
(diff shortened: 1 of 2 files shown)


timtebeek:
Would we want to support multi module Maven projects here, where we update the children when we update the parent version, and any references between children? We could likely reuse existing recipes for such follow up changes.

joanvr:
> Would we want to support multi module Maven projects here, where we update the children when we update the parent version, and any references between children? We could likely reuse existing recipes for such follow up changes.

Good point, maybe update parents if the groupId, artifactId and oldVersion matches? But not sure if this is responsibility of this recipe or the user should build a yaml recipe with this recipe and (potentially) ChangeParentPom to accomplish it

knutwannheden:
Can we add a test that this works correctly when there is a parent pom in the project? AFAIK the version is inherited and should then not be set in the child project, if it isn't already set (I think it is optional in that case).

PR #3521, opened 2023-08-31
fix: use source file encoding for reading parser input (#3520)

### LINKED ISSUE #3520: Parsing error with ISO-8859-1 encoded source files
I am using

- OpenRewrite v8.4.2
- Maven plugin v5.4.2
- rewrite-core v8.4.2

I am using the Maven plugin, and my project is a multi module project with encoding ISO-8859-1.

Use source files with ISO-8859-1 encoding and special characters.

Successful parsing.

## What did you see instead?
A parsing error "There were problems parsing some source files..."
There is a check during parsing calling `Parser.requirePrintEqualsInput` leading to a parsing error. The error happens because of comparing the `SourceFile` with the corresponding encoding with the `Parser.Input` with UTF-8 encoding.

Yes 👍

thomaszub:
## What's changed?
Use the source file encoding for the parser input for comparing if they are equal.

## What's your motivation?
Fixes #3520 

### Checklist


timtebeek:
Looks good to me, and thanks for working out the tests to fix
- #3520 

I'm adding @jkschneider as a reviewer since I don't know if there was any intention behind using the default charset, or if this might have unintended effects downstream.

rewrite-core/src/main/java/org/openrewrite/SourceFile.java
```diff
@@ -36,7 +36,8 @@ public interface SourceFile extends Tree {
      * @return <code>true</code> if the parse-to-print loop is idempotent, <code>false</code> otherwise.
      */
     default boolean printEqualsInput(Parser.Input input, ExecutionContext ctx) {
-        return printAll().equals(StringUtils.readFully(input.getSource(ctx)));
+        Charset charset = (getCharset() == null) ? StandardCharsets.UTF_8 : getCharset();
+        return printAll().equals(StringUtils.readFully(input.getSource(ctx), charset));
```
timtebeek, line 40: There's some slight subtleties that I want to point out here, and prevent those from causing issues by suggesting an alternative implementation.

The `StringUtils.readFully(InputStream)` has an early `if (inputStream == null) return "";`, that's not present in `readFully(InputStream, Charset)`. Not sure if we rely on that, but probably best to not change behavior just yet by calling the old method when there's no charset.

There's 14+ implementations of `SourceFile` in openrewrite/rewrite alone, with further implementations in modules for Kotlin, Python and JavaScript for instance. Some of these, such as Binary and Quark, throw an `UnsupportedOperationException` when `getCharset()` is called. I don't think the print equality is actually called for those, but figured call those out here all the same in case that triggers anyone to rethink the approach.

With the above I suggest we instead go for the following, to call `getCharset()` only once, after printing, and call the respective read operation that retains the `""` for a `null` `InputStream`.
```suggestion
        String printed = printAll();
        Charset charset = getCharset();
        if (charset != null) {
            return printed.equals(StringUtils.readFully(input.getSource(ctx), charset));
        }
        return printed.equals(StringUtils.readFully(input.getSource(ctx)));
```

thomaszub:
I added your suggested change. I think for consistency the `StringUtils.readFully(inputStream, charset)` should be changed to contain the check for a `null` inputStream and the return of a empty string, so the behaviour is the same as `StringUtils.readFully(inputStream)`.

timtebeek:
> I added your suggested change. I think for consistency the `StringUtils.readFully(inputStream, charset)` should be changed to contain the check for a `null` inputStream and the return of a empty string, so the behaviour is the same as `StringUtils.readFully(inputStream)`.

Yes I'd considered that as well, but the `InputStream` is only annotated `@Nullable` on the single argument method, not in the dual argument method. Ideally the IDE should then already prevent calls where the inputStream might be null, although there's no hard guarantees of course.

I think we can keep things as they are right now without any adverse effects. Thanks for adding this change!

PR #3527, opened 2023-09-05
Issue 3275/fix remove used param v2

tclayton-newr:
## What's changed?

This is a follow on to https://github.com/openrewrite/rewrite/pull/3397 which attempted to address https://github.com/openrewrite/rewrite/issues/3275. 

This change attempts to bridge the gap between static fully qualified paths and non-static imports. In addition it addresses one concern about unfolding wildcard imports by introducing an a boolean option to "Unfold wildcard imports".

## What's your motivation?

My motivation is to resolve https://github.com/openrewrite/rewrite/issues/3275 and make the RemoveUnusedImports recipe more reliable for wider adoption.

Could you double check that I added the boolean option to this recipe correctly? I attempted to copy the pattern from other recipes. Does any documentation need to be updated or is that done as part of the build? 

Any maintainer 😄

n/a

## Any additional context

n/a

### Checklist
- [ ] I've updated the documentation (if applicable) (unsure about this one)

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
@@ -31,7 +31,7 @@ class RemoveUnusedImportsTest implements RewriteTest {
 
     @Override
     public void defaults(RecipeSpec spec) {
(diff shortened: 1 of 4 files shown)


SDNFKJSDK (commented):
kk

PR #3535, opened 2023-09-09
feat; add testcase for multiline non regex replace

SimonVerhoeven:
To reproduce https://rewriteoss.slack.com/archives/C01A843MWG5/p1694144028395779

## What's changed?

I added an extra multiline non regex testcase to expand the testing, and try and see if I could reproduce the issues with: https://rewriteoss.slack.com/archives/C01A843MWG5/p1694144028395779

## What's your motivation?

To add an extra testcase, and reproduce an issue if possible

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/text/FindAndReplaceTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/text/FindAndReplaceTest.java
@@ -92,4 +92,52 @@ class FindAndReplaceTest implements RewriteTest {
           text("test", "tested")
         );
     }
+
+    @Test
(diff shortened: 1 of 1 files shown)


timtebeek:
Looking at the Slack thread the issue appears to have been in the yaml configuration, rather than the recipe itself. Thanks a lot for helping to debug that issue!

That being said I'm afraid having an example that uses string replacements on java files might send the wrong message, as we want to nudge folks towards Java visitors instead. Hope you don't mind that I close this PR for that reason, as much as I appreciate you helping others in the community!

Look forward to seeing you at Devoxx in three weeks. :)

Issue #2735, opened 2023-01-29
Maven - Explicitly define version of a plugin when the version is not set


yeikel:
Plugin versions in Maven are optional as defaults to the latest version  reading [maven-metadata.xml](https://repo.maven.apache.org/maven2/org/codehaus/mojo/versions-maven-plugin/maven-metadata.xml)

This pattern is dangerous because creates non-reproducible builds.  

Maven will also deprecate this pattern in the future: 

> 
> [WARNING]
> [WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-surefire-plugin is missing. @ line 72, column 21
> [WARNING]
> [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
> [WARNING]
> [WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
> 
> 

```xml

<plugins>
  <plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>versions-maven-plugin</artifactId>
  </plugin>
</plugins>
```

This recipe should replace that pattern to set a specific version of the plugin

```xml
<plugins>
  <plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>versions-maven-plugin</artifactId>
     <version>2.14.2</version> 
  </plugin>
</plugins>
```

Can we also allow inputs like `LATEST` but resolve it to the latest from the maven-metadata? 

The recipe "Upgrade Maven plugin version" might be able to do this but it is not clear 

Another way we might be able to do this is by using `RemovePlugin` and `AddPlugin` but I imagine that the diff will be greater if we do so

timtebeek:
Good suggestion indeed, and thanks for the detailed issue report!

yeikel:
I just noticed that **`ChangeDependencyGroupIdAndArtifactId`** can be used for this as well so maybe we can just create a YML only recipe for this

yeikel:
@timtebeek Can you assign this to me?

timtebeek:
Hope you don't mind I unassigned you @yeikel 
I want to open this issue back up again to anyone to pick up.
Feel free to open up a draft pull request if you'd like to contribute!

timtebeek:
Implemented in
- https://github.com/openrewrite/rewrite/pull/3592

PR #3538, opened 2023-09-11
AddPlugin changes; make `version` optional and add `filePattern`

Wallman:
## What's changed?

- `version` is now optional, this is useful for example when inheriting it from a parent.
- Added `filePattern` , this is useful in many cases but for me when working with a multi module maven project.

## What's your motivation?

See section above.

Authors of addPlugin.
@jkschneider 
@sambsnyd
@BoykoAlex 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
@@ -107,10 +107,9 @@ public class AddPlugin extends Recipe {
         @Override
         public boolean isAcceptable(SourceFile sourceFile, ExecutionContext executionContext) {
             if (filePattern != null) {
-                return PathUtils.matchesGlob(sourceFile.getSourcePath(), filePattern);
-            } else {
-                return sourceFile instanceof Xml.Document;
+                return PathUtils.matchesGlob(sourceFile.getSourcePath(), filePattern) && super.isAcceptable(sourceFile, executionContext);
             }
+            return super.isAcceptable(sourceFile, executionContext);
         }
 
         @Override
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
@@ -83,7 +83,7 @@ public class AddPlugin extends Recipe {
                     "If multiple patterns are supplied any of the patterns matching will be interpreted as a match. " +
                     "When not set, all source files are searched. ",
             required = false,
-            example = "**/*.java")
+            example = "**/*-parent/grpc-*/pom.xml")
     @Nullable
     String filePattern;
 
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
@@ -48,7 +48,9 @@ public class AddPlugin extends Recipe {
 
     @Option(displayName = "Version",
             description = "A fixed version of the plugin to add.",
-            example = "1.0.0")
+            example = "1.0.0",
+            required = false)
+    @Nullable
     String version;
 
     @Language("xml")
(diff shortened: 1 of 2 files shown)


Wallman:
https://github.com/openrewrite/rewrite-docs/pull/232

timtebeek:
Thanks for getting this started @Wallman ! I agree with making the version optional, provided there's a parent pom.xml, such as not to run afoul of
- https://github.com/openrewrite/rewrite/issues/2735

Not entirely sure about the filePattern, and if that generalizes well; is your intention to only add the plugin to a specific sub project? Would it then make sense to run only against that subproject instead?

We tend to write recipes such that they can run against a large collection of projects in parallel; having a filePattern parameter that matches a particular sub module can make that slightly awkward, unless your project structure is highly standardized across an organization. Could you give us a bit more detail about the usages you see?

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
```diff
@@ -90,6 +104,15 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
 
     private class AddPluginVisitor extends MavenIsoVisitor<ExecutionContext> {
 
+        @Override
+        public boolean isAcceptable(SourceFile sourceFile, ExecutionContext executionContext) {
+            if (filePattern != null) {
+                return PathUtils.matchesGlob(sourceFile.getSourcePath(), filePattern);
+            } else {
+                return sourceFile instanceof Xml.Document;
+            }
```
timtebeek, line 113: You'll also want to call out to the super implementation defined here
https://github.com/openrewrite/rewrite/blob/de8bc15be8847027e197ad4ef73df528cc5a3f7e/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java#L55-L57

Otherwise we risk running the visitor against any matching file, or xml document, even those that are not Maven pom files.

Wallman:
> Thanks for getting this started @Wallman ! I agree with making the version optional, provided there's a parent pom.xml, such as not to run afoul of
> 
> * [Maven - Explicitly define version of a plugin when the version is not set #2735](https://github.com/openrewrite/rewrite/issues/2735)
> 
> Not entirely sure about the filePattern, and if that generalizes well; is your intention to only add the plugin to a specific sub project? Would it then make sense to run only against that subproject instead?
> 
> We tend to write recipes such that they can run against a large collection of projects in parallel; having a filePattern parameter that matches a particular sub module can make that slightly awkward, unless your project structure is highly standardized across an organization. Could you give us a bit more detail about the usages you see?

We have a "highly standardized" project structure like so;

```
car-parent
├── car-service
│   ├── pom.xml
├── grpc-car
│   ├── pom.xml
├── message-definitions-car
│   ├── pom.xml
├── pom.xml
```

Where we wanted to add the plugin only to car-parent/car-service/pom.xml. We usually have one declarative recipe for a whole "refactoring" ie we want to run the recipe on the whole project. See example below.

```
name: sh.vcm.XRecipe
recipeList:
  - org.openrewrite.maven.AddPlugin:
      groupId: org.apache.maven.plugins
      artifactId: maven-dependency-plugin
      fileMatcher: '**/*-parent/*-service/pom.xml'
  - org.openrewrite.maven.AddPlugin:
      groupId: org.xolstice.maven.plugins
      artifactId: protobuf-maven-plugin
      fileMatcher: '**/*-parent/grpc-*/pom.xml'
```

What do you think?

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
Wallman: Thanks, I commited a fix for this

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
```diff
@@ -73,6 +77,16 @@ public class AddPlugin extends Recipe {
     @Nullable
     String executions;
 
+    @Option(displayName = "File pattern",
+            description = "A glob expression that can be used to constrain which directories or source files should be searched. " +
+                    "Multiple patterns may be specified, separated by a semicolon `;`. " +
+                    "If multiple patterns are supplied any of the patterns matching will be interpreted as a match. " +
+                    "When not set, all source files are searched. ",
+            required = false,
+            example = "**/*.java")
```
timtebeek, line 86:
```suggestion
            example = "'**/*-parent/grpc-*/pom.xml")
```

timtebeek (approved):
Looks good to me; I imagine the `filePattern` argument won't be used as often, and I suspect we might remove that in the future if we restore something akin to the applicability tests we had in Rewrite 7, but for now I'll see this through as I do feel we'll see such highly structured projects across organizations more often, and I feel a custom local recipe for that is too heavy handed an ask.

PR #3542, opened 2023-09-13
Add 409 and other non retryable status codes to isClientSideException()

pstreef:
This should add these responses as empty results in the maven cache

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -837,7 +837,7 @@ public class MavenPomDownloader {
         }
 
         public boolean isAccessDenied() {
-            return responseCode != null && responseCode > 400 && responseCode <= 403;
+            return responseCode != null && 400 < responseCode && responseCode <= 403;
         }
     }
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -816,8 +817,17 @@ public HttpSenderResponseException(@Nullable Throwable cause, @Nullable Integer
...
 
+        /**
+         * All 400s are considered client-side exceptions, but we only want to cache ones that are unlikely to change
+         * if requested again in order to save on time spent making HTTP calls.
+         * <br/>
+         * For 408 TIMEOUT, 425 TOO_EARLY, and 429 TOO_MANY_REQUESTS, these are likely to change if not cached.
+         */
         public boolean isClientSideException() {
-            return responseCode != null && responseCode >= 400 && responseCode <= 404;
+            if (responseCode == null || responseCode < 400 || responseCode > 499) {
```
timtebeek, line 827: Again; a preference to show visually that responseCode is outside a range.
```suggestion
            if (responseCode == null || responseCode < 400 || 499 < responseCode) {
```

PR #3537, opened 2023-09-10
Add support for `relativePath` change on `ChangeParentPom`

jonesbusy:
## What's changed?

Add support for update/adding `relativePath` on `UpdateParentPom` receipe

Fixes https://github.com/openrewrite/rewrite/issues/3536

## What's your motivation?

Refactoring multi-module Maven projects

## Have you considered any alternatives or workarounds?

None found

## Any additional context

Tested with

I've added 3 unit tests to cover 3 new use case

- Update relative path
- Add new relative path if not set (empty value)
- Add new relative path if not set (non-empty value)

```
./gradlew :rewrite-maven:test --info --tests '*ChangeParentPomTest' --continue
```

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ IDEA auto-formatter on affected files
- [ ] I've updated the documentation (if applicable)

Let me know if ok. I can also update the documentation if needed

PR #3544, opened 2023-09-14
Fix issue with duplicate `relativePath` on `ChangeParentPom`

jonesbusy:
Found issue when testing https://github.com/openrewrite/rewrite/pull/3537 in real use case.

I ended-up with duplicate `relativePath` tag on maven pom.

I'm not sure why, I was not able to reproduce with unit test. I think the model need to be refreshed earlier.

In any case I've added a tentative fix and increase the test coverage for multi module

- Changing parent module only
- Changing child module to point to a different external pom

I've also ensured the relativePath is added after the version on the parent pom-

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
@@ -205,12 +205,15 @@ public class ChangeParentPom extends Recipe {
                                     changeParentTagVisitors.add(new ChangeTagValueVisitor<>(t.getChild("relativePath").get(), targetRelativePath));
                                 }
                                 else if (tag.getChildValue("relativePath").orElse(null) == null && targetRelativePath != null) {
+                                    Xml.Tag relativePathTag = null;
                                     if (StringUtils.isBlank(targetRelativePath)) {
-                                        changeParentTagVisitors.add(new AddToTagVisitor<>(t, Xml.Tag.build("<relativePath />")));
+                                        relativePathTag = Xml.Tag.build("<relativePath />");
                                     }
                                     else {
-                                        changeParentTagVisitors.add(new AddToTagVisitor<>(t, Xml.Tag.build("<relativePath>" + targetRelativePath + "</relativePath>")));
+                                        relativePathTag = Xml.Tag.build("<relativePath>" + targetRelativePath + "</relativePath>");
                                     }
+                                    doAfterVisit(new AddToTagVisitor<>(t, relativePathTag, new MavenTagInsertionComparator(t.getChildren())));
+                                    maybeUpdateModel();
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
```diff
@@ -205,12 +205,15 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                                     changeParentTagVisitors.add(new ChangeTagValueVisitor<>(t.getChild("relativePath").get(), targetRelativePath));
                                 }
                                 else if (tag.getChildValue("relativePath").orElse(null) == null && targetRelativePath != null) {
+                                    Xml.Tag relativePathTag = null;
```
timtebeek, line 208:
```suggestion
                                    final Xml.Tag relativePathTag;
```

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
```diff
@@ -236,6 +236,181 @@ void changeParentAddRelativePathNonEmptyValue() {
         );
     }
 
+    @Test
+    @RepeatedTest(10)
```
timtebeek, line 240: There's no need to tag a test with both `@Test` and `@RepeatedTest`; I'm assuming you wanted to retain the `@RepeatedTest`, but we can also go for `@Test` here.
```suggestion
    @RepeatedTest(10)
```

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
```diff
@@ -236,6 +236,181 @@ void changeParentAddRelativePathNonEmptyValue() {
...
+                </project>
+                """
+            )
+          )
+        )
+      );
+  }
+
+  @Test
+  @RepeatedTest(10)
```
timtebeek, line 325:
```suggestion
  @RepeatedTest(10)
```

timtebeek (approved):
Mostly looks good to me; thanks for coming back with these fixes!
Some minor suggestions. Feel free to apply those as you see fit an I'll merge after that.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
timtebeek: Since we're assigning in both paths of the if, and using the result regardless I figured make it explicit that `null` is never used or expected.

timtebeek:
Thanks again! You can try out this change through [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

Issue #3559, opened 2023-09-22
Caused by: java.lang.RuntimeException: Error while visiting build.gradle: java.lang.IllegalStateException: CompilationFailedException


safetytrick:
## What version of OpenRewrite are you using?

I am using:

- Gradle plugin v6.3.8
- rewrite-spring v5.0.10

## How are you running OpenRewrite?

I am using the Gradle plugin through an init script, my project is a single module project.
```groovy
initscript {
    repositories {
        maven { url "https://plugins.gradle.org/m2" }
    }
    dependencies { classpath("org.openrewrite:plugin:6.3.8") }
}
rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite.recipe:rewrite-spring:5.0.10") // I also tried 5.0.8
    }
    rewrite {
        activeRecipe("org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_1") // I also tried going to 3_0 first
    }
    afterEvaluate {
        if (repositories.isEmpty()) {
            repositories {
                mavenCentral()
            }
        }
    }
}

```

## What is the smallest, simplest way to reproduce the problem?

I will come back to this, my build.gradle is very simple and I've tried removing most of everything from the file to eliminate the problem but nothing seems to change the problem. 

## What is the full stack trace of any errors you encountered?

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':rewriteRun'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:142)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:140)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:128)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
[...]

PR #3563, opened 2023-09-25
Rewrite escaped groovy method names

safetytrick:
And use the escaped groovy method name rewriting to update dependencies with the gradle default scope (although why this scope is being chosen by the java update recipe is unclear).

## What's changed?

Added handling for escaped method names in GroovyParserVisitor, then check for configurations that require escaping in the AddDependencyVisitor

## What's your motivation?
This is an enhancement that is also a workaround for https://github.com/openrewrite/rewrite/issues/3559.

A more complete fix would involve determining why the AddJaxbRuntime recipe is selecting certain configurations like default and rewrite to add dependencies to. 

@shanman190 @timtebeek 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
@@ -16,6 +16,7 @@
 package org.openrewrite.gradle;
 
 import lombok.RequiredArgsConstructor;
+
 import org.openrewrite.*;
 import org.openrewrite.gradle.internal.InsertDependencyComparator;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
@@ -37,7 +38,6 @@ import org.openrewrite.maven.tree.*;
 import org.openrewrite.semver.*;
 import org.openrewrite.tree.ParseError;
 
-import java.text.ParseException;
 import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
```diff
@@ -16,6 +16,8 @@
 package org.openrewrite.gradle;
 
 import lombok.RequiredArgsConstructor;
+
+import org.codehaus.groovy.syntax.Types;
```
safetytrick, line 20: I noticed this import was the only one outside of openrewrite and java, is this considered bad form? Where would you recommend putting this dependency?

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
```diff
@@ -333,6 +334,13 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
         }
     }
 
+    private String escapeIfNecessary(String configurationName) {
```
safetytrick, line 337: In theory a gradle configuration could have a name with spaces and potentially escape characters but I chose not to attempt full string escaping.

shanman190: So we may need to be a little more explicit here rather than relying on the Groovy type names themselves. As an example, the `rewrite-gradle` module depends upon `rewrite-groovy` which presently uses the Groovy 3.x api surface. However, at runtime `rewrite-gradle` might be using Groovy 2.x or 3.x depending on what version of Gradle it is operating within.

shanman190 (approved):
LGTM

timtebeek (approved):

timtebeek:
Thanks for the detailed work on this @safetytrick , and thanks for the helpful hints and review @shanman190 !

PR #3564, opened 2023-09-26
Type validation for before sources in `RewriteTest`

knutwannheden:
Add a new `RecipeSpec#afterTypeValidationOptions()` method along the existing `RecipeSpec#typeValidationOptions()` method. The former can be used to independently control the type validation options for the after sources. Otherwise the latter applies to both before and after sources.

Renames `SourceSpec.EachResult` to `SourceSpec.ValidateSource`.

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/test/TypeValidation.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/TypeValidation.java
@@ -46,20 +46,23 @@ public class TypeValidation {
     @Builder.Default
     private boolean constructorInvocations = true;
 
+    public static TypeValidation all() {
+        return new TypeValidation();
+    }
+
     public static TypeValidation none() {
         return new TypeValidation(false,false,false,false,false,false);
     }
 
     static TypeValidation before(RecipeSpec testMethodSpec, RecipeSpec testClassSpec) {
-        TypeValidation typeValidation = testMethodSpec.getBeforeTypeValidation() != null ?
-                testMethodSpec.getBeforeTypeValidation() : testClassSpec.getBeforeTypeValidation();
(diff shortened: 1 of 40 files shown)


knutwannheden:
Created as draft because I would first proactively like to fix the violating tests in the more prominent recipe repos and also get some feedback on the implementation.

knutwannheden:
Note: This will require small changes to `rewrite-kotlin` and `rewrite-javascript`, as they explicitly reference `SourceSpec.EachResult.noop`.

sambsnyd (approved):
Good work!

rewrite-test/src/main/java/org/openrewrite/test/TypeValidation.java
```diff
@@ -49,4 +49,17 @@ public class TypeValidation {
     public static TypeValidation none() {
         return new TypeValidation(false,false,false,false,false,false);
     }
+
+    static TypeValidation before(RecipeSpec testMethodSpec, RecipeSpec testClassSpec) {
+        TypeValidation typeValidation = testMethodSpec.getBeforeTypeValidation() != null ?
+                testMethodSpec.getBeforeTypeValidation() : testClassSpec.getBeforeTypeValidation();
+        // TODO: Default to `after()` is for backwards compatibility
+        return typeValidation != null ? typeValidation : after(testMethodSpec, testClassSpec);
```
knutwannheden, line 57: Having the `before` validations default to the `after` validations is probably just confusing. It does have the advantage of offering backwards compatibility, however.

I am thinking we might want to integrate it like this first to make sure all receipe repos build successfully and then make the two validations independent in a second step. But unsure about that as well.

PR #3584, opened 2023-09-30
Lazily evaluate chained validations

### LINKED ISSUE #3580: NPE in SimplifyMethodChain
org.openrewrite.java.SimplifyMethodChain without arguments.

```
 java.lang.NullPointerException
	at org.openrewrite.java.SimplifyMethodChain.lambda$validate$0(SimplifyMethodChain.java:64)
	at org.openrewrite.Validated.test(Validated.java:81)
	at org.openrewrite.java.SimplifyMethodChain.validate(SimplifyMethodChain.java:62)
	at org.openrewrite.Recipe.validate(Recipe.java:283)
	at org.openrewrite.Recipe.validate(Recipe.java:286)
	at org.openrewrite.Recipe.validate(Recipe.java:286)
	at org.openrewrite.Recipe.validateAll(Recipe.java:319)
```

Yes.

Root cause is verifying the list size without verifying there is actually a list: https://github.com/openrewrite/rewrite/blob/0d526a0dc6593d3c47e0d622f9b6f7676aee6615/rewrite-java/src/main/java/org/openrewrite/java/SimplifyMethodChain.java#L64
I would just add a null check on c before accessing its size.

Bananeweizen:
To support lazy evaluation of `someValidation.and(Validate.test(predicate))`, the predicate MUST NOT be evaluated in the test method already. Rather return an anonymous subclass that only evaluates the predicate in isValid().

To still support the reporting of validation failures, the anonymous subclass must be of type Invalid, even though the validation result may be valid.

Out of the 3 new unit tests, the second one fails without this change, showcasing the original problem.

Fixes #3580.

## What's changed?
Validate.test() no longer evaluates the given predicate immediately, but only when the isValid() method is called. That allows chaining of someValidation.and(...) respectively someValidation.or(...) to return early, without evaluation the given predicate at all. That's extremely helpful if recipe authors rely on super.validate().and(someMoreValidation), where they access potentially nullable fields in someMoreValidation.

## What's your motivation?
#3580

There is another method `testNone` that has might have the same problem. However, I haven't really understood the `None` concept, therefore I have not touched that code.

### Checklist

Side note: The last item is not helpful for non IDEA users. Can you link to some command line execution of the formatter, too (in the PR message template comment)?

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Validated.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Validated.java
@@ -66,7 +66,8 @@ public interface Validated<T> extends Iterable<Validated<T>> {
 
     /**
      * Validate that the Predicate will evaluate to 'true' on the supplied value.
-     * When the Predicate evaluates to 'false' the error message will be of the form:
+     * When the Predicate evaluates to 'false' the error message will be of the
+     * form:
      * <p>
      * "[property] was '[value]' but it [message]"
      *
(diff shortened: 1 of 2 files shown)


knutwannheden:
Maybe we should add a new `Validated#lazy()` method with a corresponding `Lazy` subtype?

knutwannheden:
Here is a recipe run showing how the `test*()` methods accepting a predicate are currently being used: https://app.moderne.io/results/34ALivU2u

knutwannheden:
We can always later consider adding `lazy()` metods.

timtebeek:
Agreed; then let's close this one for now; to be revisited if there's a compelling case. Thanks again @Bananeweizen !

PR #3592, opened 2023-09-30
Add explicit Maven plugin version for reproducibility

timtebeek:
Fixes https://github.com/openrewrite/rewrite/issues/2735


timtebeek:
Also needs a fix to
https://github.com/openrewrite/rewrite/blob/0c03b0fa9ead48f5441f19c955a8266fc79fbb89/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradePluginVersion.java#L76-L82

rewrite-maven/src/main/java/org/openrewrite/maven/UpgradePluginVersion.java
```diff
@@ -178,9 +183,11 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                             doAfterVisit(new ChangeTagValueVisitor<>(versionTag.get(), newVersion));
                         }
                     }
+                } else if (addVersionIfMissing) {
+                    Xml.Tag newTag = Xml.Tag.build("<version>" + newVersion + "</version>");
+                    doAfterVisit(new AddToTagVisitor<>(tag, newTag, new MavenTagInsertionComparator(tag.getChildren())));
```
timtebeek, line 188: Somehow still out of order.

timtebeek:
The code change here mostly follow what we do for other recursively resolved dependencies and properties already. We need the pluginManagement information to be able to tell if a parent already manages a version of that plugin, such that we don't override those versions. We really only want to add plugin versions where they are not otherwise set, such that the builds become reproducible.

knutwannheden (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/ExplicitPluginVersion.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.maven.MavenVisitor;
+import org.openrewrite.maven.UpgradePluginVersion;
+import org.openrewrite.maven.tree.ResolvedPom;
+import org.openrewrite.xml.tree.Xml;
+
+public class ExplicitPluginVersion extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Explicit plugin version";
```
timtebeek, line 30:
```suggestion
        return "Add explicit plugin versions";
```

pstreef:
Sorry for the delay @timtebeek some confusion on my end. I will pick it up right away!

Issue #3582, opened 2023-09-29
Json parser doesn't support multiple bytes unicode


kunli2:
Found this issue from the user's report [here](https://moderneinc.slack.com/archives/C05TWTNB77A/p1695930586910549)

Here is an example test case
```java
    @Test
    void multiByteUnicode() {
        rewriteRun(
          json(
            """
              {
                "id": "2d681f35-9758-43f1-a26f-6d4081a7688c",
                "text": "I'm a virtual assistant here to help you. 🤖",
                "name": "GH_ARE_YOU_REAL",
              }
              """
          )
        );
    }
```

PR #3596, opened 2023-10-02
Update json parser to support multiple-bytes-unicodes

kunli2:
fixes https://github.com/openrewrite/rewrite/issues/3582
Currently, the JSON parser doesn't support multiple bytes Unicode characters (> 2 bytes).
The root cause is the index from Antlr is mismatched with the source index.
For example:
```java
{
  "🤖": "robot",
}
```
Tokens from Antlr
```
Got token: [@0,0:0='{',<1>,1:0]
Got token: [@1,4:6='"🤖"',<10>,2:2]
Got token: [@2,7:7=':',<4>,2:5]
Got token: [@3,9:15='"robot"',<10>,2:7]
Got token: [@4,16:16=',',<2>,2:14]
Got token: [@5,18:18='}',<3>,3:0]
Got token: [@6,19:18='<EOF>',<-1>,3:1]
```

we can see `4:6='"🤖"'`. However, the 🤖 has 4 bytes, so the start/end offset in the source code is not [4,6] but [4,7].

I think all Antlr-based parsers will have the same problem ( doesn’t support multiple bytes of Unicode). and this PR is a common pattern to fix this issue.

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
@@ -49,12 +49,49 @@ public class JsonParserVisitor extends JSON5BaseVisitor<Json> {
 
     private int cursor = 0;
 
-    public JsonParserVisitor(Path path, @Nullable FileAttributes fileAttributes, EncodingDetectingInputStream source) {
+    // Whether the source has multi bytes (> 2 bytes) unicode characters
+    private final boolean hasMultiBytesUnicode;
+    // Antlr index to source index mapping
+    private final int[] indexes;
+
+    public JsonParserVisitor(Path path,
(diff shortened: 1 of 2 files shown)


knutwannheden:
I think it would be better if we would make a change based on the helper function showed here: https://github.com/tunnelvisionlabs/antlr4/blob/master/doc/unicode.md#legacy-grammar-using-surrogate-code-units

kunli2:
> I think it would be better if we would make a change based on the helper function showed here: https://github.com/tunnelvisionlabs/antlr4/blob/master/doc/unicode.md#legacy-grammar-using-surrogate-code-units

Thanks for the input, I think we are using Antrl 4.11 (after 4.7) and already using `CharStreams` so seems this doesn't help.

knutwannheden:
No, simply using `CharStreams` is not what I meant. I will make a PR.

PR #3597, opened 2023-10-03
Remove `JsonParserVisitor#indexes`

knutwannheden:
The `JsonParserVisitor#indexes` array requires an additional 4 bytes of memory per code point in the source code (only in case the source has at least one surrogate). Rather than doing that, the parser can have two cursors which are kept in sync when advancing through the source code: One code point cursor and one code unit cursor. The former is to align with the indexes of the ANTLR tokens and the latter is to be able to read from the underlying source string.

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
@@ -48,50 +48,14 @@ public class JsonParserVisitor extends JSON5BaseVisitor<Json> {
     private final FileAttributes fileAttributes;
 
     private int cursor = 0;
+    private int codePointCursor = 0;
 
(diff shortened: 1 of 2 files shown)


knutwannheden:
@kunli2 I realize that this implementation can be a little bit harder to follow, but this is what I meant when I referred to https://github.com/tunnelvisionlabs/antlr4/blob/master/doc/unicode.md#legacy-grammar-using-surrogate-code-units.

I think we should either integrate this or alternatively rename things in the current implementation a bit. Because the current variable names and comments in the code are a bit misleading. The problem isn't really multi-byte characters in general. In UTF-8 only ASCII characters are single-byte characters. In Java all BMP characters (`U+0000` to `U+FFFF`) can be represented by a single `char`. The problem are the Unicode characters beyond BMP which are represented by surrogate code units and thus require multiple `char`s for a single code point.

kunli2 (approved):
A good improvement!
thanks for this update

Issue #1621, opened 2022-04-12
Maven recipe: prefix POM expressions with project.


timtebeek:
[MNG-7404] Will drop support for prefixless expressions in POM https://github.com/apache/maven/pull/696
Although it's been deprecated for quite some time already, it will only be removed in Maven 4.
It seems like OpenRewrite would be well suited to already recognize & fix unprefixed expressions.

pway99:
Thanks @timtebeek, this is very helpful

mthmulders:
In fact, there are two things changing in Maven 4:
1. `${version}` will no longer work - should become `${project.version}`
1. `${pom.version}` will no longer work - should also become `${project.version}`

The latter one is probably simple, the former one can be more challenging...

timtebeek:
> Can someone check how much of the new recipes overlaps with https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/NormalizeMavenVariables.java? To my understanding that was already half of the work at least.

You're right! Replaced now in #3600

PR #3600, opened 2023-10-03
Merge NormalizeMavenVariables into PrefixlessExpressions

timtebeek:
Due to overlap, and the latter replacing more instances. RenamePropertyKey was added later, and is a good fit to only maintain such logic once.

As indicated in https://github.com/openrewrite/rewrite/issues/1621#issuecomment-1743167485 by @Bananeweizen


rewrite-maven/src/main/java/org/openrewrite/maven/RenamePropertyKey.java
```diff
@@ -50,13 +50,7 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(new MavenVisitor<ExecutionContext>() {
-            @Override
-            public Xml visitDocument(Xml.Document document, ExecutionContext executionContext) {
-                // Scanning every tag's value is not an efficient applicable test, so just accept all maven files
-                return SearchResult.found(document);
```
timtebeek, line 57: Not quite sure what this was for, but it seemed too broad, and not to limit recipe execution by much.

joanvr (approved):

PR #3603, opened 2023-10-04
[TypeUtils.findOverridenMethod] Skip found methods with default access and not same package

### LINKED ISSUE #3599: `TypeUtils.findOverriddenMethod` misidentifies package scoped method from class in different package
Greetings. 

We've found a problem in the [MissingOverrideAnnotation ](org.openrewrite.staticanalysis.MissingOverrideAnnotation)recipe, by which in very specific circunstances, a method will be wrongly annotated, generating compilation issues in processed projects. 

The error can be observed whenever:

- A class extends/implements from a supertype.
- The supertype contains a method which is not exposed externally (that is, with package/default visibility)
- The extending class declares a new method with the same signature.
- The recipe will wrongly consider this copycat method an override and assign the annotation, resulting in compilation failures.

staticAnalysis 1.0.8
RecipeBom 2.3.0
RewriteMavenPlugin 5.7.1

Through maven plugin in multimodule projects.

Given the following example superclass:

```java
package org.example;

public class A {

    public void accessible() {
       // Open for overriding
    }

    void notAccessible() {
        // This method is not visible from subclass due to default access and different package.
    }
}
```

And the following extending class:

```java
package org.example.impl;

import org.example.A;

class B extends A {

    public void accessible() {
        super.accessible();
    }

    void notAccessible() {
        // Same signature than super method but not an override.
    }
}
```

The recipe should add the override annotation to the accesible method.

```java
package org.example.impl;

import org.example.A;

class B extends A {

    @Override
    public void accessible() {
        super.accessible();
    }

    void notAccessible() {
        // Same signature than super method but not an override.
    }

}
```

## What did you see instead?

The notAccessible() method is considered an override and is annotated.
This results in compilation failure as notAccesible() can't override the supertype.

```java
package org.example.impl;

import org.example.A;

class B extends A {

    @Override
    public void accessible() {
        super.accessible();
    }

    @Override
    void notAccessible() {
        // Same signature than super method but not an override.
    }

}
```

Yes, we can contribute a fix to skip override annotations for these specific cases.

joteri:
## What's your motivation?

Fixes #3599 

Since it is an utility method that is invoked from different scenarios check if this change can affect any other behaviors.

@timtebeek

### Checklist
- It seems affected method is being invoked from another unit tests already so specific test was not added.
- Should add a specific unit test for `MissingOverrideAnnotation` recipe to verify this bugfix.
- no new files were added

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
@@ -407,6 +407,12 @@ public class TypeUtils {
                 }
             }
         }
+
+        // Check if access level is default then check if subclass package is the same from parent class
+        if (isDefaultAccessLevel(method.getFlags())) {
+            methodResult = methodResult.filter(m -> m.getDeclaringType().getPackageName().equals(dt.getPackageName()));
+        }
+
         return methodResult.filter(m -> !m.getFlags().contains(Flag.Private) && !m.getFlags().contains(Flag.Static));
     }
 
@@ -524,6 +530,10 @@ public class TypeUtils {
             return JavaType.Unknown.getInstance();
         }
         return t;
+    }
+
+    public static boolean isDefaultAccessLevel(final Set<Flag> flags) {
+        return !flags.contains(Flag.Public) && !flags.contains(Flag.Protected) && !flags.contains(Flag.Private);
     }
 
     static boolean deepEquals(@Nullable List<? extends JavaType> ts1, @Nullable List<? extends JavaType> ts2) {
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
@@ -532,7 +532,7 @@ public class TypeUtils {
         return t;
     }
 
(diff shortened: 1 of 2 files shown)


knutwannheden:
It would be very nice if we could get a unit test for this. Something like this:

```java
    @Test
    void packagePrivateMethodOverride() {
        rewriteRun(
          java("..."),
          java("...",
                assertThat(...);
            })
          )
        );
    }
```

rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -526,6 +532,10 @@ public static JavaType unknownIfNull(@Nullable JavaType t) {
         return t;
     }
 
+    public static boolean isDefaultAccessLevel(final Set<Flag> flags) {
```
timtebeek, line 535:
```suggestion
    private static boolean isDefaultAccessLevel(final Set<Flag> flags) {
```

timtebeek (approved):
Thanks a lot for this fix! I've reduced the visibility on the new method you added, just to limit the public API we'd need to support going forward.

timtebeek:
Thanks for getting this started @joteri ! I've added a test and pushed a small change such that there's less negations in the code. Hope you agree!

PR #2264, opened 2022-09-27
Improved YamlPrinter behavior for scalars with double-quoted values containing escape sequences

nmck257:
Handled all supported sequences specified in YAML spec: https://github.com/yaml/yaml-grammar/blob/master/yaml-spec-1.2.txt#L1914

...except for \x20 (space) and arbitrary unicode escapes (eg \x00, \u0000, \U00000000), because they were causing tricky interactions

Fixes #2154

PR #3604, opened 2023-10-05
Fix parsing of unicode escape sequences in YAML

knutwannheden:
A YAML scalar value can contain unicode escape sequences like `\u0051`. Currently, this throws off the parser.

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
@@ -400,34 +400,30 @@ class MappingTest implements RewriteTest {
       " '\\n' ",
       " '\n' ",
       " \n ",
-      " \"\\.\" ",
       " \"\\0\" ",
       " \"\\0\" ",
       " \"\\a\" ",
       " \"\\a\" ",
       " \"\\b\" ",
-      " \"\b\" ",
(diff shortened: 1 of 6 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
```diff
@@ -400,34 +400,30 @@ void mappingAnchor() {
       " '\\n' ",
       " '\n' ",
       " \n ",
-      " \"\\.\" ",
```
knutwannheden, line 403: Note: I've removed the escape sequences which are illegal according to the SnakeYAML parser.

knutwannheden:
@nmck257 In https://github.com/openrewrite/rewrite/pull/2264 you added logic to handle escape sequences. Unfortunately the tests inadvertently got disabled when we switched from Kotlin to Java. In this PR I have re-enabled the tests again, but I had to remove some of the cases, since the escape sequences appear to be invalid (at least according to the SnakeYAML parser).

I was wondering if you could have a look at this PR and make sure that the implementation still meets your requirements.

nmck257:
@knutwannheden yep, no concerns. I'll gladly trust the SnakeYaml parser; maybe some patch version over the past ~year helped it catch invalid edge cases which it did not notice before.

kunli2 (approved):

PR #3606, opened 2023-10-07
Add XSLT transformation recipe

ammachado:
## What's changed?

Allow ChangePluginConfiguration to perform XSLT transformation on plugin configuration. Also, added a new recipe for arbitrary XSLT transformations.

## What's your motivation?

Perform XSLT transformations on XML files and on Maven plugin configuration.

No

No

FindAndReplace works for simple refactoring, but this allow more complex scenarios.

## Any additional context

No

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/ChangePluginConfiguration.java
```diff
@@ -74,24 +91,45 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
                             )
                             .findAny();
                     if (maybePlugin.isPresent()) {
                         Xml.Tag plugin = maybePlugin.get();
-                        if (configuration == null) {
+                        if (configuration == null && xslt == null && xsltResource == null) {
                             plugins = filterChildren(plugins, plugin,
                                     child -> !(child instanceof Xml.Tag && "configuration".equals(((Xml.Tag) child).getName())));
-                        } else {
+                        } else if (configuration != null && xslt == null && xsltResource == null) {
```
timtebeek, line 103: I'm wondering if we should simplify the logic here such that it's easier to follow and extend. Right now when there's both a configuration and xslt supplied, we don't do anything. That's perhaps confusing to folks. Instead I propose we let configuration to precedence, and update the logic to reflect that. Would you agree to that?
```suggestion
                        } else if (configuration != null) {
```

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePluginConfiguration.java
```diff
@@ -74,24 +91,45 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
                             plugins = addOrUpdateChild(plugins, plugin,
                                     Xml.Tag.build("<configuration>\n" + configuration + "\n</configuration>"),
                                     getCursor().getParentOrThrow());
+                        } else if (configuration == null) {
```
timtebeek, line 107:
```suggestion
                        } else if (xslt != null || xsltResource != null) {
```

rewrite-xml/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ dependencies {
     api("com.fasterxml.jackson.core:jackson-annotations")
 
     compileOnly(project(":rewrite-test"))
+    compileOnly("org.slf4j:slf4j-api:1.7.+")
```
timtebeek, line 23: We don't often log from our recipes, as it's hard to uniformly present that output in the various ways recipes might run. Instead we often opt to add markers, which we might then in turn log, or present in some other way to the user, as appropriate. If at all possible I'd opt for markers, or no logging statements here too. What are your thoughts on that?

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
```diff
@@ -0,0 +1,126 @@
...
+    String filePattern;
+
+    @Override
+    public String getDisplayName() {
+        return "XSLT transformation";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Apply the specified XSLT transformation on.";
```
timtebeek, line 69: Did you intend to add more of a description here? At first glance it seemed like that to me.

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
```diff
@@ -0,0 +1,126 @@
...
+    @Option(displayName = "XSLT Configuration transformation classpath resource",
+            description = "Recipe transformation provided as a classpath resource.")
+    String xsltResource;
+
+    @Option(displayName = "File pattern",
+            description = "A glob expression that can be used to constrain which directories or source files should be searched. " +
+                          "Multiple patterns may be specified, separated by a semicolon `;`. " +
+                          "If multiple patterns are supplied any of the patterns matching will be interpreted as a match.",
+            example = "**/*.xml")
+    String filePattern;
```
timtebeek, line 60: I'm wondering if it would make sense to add an xpath expression option to this recipe, such that we might not have to add the xslt options to `ChangePluginConfiguration`. This recipe can then match particular tags within specific files, and be used in isolation. Otherwise I might foresee XSLT options being added to more recipes, whereas that would complicate maintenance a bit going forward. Curious to hear how you'd feel about that.

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
```diff
@@ -0,0 +1,126 @@
...
+        TreeVisitor<?, ExecutionContext> visitor = new XsltTransformationVisitor(loadResource(xslt, xsltResource));
+
+        @SuppressWarnings("unchecked")
+        TreeVisitor<?, ExecutionContext> check = Preconditions.or(Arrays.stream(filePattern.split(";"))
+                .map(HasSourcePath<ExecutionContext>::new)
+                .toArray(TreeVisitor[]::new));
+
+        visitor = Preconditions.check(check, visitor);
+
+        return visitor;
```
timtebeek, line 94:
```suggestion
        return Preconditions.check(check, visitor);
```

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
```diff
@@ -0,0 +1,126 @@
...
+        Validated<Object> validated = super.validate();
+
+        if (StringUtils.isBlank(xslt) && StringUtils.isBlank(xsltResource)) {
+            validated = validated.and(Validated.missing("xslt", xslt, "XSLT")
+                    .or(Validated.missing("xsltResource", xsltResource, "XSLT Classpath resource")));
+        }
+
+        if (!StringUtils.isBlank(xslt) && !StringUtils.isBlank(xsltResource)) {
+            validated = validated.and(Validated.missing("xslt", xslt, "XSLT")
+                    .and(Validated.missing("xsltResource", xsltResource, "XSLT Classpath resource")));
```
timtebeek, line 108: Do the validated method calls here not imply the calls to `StringUtils.isBlank` ? I'd be quite surprised if we need to check these explicitly, but I'm not in my IDE as I go through this first review pass.

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformationVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+            try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+                transformer.transform(text, new StreamResult(os));
+                if (log.isDebugEnabled()) {
+                    log.debug("Original document: {}", originalDocument);
+                    log.debug("New document: {}", os);
+                }
+                return document.withRoot(Xml.Tag.build(os.toString().replace("\r", "")));
+            }
+        } catch (IOException | TransformerException e) {
+            log.error(e.getMessage(), e);
```
timtebeek, line 76:
```suggestion
```

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformationVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+            Transformer transformer = TransformerFactory.newInstance().newTransformer(xsltSource);
+
+            String originalDocument = document.printAll();
+            Source text = new StreamSource(new ByteArrayInputStream(originalDocument.getBytes()));
+            try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+                transformer.transform(text, new StreamResult(os));
+                if (log.isDebugEnabled()) {
+                    log.debug("Original document: {}", originalDocument);
+                    log.debug("New document: {}", os);
+                }
```
timtebeek, line 72:
```suggestion
```

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformationVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.transform.stream.StreamSource;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Objects;
+
+import static org.openrewrite.Tree.randomId;
+
+@Slf4j
```
timtebeek, line 38:
```suggestion
```

timtebeek (commented):
Impressive work already; thanks for getting this started! Curious to hear what you intend to use if for. I've gone through for a first round of review comments, let me know how you feel about what I've raised so far!

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformationVisitor.java
ammachado: Do you have an example on how can I use Markers instead of slf4j?  I'm using slf4j here to get this information when I run maven with execution debug output on (`mvn -X`).

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
ammachado: The idea here was to not let the user specify both `xslt` and `xsltResource` at the same time.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePluginConfiguration.java
ammachado: Yes, that's a great idea. I had a clone of the ChangePluginConfiguration doing only the XSLT transformations on my branch.

rewrite-xml/build.gradle.kts
ammachado: Logging is simpler, requires less knowledge of the intrinsics of OR.  But I'm interested in learning a different way to do this type of communication / user feedback. Is there any good example from where I can learn?

rewrite-xml/src/main/java/org/openrewrite/xml/XsltTransformation.java
ammachado: The ChangePluginConfiguration, in my use case, its' the simplest transformation. The general case XsltTransfomation is being used in my setup to convert OSGI / Karaf blueprints to Spring Boot.  The XsltTransformation can do search & replace on XML files without converting them to PlainText files as wel..

rewrite-xml/build.gradle.kts
jkschneider: Showing how markers are used. Agree with @timtebeek we cannot have logging statements in core recipes: https://app.moderne.io/recipes/org.openrewrite.java.search.FindMethods?organizationId=T3BlblJld3JpdGU%3D&defaults=W3sibmFtZSI6Im1ldGhvZFBhdHRlcm4iLCJ2YWx1ZSI6Im9yZy5vcGVucmV3cml0ZS5tYXJrZXIuU2VhcmNoUmVzdWx0IGZvdW5kKC4uKSJ9LHsibmFtZSI6Im1hdGNoT3ZlcnJpZGVzIiwidmFsdWUiOiJUcnVlIn1d

ammachado: Your example uses DataTables which are not available on OpenRecipe yet, just on Moderne.io.

timtebeek: Resolving this discussion as the logging was removed, and data tables are now available:
https://docs.openrewrite.org/running-recipes/data-tables

timtebeek (approved):
Great work! Added some minor polishing but otherwise good to go if you ask me. I'll leave a little room for further feedback before I merge, but will merge before the next release.

And thank for the kind contribution, explaining your intended use, and patience in seeing this through.

PR #3609, opened 2023-10-09
RecipeSpec recipe caching

nmck257:
## What's changed?
`RecipeSpec` methods `recipeFromYaml`, `recipeFromResource`, and `recipeFromResources` now memoize Recipes to avoid repeated redundant Environment instantiation / classpath scanning

## What's your motivation?
Speed of unit tests; these methods (especially when calling `scanYamlResources`) can cost a few hundred milliseconds, which adds up fast if a large test class is using them.

- For a lighter touch, we could store only the `Environment` built with the `scanYamlResources` call in a lazy-init static field; this probably delivers ~most of the speed benefits with less code and less mem
- Using a memoization/caching library might be nicer, but, at the cost of another dependency
- Hashing together the relevant Strings (instead of joining them) might make for cheaper keys, but I struggled to implement that concisely since doing fancy stuff with varargs is a bit painful in java (no spread operator), esp 8 (fewer helper methods for instantiating lists, etc)
- I tried caching the customized `testClassSpec` objects used by `rewriteRun`, but that was impractical since `RewriteTest` is an interface and can't support instance variables
- A step further on the above, I even experimented with a static `WeakHashMap<Consumer<RecipeSpec>, RecipeSpec>` to cache `testClassSpec` objects based on the customizer (eg `this::defaults`) which would produce them, but I couldn't get keys with consistent memory references. And even if I *could* figure out clever lambda syntax to make that happen, it would depend on the test class being configured to as `TestInstanceLifecycle(PER_CLASS)`, which is non-default and has other ramifications

### Checklist


nmck257:
test failures look unrelated (something about the rewrite-gradle model failing to resolve the rewrite-core:8.8.0-SNAPSHOT dep)

sambsnyd:
Thanks @nmck257 !

PR #3610, opened 2023-10-09
Added recipe options as JSON to FindRecipes.

traceyyoshima:
Changes:

- FindRecipes will include the recipe options as JSON.

Format:
```json
[
  {
    "name": "optionName",
    "displayName": "...",
    "description": "...",
    "example": "...",
    "required": false,
    "valid": ["private", "protected", "package", "public"]
  }
]
```


knutwannheden:
To deal with escaping that may be necessary (see e.g. https://github.com/openrewrite/rewrite/blob/596cb39a71a1902b9f14f8bd9cda44c5069831c9/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java#L45), I changed the code to use Jackson to construct the JSON. I also changed the test to test for escaped quotes within a string value.

PR #3615, opened 2023-10-13
Expand assertion to print failure message to make it easier to debug …

dawngerpony:
…the failed test.

## What's changed?

A tiny PR to fix a problem in my environment where my local `.m2` directory contained a commons-lang directory with a POM but no JAR:

```shell
ls -l /Users/dawngerpony/.m2/repository/org/apache/commons/commons-lang3/3.11/commons-lang3-3.11.jar
ls: /Users/dawngerpony/.m2/repository/org/apache/commons/commons-lang3/3.11/commons-lang3-3.11.jar: No such file or directory
```

With the expanded assertion I was able to discover that this JAR was the problem and delete the directory, which made the test pass.

## What's your motivation?

I wanted to raise my first PR against the codebase and felt it should be a small one! This was just me trying to get my build to pass.

Side note/question: why does the state of my local `.m2` directory affect the outcome of what should be a unit test? Surely unit tests should not rely on a particular filesystem configuration?

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/utilities/MavenArtifactDownloaderTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/utilities/MavenArtifactDownloaderTest.java
@@ -32,6 +32,7 @@ import java.nio.file.Path;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 class MavenArtifactDownloaderTest {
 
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks again for making this change! I'd merged it while on the go, and didn't add a comment then, but just wanted to say we appreciate you pitching in here!

I understand you're surprised to see a test affected by the contents of `~/.m2`; We're starting to see more issues associated to that dependency, also with recipe runs, but it would be a big effort to fully disconnect, with potential downsides to the test duration too. We'll monitor this going forward and might come back to this if it remains problematic.

PR #3625, opened 2023-10-17
Add recipes to create new files

ammachado:
## What's changed?

Added recipes to create new files (Java classes and properties, XML and YAML files)

## What's your motivation?

Creating properties for a new profile on a Spring Boot project.

Some tests are still failing (for overwrites)

There is a recipe for creating generic text files, but files must be created using the correct `Tree` subclass in order to be modified using other recipes.

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/CreateJavaClass.java
```diff
@@ -0,0 +1,156 @@
...
+
+                    return newCu.withSourcePath(path);
+                }
+                return cu;
+            }
+        };
+    }
+
+    private @NonNull String getSourcePath() {
+        return String.format("src/%s/java/%s/%s.java", sourceRoot, packageName.replace('.', '/'), className);
```
shanman190, line 154: It may be better to just allow the recipe to specify the source path itself as an input option. As written presently, this would only be able to add a Java class file to the root project.

rewrite-java/src/main/java/org/openrewrite/java/CreateJavaClass.java
```diff
@@ -0,0 +1,182 @@
...
+            @JsonProperty("overwriteExisting") @Nullable Boolean overwriteExisting,
+            @JsonProperty("relativePath") @Nullable String relativePath
+    ) {
+        this.sourceRoot = sourceRoot;
+        this.packageName = packageName;
+        this.className = className;
+        this.classTemplate = classTemplate;
+        this.overwriteExisting = overwriteExisting;
+        this.relativePath = relativePath;
+    }
```
sambsnyd, line 107: This constructor is not necessary. An equivalent constructor would be generated by lombok via `@Value`. If you are getting errors relating to serialization when running tests in your IDE it is because you are missing the `-parameters` argument to the Java compiler.

Please also remove the similarly redundant constructors from the other classes.

rewrite-java/src/main/java/org/openrewrite/java/CreateJavaClass.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class CreateJavaClass extends ScanningRecipe<AtomicBoolean> {
```
sambsnyd, line 47: This is missing a way for the parser classpath to be specified, which will be necessary for most non-trivial java classes.

rewrite-java/src/main/java/org/openrewrite/java/JavaSourceRoot.java
```diff
@@ -0,0 +1,41 @@
...
+package org.openrewrite.java;
+
+import org.openrewrite.internal.lang.NonNull;
+import org.openrewrite.internal.lang.Nullable;
+
+enum JavaSourceRoot {
+
+    Main,
+
+    Test;
```
sambsnyd, line 25: Build systems such as Gradle allow arbitrary source sets to be defined, not just "main" and "test". Suggest using a plain `String` instead of creating this class.

rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateLanguageExtension.java
```diff
@@ -40,8 +40,8 @@ public abstract class JavaTemplateLanguageExtension {
     Space.Location loc;
     JavaCoordinates.Mode mode;
 
-    public JavaTemplateLanguageExtension(JavaTemplateParser templateParser, Substitutions substitutions,
-                                         String substitutedTemplate, JavaCoordinates coordinates) {
+    protected JavaTemplateLanguageExtension(JavaTemplateParser templateParser, Substitutions substitutions,
```
sambsnyd, line 43: Do not reduce this visibility. This is used in parsers for other languages not yet merged into the main rewrite repository.

rewrite-properties/src/main/java/org/openrewrite/properties/CreatePropertiesFile.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public @NonNull String getDisplayName() {
+        return "Create YAML file";
+    }
+
+    @Override
+    public @NonNull String getDescription() {
+        return "Create a new YAML file.";
+    }
```
sambsnyd, line 72: Copy and paste error, these should be "properties"

rewrite-properties/src/main/java/org/openrewrite/properties/CreatePropertiesFile.java
```diff
@@ -0,0 +1,107 @@
...
+import java.nio.file.Paths;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class CreatePropertiesFile extends ScanningRecipe<AtomicBoolean> {
```
sambsnyd, line 42: Seems to be missing a `String` field to specify the initial contents of the properties file. Could optionally be set with a `Properties` object.

rewrite-xml/src/main/java/org/openrewrite/xml/CreateXmlFile.java
```diff
@@ -0,0 +1,108 @@
...
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class CreateXmlFile extends ScanningRecipe<AtomicBoolean> {
+
+    @Option(displayName = "Relative file path",
+            description = "File path of new file.",
+            example = "foo/bar/baz.yaml")
```
sambsnyd, line 46: example path shows yaml extension

sambsnyd (commented):
Thanks for your contribution, I think these will be useful recipes to have.

rewrite-java/src/main/java/org/openrewrite/java/CreateJavaClass.java
ammachado: How that can be done? I tried an implementation here, but the scanning recipe was passing along `SourceSpec`s with parse errors.

I also tried to enable logging for compilation errors (the parser supports it), but 1) there is no logging implementation on `rewrite-java`; and 2) I tried adding `testRuntimeOnly("org.slf4j:slf4j-simple:1.7+")` on `rewrite-java`'s `build.gradle.kts`, and it also didn't work, because archunit is [pulling slf4j-api 2.0.3](https://central.sonatype.com/artifact/com.tngtech.archunit/archunit/1.0.1/dependencies)and SLF4J 1.x bindings don't work on SLF4J 2.x.  Any advice?

Another question, how do I enable `TypeValidation` on the tests? Adding `spec.typeValidationOptions(TypeValidation.builder().build())` didn't work.

timtebeek:
Perhaps we can pull `rewrite-java/src/main/java/org/openrewrite/java/CreateJavaClass.java` out into a separate follow up PR to get the other recipes in already? Should help not to hold up the other improvements given that there seems to be some open questions specific to the Java variant still.

ammachado:
I refactored the `CreateJavaClass` to create empty java classes instead, for now we can use other existing recipes to add implementation as required.

timtebeek (approved):
Thanks a lot @ammachado ! I've pushed up a few small changes and fixes, hope you agree!

PR #3634, opened 2023-10-19
Fixed property usage pattern matching search results

joanvr:
## What's changed?
Previously, all property usages were printed the actual value of the property, not only the ones matching the propertyPattern. Now we also apply the propertyPattern to find usages.

## What's your motivation?
Another recipe is using this recipe to find java.version property, and it was always returning results if there were any property being used in the pom.

@timtebeek @knutwannheden @pstreef 

I've seen that there is a find method at the bottom, that only returns property definitions, but not property usages, that has duplicated code. To avoid propagating a change in behavior down, I've left the method as it is, but I found kinda inconsistent to have this code partially duplicated and with different behavior than in the actual recipe.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
@@ -76,10 +76,14 @@ public class FindProperties extends Recipe {
         };
     }
 
+    /**
+     *
+     * @param xml The xml document of the pom.xml
+     * @param propertyPattern Regular expression pattern used to match property tag names
+     * @return Set of Maven project property tags that matches the {@code propertyPattern} within a pom.xml
+     */
     public static Set<Xml.Tag> find(Xml.Document xml, String propertyPattern) {
-        Pattern propertyMatcher = Pattern.compile(propertyPattern
-                .replace(".", "\\.")
-                .replace("*", ".*"));
+        Pattern propertyMatcher = Pattern.compile(propertyPattern);
         Set<Xml.Tag> found = new HashSet<>();
         new MavenVisitor<Set<Xml.Tag>>() {
             @Override
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
@@ -61,11 +61,11 @@ public class FindProperties extends Recipe {
             @Override
             public Xml visitTag(Xml.Tag tag, ExecutionContext context) {
                 Xml.Tag t = (Xml.Tag) super.visitTag(tag, context);
-                if (isPropertyTag() && propertyMatcher.matcher(tag.getName()).matches()) {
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
```diff
@@ -36,9 +36,7 @@
 @Value
 public class FindProperties extends Recipe {
 
-    @Option(displayName = "Property pattern",
-            description = "Regular expression pattern used to match property tag names.",
-            example = "guava*")
+    @Option(displayName = "Property pattern", description = "Regular expression pattern used to match property tag names.", example = "guava*")
```
timtebeek, line 39: I know this isn't part of your changes, but I find the description example value here a little confusing.
- We say the pattern is a regular expression, yet below we replace `.` with `\\.` and replace `*` with `.*`. Would that not make this a glob pattern instead?
- If the value _is_ a regular expression, then `guava*` doesn't make much sense; that'd match `guav`, `guava` & `guavaaaa` for example, were it not for the above special handling.
- How does the algorithm below handle proper regular expression patterns like `\\.` or `.*`? it seems we'd duplicate the replacements, which is at best confusing, at worst wrong.

Perhaps we should reevaluate how `propertyPattern` is used, and adjust our description here accordingly.

rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
```diff
@@ -55,8 +55,11 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        Pattern propertyMatcher = Pattern.compile(propertyPattern.replace(".", "\\.")
+        Pattern propertyMatcher = Pattern.compile(propertyPattern
+                .replace(".", "\\.")
                 .replace("*", ".*"));
```
knutwannheden, line 60: It looks like we should modify the description of this option, because it really isn't a regular expression at all. It is more like a glob. Then if someone asks for regex support I guess we have to add a boolean option to control whether the pattern is a glob or regex.

rewrite-maven/src/main/java/org/openrewrite/maven/search/FindProperties.java
```diff
@@ -55,18 +55,18 @@ public String getDescription() {
...
                 Xml.Tag t = (Xml.Tag) super.visitTag(tag, context);
-                if (isPropertyTag() && propertyMatcher.matcher(tag.getName()).matches()) {
+                if (isPropertyTag() && propertyMatcher.matcher(t.getName()).matches()) {
                     t = SearchResult.found(t);
                 }
 
-                Optional<String> value = tag.getValue();
-                if (t.getContent() != null && value.isPresent() && value.get().contains("${")) {
+                Optional<String> value = t.getValue();
+                if (value.isPresent() && propertyUsageMatcher.matcher(value.get()).matches()) {
```
joanvr, line 69: No need to check for `t.getContent != null` since it's already checked in `t.getValue()` returning an `Optional.empty()`

knutwannheden (approved):

Issue #3626, opened 2023-10-18
ResolvedDependency.hashCode() causes StackOverflowError with circular dependencies


edudar:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.7.4
- Gradle plugin v6.3.19
- rewrite-maven v8.7.4

`org.openrewrite.maven.tree.ResolvedDependency` uses Lombok-generated `hashCode` that calls for hash codes of transitive dependencies. This leads to StackOverflowError when dependencies are "circular". 

For example, [io/grpc/grpc-core/1.58.0](https://repo1.maven.org/maven2/io/grpc/grpc-core/1.58.0/grpc-core-1.58.0.pom) depends on `grpc-util` and [io/grpc/grpc-util/1.58.0](https://repo1.maven.org/maven2/io/grpc/grpc-util/1.58.0/grpc-util-1.58.0.pom) depends on `grpc-core`. The scopes are different (`compile` vs. `runtime`), so Gradle resolves this just fine.

knutwannheden:
@edudar Thanks for reporting this issue. Can you tell us exactly where you encountered this `StackOverflowError`?

edudar:
Nothing special to report here; I was simply running `gradle rewriteRun` on the project. The active recipe does not matter as this error happens on any of them. I first encountered it with `guava` and managed to resolve it on my own to then stumble upon `grpc` case. We don't define these explicitly and get them transitively via one of `gcp` dependencies for logging.

knutwannheden:
Makes sense. If you could tell me what recipe you ran, I may be able to reproduce the problem.

edudar:
`org.openrewrite.java.RemoveObjectsIsNull`

knutwannheden:
Thanks. I was expecting a Maven recipe here. But I will try to reproduce using this.

edudar:
I believe the maven part runs regardless of whether the recipe is maven-related or not. I tried multiple recipes: as simple as `RemoveObjectsIsNull` to junit migrations to spring boot migrations. They all were failing with the same error pointing to rewrite-maven.

knutwannheden:
We detected a few more situations where these dependency cycles were problematic. Should be fixed now with #3678

PR #3635, opened 2023-10-19
Add missing recurring dependencies to `ResolvedDependency`

knutwannheden:
When `ResolvedPom#resolveDependencies()` processes the dependencies, it skips any dependencies it has already seen. But it must still make sure to add them to the list of dependencies of the artifact that requested it.

Also exclude `ResolvedDependency#dependencies` from `hashCode()` since dependencies can be circular (although not within one scope).

Fixes: #3626


timtebeek (approved):

sambsnyd:
I would rather not have cycles unless there's a strong argument we need them

knutwannheden:
Closing this for now.

PR #3636, opened 2023-10-19
Ignore local pom if the jar file does not exist.

pstreef:


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -516,6 +516,14 @@ public Pom download(GroupArtifactVersion gav,
                             RawPom rawPom = RawPom.parse(fis, Objects.equals(versionMaybeDatedSnapshot, gav.getVersion()) ? null : versionMaybeDatedSnapshot);
                             Pom pom = rawPom.toPom(inputPath, repo).withGav(resolvedGav);
 
+                            if("jar".equals(pom.getPackaging())){
+                                Path jar = f.toPath().getParent().resolve(gav.getArtifactId() + '-' + versionMaybeDatedSnapshot + ".jar");
+                                if (!Files.exists(jar)) {
+                                    // This is a corrupt dependency.
```
pstreef, line 522: It's not really corrupted. There are valid usecases for the jar to be missing (like output of mvn dependencyTree)

knutwannheden, line 522: As we've seen, this typically isn't the case. It just means that the jar has not been downloaded.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -516,6 +516,14 @@ public Pom download(GroupArtifactVersion gav,
                             RawPom rawPom = RawPom.parse(fis, Objects.equals(versionMaybeDatedSnapshot, gav.getVersion()) ? null : versionMaybeDatedSnapshot);
                             Pom pom = rawPom.toPom(inputPath, repo).withGav(resolvedGav);
 
+                            if("jar".equals(pom.getPackaging())){
+                                Path jar = f.toPath().getParent().resolve(gav.getArtifactId() + '-' + versionMaybeDatedSnapshot + ".jar");
```
knutwannheden, line 520:
```suggestion
                                Path jar = f.toPath().resolveSibling(gav.getArtifactId() + '-' + versionMaybeDatedSnapshot + ".jar");
```

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -516,6 +516,14 @@ public Pom download(GroupArtifactVersion gav,
                             RawPom rawPom = RawPom.parse(fis, Objects.equals(versionMaybeDatedSnapshot, gav.getVersion()) ? null : versionMaybeDatedSnapshot);
                             Pom pom = rawPom.toPom(inputPath, repo).withGav(resolvedGav);
 
+                            if("jar".equals(pom.getPackaging())){
```
knutwannheden, line 519:
```suggestion
                            if (pom.getPackaging() == null || "jar".equals(pom.getPackaging())) {
```

knutwannheden: Note that `jar` is the default packaging. So we should also check that the jar exists, when the value is `null`.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
pstreef:
```suggestion
                                    // The jar has not been downloaded, making this dependency unusable.
```

knutwannheden (approved):

joanvr:
Added check for file size == 0, since I've also found some of this broken artifacts on my local maven repository that were causing issues.
Also fixed the test that was failing and added a couple more that tests those broken artifacts

joanvr (approved):

PR #3638, opened 2023-10-19
Remove unused annotation parameter imports

ammachado:
## What's changed?

This change the `RemoveAnnotation` recipe/visitor to remove unused annotation parameter imports

### Checklist


knutwannheden:
I suspect this might not work when an annotation attribute has a `Class` type or a primitive type, where the expression references a constant from another type.

Also, there could be static imports to consider. I think the implementation should visit the expression trees and also collect field types to remove imports for.

sambsnyd:
Thanks @ammachado !

PR #3643, opened 2023-10-23
Proof of concept for preconditions on declarative recipes

sambsnyd:
In rewrite 7 "applicability tests" allowed recipe authors to conditionally execute a visitor based on the result of another visitor. This was a `Recipe`-level concept, and it substantially complicated `RecipeScheduler`. 

In rewrite 8 we renamed applicability tests to preconditions, and declarative recipes lost the ability to declare them. This greatly simplified recipe scheduling and allowed us to implement `ScanningRecipe`. 

This proposal brings back the cherished and long-lost ability for declarative recipes to have preconditions, without any impact to `Recipe` or `RecipeScheduler`. Inside `DeclarativeRecipe` I gather up all the preconditions and put one recipe at the beginning of the recipe list to evaluate them, I call this synthetic recipe the "bellwether". Then every subsequent recipe in the list is decorated with a precondition which looks solely at the result of the bellwether. This ensures that the preconditions are evaluated exactly once and at the appropriate time, which is important for both correctness and performance.

There are various ways this is incomplete, pending agreement that this is how we want to proceed:
- `ScanningRecipe` preconditions are not handled. I imagine we would initially disallow them, as supporting them is complex and not necessary for the things we currently use preconditions for.
- `YamlResourceLoader` logic to deserialize yaml preconditions not yet implemented 
- `PreconditionDecoratedRecipe` and its scanning counterpart 
- We could optionally support a separate set of preconditions that apply only to the scanning portion of scanning recipes. Or apply the one set of preconditions both to scanning and visiting. Not entirely clear to me what our desired behavior here is.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
@@ -20,12 +20,11 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
+import lombok.experimental.NonFinal;
 import org.intellij.lang.annotations.Language;
-import org.openrewrite.Contributor;
-import org.openrewrite.Maintainer;
-import org.openrewrite.Recipe;
-import org.openrewrite.Validated;
+import org.openrewrite.*;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.marker.SearchResult;
 
 import java.net.URI;
 import java.time.Duration;
@@ -67,6 +66,11 @@ public class DeclarativeRecipe extends Recipe {
 
     private final List<Recipe> uninitializedRecipes = new ArrayList<>();
     private final List<Recipe> recipeList = new ArrayList<>();
+    private final List<Recipe> preconditions = new ArrayList<>();
+
+    public void addPrecondition(Recipe recipe) {
+        preconditions.add(recipe);
+    }
 
     @JsonIgnore
     private Validated<Object> validation = Validated.test("initialization",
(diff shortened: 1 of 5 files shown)


nmck257:
This approach makes sense to me so far. Regarding the open questions, it's helping me to think back in terms of rewrite 7 (hopefully I'm remembering accurately):
- I agree that not supporting `ScanningRecipe`s as preconditions in the first implementation makes sense, and would probably throw an exception explicitly if that happens at runtime. I think that any use case where you'd want to use a `ScanningRecipe` as a declarative precondition would be roughly congruous to what was previously called a (non-single-source) applicability test (ie, "I need to check a different file to know whether I should change this file"), and from prior convos, I think we're of the opinion that those were rarely useful
  - This is further evidenced by that a `ScanningRecipe` in rewrite 8 would have implemented `visit(List<SourceFile>)` and *not* `getVisitor` in rewrite 7, therefore it would have been incompatible for use as an applicability test
- For how to apply (declarative) preconditions to `ScanningRecipe`s, my hunch would be for the default behavior to only apply to the Visitor, not the Scanner. If I'm consuming recipes in a declarative context, I'm not super-aware of whether those recipes are scanning or traditional; I'm just focused on the desired output (ie the Visitor's work), and that's likely what I'd want to filter with a precondition. The Scanner in a `ScanningRecipe` feels like an implementation detail which should be encapsulated away from a declarative recipe author; reusing the precondition from the visitor feels like it'll lead to obscure bugs which would be super non-obvious to the user
  - Another observation here is that if preconditions came from single-source applicability tests and scanning recipes came from `visit(List<SourceFile>)` recipes, then those two things would never have intersected in rewrite 7. So, applying preconditions to scanning recipes is indeed a new/novel question
  - That said, I'm not sure how I'd envision selecting the non-default behavior in the Yaml (eg is it a separate field on the recipe declaration for `scanningPrecondition`, or does the new `precondition` field have some `scanningRecipeStrategy` child field)

knutwannheden:
@nmck257 I think what you are saying regarding the scanning recipes makes sense, ad some scanning recipes will likely have to gather information from sources which the user may not want to touch and would thus filter out using preconditions.

sambsnyd:
I've gone back and forth on whether or not `ScanningRecipe` should have preconditions apply to their scanner.

We use `ScanningRecipe` in several places for data table production, where `getScanner()` and `generate()` do all the work and a precondition doesn't mean anything. 

One thing we want to use preconditions for is skipping whole repositories not of interest to a particular search/data table aggregation. If preconditions apply only to the visitor that could end up with a bunch of pointless scanning / data table entries we'd want filtered out.

I also considered having separate preconditions just for scanners, and maybe we will add that later. But for now I didn't want to deal with the complexity of implementing and explaining the nuances.

ia3andy:
@sambsnyd you still plan to do YamlResourceLoader right?

Issue #3651, opened 2023-10-31
Intersection type casts are not supported


knutwannheden:
As of Java 8 a type cast expression can as its target type declare an intersection type as in the following example:

```java
    @Test
    void intersectionCast() {
        rewriteRun(
          java(
            """
              import java.io.Serializable;
              import java.util.function.BiFunction;
              
              class Test {
                  Serializable s = (Serializable & BiFunction<Integer, Integer, Integer>) Integer::sum;
              }
              """
          )
        );
    }
```

Our current Java parsers don't support this and will result in a parse-to-print failure. The cast expression ends up getting serialized as `( & BiFunction<Integer, Integer, Integer>) Integer::sum` (note the missing `Serializable`).

knutwannheden:
To fix this we will both require a new `J.IntersectionType` as well as a new `JavaType.IntersectionType`.

PR #3652, opened 2023-10-31
Support intersection type casts

knutwannheden:
Supported since Java 1.8: https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.16

Fixes: #3651


sambsnyd (approved):
Looks reasonable to me

sambsnyd:
I don't think intersection can extend `FullyQualified` since it represents multiple types

knutwannheden:
@sambsnyd In the Java compiler's model the intersection type does extend ClassType. I believe it has to do with the fact that an intersection type can only have one class and the others need to be interfaces. Alternatively all interfaces and Object as the implicit class. That way an intersection is pretty much the same as a regular class type and the fully qualified name would be the class name. But I am also unsure about this and wonder what problems we might get if we do this or don't do it...

kunli2 (approved):

PR #3656, opened 2023-11-02
Fixed error when no file matcher was present

motu55:
## What's changed?
the noop treevisitor was replaced with a hassource visitor which always matches

## What's your motivation?
the following line
(https://github.com/openrewrite/rewrite/blob/d5ed2c5a1cc64b412afcee87c32ba6ab1b344758/rewrite-core/src/main/java/org/openrewrite/Preconditions.java#L41)
would always result to false on an xml file
`!(tree instanceof SourceFile)` would be false in case of an xml.document which inherits from source file
  `check.visit(tree, ctx) != tree` is false since check is the noop treevisitor which returns the tree
  since both are false the whole check is false

## Any additional context
also added a test for it

### Checklist
- [ ] I've added the license header to any new files through `./gradlew licenseFormat` (no new files)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files (no since it restructured all the files so better let it be)


timtebeek (approved):
Thanks for the detailed report & fix @motu55!

PR #3660, opened 2023-11-03
update method type on DeleteMethodArgument

joanvr:
## What's changed?
Updating method type list of arguments to avoid next calls to the recipe to keep method matching

## What's your motivation?
Right now this recipe is not idempotent, thus we need to specify in the tests that we only want one cycle. This is not very convenient when using this recipe from other recipes, and also leaves the method in an inconsistent state.

@timtebeek 

## Any additional context
I had to change a test, that was relying specifically on the fact that the method type is not updated and that subsequent calls to the recipe will still match and delete the parameter. I do not know if this is also used in recipes that are using this one.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/DeleteMethodArgumentTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/DeleteMethodArgumentTest.java
@@ -37,8 +37,7 @@ class DeleteMethodArgumentTest implements RewriteTest {
     @Test
     void deleteMiddleArgumentDeclarative() {
         rewriteRun(
-          spec -> spec.recipe(new DeleteMethodArgument("B foo(int, int, int)", 1))
-            .cycles(1).expectedCyclesThatMakeChanges(1),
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/DeleteMethodArgumentTest.java
```diff
@@ -65,7 +63,7 @@ void deleteArgumentsConsecutively() {
         rewriteRun(
           spec -> spec.recipes(
             new DeleteMethodArgument("B foo(int, int, int)", 1),
-            new DeleteMethodArgument("B foo(int, int, int)", 1)
+            new DeleteMethodArgument("B foo(int, int)", 1)
```
joanvr, line 66: Since we are updating the methodType, now it is required to match the new resulting "signature"

knutwannheden: Here is one case where that was required: https://github.com/openrewrite/rewrite-testing-frameworks/commit/47ccd370247f1171fa9df005da8a9a3342d19f3f

Issue #3661, opened 2023-11-04
UseStaticImport replaces non-static calls


Bananeweizen:
UseStaticImport replaces everything that matches the method pattern. Even non static methods. That's a problem if the method pattern is a wildcard for a complete class like `java.util.function.Predicate *(..)`, where the class contains both static and non-static methods.

## What is the smallest, simplest way to reproduce the problem?

```java
import java.util.function.Predicate;

public class Reproducer {
    void reproduce() {
        final Predicate<Object> predicate = x -> false;
        final var okay = Predicate.not(predicate);
        final boolean error = predicate.test(null);
    }
}

```

## What did you expect to see?
Only the second line of the method being modified, but not the third.

## What did you see instead?

```java
import java.util.function.Predicate;

import static java.util.function.Predicate.not;
import static java.util.function.Predicate.test; // compiler error

public class Reproducer {
    void reproduce() {
        final Predicate<Object> predicate = x -> false;
        final var okay = not(predicate);
        final boolean error = test(null); // compiler error
    }
}
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Yes. Looks like fix is just one more condition in the if.

timtebeek:
Thanks for reporting this issue, and the offer to help fix it as well! Indeed looks well to do, especially given your background to the project already. Do let us know if you need any help, or can't get to it in the near future!

PR #3662, opened 2023-11-04
Restrict UseStaticImport to static methods (#3661)

Bananeweizen:
* add tests for good and bad cases
* fix issue by checking method to be static
* verified that the bad case fails without the fix and passes with the fix

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
@@ -20,6 +20,7 @@ import lombok.experimental.FieldDefaults;
 import org.openrewrite.*;
 import org.openrewrite.java.search.DeclaresMethod;
 import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Flag;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.Javadoc;
@@ -73,6 +74,10 @@ public class UseStaticImport extends Recipe {
                     return m;
                 }
                 if (m.getMethodType() != null) {
+                    if (!m.getMethodType().hasFlags(Flag.Static)) {
+                        return m;
+                    }
+
                     JavaType.FullyQualified receiverType = m.getMethodType().getDeclaringType();
                     maybeRemoveImport(receiverType);
 
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
@@ -146,79 +146,6 @@ class UseStaticImportTest implements RewriteTest {
         );
     }
 
-    @Test
-    void methodInvocationsHavingNullSelect() {
-        rewriteRun(
-          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
-              @Override
-              public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
-                  return super.visitClassDeclaration(classDecl, ctx).withExtends(null);
-              }
-
-              @Override
-              public J.Import visitImport(J.Import _import, ExecutionContext executionContext) {
-                  //noinspection ConstantConditions
-                  return null;
-              }
-
-              @Override
-              public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
-                  return super.visitMethodInvocation(method, executionContext)
-                    .withDeclaringType(JavaType.ShallowClass.build("asserts.Assert"));
-              }
-
-              @Override
-              public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext executionContext) {
-                  doAfterVisit(new UseStaticImport("asserts.Assert assert*(..)").getVisitor());
-                  return super.visitCompilationUnit(cu, executionContext);
-              }
-          })).cycles(2).expectedCyclesThatMakeChanges(2),
-          java(
-            """
-              package asserts;
-
-              public class Assert {
-                  public static void assertTrue(boolean b) {}
-                  public static void assertEquals(int m, int n) {}
-              }
-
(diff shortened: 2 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
```diff
@@ -58,7 +59,7 @@ private class UseStaticImportVisitor extends JavaIsoVisitor<ExecutionContext> {
         public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
             MethodMatcher methodMatcher = new MethodMatcher(methodPattern);
             J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
-            if (methodMatcher.matches(m)) {
+            if (methodMatcher.matches(m) && method.getMethodType().hasFlags(Flag.Static)) {
```
knutwannheden, line 62: Would be better to place this in the next condition, where we know the method type to be non-null. Also, it should be retrieved from `m` rather than `method`, so that it also supports nested calls.

rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
```diff
@@ -58,7 +59,7 @@ private class UseStaticImportVisitor extends JavaIsoVisitor<ExecutionContext> {
...
                 if (m.getMethodType() != null) {
```
timtebeek, line 63:
```suggestion
            if (methodMatcher.matches(m)) {
                if (m.getMethodType() != null && m.getMethodType().hasFlags(Flag.Static)) {
```

timtebeek: Apply suggestions from https://github.com/openrewrite/rewrite/pull/3662/files#r1382445827

rewrite-java/src/test/java/org/openrewrite/java/UseStaticImportTest.java
timtebeek, line 1: I'm going to move these tests to https://github.com/openrewrite/rewrite/blob/e5c77d5afe17070474d7b5cf6e728600c5f776af/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java#L30

rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
timtebeek: Turns out it's a little more complicated when we look at 
https://github.com/openrewrite/rewrite/blob/e5c77d5afe17070474d7b5cf6e728600c5f776af/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java#L76-L78

rewrite-java/src/test/java/org/openrewrite/java/UseStaticImportTest.java
Bananeweizen: Sure. I must admit I'm not really certain how things are structured in the project. I didn't discover the tests in that subproject before. You may want to incorporate the suggestion of Knut when moving.

rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
timtebeek: There's an unusual test which fails when we add a limitation to static methods in this recipe https://github.com/openrewrite/rewrite/blob/e5c77d5afe17070474d7b5cf6e728600c5f776af/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java#L76-L78

timtebeek:
Capturing the failure here to save folks a click
```
UseStaticImportTest > methodInvocationsHavingNullSelect() FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "test/Test.java":
    diff --git a/test/Test.java b/test/Test.java
    index 462a2c3..259aa9b 100644
    --- a/test/Test.java
    +++ b/test/Test.java
    @@ -1,8 +1,5 @@ 
     package test;
 
    -import static asserts.Assert.assertEquals;
    -import static asserts.Assert.assertTrue;
    -
     class Test {
         void test() {
             assertTrue(true);
    ] 
    expected: 
      "package test;
  
      import static asserts.Assert.assertEquals;
      import static asserts.Assert.assertTrue;
  
      class Test {
          void test() {
              assertTrue(true);
              assertEquals(1, 2);
          }
      }"
     but was: 
      "package test;
  
      class Test {
          void test() {
              assertTrue(true);
              assertEquals(1, 2);
          }
      }"
```

Bananeweizen:
I rebased, squashed, and investigated the failing test `methodInvocationsHavingNullSelect()`. To me that looks like a bug in some other area of the code. `Assert.assert(boolean)` is marked as static initially (flags=9). But when reaching the breakpoint in this recipe, the method is no longer marked static (flags=1).  I believe that https://github.com/openrewrite/rewrite/blob/eccdbd4e229b9fc057dd5caaa8b7d9f088a4462d/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java#L167 is the culprit. It somehow recreates the method, but seems to miss copying the flag. Since I have no clue of how to work with the generated (private) with-method from lombok in IntelliJ, I'm not sure how to fix that.

timtebeek:
I'm looking at that test now and I'm inclined to say the test is quite odd; it eagerly replaces calls to instance methods with calls to static methods called within those instance methods, based on the `.withExtends(null)` and `.withDeclaringType(JavaType.ShallowClass.build("asserts.Assert"))`. That's not what the `UseStaticImport` recipe should be about, and the check on methods being static here correctly makes that fail in my opinion. Would you agree there @knutwannheden ?

https://github.com/openrewrite/rewrite/blob/eccdbd4e229b9fc057dd5caaa8b7d9f088a4462d/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java#L150-L220

knutwannheden:
That test indeed looks rather peculiar. It also looks like it is very old.

timtebeek:
Then let's remove it; if it was necessary even after this change I suspect something downstream will pick that up.

timtebeek:
Thanks a lot yet again @Bananeweizen !

PR #3665, opened 2023-11-06
Upgrade versions defined in `gradle.properties`

timtebeek:
As reported through https://rewriteoss.slack.com/archives/C01A843MWG5/p1699263596812209
Test provided by Matt Johnson there, who was not able to open this PR himself due to policy issues.

## What's changed?
Add a test to replicate issue reported in Slack

## What's your motivation?
We want to support upgrading version properties in `gradle.properties`.
A test is a first step towards that.

Next we likely need to update the recipe implementation to also look at `gradle.properties` files in the same project.

Any hints towards the implementation?
Should we also look at parent projects for gradle properties?

Not yet.


timtebeek:
_Adding replies made by @shanman190 on Slack here as background, and to ensure they're kept beyond the Slack history cut off_

So I'll chime in since I've been pretty active in the Gradle space. So to expand upon what Tim has stated already, Gradle builds are full and complete programs. Prior to OpenRewrite 8.x there wasn't a good way to collect data from one file, then enact change in a completely different file. This is where the ScanningRecipe comes in. So when the upgrade recipes were mostly defined, they had to work upon just the build.gradle source file itself which meant not being able to lookup or alter values from any other place where you could define a version in Gradle.

Just to note a few places where you could hide a version:

1. On the dependency declaration in build.gradle(.kts) (testImplementation("junit:junit:4.12"))
2. In a variable declaration in the same source file (def springBootVersion = "2.7.0")
3. In an extension variable in the same source file (ext.springBootVersion = "2.7.0")
4. In a gradle.properties file in the same project (springBootVersion=2.7.0)
5. In an extension variable in any parent module
6. In a gradle.properties file in any parent module
7. In the Gradle Version Catalog in settings.gradle(.kts) (https://docs.gradle.org/current/userguide/platforms.html#sub:version-catalog-declaration)
8. In the Gradle Version Catalog in libs.versions.toml (https://docs.gradle.org/current/userguide/platforms.html#sub:conventional-dependencies-toml)
9. In an internal Gradle Platform BOM
10. In an external BOM
11. Defined in an included Gradle build script (apply from: "./versions.gradle")
12. Defined in a precompiled buildSrc plugin
13. Defined in a binary plugin

And being honest, I'm still sure that I've missed places. 😅 

The recipe as it's setup today, tries to handle cases 1, 2, and 3 since those only effect the current source file. Cases 4-9 and 11-12 could now likely also be handled by taking advantage of the ScanningRecipe mechanism. I personally haven't seen a lot of cases, but there certainly are some where what is presently available doesn't handle the majority of cases.

I know the Micronaut group as a case in point, put in their migration recipes the UpdateGradleProperty invocation since from their [Micronaut Launch](https://micronaut.io/launch) application defines the version in gradle.properties. This would be where via the OSS build plugins the rewrite.yml file can come into play to enable you to define multiple recipes to enact composite behaviors specific to your projects. So to use your example today, I'd probably solve that by including the Spring Boot 3.1 migration recipe AND then an UpdateGradleProperty to also handle the custom properties as an aside step. Definitely not quite as nice as it being fully autonomous, but it gets the job done.

To illustrate for a Spring Boot project, most applications tend to take the versions as defined within the Spring Boot Gradle Plugin (binary plugin that is version aligned as it's published out of Spring Boot itself).
```groovy
plugins {
  id "org.springframework.boot" version "2.7.0"
  id "io.spring.dependency-management" version "1.1.0"
}

[ ... omitted for brevit ...]

dependencies {
    implementation("org.springframework.boot:spring-boot-starter-web")
}
```
OR
```groovy
plugins {
  id "org.springframework.boot" version "2.7.0"
}

[ ... omitted for brevit ...]

dependencies {
    implementation(platform(org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES))
    implementation("org.springframework.boot:spring-boot-starter-web")
}
```
In all of the Spring Boot migration recipes, they are setup such that it also effects the Gradle plugin definitions like above and updates those versions correctly. This helps to ensure that you only have the version defined in one place while you're also taking Spring's recommendations for it's paired version catalog.

NOTE: In subproject builds, Gradle doesn't want for you to specify the version of plugins, so those have the form of id "org.springframework.boot" (note no version method). This furthers the case of only having the version defined once except also covering some of the cases above in a different way conveniently.

ajohnsonz:
Summarising the suggestions from the Slack thread with @timtebeek and @shanman190 for posterity:

The `GradleProject` marker is only holding onto the requested/resolved literal values. Unfortunately, it doesn't know the relationship in terms of the properties.

Initially, I'm thinking that we'd have to gather all properties by source path AND all matching dependency declarations GA+variable name might be the easiest, then during the edit phase....
- If we are on a `gradle.properties` file, use the variable key to lookup the GA, resolve the newVersion, then conditionally update the property value.
- If we are on a `build.gradle(.kts)` file, no changes here.

To gather properties by source path and all matching dependency declarations:
- have the existing recipe extend `ScanningRecipe` (instead of `Recipe`)
- with a custom accumulator object that holds those values as fields, returned from `getInitialValue`
- then run two visitors in `getScanner`, one `PropertiesVisitor` and one `GroovyVisitor`, and have those fill your accumulator object
- then in `getVisitor` run a `PropertiesVisitor` (as well as the existing `GroovyVisitor`) that correlates the values in your custom accumulator, to make the required changes on `gradle.properties`

Here's an existing `ScanningRecipe` which scans and edits multiple file types that you can base that side off of, along with the associated test class:
- https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
- https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java

timtebeek:
Continued in https://github.com/openrewrite/rewrite/pull/3901

PR #3669, opened 2023-11-07
fix collision with nested class references

joanvr:
## What's changed?
Checking for the root identifier on fieldAccess to add to usedTypes when referencing a nested type.

## What's your motivation?
https://github.com/moderneinc/support-app/issues/20

@knutwannheden 

### Checklist


knutwannheden (approved):
LGTM. Thanks for fixing this!

PR #3676, opened 2023-11-08
Streamline local Maven repo URI closes #3671

### LINKED ISSUE #3671: Unify path of MAVEN_LOCAL_DEFAULT in MavenRepository
The repository URL returned by `MavenRepository.MAVEN_LOCAL_DEFAULT` has no host information 
`file:/path` whereas all (?) other places use empty host `file:///path` syntax. 
Both are valid, I'll still provide a tiny PR to streamline it.

fabapp2:
## What's changed?

`MavenRepository.MAVEN_LOCAL_DEFAULT` cretaes the repository URI with `file://path` instead of `file:path` now.

## What's your motivation?

Both are valid URIs but this change streamlines the representation with other occurrences.

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
```diff
@@ -37,7 +37,7 @@
 public class MavenRepository implements Serializable {
 
     public static final MavenRepository MAVEN_LOCAL_USER_NEUTRAL = new MavenRepository("local", new File("~/.m2/repository").toString(), "true", "true", true, null, null, false);
-    public static final MavenRepository MAVEN_LOCAL_DEFAULT = new MavenRepository("local", new File(System.getProperty("user.home") + "/.m2/repository").toURI().toString(), "true", "true", true, null, null, false);
+    public static final MavenRepository MAVEN_LOCAL_DEFAULT = new MavenRepository("local", "file://" + System.getProperty("user.home") + "/.m2/repository", "true", "true", true, null, null, false);
```
timtebeek, line 40: My immediate thought here is what about Windows? From what I can find the value there is similar to
```
c:\Users\Foo
```
Would this then still work with that prefix? Would the same test pass or fail on Windows?

knutwannheden: I suspect it will be safest to create a `Path` and then an URI from that using `toURI()`.

shanman190: The original implementation or using `Path` are the only correct mechanisms to get a valid URI for all operating systems.

On Mac and *nix, the uri looks like:
```
# macOS
user.home=/Users/Foo
file:///Users/Foo/.m2/repository

# *nix
user.home=/home/Foo
file:///home/Foo/.m2/repository
```

For windows, it looks like:
```
user.home=C:\Users\Foo
file:///C:/Users/Foo/.m2/repository
```

By using `File` or `Path` as an intermediate step ensures that the result URI is valid for all operating systems.

timtebeek: This indeed seems to work for me on linux; anyone able to try the same on Windows?
```
jshell> Paths.get(System.getProperty("user.home"), ".m2", "repository").toUri().toString()
$1 ==> "file:///home/tim/.m2/repository/"
```

```suggestion
    public static final MavenRepository MAVEN_LOCAL_DEFAULT = new MavenRepository("local", Paths.get(System.getProperty("user.home"), ".m2", "repository").toUri().toString(), "true", "true", true, null, null, false);
```

shanman190: From Windows:
```
jshell> Paths.get(System.getProperty("user.home"), ".m2", "repository").toUri().toString();
$2 ==> "file:///C:/Users/tso4045/.m2/repository/"
```

timtebeek (approved):
Ok to merge if you are @fabapp2

rewrite-maven/src/test/java/org/openrewrite/maven/tree/MavenRepositoryTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class MavenRepositoryTest {
+    @Test
+    @DisplayName("verify URI of local Maven repository")
+    void expectedLocalMavenRepository() {
+        String expectedPath = "file://" + System.getProperty("user.home") + "/.m2/repository";
```
shanman190, line 27: @timtebeek and @fabapp2, this test also needs to be updated accordingly if it's kept. Otherwise, the tests will fail on Windows.

shanman190 (changes requested):

rewrite-maven/src/test/java/org/openrewrite/maven/tree/MavenRepositoryTest.java
timtebeek: I'm fine to remove the test class; no need to test `Paths.get`.

shanman190 (approved):

PR #3678, opened 2023-11-09
Fix `StackOverflowError` in `ResolvedDependency#findDependency()`

knutwannheden:


knutwannheden:
Depends on openrewrite/rewrite-gradle-tooling-model#10.

PR #3679, opened 2023-11-09
fix defaultClasspath for GradleParser

joanvr:
## What's changed?
Fixed default classpath for settings parser.

## What's your motivation?
When trying to parse a `settings.gradle` file it was failing with Class not found exceptions due to not properly setting the classpath on the parser.

@knutwannheden @jkschneider

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
@@ -44,14 +44,15 @@ public class GradleParser implements Parser {
     @Override
     public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path relativeTo, ExecutionContext ctx) {
         if (buildParser == null) {
-            if (base.buildscriptClasspath == null) {
+            Collection<Path> buildscriptClasspath = base.buildscriptClasspath;
+            if (buildscriptClasspath == null) {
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
```diff
@@ -44,8 +44,10 @@ public class GradleParser implements Parser {
     @Override
     public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path relativeTo, ExecutionContext ctx) {
         if (buildParser == null) {
-            if (base.buildscriptClasspath == null && defaultClasspath == null) {
-                defaultClasspath = loadDefaultClasspath();
+            if (base.buildscriptClasspath == null) {
+                if (defaultClasspath == null) {
+                    defaultClasspath = loadDefaultClasspath();
+                }
                 base.buildscriptClasspath = defaultClasspath;
```
shanman190, line 51: It may be safer to instead use a local variable that way if the end user has a handle to the base builder, then we aren't modifying the state of that instance from here as a side effect.

joanvr: Good point 👍

knutwannheden (approved):
@joanvr Would be cool if we can add a simple test for the recipe, so we don't again accidentally introduce a regression.

PR #3682, opened 2023-11-12
CreateTextFile should override files of different type

timtebeek:
## What's changed?
Allow `CreateTextFile` to overwrite files in a different format than `PlainText`.

## What's your motivation?
We recently started parsing `Jenkinsfile` as Groovy, which unfortunately broke the `CreateTextFile` used in `ModernizeJenkinsfile`
https://github.com/openrewrite/rewrite-jenkins/blob/8914ccd62655d65f2c54e2e40ed6289f13f69542/src/main/resources/META-INF/rewrite/rewrite.yml#L53-L70
Fixes https://github.com/openrewrite/rewrite-jenkins/issues/52

Called out below.

Not really; looked towards `FindAndReplace` as that also allows text replace regardless of file type.

## Any additional context
- https://github.com/openrewrite/rewrite-jenkins/issues/52

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
@@ -18,7 +18,10 @@ package org.openrewrite.text;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
+import org.openrewrite.binary.Binary;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.quark.Quark;
+import org.openrewrite.remote.Remote;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -81,7 +84,7 @@ public class CreateTextFile extends ScanningRecipe<AtomicBoolean> {
 
     @Override
     public Collection<SourceFile> generate(AtomicBoolean shouldCreate, ExecutionContext ctx) {
-        if(shouldCreate.get()) {
+        if (shouldCreate.get()) {
             return PlainTextParser.builder().build().parse(fileContents)
                     .map(brandNewFile -> (SourceFile) brandNewFile.withSourcePath(Paths.get(relativeFileName)))
                     .collect(Collectors.toList());
@@ -92,13 +95,17 @@ public class CreateTextFile extends ScanningRecipe<AtomicBoolean> {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(AtomicBoolean created) {
         Path path = Paths.get(relativeFileName);
-        return new PlainTextVisitor<ExecutionContext>() {
+        return new TreeVisitor<SourceFile, ExecutionContext>() {
             @Override
-            public PlainText visitText(PlainText text, ExecutionContext ctx) {
-                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(text.getSourcePath().toString())) {
-                    return text.withText(fileContents);
+            public @Nullable SourceFile visit(@Nullable Tree tree, ExecutionContext executionContext) {
+                SourceFile sourceFile = (SourceFile) requireNonNull(tree);
+                if (sourceFile instanceof Quark || sourceFile instanceof Remote || sourceFile instanceof Binary) {
+                    return sourceFile;
                 }
-                return text;
+                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(sourceFile.getSourcePath().toString())) {
+                    return PlainTextParser.convert(sourceFile).withText(fileContents);
+                }
+                return sourceFile;
             }
         };
     }
--- a/rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
@@ -18,10 +18,7 @@ package org.openrewrite.text;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
-import org.openrewrite.binary.Binary;
 import org.openrewrite.internal.lang.Nullable;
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
```diff
@@ -92,13 +95,17 @@ public Collection<SourceFile> generate(AtomicBoolean shouldCreate, ExecutionCont
...
+        return new TreeVisitor<SourceFile, ExecutionContext>() {
             @Override
-            public PlainText visitText(PlainText text, ExecutionContext ctx) {
-                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(text.getSourcePath().toString())) {
-                    return text.withText(fileContents);
+            public @Nullable SourceFile visit(@Nullable Tree tree, ExecutionContext executionContext) {
+                SourceFile sourceFile = (SourceFile) requireNonNull(tree);
+                if (sourceFile instanceof Quark || sourceFile instanceof Remote || sourceFile instanceof Binary) {
+                    return sourceFile;
+                }
```
timtebeek, line 104: Saw this in `FindAndReplace`, likely to prevent issues with `PlainTextParser.convert` below.

rewrite-test/build.gradle.kts
```diff
@@ -12,4 +12,6 @@ dependencies {
     implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-csv")
     implementation("org.slf4j:slf4j-api:1.7.36")
     implementation("org.eclipse.jgit:org.eclipse.jgit:5.13.+")
+
+    testImplementation(project(":rewrite-groovy"))
```
timtebeek, line 16: Added as a test dependency; for `java(...)` we would have similarly had to add for instance `rewrite-java-17`. Figured have the test below reflect what triggered the change, but open to change this. Didn't spot a second source file type in rewrite-core to use here.

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
sambsnyd: Yes, that is why it is there.

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
```diff
@@ -92,13 +95,17 @@ public Collection<SourceFile> generate(AtomicBoolean shouldCreate, ExecutionCont
...
+                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(sourceFile.getSourcePath().toString())) {
+                    return PlainTextParser.convert(sourceFile).withText(fileContents);
```
sambsnyd, line 106: Since you're wholly overwriting the existing file then there's no need to use `PlainTextParser.convert()` since you aren't interacting with its contents in any way. You can just new up a `PlainText` with an appropriate source path. Then you don't even need the guard against quark/binary/remote.

rewrite-test/build.gradle.kts
sambsnyd: This is fine.

sambsnyd (approved):
Looks good to me

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
knutwannheden: It would be nice if we had a registry of our parsers so that we could parse the source. But since we don't have access to the compile-time environment at this point, it would at best end up producing a not properly type-attributed LST.

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
```diff
@@ -92,13 +95,17 @@ public Collection<SourceFile> generate(AtomicBoolean shouldCreate, ExecutionCont
...
             @Override
-            public PlainText visitText(PlainText text, ExecutionContext ctx) {
-                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(text.getSourcePath().toString())) {
-                    return text.withText(fileContents);
+            public @Nullable SourceFile visit(@Nullable Tree tree, ExecutionContext executionContext) {
+                SourceFile sourceFile = (SourceFile) requireNonNull(tree);
+                if (sourceFile instanceof Quark || sourceFile instanceof Remote || sourceFile instanceof Binary) {
+                    return sourceFile;
+                }
+                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(sourceFile.getSourcePath().toString())) {
```
knutwannheden, line 105: Do we need the `Path#toString()` calls. Seems like we should be able to rely on `Path#equals(Object)`.

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
```diff
@@ -92,13 +95,17 @@ public Collection<SourceFile> generate(AtomicBoolean shouldCreate, ExecutionCont
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(AtomicBoolean created) {
         Path path = Paths.get(relativeFileName);
-        return new PlainTextVisitor<ExecutionContext>() {
+        return new TreeVisitor<SourceFile, ExecutionContext>() {
             @Override
-            public PlainText visitText(PlainText text, ExecutionContext ctx) {
-                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.toString().equals(text.getSourcePath().toString())) {
-                    return text.withText(fileContents);
+            public @Nullable SourceFile visit(@Nullable Tree tree, ExecutionContext executionContext) {
```
knutwannheden, line 100: AFAICT the `@Nullable` annotation isn't required here.

timtebeek: I've removed the first one on the return type; the second on on the first visit argument seems common throughout our code base.

PR #3685, opened 2023-11-13
Add BitbucketBuildEnvironment for better progress indicators

timtebeek:
## What's changed?
Add Bitbucket BuildEnvironment

## What's your motivation?
Recognize bitbucket CI, such that the CLI progress bar adjusts output.

Any reason we didn't have this before?
Anything we should be mindful of in terms of detection order?

We could detect `CI` only as a catch all CI env; but then we lack input for `GitProvenance`.

## Any additional context
As per https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/


knutwannheden (approved):

timtebeek:
Looks like this needs another coordinate merge and deploy, because we're adding a marker.
@pstreef would you mind merging this one whenever is convenient for you as part of on-call?

PR #3688, opened 2023-11-15
Add ExcludeTransitiveDependency recipe

kunli2:
Add a recipe to exclude transitive gradle dependencies.


timtebeek:
Cross linking an alternative interpretation in
- #3689

PR #3689, opened 2023-11-15
Alternative `ExcludeTransitiveDependency`

knutwannheden:


knutwannheden:
Note: This is still WIP. I don't think a lot is missing.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ExcludeTransitiveDependencyTest.java
```diff
@@ -0,0 +1,220 @@
...
+                      exclude group: "commons-collections", module: "commons-collections"
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void excludeAllTransitiveDependencies() {
```
shanman190, line 169: Normally when I see multiple of these excludes in the manner shown here, it makes me start to wonder if a global exclude isn't a better option.

I could see a global exclude dependency recipe satisfying that case though as another potential option.

knutwannheden: That is what this recipe should be doing, when no `configuration` option is set. Does that make sense?

shanman190: I'll clarify the difference that I'm talking about a little bit. 

I would consider each of these as a dependency-specfic exclusion even if the recipe applies said exclusion to all dependencies.

A global exclusion looks like:
```
configurations {
    all {
        exclude group: "commons-collections", module: "commons-collections"
    }
}

dependencies {
                     implementation("commons-beanutils:commons-beanutils:1.9.4") 
                   implementation("com.opencsv:opencsv:4.6") 
                    
                   testImplementation("commons-beanutils:commons-beanutils:1.9.4") 
                   testImplementation("com.opencsv:opencsv:4.6")
}
```

knutwannheden: @shanman190 This looks really good! Possibly we don't even need any configuration-specific exclusions, as I think that will be a rather esoteric use case. Possibly we could have some kind of `all` option to control whether to do it this way or whether to add exclusions to the individual dependencies (defaulting to modify the `all` configuration).

shanman190: I personally see both as being useful. The use cases in front of someone that needs to add the exclusion often guide to which implementation to use, but if it were me I'd probably make the following edits to support both styles.

* Rename `groupId` to `excludeGroupId`
* Rename `artifactId` to `excludeArtifactId`
* Introduce an optional `groupId` (more in a moment)
* Introduce an optional `artifactId` (more in a moment)

When `groupId` and `artifactId` are empty, we would create a global exclusion either in `all` if `configuration` is null or blank, otherwise in the specified `configuration`. When `groupId` and `artifactId` are present, then we add the exclusion only to the targeted dependencies using the same selection criteria.

shanman190: Also technically only one of `excludeGroupId` or `excludeArtifactId` is required. So you can add exclusions like the following:

```
configurations {
    all {
        exclude group: "commons-collections"
        exclude module: "commons-collections"
    }
}

dependencies {
    implementation("commons-beanutils:commons-beanutils") {
        exclude group: "commons-collections"
        exclude module: "commons-collections"
    }
}
```

knutwannheden:
Probably not applicable anymore.

PR #3690, opened 2023-11-15
Maven local could be in a different location

pstreef:


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -526,6 +526,11 @@ public Pom download(GroupArtifactVersion gav,
                         if (!f.exists()) {
                             continue;
                         }
+                        String fullDefaultMavenLocalUri = MavenRepository.MAVEN_LOCAL_USER_NEUTRAL.getUri().replace("~",System.getProperty("user.home"));
+                        //not the default maven local, we're skipping this for now
+                        if (!repo.getUri().startsWith(fullDefaultMavenLocalUri)) {
+                            continue;
```
timtebeek, line 532:
```suggestion
                        String fullDefaultMavenLocalUri = MavenRepository.MAVEN_LOCAL_USER_NEUTRAL.getUri().replace("~", System.getProperty("user.home"));
                        if (!repo.getUri().startsWith(fullDefaultMavenLocalUri)) {
                            // Non-default local Maven dependencies can not be shared between users, so we skip the repo
                            continue;
```

timtebeek (approved):

PR #3691, opened 2023-11-15
do not overwrite the repo if it's not the default local one

joanvr:
## What's changed?
Do not replace the repo with the user-agnostic ~/.m2/repository if it is not the local default one

## What's your motivation?
If a custom local maven repo is configured, we cannot assume that we can replace it with ~/.m2/repository, because it will point to a different location.

@timtebeek @pstreef 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -539,8 +539,7 @@ public class MavenPomDownloader {
                                 }
                             }
 
-                            String fullDefaultMavenLocalUri = MavenRepository.MAVEN_LOCAL_USER_NEUTRAL.getUri().replace("~", System.getProperty("user.home"));
-                            if (repo.getUri().equals(fullDefaultMavenLocalUri)) {
+                            if (repo.getUri().equals(MavenRepository.MAVEN_LOCAL_DEFAULT.getUri())) {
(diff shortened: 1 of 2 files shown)


pstreef:
Maybe add a unit test for this?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -544,8 +539,11 @@ public Pom download(GroupArtifactVersion gav,
                                 }
                             }
 
-                            // so that the repository path is the same regardless of username
-                            pom = pom.withRepository(MavenRepository.MAVEN_LOCAL_USER_NEUTRAL);
+                            String fullDefaultMavenLocalUri = MavenRepository.MAVEN_LOCAL_USER_NEUTRAL.getUri().replace("~", System.getProperty("user.home"));
+                            if (repo.getUri().equals(fullDefaultMavenLocalUri)) {
+                                // so that the repository path is the same regardless of username
+                                pom = pom.withRepository(MavenRepository.MAVEN_LOCAL_USER_NEUTRAL);
```
timtebeek, line 545: Don't we serialize this information, and therefor might not want a machine-specific path?

joanvr: but anyway, it does not make sense to replace it if it's not actually the home one, right?

timtebeek (approved):

PR #3698, opened 2023-11-16
Add modifiers to TypeParameter for kotlin

kunli2:
Add a list of `J.Modifier` to `J.TypeParameter` for Kotlin.
to support variance cases like this :
```kotlin
            @Target(AnnotationTarget.TYPE_PARAMETER)
            annotation class Anno
            class Foo  <   @Anno     out     T :  Any    >  () {}
```
`out` will be a modifier of `T :  Any` which is a  `J.TypeParameter` here.

as same as to `in` and `reified` keywords.

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2433,7 +2433,7 @@ public class GroovyParserVisitor {
                     .withAfter(EMPTY));
             bounds = JContainer.build(boundsPrefix, convertedBounds, Markers.EMPTY);
         }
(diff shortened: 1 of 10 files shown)


sambsnyd (approved):

knutwannheden (commented):
@kunli2 I think you may have forgotten to update `JavaVisitor`. Won't matter for Java, but for Kotlin or other derived languages, unless we there override the corresponding visit method.

kunli2:
> @kunli2 I think you may have forgotten to update `JavaVisitor`. Won't matter for Java, but for Kotlin or other derived languages, unless we there override the corresponding visit method.

Thanks, added here https://github.com/openrewrite/rewrite/commit/4f040664be8a836dbaf1d03a4bfa22c1f7988593

Issue #3699, opened 2023-11-17
ClassCastException in MavenParser in case of malformed POM


api-from-the-ion:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v5.8.0 in Maven plugin
- Maven plugin v5.11.0

## How are you running OpenRewrite?

I am using the Maven plugin, calling from the shell.
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:5.11.0:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE \
  -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit4to5Migration    ... 
```

## What is the smallest, simplest way to reproduce the problem?

We have malformed POM with something like this:
```xml
<configuration>
  <url>jdbc:postgresql://localhost</url>&gt;
</configuration>

```

## What did you expect to see?

I expect Rewrite to work and not to break with ClassCastException.
Additionally, it would be nice to have a logging possibility here, so I wouldn't have to clone the source code and remote debug the Maven plugin here to get to know what exactly is wrong at this place.

## What did you see instead?

Instead, I see an exception; see the stack trace below.

## What is the full stack trace of any errors you encountered?

```
java.lang.ClassCastException: org.openrewrite.tree.ParseError cannot be cast to org.openrewrite.xml.tree.Xml$Document
    at org.openrewrite.maven.MavenParser.parseInputs (MavenParser.java:87)
    at org.openrewrite.Parser.parse (Parser.java:58)
    at org.openrewrite.maven.MavenMojoProjectParser.parseMaven (MavenMojoProjectParser.java:478)
    at org.openrewrite.maven.MavenMojoProjectParser.listSourceFiles (MavenMojoProjectParser.java:150)
    at org.openrewrite.maven.AbstractRewriteMojo.loadSourceSet (AbstractRewriteMojo.java:257)
    at org.openrewrite.maven.AbstractRewriteMojo.listResults (AbstractRewriteMojo.java:239)
    at org.openrewrite.maven.AbstractRewriteRunMojo.execute (AbstractRewriteRunMojo.java:53)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
[...]

PR #3701, opened 2023-11-17
Do not log `ClassCastException` in MavenParser in case of malformed POM

api-from-the-ion:
## What's changed?

Wrote the unit test for the case #3699 and proposed fix

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite/issues/3699

No

No

I don't think there are any, but I'm glad to hear about them.

## Any additional context

After this fix, there is no ClassCastException. But the test asserts fails anyway because the POM is changed here - see the results of the tests. And now I don't know what to do with this. Should we suppress it or ignore it in the code itself? So that test would expect such changes or ignore them? Or is here some other way? Correct the parser, so these changes wouldn't appear and there would be no difference? 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
@@ -82,12 +82,18 @@ public class MavenParser implements Parser {
                 pom.getProperties().put("project.basedir", baseDir);
                 pom.getProperties().put("basedir", baseDir);
 
-                Xml.Document xml = (Xml.Document) new MavenXmlParser()
+                SourceFile sourceFile = new MavenXmlParser()
                         .parseInputs(singletonList(source), relativeTo, ctx)
                         .iterator().next();
 
-                projectPoms.put(xml, pom);
-                projectPomsByPath.put(pomPath, pom);
+                if (sourceFile instanceof Xml.Document) {
+                    Xml.Document xml = (Xml.Document) sourceFile;
+
+                    projectPoms.put(xml, pom);
+                    projectPomsByPath.put(pomPath, pom);
+                } else {
+                    parsed.add(sourceFile);
+                }
             } catch (Throwable t) {
                 ctx.getOnError().accept(t);
                 parsed.add(ParseError.build(this, source, relativeTo, ctx, t));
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for this improvement! I've gone ahead and updated your test to use the parser directly, as that's more convenient when asserting ParseErrors. Before your code change we returned a different ParseError from the `catch(Throwable)`, which we entered because of the `ClassCastException`. As a side effect that also logged the stacktrace you mentioned in
- #3699

In the new situation we still return a `ParseError`, but now the one created by the `MavenXmlParser` which extends `XmlParser`. That in turn might still log an exception, but hopefully with more details that help pinpoint and resolve the issue.
https://github.com/openrewrite/rewrite/blob/81582279dc825aa07a13c5609126dfbd2fcf6bd9/rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java#L60-L62

timtebeek (approved):

api-from-the-ion:
Thank you for review and merge. I have couple of questions. So now the exception from the XML parsers itself should be logged and could be seen by the end user? Because this was a major PITA ;-) to understand and find out where this came from and which POM should be corrected and where. It the shame that we have a ParserError with a diff with the exact description where this went wrong. But it was obscured by the ClassCastException. Because I can't test it right now - are you sure that this is this way now?

Another question: you mentioned and closed another issue with this PR; are you sure that there we have the same cause from the same place in code?

The final question is like in my another ticket. The cause of all of this was that the parser parsed the original XML, but after the comparing the result with original XML some differences were found. Which causes a ParseError. And the question is: could we also change the behavior of this parser to behave more like the Maven one? So such errors wouldn't appear?

timtebeek:
Hi @api-from-the-ion ; You were right, the fix was incomplete; it's now been fixed in https://github.com/openrewrite/rewrite-maven-plugin/commit/f6ffdc328c6346e6d3bc98ed69cc7e9552731d89 such that you should, finally, be able to see the original parser errors when problems occur. Thanks again!

> The final question is like in my another ticket. The cause of all of this was that the parser parsed the original XML, but after the comparing the result with original XML some differences were found. Which causes a ParseError. And the question is: could we also change the behavior of this parser to behave more like the Maven one? So such errors wouldn't appear?

That'll depend on what errors occur, and I hope my changes will help surface and resolve those errors. We have our own parser based on our XmlParser, which differs from the implementation Apache Maven uses. Without knowing the details of the errors you had been getting I can't say for certain which is better.

I hope the above change help surface your issues, and please keep reporting any problems that you find; really helpful to have detailed feedback.

PR #3706, opened 2023-11-19
Ignore methods with type parameters in UseStaticImport

### LINKED ISSUE #3705: UseStaticImport fails on <Generic>calls()
`Collections.<Object>emptyList()` and similar explicitly typed generic methods are not handled correctly in `UseStaticImport`.

```java
    @Test
    void ecplicitGenerics() {
        rewriteRun(
          spec -> spec.recipe(new UseStaticImport("java.util.Collections emptyList()")),
          java(
            """
            import java.util.Collections;
            import java.util.List;

            public class Reproducer {
                public void reproduce() {
                    List<Object> list = Collections.emptyList();
                    List<Object> list = Collections.<Object>emptyList();
                }
            }
            """,
            """
            import java.util.List;

            import static java.util.Collections.emptyList;

            public class Reproducer {
                public void reproduce() {
                    List<Object> list = emptyList();
                    List<Object> list = emptyList();
                }
            }
            """
          )
        );
    }
```

The second call either being shortened to just the method name or not changed at all.

## What did you see instead?

The second line with the generic becomes this non compiling code:
```java
List<Object> list = <Object>emptyList();
```

Maybe, but I haven't been able to fix the other ones I raised for the same class, yet.

Bananeweizen:
Fixes #3705.

## What's changed?
Avoids the recipe in case of type parameters, as suggested by @knutwannheden 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
@@ -18,6 +18,7 @@ package org.openrewrite.java;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Issue;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Nicely done! Thanks for the quick fix.

PR #3707, opened 2023-11-19
Tentative fix for missing "//" operator on `XPathMatcher`

ammachado:
## What's changed?
Changes `XPathMatcher` to allow `//` syntax.

## What's your motivation?
https://github.com/openrewrite/rewrite-migrate-java/issues/353

This is just enough to fix the before mentioned issue, definitely needs more testing and a better implementation.

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.xml;
 
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.Cursor;
 import org.openrewrite.xml.search.FindTags;
 import org.openrewrite.xml.tree.Xml;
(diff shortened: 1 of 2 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
```diff
@@ -78,6 +88,33 @@ public boolean matches(Cursor cursor) {
             Collections.reverse(path);
             String[] parts = expression.substring(1).split("/");
 
+            // Deal with the two forward slashes in the expression; works, but I'm not proud of it.
```
timtebeek, line 91: Missed chance at using [our `@SelfLoathing(name= "...")` annotation](https://github.com/openrewrite/rewrite/blob/e4f66c2ba20fd7561cae6d0c5631cae581b59edf/rewrite-core/src/main/java/org/openrewrite/internal/SelfLoathing.java#L28).

timtebeek:
Decided to merge as is to not keep this open; since we specifically look at cases where the expression contains `//` I don't expect any adverse effects. Thanks for identifying the problem and providing the fix!

PR #3708, opened 2023-11-19
Remove default `jar` packaging when using `ChangePackaging`

ammachado:
## What's changed?
Omit `jar` packaging on maven pom.xml files (it is the default)

## What's your motivation?
Omit the `<packaging>jar</packaging>` tag on pom.xml files when using the `ChangePackaging` recipe.

### Checklist


timtebeek (approved):
Makes sense, yes. Thanks for adding this!

PR #3718, opened 2023-11-21
Change TypeParameter to be a Expression, to support kotlin intersection type

kunli2:
This change is to support Kotlin intersection type like this

```kotlin
              import java.util.*

              @Suppress("UNUSED_PARAMETER")
              fun <T : Any?> test(n: Optional <  T   &    Any > = Optional.empty<T>()) {}
```

`Optional <  T   &    Any >` is a `J.ParameterizedType`, 
If we parse `T   &    Any` as a `J.TypeParameter`, then it needs to be a `J.Expression` since  `J.ParameterizedType` requires a `JContainer<Expression>` for type parameters.

There is an alternative way to avoid this change, which is to parse `T   &    Any` to a `K.Binary` with a new binary type.


knutwannheden:
I think we should first finish and integrate #3652 which will introduce `J.IntersectionType`.

kunli2:
yes, let's use `J.IntersectionType` instead.

PR #3720, opened 2023-11-21
Change annotation attribute name to support Swagger migration

timtebeek:


rewrite-java/src/main/java/org/openrewrite/java/ChangeAnnotationAttributeName.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Option(displayName = "Annotation Type",
+            description = "The fully qualified name of the annotation.",
+            example = "org.junit.Test")
+    String annotationType;
+
+    @Option(displayName = "Old attribute name",
+            description = "The name of attribute to change.",
+            required = false,
```
pstreef, line 45: I presume if this is not passed the default `value` is renamed to the `newAttributeName `? Perhaps that would be good to add in the description?

pstreef: Actually, it looks like it might be required after all? There is no nullCheck for this field below

timtebeek: Good catch, thanks! Copy paste error from the source recipe.

rewrite-java/src/main/java/org/openrewrite/java/ChangeAnnotationAttributeName.java
```diff
@@ -0,0 +1,75 @@
...
+            example = "waitFor")
+    String newAttributeName;
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>(annotationType, false), new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext context) {
+                J.Annotation a = super.visitAnnotation(annotation, context);
+                if (!new AnnotationMatcher(annotationType).matches(a)) {
```
joanvr, line 59: Does it make sense to move this AnnotationMatcher as a field in the JavaIsoVisitor to not instantiate it every (and compile the pattern) on every annotation?

timtebeek: ah yes, I'd been meaning to extract that to a constant as I typed that matcher, but forgot. thanks!

timtebeek:
Thanks for the feedback both; merged with the reasonable assumption that there's no further feedback; let me know otherwise!

PR #3724, opened 2023-11-22
AutoFormat bug with method parameter alignment when using tabs

shanman190:
## What's changed?
Test case showing bug with continuation indents improperly being auto formatted.

## Any additional context
I've seen another case related to continuation indents for method arguments as well as method chains, but haven't been able to reproduce that case yet.

### Checklist


knutwannheden:
This indeed appears to be related to `TabsAndIndentsStyle#useTabCharacter` set to `true`, because if I set it to `false` and change all tabs into four spaces, then the test passes.

PR #3726, opened 2023-11-22
Migrate single-source applicability tests to preconditions

knutwannheden:


knutwannheden:
@timtebeek Oops! Looks like we have a conflict here. I always mix up any-source and single-source tests, but I hope I remember it correctly here.

I wasn't sure if we want to rename the recipe or not. But that is of course also an option.

timtebeek:
Ah feel free to overwrite my changes! I'd abandoned my attempts at using `ChangePropertyKey` as I thought that did not allow to change the hierarchy, only the last key in a path, whereas here we want to flatten `applicability.singleSource` to `preconditions`. I didn't think it worth the trouble to automate the change this long after the switch to Rewrite 8, but perhaps the Red Hat folks would be grateful still.

timtebeek (approved):

PR #3727, opened 2023-11-22
Prevent NPE as reported due to null normalized repository

timtebeek:
Possibly due to mirrors applied with property placeholders. This is part exploratory until I know exactly what's lead to the problem in the first place, but the changes here ought to help regardless.


timtebeek:
As discussed in Slack; we'll merge and see if this turns up more.

PR #3729, opened 2023-11-22
Don't use static import on collision with local method

Bananeweizen:
If a class declares a method with the same name as a candidate for a static import, then that will always lead to compile errors if the static import is applied. Therefore the recipe must check for not having any similarly named method locally.

## What's changed?
Additional precondition for static import, namely `not(DeclaresMethod("* sameNameAsMethodPattern(..)"`.

## What's your motivation?
Quite some compile errors on larger code bases with UseStaticImport recipe. The included unit test demonstrates this and leads to uncompilable code without the change.

The method pattern is created via String manipulation. Not sure if there are better ways using MethodPattern, MethodPatternParser etc. but that was hard to explore.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
@@ -93,6 +93,29 @@ class UseStaticImportTest implements RewriteTest {
           )
         );
     }
+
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the clear example and fix! I don't think there's any other way to check really check the same from what you've done.

PR #3734, opened 2023-11-24
Stop over eager exclude of UseStaticImport

timtebeek:
Fixes [org.openrewrite.java.testing.junit5.AssertToAssertionsTest#staticallyImportAssertions](https://github.com/openrewrite/rewrite-testing-frameworks/blob/39dc67cab9a2d32ed000e5ad1348d31cae3979cd/src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java#L318) after
- https://github.com/openrewrite/rewrite/pull/3729

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
@@ -51,10 +51,15 @@ public class UseStaticImport extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        int indexSpace = methodPattern.indexOf(' ');
-        int indexBrace = methodPattern.indexOf('(', indexSpace);
-        String identicalMethodName = "* " + methodPattern.substring(indexSpace, indexBrace) + "(..)";
-        return Preconditions.check(Preconditions.and(new UsesMethod<>(methodPattern), Preconditions.not(new DeclaresMethod<>(identicalMethodName))), new UseStaticImportVisitor());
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
```diff
@@ -51,10 +51,15 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        int indexSpace = methodPattern.indexOf(' ');
-        int indexBrace = methodPattern.indexOf('(', indexSpace);
-        String identicalMethodName = "* " + methodPattern.substring(indexSpace, indexBrace) + "(..)";
-        return Preconditions.check(Preconditions.and(new UsesMethod<>(methodPattern), Preconditions.not(new DeclaresMethod<>(identicalMethodName))), new UseStaticImportVisitor());
+        TreeVisitor<?, ExecutionContext> preconditions = new UsesMethod<>(methodPattern);
+        if (!methodPattern.contains(" *(")) {
```
knutwannheden, line 55: I wasn't aware we did this kind of parsing in this recipe. Would be cleaner to use `MethodSignatureParser`, but probably a bit more work.

timtebeek: Yes that was a recent addition; I had a brief look at using `MethodSignatureParser`, but as that takes a `TokenStream` I don't think it's easily adopted here. I don't think `" *("` could mismatch easily here; at worst we'd then not use a static import. It works for the downstream tests in rewrite-testing-frameworks.

PR #3735, opened 2023-11-25
Ensure SourceSpec text blcks start on a new line

timtebeek:
## What's changed?
Add a newline & autoformat when SourceSpec text block has no whitespace or comment prefix.

## What's your motivation?
Commonly pointed out in review; by automating the code change & review we reduce review burden, and get more consistent code.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/recipes/SourceSpecTextBlockIndentationTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/recipes/SourceSpecTextBlockIndentationTest.java
@@ -23,7 +23,7 @@ import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
 
-public class SourceSpecTextBlockIndentationTest implements RewriteTest {
+class SourceSpecTextBlockIndentationTest implements RewriteTest {
 
     @Override
     public void defaults(RecipeSpec spec) {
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/recipes/SourceSpecTextBlockIndentationTest.java
```diff
@@ -83,4 +83,56 @@ void test() {
...
+                    rewriteRun(
+                       text(\"""
+                           class Test {
+              \s
+                              \s
+                               void test() {
+                                   System.out.println("Hello, world!");
+                               }
+                           }
+                           \"""
```
knutwannheden, line 106: Possibly we could also change the recipe to put the closing parenthesis following the text block on its own line. Quite often I see it be on the same line.

knutwannheden (approved):

rewrite-java-test/src/test/java/org/openrewrite/java/recipes/SourceSpecTextBlockIndentationTest.java
timtebeek: Good addition indeed; I've also thought that maybe we should detect placement of text blocks on a new line as a style option, and aim for consistency that way rather than with this addition here. Might be more involved though.

timtebeek: I'm going to merge this one as is, as I don't immediately see an easy way to add a suffix to the text block under the right conditions, and as implemented now it's already an improvement that might prompt a rethink of newlines on contributions. Thanks for the suggestion; can be picked up separately.

PR #3743, opened 2023-11-29
Added a few type validation checks to FindMissingTypes.

### LINKED ISSUE #3739: Improve FindMissingTypes to check NameTree(s) types based on the parent tree.
Several types of type attribution checks need to be included that will help to identify issues updating types via recipes and find type attribution issues in other implementations of J.

I.E.
- `J.ParameterizedType#clazz` and J.Annotation#annotatedType should have a `JavaType$Class` if the `NameTree` is a `J.Identifier`.
- The name of J.MethodInvocation and J.MethodDeclaration should have the method type.

traceyyoshima:
Changes:

- FindMissingTypes will validate the `name` field `MethodInvocation` and `MethodDeclaration` when the method type is not null and the type on the identifier is not null.
- FindMissingTypes will validate the `clazz` field on `J.ParameterizedType` and `J.NewClass` when the clazz is a `J.Identifier` and the type is not null.
- ChangeMethodName, ChangeMethodTargetToStatic, ChangeMethodTargetToVariable, ChangePackage, ChangeType, SimplifyMethodChain, and ImplementInterface will update the old type with the new type on applicable identifiers.

These types occur from the Java compiler and are expected on the LST. The changes help to identify unexpected changes through recipes and misaligned type mappings in other implementations of `J`.
fixes #3739

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest3Test.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest3Test.java
@@ -369,7 +369,7 @@ class JavaTemplateTest3Test implements RewriteTest {
               public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                   J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
                   if (mi.getSimpleName().equals("acceptInteger")) {
-                      return JavaTemplate.builder("acceptString(#{any()}.toString())").contextSensitive()
+                      mi = JavaTemplate.builder("acceptString(#{any()}.toString())").contextSensitive()
                         .javaParser(JavaParser.fromJavaVersion()
                           .dependsOn(
                             """
@@ -384,6 +384,7 @@ class JavaTemplateTest3Test implements RewriteTest {
                         )
                         .build()
                         .apply(updateCursor(mi), mi.getCoordinates().replaceMethod(), mi.getArguments().get(0));
+                      mi = mi.withName(mi.getName().withType(mi.getMethodType()));
                   }
                   return mi;
               }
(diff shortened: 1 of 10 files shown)


traceyyoshima:
The type checks revealed a couple of old types that were left behind by recipes.
It's possible downstream projects will have similar issues, but I'll fix those after the changes are merged.

knutwannheden (approved):
I am wondering if requiring referential equality might be a bit too strict. Can we try to gauge how many recipes will be affected by these validations by locally checking rewrite-static-analysis and rewrite-migrate-java?

traceyyoshima:
> I am wondering if requiring referential equality might be a bit too strict. Can we try to gauge how many recipes will be affected by these validations by locally checking rewrite-static-analysis and rewrite-migrate-java?

Sure thing, no problem, I'll check! We can also loosen the comparison by comparing the signatures, which'll ensure the types have been updated without requiring referential equality.

traceyyoshima:
> Can we try to gauge how many recipes will be affected by these validations by locally checking rewrite-static-analysis and rewrite-migrate-java?

@knutwannheden

`rewrite-migrate-java`:
10/588 tests fail after the type validation changes from 5 recipes.

`rewrite-static-analysis`:
24/1133 tests fail after the type validation changes from 6 recipes.

I already planned to fix downstream projects, and I don't anticipate the revisions will take much time.
I'll merge the changes tomorrow and fix the downstream tests afterward.

Note:
I only looked at a couple of the test failures, but one of the issues appears to be caused by `JavaTemplate`.
`ReplaceRedundantFormatWithPrintf` uses the `JavaTemplate` [here](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/ReplaceRedundantFormatWithPrintf.java#L154) to update the method and leaves the old type in the name. In this case, it might be better to fix the `JavaTemplate` so that old types aren't left in the LST.

knutwannheden:
@traceyyoshima Sounds good. I've done the same thing a few times already. Typically, these two repos are those where you get the most problems.

PR #3749, opened 2023-12-01
Change HasSourcePath from a Visitor into a Recipe for use in yaml preconditions

timtebeek:
We already had HasSourcePath as visitor, which was usable from `Preconditions.check`.
Now it's still usable from `Preconditions.check` (through a different method), and also usable from [yaml preconditions](https://docs.openrewrite.org/reference/yaml-format-reference#preconditions).


rewrite-core/src/main/java/org/openrewrite/HasSourcePath.java
```diff
@@ -35,34 +44,52 @@ public HasSourcePath(@Nullable String filePattern) {
...
-            SourceFile sourceFile = (SourceFile) tree;
-            Path sourcePath;
-            if ("glob".equals(syntax) && filePattern.startsWith("**")) {
-                sourcePath = Paths.get(".").resolve(sourceFile.getSourcePath().normalize());
-            } else {
-                sourcePath = sourceFile.getSourcePath().normalize();
-            }
+    @Override
+    public String getDescription() {
+        return "Matches a source file based on its path, for use as preconditions. " +
```
timtebeek, line 59:
```suggestion
        return "Matches a source file based on its path, for use as precondition. " +
```

timtebeek:
As discussed in Slack; seeing this one through.

Issue #3759, opened 2023-12-03
Fallthrough case statement in groovy causes paring error


reallyinsane:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.8.0
- Maven/Gradle plugin v5.14.1
- rewrite-groovy_module_ v8.8.0

## How are you running OpenRewrite?
I'm trying to build my own recipes manipulating Jenkinsfile (groovy). Therefore I run tests and execute the recipe using the Maven plugin.

## What is the smallest, simplest way to reproduce the problem?

The simplest way is having a switch statement in a groovy file with a fallthrough case statement. The parser of openrewrite expects a blockstatement following the case statement, so it fails always.

You can use the test org.openrewrite.groovy.tree.SwitchTest#fallthroughCase in my fork.

```java
switch("foo") {
  case "foo":
  case "bar": {
    break
  }
}
```

## What did you expect to see?

I'd expect that a fallthrough statement is accepted.

## What did you see instead?
A parsing error.

```
Failed to parse file.groovy at cursor position 31. The next 10 characters in the original source are `
    case `
org.openrewrite.groovy.GroovyParsingException: Failed to parse file.groovy at cursor position 31. The next 10 characters in the original source are `
    case `
	at app//org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:174)
	at app//org.openrewrite.groovy.GroovyParser.lambda$parseInputs$4(GroovyParser.java:154)
	at app//org.openrewrite.groovy.GroovyParser$$Lambda$480/0x0000000800dace00.apply(Unknown Source)
	at java.base@17.0.2/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base@17.0.2/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base@17.0.2/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845)
	at java.base@17.0.2/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base@17.0.2/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base@17.0.2/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base@17.0.2/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #3760, opened 2023-12-03
Fix for #3759

reallyinsane:
Allow empty statement to support case-fallthrough.

## What's changed?
Groovy parsing case-statement allows empty statement instead of just blockstatement.

## What's your motivation?
We're having Jenkinsfiles we want to change with openrewrite. These contain fallthrough-case statements. Because of #3759 recipes are failing with parsing errors.

I'm not quite sure which kind of other statement may follow a case-statement apart from EmptyStatement and BlockStatement. My code just adds EmptyStatement support.

As I want to change the Jenkinsfiles automatically changing them manually is not an option for me.

### Checklist
- [x ] I've added unit tests to cover both positive and negative cases


timtebeek (approved):
Thanks a lot for reporting and immediately fixing the issue. I've merely reformatted and added a test case what wasn't covered yet. Good to merge!

PR #3764, opened 2023-12-04
Support `@OptionGroup` on `Recipe`

JLLeitschuh:
Enables composing `Recipe` of one or more objects annotated with `@OptionGroup` to support
re-use of recipe options between multiple recipes.

Supports serialization/deserialization correctly through the use of Jackson's `@JsonUnwrapped` annotation.

## What's changed?

Modifies `Recipe`'s `getOptionDescriptors` to support a new `@NestedOption` annotation.

## What's your motivation?

As I'm building out recipes within `rewrite-java-security` I'm finding a desire to share the same configuration
options between multiple recipes.

In general, for security fix recipe, you only want to apply those recipes when both test and production source is modified.
Before 8.0 that heavily refactored recipes and how many source files were loaded into memory, this was possible with the `anySource` applicability test.

Eg:
```yaml
name: com.example.SecurityModifyRecipe
displayName: Apply `SecureTempFileCreation`
description: Applies the `SecureTempFileCreation` to non-test sources first, if changes are made, then apply to all sources.
applicability:
  anySource:
    - org.openrewrite.java.search.IsLikelyNotTest
    - org.openrewrite.java.security.SecureTempFileCreation
recipeList:
  - org.openrewrite.java.security.SecureTempFileCreation
```

See: https://github.com/openrewrite/rewrite/discussions/2849

However, in 8.0 this was broken. As such, the current advice is to put this filtering into each security recipe individually.

In order to achive this, and reduce code-duplication, being able to compose recipes with a common set of options, will enable that, while reducing code duplication.

As this feature will need to integrate with the Moderne SaaS as the interface for how these new composed options are tested, I'd want to sanity check that
SaaS invocation of Recipes uses jackson deserialization to instantiate them as I expect. If that is the case, then this feature should work seemlessly with the Moderne SaaS
without any needed changes there to support this functionality.

This feature should have no impact on any existing recipes or how those recipes are invoked by the SaaS.

I will test and verify this functionality works as-expected by modifying the recipes in `rewrite-java-security` to verify this works with the Moderne SaaS as expected.

@sambsnyd

A single base-class with base options doesn't work as the fields on the base-class still need to be surfaced in the constructor of the subclasses for proper deserialization by Jackson. Additionally, `Recipe#getOptionDescriptors` currently uses `Class#getDeclaredFields` which, as documented, does not include inherited fields, so would not be seen when creating the `OptionDescriptor`

Lots of duplicate code, including duplicating option annotations between multiple recipes.
Using multiple hard-coded constants for each duplicated annotation.

This solution is more elegant IMHO.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/OptionGroup.java
+++ b/rewrite-core/src/main/java/org/openrewrite/OptionGroup.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2023 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,7 +15,8 @@
  */
 package org.openrewrite;
 
-import org.intellij.lang.annotations.Language;
+import com.fasterxml.jackson.annotation.JacksonAnnotationsInside;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
@@ -23,17 +24,17 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Options are used to configure recipes. They are annotated on fields of recipes and are used to generate
- * documentation and to validate user input.
+ * Nested options are used to configure recipes.
  * <p>
- * For nested options, see {@link OptionGroup}.
+ * They allow for options to be re-used across multiple {@link Recipe} implementations through object composition.
+ * <p>
+ * Fields annotated with this method <i>may</i> also be annotated with {@link JsonUnwrapped} for further configuration
+ * of the deserialized JSON.
  */
+@Incubating(since = "8.12.0")
 @Target({ElementType.FIELD, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
-public @interface Option {
-    @Language("markdown") String displayName() default "";
-    @Language("markdown") String description() default "";
-    String example() default "";
-    String[] valid() default "";
-    boolean required() default true;
+@JsonUnwrapped
+@JacksonAnnotationsInside
+public @interface OptionGroup {
 }
--- a/rewrite-core/src/test/java/org/openrewrite/OptionGroupTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/OptionGroupTest.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.config.Environment;
+import org.openrewrite.config.RecipeDescriptor;
+import org.openrewrite.config.YamlResourceLoader;
+import org.openrewrite.test.RewriteTest;
+
+import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class OptionGroupTest implements RewriteTest {
+
+    @Value
+    public static class AnOptionGroup {
+        @Option(displayName = "Option within a nested class", description = "Option")
+        String aNestedOption;
+    }
+
+    @Value
+    @EqualsAndHashCode(callSuper = true)
+    public static class ARecipe extends Recipe {
+
+        @Option(displayName = "An option", description = "Option")
+        String anOption;
+
+        @OptionGroup
+        AnOptionGroup aGroup;
+
+        @Override
+        public String getDisplayName() {
+            return "A Recipe";
+        }
+
+        @Override
+        public String getDescription() {
+            return "A Recipe Description";
+        }
+    }
+
+    @Test
+    public void nestedOption() {
+        ARecipe recipe = new ARecipe(
+                "An option",
+                new AnOptionGroup("A nested option")
+        );
+        RecipeDescriptor descriptor = recipe.getDescriptor();
+        assertEquals(2, descriptor.getOptions().size(), "Option count");
+    }
+
+    @Test
+    void loadingNested() {
+        Environment env = Environment.builder()
+                .load(new YamlResourceLoader(new ByteArrayInputStream(
+                        //language=yml
+                        """
+                                type: specs.openrewrite.org/v1beta/recipe
+                                name: test.LoadNestedRecipe
+                                displayName: Load Nested Recipe
+                                recipeList:
+                                    - org.openrewrite.OptionGroupTest$ARecipe:
+                                        anOption: Hello!
+                                        aNestedOption: World!
+                                """.getBytes()
+                ), URI.create("rewrite.yml"), new Properties()))
+                .build();
+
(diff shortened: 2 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/MethodMatcherOption.java
```diff
@@ -0,0 +1,41 @@
...
+    @Option(displayName = "Match on overrides",
+            description = "When enabled, find methods that are overrides of the method pattern.",
+            required = false)
+    @Nullable
+    Boolean matchOverrides;
+
+    public MethodMatcher methodMatcher() {
+        return new MethodMatcher(methodPattern, matchOverrides != null && matchOverrides);
+    }
+}
```
JLLeitschuh, line 41: IMHO this is the best example of how this option can be used. I'd want to verify that this works with the SaaS as I expect it to before refactoring all of the existing recipes to use this instead. But for now, this demonstrates a solid example of how this feature can be used.

rewrite-java/src/main/java/org/openrewrite/java/MethodMatcherOption.java
JLLeitschuh, line 1: IMHO this is the best example of how this new `@NestedOption` can be used for `rewrite-java`. I'd want to verify that this works with the SaaS as I expect it to before refactoring all of the existing recipes to use this instead. But for now, this demonstrates a solid example of how this feature can be used.

JLLeitschuh:
@sambsnyd thoughts on merging this?

knutwannheden:
I think `@OptionGroup` would be a better name and the tight coupling with the `@JsonUnwrapped` annotation is not so nice. As an alternative we could check if the `@JacksonAnnotationsInside` meta annotation works for us. https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JacksonAnnotationsInside.html

Further, I also wonder if we should possibly move the annotation to the option group type instead. Then it would be very similar to how this all works in Quarkus with `@ConfigItem` and `@ConfigGroup`, IIRC. Although, I have no idea if this aligns with how `@JsonUnwrapped` works or not.

knutwannheden:
I just checked: `@JsonUnwrapped` has to be declared on the property side. But maybe the meta annotation approach works?

JLLeitschuh:
Already using `@JacksonAnnotationsInside`

https://github.com/openrewrite/rewrite/pull/3764/files#diff-e4085063acbc846936645bf5269d26d05193cd41027e06014841b7afa02ee32bR38

JLLeitschuh:
Take a look at the current code and see if that aligns with your vision or not

JLLeitschuh:
Anything else needed to get this across the finish line?

JLLeitschuh:
@timtebeek anything we can do to get this across the finish line?

rewrite-core/src/main/java/org/openrewrite/OptionGroup.java
```diff
@@ -0,0 +1,40 @@
...
+ * Fields annotated with this method <i>may</i> also be annotated with {@link JsonUnwrapped} for further configuration
+ * of the deserialized JSON.
+ */
+@Incubating(since = "8.12.0")
+@Target({ElementType.FIELD, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@JsonUnwrapped
+@JacksonAnnotationsInside
+public @interface OptionGroup {
+}
```
ammachado, line 40: I like this idea too, @JLLeitschuh. Perhaps you could also add a few fields to it:
```suggestion
import com.fasterxml.jackson.annotation.JacksonAnnotationsInside;
import org.intellij.lang.annotations.Language;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

/**
 * Nested options are used to configure recipes.
 * <p>
 * They allow for options to be re-used across multiple {@link Recipe} implementations through object composition.
 * <p>
 * Fields annotated with this method <i>may</i> also be annotated with
 * {@link com.fasterxml.jackson.annotation.JsonProperty} or {@link com.fasterxml.jackson.annotation.JsonUnwrapped}
 * for further configuration of the deserialized JSON.
 */
@Incubating(since = "8.12.0")
@Target({ElementType.FIELD, ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
@JacksonAnnotationsInside
public @interface OptionGroup {
    @Language("markdown")
    String displayName() default "";
    @Language("markdown")
    String description() default "";
    boolean required() default true;
}
```

rewrite-core/src/test/java/org/openrewrite/OptionGroupTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class OptionGroupTest implements RewriteTest {
```
ammachado, line 35:
```suggestion
import com.fasterxml.jackson.annotation.JsonProperty;
import lombok.EqualsAndHashCode;
import lombok.Value;
import org.assertj.core.api.InstanceOfAssertFactories;
import org.junit.jupiter.api.Test;
import org.openrewrite.config.CompositeRecipe;
import org.openrewrite.config.DeclarativeRecipe;
import org.openrewrite.config.Environment;
import org.openrewrite.config.OptionDescriptor;
import org.openrewrite.config.RecipeDescriptor;
import org.openrewrite.config.YamlResourceLoader;
import org.openrewrite.internal.lang.Nullable;
import org.openrewrite.test.RewriteTest;

import java.io.ByteArrayInputStream;
import java.net.URI;
import java.util.Properties;

import static org.assertj.core.api.Assertions.assertThat;
import static org.junit.jupiter.api.Assertions.assertEquals;

public class OptionGroupTest implements RewriteTest {
```

rewrite-core/src/test/java/org/openrewrite/OptionGroupTest.java
```diff
@@ -0,0 +1,110 @@
...
+
```
ammachado, line 36:
```suggestion
    @Value
    public static class ANestedObject {
        @Option(displayName = "Option A", description = "Option A")
        String optionA;
        @Nullable
        @Option(displayName = "Option B", description = "Option B", required = false)
        Integer optionB;
    }
```

rewrite-core/src/main/java/org/openrewrite/OptionGroup.java
JLLeitschuh: Could you elaborate on what the behavior of those additional arguments would be?

The sub elements of the @OptionGroup will already have those fields on them. Would these be prefixes or suffexes for the children items?

ammachado: This is how I would use it, but I would also add it at least to maintain consistency with existing `@Option`s.

One detail that I think is missing in your implementation is the validation of nested fields.

I removed my comments to keep the review clean, but please let me know if I can help.

JLLeitschuh: The goal was to, out of the box, maintain consistent support with the Moderne SaaS. If `@OptionGroup` required specifying nested json/yaml fields, then the Moderne team would need to make changes to the SaaS to support it, which I wanted to avoid

JLLeitschuh: This is why I chose to have `@OptionGroup` contain `@JsonUnwrap` by default

JLLeitschuh:
@timtebeek can we push this forward?

rewrite-core/src/main/java/org/openrewrite/OptionGroup.java
JLLeitschuh: > One detail that I think is missing in your implementation is the validation of nested fields.

Can you elaborate? AFAIK, I'm not doing anything different from existing recipe validation. If I'm wrong, can you point me towards where I should be looking?

JLLeitschuh: @sambsnyd @knutwannheden what are your thoughts here. My intention was to add this annotation such that it required no SaaS changes to support.

Would you prefer to be able to actually support nested JSON inputs OR flattened inputs. Or would you prefer to go with what I have currently which is that `@OptionGroup` always flattens the inputs

timtebeek:
Appreciate the work done here @JLLeitschuh , but I've just discussed your open PRs with Sam and we've decided for now not to support option groups. There would have to be additional changes both on the OpenRewrite side as well as the Moderne side to for instance recipe loading from Yaml, that we feel would add too much complexity for a use case that we feel is relatively rare. 
Thanks for both the initial work done here and the above discussion!

PR #3768, opened 2023-12-05
added plugins to ResolvedPom + some tests

joanvr:
## What's changed?
Added plugins info to ResolvedPom
Also added some extra tests on MavenParser regarding plugins and pluginManagement

I've found that UpdateMavenModel does not update pluginManagement right now. Maybe we should add there plugins and pluginManagement?

@timtebeek @knutwannheden 

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -470,6 +487,18 @@ private void mergeRequestedDependencies(List<Dependency> incomingRequestedDepend
...
+            if (!incomingPlugins.isEmpty()) {
+                if (plugins == null || plugins.isEmpty()) {
+                    //It is possible for the plugins to be an empty, immutable list.
+                    //If it's empty, we ensure to create a mutable list.
+                    plugins = new ArrayList<>(incomingPlugins);
+                } else {
+                    plugins.addAll(incomingPlugins);
+                }
+            }
+        }
```
timtebeek, line 500: What would happen if the current project and a parent both have a `maven-compiler-plugin`? Are these then both added to this list?

What about a plugin that occurs both in `<build><pluginManagement>` and `<build><plugins>` with configuration in both?

joanvr: oh, it seems that there is an interesting fine-tunning in inheritance of plugins from parent: https://www.baeldung.com/maven-plugin-override-parent.

joanvr: Implemented some merging logic. Missing merging of configurations and executions.
For configurations I think we won't be able to do the fine-grained merging, since we are missing the xml attributes in the JsonNode object.
For executions still not clear to me how to merge them exactly.

kmccarp: @timtebeek is there any reason not to merge this as-is and then work on the better merging?

timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
timtebeek: Ok to merge; I'd held off as every change to tree classes requires a coordinated deploy of the platform, right?
Figured best to limit the amount of coordinated releases and deployments, but haven't found time to dive deeper here.

joanvr:
I'm probably going to finish the "basic" merging of configurations, to at least have the ones in the parent in the child too, but if the `combine.self` xml attribute is not available in JsonNode, probably I'm going to not cover it in this PR yet.
Also, regarding executions, we do not actually need them right now to detect the JDK (the main purpose of this task), so probably just not even gonna try to merge them, and just keep the one in the child.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -325,7 +345,7 @@ void resolveParentsRecursively(Pom requested) throws MavenDownloadingException {
             if (initialRepositories != null) {
                 mergeRepositories(initialRepositories);
             }
-            resolveParentPropertiesAndRepositoriesRecursively(pomAncestry);
+            resolveParentPropertiesAndRepositoriesRecursively(new ArrayList<>(pomAncestry));
```
joanvr, line 348: this avoid the need to clear the list and populate it after every call to a `resolveParent*Recursively`, which was missing for resolveParentPluginManagementRecursively causing not proper resolution.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -416,10 +433,18 @@ private void resolveParentDependenciesRecursively(List<Pom> pomAncestry) throws
...
+        private void resolveParentPluginsRecursively(List<Pom> pomAncestry) throws MavenDownloadingException {
             Pom pom = pomAncestry.get(0);
 
-            mergePluginManagement(pom.getPluginManagement());
+            for (Profile profile : pom.getProfiles()) {
+                if (profile.isActive(activeProfiles)) {
+                    mergePluginManagement(profile.getPluginManagement());//, pom);
+                    mergePlugins(profile.getPlugins());
+                }
+            }
```
joanvr, line 444: also resolving profiles too

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -470,15 +496,153 @@ private void mergeRequestedDependencies(List<Dependency> incomingRequestedDepend
...
+            Set<GroupArtifact> uniqueDependencies = new HashSet<>();
+            for (Dependency dependency : dependencies) {
+                merged.add(dependency);
+                uniqueDependencies.add(new GroupArtifact(dependency.getGroupId(), dependency.getArtifactId()));
+            }
+            for (Dependency dependency : incomingDependencies) {
+                if (!uniqueDependencies.contains(new GroupArtifact(dependency.getGroupId(), dependency.getArtifactId()))) {
+                    merged.add(dependency);
+                }
+            }
```
joanvr, line 530: deduplicating at merging time, as opposed to dependencies, since I think it's better than adding the logic to `deduplicate()`

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -470,15 +496,153 @@ private void mergeRequestedDependencies(List<Dependency> incomingRequestedDepend
...
+            JsonNode listCombine = list.get("combine.children");
+            if (listCombine != null) {
+                ret.set("combine.children", listCombine);
+            }
+
+            return ret;
+        }
+
+        private List<Plugin.Execution> mergePluginExecutions(List<Plugin.Execution> executions, List<Plugin.Execution> incomingExecutions) {
+            return executions.isEmpty() ? incomingExecutions : executions; // TODO
```
joanvr, line 593: we are not going to use executions right now, but would be a good idea to implement this properly at some point.

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -2300,7 +2302,514 @@ void malformedPom() {
...
+                                    <artifactId>maven-compiler-plugin</artifactId>
+                                    <version>3.11.0</version>
+                                    <configuration>
+                                        <release>11</release>
+                                    </configuration>
+                                </plugin>
+                            </plugins>
+                        </build>
+                    </project>
+                """,
```
timtebeek, line 2326: nitpick: I think we have a recipe that minimizes this leading white space; might be good to apply that here to put the closing `"""` at the same column as `</project>`

timtebeek (approved):
Nicely done! Haven't tested it here, but I trust you've tested it there

PR #3769, opened 2023-12-05
Replace `org.junit.platform.commons.util.StringUtils` with `org.openrewrite.internal.StringUtils`

ammachado:
## What's changed?

Replace StringUtils used by the recipes that create files

## What's your motivation?

Running the recipes with the Maven plugin was showing a `A required class was missing` exception

### Checklist


timtebeek:
Thanks a lot! Not sure how this managed to sneak in; we should ideally not have those classes available on the src/main classpath.. 🤔

PR #3771, opened 2023-12-05
`AddPropery` now accepts a comment

ammachado:
## What's changed?

`AddSpringProperty` from `rewrite-spring` accepts comments, but only applies it's a YAML file. This PR changes `AddProperty` to accept comments.

## What's your motivation?

Add properties to Spring applications.

### Checklist

### DIFF
--- a/rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
+++ b/rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
@@ -58,16 +58,6 @@ public class AddProperty extends Recipe {
     @Nullable
     String delimiter;
 
-    /**
-     * Keeping this constructor just for compatibility purposes
-     * @deprecated Use {@link AddProperty#AddProperty(String, String, String, String)}}
-     */
-    @Deprecated
-    public AddProperty(String property, String value, @Nullable String delimiter) {
-        this(property, value, null, delimiter);
-    }
-
-    @JsonCreator
     public AddProperty(String property, String value, @Nullable String comment, @Nullable String delimiter) {
         this.property = property;
         this.value = value;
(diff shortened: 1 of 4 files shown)


rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
```diff
@@ -41,13 +44,37 @@ public class AddProperty extends Recipe {
...
+    /**
+     * Keeping this constructor just for compatibility purposes
+     * @deprecated Use {@link AddProperty#AddProperty(String, String, String, String)}}
+     */
+    @Deprecated
+    public AddProperty(String property, String value, @Nullable String delimiter) {
+        this(property, value, null, delimiter);
+    }
+
+    @JsonCreator
```
timtebeek, line 70: I could even lean towards removing this old constructor and making the change downstream if needed (which I'm open to do). That way there's no confusion going forward, especially as the three and four argument constructors have a different third argument, which could be confusing.
```suggestion
```

rewrite-properties/src/main/java/org/openrewrite/properties/AddProperty.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.properties;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
```
timtebeek, line 18: Could be removed if we delete the backwards compatible constructor.
```suggestion
```

timtebeek (approved):
Nice addition, thanks! Approved already; let me know how you'd feel about not having a backwards compatible constructor.

ammachado:
> Nice addition, thanks! Approved already; let me know how you'd feel about not having a backwards compatible constructor.

It makes sense. I removed the constructor.

Issue #3745, opened 2023-11-30
MavenDownloadingExceptions in Maven Plugin 5.14.1


api-from-the-ion:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v5.9.9 in Maven plugin
- Maven plugin v5.14.1

## How are you running OpenRewrite?

I am using the Maven plugin, calling from the shell.
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:5.13.0:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE \
  -Drewrite.activeRecipes=org.openrewrite.java.testing.hamcrest.HamcrestMatcherToAssertJ
```

## What is the smallest, simplest way to reproduce the problem?

Code isn't important, it just happens.

## What did you expect to see?

The working plugin.

## What did you see instead?

Exception.

## What is the full stack trace of any errors you encountered?

```
org.openrewrite.maven.MavenDownloadingExceptions
    at org.openrewrite.maven.MavenDownloadingExceptions.append (MavenDownloadingExceptions.java:44)
    at org.openrewrite.maven.tree.MavenResolutionResult.resolveDependencies (MavenResolutionResult.java:179)
    at org.openrewrite.maven.MavenParser.parseInputs (MavenParser.java:113)
    at org.openrewrite.Parser.parse (Parser.java:59)
    at org.openrewrite.maven.MavenMojoProjectParser.parseMaven (MavenMojoProjectParser.java:477)
    at org.openrewrite.maven.MavenMojoProjectParser.listSourceFiles (MavenMojoProjectParser.java:150)
    at org.openrewrite.maven.AbstractRewriteMojo.loadSourceSet (AbstractRewriteMojo.java:258)
    at org.openrewrite.maven.AbstractRewriteMojo.listResults (AbstractRewriteMojo.java:240)
    at org.openrewrite.maven.AbstractRewriteRunMojo.execute (AbstractRewriteRunMojo.java:53)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
[...]

PR #3775, opened 2023-12-06
Bug/3745 maven downloading exceptions wrong maven order

api-from-the-ion:
## What's changed?

I wrote the test for this issue.

## What's your motivation?

Fix this issue
- https://github.com/openrewrite/rewrite/issues/3745

I analyzed this issue and found the culprit. I wrote the test, so everybody could see it. The issue is that we have artifact `a` using artifact `d` without specifying its version. Artifact `b` is the parent of artifact `a` and first imports artifact `c` and then depends on `d`, specifying the version of `d`. Artifact `c` also depends on `d` but specifies an earlier version.

According to [this](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies) we should have a version of `d` in the `a` that is specified in `b`, because `b` is near to the `a` then `c` and therefore overrides the version of `d`. But right now, this test fails.

This is because in ResolvedPom#getManagedVersion we simply go through the list and get the version of the first matched artifact. In our case, it is an imported one because it came earlier in the pom.

The exception comes because the wrong version of the artifact isn't in the repository. It is overridden and not needed by the normal build process. As soon as we correct an issue with a wrongly fetched version, the exception should also disappear.

What we clearly need here is the depth information. We need a map instead of the list, where keys would be the elements of the current list and the value is the distance to the artifact from the current artifact. So we would filter all the matched artifacts, sort them and get the nearest. In case there is more than one artifact with the same near depth, the first one that occurred in the pom should be used according to the description from Maven, see above.

I don't know if #getManagedExclusions, #getManagedScope need the same treatment.

And it would be nice to write this with the stream API.

My problem right now is that because of the end of the year, I wouldn't have enough time for this until January. So if somebody would like to correct this it would be nice.

BTW, I have to add `org.jetbrains.kotlin.jvm` to the Gradle dependencies at least once. Otherwise, this package is missing, and some parts of the OKHTTP tests use it.

Not right now.

### Checklist


timtebeek:
Thanks a lot for getting this started @api-from-the-ion ! Good to have a clear test to replicate the issue. Indeed Maven can be quite particular in the way it resolves dependency versions, and we'll need to replicate that perfectly to fix this issue.

timtebeek:
Went through this again; your test looks really good to replicate the issue; indeed we'd need to take proximity into account when parsing to correctly resolve the version of `d`. If you want we can merge your test as a disabled test first, to prevent conflicts with the main branch, and then iterate on a fix separately.

api-from-the-ion:
Yes, we can merge the test, disabled. Hopefully I would have some time next week to fix this.

I describe once again the real situation here. We have a project with some modules, which use `javax.enterprise:cdi-api` (d). Modules (a) only describe that this dependency is provided, but don't define the version. They all have a project main pom as a parent (b). Here we first import the common pom of all our projects (c); this one declares an older version of the `cdi-api`. But then we directly declare the `cdi-api` dependency in the common parent with a newer version. As a result, we only use a new version of `cdi-api` everywhere in the project. Looks like a direct declaration also overwrites an imported one; in this aspect, an import isn't different from the dependency and has higher depth. So our direct declaration would override the version because it has a lower depth. 

In reality, the older version isn't even downloaded in my cache, which at the end causes the final exception. I hope my test reproduces this correctly.

I once again took a look at the usage of `dependencyManagement`of `ResolvedPom` and decided against the map scenario. `ResolvedManagedDependency` will be extended with proximity property, and then we can use it to group dependencies by the proximity in the cases, where version is used. But in more cases using `dependencyManagement` there are no operations with the version value, so they can remain as they are now, just working with the list without previous grouping.

One more question: I see `mergeDependencyManagement` is only called from the `resolveParentDependenciesRecursively`. So it is filled from the POM relations caused be the parent graph. But where is the merging from the relations of the dependency graph itself? So my project depends on the something, which itself depends on something and so on? I can't find it. This should also be important in this scenario?

api-from-the-ion:
I extended the tests (it wasn't technically complete) and made the fix for exactly this issue. Then I rewrote two other managed methods of the `ResolvedPom` with stream API. Hope this one looks better. This can also be partly reviewed; hope I didn't change the logic and everything else is also OK.

I used a wrong bug ticket reference (twice) in the commit message, but now it's probably too late to change it?

Now I took a look once again at places where this `dependencyManagement` or its getter is used. I see two kinds of usage and connected possible issues:

1. The version will be determined, so it just the same scenario as with my bug here, but the logic is elsewhere, and therefore I didn't get the bug right now. I'll fix it the same way - with usage of proximity information.
2. The cases, where `ResolvedManagedDependency` from `dependencyManagement`itself is a return value of the method. The problem here: we don't know where this object will be used. It is possible, that this object will be used to determine the version, and then we once again will have the same issue. So, my suggestion for this second step: also give the result regarding the proximity.

I'll probably hide the `dependencyManagement` by removing the getter and allowing the access thought some methods. Maybe some method with predicate for filtering, or something else. If this gets too complex or running time consumption will be too high, we can change the list to the map anytime and let the calling methods deal with this change. But in this scenario, we couldn't force the use of the right proximity one.

timtebeek:
Did you mean to close this PR @api-from-the-ion ?

api-from-the-ion:
No - I just got a merge conflict and pressed the wrong button to solve it. But now all is solved and pushed, and we can start a review. I think.

I fixed the described bug, and then another situations like this. And then I have done a little bit cleanup. If you have some comments - just write it here. We can change or drop some things.

rewrite-maven/build.gradle.kts
```diff
@@ -38,6 +39,7 @@ dependencies {
 
     testImplementation(project(":rewrite-test"))
     testImplementation("com.squareup.okhttp3:mockwebserver:4.+")
+//    testImplementation("org.jetbrains:kotlin.jvm:1.9.21")
```
timtebeek, line 42:
```suggestion
    testImplementation("org.jetbrains:kotlin.jvm:1.9.21")
```

rewrite-maven/build.gradle.kts
```diff
@@ -2,6 +2,7 @@ import nl.javadude.gradle.plugins.license.LicenseExtension
 
 plugins {
     id("org.openrewrite.build.language-library")
+//    id("org.jetbrains.kotlin.jvm") version "1.9.21"
```
timtebeek, line 5:
```suggestion
    id("org.jetbrains.kotlin.jvm") version "1.9.21"
```

api-from-the-ion:
Sorry, but I have to comment your changes. You removed the new package from Gradle completely. As I already described, I had problems running tests because this is needed by OKHTTP. So at least before the first run this should be in the kts. After that, one can comment it out, or it could be removed. Both locally. But not for the first initialization.

At the end we don't need both lines, they do just the same but through the different means. But one of them should be left inside, so the next person cloning this project could run the unit tests without such troubles.

timtebeek:
Hi yes those commented out lines stuck out as weird to me; would have to dive in to know exactly what you're referring to, but I'm out part of the day today.

api-from-the-ion:
This is only IntelliJ bug, here is the [link](https://youtrack.jetbrains.com/issue/KTIJ-24991). So using another IDE or command line wouldn't probably cause this issue. But for IJ users it probably would. So if we don't like to have this dependency directly and in old version - we need to document this issue, so the new developers wouldn't have to search for the reason of not running unit tests in the freshly cloned project.

api-from-the-ion:
> Hi yes those commented out lines stuck out as weird to me; would have to dive in to know exactly what you're referring to, but I'm out part of the day today.

Don't want to bother you much, but is it possible to review this further and to merge this? Because I already solved three merge conflicts and there would be some in the future as long as this isn't reviewed and merged.

timtebeek:
> > Hi yes those commented out lines stuck out as weird to me; would have to dive in to know exactly what you're referring to, but I'm out part of the day today.
> 
> Don't want to bother you much, but is it possible to review this further and to merge this? Because I already solved three merge conflicts and there would be some in the future as long as this isn't reviewed and merged.

Hi yes, I really should and thanks for your patience & resolving conflicts while this has been open. It's tough getting to all open PRs, and this one in particular contains a lot of changes so needs more attention.

One thing that I'd spotted is that you've adopted the streams API where we had for loops before. That's something I'm likely to revert in a review, and wouldn't ask of you to change given how much effort you've already put into this PR. While I like the readability aspects of the streams API, on a whole in OpenRewrite we tend to favor simpler loop constructs as they come with fewer allocations, which we're sensitive to in terms of performance. Figure give you that context before you see the change on this PR.

Know that your work is much appreciated! Just need to allocate time for a proper review to get this across the line. 🙏🏻

api-from-the-ion:
> One thing that I'd spotted is that you've adopted the streams API where we had for loops before. That's something I'm likely to revert in a review, and wouldn't ask of you to change given how much effort you've already put into this PR. While I like the readability aspects of the streams API, on a whole in OpenRewrite we tend to favor simpler loop constructs as they come with fewer allocations, which we're sensitive to in terms of performance. Figure give you that context before you see the change on this PR.

Thank you for your replay and explanation. I would like to advocate for the stream changes a little bit. ;-)

Foremost: you are responsible for this product, so the final decision is always yours. ;-)

On the other hand: if we use parallel stream, some operations like filtering should run quicker than in simple for each loop with some ifs. I wouldn't like to introduce this change directly, but we could do it. Because filtering and collecting is really independent here and could be done parallel.

If you still want to change this, then could we replace this with older language constructs like ifs and loops, but still retain the functional interfaces? As I understand you correctly, you don't like the streams in some scenarios because the build up of the stream implementation takes more time than we gain by using them? Beside the better readability? But this isn't an issue with predicates and consumers. I eliminated the repeating code of the metadata with wrapper, so one just see the different part of code at the place of the wrapper's usage - the common code is in the wrapper. And the wrapper API (the interface, the method's parameters) are functional interfaces and totally independent of the inner stream implementation.

I took a look at the current master, and there are some places beside the tests where stream API is used. So this could be a really fresh decision not to use the stream API? I know that nobody does RTFM, but maybe this information for developers should be documented somewhere?

timtebeek:
I didn't get to the above changes before my holiday unfortunately; I'll try to pick this up when I get back!

timtebeek (commented):
Hi @api-from-the-ion ! It's been a while; apologies for the wait and thanks for your patience!

I've recently learned more about how Maven expects managed dependencies to be ordered,  following #4337 . I _think_ the unit test added here has a flaw in the logic.

After reading #4337 it seems to override a BOM version one ought to place the managed dependency before the BOM import that includes that same version. e.g. update the test to:

```java
// a depends on d without specifying version number. a's parent is b
// b imports c first into its dependencyManagement section
// and b depends on d in its dependencyManagement section; here another version of d is specified
// c's dependencyManagement specifies an earlier version of d to use
// So if all goes well we will have the version of d from b's dependencyManagement because it's nearest to our pom
// and therefore overrides the version from c
@Issue("https://github.com/openrewrite/rewrite/issues/3745")
@Test
void versionConflictBetweenParentAndImportedFromParent() {
    rewriteRun(
      mavenProject("parent-b",
        pomXml(
          """
                <project>
                    <artifactId>parent-b</artifactId>
                    <version>0.1.0-SNAPSHOT</version>
                    <packaging>pom</packaging>
                    <properties>
                        <version.d>0.2.0-SNAPSHOT</version.d>
                    </properties>
                    <dependencyManagement>
                        <dependencies>
                            <dependency>
                                <artifactId>d</artifactId>
                                <version>${version.d}</version>
                            </dependency>
                            <dependency>
                                <artifactId>c</artifactId>
                                <version>0.1.0-SNAPSHOT</version>
                                <type>pom</type>
                                <scope>import</scope>
                            </dependency>
                        </dependencies>
                    </dependencyManagement>
                </project>
            """
        )
      ),
      mavenProject("child-a",
        pomXml(
          """
                <project>
                    <parent>
                        <artifactId>parent-b</artifactId>
                        <version>0.1.0-SNAPSHOT</version>
                        <relativePath />
                    </parent>
                    <artifactId>child-a</artifactId>
                    <dependencies>
                        <dependency>
                            <artifactId>d</artifactId>
                        </dependency>
                    </dependencies>
                </project>
            """,
          spec -> spec.afterRecipe(pomXml ->
            assertThat(pomXml.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow().getDependencies()
              .get(Scope.Compile).get(0).getVersion())
              .isEqualTo("0.2.0-SNAPSHOT")
          )
        )
      ),
      mavenProject("c",
        pomXml(
          """
                <project>
                    <artifactId>c</artifactId>
                    <version>0.1.0-SNAPSHOT</version>
                    <packaging>pom</packaging>
                    <dependencyManagement>
                        <dependencies>
                            <dependency>
                                <artifactId>d</artifactId>
                                <version>0.1.0-SNAPSHOT</version>
                            </dependency>
                        </dependencies>
                    </dependencyManagement>
                </project>
            """
        )
      ),
      mavenProject("d-v1",
        pomXml(
          """
                <project>
                    <artifactId>d</artifactId>
                    <version>0.1.0-SNAPSHOT</version>
                </project>
            """
        )
      ),
      mavenProject("d-v2",
        pomXml(
          """
                <project>
                    <artifactId>d</artifactId>
                    <version>0.2.0-SNAPSHOT</version>
                </project>
            """
        )
      )
    );
}
```

When I run the above test without any further changes to `src/main` we _do_ see the correct expected `d:0.2.0-SNAPSHOT` version. Do you agree with that analysis?

timtebeek:
If you do agree, then given the above I'd propose we close this PR. As much as the appreciate the work you've put in for these refactorings, I think it's best we keep the original logic as is to match how Maven picks the first managed version. We're also considering an extension to `org.openrewrite.maven.RemoveRedundantDependencyVersions`  to clear out such unused dependencyManagement entries, to help prevent any such confusion. https://docs.openrewrite.org/recipes/maven/removeredundantdependencyversions

timtebeek:
Closing this PR based on the last two comments above, and due to the number of conflicts this has developed since. Thanks again for helping explore this issue @api-from-the-ion !

Issue #3783, opened 2023-12-07
MethodParamPad breaks Jenkinsfile


hazendaz:
Using static analysis 1.1.2 and has been occurring last few versions best I can tell.  

Using rewrite-maven-plugin 5.15.2

Only change picked up in repo I'm working with is that it touched my Jenkinsfile unexpectedly.  It has taken the trailing space after every occurrence of 'library ' which causes it to be squashed to the next word and thus breaks Jenkins.

Only two recipe's ran as part of this

org.openrewrite.staticanalysis.CodeCleanup
org.openrewrite.java.RemoveUnusedImports

Broken items

```library '...'``` becomes ```library'...'```
```library identifier``` becomes ```libraryidentifier```

timtebeek:
hi @hazendaz ; sorry to hear about these issues! I've moved the issue as I think the problem lies in our Groovy handling, not as much the individual recipe. We had a similar report the other day in #3777 and #3737, after we started parsing Jenkinsfiles as Groovy in https://github.com/openrewrite/rewrite/commit/4d74818d9d76eed72af6fc57b937ee6913ebfc57 and https://github.com/openrewrite/rewrite/commit/7ed7222b17f689d57c9392fd6430f82fe301da1b. I've asked the team if there's a common fix we can apply to get this resolved, such that you can still target Jenkinsfiles with Java/Groovy recipes, without these issues.

As a workaround you can force Jenkinsfiles to be parsed as text by passing in a plaintext mask to the OSS Maven and Gradle plugins, or excluding Jenkinsfiles from parsing for now while we work through this issue.

hazendaz:
@timtebeek This issue is still not resolved.  I think your PR that you closed #3789 would have solved by simply stop touching Jenkinsfile.  Instead comment was made that others fixes correct but none of that did.  I've updated every single version since including that from tonight and its still a problem right now.

timtebeek:
Sounds like a very similar case to what we discussed over on https://github.com/openrewrite/rewrite-static-analysis/issues/233#issuecomment-1902634285
That was fixed in this PR, which I think applies here as well; thanks again for bringing this up and testing intermediate versions!
- https://github.com/openrewrite/rewrite/pull/3940

PR #3790, opened 2023-12-07
Don't remove leading space on omitted parentheses

knutwannheden:
In Groovy method calls don't always have parentheses around the arguments. In that case the space between the method name and first argument must not be removed.

Fixes: #3783


knutwannheden:
Isn't quite this easy unfortunately. The Groovy parsing would indeed have to be overhauled to for example also handle trailing lambdas.

PR #3793, opened 2023-12-08
Fix for AddPlugin when plugin version is managed

ammachado:
## What's changed?

`AddPlugin` throws exception if the plugin being added is already present on pom.xml and it's version is managed by a parent pom.

No

No one.

### Checklist


timtebeek (approved):

timtebeek:
Makes sense, thanks a lot!

PR #3794, opened 2023-12-08
Skip updating Gradle when already updated

shanman190:
## What's changed?
Skip performing extra work when the current Gradle version was updated earlier in the cycle.

## What's your motivation?
Overall reduce what would be considered duplicate work due to ordering within the recipe stack. For most cases, the stack will include Gradle updates that incrementally go higher. However, users may pull in recipes in varying orders, so we should ensure that we do the minimum amount of work necessary.

Any of: @jkschneider @sambsnyd @timtebeek 

Only one really is to try to strictly order recipes in the recipe stack as a user, so that things always go up and never down. Though this can be harder than it sounds.

### Checklist


timtebeek (approved):
Looks good to me; thanks for adding this logic!

PR #3795, opened 2023-12-08
Find JavaVersion markers on any source file

timtebeek:
## What's changed?
Let `HasJavaVersion` return a `TreeVisitor` instead of a `JavaIsoVisitor`

## What's your motivation?
Set a property for use of virtual threads if on Java 21, with a non-scanning recipe.
- https://github.com/openrewrite/rewrite-spring/pull/446

Necessary for proper functioning of this recipe & tests:
- https://github.com/openrewrite/rewrite-spring/pull/460

## Any additional context
We add the Java version marker to all source files in
https://github.com/openrewrite/rewrite-maven-plugin/blob/1d205eac8f43a17395ed317b8e8984aa25990a59/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L296


knutwannheden (approved):
Note that the marker won't get added processed by the native build, but I guess that won't be an issue.

PR #3801, opened 2023-12-09
Fix `StackOverflowError` in `MavenPomDownloader`

knutwannheden:
When building gridgain/gridgain@main there is a `StackOverflowError` because the root project references `parent/pom.xml` as its parent and the latter contains `<relativePath/>`. But that actually means that Maven should *not* look for the parent in the current project.

It seems difficult to get this 100% correct right now, as we with our current model we deserialize using Jackson into `RawPom.Parent` can't tell the difference between an absent `<relativePath>` element and a present but empty `<relativePath>` element, even though that is significant in Maven.

As a shortcut we can check if the parent (in the file system) has a matching GA or not.


knutwannheden:
It seems difficult to get this 100% correct right now, as we with our current model we deserialize using Jackson into `RawPom.Parent` can't tell the difference between an absent `<relativePath>` element and a present but empty `<relativePath>` element, even though that is significant in Maven.

timtebeek (approved):

PR #3804, opened 2023-12-11
YamlParser fails print idempotency for values containing `:@a@`

Laurens-W:
## What's changed?

Added a testcase for a short piece of yaml that should be parsed correctly (Spring Boot correctly interprets this)

## What's your motivation?

Ran into an issue where the application.yaml isn't parsed because of a file reference that uses `@project.basedir@`

Is this valid yaml?

@timtebeek 

Referring to the file as a classpath resource works like a charm

### Checklist


rewrite-yaml/src/test/java/org/openrewrite/yaml/YamlParserTest.java
```diff
@@ -129,4 +129,14 @@ void unicodeEscapes() {
...
         );
     }
+
+    @Test
+    void projectBasedir() {
+    rewriteRun(
+        yaml(
+            """
+              server:
+                  keyStore: file:@project.basedir@/src/test/resources/keystore.jks
```
timtebeek, line 139: Fails with
```
org.opentest4j.AssertionFailedError: [Bug in source parser or printer resulted in the following difference for "file.yaml"] 
expected: 
  "server:
      keyStore: 'file:@project.basedir@/src/test/resources/keystore.jks'"
 but was: 
  "server:
      keyStore: 'file:1fead784-d4ff-48b9-a9f6-34edbb297d12/src/test/resources/keystore.jks'"
```

timtebeek: Tried a few iterations:
- no improvement when quoting the value
- no improvement when using `@foo@`
- `@@` works,
- `@foo` works, as does `foo@`

timtebeek:
Still fails with these two failures on that last two tests added.
```diff
diff --git a/file.yaml b/file.yaml
index 9e749ca..3812502 100644
--- a/file.yaml
+++ b/file.yaml
@@ -1 +1 @@ 
-a: b:@c@d
\ No newline at end of file
+a: b:f73fe741-95d9-479a-9829-235836e2a6e2
\ No newline at end of file
```

```diff
diff --git a/file.yaml b/file.yaml
index 649c05f..47d68eb 100644
--- a/file.yaml
+++ b/file.yaml
@@ -1 +1 @@ 
-a: 'b:@c@d'
\ No newline at end of file
+a: 'b:3002553c-c619-48c3-a1b6-c2d70f833700d'
\ No newline at end of file
```

Laurens-W:
@timtebeek maybe we could still merge this to have the regression test case?

timtebeek:
Had a brief look, but to me the test that was added for #4917 looks very similar, especially the first line has the same failure mode:
https://github.com/openrewrite/rewrite/blob/bf0519d9be4ae246cf2164f87ca69fc6936f72fa/rewrite-yaml/src/test/java/org/openrewrite/yaml/YamlParserTest.java#L224-L231

To me that seems then sufficiently covered by that other existing test, but do let me know if you feel I've missed something.

Laurens-W:
I hadn't looked in detail as Greg contacted me and I left it up to him, but yeah that sufficiently covers the initial issue I ran into!

PR #3812, opened 2023-12-12
Add `J.NullableType` for Kotlin / C#

kunli2:
Languages such as Kotlin, JavaScript, and C# support nullable types denoted by appending a ? after the type, indicating potential nullability. Introducing J.NullableType to accommodate these scenarios.
The according change for Kotlin to deprecate `IsNullable` marker: https://github.com/openrewrite/rewrite-kotlin/pull/497


rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
```diff
@@ -843,6 +843,16 @@ public J visitNewClass(NewClass newClass, PrintOutputCapture<P> p) {
         return newClass;
     }
 
+    @Override
+    public J visitNullableType(J.NullableType nt, PrintOutputCapture<P> p) {
+        beforeSyntax(nt, Space.Location.UNARY_PREFIX, p);
```
knutwannheden: Should we use `NULLABLE_PREFIX` here?
```suggestion
        beforeSyntax(nt, Space.Location.NULLABLE_PREFIX, p);
```

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
```diff
@@ -843,6 +843,16 @@ public J visitNewClass(NewClass newClass, PrintOutputCapture<P> p) {
...
+        visit(nt.getTypeTree(), p);
+        visitSpace(nt.getPadding().getTypeTree().getAfter(), Space.Location.UNARY_OPERATOR, p);
```
knutwannheden: Then probably also a dedicated `NULLABLE_OPERATOR`.

Although, possibly we should instead throw an `UnsupportedOperationException` and require extending languages to implement 🤷

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -997,6 +997,24 @@ public J visitNewClass(J.NewClass newClass, P p) {
         return n;
     }
 
+    public J visitNullableType(J.NullableType nullableType, P p) {
```
traceyyoshima, line 1000: Info: any addition to a tree where the visitor is added should also have an associated visit in the ISO visitor.

Please add the visit to JavaIsoVisitor as well

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -4423,6 +4423,92 @@ public NewClass withArguments(JContainer<Expression> arguments) {
...
+
+        JRightPadded<TypeTree> typeTree;
+
+        public TypeTree getTypeTree() {
+            return typeTree.getElement();
+        }
+
+        @Override
+        public @Nullable JavaType getType() {
+            return typeTree.getElement().getType();
```
knutwannheden, line 4457: Hmm. This will actually be wrong for Kotlin, as the nullability is lost in the type attribution. Possibly we already now have to add `JavaType.Nullable` or alternatively some kind of `nullable` flag.

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
kunli2: this has been fixed.

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
kunli2: updated to `Space.Location.NULLABLE_PREFIX`

kunli2: @traceyyoshima  what do you think about this, does JS and C# print it the same way? 
if the answer is yes, we probably want to have a shared print method here.

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
kunli2: added to `JavaIsoVisitor`

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
kunli2: do you mean like add a `JavaType.NullableType` like this
```java
    @RequiredArgsConstructor
    @Getter
    @With
    class NullableType implements JavaType {
        private JavaType raw;
    }
```
and changed here to

```java
        @Override
        public @Nullable JavaType getType() {
            return new JavaType.NullableType(typeTree.getElement().getType());
        }
```

knutwannheden: Either that (although for consistency it should probably just be called `Nullable`) or have it encoded in `JavaType.Class#flagsBitMap`. I think the latter might not work, as we can in Kotlin also have something like
```kt
fun <T> foo(): T? = ...
```
where a generic type variable is nullable.

But `Nullable` should also extend `JavaType.FullyQualified` otherwise there will be a lot of recipes that won't work anymore.

Let's see if we can get @sambsnyd to chime in on this.

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
traceyyoshima: C# shares the same syntax: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#declaration-and-assignment.

A `?` in TS represents an [optional property](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties), nullability is defined by using a union type.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
traceyyoshima: I think extending `JavaType$FullyQualified` (FQ) may lead to unexpected results when applied to a `JavaType$GenericTypeVariable` (GTV). The `FullyQualified` class contains a lot of methods that do not apply to a GTV. It's relatively common to check if the type is an `FQ` and progress further with a condition in a recipe, so I'd be cautious.

knutwannheden: Let's leave the type attribution as is for now. I am thinking we should maybe instead add annotations to the type attribution, as I suspect that is what Kotlin also uses in the bytecode and would also work for the various nullable annotations out there. But this is a bigger piece of work outside the scope of this PR.

knutwannheden: AFAICT the Java bytecode indeed contains `org.jetbrains.annotations.Nullable` annotations, so having that available in the type attribution is probably the way forward.

PR #3817, opened 2023-12-13
Add new `SourceFileService`

knutwannheden:
The service implements the methods `computeWeight()` and `newVisitor()`. This is then in turn used by `SourceFile#getWeight()` with the idea that it will automatically use a visitor that doesn't need to be adapted using `TreeVisitorAdapter`, which behind the scenes generates a new class.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -17,13 +17,14 @@ package org.openrewrite;
 
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.service.SourceFileService;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.style.Style;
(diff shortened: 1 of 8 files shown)


knutwannheden:
Closing for now.

JLLeitschuh:
Use cases for services:
 - Control Flow analysis will be, likely, unique per-language as the characteristics around dataflow are unique based upon language.
 - Custom logic that surrounds printing the AST. For example:
   - Control Flow Analysis: https://github.com/openrewrite/rewrite-analysis/blob/main/src/main/java/org/openrewrite/analysis/controlflow/ControlFlowJavaPrinter.java
   - Detecting AST elements by line/column number: https://github.com/openrewrite/rewrite-analysis/pull/41

knutwannheden:
For control flow analysis we would likely add another dedicated service. We already have a few services, but for now we have limited this mechanism to `J` and extending languages. But I think it is soon time to pull this up into the core.

@JLLeitschuh Thanks for your use cases.

PR #3822, opened 2023-12-14
Maven POM downloader optimizations

ammachado:
## What's changed?

This PR introduces two small optimizations to `MavenPomDownloader`:
1. Add a trailing '/' to Maven repository URL checks to avoid HTTP 302 redirects on slow Artifactory instances. Also tag the repository URL as known, if able to connect to it.
2. When deriving metadata for artifacts present on the local maven repository, consider only successfully downloaded artifacts.

This PR also fixes the counting of `deriveMetadata` successful calls.

## What's your motivation?

Increased latency when working on network restricted environments with slow Artifactory proxies.

No.

No workarounds for environment issues

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
@@ -65,6 +65,7 @@ public class MavenRepository implements Serializable {
 
     @EqualsAndHashCode.Include
     @With
+    @NonFinal
     boolean knownToExist;
 
     // Prevent user credentials from being inadvertently serialized
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
@@ -63,6 +63,8 @@ public class MavenRepository implements Serializable {
     @Nullable
     String snapshots;
 
+    @EqualsAndHashCode.Include
+    @With
     @NonFinal
     boolean knownToExist;
 
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
@@ -65,7 +65,6 @@ public class MavenRepository implements Serializable {
 
     @EqualsAndHashCode.Include
     @With
-    @NonFinal
     boolean knownToExist;
 
     // Prevent user credentials from being inadvertently serialized
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepositoryMirror.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepositoryMirror.java
@@ -104,7 +104,7 @@ public class MavenRepositoryMirror {
     }
 
     public MavenRepository apply(MavenRepository repo) {
-        if (repo.getUri().equals(url) && Objects.equals(id, repo.getId()) || !matches(repo)) {
+        if (Objects.equals(id, repo.getId()) && (repo.getUri().equals(url) || repo.isKnownToExist()) || !matches(repo)) {
             return repo;
         } else {
             MavenRepository repoWithMirror = repo.withUri(url)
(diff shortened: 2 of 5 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
```diff
@@ -63,6 +63,8 @@ public class MavenRepository implements Serializable {
     @Nullable
     String snapshots;
 
+    @EqualsAndHashCode.Include
```
ammachado, line 66: I think that this field can be excluded from `equals`/`hashCode`.

timtebeek: Would you then want to remove this line again, given the `@EqualsAndHashCode(onlyExplicitlyIncluded = true)` at on the class?
```suggestion
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
```diff
@@ -63,6 +63,8 @@ public class MavenRepository implements Serializable {
...
+    @With
     @NonFinal
```
timtebeek, line 68: I'd expect the `@With` to imply `@NonFinal`, and don't see that on other fields that have `@With` in this class.
```suggestion
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepositoryMirror.java
```diff
@@ -104,7 +104,7 @@ public static MavenRepository apply(Collection<MavenRepositoryMirror> mirrors, M
     }
 
     public MavenRepository apply(MavenRepository repo) {
-        if (repo.getUri().equals(url) && Objects.equals(id, repo.getId()) || !matches(repo)) {
+        if (Objects.equals(id, repo.getId()) && (repo.getUri().equals(url) || repo.isKnownToExist()) || !matches(repo)) {
```
timtebeek, line 107: I'm having a little trouble following the addition of `repo.isKnownToExist())` here. Could you add some context as to why this is added? This seems to not apply a mirror if a original repository exists, but I don't fully understand why. I'm thinking of scenarios where for instance the mirror limits outbound traffic, where both might exist, but the mirror is still preferred.

It might just be my unfamiliarity with the classes here; from all your earlier work I've come to trust your judgement; just want to be sure I understand why, and that the changes are as intended.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
ammachado: The issue here is that you can have the same repository defined as a `<mirror>` or as a `<repository>` on a Maven's `settings.xml` file. You can't set the release/snapshot policy in a `<mirror>` configuration.

ammachado: Here's a `settings.xml` that shows the issue:

```xml
<settings xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd" xmlns="http://maven.apache.org/SETTINGS/1.1.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <mirrors>
    <mirror>
      <mirrorOf>*,!snapshot-repos</mirrorOf>
      <id>release-repos</id>
      <name>maven-release-repos</name>
      <url>https://my.artifactory/artifactory/maven-release-repos/</url>
    </mirror>
  </mirrors>
  <profiles>
    <profile>
      <id>artifactory</id>
      <repositories>
        <repository>
          <id>release-repos</id>
          <name>maven-release-repos</name>
          <url>https://my.artifactory/artifactory/maven-release-repos/</url>
          <snapshots>
            <enabled>false</enabled>
          </snapshots>
        </repository>
        <repository>
          <id>snapshot-repos</id>
          <name>maven-snapshot-repos</name>
          <url>https://my.artifactory/artifactory/maven-snapshot-repos/</url>
          <releases>
            <enabled>false</enabled>
          </releases>
          <snapshots>
            <enabled>true</enabled>
          </snapshots>
        </repository>
      </repositories>
    </profile>
  </profiles>
  <activeProfiles>
    <activeProfile>artifactory</activeProfile>
  </activeProfiles>
</settings>
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepositoryMirror.java
ammachado: `knownToExist` is a flag to test if the repo URL a) had any property substituted; and b) is reachable, in case of a HTTP[S] remote repository. Looks like this was used as some sort of cache shortcut mechanism, but there's also a cache in place for it. This addition avoids any further testing if this repo was tested before.

A few lines below this change, `knownToExist` is set to `false` in case of the URL is replaced by a mirror, and this is the only place this field was set before this change.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
ammachado: The annotation is required because Lombok considers it as a required constructor argument if it's not present. Changing the field to `Boolean` (capital B) would be a better (and bigger) fix.

timtebeek:
Thanks a lot @ammachado !

Issue #3807, opened 2023-12-11
Add glob curly-brace support to FindSourceFiles


nmck257:
Eg `{.md,.txt}`, like `HasSourcePath` used to support

_Originally mentioned by @timtebeek in https://github.com/openrewrite/rewrite/issues/3755#issuecomment-1838623921_

timtebeek:
Agree it would be best to reinstate those optional selectors. In the move from `HasSourcePath` to `FindSourceFiles` we now use `PathUtils`, which does not support that yet, but is stable across operating systems.
https://github.com/openrewrite/rewrite/blob/bbc753423476a5ec557ff70c64593eb28d1c406f/rewrite-core/src/main/java/org/openrewrite/PathUtils.java#L66-L82

nmck257:
@timtebeek - semi-related - would you have any concerns if I widened `FindSourceFiles` `filePattern` to `@Nullable`, and moved logic like this into `FindSourceFiles` directly?
https://github.com/openrewrite/rewrite/blob/b19730693662b69321980e2559a7cab90ccd2512/rewrite-xml/src/main/java/org/openrewrite/xml/RemoveXmlTag.java#L51

At the same time, I'd probably adjust such that empty-string causes that same "match everything" behavior; currently, empty-string seems to match nothing, and afaik, most glob specifications interpret empty-string as "match everything" (and that's also how `HasSourcePath` worked)

timtebeek:
Seems reasonable to me, yes thanks!

nmck257:
I'm also noticing that the old `HasSourcePath` relied on a `PathMatcher` from the `FileSystem`, whereas `FindSourceFiles` uses a homegrown glob implementation in `PathUtils`.

Do you happen to know if our homegrown impl has desirable qualities (perf, etc) over the "native" impl?

Looks like @shanman190 wrote much of the `PathUtils::matchesGlob`

shanman190:
@nmck257 it's more like fixed up a bit. Haha

The biggest part is that the Java `PathMatcher` is OS-specific due to being `FileSystem` specific, so this means that for the OSS build plugins everything works just fine as all LST and recipe execution happens on the immediate host OS, but when you have to have interoperability (such as the Moderne SaaS) between operating systems then it begins to fail without some massaging/assistance.

nmck257:
Ah, that's good context. Okay, then adding curly-brace support to our custom `PathUtils` impl feels like the right choice.

[...]

PR #3823, opened 2023-12-14
filePattern on FindSourceFiles is now nullable, and matches all files…

nmck257:
… on blank input, to make it a little more convenient to use in place of HasSourcePath

## Any additional context

Brief discussion over here https://github.com/openrewrite/rewrite/issues/3807#issuecomment-1850976790

### Checklist

### DIFF
--- a/rewrite-test/src/test/java/org/openrewrite/FindSourceFilesTest.java
+++ b/rewrite-test/src/test/java/org/openrewrite/FindSourceFilesTest.java
@@ -18,6 +18,7 @@ package org.openrewrite;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullSource;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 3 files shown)


rewrite-test/src/test/java/org/openrewrite/FindSourceFilesTest.java
```diff
@@ -76,7 +77,8 @@ void windows() {
     }
 
     @ParameterizedTest
-    @ValueSource(strings = {"hello.txt", "/hello.txt", "\\hello.txt", "./hello.txt", ".\\hello.txt"})
+    @ValueSource(strings = {"hello.txt", "/hello.txt", "\\hello.txt", "./hello.txt", ".\\hello.txt", ""})
+    @NullSource
     void findRoot(String filePattern) {
```
timtebeek, line 82: Two failures in this test it seems https://ge.openrewrite.org/s/akobtpkwdhh44/tests/overview?outcome=FAILED

nmck257: will check back over; was having issues with my local build the other day

nmck257: fixed; that's what I get for only skimming the test case I was adding on to :)

timtebeek (approved):

rewrite-test/src/test/java/org/openrewrite/FindSourceFilesTest.java
timtebeek: All good, thanks for the quick fix!

timtebeek:
Thanks again! Good to see this match-all option restored as compared to HasSourcePath before.

PR #3827, opened 2023-12-16
`SemanticallyEqual`: Fix bug in `visitFieldAccess()`

knutwannheden:
When comparing `J.FieldAccess` instances, the targets also needs to be compared.


timtebeek (approved):

knutwannheden:
The fix I made is not quite right yet. I will look into it next week.

PR #3829, opened 2023-12-16
Add GoToLine Recipe + Utility Methods

JLLeitschuh:
Adds a recipe that will find the LST nodes in the tree at a given
line/column.

This is useful for matching the OpenRewrite LST against outputs from
other scanners, for example security scanners that generate SARIF files.

## What's changed?

Added a `GoToLine` recipe.

## What's your motivation?

Being able to find LST nodes by their line/column coordinates can be helpful when matcing up the LST against scanners.

Are there any unit tests that I'm missing that I should really make sure I include?

This strategy, unfortunately, doesn't inherintly support other languages. Is there any strategy that would make this multi-language?

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/GoToLine.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/GoToLine.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.search;
+
+import lombok.*;
+import org.openrewrite.*;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaPrinter;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.SearchResult;
+
+import java.util.*;
+
+@Incubating(since = "8.12.1")
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class GoToLine {
+    /**
+     * Set to true to print the file as it is being searched. Useful for debugging.
+     * The JIT will completely remove code flagged behind this when set to false because it is unreachable,
+     * and this is a constant.
+     */
+    private static final boolean debug = false;
+
+    /**
+     * Find all elements in the LST at the given line.
+     * <p>
+     * <strong>NOTE:</strong> line number is 1-based, which matches the behavior of most editors.
+     * </p>
+     *
+     * @param sourceFile The source file to search.
+     * @param line       The line number to search. 1-based.
+     * @return The elements found at the given line, or an empty collection if no elements were found.
+     */
+    public static Collection<J> findLine(JavaSourceFile sourceFile, int line) {
+        if (line < 1) {
+            throw new IllegalArgumentException("Line numbers must be 1-based");
+        }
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
Supersedes https://github.com/openrewrite/rewrite-analysis/pull/41

knutwannheden:
@JLLeitschuh I am surprised that we provide a recipe for this. What is the usefulness of that? Especially since the recipe doesn't have a path option for the file. Would it not make more sense to only have a visitor?

JLLeitschuh:
> Especially since the recipe doesn't have a path option for the file.

I figure if you want a path option that should, likely, be it's own recipe that flags the file as a precondition, then this is applied.

JLLeitschuh:
Screenshot 2023-12-16 at 1 56 12 AM

I was implying that this should be a recipe from this comment. But I guess it was never explicitly stated by @jkschneider.

Idk. I thought it may be useful to demonstrate the capability, but I'm not attached to keeping the functionality in a recipe. The recipe needs to exist in some form for unit testing though.

If it's not a recipe, where would the API live? I had hoped on `SourceFile` but that's in `core`, which I wouldn't be able to do without services for other languages.

knutwannheden:
But when would you be using this recipe as is? I thought your use case would be programmatic use. How would you be using this?

Having line and column number options but no path option seems kind of weird.

knutwannheden:
For unit testing you can just use the `toRecipe()` utility which wraps a visitor into a recipe.

knutwannheden:
I think we should move this code into the core as soon as we have agreed on how this services API should look like. That has now also been incubating in rewrite-java for some time.

JLLeitschuh:
> But when would you be using this recipe as is?

Quick and dirty testing on the SaaS. But in general, not often. It doesn't need to be a recipe. I'd like a suggestion on where it _should_ live though.

knutwannheden:
I think the visitor might be interesting as a separate utility. If we additionally want a recipe, I think we should also have a path option.

Apart from that I think the code is in the right place. The service is something we would build on top.

rewrite-java/src/main/java/org/openrewrite/java/tree/Space.java
```diff
@@ -444,7 +445,17 @@ public enum Location {
         WHILE_PREFIX,
         WILDCARD_BOUND,
         WILDCARD_PREFIX,
-        YIELD_PREFIX,
+        YIELD_PREFIX;
+
+        @Incubating(since = "8.11.3")
+        public boolean isPrefix() {
+            return this.toString().endsWith("_PREFIX");
```
knutwannheden, line 452: Do we get a lot of false positives without this?

JLLeitschuh: The logic does not work correctly without a check to see if the `Location` is a prefix.

Without it, we end up findings AST elements on other lines

rewrite-java/src/main/java/org/openrewrite/java/search/GoToLine.java
```diff
@@ -0,0 +1,245 @@
...
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaPrinter;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.SearchResult;
+
+import java.util.*;
+
+@Incubating(since = "8.11.3")
```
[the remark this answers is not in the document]
JLLeitschuh: Resolved! Much appreciated!

JLLeitschuh:
@knutwannheden PTAL at the changes I made yesterday and let me know if they are good for you to merge

JLLeitschuh:
@timtebeek anything that can be done to get this over the finish line?

timtebeek:
Hi @JLLeitschuh ! Thanks for proposing to add this upstream; after some more discussion we feel this would fit better in rewrite-analysis or rewrite-java-security after all. Sorry about any rework; I know you started out there originally. Look forward to seeing it there, and what you'll build once it's available.

JLLeitschuh:
Much appreciated. Any qualms on accepting the changes to `Space` for `isPrefix' and `isSuffix`?

timtebeek:
> Any qualms on accepting the changes to `Space` for `isPrefix' and `isSuffix`?

Given the coordination required for any changes to tree packages, and our general lean towards a minimal public API it might be best to keep the pre/suffix checks within the GoToLine utility too, at least for now. Thanks again!

JLLeitschuh:
Fair enough. Especially because, as of currently, this is likely fairly niche for now.

@knutwannheden do you have insights into how stable the `[LANGUAGE]Printer`'s are for all the other languages? In order for this functionality to work for multiple languages, I'm going to have to build on top of those, so having them break out from under `rewrite-analysis` would be annoying

PR #3831, opened 2023-12-16
Option to add classifier to managed dependency

### LINKED ISSUE #3830: Maven: Add ChangeManagedDependencyClassifier
Currently there is https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyClassifier.java

Which changes a Maven Dependency classifier for example:

**From:**
```xml
<dependency>
	<groupId>org.primefaces</groupId>
	<artifactId>primefaces</artifactId>
	<version>13.0.3</version>
</dependency>
```

**To: (Add classifier jakarta)**
```xml
<dependency>
	<groupId>org.primefaces</groupId>
	<artifactId>primefaces</artifactId>
	<version>13.0.3</version>
        <classifier>jakarta</classifier>
</dependency>
```

However there is no `ChangeManagedDependencyClassifier` to change a managed dependency classifier?  Ideally the above JAVA should do both at the same time or a new `ChangeManagedDependencyClassifier` could be introduced which is exactly like `ChangeDependencyClassifier` which a few lines of code changed to check for Managed depenency.

timtebeek:
## What's changed?
Added `@Option Boolean includeManaged` to `.../maven/ChangeDependencyClassifier`.

## What's your motivation?
Fixes #3830

@melloware

Separate recipe seems too heavy-handed; 

## Any additional context
- #3830

## TODO 
- [ ] Only add in managed when present
- [ ] Possibly do the same for Gradle


timtebeek:
Not quite pleased with this approach yet: ideally we detect if a dependency is managed (in a pom part of the git repository), and only then add the classifier there, without passing in an explicit option toggle. 🤔

melloware (approved):
I like this better than not having it at all!

melloware (approved):

timtebeek:
I'm going to close this attempt; the downside of having something sub-optimal is that folks will build it into their own recipes, making it hard to remove. I think instead we should opt for detecting if a dependency is managed, and if so add the classifier to the managed dependency if we can, and if we can not then add the classifier where it's used instead. I'll close this attempt for now .

PR #3834, opened 2023-12-19
Add utility to print recipe tree

ammachado:
## What's changed?
Added a method on `RewriteTest` to print the nested recipe tree for a given recipe

## What's your motivation?
While working on a large declarative recipe, I identified a few recipes being called more than once. This list the recipe tree and helps identify the issue.

No

No one

No

## Any additional context
No

### Checklist


timtebeek:
I like the initiative here @ammachado and can definitely see the value in being able to easily print the recipe tree from a test. The only thing that gives me pause is the amount of new public API surface that we're adding and would need to support going forward, and how that's used from a user's perspective.

In the current version we'd add five new methods to `RewriteTest`, all of which can be overridden by end users. That also means, that even if you only want to print a recipe tree once, you need to override a method and remove that again. And while there is precedence in RewriteTest of having methods to override, it might not be ideal going forward.

What would you say if we instead add a `print(Consumer<String>)` method to [RecipeSpec](https://github.com/openrewrite/rewrite/blob/d0d73e65bf63aea697ec29756d8df82962a2fb8a/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java#L44)? That's already where we configure the recipe, such that it might feel natural to tag on a print call there. We can then call that consumer in [RewriteTest.rewriteRun](https://github.com/openrewrite/rewrite/blob/d6fe17761edb4f236b1d5b201e9b9d737e751cce/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java#L156-L159), such that it's printed just before it's executed. We'd only be adding a single new method to the API, while still allowing for reasonably flexible use through the `Consumer<String>`.

Note that I'm not saying to change anything just yet, but opening a discussion and curious to hear your thoughts.

shanman190:
Just as an expansion question @timtebeek, might there be a possibility that someone might want to see the recipe materialized? 

In specific, I'm wondering if there might be a way to just pass the recipe into a consumer itself, then have a utility method that prints the recipe stack. This would be kinda similar to what I did previously with the Gradle `withToolingApi(..)` expansion. So it may unlock even more use cases beyond just recipe printing itself.

timtebeek:
Certainly an interesting suggestion, and indeed the one with the most flexibility. I do wonder about potential misuse though, as it might lead to a discoverable but incorrect API use if folks use something like `RecipeSpec.consume(Consumer<Recipe>)` and then start to call `.run(..)` or `.validate(..)`, or don't correctly handle `AdHoc*Recipe`, `CompositeRecipe` and other special cases. Guess it goes to show that even comparatively small additions can have interesting effects. Not quite sure yet what's the right way forward, but do really appreciate both of your's input.

ammachado:
@timtebeek I thought about it, but just took the shortest route to see how it would look in an official version. I can find some time to work on a better implementation. In my original implementation, I only printed `DeclarativeRecipe` and `CompositeRecipe` if the compound recipe had a single recipe grouped together.

I originally used this to debug which ended up being fixed in https://github.com/openrewrite/rewrite/pull/3822, one of the dependencies was being looked up multiple times so I discovered I had a duplicate recipe in the recipe tree.

timtebeek:
Thanks @ammachado ! I've gone over what you've added and played around a bit to see if I could come up with a more minimal API, without too much loss of function. Here's what I came up with, but intentionally hadn't pushed yet as to not undo your work.

```diff
diff --git a/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java b/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
index e0f0f79a3..4dcbf2d39 100644
--- a/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
@@ -19,17 +19,13 @@
 import lombok.Value;
 import lombok.With;
 import org.openrewrite.Contributor;
-import org.openrewrite.Incubating;
 import org.openrewrite.Maintainer;
 import org.openrewrite.internal.lang.Nullable;
 
 import java.net.URI;
 import java.time.Duration;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
-import java.util.function.Consumer;
-import java.util.stream.Collectors;
 
 @Value
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
@@ -62,23 +58,4 @@ public class RecipeDescriptor {
     List<RecipeExample> examples;
 
     URI source;
-
-    @Incubating(since = "8.11.3")
-    public void print(Consumer<String> consumer) {
-        StringBuilder recipeString = new StringBuilder(name);
-
-        if (options != null && !options.isEmpty()) {
-            String opts = options.stream().map(option -> {
-                if (option.getValue() != null) {
-                    return String.format("%s=%s", option.getName(), option.getValue());
-                }
-                return null;
-            }).filter(Objects::nonNull).collect(Collectors.joining(", "));
-            if (!opts.isEmpty()) {
-                recipeString.append(String.format(": {%s}", opts));
-            }
-        }
-
-        consumer.accept(recipeString.toString());
-    }
 }
diff --git a/rewrite-test/src/main/java/org/openrewrite/test/RecipePrinter.java b/rewrite-test/src/main/java/org/openrewrite/test/RecipePrinter.java
index c4df19720..9a621d26b 100644
--- a/rewrite-test/src/main/java/org/openrewrite/test/RecipePrinter.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RecipePrinter.java
@@ -15,57 +15,64 @@
  */
 package org.openrewrite.test;
 
-import lombok.Value;
 import org.openrewrite.Incubating;
 import org.openrewrite.Recipe;
 import org.openrewrite.config.OptionDescriptor;
 import org.openrewrite.config.RecipeDescriptor;
 
-import java.util.*;
+import java.util.List;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 @Incubating(since = "8.12.1")
-public interface RecipeTreePrinter {
+public interface RecipePrinter {
 
-    RecipeTreePrinter DEFAULT = () -> System.out::println;
-
-    Consumer<String> consumeRecipeDescriptor();
+    /**
+     * The consumer to print the recipe tree to, for instance {@code () -> System.out::println}.
+     *
+     * @return the consumer
+     */
+    Consumer<CharSequence> consumer();
 
     /**
-     * Print the recipe tree for a recipe.
+     * Print the recipe tree for a recipe to {@link #consumer()}.
      *
-     * @param recipe the
+     * @param recipe the recipe to print
      */
     default void printTree(Recipe recipe) {
-        printRecipe(recipe.getDescriptor(), "");
+        consumer().accept(printRecipe(recipe.getDescriptor(), ""));
     }
 
     /**
-     * Print the recipe tree for a recipe descriptor.
+     * Internal method to print a recipe tree. This is used recursively to print the tree. Not intended for external use.
      *
      * @param rd     the recipe descriptor
      * @param prefix the indentation prefix
      */
-    default void printRecipe(RecipeDescriptor rd, String prefix) {
+    default CharSequence printRecipe(RecipeDescriptor rd, String prefix) {
         List<OptionDescriptor> options = rd.getOptions();
-        StringBuilder recipeString = new StringBuilder(rd.getName());
+        StringBuilder recipeString = new StringBuilder(prefix);
+        recipeString.append(rd.getName());
         if (options != null && !options.isEmpty()) {
-            String opts = options.stream().map(option -> {
-                if (option.getValue() != null) {
-                    return String.format("%s=%s", option.getName(), option.getValue());
-                }
-                return null;
-            }).filter(Objects::nonNull).collect(Collectors.joining(", "));
+            String opts = options.stream()
+                    .map(option -> {
+                        if (option.getValue() != null) {
+                            return String.format("%s=%s", option.getName(), option.getValue());
+                        }
+                        return null;
+                    }).
+                    filter(Objects::nonNull)
+                    .collect(Collectors.joining(", "));
             if (!opts.isEmpty()) {
                 recipeString.append(String.format(": {%s}", opts));
             }
         }
-
-        consumeRecipeDescriptor().accept(recipeString.toString());
+        recipeString.append(System.lineSeparator());
 
         for (RecipeDescriptor child : rd.getRecipeList()) {
-            printRecipe(child, prefix + "  ");
+            recipeString.append(printRecipe(child, prefix + "  "));
         }
+        return recipeString;
     }
 }
diff --git a/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java b/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java
index 73e09fdf7..6b67efe59 100644
--- a/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java
@@ -99,7 +99,7 @@ public static RecipeSpec defaults() {
     Function<List<SourceFile>, LargeSourceSet> sourceSet;
 
     @Nullable
-    RecipeTreePrinter recipeTreePrinter;
+    RecipePrinter recipePrinter;
 
     /**
      * Configuration that applies to all source file inputs.
@@ -287,8 +287,8 @@ ExecutionContext getExecutionContext() {
     }
 
     @Incubating(since = "8.11.3")
-    public RecipeSpec printRecipe(RecipeTreePrinter recipeTreePrinter) {
-        this.recipeTreePrinter = recipeTreePrinter;
+    public RecipeSpec printRecipe(RecipePrinter recipePrinter) {
+        this.recipePrinter = recipePrinter;
         return this;
     }
 }
diff --git a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
index af6daeca1..670fbbe7f 100644
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -182,15 +182,10 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
             validateRecipeOptions(recipe);
         }
 
-        RecipeTreePrinter recipeTreePrinter = null;
-        if (testMethodSpec.getRecipeTreePrinter() != null) {
-            recipeTreePrinter = testMethodSpec.getRecipeTreePrinter();
-        } else if (testClassSpec.getRecipeTreePrinter() != null) {
-            recipeTreePrinter = testClassSpec.getRecipeTreePrinter();
-        }
-
-        if (recipeTreePrinter != null) {
-            recipeTreePrinter.printTree(recipe);
+        if (testMethodSpec.getRecipePrinter() != null) {
+            testMethodSpec.getRecipePrinter().printTree(recipe);
+        } else if (testClassSpec.getRecipePrinter() != null) {
+            testClassSpec.getRecipePrinter().printTree(recipe);
         }
 
         int cycles = testMethodSpec.cycles == null ? testClassSpec.getCycles() : testMethodSpec.getCycles();
diff --git a/rewrite-test/src/test/java/org/openrewrite/test/RecipePrinterTest.java b/rewrite-test/src/test/java/org/openrewrite/test/RecipePrinterTest.java
index 183625e5b..b75da49fb 100644
--- a/rewrite-test/src/test/java/org/openrewrite/test/RecipePrinterTest.java
+++ b/rewrite-test/src/test/java/org/openrewrite/test/RecipePrinterTest.java
@@ -29,7 +29,7 @@
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
-class RecipeTreePrinterTest implements RewriteTest {
+class RecipePrinterTest implements RewriteTest {
 
     private StringBuilder sb;
 
@@ -75,24 +75,24 @@ void printRecipeTreeForRecipeWithNestedRecipes() {
 
         assertThat(output).isEqualTo(expected);
     }
-}
 
-@Value
-@EqualsAndHashCode(callSuper = true)
-class TestRecipe extends Recipe {
+    @Value
+    @EqualsAndHashCode(callSuper = true)
+    private static class TestRecipe extends Recipe {
 
-    @Option(displayName = "An option",
-      description = "A sample option.",
-      example = "Some text.")
-    String theOption;
+        @Option(displayName = "An option",
+          description = "A sample option.",
+          example = "Some text.")
+        String theOption;
 
-    @Override
-    public String getDisplayName() {
-        return "Test recipe";
-    }
+        @Override
+        public String getDisplayName() {
+            return "Test recipe";
+        }
 
-    @Override
-    public String getDescription() {
-        return "Test recipe.";
+        @Override
+        public String getDescription() {
+            return "Test recipe.";
+        }
     }
 }
```

My intention here is to
- limit the public API going forward,
- contain the changes to the `rewrite-test` module & `RewriteTest` as much as possible
- revert the changes to `org.openrewrite.config.RecipeDescriptor`
- have only two public methods on `RecipePrinter`
- in naming don't assume we'll print a tree
- make `printRecipe(RecipeDescriptor rd, String prefix)` internal
- switch to `Consumer<CharSequence>` to allow direct `StringBuilder` use
- remove unused `RecipeTreePrinter DEFAULT`; `() -> System.out::println`
- remove `SystemOutRecipeTreePrinter` and `recipesToSkipChildren`

If you're ok with these changes, at least to start, then I can apply them to your branch and see this through.

Note that I'm thankful for the work you've put in! Merely trying to limit the API surface to support.

ammachado:
LGTM

rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java
```diff
@@ -282,4 +285,10 @@ public RecipeSpec afterTypeValidationOptions(TypeValidation typeValidation) {
     ExecutionContext getExecutionContext() {
         return executionContext;
     }
+
+    @Incubating(since = "8.11.3")
```
timtebeek, line 289:
```suggestion
    @Incubating(since = "8.12.1")
```

timtebeek (approved):
Thanks a lot! Nice addition to help troubleshoot large recipe hierarchies.

PR #3836, opened 2023-12-19
Feat:token length estimator

justine-gehring:
Adding two recipes which estimates the number of token for class definitions and method definitions, and inputs those estimates into datatables. 

## Any additional context

This is to estimate latency for the ai agent recommender.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ClassDefinitionLength.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ClassDefinitionLength.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.search.DeclaresMethod;
+import org.openrewrite.java.table.TokenCount;
+import org.openrewrite.java.tree.J;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class ClassDefinitionLength extends Recipe {
+    @Override
+    public String getDisplayName() {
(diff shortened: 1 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ClassDefinitionLength.java
```diff
@@ -0,0 +1,57 @@
...
+        return "Locates class definitions and predicts the number of token in each.";
+    }
+
+    transient TokenCount tokens = new TokenCount(this);
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration clazz, ExecutionContext ctx) {
+                J.ClassDeclaration cd = getCursor().firstEnclosing(J.ClassDeclaration.class);
```
knutwannheden, line 47: @justine-gehring Please note that this will always return the class declaration `clazz` itself. To me it isn't quite clear what the purpose is of this, but I assume it was copied-pasted from the other recipe and can thus be removed. Still, you might want to consider if there should also be an entry for nested classes or not (currently won't be the case).

rewrite-java/src/main/java/org/openrewrite/java/MethodDefinitionLength.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+
+        return new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                J.MethodDeclaration md = super.visitMethodDeclaration(method, ctx);
+                J.ClassDeclaration cd = getCursor().firstEnclosing(J.ClassDeclaration.class);
+                if(cd == null) {
+                    return md;
```
knutwannheden, line 50: I suspect that this should never be possible, so we could probably also just remove this check.

rewrite-java/src/main/java/org/openrewrite/java/MethodDefinitionLength.java
```diff
@@ -0,0 +1,58 @@
...
+                }
+                int numberOfTokens = (int) ( md.printTrimmed(getCursor()).length() / 3.5);
+                tokens.insertRow(ctx, new TokenCount.Row( md.getSimpleName(), numberOfTokens));
```
knutwannheden, line 53: Note that Java supports method overloading, which means that you would end up with multiple rows with the same value. I think it might be better to instead use a signature. To do this you can have a field like this in your visitor:
```java
final JavaTypeSignatureBuilder signatureBuilder = new DefaultJavaTypeSignatureBuilder();
```
and then use it something like this:
```java
String signature = signatureBuilder.signature(md.getType());
```

knutwannheden (commented):
@justine-gehring Added a few comments to your PR.

rewrite-java/src/main/java/org/openrewrite/java/ClassDefinitionLength.java
justine-gehring: I believe no entry for nested classes is fine for what we're estimating. But thank you for all your comments, it clears up some things.

PR #3837, opened 2023-12-19
New `AnnotationService`

knutwannheden:
Instead of calling the `getAllAnnotations()` and `getAnnotations()` methods directly, recipes and visitors should be using the `AnnotationService`. This will allow `rewrite-kotlin` to properly support multi annotations and annotations with use-site targets.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
@@ -23,9 +23,7 @@ import org.openrewrite.java.service.ImportService;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 
-import java.util.Iterator;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 
 @SuppressWarnings("unused")
 public class JavaVisitor<P> extends TreeVisitor<J, P> {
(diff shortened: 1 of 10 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -137,7 +137,12 @@ public void maybeAddImport(@Nullable String packageName, String typeName, @Nulla
 
     @Incubating(since = "8.2.0")
     public <S> S service(Class<S> service) {
-        return getCursor().firstEnclosingOrThrow(JavaSourceFile.class).service(service);
+        for (Cursor c = getCursor(); c.getParent() != null; c = c.getParent()) {
+            if (c.getValue() instanceof JavaSourceFile) {
+                return ((JavaSourceFile) c.getValue()).service(service);
+            }
+        }
```
knutwannheden, line 144: Since `service()` needs to traverse to the top of the tree, I wonder if it should leave "breadcrumbs" if the depth exceeds a certain level. E.g. if the depth exceeds 100, it could leave a cached entry at half way up the tree (as a message in the corresponding cursor). The next access would then only have to traverse half the way up (although it would have to check the messages of each cursor). The checking for messages could also be delayed until the depth of 50 has been exceeded.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -134,6 +134,7 @@ public AnnotatedType withType(@Nullable JavaType type) {
             return withTypeExpression(typeExpression.withType(type));
         }
 
+        @Deprecated
```
timtebeek, line 137: Might be kind to add a note here to each of these deprecated classes why they where deprecated, and what to use now. Folks are likely to stumble across these deprecations in their existing recipes, and would otherwise be left wondering what to use instead.

timtebeek (approved):

traceyyoshima (approved):
LGTM - really, like the addition of the `matches` method.

PR #3842, opened 2023-12-20
Fix the plugin default groupId

gsmet:
The default groupId for Maven plugins is not the one from the current pom but `org.apache.maven.plugins`.

## What's changed?

The default groupId for Maven plugins was the one from the currently resolved POM, while it should be `org.apache.maven.plugins`.

## What's your motivation?

Correctness :).
See https://stackoverflow.com/questions/65527291/is-groupid-required-for-plugins-in-maven-pom-xml for the details.

@timtebeek hey, found this while working on the update recipes for Quarkus 3.7. I can work around it for now but I think it's worth a fix upstream. Thanks!

### Checklist
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files -> I use Eclipse but I was careful about not changing the formatting.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -198,7 +198,7 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
 
     private boolean hasPluginGroupId(String groupId) {
         Xml.Tag tag = getCursor().getValue();
-        boolean isGroupIdFound = matchesGlob(tag.getChildValue("groupId").orElse(getResolutionResult().getPom().getGroupId()), groupId);
+        boolean isGroupIdFound = matchesGlob(tag.getChildValue("groupId").orElse("org.apache.maven.plugins"), groupId);
         if (!isGroupIdFound && getResolutionResult().getPom().getProperties() != null) {
             if (tag.getChildValue("groupId").isPresent() && tag.getChildValue("groupId").get().trim().startsWith("${")) {
                 String propertyKey = tag.getChildValue("groupId").get().trim();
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot! I'm out tonight but will try to get a patch release out tomorrow-ish. Thanks for the help!

gsmet:
No pressure from our side. We have a workaround and 3.7 is planned for the end of January.

gsmet:
Thanks for merging.

timtebeek:
Through the magic of time zones and a distributed team I woke up to find there's already a new release out containing your fix: https://github.com/openrewrite/rewrite/releases/tag/v8.11.4
Good to have you back on the latest versions and contributing here!

Let me know if there's anything I can do to help smoothen towards your 3.7 release. I imagine you should be able to leverage [Migrate to Java 17](https://docs.openrewrite.org/recipes/java/migrate/upgradetojava17) and the recently added
- https://github.com/openrewrite/rewrite-quarkus/pull/68

gsmet:
Ah that's interesting.

I missed the Java 17 recipe and implemented the update of the Maven Compiler properties yesterday. I suppose I will drop my work and use yours.
I think I will cherry-pick things a bit though as I don't want to enforce a text block migration on our users. It could generate a lot of noise when reviewing the diff.

As for the Quarkus recipe, very nice, I don't see it there though: https://docs.openrewrite.org/recipes/quarkus, has it been released? I will add a comment there because the OP might be interested in expanding it further.

timtebeek:
Yes seems reasonable to cherry pick; there's a lot we've tied together but I could see you wanting to create a minimal diff at first. There's some other helpful bits in there too such as [upgrading your GitHub Actions to use Java 17](https://github.com/openrewrite/rewrite-migrate-java/blob/217fc8eadc43393e59c9127e34cf38af89a9823b/src/main/resources/META-INF/rewrite/java-version-17.yml#L32-L33), such that any first PR will work. We're continuously expanding those recipes to cover more cases, which you might miss out on when cherry picking, but I do understand that you mostly want to get folks to Quarkus 3.7 without any chance of pushback. Feel free to direct them to us for further migration steps then. My recent favorite is [moving to getFirst and getLast](https://github.com/openrewrite/rewrite-migrate-java/blob/217fc8eadc43393e59c9127e34cf38af89a9823b/src/main/resources/META-INF/rewrite/java-version-21.yml#L56-L59) on Java 21. :)

I hadn't yet done a new release of rewrite-quarkus since @ammachado added the properties recipe there; I'll do one soon-ish (it requires a few others to go first), and definitely well before your January deadline.

PR #3843, opened 2023-12-21
Instance names

jkschneider:
Instance names for recipes help to display human readable text for instantiated recipes participating in recipe runs, including their options where applicable.


zieka (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePluginGroupIdAndArtifactId.java
```diff
@@ -35,41 +34,38 @@ public class ChangePluginGroupIdAndArtifactId extends Recipe {
...
 
-    @Option(displayName = "New artifactId",
-            description = "The new artifactId to use. Defaults to the existing artifact id.",
+    @Option(displayName = "New artifact ID",
+            description = "The new artifact ID to use. Defaults to the existing artifact ID.",
             example = "my-new-maven-plugin",
             required = false)
     @Nullable
-    String newArtifactId;
+    String newArtifact;
```
k-seth, line 59: Out of curiosity, was this change intended? Any existing declarative recipes that use this recipe will no longer provide the correct key, not to mention the naming is now inconsistent with other recipes like `ChangeDependencyGroupIdAndArtifactId`.

nmck257: encountered this organically upon upgrade, and I agree, it does feel like an accident

timtebeek: Here's a PR that reverts the change in a backwards compatible manner
- https://github.com/openrewrite/rewrite/pull/4421

PR #3844, opened 2023-12-21
Replace resilience4j with failsafe

pstreef:
## What's changed?
`resilience4j-retry` replaced with `dev.failsafe`

## What's your motivation?
`resilience4j-retry` has a lot of transitive dependencies and the latest version depends on java 17. Having to pin this to 1.7.0 can cause issues with consumers.

[FailSafe](https://github.com/failsafe-lib/failsafe) is a "lightweight, zero-dependency library for handling failures in Java 8+" which can handle the retries in stead

I need to look into metrics. If the old retries had them and if we want them on the new.

Implement an internal retrier.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -33,10 +33,7 @@ import org.openrewrite.maven.MavenSettings;
 import org.openrewrite.maven.cache.MavenPomCache;
 import org.openrewrite.maven.tree.*;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
+import java.io.*;
 import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.nio.file.DirectoryStream;
@@ -59,6 +56,7 @@ import static java.util.stream.Collectors.toList;
 public class MavenPomDownloader {
     private static final RetryPolicy<Object> retryPolicy = RetryPolicy.builder()
             .handle(SocketTimeoutException.class, TimeoutException.class)
+            .handleIf(throwable -> throwable instanceof UncheckedIOException && throwable.getCause() instanceof SocketTimeoutException)
             .withDelay(Duration.ofMillis(500))
             .withJitter(0.1)
             .withMaxRetries(5)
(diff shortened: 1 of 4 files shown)


timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -140,26 +131,33 @@ public MavenPomDownloader(Map<Path, Pom> projectPoms, HttpSender httpSender, Exe
...
+                    }
+                    return response.getBodyAsBytes();
+                }
+            });
+        } catch (Throwable t) {
+            if (t.getCause() != null) {
+                if (t.getCause() instanceof HttpSenderResponseException) {
+                    throw (HttpSenderResponseException) t.getCause();
+                }
+                throw new HttpSenderResponseException(t.getCause(), null);
```
timtebeek, line 153: I worry a bit about this wrapping any `Throwable.cause` in a `HttpServerResponseException`, since we use the response exception to decide that a server existed on these lines for instance: https://github.com/openrewrite/rewrite/blob/e17faa5222a99b29bbae6c9c20e813faa2d99051/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java#L753-L754

I think we'd best reserve `HttpServerResponseException` for cases where there is a http response code to pass in, as there's a lot of nested try/catches below that are already hard to reason about, which would be even more difficult if there's `HttpServerResponseException` without http responses.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -140,26 +131,33 @@ public MavenPomDownloader(Map<Path, Pom> projectPoms, HttpSender httpSender, Exe
...
+        try {
+            return Failsafe.with(retryPolicy).get(() -> {
+                try (HttpSender.Response response = httpSender.send(request)) {
+                    if (!response.isSuccessful()) {
+                        throw new HttpSenderResponseException(null, response.getCode());
+                    }
+                    return response.getBodyAsBytes();
+                }
+            });
+        } catch (Throwable t) {
```
timtebeek, line 148: I wonder if it makes sense to catch `FailsafeException` separately, and if caught throw the underlying cause instead of wrapping that in a `HttpSenderResponseException`. I'll see if I can whip up an alternative.

pstreef: great minds and all. See latest commit

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenArtifactDownloader.java
```diff
@@ -108,7 +108,7 @@ public Path downloadArtifact(ResolvedDependency dependency) {
                 bodyStream = Files.newInputStream(Paths.get(URI.create(uri)));
             } else {
                 HttpSender.Request.Builder request = applyAuthentication(dependency.getRepository(), httpSender.get(uri));
-                try (HttpSender.Response response = sendRequest.apply(request.build());
+                try (HttpSender.Response response = Failsafe.with(retryPolicy).get(() -> httpSender.send(request.build()));
```
timtebeek, line 111: Would it make sense to additionally catch `FailsafeException` below and throw it's cause when encountered? Just to make sure that when `ExceptionUtils.sanitizeStackTrace` truncates the stacktrace, that we're left with the informative bits, and not the retry.

timtebeek (approved):
Already approved, but with a final suggestion left in my last comment.

Nice to see how you implemented exactly what I had in mind after my first comment, and good to have a long standing retry bug fixed.

PR #3850, opened 2023-12-25
Correctly handle `Recipe#maxCycles()` in `RecipeStack`

knutwannheden:
When a recipe's `maxCycles()` declares that a recipe should not be executed, then none of its sub recipes (returned by `getRecipeList()`) should be executed either. This is now corrected in the recipe scheduler's `RecipeStack` component.

### DIFF
--- a/rewrite-core/openrewrite.json
+++ b/rewrite-core/openrewrite.json
@@ -49,10 +49,6 @@
             "type": "string",
             "format": "duration"
           },
-          "maxCycles": {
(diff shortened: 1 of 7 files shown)


knutwannheden:
@timtebeek This PR also fixes a bug that could be committed separately if we are unsure about the `maxCycles` YAML property (although as pointed out in the issue I think there are valid use cases for it).

timtebeek (approved):

PR #3851, opened 2023-12-27
Annotation added on method parameter lacks whitespace separation with next annotation

jeonyeonbin:
## What's changed?

it is only provide testing.

## What's your motivation?

i think autoFormat does between Annotation space but it doesnt` work

@timtebeek 

### Checklist


rewrite-java/src/test/java/org/openrewrite/java/AddAnnotationInMethodParameterTest.java
```diff
@@ -0,0 +1,109 @@
...
+                  public void method(@Deprecated String method) {
+                  }
+              }
+              """,
+            """
+              package org.openrewrite.java;
+                          
+              class Example {
+                  
+                  public void method(@DeprecatedMethodParameter @Deprecated String method) {
```
jeonyeonbin, line 101: in this case i want to like this 

but result is
```
public void method(@DeprecatedMethodParameter@Deprecated String method) {
```

have you ever provide solutions?

seonwoo960000:
Hi, @koppor, shouldn't the consecutive annotations be separated? Does rewrite provides this kind of features 🤔 ?

koppor:
@seonwoo960000 I am "only" a user of OpenRewrite and don't know about all features. Since you are in the code, maybe, you can check the code?

PR #3860, opened 2023-12-29
Add annotations to `JavaType$Array`

traceyyoshima:
Changes:
- Added `FullyQualified[]` annotations field to `JavaType$Array`
- Updated Java type mappers and signature builders to support annotations on JavaType$Array.
- Updated Java parsers and printers to parser and print annotated dimensions in source order.

TODO:
- Update `RemoveAnnotationVisitor` the visitor removes annotations from the LST, but does not update the types.


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1339,42 +1339,68 @@ private TypeTree annotatedTypeTree(Tree node, Map<Integer, JCAnnotation> annotat
         } else if (node instanceof JCAnnotatedType) {
```
knutwannheden, line 1339: I was wondering if we need this whole if block or if it is covered by the `convert()`. If it is required, then I wonder if a test case is missing, as a breakpoint on the `convert()` statement never hit when running the rewrite-java-tck tests. Even when running all the rewrite-java-test tests the breakpoint was also just hit twice, but removing this block doesn't seem to cause any problems.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1290,7 +1290,7 @@ public J visitAnnotatedType(AnnotatedTypeTree node, Space fmt) {
             if (node.getUnderlyingType() instanceof JCFieldAccess) {
                 return new J.AnnotatedType(randomId(), fmt, Markers.EMPTY, leadingAnnotations, annotatedTypeTree(node.getUnderlyingType(), annotationPosTable));
             } else if (node.getUnderlyingType() instanceof JCArrayTypeTree) {
```
knutwannheden, line 1292: Alternatively to my comment on line 1339 I suspect that this block here could be removed (didn't test this however).

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
traceyyoshima: I tested this and this block isn't necessary anymore -- I'll remove it in a moment

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
traceyyoshima: This is necessary, otherwise code will be parsed into whitespace.

In some of our parsers, we confirm that whitespace does not contain non-whitespace characters. I don't know if we ever added that to Java or if it was removed, but we don't check at this time.

I've found a handful of cases where source code was parsed into whitespace when it shouldn't have. So, it may be a good idea to add that to `Assertions#java`.

knutwannheden (approved):

Issue #3856, opened 2023-12-28
MergeYaml recipe fileMatcher property is present in 7.x version but removed in 8.x versions


vijay-chavakula:
MergeYaml recipe **_fileMatcher_** is present in 7.x version but removed in 8.x version
causing the recipe changes to apply to all yml files in project and not only to specific files.

I am using

- OpenRewrite recipe bom V2.3.1
- Maven compiler plugin v3.8.0
- rewrite-yaml v8.7.2 and 8.11.1

## How are you running OpenRewrite?

Are you using the Maven plugin, Gradle plugin, Moderne CLI, Moderne.io or something else?
Maven plugin 
Is your project a single module or a multi-module project?
Single Module
Can you share your configuration so that we can rule out any configuration issues?

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>3.8.0</version>
</plugin>
<dependency>
      <groupId>org.openrewrite</groupId>
      <artifactId>rewrite-yaml</artifactId>
      <scope>compile</scope>
</dependency>
```
```yaml
org.openrewrite.yaml.MergeYaml:
      key: $
      yaml: >
        tasks:
          preBuild:
            - closure: startServer
         
      acceptTheirs: false
      fileMatcher: '**/jules.yml'
      objectIdentifyingProperty: name
```

## What is the smallest, simplest way to reproduce the problem?
using the declarative approach of the above mentioned recipe in creating our own recipe and compile and run the jar on our projects to will apply the yaml changes to all the yml files in the project structure not only to jules.yml. 

## What did you expect to see?
We are expecting to revert the changes done to **_MergeYaml_** recipe to include back **_fileMatcher_** property and apply changes only to those files.
Like in above example only to jules.yml in all directories of project.

## What did you see instead?
Right now after 8.x versions as the fileMatcher property is removed we see the changes getting applied to all the yml files in the project which is not expected. 

## What is expected to be done to fix it?
Earlier the code for the 7.11.0 and above version is 
```java
 @Override
    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
        return new HasSourcePath<>(fileMatcher);
    }
```
We need similar code to check which files to have the changes applied.

[...]

PR #3866, opened 2023-12-29
#3856 fileMatcher added back to MergeYaml

vijay-chavakula:
## What's changed?
fileMatcher option which is removed has been added back

## What's your motivation?
as the preconditions are not working in declarative approach for the issue #3856 
we have to scan the files and apply recipe only on the files which matches the regular expression.

Added testcases which shows files which matches regular expression are only touched with this approach, 
there is one breaking change, users who are using MergeYaml in Java custom recipe now should pass one more extra parameter in the constructor.

@timtebeek 

There are no alternatives for this issue. 

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -37,6 +37,14 @@ public class MergeYaml extends Recipe {
     @Language("yml")
     String yaml;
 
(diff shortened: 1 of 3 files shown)


timtebeek:
As also mention in Slack:  intuitively I'd say there's an issue with how preconditions apply to scanning recipes that we need to fix instead, which would remove the need for `fileMatcher` on `MergeYaml`; let's explore that first.

vijay-chavakula:
this is not an intended fix

PR #3868, opened 2023-12-29
Preserve space and comments after import in RemoveImport

tobli:
If the removed import was a static import and was followed by a separating blank line, that blank line would be collapsed. Also if the removed import was followed by a comment, that comment would be lost.

This commit addresses these two issues by:
- not overwriting the prefix spacing if it'd remove whitespace
- only copy whitespace from removed import, don't overwrite comments

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveImportTest.java
@@ -331,6 +331,162 @@ class RemoveImportTest implements RewriteTest {
         );
     }
 
+    @Test
+    void preservesWhitespaceAfterRemovedImportForWindowsWhitespace() {
+        rewriteRun(
+          spec -> spec.recipe(removeImport("java.util.List")),
+          java(
+              "package com.example.foo;\r\n" +
+              "\r\n" +
+              "import java.util.Collection;\r\n" +
+              "import java.util.List;\r\n" +
+              "\r\n" +
+              "import java.util.ArrayList;\r\n" +
+              "\r\n" +
+              "public class A {\r\n" +
+              "}\r\n",
+              "package com.example.foo;\r\n" +
+              "\r\n" +
+              "import java.util.Collection;\r\n" +
+              "\r\n" +
+              "import java.util.ArrayList;\r\n" +
+              "\r\n" +
+              "public class A {\r\n" +
+              "}\r\n"
+          )
+        );
+    }
+
+    @Test
+    void preservesWhitespaceAfterRemovedImportForMixedWhitespaceBefore() {
+        rewriteRun(
+          spec -> spec.recipe(removeImport("java.util.List")),
+          java(
+              "package com.example.foo;\n" +
+              "\n" +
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/RemoveImportTest.java
```diff
@@ -331,6 +331,86 @@ public class A {
         );
     }
 
+    @Test
+    void doNotLeaveLeaveInitialBlankLine() {
```
tobli, line 335: Added this test case to assert a case (removing the first import, with no package statement) that was present in other test classes that uses RemoveImport.

timtebeek:
Quality work here @tobli ! Thanks for adding these fixes and the detailed description of your intention here and in the test cases added. I'm tagging @traceyyoshima just in case I'd missed something; he seems to have added the conditional on line 146 that's removed.
https://github.com/openrewrite/rewrite/blob/3da2178f7ca70d421175e9cc6471e478ff63081e/rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java#L145-L149

Otherwise ok to merge on my end; appreciate you matching our code style too in your changes!

tobli:
Thanks!
I replaced the conditional, and the counting of line breaks, with the length check when applying the leftover white space. It's not exactly identical, the new code handled some scenarios better. That said, there are some edge cases where some more complex conditionals (combining the two approaches) would be even more correct, but I found this to be a pragmatic trade off. Very much open to tweaking that however, so appreciate the review.

traceyyoshima:
Hi @tobli, thank you for creating this PR! I think Jonathan added the condition, regardless, the case that the original code covers is mixed CRLF and LF, and trailing whitespace.

I.E. if the removed prefix is `\n\n` and the current prefix is `\r\n` or `\n  \n`, the condition will fail. I think mixed whitespace is common enough that we should cover those cases.

tobli:
@traceyyoshima totally valid points, I adapted and added some test cases that should cover those cases.

tobli:
@traceyyoshima @timtebeek anything more you'd like to see in this PR for it to be ready?

rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java
```diff
@@ -172,6 +175,10 @@ public RemoveImport(String type, boolean force) {
         return j;
     }
 
+    private long countTrailingLinebreaks(Space space) {
+        return space.getLastWhitespace().chars().filter(s -> s == '\n').count();
```
knutwannheden, line 179: I am a bit unsure why this doesn't fail the build as `String#chars()` was added in Java 9 and we should have the build configured to require Java 8.

Any ideas, @timtebeek ?

knutwannheden: Hmm... Apparently `CharSequence#chars()` was already added as a default method in Java 8. What was added in Java 9 was just the `String#chars()` override for better performance.

knutwannheden (approved):

PR #3871, opened 2023-12-31
Add break statement to toCamelCase

seonwoo960000:
## What's changed?
Add a break statement in for loop. 

## What's your motivation?
We don't have to check all characters in a string to see whether the string is `allUpperCase`. We can break the for loop when any character is lowercase. 

N/A 

N/A 

N/A 

## Any additional context
N/A 

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases -> existing tests will verify whether change is valid 

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
@@ -148,18 +148,19 @@ public enum NameCaseConvention {
      */
     private static String toCamelCase(String str, boolean lowerCaseFirstLetter) {
         boolean allUpperCase = true;
-        for (char c : str.toCharArray()) {
+        final int strLength = str.length();
+        for (int i = 0; i < strLength; i++) {
+            final char c = str.charAt(i);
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
```diff
@@ -149,11 +149,13 @@ private static String lowerUnderscore(String str) {
     private static String toCamelCase(String str, boolean lowerCaseFirstLetter) {
         boolean allUpperCase = true;
         for (char c : str.toCharArray()) {
```
knutwannheden, line 151: It would be even better if we would here use a regular `for` loop and access the characters using `String#charAt(int)` rather then converting to `char[]` first.

seonwoo960000: Sure, I'll fix that as well

timtebeek (approved):
Thanks for noticing and providing this improvement!

knutwannheden (commented):
👍

Issue #3744, opened 2023-11-16
The `v` from `var` is parsed into prefix when assigned an unknown type, leading to recipe issues further down


kevin0x90:
## What version of OpenRewrite are you using?
I use the current RELEASE version.

I am using

- OpenRewrite
- Maven plugin
- org.openrewrite.staticanalysis.FinalizeLocalVariables

## How are you running OpenRewrite?
I ran the following maven command:
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-static-analysis:RELEASE \
  -Drewrite.activeRecipes=org.openrewrite.staticanalysis.FinalizeLocalVariables
```

## What is the smallest, simplest way to reproduce the problem?
The issue is reproducible with the following test case
```java
@Test
    void shouldFinalizeCorrectly() {
        rewriteRun(
          s -> s.typeValidationOptions(TypeValidation.none()),
          version(
            //language=java
            java(
              """
                public class Main {
                    private boolean test(MessagePayload payload) {
                        var a = DomainObject.from(payload.getValue());
                
                        return a != null
                            && Integer.valueOf(999).equals(a.getField1())
                            && Integer.valueOf(101).equals(a.getField2());
                    }
                }
                  """,
              """
                public class Main {
                    private boolean test(MessagePayload payload) {
                        final var a = DomainObject.from(payload.getValue());
                    
                        return a != null
                            && Integer.valueOf(999).equals(a.getField1())
                            && Integer.valueOf(101).equals(a.getField2());
                    } 
                }
                  """
            ), 17)
        );
    }
```

The above test fails with the following diff:
![Screenshot from 2023-11-16 12-48-54](https://github.com/openrewrite/rewrite-static-analysis/assets/6693174/6098601f-9f10-4eb0-993e-56b3e47b4946)

seonwoo960000:
I'm currently looking into the issue, but stuck where the "var" is replaced into "(ERROR)" 
- Method: `ReloadableJava17ParserVisitor.visitVariables`
![image](https://github.com/openrewrite/rewrite/assets/69591622/6e57a225-70a6-4fa0-9e3d-3d80bf2a06fc)

PR #3872, opened 2024-01-01
Check "var" keyword when varType is null or JCErroneous

seonwoo960000:
## What's changed?
When `var` is used with unknown type, rewrite consider `var` as `Space`. We need to add additional logic to identifyt `var` keyword. 

## What's your motivation?
[Issue link](https://github.com/openrewrite/rewrite/issues/3744)

Below test code doesn't work. It rewrites `var a = b` into `finalvar a = b` (should have space between final and var) 
```java
@Test
@Issue("https://github.com/openrewrite/rewrite/issues/3744")
void missingTypePlusVarNotMissingSpace() {
    rewriteRun(
      s -> s.typeValidationOptions(TypeValidation.none()),
      //language=java
      java(
        """
          import foo.Unknown;
          class Bar {
              void test(Unknown b) {
                  var a = b;
              }
          }
            """,
        """
          import foo.Unknown;
          class Bar {
              void test(Unknown b) {
                  final var a = b;
              } 
          }
            """
      )
    );
}
```

- Whether it's ok to put the checking logic into
  - `ReloadableJava11ParserVisitor`
  - `ReloadableJava17ParserVisitor`
  - `ReloadableJava21ParserVisitor`

N/A 

N/A 

## Any additional context
AS-IS 
image

TO-BE 
image

### Checklist

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -1508,14 +1508,16 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
         List<J.Annotation> typeExprAnnotations = collectAnnotations(annotationPosTable);
 
         TypeTree typeExpr = null;
-        if (vartype == null || vartype instanceof JCErroneous) {
+        if (vartype == null) {
+            typeExpr = null;
+        } else if (vartype instanceof JCErroneous) {
             try {
-                if (source.substring(node.startPos, node.pos).contains("var")) {
+                if (source.substring(node.startPos, node.pos).startsWith("var")) {
                     // "var x = unknownType" can be translated into "(ERROR) var x = unknownType"
                     typeExpr = new J.Identifier(randomId(), sourceBefore("var"), Markers.EMPTY, emptyList(), "var", typeMapping.type(vartype), null);
                     typeExpr = typeExpr.withMarkers(typeExpr.getMarkers().add(JavaVarKeyword.build()));
                 }
-            } catch (StringIndexOutOfBoundsException e) {
+            } catch (Exception e) {
                 typeExpr = null;
             }
         } else if (endPos(vartype) < 0) {
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -1565,7 +1565,6 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
         for (int i = 0; i < nodes.size(); i++) {
             JCVariableDecl n = (JCVariableDecl) nodes.get(i);
 
-            // how to capture var ?
             Space namedVarPrefix = sourceBefore(n.getName().toString());
 
             JavaType type = typeMapping.type(n);
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -1507,19 +1507,9 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
 
         List<J.Annotation> typeExprAnnotations = collectAnnotations(annotationPosTable);
 
-        TypeTree typeExpr = null;
+        TypeTree typeExpr;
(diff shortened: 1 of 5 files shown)


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1507,9 +1507,19 @@ private J.VariableDeclarations visitVariables(List<VariableTree> nodes, Space fm
...
+        } else if (vartype instanceof JCErroneous) {
+            try {
+                if (source.substring(node.startPos, node.pos).startsWith("var")) {
+                    // "var x = unknownType" can be translated into "(ERROR) var x = unknownType"
+                    typeExpr = new J.Identifier(randomId(), sourceBefore("var"), Markers.EMPTY, emptyList(), "var", typeMapping.type(vartype), null);
+                    typeExpr = typeExpr.withMarkers(typeExpr.getMarkers().add(JavaVarKeyword.build()));
+                }
+            } catch (Exception e) {
+                typeExpr = null;
+            }
```
timtebeek, line 1522: @knutwannheden is probably the better judge of any possible adverse effects if we were to adopt this, but I found it also works without the extra conditional block, by falling into the var-handling else block just below.

Should we keep this try-catch or can we rely on the existing logic, effectively only removing the ` || vartype instanceof JCErroneous` above?
```suggestion
```

timtebeek (commented):
Thanks a lot @seonwoo960000 ! Promising work; I've tagged @knutwannheden for an additional look following my suggestion, but this this looks good overall already.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1555,6 +1565,7 @@ private J.VariableDeclarations visitVariables(List<VariableTree> nodes, Space fm
         for (int i = 0; i < nodes.size(); i++) {
             JCVariableDecl n = (JCVariableDecl) nodes.get(i);
 
+            // how to capture var ?
```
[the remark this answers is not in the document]
seonwoo960000: Ah, missed that point. I'll remove it right away.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
seonwoo960000: Looks cleaner. Without any adverse effects, let's apply your suggestion.

knutwannheden:
Can we have a test for this in rewrite-java-tck?

traceyyoshima:
@knutwannheden @timtebeek I'm going to take a quick look at the node from the 11, 17, and 21 parser. The compiler should know it's a `var` regardless of an erroneous type. So, I suspect it's possible to check the symbol on the node rather than compare the source position.

traceyyoshima:
> @knutwannheden @timtebeek I'm going to take a quick look at the node from the 11, 17, and 21 parser. The compiler should know it's a `var` regardless of an erroneous type. So, I suspect it's possible to check the symbol on the node rather than compare the source position.

We'll need to use the source code at the cursor position to detect `var`.

@seonwoo960000, please add a test to [VariableDeclarationsTest](https://github.com/openrewrite/rewrite/blob/main/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/VariableDeclarationsTest.java#L64C22-L64C22)

```java
    @Test
    @MinimumJava11
    void unknownVar() {
        rewriteRun(
          spec -> spec.typeValidationOptions(TypeValidation.none()),
          java(
            """
              class Test {
                  void test(Unknown b) {
                      final var a = b;
                  }
              }
              """
          )
        );
    }
```

seonwoo960000:
> Can we have a test for this in rewrite-java-tck?

Hi, @knutwannheden 😄 . Should I add the same tests in `rewrite-java-tck` as well?

seonwoo960000:
Hi @traceyyoshima , I've added the test you provided!

traceyyoshima:
> Hi @traceyyoshima , I've added the test you provided!

Thank you!

> Hi, @knutwannheden 😄 . Should I add the same tests in rewrite-java-tck as well?

Ah, the `VariableDeclarationTest` is located in `rewrite-java-tck`. I agreed with Knut and saved you a few steps.

knutwannheden:
@seonwoo960000 Just for your understanding: The tests in rewrite-java-test don't get executed against all parser versions, so when changing the parser we try to create tests in rewrite-java-tck instead.

knutwannheden (approved):

seonwoo960000:
Thank you for the clarification 😄 , helped me understand better.

traceyyoshima (approved):

PR #3878, opened 2024-01-02
Simplify `!(false)` and `!((false))`

timtebeek:
Simplify additional cases as discovered on
 - https://github.com/openrewrite/rewrite-static-analysis/pull/231

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
@@ -103,29 +103,35 @@ public class SimplifyBooleanExpressionVisitor extends JavaVisitor<ExecutionConte
         J.Unary asUnary = (J.Unary) j;
 
         if (asUnary.getOperator() == J.Unary.Type.Not) {
-            Expression expr = asUnary.getExpression();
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
```diff
@@ -103,33 +103,39 @@ public J visitUnary(J.Unary unary, ExecutionContext ctx) {
...
-            }
+            j = unpackExpression(asUnary.getExpression(), j);
         }
         if (asUnary != j) {
             j = j.withPrefix(asUnary.getPrefix());
         }
         return j;
     }
 
+    private J unpackExpression(Expression expr, J j) {
```
knutwannheden, line 114: Should this also be called something with "negate" maybe?

knutwannheden (approved):

rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
timtebeek: I'd doubted about the name too, but because of the recursive invocation on line 134 I went for unpack instead, as we don't always negate. Open to other suggestions still.

PR #3885, opened 2024-01-05
Missing type tag in AddDependencyVisitor

leveretconey:
## What's changed?

AddDependencyVisitor should add \<type\> tag during XML generation

## What's your motivation?

I've noticed that the AddDependencyVisitor does not generate \<type\>, which prevents us from adding dependencies of the pom type. I believe this is not intentional, as the AddManagedDependencyVisitor is capable of generating \<type\>.


timtebeek (approved):
Perfect, thanks for noticing and fixing!

PR #3887, opened 2024-01-05
Supporting changing name of implicit `value` attributes in ChangeAnnotationAttributeName

tobli:
The `ChangeAnnotationAttributeName` recipe supports renaming annotation attributes (like `@Foo(x = "y")`). If the attribute name is `value` it's possible to not write an assignment and just write the value (like `Bar("z")`), and this case was not handled previously.

## What's changed?
This change adds the following (validated by new tests):
- support for renaming implicit value attributes (`Bar("z")` to `Bar(foo = "z")`)
- support for expanding implicit value attributes (`Bar("z")` to `Bar(value = "z")`)
- ensures the tree is unchanged for an identical rename (`@Foo(x = "y")` to `@Foo(x = "y")`)

## What's your motivation?
Noticed this comment: https://github.com/openrewrite/rewrite-openapi/issues/1#issuecomment-1877633407

How I created the `J.Assignment`, I hacked it to get what I wanted, but there might be a better way. Should there be an auto-format in case the padding of the `=` is different in the current code style?

### Checklist

### DIFF
--- a/rewrite-java/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
@@ -38,4 +38,104 @@ class ChangeAnnotationAttributeNameTest implements RewriteTest {
             )
         );
     }
+
+    @Test
+    void doNotChangeIdenticalAssignment() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeAnnotationAttributeName("java.lang.SuppressWarnings", "value", "value")),
+          java(
+            """
+              @SuppressWarnings(value = {"foo", "bar"})
+              class A {}
+              """
+          )
+        );
+    }
+
+    @Test
+    void renameValueAttribute() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeAnnotationAttributeName("org.example.Foo", "value", "bar")),
+          java(
+            """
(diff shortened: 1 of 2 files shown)


rewrite-java/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
```diff
@@ -38,4 +38,104 @@ class A {}
             )
         );
     }
+
+    @Test
+    void doNotChangeIdenticalAssignment() {
```
tobli, line 43: This would fail with the error
> An empty diff was generated. The recipe incorrectly changed a reference without changing its contents.

rewrite-java/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
```diff
@@ -38,4 +38,104 @@ class A {}
...
+            """
+              @SuppressWarnings(value = {"foo", "bar"})
+              class A {}
+              """
+          )
+        );
+    }
+
+    @Test
+    void rewriteEnumValue() {
```
tobli, line 103: This test case may not be strictly necessary, it was helpful to me during development to get the value assignment case working.

timtebeek (approved):
Love the attention to detail in the tests, and subtle addition of function using only existing parameters. Thanks a lot!

PR #3889, opened 2024-01-05
Validate that Change(Managed)DependencyGroupIdAndArtifactId updates GAV

tobli:
## What's changed?
ChangeDependencyGroupIdAndArtifactId and ChangeManagedDependencyGroupIdAndArtifactId are intended to be set up to *change* maven coordinates. If the pre- and post-coordinates are the same, it's likely that the UpgradeDependencyVersion recipe was intended. This change adds validation of the newGroupId/newArtifactId options.

## What's your motivation?
See https://github.com/openrewrite/rewrite-java-dependencies/issues/55

That the use of `Validated` is correct.

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
@@ -29,6 +29,9 @@ import org.openrewrite.xml.tree.Xml;
 
 import java.util.*;
 
+import static org.openrewrite.Validated.test;
+import static org.openrewrite.internal.StringUtils.isBlank;
+
 @Value
 @EqualsAndHashCode(callSuper = true)
 public class ChangeManagedDependencyGroupIdAndArtifactId extends Recipe {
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
```diff
@@ -100,6 +103,17 @@ public Validated<Object> validate() {
         if (newVersion != null) {
             validated = validated.and(Semver.validate(newVersion, versionPattern));
         }
+        validated =
```
timtebeek, line 106: As compared to the non-managed variant this one lacks the check that there's a `newGroupId` or `newArtifactId`. Was that a conscious choice, or would that make sense to add here too?
```suggestion
        validated = validated.and(required("newGroupId", newGroupId).or(required("newArtifactId", newArtifactId)));
        validated =
```

timtebeek: Nevermind, was reading top to bottom and the tests do indicate this is checked.

timtebeek (approved):
Perfect, looks good to me! I'm fine with breaking downstream temporarily while we make the necessary changes there. Or maybe best to do the downstream changes first, to ensure that everything works as expected with `UpgradeDependencyVersion` there.

timtebeek:
Adding a note for myself Monday: add the same change to the rewrite-java-dependencies variants, do another scan of misuse across projects, and then merge this one.

timtebeek:
Hurdle to clear as identified on Slack
> just realised RetainVersions#plan will always violate this: https://github.com/openrewrite/rewrite/blob/10081ba81d516809b3790b61905be5115aa8f81c/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/RetainVersions.java#L48-L53

PR #3890, opened 2024-01-06
Remove unnnecessary parenthesis from unary expressions

### LINKED ISSUE #3883: `UnnecessaryParenthesesVisitor` could remove parentheses around `!(falseMethod())`
This test could fit in [rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest](https://github.com/openrewrite/rewrite/blob/08b330fb163c08e130a7d53bc8e741f98fe0ae92/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest.java#L41)
```java
@Test
void unwrapNotMethodInvocation() {
    rewriteRun(
      unnecessaryParentheses(style -> style),
      java(
        """
          class Test {
              boolean trueMethod() {
                  return !(falseMethod());
              }
              boolean falseMethod() {
                  return false;
              }
          }
          """,
        """
          class Test {
              boolean trueMethod() {
                  return !falseMethod();
              }
              boolean falseMethod() {
                  return false;
              }
          }
          """
      )
    );
}
```

`UnnecessaryParenthesesVisitor` and `UnnecessaryParenthesesStyle` expanded to include negation, to then remove parentheses (only) where possible. We'd need to implement `visitUnary` and when for instance a method invocation is negated, then remove the parentheses.

Cases like `!(false)` are already handled by `SimplifyBooleanExpressionVisitor`, so likely need not be handled again here.

Dedicated recipe seemed unnecessary given that we have an existing, that's also used from other recipes and languages.

## Additional context
As discovered on
- https://github.com/openrewrite/rewrite-static-analysis/pull/231

ammachado:
## What's changed?
Fixes #3883 

## What's your motivation?
https://github.com/openrewrite/rewrite-static-analysis/pull/231

No

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest.java
@@ -965,4 +965,34 @@ class UnnecessaryParenthesesTest implements RewriteTest {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/3883")
+    @Test
+    void unwrapNotMethodInvocation() {
+        rewriteRun(
+          unnecessaryParentheses(style -> style),
+          java(
+            """
+              class Test {
+                  boolean trueMethod() {
+                      return !(falseMethod());
+                  }
+                  boolean falseMethod() {
+                      return false;
+                  }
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks!

PR #3898, opened 2024-01-09
3623 $ as inner class

MBoegers:
## What's changed?
* TypeUtils handle $ better if determing the FQN
* use TypeUtils in rewrite-java*

## What's your motivation?
I want the fix the handling of $ as class name.

## Any additional context
If finished this should enable https://github.com/openrewrite/rewrite-migrate-java/pull/285

### Checklist


timtebeek:
Great to see, thanks @MBoegers ; saving others a click by copying the remaining failure here:
```
JavaTemplateSubstitutionsTest > anyForInnerClass() FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "A.java":
    diff --git a/A.java b/A.java
    index df29fc4..f075a84 100644
    --- a/A.java
    +++ b/A.java
    @@ -1,6 +1,6 @@ 
     class A {
         void foo() {
    -        $ test = new $();
    +        $ test = new A.$();
         }
         static class $ {
             static $ of() {
    ] 
    expected: 
      "class A {
          void foo() {
              $ test = new $();
          }
          static class $ {
              static $ of() {
                  return new $();
              }
          }
      }"
     but was: 
      "class A {
          void foo() {
              $ test = new A.$();
          }
          static class $ {
              static $ of() {
                  return new $();
              }
          }
      }"

```

rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -39,7 +39,14 @@ public static boolean isString(@Nullable JavaType type) {
     }
 
     public static String toFullyQualifiedName(String fqn) {
-        return fqn.replace('$', '.');
+        /// replace every $ that is not prefixed by a dot in an iterative waz to handle inner classes wich name $
```
JLLeitschuh, line 42:
```suggestion
        /// replace every $ that is not prefixed by a dot in an iterative way to handle inner classes wich name $
```

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
```diff
@@ -1013,7 +1013,7 @@ public static void main(String[] args) {
     @Issue("https://github.com/openrewrite/rewrite/issues/1698")
     @Test
     void correctlyRemoveImportsFromNestedInternalClasses() {
-        rewriteRun(
+            rewriteRun(
```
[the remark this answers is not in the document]
timtebeek:
```suggestion
        rewriteRun(
```

rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -39,7 +39,14 @@ public static boolean isString(@Nullable JavaType type) {
     }
 
     public static String toFullyQualifiedName(String fqn) {
-        return fqn.replace('$', '.');
+        /// replace every $ that is not prefixed by a dot in an iterative way to handle inner classes wich name $
```
timtebeek, line 42:
```suggestion
        // replace every $ that is not prefixed by a dot in an iterative way to handle inner classes whose name is $
```

rewrite-java/src/main/java/org/openrewrite/java/internal/template/Substitutions.java
```diff
@@ -126,7 +126,7 @@ private String substituteTypedPattern(String key, int index, TemplateParameterPa
             if (arrayType.getElemType() instanceof JavaType.Primitive) {
                 s += ((JavaType.Primitive) arrayType.getElemType()).getKeyword();
             } else if (arrayType.getElemType() instanceof JavaType.FullyQualified) {
-                s += ((JavaType.FullyQualified) arrayType.getElemType()).getFullyQualifiedName().replace("$", ".");
+                s += TypeUtils.toFullyQualifiedName(((JavaType.FullyQualified) arrayType.getElemType()).getFullyQualifiedName());
```
knutwannheden, line 129: A `JavaType.FullyQualified` instance currently typically uses the [binary name](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.1), which in the compiler's AST is represented by `Symbol#flatName()`.

I think that instead of changing and using `TypeUtils.toFullyQualifiedName()` to get the fully qualified name of a instances implementing `JavaType.FullyQualified` we would be better off either changing the internal representation or deriving the name from the current representation.

Changing the current internal representation is trickier (with regards to backward compatibility), but adding a utility that can return the "source name" for a `JavaType.FullyQualified` instance should be possible by combining what is returned by `getOwningClass()` and by `getFullyQualifiedName()`.

I am saying this because a class could also be called `Foo$Bar` if you wanted to and that is another `$` we shouldn't be replacing.

PR #3901, opened 2024-01-10
Add support for gradle dependency versions to be updated in gradle.properties

ajohnsonz:
### Checklist


ajohnsonz:
Copying this over from #3665 for ease of reference:

Summarising the suggestions from the Slack thread with timtebeek and shanman190 for posterity:

The `GradleProject` marker is only holding onto the requested/resolved literal values. Unfortunately, it doesn't know the relationship in terms of the properties.

Initially, I'm thinking that we'd have to gather all properties by source path AND all matching dependency declarations GroupId/ArtifactId+variable name might be the easiest, then during the edit phase....
- If we are on a `gradle.properties` file, use the variable key to lookup the GroupId/ArtifactId, resolve the newVersion, then conditionally update the property value.
- If we are on a `build.gradle(.kts)` file, no changes here.

To gather properties by source path and all matching dependency declarations:
- have the existing recipe extend `ScanningRecipe` (instead of `Recipe`)
- with a custom accumulator object that holds those values as fields, returned from `getInitialValue`
- then run two visitors in `getScanner`, one `PropertiesVisitor` and one `GroovyVisitor`, and have those fill your accumulator object
- then in `getVisitor` run a `PropertiesVisitor` (as well as the existing `GroovyVisitor`) that correlates the values in your custom accumulator, to make the required changes on `gradle.properties`

Here's an existing `ScanningRecipe` which scans and edits multiple file types that you can base that side off of, along with the associated test class:
- https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
- https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,275 @@ public Validated<Object> validate() {
...
+                                    String gav = (String) literal.getValue();
+                                    if (gav == null) {
+                                        getCursor().putMessage(UPDATE_VERSION_ERROR_KEY, new IllegalStateException("Unable to update version"));
+                                        return arg;
+                                    }
+                                    Dependency dep = DependencyStringNotationConverter.parse(gav);
+                                    if (dependencyMatcher.matches(dep.getGroupId(), dep.getArtifactId())
+                                        && dep.getVersion() != null
+                                        && !dep.getVersion().startsWith("$")) {
+                                        GradleProject gradleProject = getCursor().firstEnclosingOrThrow(JavaSourceFile.class)
```
shanman190, line 298: We can use the existing `GradleProject` marker from the accumulator here as a simplification.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,275 @@ public Validated<Object> validate() {
...
+                                J.Literal groupLiteral = (J.Literal) groupValue;
+                                J.Literal artifactLiteral = (J.Literal) artifactValue;
+                                //noinspection DataFlowIssue
+                                if (!dependencyMatcher.matches((String) groupLiteral.getValue(), (String) artifactLiteral.getValue())) {
+                                    return m;
+                                }
+                                G.MapEntry versionEntry = (G.MapEntry) depArgs.get(2);
+                                Expression versionExp = versionEntry.getValue();
+                                if (versionExp instanceof J.Literal && ((J.Literal) versionExp).getValue() instanceof String) {
+                                    GradleProject gradleProject = getCursor().firstEnclosingOrThrow(JavaSourceFile.class)
```
shanman190, line 350: Same here with using the existing accumulator GradleProject marker.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,275 @@ public Validated<Object> validate() {
...
-                                    .findFirst(GradleProject.class)
-                                    .orElseThrow(() -> new IllegalArgumentException("Gradle files are expected to have a GradleProject marker."));
-                            String version = dep.getVersion();
-                            try {
-                                String newVersion = "classpath".equals(method.getSimpleName()) ?
-                                        findNewerPluginVersion(dep.getGroupId(), dep.getArtifactId(), version, gradleProject, ctx) :
-                                        findNewerProjectDependencyVersion(dep.getGroupId(), dep.getArtifactId(), version, gradleProject, ctx);
-                                if (newVersion == null || version.equals(newVersion)) {
-                                    return arg;
+                        return equalIgnoringSeparators(sourceFile.getSourcePath(), Paths.get(GRADLE_PROPERTIES_FILE_NAME));
```
shanman190, line 180: We will either want to...

* Use a `Preconditions.check(new FindSourceFiles("**/gradle.properties"), <existing visitor>)`
* Update this comparator to allow matching for any file with a name `gradle.properties` as they can appear multiple times throughout a Gradle multi module project.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -435,4 +436,29 @@ void defaultsToLatestRelease() {
           )
         );
     }
+
+    @Test
+    void versionInPropertiesFile() {
```
shanman190, line 441: I think we should also add a test case that models a multi module project layout where the version is found within a child's `gradle.properties` file.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
ajohnsonz: @shanman190 what about something like this which is used in AddProperty recipe?
`sourceFile.getSourcePath().endsWith("gradle.properties"))`

shanman190: @ajohnsonz, yep! That fits with option 2 that I mentioned previously.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,271 @@ public Validated<Object> validate() {
...
+    @Override
+    public DependencyVersionState getInitialValue(ExecutionContext ctx) {
+        // create the accumulator object
+        return new DependencyVersionState();
+    }
+
     @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor() {
+    public TreeVisitor<?, ExecutionContext> getScanner(DependencyVersionState acc) {
+        // run two visitors in getScanner, one PropertiesVisitor and one GroovyVisitor, and have those fill your accumulator object
```
shanman190, line 139:
```suggestion
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,271 @@ public Validated<Object> validate() {
...
+            }
+        };
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor(DependencyVersionState acc) {
+        MethodMatcher dependencyDsl = new MethodMatcher("DependencyHandlerSpec *(..)");
+        DependencyMatcher dependencyMatcher = new DependencyMatcher(groupId, artifactId, null);
+        return
+            Preconditions.or(
```
shanman190, line 173:
```suggestion
        return Preconditions.or(
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,271 @@ public Validated<Object> validate() {
...
+                                J.Literal groupArtifact = (J.Literal) strings.get(0);
+                                G.GString.Value versionValue = (G.GString.Value) strings.get(1);
+                                if (!(versionValue.getTree() instanceof J.Identifier) || !(groupArtifact.getValue() instanceof String)) {
+                                    return entry;
+                                }
+
+                                // here we have "groupArtifact" which is an object that contains the field "value" which is "com.google.guava:guava:"
+                                // we also have "versionValue" which has a field called "tree" which has a field "simpleName" which is "guavaVersion"
+
+                                Dependency dep = DependencyStringNotationConverter.parse((String) groupArtifact.getValue());
```
shanman190, line 201:
```suggestion
                                Dependency dep = DependencyStringNotationConverter.parse((String) groupArtifact.getValue());
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +123,271 @@ public Validated<Object> validate() {
...
+                                if (dependencyMatcher.matches(dep.getGroupId(), dep.getArtifactId())) {
+                                    String versionVariableName = ((J.Identifier) versionValue.getTree()).getSimpleName();
+                                    if (entry.getKey().equals(versionVariableName)) {
+                                        if (!StringUtils.isBlank(newVersion)) {
+                                            String resolvedVersion = null;
+                                            try {
+                                                resolvedVersion = findNewerProjectDependencyVersion(dep.getGroupId(), dep.getArtifactId(), newVersion, acc.gradleProject, ctx);
+                                            } catch (MavenDownloadingException e) {
+                                                // TODO warn about this exception but continue?
```
shanman190, line 210:
```suggestion
                                                return e.warn(entry);
```

shanman190 (approved):
I think that there are a couple of unused imports, but it looks good. Thanks, @ajohnsonz!

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +121,259 @@ public Validated<Object> validate() {
...
                     List<Expression> depArgs = m.getArguments();
-                    if (depArgs.get(0) instanceof J.Literal || depArgs.get(0) instanceof G.GString || depArgs.get(0) instanceof G.MapEntry) {
-                        m = updateDependency(m, ctx);
-                    } else if (depArgs.get(0) instanceof J.MethodInvocation &&
-                            (((J.MethodInvocation) depArgs.get(0)).getSimpleName().equals("platform") ||
-                                    ((J.MethodInvocation) depArgs.get(0)).getSimpleName().equals("enforcedPlatform"))) {
-                        m = m.withArguments(ListUtils.mapFirst(depArgs, platform -> updateDependency((J.MethodInvocation) platform, ctx)));
+                    if (depArgs.get(0) instanceof J.Literal ||
+                        depArgs.get(0) instanceof G.GString ||
+                        depArgs.get(0) instanceof G.MapEntry) {
```
timtebeek, line 145: Perhaps I'm missing something, but further down we seem to only process `G.GString`; could we then not add the others here?
```suggestion
                    if (depArgs.get(0) instanceof G.GString) {
```

timtebeek: If so we can also change the field type on `DependencyVersionState`, and remove the type cast in `visitEntry` below.

timtebeek (commented):
Great to see these changes! Thanks a lot, also to Shannon for guidance and review. I've pushed some small changes and have one remaining question, mostly for my own understanding.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
shanman190: So I agree partially. Both the `G.GString` and `G.MapEntry` could have an identifier for the version value. Only the `J.Literal` would be valid to skip.

timtebeek: Should we then add handling for G.MapEntry below as well? Or is that out of scope for now?

shanman190: You reminded me about missing that, I've added a comment below to that effect. :)

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +121,259 @@ public Validated<Object> validate() {
...
-                        } catch (MavenDownloadingException e) {
-                            return e.warn(m);
-                        }
-                        if (newVersion == null || version.equals(newVersion)) {
-                            return m;
+
+                    @Override
+                    public org.openrewrite.properties.tree.Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
+                        for (Expression dependencyExpression : acc.dependencyExpressions) {
+                            if (dependencyExpression instanceof G.GString) {
```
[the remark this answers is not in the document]
ajohnsonz: What does a G.MapEntry look like in the real world? (i.e. what would a unit test look like to make sure it's correctly covered?)

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
timtebeek: Great; then I suppose the proper suggestion here is to use
```suggestion
                    if (depArgs.get(0) instanceof G.GString ||
                        depArgs.get(0) instanceof G.MapEntry) {
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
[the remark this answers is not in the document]
shanman190: @ajohnsonz, so here's a test case from the existing test class that showcases map notation.

https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Ftest%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FUpgradeDependencyVersionTest.java#L238

In practice, you'd replace the version with a variable or `G.GString` reference to the variable name in the gradle.properties file.

So what that means would be something like so:
```
implementation group: "com.google.guava", name: "guava", version: guavaVersion
// OR
implementation group: "com.google.guava", name: "guava", version: "${guavaVersion}"
```

Both of the above cases, I see pretty regularly

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -116,189 +120,295 @@ public Validated<Object> validate() {
...
+                            String valueValue = null;
+                            if(arg.getValue() instanceof J.Literal) {
+                                J.Literal value = (J.Literal) arg.getValue();
+                                if(value.getValue() instanceof String) {
+                                    valueValue = (String)value.getValue();
+                                }
+                            } else if(arg.getValue() instanceof J.Identifier) {
+                                J.Identifier value = (J.Identifier) arg.getValue();
+                                valueValue = value.getSimpleName();
+                            }
```
shanman190, line 165: If you're up for it, we should also go ahead and attempt to cover the `G.GString` case coming in here for the version. Much like what was described in on of the PR comments:

```
implementation group: "com.google.guava", name: "guava", version: "${guavaVersion}"
```

shanman190:
I think we've covered everything that I can think of, so unless @timtebeek has anything else, I think we can get this merged and start letting folks bang away at it.

timtebeek (approved):
Great work! Really neat to see this come together.

PR #3905, opened 2024-01-11
`SpacesVisitor`: Remove spaces before statement semicolon

knutwannheden:


knutwannheden:
There is quite a few more places we could make sure to remove extra spaces before semicolon.

Issue #3333, opened 2023-06-18
NameCaseConvention.LOWER_CAMEL should handle mixed case more properly


koppor:
Actual:

```diff
-      PdfDocumentViewModel PDFviewModel = new PdfDocumentViewModel(mockPDF);
+      PdfDocumentViewModel pDFviewModel = new PdfDocumentViewModel(mockPDF);
```

Expected

```diff
-      PdfDocumentViewModel PDFviewModel = new PdfDocumentViewModel(mockPDF);
+      PdfDocumentViewModel pdfViewModel = new PdfDocumentViewModel(mockPDF);
```

(This was originally reported at https://github.com/openrewrite/rewrite-static-analysis/issues/12#issuecomment-1588208014)

seonwoo960000:
According to the `Expected` behavior you explained, it says `PDFviewModel` should be changed to `pdfViewModel`. But shouldn't the original variable name be `PDFViewModel`(`v` -> `V`)?

seonwoo960000:
What would you expect if the variable name was `PDFViewModel`? (`v` to uppercase `V`). Will the result of the variable name changed into `pdfvIewModel`?
```java
-      PdfDocumentViewModel PDFViewModel = new PdfDocumentViewModel(mockPDF);
+      PdfDocumentViewModel pdfvIewModel = new PdfDocumentViewModel(mockPDF);
```

koppor:
I expect `PdfViewModel`. The assumption is that the programmers uppercase the abbreviations and then continue with an upper case letter for the next word. E.g., `XMLParser` -> `XmlParser`, `ISOStandard` -> `IsoStandard`, `TOSCAAnalyzer` -> `ToscaAnalyzer`.

Maybe, you can configure the behavior? I there could also be code which reads `ISOstandard`, `TOSCAanalyzer`, ..

koppor:
**Update** I see that my assumption writing the comment. OMG. Maybe, the recipe should have some build-in abbrevations (PDF, ISO, XML, JSON) and offer a configuration of the known abbrevations?

seonwoo960000:
@koppor, wouldn't checking for built-in abbreviations be expensive? `rewrite` would have to check all variables to see if they contain any built-in abbreviations. 

What are your thoughts on allowing users to define their own built-in abbreviations? This approach offers the following advantages: 
- We don't need to specify abbreviations in the code and can track users' preferences. 
- If there are no abbreviations to check, there won't be any additional cost.

[...]

PR #3906, opened 2024-01-12
Fix LOWER_CAMEL, UPPER_CAMEL to handle uppercase abbreviation better

seonwoo960000:
## What's changed?
`LOWER_CAMEL` and `UPPER_CAMEL` can now handle uppercase abbreviation better 

AS-IS 
```
-      PdfDocumentViewModel PDFviewModel = new PdfDocumentViewModel(mockPDF);
+      PdfDocumentViewModel pDFviewModel = new PdfDocumentViewModel(mockPDF);
```

TO-BE(`LOWER_CAMEL`)  
```
-      PdfDocumentViewModel PDFviewModel = new PdfDocumentViewModel(mockPDF);
+      PdfDocumentViewModel pdfViewModel = new PdfDocumentViewModel(mockPdf);
```

## What's your motivation?
https://github.com/openrewrite/rewrite/issues/3333

@koppor

## Any additional context
[Issue](https://github.com/openrewrite/rewrite/issues/3333#issuecomment-1888128281)

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
@@ -56,6 +56,7 @@ public enum NameCaseConvention {
     UPPER_UNDERSCORE;
 
     private static final Pattern CAMEL_CASE_SPLIT = Pattern.compile("[\\s_-]");
+    private static final int uppercaseAbbreviationMinLength = 3;
 
     /**
      * Formats the input to the style of this {@link NameCaseConvention}.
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
```diff
@@ -149,18 +150,31 @@ private static String lowerUnderscore(String str) {
...
             if (Character.isLowerCase(c)) {
                 allUpperCase = false;
                 break;
             }
+
+            if ((i + 1) >= uppercaseAbbreviationMinLength) {
+                uppercaseAbbreviationIncluded = true;
+                uppercaseAbbreviationEndIndex = i - 1;
+            }
         }
```
koppor, line 167: The check for `i` needs to be done at the end of checking. Thus, following proposal:

```suggestion
        int i = 0;
        while (i < strLength) {
            final char c = str.charAt(i);
            if (Character.isLowerCase(c)) {
                allUpperCase = false;
                break;
            }
            i++;
        }
        if (i > uppercaseAbbreviationMinLength) {
            uppercaseAbbreviationIncluded = true;
            uppercaseAbbreviationEndIndex = i - 1;
        }
```

Note that I also changed the if condition. I got rid of `+ 1` by using another comparison.

rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
```diff
@@ -201,7 +215,8 @@ private static String nameCaseJoiner(String str, boolean lowerCaseFirstLetter, c
                         builder.append(lowerCaseChar);
                     } else if (Character.isUpperCase(last) || last == '.') {
                         builder.append(lowerCaseChar);
-                    } else if (Character.isDigit(last) && (Character.isUpperCase(secondLast) || secondLast == separatorChar)) {
+                    } else if (Character.isDigit(last) && (Character.isUpperCase(secondLast)
+                                                           || secondLast == separatorChar)) {
```
koppor, line 219: I would not reformat code.

koppor (changes requested):
Code looks good

I think, even for my code, I don't need an exception list. Let's see.

---

IT Smalltalk: I needed to file an IntelliJ issue report, because the RegEx matching there does not work for me. - https://youtrack.jetbrains.com/issue/IDEA-342866/Regex-pattern-matching-does-not-work-with-occurence-count



---

Side note: I had to open an issue at IntelliJ for being able to review: https://youtrack.jetbrains.com/issue/IDEA-342866/Regex-pattern-matching-does-not-work-with-occurence-count

rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
seonwoo960000: Great idea. Let's go for that 👍

rewrite-core/src/main/java/org/openrewrite/internal/NameCaseConvention.java
seonwoo960000: Sorry for that unrelated work. Reverted it 🙇

koppor:
Please check CircleCI output.

```
NameCaseConventionTest > lowerCamel(String, String) > [17] input=PDFViewModel, expected=pdfViewModel FAILED
    expected: "pdfViewModel"
     but was: "pdfvIewModel"
```

PR #3917, opened 2024-01-15
Find `@Option` annotations without `example` values

timtebeek:
As came up in Slack: https://github.com/openrewrite/gh-automation/issues/45#issuecomment-1891827588

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/recipes/MissingOptionExampleTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/recipes/MissingOptionExampleTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.recipes;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class MissingOptionExampleTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new MissingOptionExample())
+          .parser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()));
+    }
+
+    @DocumentExample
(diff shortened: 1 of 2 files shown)


timtebeek:
@koppor / @MBoegers since you were involved; if this gets a thumbs up I'll merge and deploy. :)

rewrite-java-test/src/test/java/org/openrewrite/java/recipes/MissingOptionExampleTest.java
```diff
@@ -0,0 +1,129 @@
...
+                      return "Find `@Option` annotations that are missing `example` values.";
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void skipBoolean() {
```
koppor, line 105: I don't understand the intention of this test

koppor (commented):
Looks good. I miss some documentation for one test case

rewrite-java-test/src/test/java/org/openrewrite/java/recipes/MissingOptionExampleTest.java
koppor: Ah, I get it. Booleans don't need ab example.

koppor (approved):

rewrite-java-test/src/test/java/org/openrewrite/java/recipes/MissingOptionExampleTest.java
timtebeek: indeed, we already insert [`false` by default for booleans](https://github.com/openrewrite/rewrite-recipe-markdown-generator/blob/821ab474934d72d21a5ec5b956ea92b2c155759f/src/main/kotlin/org/openrewrite/RecipeMarkdownGenerator.kt#L1064-L1065).

MBoegers:
nice! just out of curiositiy, wouldn't be a `//todo:` be easier to find?

timtebeek:
Right now we produce a search marker, which has separate handling in the Moderne Platform to show the results there. That would be my main goal with this recipe to have a list of missing examples, and a hint at how to resolve that on a PR. Inserting a comment is also an option, but lacks that more sophisticated handling.

MBoegers:
Yes, I understand that they are SearchResults but simply adding “todo” to the comment would make visible for InteliJs TODO tab and combined with a Issue link it would be possible to have an automated todo list. But it will also work this way 😉

Von: Tim te Beek ***@***.***>
Antworten an: openrewrite/rewrite ***@***.***>
Datum: Montag, 15. Januar 2024 um 20:27
An: openrewrite/rewrite ***@***.***>
Cc: "Bögershausen, Merlin" ***@***.***>, Mention ***@***.***>
Betreff: Re: [openrewrite/rewrite] Find ***@***.***` annotations without `example` values (PR #3917)


Right now we produce a search marker, which has separate handling in the Moderne Platform to show the results there. That would be my main goal with this recipe to have a list of missing examples, and a hint at how to resolve that on a PR. Inserting a comment is also an option, but lacks that more sophisticated handling.

timtebeek:
Adding a link to the recipe as it can be run from the platform:
https://app.moderne.io/recipes/org.openrewrite.java.recipes.MissingOptionExample
From this initial run it appears we only have a handful of cases still to cover 🎉 
https://app.moderne.io/results/cCwlMe3wr

image

timtebeek:
Looking over the results once more we can flag even fewer cases by ignoring primitives https://github.com/openrewrite/rewrite/commit/b044c5de771a6161d27d5c6170c2487ff5c47d90

Issue #3916, opened 2024-01-15
RemoveObjectsIsNull should remove unused import java.util.Objects


koppor:
When applying RemoveObjectsIsNull, checkstyle says

    [ERROR] /home/runner/work/jabref/jabref/src/main/java/org/jabref/gui/shared/SharedDatabaseUIManager.java:4:8: Unused import - java.util.Objects. [UnusedImports]

This should not happen. Can the recipe simply call https://docs.openrewrite.org/recipes/java/removeunusedimports afterwards? Maybe with a parameter, to only handle `Objects`?

Detected at: https://github.com/JabRef/jabref/pull/10788

Issue #3915, opened 2024-01-15
RemoveObjectsIsNull also rewrites boxed ints


koppor:
We have an issue in the code, where we have `Objects.isNull(i)`, where `i` is an `int`.

Then, Java boxes the integer. when `i != null`  is written, the auto boxing does not work and leads to a compile error:

    operand types for binary operator '!='

Wish: when the method parameter type is not an object, do not rewrite.

See https://github.com/JabRef/jabref/pull/10788

timtebeek:
Picked up in 
- https://github.com/openrewrite/rewrite/pull/3918

PR #3918, opened 2024-01-15
Change primitive handling in RemoveObjectsIsNull

timtebeek:
Fixes https://github.com/openrewrite/rewrite/issues/3915
Fixes https://github.com/openrewrite/rewrite/issues/3916

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
@@ -36,18 +36,17 @@ class RemoveObjectsIsNullTest implements RewriteTest {
         rewriteRun(
           java(
             """
-              import static java.util.Objects.isNull;
+              import java.util.Objects;
               public class A {
                   public void test() {
                       Boolean a = true;
-                      if (java.util.Objects.isNull(a)) {
+                      if (Objects.isNull(a)) {
                           System.out.println("a is null");
                       }
                   }
               }
               """,
             """
-              import static java.util.Objects.isNull;
               public class A {
                   public void test() {
                       Boolean a = true;
@@ -117,7 +116,6 @@ class RemoveObjectsIsNullTest implements RewriteTest {
               }
               """,
             """
-              import static java.util.Objects.nonNull;
               public class A {
                   public void test() {
                       Boolean a = true;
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
```diff
@@ -63,58 +62,35 @@ public void test() {
...
-            """
-              import static java.util.Objects.nonNull;
+              import static java.util.Objects.isNull;
               public class A {
-                  public void test() {
-                      boolean a = true;
-                      if (a != null) {
-                          System.out.println("a is non-null");
+                  public void test(boolean a) {
+                      if (isNull(a)) {
```
koppor, line 92: This should be changed, shouldn't it?

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
```diff
@@ -125,27 +101,26 @@ public void test() {
 
     @Issue("https://github.com/openrewrite/rewrite/issues/1547")
     @Test
-    void transformCallToNonNullNeedsParentheses() {
```
koppor, line 128: Was this test completely removed by intention?

koppor (commented):
Sponteneous comments just checking the diff.

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
timtebeek: I had my doubts about this case; when we upcast from `boolean` to `Boolean` it doesn't really make sense to even call `isNull` with an upcast to `Boolean` for that method call. Perhaps we should completely remove the method call in those cases?

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
timtebeek: Yes after the recipe changes this was a duplicate. We can revise this of course based on the other feedback here, but up to this commit it was intentional.

rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
koppor: Oh, yeah! And replace !isNull by `false` 😅

timtebeek: Yup; I've changed the logic now to add simple `true` or `false` instead. [In rewrite-launchdarkly](https://github.com/openrewrite/rewrite-launchdarkly/blob/87279dabe8a7d57260575381f4da40dd2262f251/src/main/java/org/openrewrite/launchdarkly/RemoveBoolVariation.java#L67) we follow such a replacement up with `SimplifyConstantIfBranchExecution`, but that would require this recipe to move to rewrite-static-analysis. 🤔

PR #3920, opened 2024-01-16
Allow regular expressions on `ChangeTagAttribute`

ammachado:
## What's changed?
Allow the usage of regular expressions on `ChangeTagAttribute`

## What's your motivation?
Replace namespaces/namespaces URIs on `schemaLocation`.

No

No

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
@@ -88,7 +88,7 @@ public class ChangeTagAttribute extends Recipe {
                 }
 
                 String stringValue = attribute.getValueAsString();
-                if (oldValue != null && !stringValue.startsWith(oldValue)) {
+                if (oldValue != null) {
                     if (Boolean.TRUE.equals(regex) && !Pattern.matches(oldValue, stringValue)) {
                         return attribute;
                     } else if ((regex == null || Boolean.FALSE.equals(regex)) && !stringValue.startsWith(oldValue)) {
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
@@ -91,7 +91,8 @@ public class ChangeTagAttribute extends Recipe {
                 if (oldValue != null) {
                     if (Boolean.TRUE.equals(regex) && !Pattern.matches(oldValue, stringValue)) {
                         return attribute;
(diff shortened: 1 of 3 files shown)


timtebeek:
Thanks for getting this started ; what part of this makes it still a draft? Do you want to try it in use first before a review & merge?

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
```diff
@@ -75,25 +83,31 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
+                    }
                 }
 
                 if (newValue == null) {
                     //noinspection DataFlowIssue
                     return null;
                 }
 
-                String changedValue = (oldValue != null) ? attribute.getValueAsString().replace(oldValue, newValue): newValue;
+                String changedValue = (oldValue != null) ? stringValue.replaceAll(oldValue, newValue) : newValue;
```
timtebeek, line 104: I wonder if we should have separate the handling here between regex and non-regex code paths. `replace(a,b)` did exact replacements, whereas `replaceAll(a,b)` introduces regex matching. Seeing how `oldValue` can contain any string, that might then change the behavior around special characters like dots.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
```diff
@@ -75,25 +83,31 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
                     return attribute;
                 }
-                if(oldValue != null && !attribute.getValueAsString().startsWith(oldValue)) {
-                    return attribute;
+
+                String stringValue = attribute.getValueAsString();
+                if (oldValue != null) {
+                    if (Boolean.TRUE.equals(regex) && !Pattern.matches(oldValue, stringValue)) {
+                        return attribute;
+                    } else if ((regex == null || Boolean.FALSE.equals(regex)) && !stringValue.startsWith(oldValue)) {
```
timtebeek, line 94:
```suggestion
                    }
                    if ((regex == null || Boolean.FALSE.equals(regex)) && !stringValue.startsWith(oldValue)) {
```

timtebeek (commented):
Minor comment that I think we should be able to work through quickly. Thanks for getting this started!

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttribute.java
ammachado: You're right, I just remembered https://github.com/openrewrite/rewrite-logging-frameworks/pull/120. 😅

timtebeek (approved):
Perfect, thanks a lot!

PR #3921, opened 2024-01-16
Remove the concept of cycles from OpenRewrite

sambsnyd:
The concept of cycles within the `RecipeScheduler` is broadly obsolete. For a while the implicit expectation has been that recipes do all of their work within a single cycle. This expectation is crucial for good performance of complex composite recipes on large repositories. This PR formalizes this expectation by removing the multi-cycle infrastructure from `RecipeScheduler` and elsewhere the concept touched.

### Recipe Authoring

Some recipes are deliberately designed to be recursive, requiring multiple applications to get to their desired result. 
For these situations I have introduced `Repeat.repeatUntilStable()` to be used within their `getVisitor()` method to internally run a visitor in a loop until the output is stable. In this way a recipe can still be designed in this recursive style without causing an entirely new run of all other recipes. 

### Testing

`RecipeTest` still validates that recipes are stable, not making changes after the first application of the recipe. This check is very valuable for catching a swathe of recipe authoring mistakes where the recipe fails to leave code it shouldn't change alone. This is a particularly common category of mistake among new recipe authors. 

Previously when recipe authors had an unstable recipe they would get a cryptic error similar to:
```
Expected:
Hello world
Actual:
Hello worldHello world
```

Now they'll get this descriptive and actionable message:
```
Running the recipe on its expected output resulted in additional changes.
For correctness and performance recipes are expected to be stable, making all of their changes in a single application.
This error is typically resolved by adding defensive guards against unnecessary changes.
You can opt out of this validation by setting recipeOutputStabilityValidation(false).
The recipe produced unstable results for these sources: 

------------------------------------------------------------------------
file.txt after first recipe application
------------------------------------------------------------------------
Hello world
------------------------------------------------------------------------
file.txt after second recipe application
------------------------------------------------------------------------
Hello worldHello world
```

The most common place that users directly interact with recipe cycles is in tests, where they could configure `RecipeSpec.cycles()` and `RecipeSpec.expectedCyclesThatMakeChanges()`. To reduce breakage I have marked these `@Deprecated` and effectively made them aliases of the new opt-out `RecipeSpec.recipeOutputStabilityValidation(false)`. Whenever we decide to fully remove those methods we can simultaneously release a recipe to complete this replacement.


rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java
```diff
@@ -60,57 +57,39 @@ private LargeSourceSet runRecipeCycles(Recipe recipe, LargeSourceSet sourceSet,
...
-            } finally {
-                // Clear any messages that were added to the root cursor during the cycle. This is important
-                // to avoid leaking memory in the case when a recipe defines a static TreeVisitor. That
-                // TreeVisitor will still contain a reference to this rootCursor and any messages in it
-                // after recipe execution completes. The pattern of holding a static TreeVisitor isn't
-                // recommended, but isn't possible for us to guard against at an API level, and so we are
-                // defensive about memory consumption here.
-                rootCursor.clearMessages();
-            }
+            after.afterCycle(true);
```
knutwannheden, line 80: If we proceed with this, we should probably at least also rename this method. We probably still need some kind of life cycle method, so I don't think we can remove it.

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -236,7 +236,7 @@ public Collection<Recipe> listRecipes() {
                 }
             }
             DeclarativeRecipe recipe = new DeclarativeRecipe(name, displayName, description, tags,
-                    estimatedEffortPerOccurrence, source, (boolean) r.getOrDefault("causesAnotherCycle", false), maintainers);
```
knutwannheden, line 239: We also need to update the `/rewrite-core/openrewrite.json` accordingly.

PR #3923, opened 2024-01-16
Move `MavenDownloadingException` to `internal` package

knutwannheden:


knutwannheden:
Went for a different solution.

PR #3924, opened 2024-01-16
Gradle `AddDependencyVisitor` should not use default Maven repos

knutwannheden:
When downloading Maven metadata the `AddDependencyVisitor` should not configure any default repos and must therefore use the `MavenPomDownloader(ExecutionContext)` constructor.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
@@ -393,7 +393,7 @@ public class AddDependencyVisitor extends GroovyIsoVisitor<ExecutionContext> {
     }
 
     private static MavenMetadata downloadMetadata(String groupId, String artifactId, List<MavenRepository> repositories, ExecutionContext ctx) throws MavenDownloadingException {
-        return new MavenPomDownloader(emptyMap(), ctx, null, null)
+        return new MavenPomDownloader(ctx)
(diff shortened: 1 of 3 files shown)


shanman190:
@knutwannheden, this should work just fine as the total list of repositories coming from Gradle will be in the list handed to the download methods.

knutwannheden:
@shanman190 Thanks! I just noticed today that we don't properly support repos defined in a `subprojects` block and that issue was masked by the fact that default repos got applied.

shanman190:
So actually, I think that is handled already. The repositories are pulled from the `RepositoryHandler` when the Gradle marker is being built. So when we are processing a subproject those will be there and available. ;)

https://github.com/openrewrite/rewrite-gradle-tooling-model/blob/b6a81d97772d82d19f31ee2dff38bec1726983f4/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java#L58

knutwannheden:
Yes, kind of. IIRC the issue was when the root project in addition to the repositories also declared dependencies inside the `subprojects` block. When running a recipe like UpgradeDependencyVersion across the root project it would fail if the root project doesn't have repos defined of ita own.

shanman190:
Ahh, I see, I see. Yeah, for that case there would need to be something to contend with what Gradle proper would consider an anti pattern today. For the subproject block, we would want to consider the dependencies block from the perspective of the child repositories rather than the root project repositories, but as you can imagine that's going to be a little... difficult.

knutwannheden:
Thanks for confirming. Sam also suggests we wait with supporting this.

PR #3929, opened 2024-01-17
Fixed ChangeType for stackoverflow

zacscoding:
## What's changed?

Check if `classType` is the same as `classType.getOwningClass()` in ChangeType::getTopLevelClassName()

## What's your motivation?

Fixed `ChangeType` recipe to prevent stack over flow.

## Any additional context

I used `ChangeType` recipe to change `@ApiModelProperty` annotation to `@Schema` and below is my sample kotlin code.

```java
class MyClass {
  @ApiModelProperty("description")
  private val code: String? = null

  companion object {
    const val MY_CODE = "001"
  }
}
```

Above sample code is working well in the unit test code, but stack over flow error occurs if i run rewrite recipes.
More details, in the first call, `classType` is `MyClass$Companion` and `classType.getOwningClass()` is `MyClass`.
In the recursive calls, `classType` and `classType.getOwningClass()` is the MyClass repeatedly.


rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -634,7 +634,8 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, At
     }
 
     public static JavaType.FullyQualified getTopLevelClassName(JavaType.FullyQualified classType) {
-        if (classType.getOwningClass() == null) {
+        if (classType.getOwningClass() == null || TypeUtils.fullyQualifiedNamesAreEqual(classType.getFullyQualifiedName(),
+                classType.getOwningClass().getFullyQualifiedName())) {) {
```
timtebeek, line 638:
```suggestion
                classType.getOwningClass().getFullyQualifiedName())) {
```

timtebeek:
Thanks for the proposed fix @zacscoding ! Seems like this is specific to Kotlin, so I'm doing a quick tag of @knutwannheden in case he'd like to solve this differently given the context above.

zacscoding:
Thanks @timtebeek , I look forward to the problem being resolved :)

knutwannheden:
@zacscoding As you are saying a unit test like the following passes without issues. I wonder what the problem is when run differently.

```java
    @Test
    void companion() {
        rewriteRun(
          spec -> spec.recipe(new ChangeType("a.ApiModelProperty", "a.Schema", true)),
          kotlin(
            """
              package a

              annotation class ApiModelProperty
              """,
            SourceSpec::skip
          ),
          kotlin(
            """
              import a.ApiModelProperty

              class MyClass {
                @ApiModelProperty
                private val code: String? = null
                            
                companion object {
                  const val MY_CODE = "001"
                }
              }
              """,
            """
              import a.Schema

              class MyClass {
                @Schema
                private val code: String? = null
                            
                companion object {
                  const val MY_CODE = "001"
                }
              }
              """
          )
        );
    }
```

zacscoding:
@knutwannheden  Sorry for the late response. I just tried to execute rewriteRun command with our recipe jar :(

zacscoding:
@knutwannheden @timtebeek  
Could you please review this PR again?
I think this PR is appropriate as defensive code as well.

jkschneider:
I don't see this PR as harmful at any rate, even if we can't fully reproduce. If it solves @zacscoding issue, I think it should be merged without further ado.

PR #3933, opened 2024-01-18
Adding support for plugin versions in gradle.properties file

ajohnsonz:
## What's changed?
Changing the `UpgradePluginVersion` recipe from a Recipe to a ScanningRecipe so that we can match plugin version variables with anything declared in a `gradle.properties` file

@timtebeek @shanman190 who are aware of the context behind this and the similar improvements that were made to `UpgradeDependencyVersion`

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/IsBuildGradle.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/IsBuildGradle.java
@@ -16,9 +16,8 @@
 package org.openrewrite.gradle;
 
 import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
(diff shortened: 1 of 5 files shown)


shanman190:
One question that I have is if we should borrow the similar code from https://github.com/openrewrite/rewrite/commit/6f670f821f32ac97634df5899d743cf06fde40e5 for the `GroovyVisitor` within the `getVisitor(..)` block. There's a minor performance benefit to not having to look up those values a second time, but we should if we do choose to go that path we should take the null checking from Knut's commit.

shanman190 (approved):

timtebeek:
Thanks a lot @ajohnsonz ! Great to see support added here. 🎉

PR #3935, opened 2024-01-18
Recipe to comment imports

ammachado:
## What's changed?
Added a new recipe to comment imports

## What's your motivation?
Add a comment to types that were removed

- The `import` statement is not included in the comment as expected
- The failing test output is not showing the diffs anymore

@timtebeek 

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/CommentImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/CommentImport.java
@@ -1,0 +1,81 @@
+package org.openrewrite.java;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import lombok.With;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.search.UsesType;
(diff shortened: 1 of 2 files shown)


anandfresh (approved):

rewrite-java/src/main/java/org/openrewrite/java/CommentImport.java
```diff
@@ -0,0 +1,96 @@
...
+                        import_ = (J.Import) super.visitImport(import_, ctx);
+                        if (import_.getMarkers().findFirst(ImportAlreadyCommented.class).isPresent()) {
+                            return import_;
+                        }
+                        if (import_.getTypeName().equals(fullyQualifiedTypeName)) {
+                            boolean multilineComment = Boolean.TRUE.equals(multiline);
+                            String formattedComment = multilineComment ? "\n" + commentText + "\n" : commentText;
+                            Comment comment = new TextComment(multilineComment, formattedComment, import_.getPrefix().getWhitespace(), Markers.EMPTY);
+                            return import_.withComments(ListUtils.concat(import_.getComments(), comment))
+                                    .withMarkers(import_.getMarkers().add(new ImportAlreadyCommented(UUID.randomUUID())));
```
timtebeek, line 83: Looking at this now it appears you'd want to comment out the original import, instead of adding comment to the import; is that correct? In that case I think you'd have to remove this import, find whatever is the next LST element, and merge any comments on this import, that next LST element and a commented out import there.

Slightly more involved; but we'd used similar logic when we for instance add or remove an import: that also needs to shuffle around any existing comments and tag them onto the intuitively right LST element.

PR #3936, opened 2024-01-18
`MavenParser` allows skipping dependency resolution

knutwannheden:
This could be useful when loading a Maven pom.xml only to iterate over its properties, plugins, etc.


knutwannheden:
Probably isn't very useful after all. Closing again.

timtebeek (approved):

Issue #3938, opened 2024-01-20
`UpgradeSpringBoot_3_2` changes `build.gradle` `sourceCompatibility` and `targetCompatibility` from 21 back to 17


aholland:
Title says it all - sorry for the brief report! Also I think they should ideally be JavaVersion.21 in Gradle 8.5 (to be fair you might already be detecting which convention to use - I had a plain `21` before I ran the script).

timtebeek:
Picked up in
- https://github.com/openrewrite/rewrite/pull/3939

PR #3939, opened 2024-01-21
Do not downgrade SourceCompatibility by default

timtebeek:
Fixes https://github.com/openrewrite/rewrite/issues/3938

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
@@ -57,6 +57,12 @@ public class UpdateJavaCompatibility extends Recipe {
     @Nullable
     DeclarationStyle declarationStyle;
 
+    @Option(displayName = "Allow Downgrade",
+            description = "Allow downgrading the Java version.",
+            required = false)
+    @Nullable
+    Boolean allowDowngrade;
+
     @Override
     public String getDisplayName() {
         return "Update Gradle project Java compatibility";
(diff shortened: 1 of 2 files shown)


shanman190:
@timtebeek, something interesting here is that the intermediate recipe already has a strict check for strict upgrade. 

https://github.com/openrewrite/rewrite-migrate-java/blob/main/src%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fjava%2Fmigrate%2FUpgradeJavaVersion.java#L69

timtebeek:
Hmm that is odd indeed; that conditional looks for the JavaVersion marker on `build.gradle files`, but when I look at the Gradle plugin to see [we create the JavaVersion marker](https://github.com/openrewrite/rewrite-gradle-plugin/blob/dd0617f3ecfe4d332f6a06665229c4ddc2345b29/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java#L664-L667), but at first glance I only see that added to Java, Kotling and `.groovy` files, not the build.gradle files we evaluate here. Am I missing something? Or should we add those markers to Gradle files too still?

I think this might have only shown up now as folks start to run recipes that chain in a Java 17 migration while being on Java 21; it's been rare that we had cases where folks are on newer Java versions from what we prescribe.

shanman190:
Ok, now I see what happened. The `UpgradeJavaVersion` recipe pre-8.0 actually added the `JavaVersion` marker to every LST element as if performed that targeted upgrade. This was then used to block the downgrading automatically. When it was converted to 8.x it now is only conditionally adding the markers. So to your point, if things aren't aligned with the prescriptive guidance, then it now acts funny. 

So longer story short, I think your change does in fact fix the issue fully, but we do have some dead code in the `UpgradeJavaVersion` recipe to now misleading behavior.

timtebeek:
Picked that up in https://github.com/openrewrite/rewrite-migrate-java/pull/400 to prevent any confusion, thansk!

PR #3942, opened 2024-01-22
Fix `AddDependency` when dependent on other recipes

knutwannheden:
The `AddDependency#onlyIfUsing` option currently only works if the source files already in the initial state (before any other recipes were applied) already contained references to what is specified in the `onlyIfUsing` option. Thus, if another recipe were to modify the code such that it adds a reference to code matching `onlyIfUsing`, this edit on its own would not lead to the dependency being added.

The problem is that `AddDependency` only runs in one cycle and thus uses the scanning phase to decide if a dependency needs to be added. To properly describe the use case described above, it should however request a second cycle in some scenarios.

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
@@ -23,6 +23,7 @@ import org.openrewrite.Issue;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
 import org.openrewrite.gradle.marker.GradleProject;
 import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.ChangePackage;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -116,6 +116,8 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
     @Nullable
     Boolean acceptTransitive;
 
+    transient boolean requiresAnotherCycle = true;
```
knutwannheden, line 119: In a first iteration I was planning on having this transient field (initially `false`) which the recipe in the edit phase sets to `true` if it detects any source files which match `onlyIfUsing` so that a second cycle gets run.

A more proper fix would be a dedicated API. One idea would be to add an `ExecutionContext` parameter to the existing `Recipe#causesAnotherCycle()` method, which would allow the recipe to store something in the `ExecutionContext`. Another idea would be to add a new API to `RecipeRunCycle` (can be obtained via `ExecutionContext#getCycleDetails()`) or directory on `ExecutionContext`. I am thinking something like a `requestAnotherCycle()`. But the existing  `Recipe#causesAnotherCycle()` would likely have to remain for declarative recipes.

Depending on the approach, some additional changes may be required to the recipe scheduler, as it currently expects the recipe requesting another cycle to also have performed changes in the current cycle. This may not always be the case.

shanman190:
In the past the detail on `onlyIfUsing` was that the type was already being used on the _before_ side, rather than the _after_ side.

knutwannheden:
@shanman190 I would have to look at the history of the recipe, but assuming it was a "scanning recipe" before 8.0, it would still have picked up on changes performed by earlier recipes, as the scanning recipes worked entirely different back then. But it would be worth it to verify this.

shanman190:
Ahh, right right. Change type would put the type in use, then the `onlyIfUsing` would have tripped automatically. Now I see the bug you're addressing.

knutwannheden:
I still think this could be a valid PR, but as long as nobody reports this, I think I will just close it.

PR #3954, opened 2024-01-29
`@PackageRecipeTags` annotation on `package-info.java`

JLLeitschuh:
Adds a package level annotation for adding tags to recipes at the
package-level scope.

## What's changed?

`Recipe#getDescriptor` now looks for Recipe tags on `package-info.java`, as well as parent `package-info.java`.

## What's your motivation?

When writing recipes that all fall into a similar set of tags, for example `security` recipes, you likely want to
tag recipes at the package level, so you don't need to add the same exact tag to every single `Recipe`.

Is the `PackageRecipeTags` annotation name fine?

Anyone!

Could write an OpenRewrite-OpenRewrite recipe that modifies every recipe in a package to add a tag.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/PackageRecipeTags.java
+++ b/rewrite-core/src/main/java/org/openrewrite/PackageRecipeTags.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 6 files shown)


jkschneider:
I like the concept, but rather than adding a package level annotation, I think this concept would better fit in our `category` resource concept. Categories are tied to packages so putting tags on the category should apply to all recipes in that category. There is a [category yml](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/resources/META-INF/rewrite/java-categories.yml) format.

JLLeitschuh:
Can you elaborate upon what the difference between categories and tags, at a high level, are supposed to be.

I also had a similar realization, once I realized that categories existed. The annoying downside is that those category added tags won't show up when you call `getDescriptor().getTags()` but I don't think that's a big deal.

PR #3955, opened 2024-01-29
fix: prevent all referenced properties from updating to latest regardless of recipe

ajohnsonz:
## What's changed?

Fixing a bug in the Properties visitor for the rewrite-gradle plugin which failed to check whether the property currently being visited is referenced by the GroupArtifact that the recipe refers to

@timtebeek and @shanman190  have the context for this

### Checklist


shanman190 (approved):

shanman190:
Thanks, @ajohnsonz!

PR #3962, opened 2024-02-01
Also update Maven managed dependency by default in `ChangeDependencyGroupIdAndArtifactId`

timtebeek:
## What's changed?
Also update Maven managed dependencies by default in ChangeDependencyGroupIdAndArtifactId 

## What's your motivation?
By customer request to be less surprising by default: a change in the GAV should propagate to managed dependencies too by default, not require the same parameters passed into `ChangeManagedDependencyGroupIdAndArtifactId` a second time.

There was a surprising bit of depth to this into how the various arguments work together, and what might be desired from a user's perspective; I hope the named variables in the decision logic help follow along and troubleshoot if there are any issues.

I've chosen to only update the managed dependency GAV is a version argument is already provided, as it's quite unlikely that a GAV change uses the same version, and we don't want to have failures to resolve the managed dependency. I specifically chose not to insert a version ourselves if none is provided; it is then better not to change the managed dependency at all.

This does what a user would expect by default: a changed GAV extends to managed dependencies unless explicitly disabled.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -88,18 +88,18 @@ public class ChangeDependencyGroupIdAndArtifactId extends Recipe {
     @Nullable
     Boolean overrideManagedVersion;
 
+    @Option(displayName = "Update dependency management",
+            description = "Also update the dependency management section. The default for this flag is `true`.",
+            required = false)
+    @Nullable
+    Boolean changeManagedDependency;
+
     public ChangeDependencyGroupIdAndArtifactId(String oldGroupId, String oldArtifactId, @Nullable String newGroupId, @Nullable String newArtifactId, @Nullable String newVersion, @Nullable String versionPattern) {
-        this.oldGroupId = oldGroupId;
-        this.oldArtifactId = oldArtifactId;
-        this.newGroupId = newGroupId;
-        this.newArtifactId = newArtifactId;
-        this.newVersion = newVersion;
-        this.versionPattern = versionPattern;
-        this.overrideManagedVersion = false;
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -146,18 +147,28 @@ public Validated<Object> validate() {
...
             @Nullable
             final VersionComparator versionComparator = newVersion != null ? Semver.validate(newVersion, versionPattern).getValue() : null;
             @Nullable
             private Collection<String> availableVersions;
 
             @Override
-            public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
+            public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
+                // Any managed dependency change is unlikely to use the same version, so only update selectively.
+                if ((changeManagedDependency == null || changeManagedDependency) && newVersion != null || versionPattern != null) {
```
timtebeek, line 159: This seemed a reasonable limit to impose, as the recipe itself also has a `if (newVersion != null)`.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -146,18 +147,28 @@ public Validated<Object> validate() {
...
+                    doAfterVisit(new ChangeManagedDependencyGroupIdAndArtifactId(
+                            oldGroupId, oldArtifactId,
+                            Optional.ofNullable(newGroupId).orElse(oldGroupId),
+                            Optional.ofNullable(newArtifactId).orElse(oldArtifactId),
```
timtebeek, line 163: These arguments are not nullable on `ChangeManagedDependencyGroupIdAndArtifactId`, and either one must already be different as per the validation here.

knutwannheden (approved):

PR #3963, opened 2024-02-01
`JavaTemplate` matching on parameterized types

jkschneider:
Adding support for parameterized types in the grammar of the template parser and for matching as performed by `JavaTemplate#matches()`.


knutwannheden:
I think now it is basically working. I had to change the internal representation we use for template parameters to use a `JavaType` rather than a plain `String` for the parameter type, otherwise we can't match parameterized types.

In this area there is still some performance improvements that can be made: We should change the `JavaTemplate#matcher()` instance method so that it caches the matcher in a precompiled form (like `Pattern#compile()` basically). That would then end up caching the template parameters we create in `JavaTemplateSemanticallyEqual#createTemplateParameters()`: https://github.com/openrewrite/rewrite/blob/6acddbf161bd51e92d034dec12943a095fe2c483/rewrite-java/src/main/java/org/openrewrite/java/JavaTemplateSemanticallyEqual.java#L65-L124

But I will leave this optimization for another PR.

Further, I am quite sure there are still bugs in the area of `TypeUtils#isAssignableTo()` (which has grown very organically), but also that I will leave for another day.

knutwannheden (approved):
LGTM.

PR #3965, opened 2024-02-02
Make ChangePropertyValue regex allow for multiple replacements of same regex pattern in single value

### LINKED ISSUE #3964: ChangePropertyValue regex does not allow for multiple replacements of same regex pattern.
In yaml’s ChangePropertyValue I am having issues finding the correct regex that works.
Based on [this line](https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java#L107), it should replace all occurences of the regex with its replacement.
However, replacement is not triggered as [this matchesOldValue condition](https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java#L89), renders false.

Besides that, I found that the properties recipe behaves similar to yaml recipe, but does not have the same checks/safeties built-in. 

Would it be an option if a create PR that accepts another boolean as optional input: eg. partialMatch or partialReplace.
That boolean would trigger for regex and string a contains mechanism instead of equals.
For replacement it would trigger all occurences to be changed for regex and for String (rather than sending back new value)

Triggering this recipe multiple times with regex that does prefix/suffix optionals. (but then, how much occurences of the pattern can you have?)

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
PR is following

Jenson3210:
## What's changed?
Made properties and yaml recipe behave the same + altered check if a property value matches the regex. ISO making it complete match, it is now searching if there is a partial match of the regex. This way it's opened up for multiple times te same replacement in single run.

## What's your motivation?
Fix #3964 

Running same recipe multiple times after each other, but that's not really a good approach (I cannot know up front how often a particular replacement needs to happen)

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddProperty.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddProperty.java
@@ -127,7 +127,7 @@ public class AddProperty extends ScanningRecipe<AddProperty.NeedsProperty> {
                         sourceFile.getSourcePath().endsWith("gradle.properties")) {
                         Tree t = !Boolean.TRUE.equals(overwrite) ?
                                 sourceFile :
-                                new ChangePropertyValue(key, value, null, false, null)
+                                new ChangePropertyValue(key, value, null, false, false, null)
                                         .getVisitor().visitNonNull(sourceFile, ctx);
                         return new org.openrewrite.properties.AddProperty(key, value, null, null)
                                 .getVisitor()
@@ -136,7 +136,7 @@ public class AddProperty extends ScanningRecipe<AddProperty.NeedsProperty> {
                 } else if (sourceFile.getSourcePath().endsWith("gradle.properties")) {
                     Tree t = !Boolean.TRUE.equals(overwrite) ?
                             sourceFile :
-                            new ChangePropertyValue(key, value, null, false, null)
+                            new ChangePropertyValue(key, value, null, false, false, null)
                                     .getVisitor().visitNonNull(sourceFile, ctx);
                     return new org.openrewrite.properties.AddProperty(key, value, null, null)
                             .getVisitor()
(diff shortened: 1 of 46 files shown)


timtebeek:
Thanks for getting this started @Jenson3210 ! While your change is indeed the safe way to go about this, I've also glanced at how these recipes are currently used, and don't often see `oldValue` and `regex` used, such that we might get away with not adding `partialMatch`, but switch from using `String.matches` to indeed `Matcher.find`. Folks that want to old behaviour can then switch from `my:regex.*` to using `^my:regex.*$`. Any thoughts on that?

Jenson3210:
This crossed my mind as well. 
But wanted to go for the "safe" option as I do not feel comfortable making breaking changes on something which is not mine. 

That being said: 
If we would do the way you mention, we would still not have the (wanted/similar?) behaviour/option for non-regex, partialMatch combination. 
```java
    @Test
    void partialMatch() {
        rewriteRun(
          spec -> spec.recipe(new ChangePropertyValue("*", "replaced", "replaceme", null, true,null)),
          properties("""
            multiple=[replaceme:1][replaceme:2]
            multiple-prefixed=test[replaceme:1]test[replaceme:2]
            multiple-suffixed=[replaceme:1]test[replaceme:2]test
            multiple-both=test[replaceme:1]test[replaceme:2]test
            """, """
            multiple=[replaced:1][replaced:2]
            multiple-prefixed=test[replaced:1]test[replaced:2]
            multiple-suffixed=[replaced:1]test[replaced:2]test
            multiple-both=test[replaced:1]test[replaced:2]test
            """)
        );
    }
```

timtebeek:
Hmm; would that not be covered by using the regex option but escaping all special characters in a regex?

Jenson3210:
you mean to say something like: 
```yaml
  - org.openrewrite.yaml.ChangePropertyValue:
      propertyKey: "**.*"
      newValue: replaced
      oldValue: replaceme
      regex: "True"
```

Certainly possible, but it could yield issues towards the regex field not set/understanding. People might think it is not a regex -> not required to set.
Escaping special characters, looks to be overly complex for a find/replace. 
But you're the "owner", I can do the changes if you prefer this approach. ;)

Jenson3210:
@timtebeek made the necessary changes. Let me know if any other recipes pop into your mind that would need ^...$ surrounding.

timtebeek (approved):
Thanks again! Nice to be able to add functionality with a minimal amount of changes.

Jenson3210:
That's kuddo's to the original author more than to this one 😄

PR #3971, opened 2024-02-03
Submit dependency graph to GitHub after publish

timtebeek:
## What's changed?
Added a workflow that will submit a dependency graph to the GitHub API after publish.
As per 
- https://github.com/gradle/actions/blob/main/dependency-submission/README.md
- https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/exploring-the-dependencies-of-a-repository

## What's your motivation?
Want to get insight into dependents, to surface who all are using OpenRewrite on public projects.
- https://github.com/openrewrite/rewrite-docs/issues/131

Additionally this will allow us to explore dependencies and vulnerabilities from within GitHub on the Security and Insights tabs.

@natedanner or @sjungling, as you both frequently go through the vulnerability reports, which are also analyzed with


timtebeek:
Figured adding it here would have the most impact; beyond this I think it might only make sense on the OSS Maven/Gradle plugins, as they are separately depended upon, or at most rewrite-recipe-bom as a catch all there.

Because of the permission needed to submit to the API I added this to a separate workflow that's only triggered after a release, not on any pull request, or attached to any other workflow.

.github/workflows/dependency-submission.yml
```diff
@@ -0,0 +1,24 @@
...
+  dependency-submission:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+      - name: Generate and submit dependency graph
+        uses: gradle/actions/dependency-submission@v3
+        env:
+          # Exclude all dependencies that originate solely in the 'buildSrc' project
+          DEPENDENCY_GRAPH_EXCLUDE_PROJECTS: ':buildSrc'
```
timtebeek, line 22: We could remove these lines, or change them to exclude `:tools` for instance; kept it in as an example for discussion here.
```suggestion
```

sjungling (approved):

natedanner (approved):

timtebeek:
Ok that seems to have worked: https://github.com/openrewrite/rewrite/network/dependencies
Now wait to see if we similarly [see any dependents show up](https://github.com/openrewrite/rewrite/network/dependents); which might take a while for this new version.

timtebeek:
Also added to 
- https://github.com/openrewrite/rewrite-recipe-bom/pull/25
- https://github.com/openrewrite/rewrite-gradle-plugin/commit/49361f7afc2150c7ce7719738ef8f77d6e77efab

PR #3972, opened 2024-02-03
Remove duplicate plugin versions. 

### LINKED ISSUE #3932: RemoveRedundantDependencyVersions for Plugins

Like dependencies, plugins can also have redundant versions when compared to the `pluginManagement`. But, `RemoveRedundantDependencyVersions` only touches dependencies.

Can't think of any

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>org.sample</groupId>
    <artifactId>sample</artifactId>
    <version>1.0.0</version>
    <packaging>jar</packaging>
    <parent>
        <version>2.7.18</version>
        <relativePath/>
    </parent>
    <build>
        <plugins>
            <plugin>
                <groupId>pl.project13.maven</groupId>
                <artifactId>git-commit-id-plugin</artifactId>
                <version>4.9.10</version>
            </plugin>
        </plugins>
    </build>
</project>

```

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>org.sample</groupId>
    <artifactId>sample</artifactId>
    <version>1.0.0</version>
    <packaging>jar</packaging>
    <parent>
        <version>2.7.18</version>
        <relativePath/>
    </parent>
    <build>
        <plugins>
            <plugin>
                <groupId>pl.project13.maven</groupId>
                <artifactId>git-commit-id-plugin</artifactId>
            </plugin>
        </plugins>
    </build>
</project>

```

Implementation options could include:
- Add this to `RemoveRedundantDependencyVersions`
- Same as above, but with an arg to control the plugin-handling behavior
  - Choice about whether to default the behavior to on or off
- Create a new, separate `RemoveRedundantPluginVersions`

I'd vote for adding it to `RemoveRedundantDependencyVersions`, active by default, with a new optional arg to disable checking plugins.

I might; priority TBA. This might also be a good first issue for someone else.

ammachado:
## What's changed?
Fixes #3932

## What's your motivation?
Remove duplicate plugin versions from pom files updated by OpenRewrite

No

No

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -307,6 +307,17 @@ public class ResolvedPom {
         return null;
     }
 
+    @Nullable
+    public String getManagedPluginVersion(String groupId, String artifactId) {
+        for (Plugin p : pluginManagement) {
+            if (p.matches(groupId, artifactId)) {
+                return getValue(p.getVersion());
+            }
+        }
+
+        return null;
+    }
+
     public List<GroupArtifact> getManagedExclusions(String groupId, String artifactId, @Nullable String type, @Nullable String classifier) {
         for (ResolvedManagedDependency dm : dependencyManagement) {
             if (dm.matches(groupId, artifactId, type, classifier)) {
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -40,11 +40,11 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
     static final XPathMatcher DEPENDENCY_MATCHER = new XPathMatcher("/project/dependencies/dependency");
     static final XPathMatcher PLUGIN_DEPENDENCY_MATCHER = new XPathMatcher("/project/*/plugins/plugin/dependencies/dependency");
     static final XPathMatcher MANAGED_DEPENDENCY_MATCHER = new XPathMatcher("/project/dependencyManagement/dependencies/dependency");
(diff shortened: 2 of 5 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -307,6 +307,17 @@ public String getManagedVersion(String groupId, String artifactId, @Nullable Str
...
+    public String getManagedPluginVersion(String groupId, String artifactId) {
+        for (Plugin p : pluginManagement) {
+            if (p.matches(groupId, artifactId)) {
+                return getValue(p.getVersion());
+            }
+        }
+
+        return null;
+    }
+
```
timtebeek, line 320: If we could pull out this and the change in `tree/Plugin.java` to the visitor that would help make the review and merge easier. Changes in the tree package require a coordinated release and deployment, and it seems we only use this once for now, so then having it near where it's used is preferred. We can always pull it up to the tree package later if we use this more often.

nmck257:
I'd recommend updating the recipe description (and display name?) to reflect the new behavior, but other than that (and Tim's note), I like it!

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
timtebeek: There's some eager class loading involved that then requires a coordinate deployment. If we limit this change to not make those changes I don't have to involve others (yet).

timtebeek (approved):
Thanks a lot! Also for your patience as I work through your backlog. :)

PR #3975, opened 2024-02-05
Add support for adding JavaCompatibility statements where missing to …

krlittle:
…`UpdateJavaCompatibility` recipe

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
@@ -20,6 +20,7 @@ import lombok.Value;
 import org.openrewrite.*;
 import org.openrewrite.gradle.util.ChangeStringLiteral;
 import org.openrewrite.groovy.GroovyVisitor;
+import org.openrewrite.groovy.tree.G;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.MethodMatcher;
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -63,6 +64,21 @@ public class UpdateJavaCompatibility extends Recipe {
...
+            description = "Adds Source Compatibility if one is not found.",
+            required = false)
+    @Nullable
+    Boolean addSourceCompatibilityIfMissing;
+
+    @Option(displayName = "Add Target Compatibility if missing",
+            description = "Adds Target Compatibility if one is not found.",
+            required = false)
+    @Nullable
+    Boolean addTargetCompatibilityIfMissing;
```
shanman190, line 77: Given the option that allows picking which type to update -- both, source only, or target only -- this could probably be simplified to just a single `addIfMissing`.

shanman190 (commented):
Just a couple of notes here. 

I wonder if we detect the Gradle tool chains feature has been configured, if we should also avoid adding the compatibility if it's been requested to be added. The reason for that is that using both will result in Gradle reporting it as an error.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
krlittle: I'm fine with that.  This way was a little more flexible but there's probably not very many cases where you would want one option and not the other as well.

krlittle: Could we accomplish both simplifying this and maintaining flexibility by changing the `Boolean` to an `Enum` with values of `SourceOnly`, `TargetOnly`, `SourceAndTarget`?

Issue #3847, opened 2023-12-22
Add negation support to FindSourceFiles


shanman190:
Eg `**/!(CertainFileName)IT.java`, like HasSourcePath used to support.

timtebeek:
Fixed in 
- https://github.com/openrewrite/rewrite/pull/3980

PR #3980, opened 2024-02-06
Support negation in globmatcher in pathutils

Glanfaloth:
## What's changed?

matchesGlob in PathUtils now supports negation with !()

## What's your motivation?

I was writing a custom recipe where I wanted to apply the changes to source files whose filename ends with `*IT.java`, while excluding a few exceptions e.g. `CucumberLauncherIT.java`. However, when I used FindSourceFiles, I realized negation was not supported in the glob patterns.

@shanman190 @timtebeek 

## Any additional context

#3847 

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/PathUtilsTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/PathUtilsTest.java
@@ -111,6 +111,17 @@ class PathUtilsTest {
         assertThat(matchesGlob(path("test/bar"), "test/{foo,bar}")).isTrue();
     }
 
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite/issues/3847")
+    void negation() {
+        assertThat(matchesGlob(path("test.txt"), "test.!(txt)")).isFalse();
+        assertThat(matchesGlob(path("test.java"), "test.!(txt)")).isTrue();
+        assertThat(matchesGlob(path("test/foo"), "test/!(foo|bar)")).isFalse();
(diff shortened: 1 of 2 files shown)


rewrite-core/src/test/java/org/openrewrite/PathUtilsTest.java
```diff
@@ -111,6 +111,17 @@ void eitherOr() {
...
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite/issues/3847")
+    void negation() {
+        assertThat(matchesGlob(path("test.txt"), "test.!(txt)")).isFalse();
+        assertThat(matchesGlob(path("test.java"), "test.!(txt)")).isTrue();
+        assertThat(matchesGlob(path("test/foo"), "test/!(foo|bar)")).isFalse();
+        assertThat(matchesGlob(path("test/bar"), "test/!(foo|bar)")).isFalse();
+        assertThat(matchesGlob(path("test/quz"), "test/!(foo|bar)")).isTrue();
+        assertThat(matchesGlob(path("test/bar"), "test/!(f*|b*)")).isFalse();
+    }
```
shanman190, line 123: So I think that I'd probably lean toward what the JVM uses as a guide for group matches.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)

This would only essentially mean swapping out `(..)` for `{..}` instead. By using `{..}` this we should also be able to handle the positive multi group as well.

shanman190 (approved):

timtebeek:
Thanks a lot @Glanfaloth ; great to see support for negations restored.

PR #3984, opened 2024-02-07
Skip dependency resolution option for `MavenParser` skips parent resolution as well

jkschneider:
Improves the skipping of dependency resolution on `MavenParser` by:

* Allowing to skip parent resolution as well
* Moving the skip option to `MavenExecutionContextView` which is already supplied to the parsing phase.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -343,7 +343,9 @@ public class ResolvedPom {
         MavenPomDownloader downloader;
 
         public ResolvedPom resolve() throws MavenDownloadingException {
-            resolveParentsRecursively(requested);
+            if (!MavenExecutionContextView.view(ctx).getSkipDependencyResolution()) {
+                resolveParentsRecursively(requested);
+            }
             return ResolvedPom.this;
         }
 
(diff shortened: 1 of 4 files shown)


timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -343,7 +343,9 @@ class Resolver {
         MavenPomDownloader downloader;
 
         public ResolvedPom resolve() throws MavenDownloadingException {
-            resolveParentsRecursively(requested);
+            if (!MavenExecutionContextView.view(ctx).getSkipDependencyResolution()) {
+                resolveParentsRecursively(requested);
```
timtebeek, line 347: To detect Java versions in the CLI it'd be helpful to still resolve parent poms without resolving dependencies. That aligns with how we've seen customers use archetypes to set the Java version.  Would it make sense to make resolving parent poms a separate option?

jkschneider: I'm not liking the options if that's the case. Almost would rather a general purpose dependency resolution interceptor pattern than an array of very specific options.

jkschneider:
Reassessing whether this makes sense.

knutwannheden:
> Reassessing whether this makes sense.

My reasoning for using the parser builder was that we during parsing typically reuse the same execution context and that the parsers return a stream, where the sources get lazily parsed. So the client would theoretically have to make sure the execution context stores the correct setting at the time it "pulls" the corresponding source from the stream.

This is a bit theoretical, as we won't be using this specific option anywhere where this would be a problem. I was just a bit worried that it might set a bad precedence for other parser options, but this was possibly unjustified.

knutwannheden:
Coincidentally there was this discussion regarding the encoding to be used by the parser just earlier today: https://github.com/openrewrite/rewrite-maven-plugin/pull/735#discussion_r1480907685

PR #3985, opened 2024-02-08
Allow non-period terminated Recipe descriptions

JLLeitschuh:
In the cases where the description ends with a list of resources
the recipe should allow that the list not end with a period.

## What's changed?

Modifies `RecipeTest#validateRecipeNameAndDescription` to support descriptions
that are not terminated in a `.` under certain conditions.

## What's your motivation?

I want to write a recipe that has a description that ends with a list of links.
The formatting is very weird if this list is required to end in a `.`

Are there other conditions that I should be testing for here?

Recipe authors could manually override `validateRecipeNameAndDescription` to change the behaviour of the validation, but this seems like a better general solution.

### Checklist


timtebeek:
Fair enough, thanks for adding & testing support!

PR #3987, opened 2024-02-08
Add other possible scopes for maven dependencies

der-eismann:
## What's changed?
Added remaining possible scopes for Maven dependencies to valid values.

## What's your motivation?
Maven has 6 possible scopes for dependencies, however only 3 of them were allowed as valid ones.

No

## Any additional context
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
@@ -78,7 +78,7 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
             description = "A scope to use when it is not what can be inferred from usage. Most of the time this will be left empty, but " +
                           "is used when adding a runtime, provided, or import dependency.",
             example = "runtime",
-            valid = {"import", "runtime", "provided"},
+            valid = {"compile", "import", "provided", "runtime", "system", "test"},
             required = false)
     @Nullable
     String scope;


timtebeek:
Hi @der-eismann ; Thanks for opening a pull request; Did you run into any cases where this was needed? Especially looking at

https://github.com/openrewrite/rewrite/blob/79327ebe77a4480bdfc4a40a1e9bca254ee870e0/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L77-L80

Whenever we add a dependency we typically expect `onlyIfUsing` to be set, such that the dependency is only ever added when a particular type is used. When that type is used, we also use the corresponding scope of where the type is being used.

https://github.com/openrewrite/rewrite/blob/cce995320d5fcd5c33ea6000fc18a5fb3d98fece/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L186-L188

While it might be surprising not to see all scope options configurable, it's worked out well enough in the past that we've kept it this way. Not having explicit configuration options for compile & test also helps prevent recipes from adding that option when they really should be relying on `onlyIfUsing`.

It might make sense to add these values if we proceed with this PR:
- https://github.com/openrewrite/rewrite/pull/4006

But I'm mostly hoping to hear what prompted you to add these values; did `onlyIfUsing` not work well enough for you?

der-eismann:
Hi @timtebeek, thanks for the detailed reply! To be honest I opened this PR before I realized the extent of what OpenRewrite is doing - I was expecting it to edit the pom.xml to my likings instead of containing so much logic, so that PR probably won't work correctly anyway. We are trying to utilize OpenRewrite to apply common rules and migrations on our ~80 projects, but we are struggling with some parts of how it works, but that would go beyond the scope of this issue, I should probably open a discussion instead.

In our case we had a library that was previously `com.example.messaging`, now split into `com.example.messaging.library` and `com.example.messaging.testing`. My goal was to remove the old dependency and add the new ones with their respective scope. `onlyIfUsing` is not feasible here, as the imports are overlapping in some places. Not setting the `test` scope would mean unnecessarily putting the testing lib in the resulting *.jar.

timtebeek:
All good; welcome to join [our Slack](https://join.slack.com/t/rewriteoss/shared_invite/zt-nj42n3ea-b~62rIHzb3Vo0E1APKCXEA) as well for any such questions; indeed we're not just making any low level change, but instead aiming for a minimal change where we can.

In your case it sounds like you might want to combine a few declarative recipes to for instance move some types to a different package (which would also update any imports) before `onlyIfUsing` can be used to selectively add the dependency to the right scope. Welcome to discuss more in Slack if that helps you.

We can either keep this open as a draft in case it's needed after #4006 (cc @ammachado ), or close this and add it there.

Thanks again for chiming in! Hope we're able to help you out there. :)

PR #3988, opened 2024-02-08
Add test for updating parent gradle.properties when dependency is in …

ajohnsonz:
…subprojects block

## What's changed?
Trying to get parents versions in `gradle.properties` to correctly update when the dependency is applied to subprojects

@shanman190 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -127,6 +127,7 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
     @Value
     public static class DependencyVersionState {
         Map<String, GroupArtifact> versionPropNameToGA = new HashMap<>();
+        // The value is either a String representing the resolved version or a MavenDownloadingException representing an error during resolution.
         Map<GroupArtifact, Object> gaToNewVersion = new HashMap<>();
     }
 
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -28,6 +28,7 @@ import org.openrewrite.groovy.GroovyIsoVisitor;
 import org.openrewrite.groovy.GroovyVisitor;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.tree.Expression;
@@ -44,10 +45,12 @@ import org.openrewrite.properties.PropertiesVisitor;
 import org.openrewrite.properties.tree.Properties;
 import org.openrewrite.semver.DependencyMatcher;
 import org.openrewrite.semver.Semver;
+import org.openrewrite.semver.VersionComparator;
 
 import java.util.*;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 
 @Value
@@ -124,7 +127,7 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
     @Value
     public static class DependencyVersionState {
         Map<String, GroupArtifact> versionPropNameToGA = new HashMap<>();
-        Map<GroupArtifact, String> gaToNewVersion = new HashMap<>();
+        Map<GroupArtifact, Object> gaToNewVersion = new HashMap<>();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -208,18 +211,15 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                         if(acc.gaToNewVersion.containsKey(ga)) {
                             return m;
                         }
-                        String oldVersion = findCurrentResolvedVersion(groupId, artifactId, m.getSimpleName(), gradleProject);
```
shanman190, line 211: This in particular was completely preventing updating the dependencies contained within the subproject because they don't exist in the current project's `GradleProject` marker.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -244,20 +244,15 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                                 if(acc.gaToNewVersion.containsKey(ga)) {
                                     continue;
                                 }
-                                String oldVersion = findCurrentResolvedVersion(dep.getGroupId(), dep.getArtifactId(), m.getSimpleName(), gradleProject);
```
shanman190, line 247: Same here.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
shanman190, line 211: This in particular was completely preventing updating the dependencies contained within the subproject because they don't exist in the current project's `GradleProject` marker.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -208,18 +211,15 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
...
-                        String oldVersion = findCurrentResolvedVersion(groupId, artifactId, m.getSimpleName(), gradleProject);
-                        if(oldVersion == null) {
-                            return m;
-                        }
                         try {
-                            String newVersion = findNewerProjectDependencyVersion(groupId, artifactId, oldVersion, gradleProject, ctx);
-                            if(oldVersion.equals(newVersion)) {
-                                return m;
-                            }
+                            String resolvedVersion = "classpath".equals(m.getSimpleName()) ?
```
shanman190, line 215: Previously, we weren't taking into account properties being used in the `buildscript` block.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddPluginVisitor.java
```diff
@@ -91,7 +90,7 @@ private static Optional<String> findNewerVersion(String groupId, String artifact
     }
 
     private static MavenMetadata downloadMetadata(String groupId, String artifactId, List<MavenRepository> repositories, ExecutionContext ctx) throws MavenDownloadingException {
-        return new MavenPomDownloader(emptyMap(), ctx, null, null)
+        return new MavenPomDownloader(ctx)
```
shanman190, line 93: Like in other commits, let's not use the default repositories on the Gradle side as that's not accurate.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -124,7 +127,7 @@ public Validated<Object> validate() {
     @Value
     public static class DependencyVersionState {
         Map<String, GroupArtifact> versionPropNameToGA = new HashMap<>();
-        Map<GroupArtifact, String> gaToNewVersion = new HashMap<>();
+        Map<GroupArtifact, Object> gaToNewVersion = new HashMap<>();
```
timtebeek, line 130: Perhaps good to document that this can contain a `GroupArtifact`, `String` and `Exception`, and how each are handled. Could otherwise be a bit puzzling to readers as it's not immediately apparent, even if it's necessary here.

timtebeek (approved):
Thanks a lot both! Perhaps good to add some comments and where pointed out below, and rename the PR to reflect the change rather than the test we started out with. Otherwise good to merge.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
shanman190: So this is keyed by `GroupArtifact`, but they value is either the `String` of the resolved version or the `MavenDownloadingException` representing an error in resolving the version to then be used to markup the dependency declaration later.

shanman190:
```suggestion
        // The value is either a String representing the resolved version or a MavenDownloadingException representing an error during resolution.
        Map<GroupArtifact, Object> gaToNewVersion = new HashMap<>();
```

Seems good to you, @timtebeek?

Issue #3982, opened 2024-02-07
UpdateMavenWrapper inconsistent when the wrapper is in a nested directory


gsmet:
## What version of OpenRewrite are you using?

I am using the latest version of the OpenRewrite Maven plugin: 5.22.0.

## How are you running OpenRewrite?

Using the Maven plugin.

## What is the smallest, simplest way to reproduce the problem?

```
git clone git@github.com:hibernate/hibernate-demos.git
cd hibernate-demos
git switch -d f4dd176ea3a128f5c308f4fef86b9e5cdecb1f36
cd hibernate-search/hsearch-feature-examples/base

cat <<\EOF > rewrite.yml
---
type: specs.openrewrite.org/v1beta/recipe
name: io.quarkus.updates.core.quarkus37.UpdateMavenWrapper
recipeList:
  - org.openrewrite.maven.UpdateMavenWrapper:
      distributionVersion: 3.9.x
      addIfMissing: false
EOF

mvn -U org.openrewrite.maven:rewrite-maven-plugin:run  -Drewrite.activeRecipes=io.quarkus.updates.core.quarkus37.UpdateMavenWrapper
```

## What did you expect to see?

The Maven wrapper properly updated in this subdirectory.

## What did you see instead?

The property file is updated in place but the `maven-wrapper.jar` file is moved to/updated in a parent directory (`../../../.mvn/`):

```
$ git status

HEAD detached at f4dd176
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    .mvn/wrapper/maven-wrapper.jar
	modified:   .mvn/wrapper/maven-wrapper.properties

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	../../../.mvn/ <---- maven-wrapper.jar is here
	rewrite.yml

no changes added to commit (use "git add" and/or "git commit -a")
```

## What is the full stack trace of any errors you encountered?

This is the output, we can see that the files have been moved:

```
[INFO] Running recipe(s)...
[WARNING] File has been moved from hibernate-search/hsearch-feature-examples/base/.mvn/wrapper/maven-wrapper.jar to .mvn/wrapper/maven-wrapper.jar by:
[WARNING]     io.quarkus.updates.core.quarkus37.UpdateMavenWrapper
[...]

PR #3989, opened 2024-02-08
Assorted UpdateMavenWrapper fixes

gsmet:
## What's changed?

- Make sure the files are updated in place when updated
- Do not escape colons in the URLs injected in the .properties file. When using `mvn wrapper:wrapper` you don't get colons and they are actually causing issues at least for the `maven-wrapper.jar` as this property is not loaded from Java
- Avoid adding the `maven-wrapper.jar` checksum by default as it's causing issues on Windows due to https://issues.apache.org/jira/browse/MWRAPPER-103 -> this is probably something we need to discussed

## What's your motivation?

- https://github.com/openrewrite/rewrite/issues/3982
- Discussions in https://github.com/quarkusio/quarkus/issues/38569

I didn't have the time to add tests for the first commit yet. I'll have a look at it in the next days as time permits if nobody beats me to it.

@timtebeek @shanman190 

### Checklist

^ Is there a way to apply the formatter with a command line? I'm all for formatting rules but not using Intellij atm. I suppose I could try to use it if there are no other option.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenWrapper.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenWrapper.java
@@ -107,6 +107,21 @@ public class UpdateMavenWrapper extends ScanningRecipe<UpdateMavenWrapper.MavenW
     @Nullable
     final Boolean addIfMissing;
 
+    /**
+     * Due to a current limitation of the Maven wrapper, checking the maven-wrapper.jar checksum
+     * can be problematic on Windows.
+     * Since one of the use case of the maven-wrapper.jar is to be portable across platforms, it
+     * seems safer to not enforce the maven-wrapper.jar checksum check by default.
+     *
+     * See https://issues.apache.org/jira/browse/MWRAPPER-103
+     */
+    @Getter
+    @Option(displayName = "Enforce checksum of maven-wrapper.jar",
+            description = "Enforce the check of the checksum of the maven-wrapper.jar. Defaults to `false` for Windows compatibility.",
+            required = false)
+    @Nullable
+    final Boolean checkMavenWrapperJarChecksum;
+
     @Override
     public String getDisplayName() {
         return "Update Maven wrapper";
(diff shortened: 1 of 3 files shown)


timtebeek:
Thanks a lot for the in depth look for these various fixes. Added a test for the subdir handling to update the same files just now. I'm not sure what the backstory was on those escaped colons for properties, but that seems to have been added in
- https://github.com/openrewrite/rewrite/pull/3392

@shanman190 do you recall what was behind that, and if we need it for any particular case still? Otherwise good to merge if you ask me.

shanman190:
So that's actually because I copied from the `UpdateGradleWrapper` recipe that I wrote. I kinda made the assumption that Maven would have scavenged at least a fair bit of the code from Gradle, but it appears in this particular case they didn't.

On the Gradle side, they read and write the properties using `java.util.Properties`. When a `store` operation is executed, it escapes the `:` character, so during a `gradle wrapper` or `gradlew wrapper` cycle, the end resulting properties file will be escaped. Now the one aspect that Gradle then goes an manipulates is by writing the output to memory, then removing the header, and finally writing that as the result to disk. Maven on the other hand appears to just write the properties using a `BufferedWriter` instead rather than the `java.util.Properties` interface. 

Now with that said, the `wrapperUrl` is parsed out of the properties by the shell scripts, then used to directly download the wrapper.jar if it's missing. Since this is used in a shell script, being escaped would likely result in a few issues. Downloading the wrapper.jar file in the script is not something that is supported on the Gradle side, so that's why this issue never presents over there.

shanman190:
I went ahead and reworded/renamed a few things with respect to the new option.

I also wanted to point out that this particular `Get-FileHash` isn't a Maven Wrapper or Windows issue in itself. `Get-FileHash` is a function that ships with the PowerShell for every Windows OS version. However, when PowerShell Core installs badly and then is used, such as is the case within GitHub runners at present it appears, then the core PSModulePath can effectively be hidden from view resulting in those core cmdlets not being available for use. So the issue isn't technically a Windows (general) incompatibility, but nonetheless it occurred in the ecosystem where folks work from. As such, I removed that note, but did add the issue link in the description markdown as a possible example implication of what enabling that feature can result in occurring.

shanman190 (approved):
Thanks for the work here to fix this issue, @gsmet!

gsmet:
Thanks to both of you!

PR #3992, opened 2024-02-11
fix/add managed dep with properties

ammachado:
- Failing test
- AddManagedDependency fix

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -208,6 +208,9 @@ public class MavenPomDownloader {
     }
 
     public MavenMetadata downloadMetadata(GroupArtifact groupArtifact, @Nullable ResolvedPom containingPom, List<MavenRepository> repositories) throws MavenDownloadingException {
+        if (containingPom != null) {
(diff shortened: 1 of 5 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -208,6 +208,9 @@ private Pom getParentWithinProject(Pom projectPom, Map<Path, Pom> projectPoms) {
     }
 
     public MavenMetadata downloadMetadata(GroupArtifact groupArtifact, @Nullable ResolvedPom containingPom, List<MavenRepository> repositories) throws MavenDownloadingException {
+        if (containingPom != null) {
+            groupArtifact = containingPom.getValues(groupArtifact);
+        }
```
timtebeek, line 213: I'm wondering if this block adds anything when the call below these lines to `downloadMetadata(GroupArtifactVersion gav, @Nullable ResolvedPom containingPom, List<MavenRepository> repositories) throws MavenDownloadingException` appears to do the same but for the GAV. Is there a reason to keep this in here and possibly get the values twice?
```suggestion
```

timtebeek (approved):
Nice addition, thanks! Like how there's only minimal API changes to make this possible.

One small question but then we're good to merge I think.

timtebeek (approved):

PR #3993, opened 2024-02-12
Allow usage of additional file extensions per parser.

ammachado:
## What's changed?
Allow parsers to parse additional file extensions.

## What's your motivation?
Parse Karaf's configuration files as properties.

No

No

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/tree/ParsingExecutionContextView.java
+++ b/rewrite-core/src/main/java/org/openrewrite/tree/ParsingExecutionContextView.java
@@ -17,12 +17,19 @@ package org.openrewrite.tree;
 
 import org.openrewrite.DelegatingExecutionContext;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Parser;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 
 import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 public class ParsingExecutionContextView extends DelegatingExecutionContext {
     private static final String PARSING_LISTENER = "org.openrewrite.core.parsingListener";
+    private static final String ADDITIONAL_EXTENSIONS = "%s.additionalExtensions";
 
     private static final String CHARSET = "org.openrewrite.parser.charset";
 
@@ -54,5 +61,30 @@ public class ParsingExecutionContextView extends DelegatingExecutionContext {
     @Nullable
     public Charset getCharset() {
         return getMessage(CHARSET);
+    }
+
+    public ParsingExecutionContextView addAdditionalExtension(Class<? extends Parser> parserClass, String extension) {
+        putMessageInSet(String.format(ADDITIONAL_EXTENSIONS, parserClass.getName()), extension);
+        return this;
+    }
+
+    public Set<String> getAdditionalExtensions(Class<? extends Parser> parserClass) {
+        String key = String.format(ADDITIONAL_EXTENSIONS, parserClass.getName());
+
+        Set<String> additionalExtensions = getMessage(key, Collections.emptySet());
+        if (additionalExtensions == Collections.EMPTY_SET) {
+            String property = System.getProperty(key, null);
+            if (StringUtils.isNotEmpty(property)) {
+                additionalExtensions = new HashSet<>(Arrays.asList(property.split(",")));
+                putMessage(key, additionalExtensions);
+            }
+        }
+
+        return additionalExtensions;
+    }
+
+    public ParsingExecutionContextView setAdditionalExtensions(Class<? extends Parser> parserClass, @Nullable Set<String> additionalExtensions) {
+        putMessage(String.format(ADDITIONAL_EXTENSIONS, parserClass.getName()), additionalExtensions);
+        return this;
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/Parser.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Parser.java
@@ -101,13 +101,38 @@ public interface Parser {
 
     boolean accept(Path path);
 
+    default boolean accept(Path path, ExecutionContext ctx) {
+        if (accept(path)) {
+            return true;
+        }
+        return ParsingExecutionContextView.view(ctx).getAdditionalExtensions(this.getClass())
+                .contains(PathUtils.getExtension(path));
+    }
+
+    /**
+     * @deprecated Use {@link #accept(Input, ExecutionContext)} instead.
+     */
+    @Deprecated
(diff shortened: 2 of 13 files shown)


knutwannheden:
I like the idea! But I wonder if the execution context shouldn't instead allow registering alternative implementations for `accept()`, so that file extensions can also be removed (let's say someone doesn't want to parse `Jenkinsfile` files anymore). This would require some further refactoring, so that an alternative implementation can still optionally delegate to the default implementation. Thoughts on this?

ammachado:
Perhaps if we just add a file exclusion pattern, similar to the additionalExtensions implemented here, would work? ... `accept(..)` would check for the default extensions, then inclusions and exclusions.

rewrite-core/src/main/java/org/openrewrite/tree/ParsingExecutionContextView.java
```diff
@@ -96,16 +95,15 @@ public ParsingExecutionContextView setExcludePattern(Class<? extends Parser> par
     }
 
     public <T extends Parser> boolean accepts(T parser, Path path) {
-        for (String pattern : getExcludePattern(parser.getClass())) {
+        for (String pattern : getExcludePattern(parser.getParserClass())) {
```
knutwannheden, line 98: This new `getParserClass()` is an unfortunate implementation detail that surfaces in the API. I wonder if we can find a cleaner solution. The `Parser.Builder#getDslName()` looks like it would have been nicer, but unfortunately we don't have access to that.

rewrite-core/src/main/java/org/openrewrite/tree/ParsingExecutionContextView.java
```diff
@@ -15,14 +15,25 @@
...
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 public class ParsingExecutionContextView extends DelegatingExecutionContext {
     private static final String PARSING_LISTENER = "org.openrewrite.core.parsingListener";
+    private static final String INCLUDE_PATTERNS = "%s.includePatterns";
+    private static final String EXCLUDE_PATTERNS = "%s.excludePatterns";
```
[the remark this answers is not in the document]
knutwannheden: I think I now understand that you did it this way so that the `Parser#accept(Path)` can remain as is and that the include and exclude patterns can be used to supply a "delta". Depending on how far we want to take this, that may indeed be a good compromise.

rewrite-core/src/main/java/org/openrewrite/tree/ParsingExecutionContextView.java
ammachado: I'm providing a better implementation.

sambsnyd:
It isn't clear to me that we need to make this extensible. 
We're happy to add to the list of extensions for each parser. Do you have a file extension you'd like to accrue to one of the parsers by default? Do you have a scenario where hard-coding is insufficient or otherwise impossible?

ammachado:
This PR adds the option to include/exclude file patterns per parser.  In my original use case, I have properties files that have `.cfg` extension instead of `.properties`, and I want to use the existing property recipes to refactor them.

jimschubert:
I also described my use case in #4062. I have a yaml file without extension. To parse this as YAML required me to not only subclass YamlParser and a few other types, but I had to pass the file pattern as plain text matchers to the Maven plugin. That functionally has been deprecated, so it's not clear to me how I'd go about targeting my desired file mappings without this PR.

sambsnyd:
I would rather add an entry to the hardcoded list than add an entire extension mechanism.
If all you need is to parse Karaf's configuration files as properties we can add it to the list of extensions that parser accepts by default.

jimschubert:
Hard coding my company's internal tooling names into OpenRewrite isn't an acceptable solution.

timtebeek:
@jimschubert Please note the `plaintextMasks` option is not deprecated; we only deprecated a duplicate option without a qualifying prefix. 

To offer an alternative that's already possible using existing mechanisms, one could try a recipe that runs first to convert any PlainText parsed files into Yaml with a recipe; this is something that's come up on conversation before internally. 

<details>
<summary>Show recipe that converts a plain text parsed file into Yaml</summary>

Note that I haven't checked the below code; just meant to illustrate the general approach.
```java
package org.openrewrite.yaml;

import lombok.Value;
import org.openrewrite.*;
import org.openrewrite.text.PlainText;
import org.openrewrite.text.PlainTextVisitor;

import java.util.Optional;

@Value
public class AsYaml extends Recipe {

    @Option(displayName = "File pattern",
            description = "Ant-style path pattern matching source files to parse as YAML.",
            example = "com/myapp/**/*.acme")
    String filePattern;

    @Override
        return "Parse as YAML";
    }

    @Override
        return "Parse an existing PlainText file as YAML. Useful for when you have a file that is not recognized as YAML, but you know it is.";
    }

    @Override
        PlainTextVisitor<ExecutionContext> plainTextVisitor = new PlainTextVisitor<ExecutionContext>() {
            @Override
            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                if (tree instanceof PlainText) {
                    PlainText pt = (PlainText) tree;
                    Optional<SourceFile> first = YamlParser.builder().build()
                            .parse(tree.print(getCursor()))
                            .findFirst();
                    if (first.isPresent()) {
                        SourceFile sourceFile = first.get();
                        if (pt.getCharset() != null) {
                            sourceFile = sourceFile.withCharset(pt.getCharset());
                        }
                        return sourceFile
                                .withCharsetBomMarked(pt.isCharsetBomMarked())
                                .withChecksum(pt.getChecksum())
                                .withFileAttributes(pt.getFileAttributes())
                                .withSourcePath(pt.getSourcePath())
                                .withMarkers(pt.getMarkers())
                                .withId(pt.getId());
                    }
                }
                return tree;
            }
        };
        return Preconditions.check(new FindSourceFiles(filePattern), plainTextVisitor);
    }
}
```
</details>

If the above recipe approach is workable then that might then remove the need for an extension mechanism, and the necessary configuration handling downstream to do that well.

jimschubert:
Thanks @timtebeek for the clarification on the depreciation and the example code! The depreciation message didn't make the clarification you made above, and although I checked the code, I must have missed that a prefixed option was still available. 

I'll give your example recipe a try. This is effectively what I'm doing already, as I found that I had to first do a search or other recipe on the file to convert. However I had to put a hack around the parser due to hard-coded extensions: https://github.com/openrewrite/rewrite/blob/98b720b802ea9d4f3d5b6c3e0ce532520c3e311f/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java#L436. I can't recall what was calling the accept method.

I'll return to work around the beginning of August and respond to the issue if my experience with the YAML files has changed after incorporating the above.

PR #3995, opened 2024-02-13
Add repository results standard data table

pstreef:
## What's changed?
A new standard data table is added (will be created for all recipes)

## What's your motivation?
The `SourcesFileResults` data table is too detailed on the one hand but lacks precision on the other. This data-table makes it possible to uniquely identify a recipe running on a repository and retrieve the aggregated occurrences and estimated time savings.

Naming, I had my doubts about the data-table name other canididates are:
`RecipeResults`
`RecipeRepositoryResults`

@jkschneider 

An alternative would be to fix the `SourcesFileResults` data table to have the options (or a unique way to identify which results belong to which recipe + options in a list of recipes)

### Checklist


rewrite-core/src/main/java/org/openrewrite/table/RepositoryResults.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public void flush(ExecutionContext ctx) {
+        maxCycle = ctx.getCycle(); //allow to write to this data table after last cycle
+        for (Map.Entry<RecipeKey, Counter> entry : counters.entrySet()) {
+            insertRow(ctx, new Row(entry.getKey().getRootName(), entry.getKey().getRecipeName(), entry.getKey().getOptions(), entry.getValue().estimatedTimeSavings, entry.getValue().occurrences, entry.getValue().filesChanged));
+        }
+    }
+
+    public static String serializeOption(String name, Object value) {
```
pstreef, line 97: doubting between this more readable serialization or json (makes matching a lot easier)

rewrite-core/src/main/java/org/openrewrite/table/RepositoryResults.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright 2022 the original author or authors.
```
knutwannheden, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

rewrite-core/src/main/java/org/openrewrite/table/RepositoryResults.java
knutwannheden: Wouldn't JSON basically be the same? Now you would get something like:
```
[foo="false", bar="1", baz="null", qux=`Foo"bar"`]
```
with JSON:
```json
{"foo":false, "bar":1, "baz":null, "qux":"Foo\"bar\""}
```
Of these I think JSON looks better and it also preserves basic data types and can again be deserialized as JSON.

Of course there will be some additional escaping when serialized as CSV. But that applies to both formats.

PR #3998, opened 2024-02-13
ChangeDependencyGroupIdAndArtifactId - fixed issue described by `managedToUnmanagedExternalizedDepMgmt` test…

nmck257:
…; previously, the changed dep would erroneously have no version (and one or more "No version provided" markers) after recipe execution

## What's your motivation?

The Spring Cloud 2022.0.x upgrade recipe is currently susceptible to the defect fixed by this PR (hence the dep values used in the new test cases)
https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/spring-cloud-2022.yml

@timtebeek 

### Checklist


timtebeek (approved):
Thanks for the fix @nmck257 ; I like how it also simplifies the logic here.

PR #4001, opened 2024-02-14
fixing defect when using `latest.patch` version syntax with ChangeDep…

nmck257:
…endencyGroupIdAndArtifactId and ChangeManagedDependencyGroupIdAndArtifactId

### Checklist


timtebeek (approved):
Thanks a lot! Not an easy one to spot either, so much appreciated.

PR #4002, opened 2024-02-14
ChangeParentPom now allows changing groupId/artifactId in the case wh…

nmck257:
…ere old and new version match, without requiring the `allowVersionDowngrades` option (also added extra logic to preserve idempotency for `relativePath` changes)

## What's your motivation?

The new test case was previously failing, with the recipe making no changes. Adding the `allowVersionDowngrades` option as `true` would allow the expected changes, but it's not really a "downgrade", so that option doesn't feel like it should be required in that case.

...And once I made that change, one of the other tests started adding duplicative `<relativePath />` tags on subsequent cycles, hence the adjusted logic in that section.

The new `return` logic in `findAcceptableVersion` is a bit cumbersome. I considered instead adjusting `VersionComparator::upgrade` to allow returning the current version -- it would certainly simplify this control flow -- but that seemed like a semantic change with implications for all other usages, and even adding an optional arg for that behavior felt excessive.

### Checklist


timtebeek (approved):
Thanks for another detailed fix!

PR #4004, opened 2024-02-14
Allow declarative preconditions for scanning recipes

knutwannheden:
When a declarative recipe contains both preconditions and scanning recipes in its list, the scanning recipe's scanner is never executed because the bellwether precondition doesn't get initialized until the execute phase. Similarly, the generate phase won't be executed either, because `BellwetherDecoratedScanningRecipe` doesn't cater for this.

This PR changes this by always running the scanning and generate phase of any contained scanning recipes. This at least allows scanning recipes to function properly.

More fine grained control of how preconditions should apply to the scanning phase are the subject of #4005.


knutwannheden:
I could not find any declarative recipes with preconditions which also directly (or indirectly) contain any scanning recipes. I suppose this is why this problem has never shown up until now.

PR #4006, opened 2024-02-15
Add dependency now works with empty maven projects

ammachado:
## What's changed?
`AddDependency` now works with empty maven projects

## What's your motivation?
Refactor Spring Boot Starters

No

No

N/A

## Any additional context
N/A

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
@@ -197,7 +197,7 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for getting this started @ammachado ; I went over the changes but have a slight concern about what this change would mean when running against projects at scale, and would like to get your perspective on this before we merge.

Consider that we have a platform that allows folks to run migration recipes at scale, against an entire organization. Up to now folks are able to run for instance a Micronaut, Quarkus, or Spring migration against the entire organization, and not see any changes in projects without any Java classes, even if `onlyIfUsing` was empty (which is hopefully rare, and questionable).

With this change folks would start to see dependencies added where they were not added before in in-house starters or parent pom.xml files. I'm afraid that might surprise them in a negative way, or would need some compensating actions or documentation. What's your thought on this?

I'm wondering if we for instance might want to add an option that explicitly enables adding dependencies to projects without any Java classes, as it's presumably rare. But at the same time: not using `onlyIfUsing` should be rare as well, so I'm a bit on the fence here.

PR #4009, opened 2024-02-16
Enhance autoformat logic for Groovy files

shanman190:
## What's changed?
Bringing back some auto formatting rules to the `G` type. This focuses on tabs VS spaces, normalizing EOL characters, blank lines, and removing trailing whitespace.

## What's your motivation?
Historically, Groovy auto formatting was a bear and would pretty often result in files becoming invalid. As a result, it was disabled fully in the `org.openrewrite.java.format.AutoFormatVisitor`, so that the Java auto formatting rules would never be applied to any non-`J` type.

While re-introducing auto format support, I found two bugs. One of which is found fixed within this PR, but the other has been noted with a pair of disabled tests.

* If a visitor used with `withClasses` or `withImports` on an instance of `G`, this would result in the `G.CompilationUnit` being recreated due to a bug in the logic. Specifically, getting the classes or imports will always return a new list which means that they will never have instance identity equality. So instead, we should verify that the elements themselves are unchanged, then take the appropriate actions.
* GroovyParser appears to be unable to parse the implicit-public constructor. This additionally may extend into constructors with default arguments, so there is a second test case to cover that specific case.

N/A

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
@@ -15,10 +15,7 @@
  */
 package org.openrewrite.groovy;
 
-import org.openrewrite.Cursor;
 import org.openrewrite.SourceFile;
-import org.openrewrite.groovy.format.MinimumViableSpacingVisitor;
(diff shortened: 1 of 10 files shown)


shanman190:
I'll figure out what is wrong with auto formatting the rewrite-gradle tests tomorrow and get that fixed.

PR #4013, opened 2024-02-16
Fixed regex in XPathMatcher to support dots

kislam01:
## What's changed?
This PR fixes the Regex Pattern in XPathMatcher to support the 'dot' character in xPaths. 

## What's your motivation?
I was trying to use the xPathMatcher to find a match for a groupId in a pom file (e.g. `dependency[groupId='com.fasterxml.jackson']`). However, I noticed that the Regex pattern defined [here](https://github.com/kislam01/rewrite/blob/main/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java#L43) does not support the 'dot' character, which is why it wasn't able to find a match for the groupId. This PR aims to resolve this issue.

### Checklist


timtebeek (approved):
Thanks for the fix! Appreciate that you immediately dove in to see this resolved. :)

PR #4014, opened 2024-02-16
FindDependency now supports an optional version filter, and, created …

nmck257:
…a corresponding FindManagedDependency recipe

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindDependency.java
@@ -53,7 +53,7 @@ public class FindDependency extends Recipe {
 
     @Option(displayName = "Version pattern",
             description = "Allows version selection to be extended beyond the original Node Semver semantics. So for example," +
-                    "Setting 'version' to \"25-29\" can be paired with a metadata pattern of \"-jre\" to select Guava 29.0-jre",
+                          "Setting 'version' to \"25-29\" can be paired with a metadata pattern of \"-jre\" to select Guava 29.0-jre",
             example = "-jre",
             required = false)
     @Nullable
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/FindDependency.java
```diff
@@ -73,11 +100,31 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+            // shouldn't happen, but if it does, fail the condition
+            return false;
+        }
+        String actualVersion = resolvedDependency.getVersion();
+        Validated<VersionComparator> validate = Semver.validate(desiredVersion, versionPattern);
+        if (validate.isInvalid()) {
+            return false;
+        }
+        assert(validate.getValue() != null);
+        return validate.getValue().isValid(actualVersion, actualVersion);
```
timtebeek, line 128: Perhaps it's just my unfamiliarity, but calling `.isValid(actualVersion, actualVersion)` with the actual version twice seemed odd to me at first, although it might be ok from seeing how `Semver.validate` returns one of many ranges. Asking here just to be sure.

nmck257: The first arg of `VersionComparator::isValid` is for the current version, which is relevant for the `LatestPatch` implementation. In this case, the version is not changing, so the `actualVersion` is an appropriate value for both the current version and the version-under-test.

...but yes, it's a surprising method signature. I made another change recently in another class which had a long-standing (small) bug with its use of the same method.

Todo edit with link

timtebeek: Thanks! I'll have a look there as well; Then we're good to merge if you think so too.

PR #4015, opened 2024-02-17
Correctly set `Cursor` in `Repeat`

JLLeitschuh:
The `Cursor` wasn't being set in `Repeat` causing issues when
used for non `SourceFile` trees.

## What's changed?

Adds a call to `setCursor` in the loop for `Repeat`.
Also adds internal validation that `Repeat` is being called with the correct
`Cursor` set.

## What's your motivation?

When using `Recipe` visitors on sub-trees, and those visitors called `autoFormat` they were failing
when the `JavaSourceFile` couldn't be located in the `Cursor` hierarchy.

### Checklist


JLLeitschuh:
@sambsnyd I think this is one for you to take a look at ideally.

PR #4016, opened 2024-02-17
Fix 🦶🔫: Validate `visitNonNull` arguments

JLLeitschuh:
I often find myself shooting myself in the foot passing the wrong `Cursor` to `visitNonNull`.

This is a very easy mistake to make and often leads to quite a bit of confusion.

This fixes that by validating that a parent cursor is being passed, not the current cursor.

## What's changed?

Adds a bit of sanity validation to `visitNonNull(Tree tree, P p, Cursor parent)`

AFAIK this method is really only called when composing one visitor of another visitor, and
isn't called in the hot-path for normal LST tree traversal.

## What's your motivation?

Reduce how often I and other's new to OpenRewrite make the mistake of passing the `Cursor` for the tree, instead of passing the parent cursor.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
@@ -200,6 +200,11 @@ public abstract class TreeVisitor<T extends Tree, P> {
     }
 
     public T visitNonNull(Tree tree, P p, Cursor parent) {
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
@sambsnyd want to take a look at this too?

knutwannheden (approved):

PR #4021, opened 2024-02-19
Add `isAcceptable()` call to `Preconditions` methods

knutwannheden:
When constructing a precondition using `Preconditions#and()`, `or()`, or `not()` then the returned visitor must also call `isAcceptable()` on the input visitor(s) in order to avoid any `ClassCastException` in case the input visitor overrides `visit(Tree, P)` and as a result probably doesn't call `isAcceptable()` on itself.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
@@ -80,6 +80,10 @@ public class Preconditions {
         return new TreeVisitor<Tree, ExecutionContext>() {
             @Override
             public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                SourceFile sourceFile = tree instanceof SourceFile ? (SourceFile) tree : null;
+                if (sourceFile != null && !v.isAcceptable(sourceFile, ctx)) {
+                    return SearchResult.found(tree);
+                }
                 Tree t2 = v.visit(tree, ctx);
                 return tree == t2 && tree != null ?
                         SearchResult.found(tree) :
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

jkschneider (changes requested):
This should not be necessary since the resulting visitor is used in `Preconditions.check`

rewrite-core/src/main/java/org/openrewrite/Preconditions.java
```diff
@@ -80,6 +80,11 @@ public static TreeVisitor<?, ExecutionContext> not(TreeVisitor<?, ExecutionConte
         return new TreeVisitor<Tree, ExecutionContext>() {
             @Override
             public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                SourceFile sourceFile = tree instanceof SourceFile ? (SourceFile) tree : null;
+                // calling `isAcceptable()` in case `v` overrides `visit(Tree, P)`
+                if (sourceFile != null && !v.isAcceptable(sourceFile, ctx)) {
+                    return SearchResult.found(tree);
```
jkschneider, line 86: Why are we adding a search result?

knutwannheden: The idea here is that a visitor (used as a precondition) which is not applicable to some source file, should also evaluate to `true`.

In the following example we have a negated `PlainTextVisitor` precondition checking for `foo` in the content:

https://github.com/openrewrite/rewrite/blob/a438606dd0398802124e41767ef51c6f652363a4/rewrite-core/src/test/java/org/openrewrite/PreconditionsTest.java#L142-L162

The assumption here is that the (negated) precondition should evaluate to `true` for all text sources not containing `foo` and also any other source files (e.g. a quark). Does that make sense?

knutwannheden: Also note that the same search result marker is used as an indicator for a match after the `visit()` call (due to the negation the precondition must make some arbitrary change to indicate a match). So I chose to use the same technique here.

jkschneider: Ah, sorry was missing the `not` because it was hidden in github review.

PR #4025, opened 2024-02-20
Add `RecipesThatMadeChanges` in scanning phase when necessary

knutwannheden:
Any exceptions caught during the scanning phase may end up getting recorded using a `Markup.Error` marker on the source file. This however also requires a corresponding `RecipesThatMadeChanges` marker.


knutwannheden:
This is almost exactly the same as in cb67df7.

PR #4036, opened 2024-02-23
Fix documentation example for maven.AddPlugin

minisu:
## What's changed?
Fix documentation example for maven.AddPlugin

## What's your motivation?
The <executions> element has to be included by the user, as is made clear in AddPluginTest: https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/test/java/org/openrewrite/maven/AddPluginTest.java#L125-L139

## Any additional context
One could possibly argue that the documentation is correct, but the implementation is wrong; but the implementation follows the same convention used by the other parameters (`configuration` and `dependencies`).


timtebeek:
Appreciate the immediate fix! Should prevent others from stumbling over this issue with our docs.

timtebeek (approved):

PR #4037, opened 2024-02-23
Only report unrecovered repositoryAccessFailed

timtebeek:
Following questions like https://github.com/openrewrite/rewrite-maven-plugin/issues/742

Now that we surface `repositoryAccessFailed` in download logs and the Maven plugin, we don't want to raise any false alarms. We seem to also have sufficiently troubleshot the issues that prompted the eager propagation of exceptions, such that a reduced case could now make sense again.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -744,11 +744,6 @@ public class MavenPomDownloader {
                         if (e.isServerReached()) {
                             normalized = repository.withUri(httpsUri);
                         }
-                        if (!"Directory listing forbidden".equals(e.getBody())) {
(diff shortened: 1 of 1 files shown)


timtebeek:
Failures unrelated and also on `main` 😞

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -731,6 +731,9 @@ public MavenRepository normalizeRepository(MavenRepository originalRepository, M
                 String httpsUri = repository.getUri().toLowerCase().startsWith("http:") ?
                         repository.getUri().replaceFirst("[hH][tT][tT][pP]://", "https://") :
                         repository.getUri();
+                if (!httpsUri.endsWith("/")) {
+                    httpsUri += "/";
```
timtebeek, line 735: Based on [this comment](https://github.com/openrewrite/rewrite-maven-plugin/issues/742#issuecomment-1961240404).

PR #4044, opened 2024-02-23
Update to use access key authentication for Develocity build cache

tylerbertrand:
## What's changed?

The Develocity configuration has been updated to use the `GradleEnterpriseBuildCache` connector over the `HttpBuildCacheConnector` and now makes use of access key authentication. The username and password environment variables have also been removed from the workflows that used them.

## What's your motivation?

The use of username and password authentication with the Develocity build cache is being deprecated in favor of access key authentication.

### Checklist
- [ ] (N/A) I've added unit tests to cover both positive and negative cases
- [ ] (N/A) I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)


timtebeek (approved):
Thanks a lot for both the guidance and the fix! I'll see if I can replicate this across our other projects.

PR #4045, opened 2024-02-24
Fix gradle build files parsing issues

thewolt:
## What's changed?
Improved the groovy parser to support:
- reference to FQNC like in the `ProcessBuilder.Redirect.to(..)` call
- groovy xml node attribute manipulation like `n.@conf = 'runtime->default;docs->docs;sources->sources'`

## What's your motivation?
I tried to use the ChangeDependency recipe on an old gradle build with lots of custom code

- in `GroovyTypeMapping#classType`, in production i've encountered some `ClassCastException` where `JavaType.Unknown` cannot be casted to `JavaType.Class` but i was not able to reproduce in unit tests. I found that downcasting to `JavaType.FullyQualified` avoids the CCE and does not seem to break something else

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -731,9 +731,10 @@ public class GroovyParserVisitor {
 
         @Override
         public void visitClassExpression(ClassExpression clazz) {
-            queue.add(TypeTree.build(clazz.getType().getUnresolvedName())
+            String unresolvedName = clazz.getType().getUnresolvedName().replace('$', '.');
+            queue.add(TypeTree.build(unresolvedName)
                     .withType(typeMapping.type(clazz.getType()))
-                    .withPrefix(sourceBefore(clazz.getType().getUnresolvedName())));
+                    .withPrefix(sourceBefore(unresolvedName)));
         }
 
         @Override
@@ -1175,6 +1176,10 @@ public class GroovyParserVisitor {
                     jType = JavaType.Primitive.String;
                     // String literals value returned by getValue()/getText() has already processed sequences like "\\" -> "\"
                     int length = sourceLengthOfNext(expression);
+                    // this is an attribute selector
+                    if (source.startsWith("@"+value, cursor)) {
+                        length += 1;
+                    }
                     text = source.substring(cursor, cursor + length);
                     int delimiterLength = 0;
                     if (text.startsWith("$/")) {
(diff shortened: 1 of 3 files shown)


timtebeek:
Thanks for diving in for a fix! Let us know when you'd consider this ready for review.

thewolt:
I've got one question though... it seems that the issue i'm trying to fix is not related to the ChangeDependency recipe specifically but rather the gradle parser in general. 

Is there a better place to put the test?

timtebeek:
Indeed looks like you're fixing a parser bug instead; that should help even more folks & cases, thanks! Might be easier to reproduce in [a smaller test as seen in these tree test examples](https://github.com/openrewrite/rewrite/tree/2055f3fd7993e6a5413e00265d474de69be81f2a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree). Does that help?

thewolt:
Yes it absolutely helped. The tests have been successfully moved there and I may have found another bug when an inner class is imported.

The last thing is about the FIXME that i left in GroovyTypeMapping... it is a private method and the signature change does not seem to bother the two usages. Unfortunately I was not able to reproduce it in unit test. Does this change require a unit test ?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyTypeMapping.java
```diff
@@ -83,12 +83,13 @@ public JavaType type(@Nullable ASTNode type) {
         throw new UnsupportedOperationException("Unknown type " + type.getClass().getName());
     }
 
-    private JavaType.Class classType(ClassNode node, String signature) {
-        JavaType.Class clazz;
+    //FIXME: reproduce CCE in unit tests
```
timtebeek, line 86:
```suggestion
```

timtebeek:
Glad to hear that helped! The new smaller tests look much cleaner indeed. I'm tagging Knut or Shannon for a review here, as they have more experience with the Groovy/Gradle parsers, and are more likely to spot any possible impact I might overlook. I'd think it's fine not to add the explicit test you mentioned if it's very awkward to trigger or test for, but perhaps they feel differently.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1175,6 +1176,10 @@ public void visitConstantExpression(ConstantExpression expression) {
                     jType = JavaType.Primitive.String;
                     // String literals value returned by getValue()/getText() has already processed sequences like "\\" -> "\"
                     int length = sourceLengthOfNext(expression);
+                    // this is an attribute selector
+                    if (source.startsWith("@"+value, cursor)) {
```
shanman190, line 1180: This feels like it should be represented in the Groovy AST in some form to indicate that the current identifier is a specialized one using the attribute operator.

The way that these operators are typically represented is by the usage of a `Marker` type on the `J.Identifier`, but let me double check the Groovy AST node here first to confirm this suspicion.

thewolt:
Build failed because:
```
Caused by:
                    org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find org.openrewrite:rewrite-gradle:8.19.0-SNAPSHOT.
                    Searched in the following locations:
                      - file:/home/runner/.m2/repository/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/maven-metadata.xml
                      - file:/home/runner/.m2/repository/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/rewrite-gradle-8.19.0-SNAPSHOT.pom
                      - https://oss.sonatype.org/content/repositories/snapshots/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/maven-metadata.xml
                      - https://oss.sonatype.org/content/repositories/snapshots/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/rewrite-gradle-8.19.0-20240227.124952-11.pom
                      - https://repo.maven.apache.org/maven2/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/maven-metadata.xml
                      - https://repo.maven.apache.org/maven2/org/openrewrite/rewrite-gradle/8.19.0-SNAPSHOT/rewrite-gradle-8.19.0-SNAPSHOT.pom
                    Required by:
                        unspecified:unspecified:unspecified > org.openrewrite.gradle.tooling:plugin:2.1.0-SNAPSHOT:20240226.180904-8
```

It does not seem related to my change ...

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jkschneider: I think I'm fine with this being calculable state on the identifier without requiring a marker.

PR #4047, opened 2024-02-24
Add `AddNullMethodArgument` recipe

pstreef:
## What's changed?
Add recipe `AddNullMethodArgument`

## What's your motivation?
When changing a method we often add a nullable argument. Sometimes overriding it as to not break existing consumers. With this recipe we can migrate more quickly by adding a null value as the argument.

line 104 of the recipe, creating the actual `null` literal.
~Also `addArgumentsConsecutively` is not working as expected. Any suggestions? If not I'll give it another try soon.~

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2024 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,23 +22,24 @@ import org.openrewrite.test.RewriteTest;
 import static org.openrewrite.java.Assertions.java;
 
 class AddNullMethodArgumentTest implements RewriteTest {
-    @Language("java")
-    String b = """
-      class B {
-         public static void foo() {}
-         public static void foo(Integer n) {}
-         public static void foo(Integer n1, Integer n2) {}
-         public static void foo(Integer n1, Integer n2, Integer n3) {}
-         public B() {}
-         public B(Integer n) {}
-      }
-      """;
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.parser(JavaParser.fromJavaVersion().dependsOn("""
+          class B {
+             public static void foo() {}
+             public static void foo(Integer n) {}
+             public static void foo(Integer n1, Integer n2) {}
+             public static void foo(Integer n1, Integer n2, Integer n3) {}
+             public B() {}
+             public B(Integer n) {}
+          }
+          """));
+    }
 
     @Test
     void addToMiddleArgument() {
         rewriteRun(
           spec -> spec.recipe(new AddNullMethodArgument("B foo(Integer, Integer)", 1)),
-          java(b),
           java(
             "public class A {{ B.foo(0, 1); }}",
             "public class A {{ B.foo(0, null, 1); }}"
@@ -54,8 +55,9 @@ class AddNullMethodArgumentTest implements RewriteTest {
             new AddNullMethodArgument("B foo(Integer, null)", 1)
           ),
           java(b),
-          java("public class A {{ B.foo(0); }}",
-            "public class A {{ B.foo(0, null, null); }}"
+          java(
+            "class A {{ B.foo(0); }}",
+            "class A {{ B.foo(0, null, null); }}"
           )
         );
     }
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import org.intellij.lang.annotations.Language;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class AddNullMethodArgumentTest implements RewriteTest {
+    @Language("java")
+    String b = """
+      class B {
+         public static void foo() {}
+         public static void foo(Integer n) {}
+         public static void foo(Integer n1, Integer n2) {}
+         public static void foo(Integer n1, Integer n2, Integer n3) {}
+         public B() {}
+         public B(Integer n) {}
+      }
+      """;
+
+    @Test
+    void addToMiddleArgument() {
+        rewriteRun(
+          spec -> spec.recipe(new AddNullMethodArgument("B foo(Integer, Integer)", 1)),
+          java(b),
+          java(
+            "public class A {{ B.foo(0, 1); }}",
+            "public class A {{ B.foo(0, null, 1); }}"
+          )
+        );
+    }
+
+    @Test
+    void addArgumentsConsecutively() {
+        rewriteRun(
+          spec -> spec.recipes(
+            new AddNullMethodArgument("B foo(Integer)", 1),
+            new AddNullMethodArgument("B foo(Integer, Integer)", 1)
+          ),
+          java(b),
+          java("public class A {{ B.foo(0); }}",
+            "public class A {{ B.foo(0, null, null); }}"
+          )
+        );
+    }
+
+    @Test
+    void addToConstructorArgument() {
+        rewriteRun(
+          spec -> spec.recipe(new AddNullMethodArgument("B <constructor>()", 0)),
+          java(b),
+          java(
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
```diff
@@ -0,0 +1,125 @@
...
+
+                args.add(argumentIndex, new J.Literal(randomId(), args.isEmpty() ? Space.EMPTY : Space.SINGLE_SPACE, Markers.EMPTY, "null", "null", null, JavaType.Primitive.Null));
+                m = m.withArguments(args);
+
+                JavaType.Method methodType = m.getMethodType();
+                if (methodType != null) {
+                    List<String> parameterNames = new ArrayList<>(methodType.getParameterNames());
+                    parameterNames.add(argumentIndex, "null");
+                    List<JavaType> parameterTypes = new ArrayList<>(methodType.getParameterTypes());
+                    parameterTypes.add(argumentIndex, JavaType.Primitive.Null);
```
timtebeek, line 112: I'm not so sure about this parameter type `Null` here; Perhaps we need another argument where folks can pass in a FQN of the target type. That way any subsequent recipes in a larger migration will also match the method invocation going forward, especially if this is one partial step of a larger migration.

By comparison, now we end up with this after a recipe execution in `void addToMiddleArgument()`
image

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
```diff
@@ -0,0 +1,125 @@
...
+            List<Expression> originalArgs = m.getArguments();
+            if (methodMatcher.matches(m) && (long) originalArgs.size() >= argumentIndex) {
+                List<Expression> args = new ArrayList<>(originalArgs);
+
+                if(args.size() == 1 && args.get(0) instanceof J.Empty) {
+                    args.remove(0);
+                }
+
+                args.add(argumentIndex, new J.Literal(randomId(), args.isEmpty() ? Space.EMPTY : Space.SINGLE_SPACE, Markers.EMPTY, "null", "null", null, JavaType.Primitive.Null));
+                m = m.withArguments(args);
```
timtebeek, line 105: Perhaps this would be a little more in line with what we use elsewhere? 
```
m = m.withArguments(ListUtils.insert(m.getArguments(), new J.Literal(...), argumentIndex));
```

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    void addArgumentsConsecutively() {
+        rewriteRun(
+          spec -> spec.recipes(
+            new AddNullMethodArgument("B foo(Integer)", 1),
+            new AddNullMethodArgument("B foo(Integer, null)", 1)
+          ),
+          java(b),
+          java("public class A {{ B.foo(0); }}",
+            "public class A {{ B.foo(0, null, null); }}"
```
timtebeek, line 58: Typically I encourage folks to remove as much from their test cases as possible, such as in this case the public modifiers.
```suggestion
          java(
            "class A {{ B.foo(0); }}",
            "class A {{ B.foo(0, null, null); }}"
```

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
```diff
@@ -0,0 +1,74 @@
...
+         public B() {}
+         public B(Integer n) {}
+      }
+      """;
+
+    @Test
+    void addToMiddleArgument() {
+        rewriteRun(
+          spec -> spec.recipe(new AddNullMethodArgument("B foo(Integer, Integer)", 1)),
+          java(b),
```
timtebeek, line 41:
```suggestion
```

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
```diff
@@ -0,0 +1,74 @@
...
+    String b = """
+      class B {
+         public static void foo() {}
+         public static void foo(Integer n) {}
+         public static void foo(Integer n1, Integer n2) {}
+         public static void foo(Integer n1, Integer n2, Integer n3) {}
+         public B() {}
+         public B(Integer n) {}
+      }
+      """;
```
timtebeek, line 35: Perhaps a matter of opinion, but I like a single `dependsOn` over repeated `java(b)` below.
```suggestion
    @Override
    public void defaults(RecipeSpec spec) {
        spec.parser(JavaParser.fromJavaVersion().dependsOn("""
          class B {
             public static void foo() {}
             public static void foo(Integer n) {}
             public static void foo(Integer n1, Integer n2) {}
             public static void foo(Integer n1, Integer n2, Integer n3) {}
             public B() {}
             public B(Integer n) {}
          }
          """));
    }
```

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2020 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright 2020 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

timtebeek (commented):
Great start! I think we can improve it with some minor touches to make it more widely applicable and correct.

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
```diff
@@ -0,0 +1,125 @@
...
+    int argumentIndex;
+
+    @Override
+    public String getInstanceNameSuffix() {
+        return String.format("%d in methods `%s`", argumentIndex, methodPattern);
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Delete method argument";
```
shanman190, line 62:
```suggestion
        return "Add a `null` method argument"
```

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Delete an argument from method invocations.";
```
shanman190, line 67:
```suggestion
        return "Add a `null` argument to method invocations."
```

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
pstreef: I copied most of the cose (including this part) from `DeleteMethodArgument` but happy to change to this!

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
pstreef: ah, now this makes a lot more sense indeed. I did not fully understand what this part was doing to be honest 😅 
I think passing the type (maybe also param name?) is OK as you should always know the exact method you are migrating to. 
I don't think there is a way to determine/find this data if the method is not defined in the project is there?

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
```diff
@@ -0,0 +1,89 @@
...
+          spec -> spec.recipe(new AddNullMethodArgument("B <constructor>()", 0, Integer.class.getName(), "arg", false)),
+          java(
+            "class A { B b = new B(); }",
+            "class A { B b = new B(null); }"
+          )
+        );
+    }
+
+    @Test
+    void addCastToConflictingArgumentType() {
```
pstreef, line 79: I've added this because of possible conflicts. However I'm not sure how to get the cast to only use the not-so-fully qualified name.

pstreef:
@shanman190 Thanks, not sure how I missed those 🙈

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
timtebeek: I think this is fine; should be rare that there's overloads, which require an explicit cast to String.

If you really want to get rid of this you can shorted fully qualified type references as we do here:
https://github.com/openrewrite/rewrite-templating/blob/e69126c2209f4edd1cb8e203d6eaeda26ff744b1/src/main/java/org/openrewrite/java/template/internal/AbstractRefasterJavaVisitor.java#L45

timtebeek (approved):
Nice addition!

rewrite-java/src/main/java/org/openrewrite/java/AddNullMethodArgument.java
timtebeek: Not as easily reliably no, especially when we start stacking migration recipes where you start out without that new method available in an older version of a library.

rewrite-java-test/src/test/java/org/openrewrite/java/AddNullMethodArgumentTest.java
pstreef: I tried this but it does not shorten the cast.

knutwannheden: For `java.lang` we indeed don't shorten fully qualified names, as that could be wrong.

knutwannheden: Although note that also `java.lang` types would end up as unqualified if the compilation unit already contains unqualified references to that type (which is often the case), so using this visitor is probably still a good idea.

PR #4048, opened 2024-02-25
Make unit test work correctly on Windows

Bananeweizen:
The String representation of a path is related to the OS (or more correctly the underlying file system implementation). Compare the path without converting to String to be OS agnostic.

## What's your motivation?
The unit test failed on Windows, because of slash-versus-backslash issues. With this change applied, all tests pass on my Windows system.


timtebeek (approved):

timtebeek:
Thanks as always! :)

PR #4049, opened 2024-02-25
Make @Issue repeatable

Bananeweizen:
## What's changed?
Allow multiple issue annotations for a test.

## What's your motivation?
I once used that accidentally because I have it in other projects, and had to fix the PR after I noticed the compile error.

It uses a nested interface for the container class to avoid it being shown in typical code completion of the IDE.

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/Issue.java
+++ b/rewrite-test/src/main/java/org/openrewrite/Issue.java
@@ -16,6 +16,7 @@
 package org.openrewrite;
 
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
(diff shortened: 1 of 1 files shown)


timtebeek:
I'd considered this the other day as well; thanks opening a PR! An alternative I'd considered was using `String[] value` to keep just the same single annotation for most cases. Any thoughts on that?

Bananeweizen:
I find copying the entire line and replacing the url more simple than adding a second URL to an existing tag (and quite honestly, I also like reading 2 similarly formatted lines more than the one tag spread over 2 lines or both URLs in the same line). But that surely is just personal preference, I had similar discussions also with colleagues and we typically see at least as many opinions as people involved, if not more. :)

I'm not sure if there are any good arguments for/against one or the other from the view point of binary compatibility, usability with reflection or other more technical aspects. I've also not checked if there are other custom tags in your code base, where we might want to align. So if you have some strong opinion for either one, we can surely change this.

timtebeek (approved):

timtebeek:
Thanks for laying out your thoughts; no strong opinions here, so then let's indeed got for repeated annotations.

PR #4053, opened 2024-02-27
Add recipe `AddLiteralMethodArgument`

pstreef:
This recipe finds a method and adds a literal argument at the specified index.

## What's changed?
Recipe was added

## What's your motivation?
In stead of only adding a null argument we can now also specify a literal. This means when adding an argument and fixing the calls with a recipe we no longer need to require the field to be boxed and nullable for primitives.

-

@timtebeek 

No

## Any additional context
It's very similar to the `AddNullMethodArgument` recipe added last week.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.apache.commons.lang3.StringUtils;
+import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.*;
+import org.openrewrite.marker.Markers;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.java.tree.JavaType.Primitive;
+
+/**
+ * This recipe finds method invocations matching a method pattern and uses a zero-based argument index to determine
+ * which argument is added with a literal value.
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
```diff
@@ -0,0 +1,155 @@
...
+    @Option(displayName = "Literal",
+            description = "The literal value that we add the argument for.",
+            example = "0")
+    Object literal;
+
+    @Option(displayName = "Parameter type",
+            description = "The type of the parameter that we add the argument for. Options are: `String`, `int`,`short`, " +
+                          "`long`, `float`, `double`, `boolean` or `char`. Defaults to `String`.",
+            required = false,
+            example = "String")
```
timtebeek, line 69:
```suggestion
            example = "String",
            valid = {"String", "int", "short", "long", "float", "double", "boolean", "char"})
```

rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
```diff
@@ -0,0 +1,155 @@
...
+    int argumentIndex;
+
+    @Option(displayName = "Literal",
+            description = "The literal value that we add the argument for.",
+            example = "0")
+    Object literal;
+
+    @Option(displayName = "Parameter type",
+            description = "The type of the parameter that we add the argument for. Options are: `String`, `int`,`short`, " +
+                          "`long`, `float`, `double`, `boolean` or `char`. Defaults to `String`.",
```
timtebeek, line 67:
```suggestion
            description = "The type of the parameter that we add the argument for. Defaults to `String`.",
```

rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
```diff
@@ -0,0 +1,155 @@
...
+     * A zero-based index that indicates which argument will be added as null to the method invocation.
+     */
+    @Option(displayName = "Argument index",
+            description = "A zero-based index that indicates which argument will be added as null to the method invocation.",
+            example = "0")
+    int argumentIndex;
+
+    @Option(displayName = "Literal",
+            description = "The literal value that we add the argument for.",
+            example = "0")
```
timtebeek, line 62: Since the type example below is a String, and to differentiate this from the index, I thought it best to use a string value here too.
```suggestion
            example = "abc")
```

rewrite-java/src/main/java/org/openrewrite/java/AddLiteralMethodArgument.java
```diff
@@ -0,0 +1,155 @@
...
+                    valueSource = String.format("'%s'", getLiteral());
+                }else {
+                    primitive = Primitive.fromKeyword(primitiveType.toLowerCase());
+                    valueSource = String.valueOf(getLiteral());
+                }
+
+                if (Arrays.asList(Primitive.Byte, Primitive.Null, JavaType.Primitive.Void, Primitive.None).contains(primitive)) {
+                    throw new IllegalArgumentException("Invalid primitive type: " + primitiveType);
+                }
+
```
timtebeek, line 137: With the `valid` above this handling should not be necessary.
```suggestion
```

timtebeek (approved):
Approved with a few quick suggestions.

pstreef:
Thanks, I didn't know `valid` existed, nice!

PR #4065, opened 2024-03-02
Parse property values starting with the delimiter character

### LINKED ISSUE #4026: foo.properties is not print idempotent due to `key==value`

I am using

- Maven/Gradle 5.2.1

foo.properties containing
```properties
key==value
```

```
[WARNING] There were problems parsing foo.properties
[WARNING] java.lang.IllegalStateException: foo.properties is not print idempotent. 
... (skipped diff header output)
-key==value
+key=value
...
  org.openrewrite.Parser.requirePrintEqualsInput(Parser.java:52)
```

I checked the Javadoc of Properties and an equals sign immediately after the first equals should be considered part of the value. Also when loading "key==value" into Java properties, it returns "=value" when asking for "key".

Not for this one.

Bananeweizen:
## What's changed?
In the state machine of the properties parser do not accept further delimiters after the first non escaped delimiter was found.

## What's your motivation?
Fixes #4026.

The actual bug fix is in the first commit, so you may want to review it separately. The second commit removes all arithmetic from the state variable and gives readable names to the states to make debugging more simple for the next one looking into this.

### Checklist

### DIFF
--- a/rewrite-properties/src/main/java/org/openrewrite/properties/PropertiesParser.java
+++ b/rewrite-properties/src/main/java/org/openrewrite/properties/PropertiesParser.java
@@ -238,16 +238,19 @@ public class PropertiesParser implements Parser {
                         break;
                     } else if (c == '=' || c == ':') {
                         delimiter = Properties.Entry.Delimiter.getDelimiter(String.valueOf(c));
+                        state++;
+                        break;
                     }
-                    state++;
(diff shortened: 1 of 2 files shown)


knutwannheden (approved):
Nice! I like what you did with the enum there. Also went through my mind last time I worked on this.

PR #4066, opened 2024-03-03
Detect instanceof pattern variables in scope

timtebeek:
## What's your motivation?
Used in `InstanceOfPatternMatch` to detect variables that exist in the surrounding context, which now fails to pick up pattern matching for instance of variables, and as such introduces two identically named variables.

- https://github.com/openrewrite/rewrite-static-analysis/pull/265
- https://github.com/openrewrite/rewrite-static-analysis/pull/175

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java
@@ -244,6 +244,17 @@ public class VariableNameUtils {
         }
 
         @Override
+        public J.InstanceOf visitInstanceOf(J.InstanceOf instanceOf, Set<String> strings) {
+            if (instanceOf.getPattern() instanceof J.Identifier) {
+                Set<String> names = nameScopes.get(currentScope.peek());
+                if (names != null) {
+                    names.add(((J.Identifier)instanceOf.getPattern()).getSimpleName());
+                }
+            }
+            return super.visitInstanceOf(instanceOf, strings);
+        }
+
+        @Override
         public J.VariableDeclarations.NamedVariable visitVariable(J.VariableDeclarations.NamedVariable variable, Set<String> strings) {
             Set<String> names = nameScopes.get(currentScope.peek());
             if (names != null) {
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/VariableNameUtilsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/VariableNameUtilsTest.java
@@ -185,7 +185,24 @@ class VariableNameUtilsTest implements RewriteTest {
               """
           )
         );
+    }
 
+    @Test
+    void detectInstanceOf() {
+        rewriteRun(
+          baseTest("ifBlock", "ifBlock,pattern,methodParam"),
+          java(
+            """
+              class Test {
+                  void method(Object methodParam) {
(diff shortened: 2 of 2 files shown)


knutwannheden:
This is probably a hood start and we could go ahead and integrate this, if it already fixes some issues. To be on the safe side, the pattern variable should also be added to the scope of the containing block, as it can in fact also be visible there, depending on what the if and else statements contain.

rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java
```diff
@@ -243,6 +243,17 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, Se
...
 
+        @Override
+        public J.InstanceOf visitInstanceOf(J.InstanceOf instanceOf, Set<String> strings) {
+            if (instanceOf.getPattern() instanceof J.Identifier) {
+                Set<String> names = nameScopes.get(currentScope.peek());
+                if (names != null) {
+                    names.add(((J.Identifier)instanceOf.getPattern()).getSimpleName());
+                }
+            }
+            return super.visitInstanceOf(instanceOf, strings);
```
timtebeek, line 254: @knutwannheden is this what you meant by adding it to the parent scope?
```suggestion
        public J.InstanceOf visitInstanceOf(J.InstanceOf instanceOf, Set<String> namesInScope) {
            if (instanceOf.getPattern() instanceof J.Identifier) {
                Set<String> names = nameScopes.get(currentScope.peek());
                if (names != null) {
                    String simpleName = ((J.Identifier) instanceOf.getPattern()).getSimpleName();
                    names.add(simpleName);
                    namesInScope.add(simpleName);
                }
            }
            return super.visitInstanceOf(instanceOf, namesInScope);
```

knutwannheden: Not exactly sure what the effect would be of this change. But For example the following code must not end up creating two pattern variables with the same name, as that would give a compile error:
```java
        Object o = "";
        if (!(o instanceof String)) {
            throw new IllegalArgumentException(((String) o));
        }
        if (o instanceof String) {
            System.out.println((String) o);
        }
```

knutwannheden: The given example is very contrived, as it would result in a runtime exception. But I wanted to get the point across regarding the compiler error. I could otherwise also come up with more sensible examples.

So, to be on the safe side, a pattern variable as in the following code should be regarded as belonging to the surrounding block (not just the if statement's then block):
```java
        Object o = "";
        if (o instanceof String s) {
            // ...
        }
```

That way the `VariableNameUtils` should not propose `s` as a variable name in any statements following the `if` statement.

timtebeek: Looking this over again and how it's used in InstanceOfPattern match I _think_ right now we don't need the above suggested fix, given how we aggregate the scope names in postVisit
https://github.com/openrewrite/rewrite/blob/0f0dc80f793d0e33f60e1e982318d82c40cc4201/rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java#L193-L213


I'll try to pick this up in rewrite-static-analysis.

PR #4074, opened 2024-03-05
`MavenVisitor` should not return incorrect `ResolutionResult` for multi module maven projects

motu55:
## What's changed?
Always fetch the resolution result since in a multi module maven project it is not guaranteed that a new mavenvisitor is generated for each pom file.

## What's your motivation?
the find maven dependency did not work in my multi module project so i checked where it fails and saw that not a new maven visitor is created for each pom file. so the find dependency on a child pom file checks with the resolution result of the parent pom file which dies not have the corresponding dependencies

the test i added works always so to reproduce a local small maven multi module project has to be created 

another way could be to guarantee that a new visitor is created for each pom file

## Any additional context
unfortunately does the test i added work also with the false code since during the tests always a maven visitor is created for each file, this is unfortunate since the tes behavior and the actual behavior are not the same 

### Checklist
- [x ] I've added unit tests to cover both positive and negative cases
- [ x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files (nope since it destroyed the formatting of the whole files)


timtebeek:
Thanks for looking into this! Indeed not having the same behavior in tests as we do in practice is definitely something to cover, as it might lead to other unexpected issues that are then hard to reproduce.

Then from there we can indeed explore the change you had earlier in https://github.com/openrewrite/rewrite/pull/4074/commits/9763a71aa43ea2eb5d9ab3b4310027b5fb3c6bce , and see if that is indeed then fixes the test here, and behavior in practice. I did wonder about any potential performance impact, but we can look at that separately; correctness is more important of course when it comes to determining to add/change/remove dependencies.

motu55:
@timtebeek 
now it is ready for a review.
i do not know why all these commits show up but the files changed view is correct

timtebeek:
Thanks! Didn't manage to fit a review in before my holiday, but I've shared the link with colleagues to have a look.

sambsnyd:
@motu55 it is already the case that each file gets its own instance of each visitor, unless a `Recipe` incorrectly returns a cached value from `getVisitor()`. If `Recipe.getVisitor()` returned a cached value it would be a bug with that recipe failing to follow the contract expected of it.

For example, the `multiModule()` test you added to `FindDependencyTest` already passes without any change to `MavenVisitor`.

So to help Recipe authors detect when they've made this category of mistake I've added logic to detect and throw: https://github.com/openrewrite/rewrite/commit/8091c9c30753aa2128c2f17f51421c085853d781

Thanks for the contribution.

PR #4087, opened 2024-03-08
Printable Recipe Datatables 

ryan-hudson:
## What's changed?
Added logic in the RewriteRun Class in Rewrite-Core to output Datatable values as a CSV

## What's your motivation?
Datatable information is currently only accessible in memory in OSS, rendering it extremely difficult to use/access.  This change (coupled with the Maven Plugin PR below) would allow easy access to any datatable information created by recipes. This is especially useful for OSS users wanting to use recipes designed to only output datatable information. 

See related [Maven Plugin PR](https://github.com/openrewrite/rewrite-maven-plugin/pull/751)


sambsnyd:
Thanks @ryan-hudson !

PR #4096, opened 2024-03-14
Compact `GitProvenance.Committer` representation

knutwannheden:
The in-memory representation of the `GitProvenance.Committer` requires quite a lot of memory for the commit statistics as captured by the `TreeMap<LocalDate, Integer>` structure. Replacing with two more compact `int[]` fields.


anandfresh (approved):

knutwannheden:
Comparing for the example of dotnet/runtime the difference in retained memory by the `GitProvenance` marker is 5594800 vs 1233192 bytes.

PR #4098, opened 2024-03-15
Add fileContentsUrl option for CreateYamlFile

### LINKED ISSUE #4011: Create YAML File with remote content
I want to create a Yaml file with a remote content

Check that the URL exists and the content type is yaml

Nothing

With an URL with content

```yaml
content:
  languages:
    - french
    - english
```

I want to have a yaml file with this content

No

PhilKes:
## What's changed?
New option `fileContentsUrl` for the `CreateYamlFile` recipe to allow file contents from a URL.
The `fileContents` option takes precedence, if both the `fileContents` and `fileContentsUrl` has a value.

## What's your motivation?
Closes #4011 

-

-

No

## Any additional context
-

### Checklist


timtebeek (approved):
Thanks for getting this started @PhilKes ! Review took a little longer due to my travels, but glad to see this implemented. I've pushed up a few small polishing commits, to for instance stop us from repeatedly downloading the file even for non-matching target files, which should help performance. I've also simplified the test to no longer require a local file.

timtebeek:
Having some trouble getting the tests to work; will revisit later. To get the same result locally be sure to clear out `~/.rewrite/remote/`.

rewrite-yaml/src/test/java/org/openrewrite/yaml/CreateYamlFileTest.java
```diff
@@ -113,4 +132,76 @@ void shouldAddAnotherFile() {
...
+                return null;
+            }
+
+            @Override
+            public @Nullable Path get(URI uri) {
+                return null;
+            }
+
+            @Override
+            public @Nullable Path put(URI uri, InputStream is, Consumer<Throwable> onError) {
```
PhilKes, line 163: Isn't the `put` the problem, because it returns null?
What about something like this:
```java
RemoteExecutionContextView.view(ctx).setArtifactCache(new RemoteArtifactCache() {
      
      @Override
      public @Nullable Path get(URI uri) {
          return null;
      }

      @Override
      public @Nullable Path put(URI uri, InputStream is, Consumer<Throwable> onError) {
          try {
              var file = File.createTempFile("rewrite", "yaml");
              Files.copy(is, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
              return file.toPath();
          } catch (IOException e) {
          }
      }
  });
```

timtebeek: That indeed works, thanks a lot!

PR #4112, opened 2024-03-28
Dont directly set context, but use defaults for the gradle/non gradle constructor

pstreef:


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -106,8 +109,8 @@ public MavenPomDownloader(Map<Path, Pom> projectPoms, ExecutionContext ctx,
      */
     public MavenPomDownloader(ExecutionContext ctx) {
         this(emptyMap(), HttpSenderExecutionContextView.view(ctx).getHttpSender(), ctx);
-        MavenExecutionContextView.view(ctx).setAddCentralRepository(false);
-        MavenExecutionContextView.view(ctx).setAddLocalRepository(false);
+        this.addCentralRepository = Boolean.TRUE.equals(MavenExecutionContextView.view(ctx).getAddCentralRepository());
+        this.addLocalRepository = Boolean.TRUE.equals(MavenExecutionContextView.view(ctx).getAddLocalRepository());
```
shanman190, line 113: This side is pretty exclusively used with the Gradle pieces which Maven Central was opted to not ever be included by default. I kinda wonder if even Maven Local shouldn't be enabled either ever.

Presently, the `GradlrProject` marker provides the repositories in full in the order with which they are configured in the Gradle script. With that said, this feels risky to do in that context.

knutwannheden:
It looks like this PR broke a few tests.

shanman190:
Yeah, I did also notice that adding the default repositories was set to `true` by default which is what led to my comment with respect to the Gradle bits.
https://github.com/openrewrite/rewrite/commit/be7176596788d2f4989ed21a82a42378ba0664b3

PR #4118, opened 2024-04-02
Remove unused imports after `RemoveObjectsIsNull`

tobli:
## What's changed?
The `RemoveObjectsIsNull` recipe will now remove unused imports.

## What's your motivation?
The `RemoveObjectsIsNull` recipe would sometimes leave unused imports, if the only usage of `java.util.Objects` was removed.

## Any additional context
I added a test case that was mentioned in the original issue when this recipe was added.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveObjectsIsNullTest.java
@@ -47,7 +47,6 @@ class RemoveObjectsIsNullTest implements RewriteTest {
               }
               """,
             """
-              import static java.util.Objects.isNull;
               public class A {
                   public void test() {
                       boolean a = true;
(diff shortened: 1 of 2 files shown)


tobli:
sorry, pushed diff before syncing fork. This PR is (largely) redundant

knutwannheden:
Thanks! I can see that the static import isn't always removed when it could be removed.

tobli:
> Thanks! I can see that the static import isn't always removed when it could be removed.

Exactly! I had an unsynced branch, and essentially made similar changes to what Tim had done upstream. That said, my test cases for removal of the static import did work prior to syncing (my fork was out of date from January) but doesn't now.
I started digging a bit deeper but hasn't solved it yet. I did notice that `TypesInUse` does group MemberReferences (like `Objects::isNull`) with method uses. I think this distinction is needed in RemoveImport to make the right decision on what imports to keep.

Issue #4123, opened 2024-04-07
Invalid pattern for dependenciesFromResources in JavaParser


ckcd:
## Description of issue

In short: the `file.getName()` and pattern `Pattern.compile("[/\\\\]" + artifactName + "-?.*\\.jar$")` is mismatch so that the pattern.matcher will always be `false`. The `file.getName()` returns a simple filename without any `/` but the pattern needs a `/`.

In this [dependenciesFromResources](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L115) method it list all files in local system (default path is ~/.rewrite/classpath and it doesn't matter for this bug), then try to compute whether these files match the artifactNames given in args.

As we can see the source code is:
```java
Pattern jarPattern = Pattern.compile("[/\\\\]" + artifactName + "-?.*\\.jar$");
File[] extracted = resourceTarget.listFiles();
if (extracted != null) {
    for (File file : extracted) {
        if (jarPattern.matcher(file.getName()).find()) {
            artifacts.add(file.toPath());
            continue nextArtifact;
        }
    }
}
```
For example:
- if artifactName = `spring-beans-5.*` and their are a `spring-beans-5.3.27.jar` in target path, then we want **match** for it;
- if artifactName = `spring-beans-4.*` and their are a `spring-beans-5.3.27.jar` in target path, then we want **NOT match** for it;

But the issue is: for now this pattern will always be **NOT match** no matter what the artifactName and local file is. The reason is the `file.getName()` and pattern `Pattern.compile("[/\\\\]" + artifactName + "-?.*\\.jar$")` is mismatch.
- `file.getName()` returns a short name without any `/`.
- but the pattern needs a `/`.

## What version of OpenRewrite are you using?

I am using

- OpenRewrite 8.1.6
- Maven/Gradle plugin Maven 3.3.9 (irrelevant)
- rewrite-java 8.1.6

## How are you running OpenRewrite?

I am using the [sbm-support-rewrite](https://github.com/spring-projects-experimental/spring-boot-migrator/tree/main/sbm-support-rewrite) in spring-boot-migrator which will call open-rewrite.

## What is the smallest, simplest way to reproduce the problem?

Any codes that using [dependenciesFromResources](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L115) can reproduce this issue. For my case it is:

```java
[...]

PR #4124, opened 2024-04-07
fix invalid jarPattern in JavaParser

ckcd:
## What's changed?

Fix invalid pattern in https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L123

The `file.getName()` and pattern `Pattern.compile("[/\\\\]" + artifactName + "-?.*\\.jar$")` is mismatch so that the pattern.matcher will always be `false`. The `file.getName()` returns a simple filename without any `/` but the pattern needs a `/`.

See more details in https://github.com/openrewrite/rewrite/issues/4123

## What's your motivation?

Fix https://github.com/openrewrite/rewrite/issues/4123

No

No

No

## Any additional context

No

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
@@ -124,7 +124,7 @@ public interface JavaParser extends Parser {
             File[] extracted = resourceTarget.listFiles();
             if (extracted != null) {
                 for (File file : extracted) {
-                    if (jarPattern.matcher(file.getName()).find()) {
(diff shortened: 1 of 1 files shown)


ckcd:
@traceyyoshima @knutwannheden could you please take a look at this PR? thanks.

timtebeek:
Thanks a lot @ckcd for the detailed write up in 
- #4124 
And the subsequent fix here. Surprised this hadn't been flagged before, but your explanation makes sense.

ckcd:
@timtebeek Thanks for your reply!

PR #4130, opened 2024-04-11
Add providedCompile to the DependencyHandlerSpec for ChangeDependency gradle 

burghduffkc:
## What's changed?

Added the providedCompile to the DependencyHandlerSpec and created a test for it
## What's your motivation?

If you run the gradle ChangeDependency recipe against a war project, it will not update the dependencies for the providedCompile configuration. 
### Checklist

### DIFF
--- a/rewrite-gradle/src/main/groovy/RewriteGradleProject.groovy
+++ b/rewrite-gradle/src/main/groovy/RewriteGradleProject.groovy
@@ -50,6 +50,8 @@ interface DependencyHandlerSpec extends DependencyHandler {
     Dependency compileOnly(Object dependencyNotation, @DelegatesTo(strategy=Closure.DELEGATE_ONLY, value= ExternalDependency) Closure closure)
     Dependency implementation(Object... dependencyNotation)
     Dependency implementation(Object dependencyNotation, @DelegatesTo(strategy=Closure.DELEGATE_ONLY, value= ExternalDependency) Closure closure)
+    Dependency providedCompile(Object... dependencyNotation)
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for adding this for WAR projects @burghduffkc ; and for your patience while I was out traveling. Looking over this now it seems [the WAR plugin adds two configurations](https://docs.gradle.org/current/userguide/war_plugin.html#sec:war_dependency_management):

> The War plugin adds two dependency configurations:
> 
> **providedCompile**
> This configuration should be used for dependencies required at compilation but which are provided by the environment in which the WAR is deployed. Dependencies declared here are thus visible to the main and test compilation classpaths.
> 
> **providedRuntime**
> This configuration should be used for dependencies required at runtime but which are provided by the environment in which the WAR is deployed. Dependencies declared here are only visible to the main and test runtime classpaths.

This PR already adds `providedCompile`; do you see any reason to add or omit `providedRuntime` ?

timtebeek (approved):
Approved already; let me know if you'd want a follow up commit to include `providedRuntime` as well. :) Thanks again!

burghduffkc:
I think that should be fine.  If you want I can add that and update the
tests.  Is there any reason why all known configurations cannot be added?

timtebeek:
Thanks for the offer to help; I've pushed that change to https://github.com/openrewrite/rewrite/commit/addfc0c656978d5dc3c2f4c88a165be911e0284b directly, so we should be good. Thanks again and let me know if you have any other suggestions for improvement! :)

PR #4140, opened 2024-04-19
Instead of throwing most MavenDownloadingException, record them into …

sambsnyd:
…the resolution result. This allows for partial success when processing poms which are invalid or contain unresolvable dependencies without hiding the existence of these problems.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
@@ -66,6 +66,12 @@ public class RawPom {
     @Nullable
     String version;
 
+    // This is an obsolete field that is no longer used, but it is present in some particularly old POMs
+    @EqualsAndHashCode.Include
+    @ToString.Include
+    @Nullable
+    String currentVersion;
+
     @EqualsAndHashCode.Include
     @ToString.Include
     @Nullable
@@ -330,7 +336,9 @@ public class RawPom {
 
     @Nullable
     public String getVersion() {
-        return version == null && parent != null ? parent.getVersion() : version;
+        return version == null ?
+                (parent == null ? currentVersion : parent.getVersion()) :
+                version;
     }
 
     public Pom toPom(@Nullable Path inputPath, @Nullable MavenRepository repo) {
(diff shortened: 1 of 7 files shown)


jkschneider (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
```diff
@@ -66,6 +66,12 @@ public class RawPom {
     @Nullable
     String version;
 
+    // This is an obsolete field that is no longer used, but it is present in some particularly old POMs
+    @EqualsAndHashCode.Include
```
jkschneider, line 70: Not sure it should be included in EqualsAndHashCode

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -55,34 +54,18 @@ public class ResolvedPom {
...
 
     @With
     Iterable<String> activeProfiles;
 
-    public ResolvedPom(Pom requested, Iterable<String> activeProfiles) {
-        this(requested, activeProfiles, emptyMap(), emptyList(), null, emptyList(), emptyList(), emptyList(), emptyList());
-    }
-
-    @JsonCreator
-    ResolvedPom(Pom requested, Iterable<String> activeProfiles, Map<String, String> properties, List<ResolvedManagedDependency> dependencyManagement, @Nullable List<MavenRepository> initialRepositories, List<MavenRepository> repositories, List<Dependency> requestedDependencies, List<Plugin> plugins, List<Plugin> pluginManagement) {
```
jkschneider, line 74: This worries me. Where is the constructor going?

sambsnyd: After this PR there are no remaining invocations of the two-argument form of the constructor within [OpenRewrite](https://app.moderne.io/results/uv5uuVFm2).
So if that's gone the remaining constructor is exactly what lombok would generate, so there is no need to preserve it.

anandfresh (approved):

jkschneider:
Closing this for now in favor of fixing resolution problems. Allowing a partial POM resolution will cause the downstream effect of dependency-manipulating recipes potentially not working correctly.

PR #4141, opened 2024-04-19
preconditions are no longer cached

motu55:
## What's changed?
Precondition Bellwether uses a supplier to fetch a new visitor each time

## What's your motivation?
This causes a bug when a certain condition is met

@sambsnyd 

to use recipes instead of suppliers but the suppliers feel more correct

## Any additional context
see attachement below

### Checklist
- [ x] I've added unit tests to cover both positive and negative cases
- [x ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Preconditions.java
@@ -18,6 +18,10 @@ package org.openrewrite;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.marker.SearchResult;
 
+import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


motu55:
[mvn_multi_module_test.zip](https://github.com/openrewrite/rewrite/files/15036947/mvn_multi_module_test.zip)
this can be used to verify the fix i was not able to run the plugin with a snapshot if the code

motu55:
added a test to reproduce the behavior not sure if it is at the right place, feel free to move or instruct how to move

PR #4143, opened 2024-04-19
ChangeParentPom's property-preservation now checks the resolved new p…

nmck257:
…arent (instead of unresolved new parent pom) when determining what properties will be available

## What's changed?

see title

## What's your motivation?

The new test case would previously fail, having added `junit.version` to the pom's properties

this is mostly a tweak on @sambsnyd 's recent code in this recipe

### Checklist


sambsnyd:
Thanks @nmck257 !

PR #4148, opened 2024-04-23
RemoveRedundantDependencyVersions - adding `onlyIfManagedVersionIs` o…

### LINKED ISSUE #4138: RemoveRedundantDependencyVersions - restore ability to remove newer version pins

Add an option for the functionality removed in this commit:
https://github.com/openrewrite/rewrite/commit/687fbb55709b7d6ee0a648c425d8b0b94fce65d5
(from @sambsnyd )

N/A

```xml
              <project>
                  <groupId>org.sample</groupId>
                  <artifactId>sample</artifactId>
                  <version>1.0.0</version>
                  <parent>
                      <version>2.7.18</version>
                      <relativePath/>
                  </parent>
                  <build>
                      <plugins>
                          <plugin>
                              <groupId>pl.project13.maven</groupId>
                              <artifactId>git-commit-id-plugin</artifactId>
                              <!-- version in pom is 4.9.10, pin is newer-->
                              <version>4.9.11</version>
                          </plugin>
                      </plugins>
                  </build>
              </project>
```

```xml
              <project>
                  <groupId>org.sample</groupId>
                  <artifactId>sample</artifactId>
                  <version>1.0.0</version>
                  <parent>
                      <version>2.7.18</version>
                      <relativePath/>
                  </parent>
                  <build>
                      <plugins>
                          <plugin>
                              <groupId>pl.project13.maven</groupId>
                              <artifactId>git-commit-id-plugin</artifactId>
                          </plugin>
                      </plugins>
                  </build>
              </project>
```

## Any additional context

This (removed) behavior can be useful if you want to strictly rely on parent-provided versions, and implicitly "distrust" the validity of existing pinned versions (or anticipate that they will become a problem eventually as the parent version increases)

Yes -- mainly opening this issue first in case anyone has design input.

The least-intrusive implementation would be a new, additional flag, such as `removeNewerVersions`, which works in conjunction with `onlyIfVersionsMatch`. But the naming/usage starts to get fuzzy -- what if someone sets `onlyIfVersionsMatch` `true` and also `removeNewerVersions` `true`? We can validate those cases out, but, it's unintuitive.

Could also argue that `onlyIfVersionsMatch` is now an inaccurate name, since on `false`, it does not remove non-matching newer versions.

Replacing this field with some kind of `removalStrategy` enum could be nice, with values like `matching`, `older`, `any`, but replacing an arg like that can be disruptive for existing usage, and naming the new arg also takes some good thought (I don't love the value names I just suggested, eg).

Eager for any opinions

nmck257:
…ption and deprecating `onlyIfVersionsMatch`

## What's changed?

see title

## What's your motivation?

Closes #4138

- Open to alternative designs for the new argument
- I can just remove the deprecated bits immediately if preferred
- Not sure if the pseudo-enum is desirable

@sambsnyd had made the previous change which this PR is in response to

### Checklist


anandfresh (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
```diff
@@ -68,6 +79,48 @@ public class RemoveRedundantDependencyVersions extends Recipe {
...
+    }
+
+    // pseudo enum to avoid (de)serialization concerns
+    public static class Comparator {
+        public static final String ANY = "any";
+        public static final String EQ = "eq";
+        public static final String LT = "lt";
+        public static final String LTE = "lte";
+        public static final String GT = "gt";
+        public static final String GTE = "gte";
```
sambsnyd, line 106: It's fine for this to be a real enum

PR #4156, opened 2024-04-29
Modify default behaviour of MavenVisitor::isPluginTag() to search everywhere, not just build

dawngerpony:
## What's changed?

* Modify the default behaviour of `MavenVisitor::isPluginTag()` to search everywhere, not just in the `<build><plugins>` section. Recipes that use this method to modify plugin definitions will now find that their `pluginManagement` sections are being updated too.
* Add a few more tests to verify the new behaviour.

## What's your motivation?

We need this behaviour in our own recipes and we'd like to push it down the stack, because I thought others would find it useful too.

Not any more since the review.

@timtebeek perhaps?

I don't see any reasonable alternatives, other than adding the matching logic to every recipe which needs it.

## Any additional context

None.

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -199,6 +202,19 @@ public void maybeUpdateModel() {
...
+        String groupId, String artifactId) {
+        Cursor cursor = getCursor();
+        for (XPathMatcher matcher : Arrays.asList(PLUGIN_MATCHER, MANAGED_PLUGIN_MATCHER,
+            PROFILE_PLUGIN_MATCHER, PROFILE_MANAGED_PLUGIN_MATCHER)) {
+            if (matcher.matches(cursor) && hasPluginGroupId(groupId) && hasPluginArtifactId(
+                artifactId)) {
+                return true;
+            }
+        }
+        return false;
```
caesar-ralf, line 215: nit-pick: you can convert these to
```
```suggestion
        return Arrays.asList(PLUGIN_MATCHER, MANAGED_PLUGIN_MATCHER,
            PROFILE_PLUGIN_MATCHER, PROFILE_MANAGED_PLUGIN_MATCHER)
            .stream()
            .anyMatch(matcher -> matcher.matches(cursor) && hasPluginGroupId(groupId) && hasPluginArtifactId(
                artifactId));
```

caesar-ralf (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
timtebeek: While that does help the readability, we try to limit our use of the streams API, especially in code paths that might get evaluated often, as the object allocations can hurt performance when running at scale.

timtebeek:
Thanks for getting this started @dawngerpony ! Reminds me of this change we did a while back
- https://github.com/openrewrite/rewrite/pull/3962

I should probably have been consistent at the time to roll that out for plugins as well; thanks for picking that up now.

For dependencies we ended up with updating dependencies wherever they are found by default; with a new option only to disable that behaviour. The motivation being that it's less surprising to users: then want to change a dependency; not have to remember to also change managed dependencies, or plugin dependencies, or profile managed plugin dependencies, and so on... 

I'm thinking we can do the same here: by default update all locations where a plugin might be found, and maybe even leave out the option to limit that until we get specific complaints. To that end I'm wondering if we can simplify the implementation here by updating this specific matcher to use for instance `//plugins/plugin`.
https://github.com/openrewrite/rewrite/blob/d25357f7c1c891567197a9dd0a07e4a7234a0b92/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java#L43

What would be your thoughts to such an approach? I know you've been careful not to make any breaking changes, but in this case I think the gain in functionality is tolerable.

dawngerpony:
> What would be your thoughts to such an approach? I know you've been careful not to make any breaking changes, but in this case I think the gain in functionality is tolerable.

If you're on board then I'm on board! Would you even recommend that we literally just update `isPlugin()` to check the new matcher? That would most certainly modify the default behaviour.

timtebeek:
> If you're on board then I'm on board! Would you even recommend that we literally just update `isPlugin()` to check the new matcher? That would most certainly modify the default behaviour.

That's what I was eyeing this afternoon indeed, as I had trouble thinking of use cases where folks would want to to do one but not the other, without resorting to a custom recipe.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
caesar-ralf: makes total sense, didn’t think about the performance of it

timtebeek (approved):
Thanks a lot! Good to be consistent, with less code to boot.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -33,14 +33,13 @@
 import static org.openrewrite.internal.StringUtils.matchesGlob;
 
 public class MavenVisitor<P> extends XmlVisitor<P> {
+
     static final XPathMatcher DEPENDENCY_MATCHER = new XPathMatcher("/project/dependencies/dependency");
     static final XPathMatcher PLUGIN_DEPENDENCY_MATCHER = new XPathMatcher("/project/*/plugins/plugin/dependencies/dependency");
```
knutwannheden, line 38: Not strictly related to this PR, but should we possibly also here match plugin dependencies in profiles?
```suggestion
    static final XPathMatcher PLUGIN_DEPENDENCY_MATCHER = new XPathMatcher("//plugins/plugin/dependencies/dependency");
```

knutwannheden: Note: The corresponding `isPluginDependencyTag()` method is being used by the `UpgradeDependencyVersion` recipe.

timtebeek: Thanks! Followed up with https://github.com/openrewrite/rewrite/commit/cfed3dcf0b75b2b56215185c4be7cccee00cd36e just now.

PR #4167, opened 2024-05-05
Support ExactVersion selector prefixed with "="

### LINKED ISSUE #4166: UpgradeDependencyVersion should support forcing exact version

As historic legacy, at my workplace internal maven artifact versions are in the format `major.minor-buildnumber`.
I'm trying to use [UpgradeDependencyVersion](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java) to upgrade to an exact specific version `2.0-3`.
But this does not work because it is interpreted as a [hyphenated range](https://docs.openrewrite.org/reference/dependency-version-selectors#hyphenated-range) semver selector. 

A way to force the version to be interpreted as an exact version.

Currently using [ChangeDependencyGroupIdAndArtifactId](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java) as a workaround.
That works, but is not intended for this purpose and results in a recipe validation error "newGroupId OR newArtifactId must be different from before".

jpraet:
Fixes #4166

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/ExactVersion.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/ExactVersion.java
@@ -18,12 +18,20 @@ package org.openrewrite.semver;
 import org.openrewrite.Validated;
 import org.openrewrite.internal.lang.Nullable;
 
+/**
+ * Version selector for matching exact version: either explicitly prefixed with "=",
+ * or implicit default when no other version selectors match.
+ */
 public class ExactVersion extends LatestRelease {
     String version;
 
     public ExactVersion(String pattern) {
         super(pattern);
-        this.version = pattern;
+        if(pattern.startsWith("=")) {
+            this.version = pattern.substring(1);
+        } else {
+            this.version = pattern;
+        }
     }
 
     @Override
@@ -32,6 +40,9 @@ public class ExactVersion extends LatestRelease {
     }
 
     public static Validated<ExactVersion> build(String pattern) {
+        if(pattern.startsWith("=")) {
+            return Validated.valid("exactVersion", new ExactVersion(pattern));
+        }
         String versionOnly;
         int hyphenIndex = pattern.indexOf('-');
         if(hyphenIndex == -1) {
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -1025,10 +1026,11 @@ void dependencyManagementResolvedFromProperty() {
         );
     }
 
-    @Test
-    void upgradeToExactVersion() {
+    @ParameterizedTest
+    @ValueSource(strings = { "3.0.12.RELEASE", "=3.0.12.RELEASE" })
+    void upgradeToExactVersion(String version) {
```
timtebeek, line 1031:
```suggestion
    @Issue("https://github.com/openrewrite/rewrite/issues/4166")
    void upgradeToExactVersion(String version) {
```

rewrite-core/src/main/java/org/openrewrite/semver/ExactVersion.java
```diff
@@ -32,6 +40,9 @@ public boolean isValid(@Nullable String currentVersion, String version) {
     }
 
     public static Validated<ExactVersion> build(String pattern) {
+        if(pattern.startsWith("=")) {
```
timtebeek, line 43:
```suggestion
        if (pattern.startsWith("=")) {
```

rewrite-core/src/main/java/org/openrewrite/semver/ExactVersion.java
```diff
@@ -18,12 +18,20 @@
...
+ * Version selector for matching exact version: either explicitly prefixed with "=",
+ * or implicit default when no other version selectors match.
+ */
 public class ExactVersion extends LatestRelease {
     String version;
 
     public ExactVersion(String pattern) {
         super(pattern);
-        this.version = pattern;
+        if(pattern.startsWith("=")) {
```
timtebeek, line 30:
```suggestion
        if (pattern.startsWith("=")) {
```

timtebeek (approved):
Thanks a lot for the quick work here. I'll apply some minor touch ups and get this merged.

PR #4169, opened 2024-05-06
Suggest alternatives when active recipes are not found

timtebeek:
## What's changed?
When active recipe(s) not found, provide a helpful `Did you mean: ...` of closest match.

## What's your motivation?
Misconfigurations are common and hard to spot, as seen in supporting OSS and even colleagues.
In particular when generated Refaster recipes are used, folks might fail to add the `Recipe` or `Recipes` suffix.
By providing the nearest match as a suggestion folks have an easier way to troubleshoot this themselves.

1. I'd expect calls to `Environment#activateRecipes(Iterable<String> activeRecipes)` to be rare enough to be ok with the streams use there. Outside perhaps tests I'd expect that to be called once per recipe run. Calling that out for review.
2. We already have a dependency on commons-lang3, which has the `getLevenshteinDistance` we use here. Didn't want to add a new dependency just for that, but in theory it's deprecated. In practice we're unlikely to move to a new major version any time soon.

We could use a different method of providing alternatives.

## Any additional context
This came up twice today; once in OSS and once internally; combined time savings would already have paid this off today.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.config;
 
+import org.apache.commons.lang3.StringUtils;
 import org.openrewrite.Contributor;
 import org.openrewrite.Recipe;
 import org.openrewrite.RecipeException;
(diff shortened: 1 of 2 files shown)


timtebeek:
Also discussed on Slack; I'll take ["makes sense"](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1715018219021289?thread_ts=1715012235.182139&cid=C01VADFPJQZ) as approval to merge.

PR #4174, opened 2024-05-08
ChangePackage fails to migrate annotation argument enum in package-info

pativa:
## What's changed?
I found [this issue](https://github.com/openrewrite/rewrite/issues/3200) describing problems with migrating annotations in `package-info.java`. We seem to still have this problem.

In this MR, I have added a (failing) test showing our issue.

Our actual issue is regarding applying jakarta migration for this class:

```
@javax.xml.bind.annotation.XmlSchema(namespace = "http://example.com/schema", elementFormDefault = javax.xml.bind.annotation.XmlNsForm.QUALIFIED)
package org.example;
```

We have made it work using `org.openrewrite.java.ChangeType` for our migration, but would prefer if `org.openrewrite.java.ChangePackage` worked for this.

Please note that this does not fix the issue, only adds a test case to it :-)

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
@@ -1582,16 +1582,16 @@ class ChangePackageTest implements RewriteTest {
             """
               @org.openrewrite.MyAnnotation(myEnum = org.openrewrite.MyEnum.BAR)
               package com.acme;
-               """,
+              """,
             """
               @org.openrewrite.test.MyAnnotation(myEnum = org.openrewrite.test.MyEnum.BAR)
               package com.acme;
-               """,
+              """,
             spec -> spec.afterRecipe(cu -> {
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot! Runnable tests are my favorite way to show a bug; did you already try to step through with a debugger to see if you could pinpoint the source of the issue?

pativa:
No, I haven't found the source of the issue. It seems to find that the types are in use, but the changes are not applied.

I can maybe check it out later if you haven't already found it!

pativa:
The fix was really simple when I finally found what was going on 🙂 

The `visitPackage` method was not calling `super.visitPackage()`, so annotations on the package statement weren't being updated.

timtebeek (approved):
Yes makes sense when you see the issue; thanks for digging in and providing the fix!

pativa:
I'm not really sure of why, but some tests were failing if `super.visitPackage()` is called.

Instead, I updated the proposed fix to only update the annotations in the same way as is done inside the super method. This seems to work with all the existing tests at least.

PR #4178, opened 2024-05-08
YAML: Support lists of lists

### LINKED ISSUE #4176: YAML: Unable to parse lists of lists
## The Problem
In YAML documents it is allowed to define lists of lists.
```yaml
listOfLists:
  - - a
    - b
```

**OpenRewrite is unable to parse this type of lists correctly.**

---

To better understand the meaning of this document, here is the same document described in JSON:
```json
{
  "listOfLists": [
    ["a", "b"]
  ]
}
```

I am using the OpenRewrite Gradle Plugin v6.12.0

I'm using the Gradle Plugin within a single module project.

To reproduce the problem, create a yaml file containing the following snippet and run the plugin.
```yaml
listOfLists:
  - - a
    - b
```

Another option is to use the unit tests in the rewrite-yaml module e.g. in the `YamlParserTest.java`.
Add the following test and execute it.

```java
@Test
void listOfLists() {
    rewriteRun(
      yaml(
        """
          listOfLists:
            - - a
            - b
          """
      )
    );
}
```

The yaml should be parseable. Or: the test should succeed.

## What did you see instead?
Parsing of the yaml fails due to:
`When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself.`

(This is because, the produced source code contains additional `-` signs.)

```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "file.yaml":
index edb1cf7..8cc0b6c 100644
--- a/file.yaml
+++ b/file.yaml
@@ -1,4 +1,4 @@ 
 root:
     listOfLists:
-      - - a
+      - - - a
         - b
\ No newline at end of file
] 
expected: 
  "root:
      listOfLists:
        - - a
          - b"
 but was: 
  "root:
      listOfLists:
        - - - a
          - b"
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at org.openrewrite.test.RewriteTest.assertContentEquals(RewriteTest.java:617)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:312)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:133)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:128)
	at org.openrewrite.yaml.YamlParserTest.listOfLists(YamlParserTest.java:70)
```

I've already identified the root cause and will create a PR containing the fix later today.

pdulich:
YAML supports lists of lists. They are not common but sometimes required, for example, in some AWS CloudFormation templates. Currently, the YamlParser doesn't correctly parse these lists. It adds additional hyphens.
```
Original     ->     After parsing
root:               root:
  - - a               - - - a
    - b                 - b
```
This happens for every line that contains multiple hyphens (like for the 'a' line in the example above).
That's because the parser interprets these hyphens twice. Once as a SequenceEntry and once as prefix for the last block of that line (in this example, the scalar 'a').

The bug is in SequenceBuilder.push. This method builds the prefix of a given block by using the substring between the FIRST hyphen and the block itself. So if there is more than one hyphen in that row, the additional hyphens will be included in the prefix, although they are already represented in the LST as SequenceEntries.

Fix it by using the substring between the LAST hyphen and the block itself. To do this, extend the functionality of the commentAwareIndexOf method, so that it can find the first or the last index. Don't create an additional method as not to duplicate the logic. Also use a dedicated enum because it is cleaner and much more expressive than using a boolean flag.

Fixes #4176

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -129,7 +129,7 @@ public class YamlParser implements org.openrewrite.Parser {
                 switch (event.getEventId()) {
                     case DocumentEnd: {
                         assert document != null;
-                        if(blockStack.size() == 1 && blockStack.peek() instanceof ScalarBuilder) {
+                        if (blockStack.size() == 1 && blockStack.peek() instanceof ScalarBuilder) {
                             // The yaml document consists of a single scalar value not in a mapping or sequence
                             ScalarBuilder builder = (ScalarBuilder) blockStack.pop();
                             lastEnd = builder.getLastEnd();
@@ -174,7 +174,7 @@ public class YamlParser implements org.openrewrite.Parser {
 
                             lastEnd = lastEnd + mappingStartEvent.getAnchor().length() + fmt.length() + 1;
                             fmt = reader.readStringFromBuffer(lastEnd, event.getEndMark().getIndex());
-                            int dashPrefixIndex = commentAwareIndexOf('-', fmt, FindIndexStrategy.FIRST);
+                            int dashPrefixIndex = commentAwareIndexOf('-', fmt);
                             if (dashPrefixIndex > -1) {
                                 fmt = fmt.substring(0, dashPrefixIndex);
                             }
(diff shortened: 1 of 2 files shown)


pdulich:
I was unable to push the changes onto a branch in the official repository but I had to create a fork for that. 
Is this intended?

timtebeek (approved):
Awesome! It's not often that we see parser fixes contributed ; good to see you worked it out and made the minimal change to get this fixed. Also appreciate how you included your reasoning in the PR description; made this a joy to review.

Issue #4182, opened 2024-05-09
ChangeType erroneously changes package of unrelated type


kennytv:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.26.0-SNAPSHOT (via `org.openrewrite.recipe:rewrite-recipe-bom:latest.integration`)
- Gradle plugin v6.13.0

## How are you running OpenRewrite?

Very simple setup with just two classes and a single recipe via the Gradle plugin. Ignoring Java target and snapshot repository declaration, the build file looks like this:
```kts
plugins {
    base
    `java-library`
    id("org.openrewrite.rewrite") version "6.13.0"
}

dependencies {
    implementation(platform("org.openrewrite.recipe:rewrite-recipe-bom:latest.integration"))
}

rewrite {
    activeRecipe(
        "test.ChangeTypeTest",
    )
}
```

## What is the smallest, simplest way to reproduce the problem?
Create two classes in the same package, where the class that should *not* be changed makes some kind of reference to a class that *should* be changed (e.g. via method parameters/return values, or generic type parameters):
```java
package com.test; // This will erroneously be changed

public final class Test {

    public A test() { // Any reference to A is required for the issue
        return new A();
    }
}
```
```java
package com.test;

public final class A { // This will correctly be moved to the subpackage
}
```

The recipe used:
```yml
---
type: specs.openrewrite.org/v1beta/recipe
name: test.ChangeTypeTest
recipeList:
  - org.openrewrite.java.ChangeType:
      oldFullyQualifiedTypeName: com.test.A
      newFullyQualifiedTypeName: com.test.sub.A
```

## What did you expect to see?
`A` to be moved without further changes to `Test`

## What did you see instead?
The package declaration of `Test` changes to `com.test.sub`. It also fails to add an import for `A`, but that might just be a result of the relevant issue here

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Sure. I am vaguely familiar with visitor patterns through ASM, although I'd imagine an active contributor might be able to fix this faster than telling me what to look for

timtebeek:
That's a very odd case indeed, thanks for reporting it in, and offering to help explore a fix. I wonder if it replicates with our unit test framework as well. 🤔

kennytv:
[...]

Issue #2528, opened 2022-12-14
ChangeType: rename java file for non-public class too if the java file name matches the class name


ge0ffrey:
Given

`src/main/java/a/A1.java`
Content:
```
package a;

class A1 {} // non-public
```

with a ChangeType from `a.A1` to `a.A2`, it will result in:

`src/main/java/a/A1.java`
Content:
```
package a;

class A2 {} // non-public
```

The resulting content is fine, but **the filename should have renamed to `A2.java`** too.

Even despite the fact that `class A1` isn't public, because the filename `A1.java` matched the original classname `A1`.

ge0ffrey:
One could argue that unlike for public classes, there is no need that `A2`, after the change is in the file `A2.java`. However in practice, this is what users _pretty much always_ want, given that the rename of the filename only happens if (and only if) the `A1` was in the file `A1.java` to begin with.

For the theoretical case that that _pretty much always_, one could argue to introduce it as a boolean configuration property `renameFileOfNonPublicClassWithSameName` but I think that's overkill (YAGTNI), so I 'd leave that out until someone asks for it. (And I predict no one will every ask for it.)

Issue #1904, opened 2022-06-08
ChangeType should also change the source path and constructors when modifying the class definition


tkvangorder:
The ChangeType recipe should also rename the source file and any constructors when it modifies the class definition (the `ignoreDefinition` is `false`) and the Class is the first, top-level class of a compilation unit.

As an example: 

Given the recipe: 

```yaml
type: specs.openrewrite.org/v1beta/recipe
name: com.example.RenameSomething:
  - org.openrewrite.java.ChangeType
      oldFullyQualifiedTypeName: com.example.OldClass
      newFullQualfiedTypeName: org.example.NewClass
```

The expectation would be for the SourceFile to have a modified path of `org.exampl.NewClass.java`

PR #4184, opened 2024-05-10
Don't modify package declaration of siblings in ChangeType

kennytv:
## What's changed?
`ChangeClassDefinition` in `ChangeType` previously changed package declaration of sibling types, this PR restricts these to only the classes that should actually be changed

## What's your motivation?
Closes https://github.com/openrewrite/rewrite/issues/4182

~~I'm not exactly sure as to whether this is the proper way to do it (tracking required modification in JavaSourceFile visits), but at the very least this should show what the root cause is. And I suppose this breaks the immutability rule~~ should be addressed now

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
@@ -509,7 +509,7 @@ public class ChangeType extends Recipe {
 
                 Path newPath = Paths.get(oldPath.replaceFirst(oldFqn, newFqn));
                 if (updatePath(cu, oldPath, newPath.toString())) {
-                    getCursor().putMessage("UPDATE_PATH", true);
+                    getCursor().putMessage("UPDATE_PACKAGE", true);
                     cu = cu.withSourcePath(newPath);
                 }
                 return super.visit(cu, ctx);
@@ -532,8 +532,8 @@ public class ChangeType extends Recipe {
 
         @Override
         public J.Package visitPackage(J.Package pkg, ExecutionContext ctx) {
-            Boolean updatePath = getCursor().pollNearestMessage("UPDATE_PATH");
-            if (updatePath != null && updatePath) {
+            Boolean updatePackage = getCursor().pollNearestMessage("UPDATE_PACKAGE");
+            if (updatePackage != null && updatePackage) {
                 String original = pkg.getExpression().printTrimmed(getCursor()).replaceAll("\\s", "");
                 if (original.equals(originalType.getPackageName())) {
                     JavaType.FullyQualified fq = TypeUtils.asFullyQualified(targetType);
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
@@ -491,6 +491,7 @@ public class ChangeType extends Recipe {
         private final JavaType.Class originalType;
         private final JavaType.Class targetType;
         private final MethodMatcher originalConstructor;
+        private boolean updatePath;
 
         private ChangeClassDefinition(String oldFullyQualifiedTypeName, String newFullyQualifiedTypeName) {
             this.originalType = JavaType.ShallowClass.build(oldFullyQualifiedTypeName);
@@ -508,7 +509,8 @@ public class ChangeType extends Recipe {
                 String newFqn = fqnToPath(targetType.getFullyQualifiedName());
 
                 Path newPath = Paths.get(oldPath.replaceFirst(oldFqn, newFqn));
-                if (updatePath(cu, oldPath, newPath.toString())) {
+                updatePath = updatePath(cu, oldPath, newPath.toString());
+                if (updatePath) {
                     cu = cu.withSourcePath(newPath);
                 }
                 return super.visit(cu, ctx);
@@ -532,7 +534,7 @@ public class ChangeType extends Recipe {
         @Override
         public J.Package visitPackage(J.Package pkg, ExecutionContext ctx) {
             String original = pkg.getExpression().printTrimmed(getCursor()).replaceAll("\\s", "");
-            if (original.equals(originalType.getPackageName())) {
+            if (updatePath && original.equals(originalType.getPackageName())) {
                 JavaType.FullyQualified fq = TypeUtils.asFullyQualified(targetType);
                 if (fq != null) {
                     if (fq.getPackageName().isEmpty()) {
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
@@ -491,7 +491,6 @@ public class ChangeType extends Recipe {
         private final JavaType.Class originalType;
         private final JavaType.Class targetType;
         private final MethodMatcher originalConstructor;
-        private boolean updatePath;
 
         private ChangeClassDefinition(String oldFullyQualifiedTypeName, String newFullyQualifiedTypeName) {
             this.originalType = JavaType.ShallowClass.build(oldFullyQualifiedTypeName);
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -491,6 +491,7 @@ private static class ChangeClassDefinition extends JavaIsoVisitor<ExecutionConte
         private final JavaType.Class originalType;
         private final JavaType.Class targetType;
         private final MethodMatcher originalConstructor;
+        private boolean updatePath;
```
timtebeek, line 494: We generally recommend to [use cursor messaging to pass state](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#use-cursor-messaging-instead-of-execution-context-messaging) between different methods of a visitor. Mutable fields might linger with old values, whereas cursor messages follow the visitor pattern, and get reset as needed.

timtebeek (commented):
Thanks a lot for getting this started. Helpful to know we might now inadvertently make changes, and a good start at getting that resolved. I've added a note about the use of cursor messaging over mutable fields; I hope that's clear enough. If not do let us know!

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
kennytv: Changed - polling when checking should be the correct behavior, though I'm not 100% on that

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -509,6 +509,7 @@ public J visit(@Nullable Tree tree, ExecutionContext ctx) {
 
                 Path newPath = Paths.get(oldPath.replaceFirst(oldFqn, newFqn));
                 if (updatePath(cu, oldPath, newPath.toString())) {
+                    getCursor().putMessage("UPDATE_PATH", true);
```
timtebeek, line 512:
```suggestion
                    getCursor().putMessage("UPDATE_PACKAGE", true);
```

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -531,17 +532,20 @@ private boolean updatePath(JavaSourceFile sf, String oldPath, String newPath) {
...
-                if (fq != null) {
-                    if (fq.getPackageName().isEmpty()) {
-                        getCursor().putMessageOnFirstEnclosing(J.CompilationUnit.class, "UPDATE_PREFIX", true);
-                        //noinspection DataFlowIssue
-                        return null;
-                    } else {
-                        String newPkg = targetType.getPackageName();
-                        return JavaTemplate.builder(newPkg).contextSensitive().build().apply(getCursor(), pkg.getCoordinates().replace());
+            Boolean updatePath = getCursor().pollNearestMessage("UPDATE_PATH");
+            if (updatePath != null && updatePath) {
```
timtebeek, line 536:
```suggestion
            Boolean updatePackage = getCursor().pollNearestMessage("UPDATE_PACKAGE");
            if (updatePackage != null && updatePackage) {
```

timtebeek (commented):
Looking good ; quick suggestion to rename the message key and variable, as path can mean the source file path too, which might be confusing.

timtebeek (approved):
Approved already; let me know how you feel about those suggestions and then we can get this merged.

kennytv:
Should be all good 👍

timtebeek:
Thanks a lot for the quick turn around on this; really helpful to get this sorted. I'll merge as soon as the tests pass and from there it'll become available in [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions) until the next release, typically in a week or two.

timtebeek:
Hmm; seems two other tests have started failing; This might need some work still. 🤔

https://github.com/openrewrite/rewrite/blob/2988966b8675e008598f4942f8ca47aa1652023e/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java#L1310-L1332

https://github.com/openrewrite/rewrite/blob/2988966b8675e008598f4942f8ca47aa1652023e/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java#L1426-L1449

kennytv:
Hmm that's strange, they run and pass locally, both individually through IntelliJ and the full module via `gradlew :rewrite-java-test:test`

timtebeek:
I did see them fail for me locally; I've triggered the tests to run again to see if that shows any difference.

kennytv:
After switching machines I got the same lol, a fresh clone might have done it as well. In any case, I'm not sure I understand the `spec -> spec.path("x/y/Z.java") ` lines. Docs say this marks the path *after* running the recipe, but I'm already getting `x/y/Z.java` as the source path in `ChangeType` - is that intended behavior? Changing the spec.path call to the origin path makes the test pass again

timtebeek:
Oh wow, I'd never noticed the Java docs on those methods; they seem to have been added deliberately, although I'd always taken those to mean the source paths before the recipe is run.
https://github.com/openrewrite/rewrite/blob/212296420cd490706a23e1b11ee6c85a2813df61/rewrite-test/src/main/java/org/openrewrite/test/SourceSpec.java#L108-L124

We have other mechanisms to verify the source path is changed after a recipe run, which hints at the JavaDoc above being wrong.

https://github.com/openrewrite/rewrite/blob/212296420cd490706a23e1b11ee6c85a2813df61/rewrite-test/src/test/java/org/openrewrite/RenameFileTest.java#L41-L43

So with that, the `spec -> spec.path("a/b/NoMatch.java")` and `spec -> spec.path("x/y/Z.java")` should set the paths on the input source files, not their moved equivalents.

The `onlyChangeTypeWithoutMatchedFilePath` test then verifies that if the filename does not match any of the contained classes, that we do not change the filename. I think that's valid and probably best to retain to avoid surprises. What do you think? The linked issue is this one
- https://github.com/openrewrite/rewrite/issues/1904

The `changePathOfNonPublicClass` test verifies that we do actually change the path, even when the class is not public, as defined in this linked issue:
- https://github.com/openrewrite/rewrite/issues/2528

With that I think it would be best if we don't change the behavior of any of the existing cases, and maybe even fix the confusing JavaDocs. What are your thoughts on the same?

kennytv:
That makes sense, thanks for the help! I pushed a commit here with how I'd go about fixing them, let me know if that should be a separate PR. The fix for the jd and the `changePathOfNonPublicClass` test are pretty straight forward, just for `onlyChangeTypeWithoutMatchedFilePath` I split the test for the non-matching class into its own call, so that the source path of the `a.b.Original` case stays correct

timtebeek:
Thanks for the quick work; I had a slightly different idea come to mind after I'd already replied above; I should have updated my comment but was in the middle of other work.

Before any of your changes, we handled the cases where there's multiple classes in a single compilation unit (file) and changed their package and source path if the recipe indicates that's necessary. The `NoMatch.java` test highlighted a separate case where neither non-public class matched the filename, and where the filename should thus not change for that test (at least up to now).

The issue you've reported, and are tackling here is a different variation: there's again two classes in the same package, but in different compilation units (files). That's where the recipe up to now appears to have been making incorrect changes, based on the new test you added where the sibling in a different file was changed as well.

I think ideally we don't modify the tests we had before, and keep their behavior the same (unless we feel that's incorrect), and instead work out how to get the new test to pass by handling the above cases separately: sibling classes in the same compilation unit versus in different compilation units.

And note that I'm certainly not trying to complicate the work you're doing! It's much appreciated; we're both learning what the recipe did before, and how that relates to what we'd like it to do going forward, and hope we agree on the approach.

kennytv:
Ah I see now. *Technically* the provided and expected results aren't invalid Java code, but in practice you get misplaced class files that don't match their package declaration, which doesn't seem to be part of the linked issues (the provided class in `changePathOfNonPublicClass`, and the result in `onlyChangeTypeWithoutMatchedFilePath` are misplaced).

An alternative to checking the source path for the sibling fix (as pushed now) would be to go through the classes in JavaSourceFile to check their fully qualified names, ignoring the actual source path. That should work and makes all tests pass, but I'm not sure of other implications

timtebeek (approved):
Thanks a lot for working through all these iterations. I think this now works as we'd like it to.

PR #4186, opened 2024-05-10
bug: Handle NoClassDefFoundError when loading YAML recipes with missing recipe dependencies

bryceatmoderne:
## What's changed?
Added logic to catch NoClassDefFoundError errors when loading YAML recipes with missing recipe dependencies

## What's your motivation?
NoClassDefFoundError errors were previously not handled compromising troubleshooting.

Potential for regressed behaviour.

@timtebeek 

No

## Any additional context
No

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
@@ -278,28 +278,42 @@ public class YamlResourceLoader implements ResourceLoader {
             }
         } else if (recipeData instanceof Map) {
             Map.Entry<String, Object> nameAndConfig = ((Map<String, Object>) recipeData).entrySet().iterator().next();
+            String recipeName = nameAndConfig.getKey();
+            Object recipeArgs = nameAndConfig.getValue();
             try {
-                if (nameAndConfig.getValue() instanceof Map) {
+                if (recipeArgs instanceof Map) {
                     try {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for taking this on! I take it you've tested this particular scenario then as well?

anandfresh (approved):

timtebeek:
Had another look; when I modify the test case here to use an invalid class, it seems we're not entering the new catch block where I had expected that: 
https://github.com/openrewrite/rewrite/blob/03c458404073082275a752063887c188d213afbe/rewrite-core/src/test/java/org/openrewrite/config/YamlResourceLoaderTest.java#L303

image

Also note that the new catch block is only added for Yaml resources that contain more than one document (or at least a `---` separator), whereas it might be quite common to find resources without such a block and handle missing classes there as well.

Perhaps we should still add a test if possible to ensure we're catching in the right place and not missing certain cases.

bryceatmoderne:
> Perhaps we should still add a test if possible to ensure we're catching in the right place and not missing certain cases.
I've added exception handling for the `String` `recipeData` case. I've also added unit testing as suggested. The tests validate the behaviour when a `NoClassDefFoundError` is thrown when attempting to instantiate a recipe, however it does so in a contrived manner that is not analogous to how the Moderne customer encountered it.

anandfresh (approved):

kmccarp (approved):

PR #4189, opened 2024-05-12
org.openrewrite.java.ChangePackage precise package match

### LINKED ISSUE #4188: org.openrewrite.java.ChangePackage precise match
https://docs.openrewrite.org/recipes/java/changepackage

When I use the above recipe to move/rename/change the types in package `com.test.model.staff` to `com.test.model.data.staff`, it picks up the types in package `com.test.model.staffdetails` too.

How can I restrict it to pick up only files from package `.staff` and **not** from `.staffdetails` ?

Dinozavvvr:
The mistake was in `isTargetRecursivePackageName(String packageName)` not in package name match in `preVisit(..)`

When I tried to fix matching condition I found that `typeInNestedPackageInheritingFromTypeInBasePackage` test fails and then I found this

Fixes #4188
- #4188

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
@@ -400,6 +400,7 @@ class ChangePackageTest implements RewriteTest {
     }
 
     @Test
+    @Issue("https://github.com/openrewrite/rewrite/pull/4189")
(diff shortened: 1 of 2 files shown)


anandfresh (approved):

timtebeek (approved):
Thanks a lot for your persistence in getting this fixed! I love how small the actual change to the recipe is, and how thorough you have tested this to ensure we're doing the right thing in various scenarios. Much appreciated!

Issue #4196, opened 2024-05-17
Bug adding import to a list of imports containing duplicate imports


nguyenhoan:
## What version of OpenRewrite are you using?
I am using the latest version from the tip of GitHub's `main` branch.

## What is the smallest, simplest way to reproduce the problem?
Add the below test case to `org.openrewrite.java.style.ImportLayoutStyleTest`
```java
    @Test
    void testAddImport() {
        ImportLayoutStyle style = new ImportLayoutStyle(
          Integer.MAX_VALUE, Integer.MAX_VALUE, Collections.emptyList(), Collections.emptyList());
        JRightPadded<J.Import> import1 = new JRightPadded<>(
          new J.Import(
            randomId(),
            Space.EMPTY,
            Markers.EMPTY,
            new JLeftPadded<>(Space.SINGLE_SPACE, true, Markers.EMPTY),
            TypeTree.build("pkg.Clazz.MEMBER_1").withPrefix(Space.SINGLE_SPACE),
            null),
          Space.EMPTY,
          Markers.EMPTY);
        JRightPadded<J.Import> import2 = new JRightPadded<>(
          new J.Import(
            randomId(),
            Space.EMPTY,
            Markers.EMPTY,
            new JLeftPadded<>(Space.SINGLE_SPACE, true, Markers.EMPTY),
            TypeTree.build("pkg.Clazz.MEMBER_1").withPrefix(Space.SINGLE_SPACE),
            null),
          Space.EMPTY,
          Markers.EMPTY);
        J.Import importToAdd = new J.Import(
          randomId(),
          Space.EMPTY,
          Markers.EMPTY,
          new JLeftPadded<>(Space.SINGLE_SPACE, true, Markers.EMPTY),
          TypeTree.build("pkg.Clazz.MEMBER_2").withPrefix(Space.SINGLE_SPACE),
          null);
        assertThat(style.addImport(List.of(import1, import2), importToAdd, null, Collections.emptyList()))
          .containsExactlyInAnyOrder(import1, import1, new JRightPadded<>(importToAdd, Space.EMPTY, Markers.EMPTY));
    }
```

## What did you expect to see?
```java
containsExactlyInAnyOrder(import1, import1, new JRightPadded<>(importToAdd, Space.EMPTY, Markers.EMPTY))
```

## What did you see instead?
```java
containsExactlyInAnyOrder(import1, import1)
```

PR #4197, opened 2024-05-17
Fix bug adding import to a list of imports containing duplicate imports.

nguyenhoan:
## What's changed?
Problem:
- The presence check does not take the member of static imports into consideration
- The comparison by size cannot handle the case where the original imports contain duplicated imports making its the size be equal to that of the ideally ordered, distinct imports even after adding a new import

Fix: check the presence using the fully qualified names of imports.

## What's your motivation?
Fix https://github.com/openrewrite/rewrite/issues/4196

### Checklist


timtebeek (approved):
Thanks for working out the fix! In hind sight comparing the original and ideal number of imports seems strange. I like your suggestion to explicitly check for presence better, thanks!

nguyenhoan:
Thank you!

PR #4202, opened 2024-05-21
Fix documentation discrepancy

ryanwalker:
Change overwrite example to 'true'

## What's changed?
I just found in the `AddProperty` gradle recipe that the docs had a string example where it should have been boolean.
Here's the current docs: https://docs.openrewrite.org/recipes/gradle/addproperty
image

Overwrite example should be `true` or `false`, not `Enable the gradle build cache`


shanman190 (approved):
Good catch!

PR #4208, opened 2024-05-23
Parse troublesome Yaml with new workaround in workaround for underying parser bug

timtebeek:
## What's changed?
Prevent `IndexOutOfBoundsException` with new workaround for issues in the underlying `YamlParser` events.

## What's your motivation?
Received reports about failures with a particular Yaml snippet; with this exact snippet you hit the new conditional.
Even if you change strings in the Yaml snippet, you will not hit the new conditional; Fun times to be had tracing that down.

Ideally we report this upstream and see it resolved there.

## Any additional context
The earlier workaround was introduced in these two commits to work around bugs with the underlying Yaml parser we use.
- https://github.com/openrewrite/rewrite/pull/4178
- https://github.com/openrewrite/rewrite/pull/4187

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -384,12 +384,8 @@ public class YamlParser implements org.openrewrite.Parser {
     If the dashes of the sequence do not have an indentation, the end mark will point to the character AFTER the dash.
     */
     private boolean shouldUseYamlParserBugWorkaround(SequenceStartEvent event) {
-        int index = event.getStartMark().getIndex();
-        if (index > event.getStartMark().getBuffer().length - 1) {
-            return false;
-        }
-        int startChar = event.getStartMark().getBuffer()[index];
-        int endChar = event.getEndMark().getBuffer()[event.getEndMark().getIndex()];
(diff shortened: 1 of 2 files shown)


timtebeek:
@pdulich you might enjoy this one; a very specific Yaml snippet managed to trigger a bug in the workaround for the parser bug you discovered earlier. Those parser events sure seem off at times.

pdulich:
Thanks for the report. I already know the cause. 

There are many other problems with the parser. Therefore, I'm completely rewriting it. 

In the meantime I can push a bugfix for this issue.

timtebeek:
Thanks a lot! This PR already adds a minimal workaround that might cause the least amount of conflicts; are you ok with me merging this in the meantime or do you see another quick option?

pdulich:
Oh haha. Sorry. I'm at work and only had a quick look on your comment. I didn't realize that this is a PR and not an issue. 

If I remember correctly, the real bugfix is to replace getIndex() by getPointer().

timtebeek:
Ah perfect, trying that now; passed locally with just that replacement, thanks!

timtebeek:
> There are many other problems with the parser. Therefore, I'm completely rewriting it.

Could you tell us a bit more about your plans here? It can at times be challenging to review large sweeping changes, as much as the effort you'd put in might be appreciated. Just want to be sure we don't end up with a PR that's hard to see through.

PR #4217, opened 2024-05-28
Add recipe to replace Annotation

MBoegers:
## What's changed?
Add a Recipe which replaces an Annotation with a JavaTemplate.

## What's your motivation?
While writing a cross Framework Migration from TestNG to JUnit Jupiter I implemented the replacement of annotations repeatedly.
This could become handy for other cross Framework Migrations as well.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
@@ -17,32 +17,28 @@ package org.openrewrite.java;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.openrewrite.*;
-import org.openrewrite.java.search.UsesType;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaCoordinates;
 import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.TypeUtils;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class ReplaceAnnotation extends Recipe {
-    @Option(displayName = "Annotation pattern to replace",
-            description = "An annotation pattern, expressed as a method pattern to replace.",
+
+    @Option(displayName = "Annotation to replace",
+            description = "An annotation matcher, expressed as a method pattern to replace.",
             example = "@org.jetbrains.annotations.NotNull(\"Test\")")
     String annotationPatternToReplace;
+
     @Option(displayName = "Annotation template to insert",
             description = "An annotation template to add instead of original one, will be parsed with `JavaTemplate`.",
-            example = "@NonNull")
+            example = "@org.jetbrains.annotations.NotNull(\"Null not permitted\")")
     String annotationTemplateToInsert;
-    @Option(displayName = "Type of inserted Annotation",
-            description = "The fully qualified class name of the annotation to insert.",
-            example = "lombok.NonNull")
-    String annotationFQN;
-    @Option(displayName = "Templates Artifact id",
-            description = "The Maven artifactId to load the inserted annotations type from, defaults to JDK internals.",
-            example = "lombok",
-            required = false)
-    String artifactId;
 
     @Override
     public String getDisplayName() {
@@ -57,14 +53,9 @@ public class ReplaceAnnotation extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(new UsesType<>(annotationFQN, false),
-                new ReplaceAnnotationVisitor(
-                        new AnnotationMatcher(annotationPatternToReplace),
-                        JavaTemplate.builder(annotationTemplateToInsert)
-                                .imports(annotationFQN)
-                                .javaParser(JavaParser.fromJavaVersion().logCompilationWarningsAndErrors(true)
-                                        .classpath(artifactId))
-                                .build()));
+        return new ReplaceAnnotationVisitor(
+                new AnnotationMatcher(annotationPatternToReplace),
+                JavaTemplate.builder(annotationTemplateToInsert).javaParser(JavaParser.fromJavaVersion()).build());
     }
 
     public static class ReplaceAnnotationVisitor extends JavaIsoVisitor<ExecutionContext> {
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
```diff
@@ -0,0 +1,97 @@
...
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class ReplaceAnnotation extends Recipe {
+    @Option(displayName = "Annotation pattern to replace",
+            description = "An annotation pattern, expressed as a method pattern to replace.",
```
knutwannheden, line 30: I was going to request a link to the corresponding documentation, but couldn't find it. @mike-solomon do you know if we've documented annotation matchers?

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
```diff
@@ -0,0 +1,97 @@
...
+            example = "@org.jetbrains.annotations.NotNull(\"Test\")")
+    String annotationPatternToReplace;
+    @Option(displayName = "Annotation template to insert",
+            description = "An annotation template to add instead of original one, will be parsed with `JavaTemplate`.",
+            example = "@NonNull")
```
knutwannheden, line 35: I think it would make sense with an example that includes a parameter, to make it clear that this is supported. Additionally, since a parameter could require a reference to a Java class, I think it would make sense for the JavaTemplate here to be required to be fully context-free and thus not rely on any imports being registered with the JavaTemplate. As a result I imagine the `annotationFQN` parameter can be dropped (didn't look at the code). Then we can apply the `ShortenFullyQualifiedNames` visitor to get imports added.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
```diff
@@ -0,0 +1,97 @@
...
+    String annotationTemplateToInsert;
+    @Option(displayName = "Type of inserted Annotation",
+            description = "The fully qualified class name of the annotation to insert.",
+            example = "lombok.NonNull")
+    String annotationFQN;
+    @Option(displayName = "Templates Artifact id",
+            description = "The Maven artifactId to load the inserted annotations type from, defaults to JDK internals.",
```
knutwannheden, line 42: We might have to add what classpath this gets loaded from, as that might not be clear to the user.

I am in fact wondering how we can get that to work unless the recipe is being used by another recipe library which provides the dependency.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>(annotationFQN, false),
+                new ReplaceAnnotationVisitor(
+                        new AnnotationMatcher(annotationPatternToReplace),
+                        JavaTemplate.builder(annotationTemplateToInsert)
+                                .imports(annotationFQN)
+                                .javaParser(JavaParser.fromJavaVersion().logCompilationWarningsAndErrors(true)
```
knutwannheden, line 65: We should remove the logging of errors and warnings.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
knutwannheden: It would be interesting if the dependency could be demand-loaded using the Maven settings. Not sure if that is feasible or desirable. What is your take on this, @timtebeek ? We could also maintain (and document) that the recipe either has to be called from a recipe jar, which itself supplies the dependency or that the result will end up containing missing type attribution. But I am not sure if `ShortenFullyQualifiedTypeReferences` is going to like that.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
mike-solomon: I’m not familiar with the term annotation matcher - but maybe it’s similar to a method pattern? If not, I don’t think we have any docs on it. 

https://docs.openrewrite.org/reference/method-patterns

knutwannheden: Thanks for confirming. It is indeed very similar in concept. But rather than matching method calls (or declarations) its purpose is to match annotations.

MBoegers: Yes it is similar and are both based on AspectJ Pointcut expressions. Would it be sufficient to create a documentation issue for AnnotationMatcher and mention this description?

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
MBoegers: I tried executing `ShortenFullyQualifiedNames` with `doAfterVisit(new ShortenFullyQualifiedTypeReferences().getVisitor());` earlier, but this didn't make the expected changes. I'll try again and post a gist if not working.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
MBoegers: wouldn't it be sufficient if the project to migrate defines the needed dependency? As I understood, if defined in the dependency management the artifact would be on the classpath and all `JavaParser` are happy.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
MBoegers: Yes. Development artifact..

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
knutwannheden: IIRC the class declares a static utility method. We should find some uses of it in rewrite-static-analysis.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
knutwannheden: Normally those classpath entries won't be available. I am assuming an exception is when the recipe is listed in the rewrite.yml file of the project where you run the migration, but I am a bit unsure about that.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
knutwannheden: Yes, I believe we can just change the description now to mention that it is an annotation matcher and then when we get around to documenting that we can insert a link here.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
MBoegers: Tried it, it works in the tests without defining the artifact explicitly.

rewrite-java/src/main/java/org/openrewrite/java/ReplaceAnnotation.java
MBoegers: With `ShortenFullyQualifiedTypeReferences#modifyOnly` the recipe looks better now, thanks. But now the tests are failing because `ShortenFullyQualifiedTypeReferences` handles annotations unexpected. 

It seems to only shorten Annotations of the pattern ´packagename.@Annotation´ (see `ShortenFullyQualifiedTypeReferencesTest#typeFullyQualifiedAnnotatedField` and `ShortenFullyQualifiedTypeReferencesTest#annotatedFieldAccess`). 
In the `ShortenFullyQualifiedTypeReferences` fails if the annotation has the pattern `@packagename.TypeName`
According to the Java Language Spec 9.7.1 Annotations in FQN format have the pattern `@packagename.TypeName`. 
I think there might be a bug in `ShortenFullyQualifiedTypeReferences` or am I missing something?

knutwannheden: Quite possible that we have a bug there. I will try to investigate next week.

sambsnyd:
Thanks @MBoegers !

sambsnyd:
I've cleaned up the formatting of the tests, enabled them, and ensured that the relevant parser classpath can configured.
https://github.com/openrewrite/rewrite/commit/4831b1fb79a9e035862afbb9e72793b2bf824365

MBoegers:
I open an issue for the unexpected behaviour of `ShortenFullyQualifiedTypeNames` [4222](https://github.com/openrewrite/rewrite/issues/4222)

PR #4220, opened 2024-05-31
UpgradeDependencyVersion (gradle) not working

ryanwalker:
The Gradle recipe `UpgradeDependencyVersion` does not work in my project, no changes are applied at all.
I debugged and foudn that the method type is null, and the method matcher therefore does not match.

```
name: LocalTesting
preconditions:
  - org.openrewrite.FindSourceFiles:
      filePattern: "sdk/build.gradle"
recipeList:
  - org.openrewrite.gradle.UpgradeDependencyVersion:
      groupId: com.company
      artifactId: sdk-generation-template
      newVersion: 0.17.+
```

I put together this PR to show a workaround. I don't know that it's the right solution, but I wanted to point it out to either get some direction or see if you know another solution.

I didn't spend the time to delve into the parsing and creation of the LST, that is probably where the root cause actually is.

### Checklist


sambsnyd:
I've committed something for this which I based on this PR. I've listed you as co-author:
https://github.com/openrewrite/rewrite/commit/246b53ecab71f05794565db3ddbd8fd2cd1cc432

Thanks, Ryan 👍

PR #4230, opened 2024-06-07
Ignore configurations for constraints

Jenson3210:
## What's changed?
As discussed offline, we need to be able to exclude certain configurations for constraint creation in certain cases (vulnerabilities etc).

## What's your motivation?
Close our issue openend through slack on moderne.

## Any additional context
I will raise fix for java-rewrite-dependencies

### Checklist
- [X] I've added unit tests to cover both positive and negative cases
- [X] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

PR #4233, opened 2024-06-07
Only consider specified configurations for UpgradeTransitiveDependencies created constraints

Jenson3210:
## What's changed?
In this version, the recipe will by default consider all transitive configurations to be adding constraints. 
One can limit this behaviour by specifying constraints that need to be considered. 
Eg. in the light of vulnerability patching, a recipe runner might only be interested in running against `runtime` jars. 
That would be feasible by specifying both default runtime available configurations(implementation, runtimeOnly) + any manually added **transitive** ones.

This change will only have impact if a configurations list was specified. We reintroduce/combine earlier behaviour by checking for the extendsFrom + must be transitive. 
Added some test cases to further explain expectations/behaviour.

See #4230 for a blacklist approach

## Any additional context
Our preference goes towards a whitelist approach as that list is afaik better maintainable for wide usage of plugins, custom configurations... 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -98,6 +98,13 @@ public class UpgradeTransitiveDependencyVersion extends Recipe {
     @Nullable
     String because;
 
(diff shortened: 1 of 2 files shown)


sambsnyd:
Thanks @Jenson3210 I've merged this and closed the alternative blacklist proposal 
https://github.com/openrewrite/rewrite/pull/4230

PR #4236, opened 2024-06-07
Add Property Resolution to RemoveRedundantDependencyVersions

ryan-hudson:
## What's changed?
Added property version resolution for dependencies/plugins using the maven property syntax to define versions and removed the early exit logic for this case.

## What's your motivation?
Redundant Dependencies and Plugin versions defined through maven properties should still be removed by this recipe instead of being ignored.


timtebeek (approved):
Nice concise improvement @ryan-hudson , thanks a lot!

Issue #4241, opened 2024-06-09
No newline in import statements added by maybeAddImport / AddImportVisitor


Jenson3210:
## What version of OpenRewrite are you using?
I am using latest openRewrite

## How are you running OpenRewrite?
Using unit test case to show the issue.

## What is the smallest, simplest way to reproduce the problem?

```java
    @Test
    void combinedChangePackageAndChangeTypeMaintainsImportsCorrectly() {
        rewriteRun(
          spec ->
            spec.recipeFromYaml("""
                                type: specs.openrewrite.org/v1beta/recipe
                                name: org.openrewrite.test
                                displayName: Test
                                description: Test.
                                recipeList:
                                  - org.openrewrite.java.ChangePackage:
                                      oldPackageName: org.apache.http.client.methods
                                      newPackageName: org.apache.hc.client5.http.classic.methods
                                  - org.openrewrite.java.ChangeType:
                                      oldFullyQualifiedTypeName: org.apache.hc.client5.http.classic.methods.CloseableHttpResponse
                                      newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.CloseableHttpResponse
                                """, "org.openrewrite.test"),
          //language=java
          java(
            """
            import org.apache.http.client.methods.CloseableHttpResponse;
            import org.apache.http.client.methods.HttpUriRequest;
            import org.apache.http.impl.client.CloseableHttpClient;
            import org.apache.http.client.methods.RequestBuilder;
            class A {
                void method() {
                    RequestBuilder requestBuilder = null;
                    HttpUriRequest request = null;
                    CloseableHttpClient instance = null;
                    CloseableHttpResponse response = null;
                }
            }
            """,
            """
            import org.apache.hc.client5.http.classic.methods.HttpUriRequest;
            import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
            import org.apache.http.impl.client.CloseableHttpClient;
            import org.apache.hc.client5.http.classic.methods.RequestBuilder;

            class A {
                void method() {
                    RequestBuilder requestBuilder = null;
                    HttpUriRequest request = null;
                    CloseableHttpClient instance = null;
                    CloseableHttpResponse response = null;
                }
            }
            """)
        );
    }
```

## What did you expect to see?
import statements added on a single line each

## What did you see instead?
[...]

PR #4242, opened 2024-06-09
Import get inserted at correct index with newline space

Jenson3210:
Closes: #4241

## What's changed?
We will close the bug explained in the mentioned issue.

## What's your motivation?
We do not want to have imports on a single line.

If we need to go for another solution, at least the test case is already pushed.

See Issue for other possibilities. Or expect from EU that always imports are sorted.

### Checklist


timtebeek:
Appreciate how detailed you've taken us through the issue in 
- #4241 

Thanks for diving in and providing the fix!

PR #4245, opened 2024-06-12
Fix `Parameterized` is hidden by `withFullyQualifiedName`

OlegDokuka:
## What's changed?

This PR ensures that withFullyQualifiedName never hides the original `Parametrized` type, which includes info about generic params
## What's your motivation?
I noticed that bug, so I decided to fix it right away by PR

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
@@ -715,7 +715,12 @@ public interface JavaType {
 
         @Override
         public FullyQualified withFullyQualifiedName(String fullyQualifiedName) {
-            return type.withFullyQualifiedName(fullyQualifiedName);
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -715,7 +715,12 @@ public String getFullyQualifiedName() {
 
         @Override
         public FullyQualified withFullyQualifiedName(String fullyQualifiedName) {
-            return type.withFullyQualifiedName(fullyQualifiedName);
+            FullyQualified qualified = type.withFullyQualifiedName(fullyQualifiedName);
+            if (type.equals(qualified)) {
```
knutwannheden, line 719:
```suggestion
            if (type == qualified) {
```

OlegDokuka: fixed. Thanks!

knutwannheden (approved):

PR #4247, opened 2024-06-12
Add `AddPluginRepository` recipe

jdmcmahan:
## What's changed?
This PR adds a recipe for `AddPluginRepository` which can be used to manage a `<pluginRepositories>` configuration in a Maven POM.

## What's your motivation?
Although the `AddRepository` recipe is sufficient for most use cases, we encountered the need to reference a `<pluginRepository>` for plugin-specific dependencies. The following example illustrates this:

```xml
<pluginRepositories>
    <pluginRepository>
        <id>my-org</id>
        <url>https://maven.myorg.com/repository/</url>
        <name>My Organization's Repository</name>
    </pluginRepository>
</pluginRepositories>

<build>
    <plugins>
        <plugin>
            <artifactId>maven-checkstyle-plugin</artifactId>
            <version>3.4.0</version>
            <executions>
                <execution>
                    <id>validate</id>
                    <phase>validate</phase>
                    <goals>
                        <goal>check</goal>
                    </goals>
                </execution>
            </executions>
            <configuration>
                <configLocation>checkstyle.xml</configLocation>
            </configuration>
            <dependencies>
                <dependency>
                    <groupId>org.myorg</groupId>
                    <artifactId>myorg-custom-checkstyle-config</artifactId>
                    <version>1.0.0</version>
                </dependency>
            </dependencies>
        </plugin>
    </plugins>
</build>
```

This change is naïve in the sense that it effectively just copies the logic from the `AddRepository` class. I'm open to any opportunities to reuse or improve the code.

`AddRepository` does not provide an option to create the repository as a plugin repository. Maven itself does not fall back on non-plugin repositories for plugin dependencies.

### Checklist


timtebeek:
Great start @jdmcmahan ! Symmetry with AddPlugin is clear indeed; from a maintainability point of view I'm wondering if we should add an option to AddRepository instead to make that add either a regular or a plugin repository. I'll think that over and update your PR before a merge. Thanks again!

timtebeek (approved):
Ended up indeed reusing AddRepository instead @jdmcmahan , just so that we have less copies to maintain over time. I've taken the example test that you had provided and added that into AddRepositoryTest. Thanks again for getting this started; with this option it should be usable already

jdmcmahan:
> Ended up indeed reusing AddRepository instead @jdmcmahan , just so that we have less copies to maintain over time. I've taken the example test that you had provided and added that into AddRepositoryTest. Thanks again for getting this started; with this option it should be usable already

Awesome!! Thanks for picking this up!

PR #4259, opened 2024-06-14
Add preconditions to RecipeDescriptor for DeclarativeRecipe

timtebeek:


timtebeek:
As discussed in Slack this might not be the way to go; @kmccarp would be better able to indicate what to do instead, and if that should land in this repository or elsewhere.

PR #4266, opened 2024-06-17
Allow J.Empty type parameters in SemanticallyEqual

timtebeek:
## What's changed?
Allow either side of type parameters to be empty while still being equal.

## What's your motivation?
We noticed mismatches in some Refaster templates which should ideally still match.


knutwannheden (approved):
This PR confirms what I suspected. LGTM!

timtebeek:
Pushed up a small additional test just to be sure https://github.com/openrewrite/rewrite/commit/c8cb0fba70b2ae0a7e48090fae8686cdf427328c

PR #4268, opened 2024-06-17
Gradle `ChangeDependencyClassifier` supports add and remove classifier

BoykoAlex:
This is to align with Maven counter part


timtebeek (approved):
Thanks a lot for bringing parity to these recipes, and covering the various styles in both the recipe implementation and tests. Helps build confidence in the change. Hope this helps the recipes you're looking to write!

PR #4273, opened 2024-06-20
Maven runtime dependency is on the compile scope

BoykoAlex:
A simple runtime dependency is on the compile scope in the `ResolutionResult`. A simple unit test is included

If the fix should go in `Scope.isInClasspathOf(Scope)` we can try that.


timtebeek:
Looks good to me! Great to see you come back with a swath of fixes across the board that must not have been easy to clue together. Know that it's very much appreciated, and I'll see them through. Hope that helps resolve a number of puzzling transitive dependency issues.

I'll ping @sambsnyd for an async check after the merge, as he recently resurfaced from [this change](https://github.com/openrewrite/rewrite/commit/4df7f3af86fd93c8295eda2def375c56668a3a8d) a few lines down, so he'd be interested to see this fix too with the context he's gained over the past week.

PR #4274, opened 2024-06-20
DependencyManagement logic bug when versions are set via Maven property

gjd6640:
## What's changed?
Adds tests to UpgradeTransitiveDependencyVersionTest. Two demonstrate current behaviors and are added to prevent regression and demonstrate to the reader what the recipe can do.

The third demonstrates a potential bug:

I think this only happens when all of the following are true:

1. A BOM file in the dependencyManagement section sets the library's version number
2. That BOM file uses a Maven property to determine the library version
3. The parent pom also specifies that property and it sets it to the lower/undesired version

The plugin appears to look thru the list of managed dependency entries, sees the one in the BOM that sets the newer/desired version, and decides that no change is needed. Of course, the Maven engine understands the full context and is choosing the older version number specified via the property in the parent POM so the recipe does need to act to override the version used.

## What's your motivation?

We're remediating CVEs across many applications and some projects have this pattern present. We don't intend to continue to refine our parent POMs nor are we ready to rip them out so are trying to remediate CVEs by overriding in the local project's POM file.

@timtebeek has been discussing this with me in Slack.

This pull request serves to port these tests over from where they were built in [this rewrite-recipe-starter fork](https://github.com/gjd6640/rewrite-recipe-starter/blob/main/src/test/java/com/yourorg/UpgradeTransitiveDependencyVersionTest.java).

### Checklist
- [ x ] I've added unit tests to cover both positive and negative cases
- [ no, sorry  :) ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ no, sorry  :) ] I've used the IntelliJ IDEA auto-formatter on affected files


timtebeek:
Very clear example @gjd6640 ; thanks for opening the issue here; hope we can see this resolved for you going forward, although it might take a detailed look first.

timtebeek (commented):
Thanks a lot for the added tests! I think the ones that pass we could already merge to help explain the cases covered.

Indeed looks like we should be making the change you've logged in the for now failing test. Did you already explore the changes needed there?

timtebeek:
Looks like we've already fixed a couple of the issues pointed out here; here's the updated failures we still see:
```
UpgradeTransitiveDependencyVersionTest > upgradeTransitiveDependencyVersion_NewIssue_NotYetFixed() FAILED
        at org.openrewrite.test.LargeSourceSetCheckingExpectedCycles.afterCycle(LargeSourceSetCheckingExpectedCycles.java:119)
        at org.openrewrite.Recipe.run(Recipe.java:438)
        at org.openrewrite.maven.UpgradeTransitiveDependencyVersionTest.upgradeTransitiveDependencyVersion_NewIssue_NotYetFixed(UpgradeTransitiveDependencyVersionTest.java:116)

UpgradeTransitiveDependencyVersionTest > upgradeTransitiveDependencyVersion_WorksForBillOfMaterialsManagedDeps() FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "pom.xml":
    diff --git a/pom.xml b/pom.xml
    index ae97899..27fcf13 100644
    --- a/pom.xml
    +++ b/pom.xml
    @@ -8,7 +8,7 @@ 
         <version>1.0.0</version>
 
       <dependencyManagement>
    -    <dependencies>
    +      <dependencies>
           <dependency>
             <artifactId>spring-boot-actuator</artifactId>
    ] 
```
That last one we can ignore; or fix the indentation to see that fixed.

The first one is [a new test](https://github.com/openrewrite/rewrite/pull/4274/files#diff-1a8b7e3ef407e2c28469f32cb84657c93207ccd0f1d1f821dc9c4b87c9f4eac4R106-R203) that appears related to what's discussed in more detail here
- https://github.com/openrewrite/rewrite/pull/5051

Let's close this draft and continue on the other PR for a fix for that issue. Thanks again @gjd6640 !

Issue #4275, opened 2024-06-20
ClassCastException in UpgradeDependencyVersion


Jenson3210:
## What version of OpenRewrite are you using?
I am using moderne / latest cli

## How are you running OpenRewrite?
Moderne

## What is the smallest, simplest way to reproduce the problem?
```groovy
plugins {
    id 'java'
}
repositories {
    mavenCentral()
}
dependencies {
    //Although this one does not belong here, the method invocation with "true" is a boolean J.Literal.
    jar {
       enabled(true)
    }
}
```

## What did you expect to see?
This line should be ignored as it is not really a dependency configuration. Since it is inside the dependencies block, `isLikelyDependencyConfiguration` will result in true, hence the visitMethodInvocation will be called for that method. 

## What did you see instead?
ClassCastException when arg is an instance of J.Literal
<img width="750" alt="Screenshot 2024-06-20 at 21 56 01" src="https://github.com/openrewrite/rewrite/assets/25441280/b4c36fea-3384-424b-b144-688af18caa32">

## What is the full stack trace of any errors you encountered?
```
# Error:
allowInsecureProtocol(true)

# Message:
class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')

# Detail:
java.lang.ClassCastException: class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')
org.openrewrite.gradle.UpgradeDependencyVersion$UpdateGroovy.lambda$updateDependency$8(UpgradeDependencyVersion.java:438)
org.openrewrite.internal.ListUtils.map(ListUtils.java:176)
org.openrewrite.gradle.UpgradeDependencyVersion$UpdateGroovy.updateDependency(UpgradeDependencyVersion.java:409)
org.openrewrite.gradle.UpgradeDependencyVersion$UpdateGroovy.visitMethodInvocation(UpgradeDependencyVersion.java:397)
org.openrewrite.gradle.UpgradeDependencyVersion$UpdateGroovy.visitMethodInvocation(UpgradeDependencyVersion.java:350)
org.openrewrite.java.tree.J$MethodInvocation.acceptJava(J.java:3945)
org.openrewrite.java.tree.J.accept(J.java:59)
org.openrewrite.TreeVisitor.visit(TreeVisitor.java:250)
[...]

PR #4276, opened 2024-06-20
Ignore method invocation arguments that are not string J.Literals

Jenson3210:
Fixes: #4275

## What's changed?
A check before casting. If it is not a string, it is most likely not a dependency notation. 

## What's your motivation?
In line with: "Do not change anything if we are not sure", I propose to just give back this argument as is without touching. 

## Any additional context
This is actually not meant to be coded like this in gradle, yet it is working unfortunately and has been copy/pasted througout some of our repo's.

### Checklist


timtebeek (approved):
Neat and localized fix, thanks!

PR #4277, opened 2024-06-20
Add PlaintextOnly Option to FindAndReplace Recipe

ryan-hudson:
## What's changed?
Added a Plaintext Only option to the find and replace recipe, which when enabled, limits the recipe to only act on plaintext files. This option defaults to false.

## What's your motivation?
FindAndReplace by default replaces text in for most sourcefile types, which may not be desirable in certain cases since this recipe coverts any sourcefile to a plaintext file, thereby causing information loss for the rest of the recipe run. Enabling this flag limits the recipe scope to plaintext files for cases where a user only wants to replace something thats already parsed as a plaintext file. 

For example, this flag could prove useful in [this](https://github.com/openrewrite/rewrite-spring/pull/538) PR where we only want to touch manifest and dockerfiles.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
@@ -15,13 +15,16 @@
  */
 package org.openrewrite.text;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
 import org.openrewrite.binary.Binary;
+import org.openrewrite.internal.lang.NonNull;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.marker.AlreadyReplaced;
 import org.openrewrite.marker.Marker;
+import org.openrewrite.marker.SearchResult;
 import org.openrewrite.quark.Quark;
 import org.openrewrite.remote.Remote;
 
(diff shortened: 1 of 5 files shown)


rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
```diff
@@ -99,6 +102,39 @@ public String getDescription() {
...
+        this.replace = replace;
+        this.regex = regex;
+        this.caseSensitive = caseSensitive;
+        this.multiline = multiline;
+        this.dotAll = dotAll;
+        this.filePattern = filePattern;
+        this.plaintextOnly = null;
+    }
+
+    @JsonCreator
```
timtebeek, line 124:
```suggestion
```

rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
```diff
@@ -15,13 +15,16 @@
  */
 package org.openrewrite.text;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
```
timtebeek, line 18:
```suggestion
```

rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
timtebeek: While I appreciate your concerns for backwards compatibility, I think it's fine in this case to add the argument and just fix it downstream if needed. I reality folks will mostly use this recipe from Yaml files anyway, such that there's rarely a need to update code.

rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
```diff
@@ -156,6 +192,15 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
 
             visitor = Preconditions.check(check, visitor);
         }
+
+        if (Boolean.TRUE.equals(plaintextOnly)) {
+            visitor = Preconditions.check(new PlainTextVisitor<ExecutionContext>(){
```
timtebeek, line 197: I hadn't seen this pattern of effectively nested `Precondition.check` before, as opposed to and/or-ed together check condition; not sure how I feel about those yet. I'd have to look a little closer in my IDE.

timtebeek (commented):
Good suggestion indeed! Prevents accidental LST type loss, so good safe addition. Some minor suggestions and a note for myself to look closer.

rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java
timtebeek: Looks ok; or at least unlikely to conflict given that one is a file pattern already, and folks are likely to use one or the other.

timtebeek (approved):
Great addition!

timtebeek:
Also applied downstream in https://github.com/openrewrite/rewrite-spring/commit/7fb21edc00e64192dc1412543729bdec362aa0f3

PR #4282, opened 2024-06-23
fix: fix bellwether decorated recipe ignoring causesAnotherCycle

aslaad:
## What's changed?
Declarative recipes with preconditions no longer ignore "causesAnotherCycle" of bundled recipes.

I don't know how to test this as I couldn't reproduce it in unit tests (seems to always run extra cycles when testing?). I'm fairly new to OpenRewrite so any guidance would be welcome.

## Any additional context
A declarative recipe I'm working on was not completing as expected.
It looked like additional cycles caused by one of the recipes in the list were not being processed.
Since I couldn't reproduce it in my unit tests, I debugged the use case and noticed that causesAnotherCycle was not properly delegated to my recipe causing the run to end prematurely.
I run OpenRewrite via the IntelliJ plugin.

### Checklist
- [X]  I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
@@ -250,6 +250,11 @@ public class DeclarativeRecipe extends Recipe {
         public List<Recipe> getRecipeList() {
             return decorateWithPreconditionBellwether(bellwether, delegate.getRecipeList());
         }
+
+        @Override
+        public boolean causesAnotherCycle() {
+            return delegate.causesAnotherCycle();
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for diving in and coming through with the pull request @aslaad ! Indeed the recipe execution flow could be ever so slightly different for the testing framework as judged from your test issues, but this seems like an oversight corrected with your change. I've added the same for `BellwetherDecoratedScanningRecipe` just to have parity there. Briefly considered `@lombok.Delegate`, but that seems like a larger change with harder to oversee effect, so we'll keep it at this for now.

timtebeek (approved):

Issue #4093, opened 2024-03-14
StackOverflowError on attempt to run against import dependencies


ZhyliaievD:
Hi,

maven-rewrite-plugin fails with `java.lang.StackOverflowError` when there is a dependency with `import` scope having a parent declaration
`org.openrewrite.maven.tree.ResolvedPom` tries to resolve parent of such dependency, which in turn contains the dependency itself, resulting in circular dependency
Of course, that's a clear problem in target project, but it is ignored by maven during build process so people can easily miss it

## What version of OpenRewrite are you using?
latest version of rewrite-maven-plugin (5.25.0)

## What is the smallest, simplest way to reproduce the problem?
Run any recipe against a sample project:
https://github.com/ZhyliaievD/rewrite-maven-plugin-soe-sample 
e.g. sample command to run:
`mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.activeRecipes=org.openrewrite.java.RemoveUnusedImports`

<details><summary>OR use the following code snippets to reproduce using the following parent&child poms:</summary>
<p>

project structure:
```
/pom.xml.  <-- parent pom
/circular-example-child/pom.xml  <-- child pom
```

Parent pom:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>org.example</groupId>
  <artifactId>circular-example</artifactId>
  <version>1.0-SNAPSHOT</version>
  <packaging>pom</packaging>

  <name>circular-example</name>
  <url>http://maven.apache.org</url>

  <modules>
    <module>circular-example-child</module>
  </modules>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
  </properties>

  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.example</groupId>
        <artifactId>circular-example-child</artifactId>
        <version>1.0-SNAPSHOT</version>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
</project>
``` 

child pom:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

PR #4284, opened 2024-06-25
Fix StackOverflowError in case of circular import deps

ZhyliaievD:
Check if import dependency was already resolved before

## What's changed?
Add a check if import dependency was already resolved before

## What's your motivation?
To avoid StackOverflowError in case of circular import deps 
- https://github.com/openrewrite/rewrite/issues/4093

Fixes https://github.com/openrewrite/rewrite/issues/4093

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -421,12 +421,12 @@ public class ResolvedPom {
 
             for (Profile profile : pom.getProfiles()) {
                 if (profile.isActive(activeProfiles)) {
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for getting this started @ZhyliaievD ! I've updated your PR a bit to add additional assertions to your test that we indeed get merged. Would you mind double checking that you indeed expect `junit` present on both the parent and child as the only managed dependency versions there?

ZhyliaievD:
Hi @timtebeek 
Thans for th adjustments(and sorry for late response - just got back home)
Regarding expected managed dependencies - correct, I only expect junit in test case (Reference from [the doc](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope) about import scope: `It indicates the dependency is to be replaced with the effective list of dependencies in the specified POM's`)

PR #4286, opened 2024-06-25
fix: Add DelegatingRecipe to find underlying recipe for loading options

pstreef:
Fixes a problem where RecipeDescriptor does not have the options on it when a recipe has preconditions.

## What's changed?
When a recipe is actually a delegate it can return that delegate to be used in option extraction.

## What's your motivation?
Currently a recipe with preconditions will have no options on the descriptors of children

Adding a specific check for the Bellwether wrapper, but I think this is cleaner.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Recipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Recipe.java
@@ -421,7 +421,7 @@ public abstract class Recipe implements Cloneable {
         }
     }
 
-    public abstract static class DelegatingRecipe extends Recipe {
-        public abstract Recipe getDelegate();
+    public interface DelegatingRecipe {
+        Recipe getDelegate();
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/Recipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Recipe.java
@@ -20,7 +20,6 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import lombok.Setter;
 import org.intellij.lang.annotations.Language;
-import org.jetbrains.annotations.Nls;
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/Recipe.java
```diff
@@ -416,4 +420,8 @@ public Object clone() {
             throw new RuntimeException(e);
         }
     }
+
+    public abstract static class DelegatingRecipe extends Recipe {
+        public abstract Recipe getDelegate();
+    }
```
sambsnyd, line 426: Any reason this is an abstract class rather than an interface? Seems like it could be an interface

pstreef: Because it extends recipe, which is an abstract class. We could remove the extends ofcourse, but I thought this made more sense

pstreef: I tried the inteface, it looks a little nicer, but it does not force the DelegatingRecipe to be a Recipe itself. Not sure which I like better. What do you think? (see last commit)

rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
```diff
@@ -168,7 +168,7 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
 
     @EqualsAndHashCode(callSuper = false)
     @Value
-    static class BellwetherDecoratedRecipe extends Recipe {
+    static class BellwetherDecoratedRecipe extends Recipe implements DelegatingRecipe {
```
timtebeek, line 171: Should `BellwetherDecoratedScanningRecipe` below implement the same interface?

pstreef: yes, good catch!

Issue #4239, opened 2024-06-07
"Add Maven Parent" recipe


rcsilva83:
## What problem are you trying to solve?
I need to add a parent POM to my projects. 

## Describe the solution you'd like
I'd like a recipe for that as there already has one for changing the Maven parent.

## Have you considered any alternatives or workarounds?
Using text recipe

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
I can try :)

timtebeek:
hi @rcsilva83 ! Thanks for the suggestion. For this one I'd assume the implementation would be rather similar to [Change Maven Parent](https://docs.openrewrite.org/recipes/maven/changeparentpom)
https://github.com/openrewrite/rewrite/blob/28e7b125655fa507d68f80ed1dafa2dee38a38fa/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java#L44

Perhaps we can start with a draft PR with some unit tests, and then I'll help guide you through what's needed beyond that.

Tests are also a great way to think through and codify how you'd want to handle certain cases, like:
- should we match just any pom.xml file? Or only matching certain criteria?
- How to handle an existing parent? Is that the same for multi module projects?

rcsilva83:
Hi @timtebeek 

I'm trying to build the project on my machine but I'm having issues:

```
A problem occurred configuring project ':tools:language-parser-builder'.
> Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=17, vendor=any, implementation=vendor-specific} for WINDOWS on x86_64.
   > No locally installed toolchains match and toolchain download repositories have not been configured.
```

I'm running with JDK21 but it I run with JDK17 I have the same problem with `languageVersion=21`. I'm trying to configure `org.gradle.java.installations.paths` on `gradle.properties` but nothing changed...

Can you help me with this?

timtebeek:
hi! That's _curious_; are you using any non standard paths for your installed JDKs? I know Gradle does some discovery that might fail to detect your JDK paths.

You might be able to work around this though by following these instructions to not load troublesome modules:
https://github.com/openrewrite/rewrite/blob/28e7b125655fa507d68f80ed1dafa2dee38a38fa/IDE.properties.tmp#L1-L7

We also have some more general IDE setup instructions on these pages.
- https://docs.openrewrite.org/reference/building-openrewrite-from-source
[...]

PR #4289, opened 2024-06-26
`AddParentPom` recipe

rcsilva83:
Tests fail with "Recipe was expected to make a change but made no changes."

## What's changed?

A new AddParentPom recipe is added.

## What's your motivation?

I need to add a parent POM to my projects.

Discover why is the recipe making no changes (as stated by the tests) and how to better share code with ChangeParentPom recipe.

@timtebeek (as he invited me to make a draft PR on #4239)

### Checklist


timtebeek:
Great start @rcsilva83 ! Thanks for kicking this off. I've pushed a change to your PR to get the tests running. You had overridden `isAcceptable`, but for the precondition check to work you'd need to instead override a visit method, and return a changed instance to trigger the guarded visitor. `SearchResult.found` does exactly that. Hope that helps you explore what the recipe actually should do, as it seemed you got stuck at this point.

rcsilva83:
Hi @timtebeek !

Thank you very much for your help. The recipe is almost done: it's working but I still need to handle new parent's properties and dependency management.

However, I have some doubts:

1. What to do when no versions are found when using wildcard version? (see line 152)
2. Should I always search for version on Maven repository or should I bypass this when the version is fixed? From what I saw on other recipes this kind of optimization isn't done.

Regards!

rcsilva83:
> I still need to handle new parent's properties and dependency management.

Actually, do you think is there anything to do on this regard? I though I needed to remove properties and dependency management declarations from the pom.xml if they exist on the new parent pom but I don't think that's the case. The `ChangeParentPom` recipe just add properties and dependency management declarations from the old parent pom, right?

If you can confirm this to me, I'll remove some unused code from the recipe and, after you feedback on the 2 questions I made before, it will be done.

timtebeek:
Thanks a lot @rcsilva83 ! To answer your questions I'll reply to each in a separate paragraph below.

In ChangeParentPom we seem to [skip with no changes](https://github.com/openrewrite/rewrite/blob/7e1130c1120ec5289300d5efc4c8db2629ab00b3/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java#L170-L177) when no acceptable version is found. Probably best to be consistent then, even if it could make for a frustrating debugging experience when there are misconfigurations.

I wouldn't skip over validating the parent; we'd like to be sure the parent exists, since we'll use it to update the model of the code, and might otherwise lead to invalid configuration.

As for updating properties: I'd maybe only go as far as doing a `doAfterVisit(new org.openrewrite.maven.RemoveRedundantDependencyVersions().getVisitor())`; that clears out one of the major reasons to add a parent pom.xml; if there's more clean ups to be done then I suggest to similarly move those to separate recipes, such that there's less complexity in each, with the opportunity for reuse.

Is there anything else you'd then still like to push before a review and merge?

rcsilva83:
> In ChangeParentPom we seem to [skip with no changes](https://github.com/openrewrite/rewrite/blob/7e1130c1120ec5289300d5efc4c8db2629ab00b3/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java#L170-L177) when no acceptable version is found. Probably best to be consistent then, even if it could make for a frustrating debugging experience when there are misconfigurations.

I'll adjust this and make a new push. Then I think it'll be ready for merge :)

rcsilva83:
@timtebeek I need your help on a new problem. After changing the change to make no changes when no suitable version is found, I still got an empty `<parent/>` in the and, because it was added in the `visitDocument` method. I couldn't find a way to remove it, so I moved all the code to `visitDocument` method and removed `visitTag` method.

Everything is fine but apparently the `doAfterVisit(new org.openrewrite.maven.RemoveRedundantDependencyVersions().getVisitor())` isn't working anymore and the versions matching with parent aren't being removed anymore.

How can I solve this? I couldn't find any situation like this in other recipes.

timtebeek:
All tests seem to pass; are we missing anything? Wondering because of your question above and the draft status. Ideally we have a failing test if there's anything to add still.

We're planning a new release tomorrow; let me know if this should make it in in the current form, or with small changes.

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Option(displayName = "File pattern",
+            description = "A glob expression that can be used to constrain which directories or source files should be searched. " +
+                          "Multiple patterns may be specified, separated by a semicolon `;`. " +
+                          "If multiple patterns are supplied any of the patterns matching will be interpreted as a match. " +
+                          "When not set, all source files are searched. ",
+            required = false,
+            example = "**/*-parent/grpc-*/pom.xml")
+    @Nullable
+    String filePattern;
```
timtebeek, line 77: Typically we [use Preconditions](https://docs.openrewrite.org/reference/yaml-format-reference#preconditions) when we want to limit a recipe execution to a particular file. In particular it seems like a precondition of `org.openrewrite.FindSourceFiles` could be a replacement for this explicit option. Any arguments against dropping this option and nudging folks towards preconditions instead? That would be better for consistency, and ensures there's not proliferation of common options implemented in multiple repositories.

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Override
+    public String getInstanceNameSuffix() {
+        return String.format("`%s:%s:%s`", groupId, artifactId, version);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Change the parent pom of a Maven pom.xml. Identifies the parent pom to be changed by its groupId and artifactId.";
```
timtebeek, line 91: Perhaps good to note here what our handling is in the case of an existing parent, or a multi module project. That would help clarify the usage patterns we expect to see, and the changes users can expect from the recipe.

rewrite-maven/src/test/java/org/openrewrite/maven/AddParentPomTest.java
```diff
@@ -0,0 +1,800 @@
...
+                  <my-project-version>${project.version}</my-project-version>
+                  <my-env-prop>${env.GIT_HOME}</my-env-prop>
+                  <my-settings-prop>${settings.offline}</my-settings-prop>
+                </properties>
+              </project>
+              """
+          )
+        );
+    }
+
```
timtebeek, line 751: These tests are not necessary I think; there's no corresponding logic that would cause these to be added, right?
```suggestion
```

timtebeek (commented):
Some initial feedback already, with a few questions that might lead to stripping out some functionality still.

rcsilva83:
Hi @timtebeek ! I'll take a look on you feedback and make the code changes needed. But I committed the failing tests I told before so you can understand the problem. I just solve them moving the code to `visitDocument` method but the `RemoveRedundantDependencyVersions` stopped working.

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
rcsilva83: OK, I'll remove it.

I'm new to OpenRewrite and didn't know about Preconditions. I added the `filePattern` field based on `AddPlugin` recipe.

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
rcsilva83: Right, I forgot that :)

Do we need to be explicit about multi module project? If the recipe find a pom.xml with a parent already defined, it'll ignore it. So, in a multi module project, it won't add to the child modules as they already have a parent.

timtebeek (approved):
Nice work here @rcsilva83 ! The reason `RemoveRedundantDependencyVersions` stopped working is because you had removed the call to `maybeUpdateModel()`. That ensures our model is aware of the Maven parent before we make subsequent changes.

I've applied a few more polishing commits, and with those I think we're ready to merge. Thanks again!

rcsilva83:
@timtebeek, and I thank **you** for the guidance and patience! Now I know why OpenRewrite is improving so fast and have such a vibrant community :)

Regards!

timtebeek:
Thank you! Couldn't do it without so many folks pitching in. You're changes have already landed in
https://github.com/openrewrite/rewrite/releases/tag/v8.29.0
I'll follow up with downstream releases of recipe modules and the OpenRewrite plugins, culminating in our rewrite-recipe-bom over the next couple hours: https://github.com/openrewrite/rewrite-recipe-bom/releases

PR #4290, opened 2024-06-27
Preserve multiple dependencyManagement with different types

askoog:
## What's changed?

Find managedDependency() did not take type into account

## What's your motivation?

It is a valid case to specify different dependencyManagement dependencies for the same artifact but with different types (used for example if you create a test-jar of a dependency and manage its version). Without this fix, the RemoveDuplicateDependencies recipe removed duplicate dependencyManagement dependencies with same artifact but different types


timtebeek (approved):
Thanks for the immediate fix and added test! Seems like a clear omission & improvement.

PR #4292, opened 2024-06-27
ChangeType leads to botched imports for nested fields

timtebeek:
After 
- https://github.com/openrewrite/rewrite/pull/4291

We continued to see issues in
- https://github.com/openrewrite/rewrite-jackson/pull/6

I've updated the test to better replicate the issue there. With this change the test fails with left imports.

```diff
Unexpected result in "A.java":
diff --git a/A.java b/A.java
index 98d3bcb..da17811 100644
--- a/A.java
+++ b/A.java
@@ -1,6 +1,8 @@ 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 
 import static com.fasterxml.jackson.databind.SerializationFeature.WRAP_ROOT_VALUE;
+import static org.codehaus.jackson.map.SerializationConfig.SerializationFeature.WRAP_ROOT_VALUE;
 
 class A {
     void test() {

```

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
@@ -181,7 +181,7 @@ class ChangeTypeTest implements RewriteTest {
               class Test {
                   List p;
                   List p2;
-                  List p3;
+                  java.util.List p3;
               }
               """
           )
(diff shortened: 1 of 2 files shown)


timtebeek:
@sambsnyd if you're up for a puzzle, here's that updated test. I didn't yet get into looking why this now happens

timtebeek:
This comparison fails 
https://github.com/openrewrite/rewrite/blob/f6473b6bab48a652629744c0633b84e20e7df2ce/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java#L272
For this test
https://github.com/openrewrite/rewrite/blob/1d8b3c3eb8616a84ecd0de1d44f2feb7f393a1ae/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java#L164-L166

Because everything up to `java.` is a `J.FieldAccess`, all of it ending up in `maybeClass`, whereas that should only contain the simple class name.
image

PR #4297, opened 2024-06-30
MergeYaml should avoid adding query with no result as key at end of file

adastraperangusta:
In case the jsonpath is a query which yelds no result, avoid merging the yaml snippet at end of file with the query as a key.

## What's changed?
In case the jsonpath has no match in a given yaml file, the MergeYaml recipe appends the yaml snippet at the end of the file under the jsonpath used as a key.
When this key is a JsonPath query using function, this creates an invalid key.

## What's your motivation?
Applying MergeYaml on  yaml files which don't match the query adds an extra block which is not needed.
Checking this type of key seems easier than using "preconditions"

## Any additional context
Maybe the MergeYaml recipe should have an optional "mergeIfKeyNotFound" property set to true by default.

### Checklist


timtebeek (approved):
Thanks for the immediate fix @adastraperangusta ! Nice to see a test that was previously disabled now passes too.

We expect to do a release either today or tomorrow, but until then you can use [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

Issue #4240, opened 2024-06-08
JSP parse errors logged to console during `mod build`


bryceatmoderne:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.27.2
- Maven plugin v3.
- rewrite-_module_ v8.27.2

## How are you running OpenRewrite?
https://github.com/2345free/video-web has several JSP files that the CLI fails to parse. Any parse errors encountered are logged to the CLI console, as opposed to build.log like so:
```
../mods/latest/mod build 2345free/video-web
        ▛▀▀▚▖  ▗▄▟▜
        ▌   ▜▄▟▀  ▐
        ▛▀▀█▀▛▀▀▀▀▜
        ▌▟▀  ▛▀▀▀▀▜
        ▀▀▀▀▀▀▀▀▀▀▀
Moderne CLI 3.4.0-SNAPSHOT

> Selecting repositories

> 2345free/video-web@master
Selected 1 repositories (0.05s)

> Building LST(s)

> 2345free/video-web@master
    Build output will be written to /Users/brycetompkins/repos/mass-ingest-300K/2345free/video-web/.moderne/build/20240528090619-nDiqh/build.log
    > Step 1 - build with Maven
        Selected the 1.8.0_412-zulu JDK
    > Step 2 - build resources using the native CLI
 26% |===> ---------| (36s) ...345free/video-web/src/main/webapp/WEB-INF/jsp/about.jspline 63:14 token recognition error at: '-'
line 63:15 token recognition error at: '-'
line 63:21 token recognition error at: '<'
line 83:6 token recognition error at: '-'
line 83:7 token recognition error at: '-'
line 83:8 token recognition error at: '<'
line 85:6 token recognition error at: '-'
line 85:7 token recognition error at: '-'
line 85:8 token recognition error at: '<'
line 86:6 token recognition error at: '-'
line 86:7 token recognition error at: '-'
line 86:8 token recognition error at: '<'
line 87:6 token recognition error at: '-'
line 87:7 token recognition error at: '-'
line 87:8 token recognition error at: '<'
line 88:6 token recognition error at: '-'
line 88:7 token recognition error at: '-'
line 88:8 token recognition error at: '<'
 46% |======> ------| (36s) ...ree/video-web/src/main/webapp/WEB-INF/jsp/configure.jspline 157:47 token recognition error at: '"<'
line 159:46 token recognition error at: '"<'
line 168:6 token recognition error at: '-'
[...]

PR #4303, opened 2024-07-05
JSP parse errors fixed

Riyazul555:
## Any additional context
Closes issue #4240 

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
@@ -42,6 +42,9 @@ public class XmlParser implements Parser {
             try (EncodingDetectingInputStream is = input.getSource(ctx)) {
                 String sourceStr = is.readFully();
 
+                // Preprocess JSP-specific syntax
+                sourceStr = preprocessJspSyntax(sourceStr);
+
                 XMLParser parser = new XMLParser(new CommonTokenStream(new XMLLexer(
                         CharStreams.fromString(sourceStr))));
 
@@ -62,6 +65,11 @@ public class XmlParser implements Parser {
                 return ParseError.build(this, input, relativeTo, ctx, t);
             }
         });
+    }
+
+    private String preprocessJspSyntax(String source) {
+        // Replace JSP comments with XML comments to avoid parser errors
+        return source.replaceAll("<%--(.*?)--%>", "<!--$1-->");
     }
 
     @Override


Riyazul555:
Hey @timtebeek do checkout this PR
Thanks

timtebeek:
That won't work here @Riyazul555 ; We want to faithfully reproduce the files as parsed; this workaround would likely lead to unintended changes, as you'll likely see when you add a unit test.

For any contribution I think it's best that we start with a unit test at what's intended, and then work from there. Typically I'd only expect a PR to be opened up for review when your tests pass, or when you're looking for help.

rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
```diff
@@ -64,6 +67,11 @@ public Stream<SourceFile> parseInputs(Iterable<Input> sourceFiles, @Nullable Pat
         });
     }
 
+    private String preprocessJspSyntax(String source) {
+        // Replace JSP comments with XML comments to avoid parser errors
+        return source.replaceAll("<%--(.*?)--%>", "<!--$1-->");
+    }
```
timtebeek, line 73: This would likely break print equality. Let's add a test first before continuing any further.

timtebeek (changes requested):

rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
Riyazul555: So you want a unit test here 
Ok I will do it @timtebeek 
thanks

Jenson3210: Important to keep in mind here in the parsing process is that you can actually not change the code. 
If you would change the input code, the parsed code differs from whatever is really in the codebase. 
You would basically be running recipes against a codebase that never existed.

Rather than replacing these tags, I would expect some logic to parse them correctly. 
Jsp comments are "valid" code and could perfectly be parsed if we extend the XMLParser to understand other tags than the current one. 

For the unit test example: Have a look at my showcase of #4306 on [my fork](https://github.com/Jenson3210/rewrite/commit/d81cb5e62a594b25fe52dec171f2ea221d522cbe) I think you could start from this unit test, add a jsp comment and use that test as input for all parser fixes. 
This would then fix #4306 and #4240

timtebeek:
I'm going to close this PR for now, as we need a very different approach to get this to work, unlikely to reuse the work done in this PR so far. If this branch is updated to contain tests then we could reopen this again. Until then let's make it clear this is not a fix just yet.

Jenson3210:
I will try to see what I can come up with "if this keeps me awake 😅" in my fork with some tests already, though I am no jsp expert.

Issue #4304, opened 2024-07-05
UseStaticImport should rewrite if method already exists


protocol7:
## What version of OpenRewrite are you using?
* org.openrewrite:rewrite-core:jar:8.29.0

## What is the smallest, simplest way to reproduce the problem?

UseStaticImport should not rewrite with an import static when there is a method in the scope with the same name (and maybe signature).

The second case below I'm not sure how to best deal with, perhaps don't rewrite the `String.valueOf`, or do what the test here does.

```java
  @Test
  void testUseStaticImport() {
    rewriteRun(
        spec -> spec.recipe(new UseStaticImport("java.util.List of(..)")),
        // language=java
        java(
            """
                    package com.helloworld;

                    import java.util.List;

                    public class Foo {

                        void of() {}

                        void foo() {
                          List.of();
                        }
                    }
            """));
  }

  @Test
  void testUseStaticImport2() {
    rewriteRun(
        spec ->
            spec.recipe(
                new CompositeRecipe(
                    List.of(
                        new UseStaticImport("java.lang.String valueOf(..)"),
                        new UseStaticImport("java.lang.Integer valueOf(..)")))),
        // language=java
        java(
            """
                    package com.helloworld;

                    public class Foo {
                        void foo() {
                          String.valueOf("1");
                          Integer.valueOf("1");
                        }
                    }
            """,
            """
                    package com.helloworld;

                    import static java.lang.String.valueOf;
        
                    public class Foo {
                        void foo() {
                          valueOf("1");
                          Integer.valueOf("1");
                        }
                    }
            """
        ));
  }
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Happy to give it a try with some pointers, as usual not sure where to start.

timtebeek:
Thanks for the report! In this case I think extending the preconditions with a separate visitor that looks at static imports used might be best. We already have a check on not declaring any similarly named methods. This would be a similar check but then looking at not having any similarly named static imports already.

https://github.com/openrewrite/rewrite/blob/c97c7f214634c78b257e6a7eb6705727f3d46900/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java#L62

Welcome to add just those two unit tests to a draft pull request to kick this off; then we can see if you work it out from there, or tag team to get this through.

protocol7:
> We already have a check on not declaring any similarly named methods

[...]

PR #4307, opened 2024-07-06
Fix method name collisions with UseStaticImport

protocol7:
Ensure that an import static is not created if a method with the same name already exists, either locally (this fixes this to take into account package name) or already imported statically.

- Fixes https://github.com/openrewrite/rewrite/issues/4304

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/UseStaticImportTest.java
@@ -18,10 +18,7 @@ package org.openrewrite.java;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.Issue;
-import org.openrewrite.config.CompositeRecipe;
 import org.openrewrite.test.RewriteTest;
-
-import java.util.List;
 
 import static org.openrewrite.java.Assertions.java;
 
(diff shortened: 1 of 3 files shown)


protocol7:
This adds failing tests for now.

timtebeek:
Thanks for kicking this off @protocol7 ; was triggered to have a look, and hope you don't mind me solving this one.

I only just now saw your https://github.com/openrewrite/rewrite/issues/4304#issuecomment-2211719725 

> > I think extending the preconditions with a separate visitor that looks at static imports used might be best

> I'm not sure how to define this. I gather I could use UsesMethod, but not sure how to write the method pattern in a way that includes any other method named, say, "foo", except for the method we are trying to rewrite.

You'll see I went for defining a new visitor that specifically looks for other static imports of the same name, and if found return a search result indicating that the `Precondition.not` should not visit the source file further. Hope that helps you on a future recipe!

https://github.com/openrewrite/rewrite/blob/f4da66ae5a48bed24ec3d889e42b5e2761f8975e/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java#L64-L72

timtebeek (approved):
Thanks a lot! The `DeclaresMethod` fix and additional tests were already a great help. Beyond that it was a matter of adding a new precondition visitor that merely looks at existing imports. The way recipes run in sequence also means that your test using two `new UseStaticImport` could replace the first case, but not the second, as the second time around we see the modified model and imports as output from the first recipe. Hope you'll agree and this helps you along.

protocol7:
Perfect, thanks for finishing this one up!

PR #4308, opened 2024-07-08
Fix/connection timeout in test

ErhardSiegl:
## What's changed?
Increased connection timeout in test.

## What's your motivation?
Build did fail with slow connections.


timtebeek:
Thanks! Hope this indeed helps build stability without adding to build duration.

PR #4309, opened 2024-07-08
`Trait` interface and `VariableAccess`/`MethodAccess` implementation

jkschneider:
## What's changed?

Introduces a `Trait` API into OpenRewrite which encapsulates semantic information on top of syntax.

## What's your motivation?

We had an initial implementation of `Trait` in rewrite-analysis written by @JLLeitschuh that really did a lot to help understand the shape of what is needed. 

In this implementation, we've removed the dependency on functionaljava, reasoning that this dependency would be too steep of a learning curve for the average user. One of the key benefits of functionaljava was in support of Jonathan's enhanced `Validation` type which has also been removed here. I think `Validation` was mostly useful in the development of the trait, and where possible traits should build on other traits so that the logic of each is fairly comprehensible.

We are also adding some API, in particular take note of how `Trait` can be turned into a visitor. Check out the test for how that is used. I'm imagining a future where many recipes would start out with a Trait that quickly hones the recipe in on a point of interest in the code.

## Example 1 (use the `higher` method to assert something about a tree's ancestors)

```java
public class MicrometerNamesUseDots extends Recipe {
    @Override
        return "Micrometer meters should use dot separation";
    }

    @Override
        return "Micrometer naming convention automatically replaces dots with underscores," +
               "camel casing, or whatever is idiomatic depending on the monitoring system " +
               "that the metrics are going to.";
    }

    @Override
        return new JavaIsoVisitor<>() {
            final MethodAccess.Matcher matcher = new MethodAccess.Matcher("io.micrometer.core.instrument.MeterRegistry *(String, ..)")
              .returns(ret -> TypeUtils.isAssignableTo("io.micrometer.core.instrument.Meter", ret));

            @Override
            public J.Literal visitLiteral(J.Literal literal, ExecutionContext ctx) {
                if (matcher.higher(getCursor()).findAny().isPresent()) {
                    //noinspection DataFlowIssue
                    String value = literal.getValue().toString();
                    if (value.contains("_")) {
                        return literal.withValue(value.replace("_", "."));
                    }
                }
                return super.visitLiteral(literal, ctx);
            }
        };
    }
}
```

## Example 2 (turn the trait into a visitor)

This is the same recipe as above, implemented differently.

```java
public class MicrometerNamesUseDots2 extends Recipe {
    @Override
        return "Micrometer meters should use dot separation";
    }

    @Override
        return "Micrometer naming convention automatically replaces dots with underscores," +
               "camel casing, or whatever is idiomatic depending on the monitoring system " +
               "that the metrics are going to.";
    }

    @Override
        return new MethodAccess.Matcher("io.micrometer.core.instrument.MeterRegistry *(String, ..)")
          .returns(ret -> TypeUtils.isAssignableTo("io.micrometer.core.instrument.Meter", ret))
          .asVisitor(ma -> {
              MethodCall call = ma.getTree();
              Expression name = call.getArguments().get(0);
              if (name instanceof J.Literal) {
                  J.Literal nameLiteral = (J.Literal) name;
                  String nameStr = nameLiteral.getValue().toString();
                  if (nameStr.contains("_")) {
                      call = call.withArguments(ListUtils.mapFirst(call.getArguments(),
                        n -> nameLiteral.withValue(nameStr.replace("_", "."))));
                  }
              }
              return call;
          });
    }
}
```


rewrite-core/src/main/java/org/openrewrite/trait/Trait.java
```diff
@@ -0,0 +1,36 @@
...
+import org.openrewrite.Cursor;
+import org.openrewrite.Incubating;
+import org.openrewrite.Tree;
+
+/**
+ * A trait captures semantic information related to a syntax element.
+ *
+ * @param <T> The type of the tree that this trait is related to. When
+ *            multiple specific types of tree are possible, this should
+ *            be the lowest common super-type of all the types.
```
JLLeitschuh, line 27: I was about to raise a concern about multiple specific types of trees, but I think you addressed it in the second sentence.

rewrite-core/src/main/java/org/openrewrite/trait/TraitMatcher.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.stream.Stream;
+
+@Incubating(since = "8.30.0")
+public interface TraitMatcher<U extends Trait<?>> {
+
+    Optional<U> get(Cursor cursor);
+
+    Stream<U> higher(Cursor cursor);
+
+    Stream<U> lower(Cursor cursor);
```
JLLeitschuh, line 30: I'd appreciate a comment helping understand if higher is towards the root or away from the root.

Trees in computer science are upside down from trees in nature, right? I always forget this. A doc indicating which is which would help as a reminder.

rewrite-java/src/main/java/org/openrewrite/java/trait/VariableAccess.java
```diff
@@ -0,0 +1,193 @@
...
+        @Nullable
+        @Override
+        protected VariableAccess test(Cursor cursor) {
+            if (!(cursor.getValue() instanceof J.Identifier)) {
+                return null;
+            }
+
+            J.Identifier ident = cursor.getValue();
+            String name = ident.getSimpleName();
+            if (name.equals("this") || name.equals("super") ||
```
JLLeitschuh, line 117: This is Java specific. Do you care at this point?

rewrite-java/src/test/java/org/openrewrite/java/trait/VariableAccessTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openrewrite.marker.SearchResult;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.test.RewriteTest.toRecipe;
+
+@SuppressWarnings("ALL")
+class VariableAccessTest implements RewriteTest {
```
JLLeitschuh, line 29: Do you want to add a test for the `this` and `super` cases?

JLLeitschuh:
Very interesting to see this. I'm curious to see other cases too!

JLLeitschuh:
Have you thought about tree navigation as well? `VariableAccess` -> `VariableDeclaration` for example? Or the other way around?

rewrite-java/src/main/java/org/openrewrite/java/trait/VariableAccess.java
jkschneider: Not yet. I saw the same in rewrite-analysis.

jkschneider:
> Have you thought about tree navigation as well? VariableAccess -> VariableDeclaration for example? Or the other way around?

I think the model you have in VarAccess is quite good. I don't see any reason why we can't build var access -> var declaration as either behavior or state of either.

rewrite-core/src/main/java/org/openrewrite/trait/TraitMatcher.java
jkschneider: Good point, added more docs in [ed7d875](https://github.com/openrewrite/rewrite/pull/4309/commits/ed7d8758ebbe4c56c225e37e237c6fc9d1ef8452).

rewrite-core/src/main/java/org/openrewrite/trait/Trait.java
jkschneider: Yeah, I have more thinking on this, and the most obvious case I know of right now is a trait for `JavaDependency` which could return a Gradle `J.MethodInvocation` or an Maven (XML) `Xml.Tag` which have `Tree` as a lowest common super-type. Similar for the trait of `SqlQuery`. I have a plan, just need to write it out.

rewrite-java/src/test/java/org/openrewrite/java/trait/VariableAccessTest.java
jkschneider: More tests are certainly warranted, but expensive in time to create. Wanted to do a quick check on whether or not this approach was even close to reasonable.

rewrite-core/src/main/java/org/openrewrite/trait/VisitFunction3.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.trait;
+
+import org.openrewrite.Cursor;
+import org.openrewrite.Tree;
+
+public interface VisitFunction3<U extends Trait<?>, P> {
+    Tree visit(U data, Cursor cursor, P p);
```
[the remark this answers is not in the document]
jkschneider: On second thought, `VisitFunction3` is superfluous since `Trait#getCursor` exists. Removing the whole interface in ee129b1d2b8a61ca4e18b35dddcf5c685a45512f.

rewrite-java/src/main/java/org/openrewrite/java/trait/VariableAccess.java
```diff
@@ -0,0 +1,193 @@
...
+    public boolean isWriteAccess() {
+        MaybeParenthesesPair pair = MaybeParenthesesPair.from(cursor);
+        if (pair.getParent() instanceof J.Assignment) {
+            J.Assignment assignment = (J.Assignment) pair.getParent();
+            return assignment.getVariable() == pair.getTree();
+        }
+        if (pair.getParent() instanceof J.Unary) {
+            J.Unary unary = (J.Unary) pair.getParent();
+            return unary.getExpression() == pair.getTree();
+        }
```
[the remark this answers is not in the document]
jkschneider: Thanks, fixed in 898ac13667bb70c6ac145b6e8dab9f0a58cd9a3e.

Issue #4135, opened 2024-04-15
Migrate from Gradle Enterprise gradle plugin to Develocity plugin 


timtebeek:
## What problem are you trying to solve?

> Starting with version 3.17, the plugin is available under the “Develocity” brand. Please see the [Develocity Gradle Plugin User Manual](https://docs.gradle.com/enterprise/gradle-plugin/current) for the latest documentation. Several core functionalities have been deprecated with the rebranding and will be removed in a future plugin version. Therefore, upgrading to the 3.17+ version of the plugin requires more than just updating the version number.

> Start by updating the plugin’s version to 3.17 and changing the plugin ID to com.gradle.develocity. If you are using the common-custom-user-data-gradle-plugin, update the plugin’s version to 2.0. After running a simple task like ./gradlew help you will be informed about the usage of the deprecated APIs:

> Most of the violations can be fixed by replacing references of gradleEnterprise with develocity. Some APIs have been replaced by more descriptive and idiomatic alternatives. In this example, those APIs are taskInputFiles and publishAlways(). See [Breaking API changes](https://docs.gradle.com/enterprise/gradle-plugin/legacy/#develocity_replaced_apis) section for this and other examples. After these changes, your build script should look like the following:

https://docs.gradle.com/enterprise/gradle-plugin/legacy/#develocity_migration

Bananeweizen:
duplicate of #4134. @timtebeek looks like you had a network glitch

timtebeek:
ah no they are subtly different; one is for Maven and the other is for Gradle; Both need their own recipe for the migration. Thanks for checking though!

Issue #4134, opened 2024-04-15
Migrate from Gradle Enterprise Maven plugin to Develocity plugin


timtebeek:
## What problem are you trying to solve?
https://docs.gradle.com/enterprise/maven-extension/legacy/#develocity_migration

> Starting with version 1.21, the extension is available under the “Develocity” brand. Please see the [Develocity Maven Extension User Manual](https://docs.gradle.com/enterprise/maven-extension/current) for the current documentation. Several APIs, system properties and configuration files have been deprecated with the rebranding and will be removed in the next major version. Therefore, please consider migrating to the new APIs where applicable to minimize the amount of necessary changes when updating to the next major version.

> Start by updating the extension’s version to 1.21 and changing the artifact ID to develocity-maven-extension, then rename gradle-enterprise.xml to develocity.xml and update your .gitignore accordingly. If you are using the common-custom-user-data-maven-extension, update the extension’s version to 2.0 After running a goal like ./mvnw test you will be informed about the usage of the deprecated APIs:

> Most of the violations can be fixed by replacing references of gradleEnterprise with develocity. Some APIs have been replaced by more descriptive and idiomatic alternatives. In this example, those APIs are goalInputFiles and publish. See [Breaking API changes](https://docs.gradle.com/enterprise/maven-extension/legacy/#develocity_replaced_apis) section for this and other examples. After these changes, your extensions.xml and develocity.xml should look like the following:

PR #4312, opened 2024-07-09
Migrate add Develocity recipes to handle the new develocity configurations

shanman190:
## What's changed?
- [x] When adding Develocity to a new Gradle project make sure to use the new Develocity configuration names as of 3.17.x
- [x] When adding Develocity to a new Maven project make sure to use the new Develocity configuration names as of 3.17.x
- [ ] Migrate an existing Gradle project configured with Develocity or Gradle Enterprise < 3.17 to use the new configurations for 3.17.x and newer (moved to #4320)
- [ ] Migrate an existing Maven project configured with Develocity or Gradle Enterprise < 3.17 to use the new configurations for 3.17.x and newer

## What's your motivation?
Gradle Enterprise was renamed to Develocity and as such the Gradle plugin has been equally renamed including with a Maven POM relocation. Additionally changes to the configuration names have been introduced as part of the product rename and must be taken by end users.

https://docs.gradle.com/enterprise/gradle-plugin/legacy/#develocity_migration
https://docs.gradle.com/enterprise/maven-extension/legacy/#develocity_migration

Not in particular.

Do the migration manually

## Any additional context
* #4134 
* #4135

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddDevelocityGradlePlugin.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddDevelocityGradlePlugin.java
@@ -144,7 +144,7 @@ public class AddDevelocityGradlePlugin extends Recipe {
                     return cu;
                 }
                 // Don't modify an existing gradle enterprise DSL, only add one which is not already present
-                if (containsGradleEnterpriseDsl(cu)) {
+                if (containsGradleDevelocityDsl(cu)) {
                     return cu;
                 }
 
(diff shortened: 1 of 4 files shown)


timtebeek:
Great start! Let me know if you'd like a review & potential merge already, or whether you'd like to continue on this branch for a little while longer. The draft status had me thinking the latter, but the tasks outlined at the top might better fit into separate PRs, so then it's slightly unclear what the exact status of this PR is.

shanman190:
@timtebeek, will do! I was thinking about it a little bit and I think a good split may be on finishing up the updates required for both "add to new project", then split the migrate off to another PR. It could possibly be better to even split the migrates up into two separate PRs for each of Gradle and Maven, but I'm not sure yet. 

I just wanted to get something out there to kick off some discussion, if needed, and elevate awareness that this is coming and will soon be required to have been completed by Develocity users.

timtebeek:
Both of those splits sound good to me; let's see if @jean-andre-gauthier has anything to add to this already, and continue down this path.

shanman190:
@timtebeek, I'm marking this ready for review as now the add recipes will correctly use the non-deprecated Develocity configurations.

I'm going to start on the migration recipe for Gradle in a new PR, so keep an eye out for that coming shortly.

timtebeek (approved):
Thanks a lot! One thought that crossed my mind is when we would phase out the "Add Gradle enterprise" parts of these recipes. Is there a good reason to keep those around? And if so; should that be an option on these existing recipes, or would a rebranded copy of these recipes make sense, such that it's easier to remove the old recipes down the line?

shanman190:
@timtebeek, that is a very valid question and I'm not sure on the answer there. From a Develocity standpoint, historically any older plugin version worked with a new server version, but the Gradle version would sometimes determine which variation you'd use. A new plugin version will often have a minimum required server version otherwise it will log an error stating that it is incompatible with the current server version. 

Maybe the folks over at Gradle can provide some more guidance on their expectations? I had considered dropping the old support, but I didn't have the answer that you're looking for either.

https://docs.gradle.com/develocity/compatibility/#gradle_build_tool_compatibility
https://docs.gradle.com/develocity/compatibility/#gradle_build_tool_compatibility

erichaagdev:
I suggest leaving the _Add Gradle Enterprise_ recipe for now. @shanman190 is right – Develocity installations have a maximum supported plugin version and it's not uncommon for some organizations to be running older Develocity versions. In those cases, they will still need to apply the Gradle Enterprise plugin.

As a point of reference, the Develocity plugin is only compatible with Develocity 2024.1+ which was only released April 2, 2024. It will take some time before the majority of users are on 2024.1+.

shanman190:
@erichaagdev, awesome! Thanks for the insight!


The current `AddDevelocityGradlePlugin` and `AddDevelocityMavenExtension` will each apply their respective Gradle Enterprise configurations when the chosen plugin version is < 3.17, then use the Develocity configuration for 3.17+.

erichaagdev:
Ah, thanks for the clarification. If they do the same thing then I don't see a reason to keep the Gradle Enterprise one.

PR #4324, opened 2024-07-12
MergeYaml should be able to merge list items

timtebeek:
## What's changed?
Add a test to show MergeYaml should be able to merge list items.

## What's your motivation?
Currently the recipe fails to make any changes.

if on the input the steps list is missing, then the new item is added.

## Any additional context
As reported by a user.

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
@@ -1135,4 +1135,32 @@ class MergeYamlTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void insertElementIntoExistingSequence() {
+        rewriteRun(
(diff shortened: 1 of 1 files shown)


shanman190:
So personally I think this is bound to result in issues... If we take note of Spring Boot's history, they used to allow merging of lists, but abandoned it for being fragile and sometimes users want to change the list entirely. Now instead, they opt to simply do full list replacement instead.

timtebeek:
Thanks yes that might be the more straightforward way of handling this. As you also indicated one on one we also have an AppendToSequence recipe: https://docs.openrewrite.org/recipes/yaml/appendtosequence
That one currently takes a String value, but we could allow complex objects there for a clear choice between either replace, or append. Something to consider as we work through what's needed here.

timtebeek:
This test actually passes provided we pass in `objectIdentifyingProperty: run`. Closing this as invalid.

PR #4326, opened 2024-07-12
feat: add shell executor interface for running shell commands during recipe execution

natedanner:
## What's changed?
Adding a default executor for running npm commands in code mod recipes.

## What's your motivation?
Standardize execution of npm commands in codemod recipes and allow for customization of the environment the commands are running in. The moderne platform needs to be able to customize the environment when running codemods recipes on the moderne platform to allow custom registries.

Duplicate shell running code in these code mod repos. 
https://github.com/moderneinc/rewrite-codemods/blob/main/src/main/java/org/openrewrite/codemods/NodeBasedRecipe.java#L111 
https://github.com/moderneinc/rewrite-codemods-ng/blob/main/src/main/java/org/openrewrite/codemods/migrate/angular/NodeBasedRecipe.java#L150

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/shell/exec/ShellExecutor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/shell/exec/ShellExecutor.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.shell.exec;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Incubating;
+import org.openrewrite.scheduling.WorkingDirectoryExecutionContextView;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+@Incubating(since = "8.30.0")
+@SuppressWarnings("unused")
+public interface ShellExecutor {
+
+    @SuppressWarnings("unused")
+    default void init() {
+    }
+
+    @SuppressWarnings("unused")
+    default Path exec(List<String> command, Path workingDirectory, Map<String, String> environment, ExecutionContext ctx) {
+        return exec(command, workingDirectory, environment, Duration.ofMinutes(5), ctx);
+    }
+
+    default Path exec(List<String> command, Path workingDirectory, Map<String, String> environment, Duration timeout, ExecutionContext ctx) {
+        Path stdOut = null, stdErr = null;
+        try {
+            ProcessBuilder builder = new ProcessBuilder();
+            builder.command(command);
+            builder.directory(workingDirectory.toFile());
+            builder.environment().putAll(environment);
+
+            stdOut = Files.createTempFile(WorkingDirectoryExecutionContextView.view(ctx).getWorkingDirectory(), "shell",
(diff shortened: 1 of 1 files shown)


rewrite-core/src/main/java/org/openrewrite/NpmExecutionContextView.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
```
timtebeek, line 16: I wonder if `rewrite-core`, and package `org.openrewrite` are the best place for this NPM specific context. We seem to be spreading our frontend related classes across `rewrite-javascript`, `rewrite-nodejs`, `rewrite-codemods` and `rewrite-codemods-ng`. Adding `rewrite-core` to that list for something NPM specific might then further create confusion as to what goes where. Would it make sense to move the NPM classes elsewhere, and only keep the `ShellExecutor` in rewrite-core?

rewrite-core/src/main/java/org/openrewrite/shell/exec/NpmExecutor.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.nio.file.Path;
+
+@SuppressWarnings("unused")
+public class NpmExecutor implements ShellExecutor {
+    public void init(Path dir) {
+    }
+
+    public void postExec(Path dir) {
+    }
```
timtebeek, line 26: This almost seems to override `postExec()` in `ShellExecutor`, but not quite as that one does not take a path argument. Should this be an override? Should this call `super.postExec()`?

rewrite-core/src/main/java/org/openrewrite/shell/exec/NpmExecutor.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.shell.exec;
+
+import java.nio.file.Path;
+
+@SuppressWarnings("unused")
+public class NpmExecutor implements ShellExecutor {
+    public void init(Path dir) {
```
timtebeek, line 22: Same as below for `postExec`, this is not quite an override, but feels like it maybe should be, depending on what you were after.

rewrite-core/src/main/java/org/openrewrite/NpmExecutionContextView.java
natedanner: Yes, we'll move the npm specific code to rewrite-nodejs

rewrite-core/src/main/java/org/openrewrite/shell/exec/ShellExecutor.java
```diff
@@ -0,0 +1,80 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+@org.openrewrite.Incubating(since = "8.30.0")
+public interface ShellExecutor {
+
+    @SuppressWarnings("unused")
+    default void init() {}
```
timtebeek, line 32: Did you want to add the `Path dir` here and below in `postExec()`that you had looked to want to override?

rewrite-core/src/main/java/org/openrewrite/shell/exec/ShellExecutor.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @SuppressWarnings("unused")
+    default Path exec(List<String> command, Path workingDirectory, Map<String, String> environment, ExecutionContext ctx) {
+        return exec(command, workingDirectory, environment, ctx, 5L);
+    }
+
+    default Path exec(List<String> command, Path workingDirectory, Map<String, String> environment, ExecutionContext ctx, Long timeoutMinutes) {
```
timtebeek, line 41: I'd prefer a Duration for the timeout here, as opposed to a Long; and elsewhere we're fairly consistent in having `ExecutionContext` be the last argument; let's do that here as well.

natedanner: Will do. Thanks @timtebeek!

timtebeek (approved):

PR #4327, opened 2024-07-12
Gradle UpgradeDependencyVersion visitor should override isAcceptable

timtebeek:
## What's changed?
We returned a TreeVisitor from `getVisitor(Accumulator)`, which then implicitly returns true, meaning we incorrectly skipped the Maven visitor for `pom.xml` files, calling the Gradle visitor instead. https://github.com/openrewrite/rewrite-java-dependencies/blob/2e3f39c4863b943cdaacb39d0e788c9c73e3f949/src/main/java/org/openrewrite/java/dependencies/UpgradeTransitiveDependencyVersion.java#L128-L133

With this change we only return true for Groovy and Properties files, as we should.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-java-dependencies/pull/106

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -301,6 +301,16 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
     public TreeVisitor<?, ExecutionContext> getVisitor(DependencyVersionState acc) {
         return new TreeVisitor<Tree, ExecutionContext>() {
             @Override
+            public boolean isAcceptable(SourceFile sf, ExecutionContext ctx) {
+                if (sf instanceof Properties) {
+                    return new UpdateProperties(acc).isAcceptable(sf, ctx);
+                } else if (sf instanceof G.CompilationUnit) {
+                    return new UpdateGroovy(acc).isAcceptable(sf, ctx);
+                }
+                return false;
+            }
+
+            @Override
             public @Nullable Tree visit(@Nullable Tree t, ExecutionContext ctx) {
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -300,6 +300,16 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
...
     public TreeVisitor<?, ExecutionContext> getVisitor(DependencyVersionState acc) {
         return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sf, ExecutionContext ctx) {
+                if (sf instanceof Properties) {
+                    return new UpdateProperties(acc).isAcceptable(sf, ctx);
+                } else if (sf instanceof G.CompilationUnit) {
+                    return new UpdateGroovy(acc).isAcceptable(sf, ctx);
+                }
+                return false;
```
sambsnyd, line 310:
```suggestion
                return new UpdateProperties(acc).isAcceptable(sf, ctx) || new UpdateGroovy(acc).isAcceptable(sf, ctx);
```

We should also use `isAcceptable()` rather than `instanceof` checks in the `visit()` override, since hypothetically those recipes could further customize `isAcceptable()`

sambsnyd (approved):
Good improvement

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
timtebeek: Hope I understood that correctly; Applied in fcf6032cd0a1871207b2f128fddde7a4506c9609

PR #4334, opened 2024-07-16
Exclude Maven built-in property from circular placeholder reference check

nguyenhoan:
## What's changed?
Exclude Maven built-in property from circular placeholder reference check

## What's your motivation?
Error `Circular placeholder reference 'XYZ' in property definitions` is thrown when a built-in property, such as `project.version`, is redefined in the root POM and used in a child in multi-module projects.
I would not expect running recipes to fail given that `mvn clean verify` succeeds.

Attached is a minimum project to reproduce the issue.
[Circular-placeholder-reference.zip](https://github.com/user-attachments/files/16256934/Circular-placeholder-reference.zip)
You can run any recipe to trigger it. I ran the one for SpringBoot
```
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2 
```

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.internal;
 
+import org.openrewrite.internal.lang.NonNull;
 import org.openrewrite.internal.lang.Nullable;
 
 import java.util.*;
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.internal;
 
+import org.openrewrite.internal.lang.NonNull;
```
timtebeek, line 18:
```suggestion
```

rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
```diff
@@ -160,4 +170,13 @@ private static boolean substringMatch(CharSequence str, int index, CharSequence
         }
         return true;
     }
+
+    private static boolean isMavenBuiltInProperty(@NonNull final String originalPlaceholder) {
```
timtebeek, line 174: We've already annotated the package; that means we likely don't have to repeat ourselves here.
https://github.com/openrewrite/rewrite/blob/86ae0f10e668b825fb1379258c3924f55980df97/rewrite-core/src/main/java/org/openrewrite/internal/package-info.java#L16-L19

```suggestion
    private static boolean isMavenBuiltInProperty(final String originalPlaceholder) {
```

timtebeek (commented):
Thanks for the quick work here! I'd expected a unit test using RewriteTest that shows we can now handle a parent and child that use such properties, perhaps very similar to what you already have in your zip file. I can explore that if you don't get to it for a more thorough review later.

nguyenhoan:
That does not sound a like unit test for the changed method. If I wanted to add a unit test, which test class would I add it to?

timtebeek:
I'd say `MavenParserTest` is the best place to guarantee that your issue is fixed, and remains fixed with a test similar to this one.
https://github.com/openrewrite/rewrite/blob/ad91222a2869e7baba013394c80ea349822397e7/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java#L3042-L3109

nguyenhoan:
Thank for the suggestion! Working on adding a test there helped me find the correct strategy.

timtebeek (approved):
Great to see the added test & fix @nguyenhoan !

And I even got the review done in time on the train. ;)

PR #4336, opened 2024-07-17
Allow JavaVisitors to operate on the J language island

kmccarp:
## What's changed?
Allow JavaVisitor and JavaIsoVisitor to operate on any `SourceFile` in the `J` language grammar island (e.g. C# soon)

## What's your motivation?
To be able to run many of the existing recipes on C# files.

Any potential risks of operating on all SourceFiles of type J as I'm not familiar with the havoc this could wreak.

@sambsnyd @timtebeek 

No

## Any additional context

N/A

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
@@ -33,7 +33,7 @@ public class JavaVisitor<P> extends TreeVisitor<J, P> {
 
     @Override
     public boolean isAcceptable(SourceFile sourceFile, P p) {
-        return sourceFile instanceof JavaSourceFile;
+        return sourceFile instanceof J;
     }
 
     @Override


timtebeek:
Adding a couple links to sources here such that we have them all in view for the review.

## Java
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L54
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L1415
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaSourceFile.java#L32

## Groovy
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java#L43
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java#L72

## Kotlin
https://github.com/openrewrite/rewrite-kotlin/blob/052dbf09b1268f59dbbf25cdff231d082f602186/src/main/java/org/openrewrite/kotlin/tree/K.java#L56
```java
public interface K extends J {
```
https://github.com/openrewrite/rewrite-kotlin/blob/052dbf09b1268f59dbbf25cdff231d082f602186/src/main/java/org/openrewrite/kotlin/tree/K.java#L86
```java
    final class CompilationUnit implements K, JavaSourceFile, SourceFile {
```

timtebeek:
Given the above and the code change made I don't immediately foresee any issues. Everything in the J grammar island had already been extending both J and JavaSourceFile. Cs would be the first to extend J, but not JavaSourceFile.

At most the methods we have on JavaSourceFile would not be available anywhere we assume there to be a JavaSourceFile.
So calls to something like this service to get annotations would fail, but I suspect we'd come across those quickly and exclusively on Cs.
https://github.com/openrewrite/rewrite/blob/f66b109c7d0c4bc96cb40277700f0b44d47206b7/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaSourceFile.java#L67

timtebeek:
As an example; this recipe would fail, as it assumes to set a field unconditionally used even when not in a JavaSourceFile.
https://github.com/openrewrite/rewrite-static-analysis/blob/6f251ab96ecd337fed6395f8047c3f30ab42a9d6/src/main/java/org/openrewrite/staticanalysis/OperatorWrap.java#L47-L63

The solution there is likely to stop traversal when not in a JavaSourceFile.

jkschneider:
This is being closed in favor of ensuring that the root LST element of J family language extensions extend `JavaSourceFile`.

Issue #4333, opened 2024-07-16
Gradle UpgradeDependencyVersion omits versionPattern if exact newVersion specified


gideon-sunbit:
When running _org.openrewrite.gradle.UpgradeDependencyVersion_ specifying both an exact _newVersion_ (e.g. `32.1.1`) and a _versionPattern_ (e.g. `-jre`)
The selected version does not include the versionPattern (i.e. returns `32.1.1` instead of `32.1.1-jre`)

## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.30.0
- Gradle plugin v6.16.4

## How are you running OpenRewrite?

I am using the Gradle plugin via the IntelliJ IDEA integration (2024.1.4)
Public repo with example: https://github.com/gideon-sunbit/openrewrite-examples-public

## What is the smallest, simplest way to reproduce the problem?

**rewrite.yml**
```yaml
---
type: specs.openrewrite.org/v1beta/recipe
name: com.yourorg.UpgradeDependencyVersionExample
displayName: Upgrade Gradle dependency versions example
recipeList:
  - org.openrewrite.gradle.UpgradeDependencyVersion:
      groupId: com.google.guava
      artifactId: guava
      newVersion: 32.1.1
      versionPattern: '-jre'
```

**build.gradle**
```groovy
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

dependencies {
    implementation('com.google.guava:guava:29.0-jre')
}
```

## What did you expect to see?

```groovy
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

dependencies {
    implementation('com.google.guava:guava:32.1.1-jre')
}
```

## What did you see instead?

```groovy
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

dependencies {
    implementation('com.google.guava:guava:32.1.1')
}
```

## Execution log:

```
21:55:23: Executing 'rewriteRun'...

Download https://plugins.gradle.org/m2/org/openrewrite/plugin/maven-metadata.xml, took 1 s 55 ms
Download https://plugins.gradle.org/m2/org/openrewrite/plugin/6.16.4/plugin-6.16.4.pom, took 1 s 298 ms
Download https://plugins.gradle.org/m2/org/openrewrite/plugin/6.16.4/plugin-6.16.4.module, took 869 ms
[...]

PR #4339, opened 2024-07-18
Fix UpgradeDependencyVersion handling of exactVersion with versionPattern

gideon-sunbit:
## What's changed?

- Add new VersionComparator implementation: _ExactVersionWithPattern_
- Add _ExactVersionWithPattern_ to _SemVer_ validation chain
- Add tests

## What's your motivation?

- Issue https://github.com/openrewrite/rewrite/issues/4333

Because _SemVer_ is in rewrite-core, any change there has a greater risk (i.e. a larger "blast radius" in case of a bug) since it is used in other places other than the Gradle UpgradeDependencyVersion recipe.

Yes, A workaround for this issue is to specify the "version pattern" as part of the newVersion (e.g. "newVersion: '32.1.1-jre'").

Also, It may be possible to fix this issue in a more limited way inside [rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyVersionSelector.java](https://github.com/openrewrite/rewrite/blob/537d0237d8e71e0f9567cb6976f8b263f18e27b8/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyVersionSelector.java#L131)
Although it would still probably require adding a new VersionComparator implementation (since ExactVersion doesn't take into account the versionPattern)

## Any additional context

Technically, this issue may not be considered a bug, and the current behavior may be what the original authors have intended.

However, as a user, I find it a bit unintuitive that an XRange `newVersion` with a `versionPattern` (e.g. "32.1.X", "-jre") behaves as expected, while an exact version (e.g. "32.1.1", "-jre") does not.
So by the "principle of least surprise" I would expect ("32.1.1", "-jre") to behave similar to ("32.1.X", "-jre")
(Or fail recipe validation with "versionPattern cannot be specified with an exact version" or a similar message).
But I may be misunderstanding the code or the documentation.

### Checklist


timtebeek:
Thanks for the work here @gideon-sunbit ! At first glance this looks good; I want to take a closer look after I get back home Sunday before a merged though. I'll keep you posted!

timtebeek:
Great start @gideon-sunbit ; probably much like yourself I'm mostly wondering why the `ExactVersion` class does not handle metadata patterns already. I'm wondering if that might be the better approach here, seeing how we have explicit handling for `ExactVersion` in a number of places, to stop resolving versions against external sources. 🤔 

https://github.com/openrewrite/rewrite/blob/14aef405998f4a07a034c7b87deb3d359b3bb0ae/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java#L299-L302

https://github.com/openrewrite/rewrite/blob/14aef405998f4a07a034c7b87deb3d359b3bb0ae/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependencyVisitor.java#L190-L194

I'll see what the impact would be if we instead add a pattern field to ExactVersion...

timtebeek:
The more I look at this the more it's clear you've provided a well measured approach to handling a potentially regex metadata pattern. Thanks!

timtebeek (approved):

Issue #4246, opened 2024-06-12
Json formatting/style doesn't seem to be respected when adding values/members


dpozinen:
## What version of OpenRewrite are you using?

- OpenRewrite 2.12.0
- Maven plugin 5.23.3

## Context
I have created a recipe that adds key+value into a json. It is done via `Json.Literal` and is inspired by `json.ChangeValue`. Here's the relevant code:

```java
JsonRightPadded<JsonKey> newKey = new JsonRightPadded<>(
    new Json.Literal(randomId(), Space.EMPTY, Markers.EMPTY, "\"" + key + "\"", key),
    Space.EMPTY,
    Markers.EMPTY
);
Json.Literal newValue = new Json.Literal(randomId(), Space.build(" ", emptyList()), Markers.EMPTY, value, unQuote(value));
Json newMember = new Json.Member(randomId(), space, Markers.EMPTY, newKey, newValue);

List<Json> members = jsonIsEmpty ? new ArrayList<>() : obj.getMembers();
members.add(newMember);
return obj.withMembers(members);
```

## Problem
* I don't seem to be able to control the padding after the value is inserted
* If this isn't the only value in the object, the automatic insertion of the comma is on its own line.

My test:
```java
...
rewriteRun(
        spec -> spec.recipe(new AddKeyValue(
            "$.", "key", "\"val\""
        )),
        //language=json
        json("""
                {
                    "l": [ 1, 2 ]
                }
            """,
            """
                {
                    "l": [ 1, 2 ],
                    "key": "val"
                }
            """)
    );
...
```
fails with:
```
expected: 
  "{
      "l": [ 1, 2 ],
      "key": "val"
  }"
 but was: 
  "{
      "l": [ 1, 2 ]
  ,
      "key": "val"}"
```

The same thing happens with Arrays and their Values.

## Expected Behaviour
Can be seen in the test failure:
* When adding an element the comma is added on the same line as the previous element.
* The closing bracket is not moved from it's initial position

In general I think this is a styling issue, which I understand is not supported for json? I could not find any `json.style.*` classes, pretty sure `TabsAndIndentsStyle` for xml solves this issue. CMIIW. Thanks 🙂

PR #4341, opened 2024-07-18
feat: recipe for adding a key value pair to a json

dpozinen:
## What's changed?
Added recipe to add a new key and value to a json.

## What's your motivation?
Currently, we can rename keys and change values, but can't add new entries to a json

- Fixes https://github.com/openrewrite/rewrite/issues/4246


rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java
```diff
@@ -0,0 +1,132 @@
...
+                return true;
+            }
+
+            private boolean keyMatches(JsonKey jsonKey, String key) {
+                if (jsonKey instanceof Json.Literal) {
+                    return key.equals(((Json.Literal) jsonKey).getValue());
+                } else if (jsonKey instanceof Json.Identifier) {
+                    return key.equals(((Json.Identifier) jsonKey).getName());
+                }
+                throw new IllegalStateException("Key is not 'Json.Literal' or 'Json.Identifier': " + jsonKey);
```
dpozinen, line 127: hey guys so I'm not sure about this peace of code, I've only ever encountered `Json.Literal` as a key in testing, but `Json.Identifier` also implements `JsonKey` so that's why it's here. lmc if this is redundant

timtebeek: Whenever we're uncertain of a change, we fallback to making no change at all. That way we ensure we [do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices).

timtebeek (commented):
Had you considered `MergeYaml` instead of this add key-value pair? I think that can already achieve the same, and it would be less code to maintain for us going forward.

timtebeek (commented):
Took me a little while longer to get to this one; looking quite good already. One thing I would still like to change is the formatting used when we append to existing elements. Right now that retains the newline before the comma introduced, where I think this will look better if that comma is on the same line as the line before it. Do you see an option to work that in still @dpozinen ?

timtebeek:
I only just now see you had logged an issue along those exact lines; thanks again! It's not always easy to keep up. :)
- https://github.com/openrewrite/rewrite/issues/4246

dpozinen:
> I only just now see you had logged an issue along those exact lines; thanks again! It's not always easy to keep up. :)
> 
> * [Json formatting/style doesn't seem to be respected when adding values/members #4246](https://github.com/openrewrite/rewrite/issues/4246)

Hi, yeah, I'm not entirely sure where to start with this issue, do I need to create a new style? I'll see when I have the time to get more in depth into how rewrite _writes_ the LST, because the comma isn't part of spacing/padding of the element and is added somewhere internally

I also see you made some changes to the logic [namely here](https://github.com/openrewrite/rewrite/pull/4341/files#diff-7ace4ba4613b641d540b10bcfb7c10c0b8d892ffa8a328fd5002685ecaaa7a5fR85). Not sure why when appending (`prepend == false`) to a non empty json (`jsonIsEmpty == false`) we are using `SINGLE_SPACE` instead of the space of the first element of that json?

greg-at-moderne:
Thanks for the contribution @dpozinen.
I have taken your code, added the formatting detection logic and packaged that into #4916.

PR #4349, opened 2024-07-22
Do not wrap single active recipe & remove `activateAll()`

timtebeek:
## What's changed?
Remove `activateAll()` and do not wrap single active recipe into a new `CompositeRecipe`.

## What's your motivation?
- Removes need for edge case handling unwrap in https://github.com/openrewrite/rewrite-maven-plugin/pull/816
- Closer to what folks would expect in the simplest case: the directly instantiated Recipe, not an indirection wrapper.

The CompositeRecipe overrides a number of methods like the estimated time and data table descriptors. I think it's fine to skip this indirection, but an extra pair of eyes can't hurt.
https://github.com/openrewrite/rewrite/blob/03c458404073082275a752063887c188d213afbe/rewrite-core/src/main/java/org/openrewrite/config/CompositeRecipe.java#L25-L71

We can leave this indirection in for consistent handling no matter the number of active recipes.

## Any additional context
- As discovered on https://github.com/openrewrite/rewrite-maven-plugin/pull/816

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -161,6 +161,9 @@ public class Environment {
                     String.join(", ", suggestions));
             throw new RecipeException(message);
         }
+        if (activatedRecipes.isEmpty()) {
+            return Recipe.noop();
+        }
         if (activatedRecipes.size() == 1) {
             return activatedRecipes.get(0);
         }
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -150,26 +150,25 @@ public class Environment {
             }
         }
         if (!recipesNotFound.isEmpty()) {
+            @SuppressWarnings("deprecation")
             List<String> suggestions = recipesNotFound.stream()
                     .map(r -> recipesByName.keySet().stream()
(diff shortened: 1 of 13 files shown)


rewrite-core/src/main/java/org/openrewrite/config/Environment.java
```diff
@@ -150,28 +150,27 @@ public Recipe activateRecipes(Iterable<String> activeRecipes) {
...
-            String message = String.format("Recipes not found: %s\nDid you mean: %s",
+            String message = String.format("Recipe(s) not found: %s\nDid you mean: %s",
                     String.join(", ", recipesNotFound),
                     String.join(", ", suggestions));
             throw new RecipeException(message);
         }
+        if (activatedRecipes.size() == 1) {
+            return activatedRecipes.get(0);
+        }
         return new CompositeRecipe(activatedRecipes);
```
timtebeek, line 167: We could return `Recipe.noop()` here if the actived recipes is empty, because an empty array was passed into `activateRecipes(String... activeRecipes)` below.

pstreef: That sounds better to me as well.

pstreef (approved):

timtebeek:
@pstreef it seems like this indirection was masking missing descriptions in our test recipes; are you still ok with this updated PR?
I imagine we'll see this issue pop up some more downstream, although it seems like we already had similar unpacking on the  Moderne side of things.

pstreef (approved):

PR #4354, opened 2024-07-25
Fix ChangeDependency to handle unknown configurations

ryanwalker:
## What's changed?
ChangeDependency does not work on configurations it does not understand. This one from my project gets skipped:
```
  integrationTestImplementation "com.github.tomakehurst:wiremock-jre8-standalone:2.+"
```
  I created a GradleDependency Trait to solve this.

### Checklist


shanman190:
With the recent introduction of the `Trait` API into rewrite proper, we can instead make use of that to provide an even stronger guarantee that the method invocation is a Gradle configuration. 

We could become as strict as to only accept dependency configurations that are meant for holding dependencies by not being consumable or resolvable.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -67,6 +67,43 @@ void relocateDependency() {
...
+              plugins {
+                  id "java-library"
+              }
+              
+              repositories {
+                  mavenCentral()
+              }
+              
+              dependencies {
+                  swaggerCodegen "commons-lang:commons-lang:2.6"
```
ryanwalker, line 85: Erroring because of this config.  But it works in another test.

shanman190:
This seems to be getting closer. I'm going to also share the commit with which the `MavenDependency` trait was introduced which should show a bit better all of the components in place and how it's utilized.
https://github.com/openrewrite/rewrite/commit/c24e9920b509a79a9189402095b905f6ccbd1cd4

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -0,0 +1,88 @@
...
+    public boolean isDependency() {
+        return isLikelyDependencyConfiguration(cursor);
+    }
+
+    public boolean isNotDependency() {
+        return !isDependency();
+    }
+}
+
+
```
timtebeek, line 88:
```suggestion
}

```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -0,0 +1,102 @@
...
+    public boolean isDependency() {
+        return isLikelyDependencyConfiguration(cursor);
+    }
+
+    public boolean isNotDependency() {
+        return !isDependency();
+    }
+}
+
```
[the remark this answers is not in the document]
timtebeek:
```suggestion
}
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        // There are cases where the configurations don't populate correctly.
+        if (gradleProject.getNameToConfiguration().isEmpty()) {
+            return isLikelyDependencyConfiguration;
+        }
+
+        if (isLikelyDependencyConfiguration) {
+            return gradleProject.getNameToConfiguration().keySet().stream()
+                    .anyMatch(configurationName -> methodInvocation.toString().startsWith(configurationName));
```
ryanwalker, line 81: testing in my project, the list of configurations is not always complete.  It only had a few items in somre cases.  Didnt have time to look into it much yrt.

ryanwalker:
All tests are passing now. Its still not prod ready yet but ill clean it up early this week.  Let me know if theres anything glaring that anyone thinks should be done differently.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -0,0 +1,181 @@
...
+
+@Value
+public class GradleDependency implements Trait<J.MethodInvocation> {
+    private static final MethodMatcher DEPENDENCY_DSL_MATCHER = new MethodMatcher("* dependencies(groovy.lang.Closure)");
+    private static final MethodMatcher DEPENDENCY_CONFIGURATION_MATCHER = new MethodMatcher("DependencyHandlerSpec *(..)");
+
+
+    Cursor cursor;
+
+    public @Nullable ResolvedGroupArtifactVersion getResolvedGroupArtifactVersion() {
```
shanman190, line 52: This method body should really be within the `Matcher#test(Cursor)` method.

rewrite-gradle/src/main/java/org/openrewrite/gradle/tree/ResolvedGroupArtifactVersion.java
```diff
@@ -0,0 +1,63 @@
...
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.maven.tree.GroupArtifact;
+import org.openrewrite.maven.tree.GroupArtifactVersion;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+@Value
+@With
+public class ResolvedGroupArtifactVersion implements Serializable {
```
shanman190, line 29: From a Gradle perspective, the dependency declaration is a requested dependency, but not necessarily the same as a resolved dependency.

This class presently has the same fields as `GroupArtifactVersion` from rewrite-maven.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -0,0 +1,181 @@
...
+                return new ResolvedGroupArtifactVersion(null, artifact, null);
+            } catch (Exception e) {
+                // todo log it
+                return null;
+            }
+        }
+        return null;
+    }
+
+    public static class Matcher extends SimpleTraitMatcher<GradleDependency> {
```
shanman190, line 146: I think we should have the Gradle-specific implementation of the trait matched built and then use that for confirming the dependency configurations rather than trying to guess at them.

shanman190:
@ryanwalker, I went ahead and extracted the `Trait` api out from this PR to make this task a little simpler. You can see that PR here: https://github.com/openrewrite/rewrite/pull/4394.

PR #4355, opened 2024-07-25
feat: Extend message on http sender response

pstreef:
Adding a body when it makes sense (server/client exceptions) and cause message if available (although this is unused in this repo).


pstreef:
this has some uninteded side-effects. Going to close it for now

PR #4356, opened 2024-07-25
fix: Dont generate the same file multiple times & fix CreateTextFile

pstreef:
## What's changed?
generate no longer creates files at the same path multiple times

## What's your motivation?
Currently the same file can be created multiple times, causing instabilities in recipe execution. For instance
running `CreateTextFile` twice in the same run causes multiple cycles changing back and forth from 1 text to the other, even if `overwriteExisting` is false

This is a change in the core recipe running and might have unwanted side-effects

@sambsnyd @timtebeek 

Fixing it on the recipe level alone, but this would require "rescanning" during the edit phase.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -112,7 +112,11 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                 ScanningRecipe<Object> scanningRecipe = (ScanningRecipe<Object>) recipe;
                 List<SourceFile> generated = new ArrayList<>(scanningRecipe.generate(scanningRecipe.getAccumulator(rootCursor, ctx), unmodifiableList(acc), ctx));
                 generated.replaceAll(source -> addRecipesThatMadeChanges(recipeStack, source));
-                acc.addAll(generated);
+                for (SourceFile source : generated) {
+                    if (acc.stream().noneMatch(s -> s.getSourcePath().equals(source.getSourcePath()))) {
+                        acc.add(source);
+                    }
+                }
                 if (!generated.isEmpty()) {
                     madeChangesInThisCycle.add(recipe);
                 }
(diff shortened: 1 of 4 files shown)


okundzich:
what is the issue at walmart?

rewrite-core/src/main/java/org/openrewrite/text/CreateTextFile.java
```diff
@@ -86,7 +86,7 @@ public TreeVisitor<?, ExecutionContext> getVisitor(AtomicBoolean created) {
             @Override
             public SourceFile visit(@Nullable Tree tree, ExecutionContext ctx) {
                 SourceFile sourceFile = (SourceFile) requireNonNull(tree);
-                if ((created.get() || Boolean.TRUE.equals(overwriteExisting)) && path.equals(sourceFile.getSourcePath())) {
```
timtebeek, line 89: Weird how we were overwriting the content again, even after already initializing with file contents.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -112,7 +112,11 @@ public LSS generateSources(LSS sourceSet) {
                 ScanningRecipe<Object> scanningRecipe = (ScanningRecipe<Object>) recipe;
                 List<SourceFile> generated = new ArrayList<>(scanningRecipe.generate(scanningRecipe.getAccumulator(rootCursor, ctx), unmodifiableList(acc), ctx));
                 generated.replaceAll(source -> addRecipesThatMadeChanges(recipeStack, source));
-                acc.addAll(generated);
+                for (SourceFile source : generated) {
+                    if (acc.stream().noneMatch(s -> s.getSourcePath().equals(source.getSourcePath()))) {
+                        acc.add(source);
+                    }
+                }
```
timtebeek, line 119: Gave me a bit of pause to loop over all generated files in a list before adding a new entry, but I think it's tolerable given the expected low volume of generated files.

timtebeek (approved):

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
jkschneider: Change the reduction operation to use a `Set` generated from a `TreeMap` whose comparator is the source path.

pstreef: I updated to using `TreeSet`. This does mean that the insertion order is no longer guaranteed to be the resulting order, although I can imagine that the LST order is also similarly ordered anyway?

PR #4358, opened 2024-07-25
fix: `MergeYaml` block indentation not as expected

pstreef:
I think I have identified a bug in the yaml parser + writer.
when we parse the following yaml:
```
some:
  object:
    script: |
      #!/bin/bash
      echo "hello
```
the value of the script mapping is:
```
      #!/bin/bash
      echo "hello
```
when writing this yaml it correctly writes back as the original because of this.
However, when we want to insert this yaml somewhere into another document the prefix is not taken into account. So if I were to insert
```
script: |
  #!/bin/bash
  echo "hello
```
into the same spot I end up with:
```
some:
  object:
    script: |
  #!/bin/bash
  echo "hello
```
I think we should remove the whitespace from the block values and use the prefix of the parent (in case of a LITERAL or FOLDER style scalar) to indent when printing

It might be better to add the current indentation to the printer context.


jevanlingen:
@sambsnyd I cleaned it up by going back to main and applying all the changes from another working branch (`4358-fix-MergeYaml-block-indentation-not-as-expected`). So now it can be reviewed.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -163,9 +165,18 @@ private boolean keyMatches(Yaml.Mapping m1, Yaml.Mapping m2) {
...
 
     private Mapping mergeMapping(Yaml.Mapping m1, Yaml.Mapping m2, P p, Cursor cursor) {
+        // Merge same key, different value together
         List<Yaml.Mapping.Entry> mergedEntries = map(m1.getEntries(), existingEntry -> {
-            for (Entry incomingEntry : m2.getEntries()) {
+            for (Yaml.Mapping.Entry incomingEntry : m2.getEntries()) {
                 if (keyMatches(existingEntry, incomingEntry)) {
+                    // multiline scalar cannot be formatted automatically, apply indent from existing value
+                    if (shouldAutoFormat && incomingEntry.getValue() instanceof Yaml.Scalar && hasLineBreak(((Scalar) incomingEntry.getValue()).getValue())) {
+                        String s = grabAfterFirstLineBreak(((Yaml.Scalar) existingEntry.getValue()).getValue());
```
jevanlingen, line 174: This solution is not very nice imho. It should 'just work' like the `add` stuff, by adding the `MultilineScalarAdded` marker, so the IndentsVisitor can pick it up.

At the moment the IndentsVisitor does not put the right idents in front, I'll will look into this next Monday.

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/IndentsVisitor.java
```diff
@@ -95,12 +99,24 @@ public IndentsVisitor(IndentsStyle style, @Nullable Tree stopAfter) {
                 y = y.withPrefix(indentComments(y.getPrefix(), indent));
             }
         }
+
+        if (y instanceof Yaml.Scalar && y.getMarkers().findFirst(MultilineScalarChanged.class).isPresent()) {
```
timtebeek, line 103: If there's no overlap with the above conditionals then it might be clearer to use `else if` here. Any thoughts as to that?
```suggestion
        } else if (y instanceof Yaml.Scalar && y.getMarkers().findFirst(MultilineScalarChanged.class).isPresent()) {
```

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/IndentsVisitor.java
```diff
@@ -95,12 +99,24 @@ public IndentsVisitor(IndentsStyle style, @Nullable Tree stopAfter) {
...
+            y = ((Yaml.Scalar) y).withValue(newValue);
+        }
+
         return y;
     }
 
     private boolean isUnindentedTopLevel() {
         return getCursor().getParentOrThrow().getValue() instanceof Yaml.Document ||
-               getCursor().getParentOrThrow(2).getValue() instanceof Yaml.Document;
+                getCursor().getParentOrThrow(2).getValue() instanceof Yaml.Document;
```
timtebeek, line 119: I'm seeing your IDE introduce these extra spaces where we tend to have the continuation on the exact same line. Could you double check your settings?
```suggestion
               getCursor().getParentOrThrow(2).getValue() instanceof Yaml.Document;
```

timtebeek (approved):
Great to see! Approved already with three minor suggestions/questions. Merge once those are resolved.

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/IndentsVisitor.java
```diff
@@ -29,6 +30,9 @@
 import java.util.regex.Pattern;
 
 public class IndentsVisitor<P> extends YamlIsoVisitor<P> {
+
+    private static final Pattern COMMENT_PATTERN = Pattern.compile("^(\\s*)(.*\n?)", Pattern.MULTILINE);
```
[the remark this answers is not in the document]
jevanlingen: Yeah, that's better indeed! You want to match windows saved document to be matched as well.

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/IndentsVisitor.java
jevanlingen: I agree. When I started writing this, there was a lot of removal / adding of code, until I got it right. In the end it could be an `else-if` after all.

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/IndentsVisitor.java
jevanlingen: I have the default settings:

image

If I change one thing, the "Default" label turns to blue (even after restarting the IDE itself). I did also "Restore Defaults" and it didn't change a thing. I even downloaded IntelliJ IDEA 2024.3 just now, still no difference. So I am pretty sure I use the default settings myself.

Are you sure about yours 😬?

PR #4359, opened 2024-07-26
fix: `MergeYaml` does not update generated files

pstreef:
## What's changed?
Added a passing and failing test showing the problem

## What's your motivation?
`MergeYaml` should also work on newly created/generated files.

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
@@ -20,6 +20,8 @@ import org.openrewrite.DocumentExample;
 import org.openrewrite.Issue;
 import org.openrewrite.test.RewriteTest;
 
+import java.nio.file.Path;
+
 import static org.openrewrite.yaml.Assertions.yaml;
 
 @SuppressWarnings({"KubernetesUnknownResourcesInspection", "KubernetesNonEditableResources"})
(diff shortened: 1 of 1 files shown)


pstreef:
This is working as intended. We create a plain text file which MergeYaml cannot work on.

rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
```diff
@@ -1135,4 +1137,77 @@ void mergeEmptyStructureFollowedByCopyValue() {
...
+            .recipeFromYaml(
+              //language=yaml
+              """
+                ---
+                type: specs.openrewrite.org/v1beta/recipe
+                name: org.openrewrite.CreateAndUpdate
+                displayName: Create and update
+                description: Create a file and update it.
+                recipeList:
+                  - org.openrewrite.text.CreateTextFile:
```
timtebeek, line 1189: Not sure where this example came from, but we do have an explicit Create YAML file recipe: `org.openrewrite.yaml.CreateYamlFile`
https://docs.openrewrite.org/recipes/yaml/createyamlfile

Might be worth either passing along that message, or testing with a Yaml file created.

pstreef: yes, I found that, an ran into the next issue 🙈

PR #4362, opened 2024-07-26
Add HCL FindAndReplaceLiteral Recipe

l-ferguson:
## What's changed?
Adding a new recipe to find and replace literal values in HCL files.

## What's your motivation?
This recipe provides an HCL language-specific alternative to the existing FindAndReplace recipe that parses files as plaintext. This recipe will find and replace literal values in HCL files and allow you to execute additional HCL language-specific recipes afterwards in a single recipe run.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -150,26 +150,28 @@ public class Environment {
             }
         }
         if (!recipesNotFound.isEmpty()) {
+            @SuppressWarnings("deprecation")
             List<String> suggestions = recipesNotFound.stream()
                     .map(r -> recipesByName.keySet().stream()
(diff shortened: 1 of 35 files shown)


timtebeek:
Hi & welcome, thanks for the immediate PR. I'll have a closer look over the weekend, but for now we have some automated code suggestions to conform to our coding standards. Great start already! To simplify review, would you mind sharing is you copied parts from elsewhere such that we can compare and review only the changes?

l-ferguson:
Hi! I copied the existing [FindAndReplace recipe](https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/text/FindAndReplace.java) as a starting point and then modified it to scope to HCL literals

timtebeek (approved):
Thanks again @l-ferguson ; I've added some polishing commits to start with the bare minimum of what we need. Hope you agree with the choices I made. We can always add options later, or work with preconditions to limit what files to apply to.

PR #4363, opened 2024-07-28
Add recipe for adding/replacing a Maven runtime config

bmuschko:
## What's changed?

Introduces a new recipe for adding and replacing [Maven runtime configuration](https://maven.apache.org/configure.html#mvn-directory) in `.mvn/maven.config` or `.mvn/jvm.config`.

## What's your motivation?

It's important for Maven projects to standardize on certain runtime flags and make them available through checked-in files the SCM. All developers that check out the code from SCM will  automatically apply those flags. For example, you may want to execute the Maven [build in parallel](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3) for better build performance and therefore introduce `-T=1C` in `maven.config`.

- The Maven documentation is not very specific which notation should be used for separating a flag from the assigned value if there is one (no separation, whitespace, `=` character. In fact, you will find all notations across different pieces of documentation. The implementation allows for providing this option.
- The Maven documentation indicates that new lines should be used to separate configuration for `maven.config`. For `jvm.config`, you should use a whitespace. The implementation takes this into consideration.
- Currently, the implementation does not take into consideration the alternative flag notation e.g. ` -U` vs. `--update-snapshots`. Please advise if you'd need this case to be handled with the initial PR and optimally "how".


timtebeek (approved):
Great start @bmuschko ; I've added a small polishing commit, mostly to make things smaller and more idiomatic. Feel free to go over the changes and let me know if you have any question.

No need to handle configuration aliases just yet; but we can see how that turns out in practice.

bmuschko:
@timtebeek Thanks for the polishing and merging. All makes sense and gives additional guidance for future PRs.

PR #4364, opened 2024-07-30
feat: Add GitProvenance support for Azure DevOps repositories

bryceatmoderne:
## What's changed?
Updated `GitProvenance` to support parsing Azure DevOps repositories.

## What's your motivation?
These changes are required to support Azure DevOps as an SCM.

Let me know of other `rewrite` classes that may require similar updates.

@timtebeek 

No.

## Any additional context
No.

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
@@ -54,24 +54,21 @@ class GitProvenanceTest {
 
     private static Stream<Arguments> remotes() {
         return Stream.of(
-          Arguments.of("ssh://git@github.com/openrewrite/rewrite.git", "openrewrite", null, "rewrite"),
-          Arguments.of("https://github.com/openrewrite/rewrite.git", "openrewrite", null, "rewrite"),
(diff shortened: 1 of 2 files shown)


rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
```diff
@@ -83,9 +91,9 @@ void getOrganizationNameWithBaseUrl(String gitOrigin, String baseUrl, String org
 
     @ParameterizedTest
     @MethodSource("remotes")
-    void getRepositoryName(String remote) {
+    void getRepositoryName(String remote, String org, String repo) {
```
timtebeek, line 94: Can we collapse this and `getOrganizationName` above to get rid of the unused argument warnings?

timtebeek (approved):
This has always been a bit of an oddity; especially the long deprecated but never fully removed `getOrganizationName()`. Thanks for extending it thoughtfully.

rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
bryceatmoderne: Done. Thanks for the suggestion!

bryceatmoderne:
This requirement was addressed by https://github.com/openrewrite/rewrite/pull/4381

Issue #2218, opened 2022-09-14
MergeYamlVisitor - inline comments not taken into account


thomaszub:
Hi,

the `MergeYamlVisitor` does not take inline comments into account at merging YAMLs. See the test:

```kotlin
@Test
fun existingEntryBlockWithComment() = assertChanged(
    recipe = MergeYaml(
        "$",
        """
            spring:
              application:
                description: a description
        """.trimIndent(),
        false,
        null
    ),
    before = """
        spring:
          application:
            name: main #Some comment
    """,
    after = """
        spring:
          application:
            name: main #Some comment
            description: a description
    """
)
```

The actual result is:
```yaml
spring:
  application:
    name: main
    description: a description #Some comment
```

Thanks and kind regards
Thomas

PR #4365, opened 2024-07-30
Merge yaml test for inline comments retention

johnpaulms:
## What's changed?
Added test case to test inline comment retention for mergeyaml recipe

## What's your motivation?
We have same issue as described here
- https://github.com/openrewrite/rewrite/issues/2218

Nothing as of now.

@timtebeek @pstreef to review this test case

There is no work around for this issue. Developers want to retain the comments in their yaml file.

### Checklist


timtebeek:
Thanks for the helpful test here @johnpaulms ! What I can see from stepping through with the debugger is that the comment is attached to the `document.end`, as opposed to the `document.block.entries[0]`, which it why it appears to shift. If there were an element after that sequence, then the comment moves to the prefix of the next element. We'l have to mimic that and move the end to a prefix accordingly to have that preserve it's location. @pstreef is this something you can fit in there?

pstreef (approved):

pstreef:
merging this into my branch so I can take it along there

PR #4367, opened 2024-08-01
feat: Determine SCM and parse clone url in GitProvenance for more accurate path/origin/organiation/groups

pstreef:
## What's changed?
Determine the SCM from a list of known urls and any manually registered SCMs. The SCM is then used to parse a typed `CloneUrl` which in turn give information about a repository. Depending on the SCM there are some expectations about the clone url that must be met.

## What's your motivation?
To determine the repository origin, path, organization, project and group path (for GitLab) is tricky, especially with self-hosted SCMs. Doing this in 1 place makes it more maintainable.

Any url patterns that are missed

@bryceatmoderne 

Implementing this in higher level code/places.

## Any additional context
This is going to be used directly by Moderne CLI, but should be useful for other projects as well. 

### Checklist


rewrite-core/src/test/java/org/openrewrite/scm/AzureDevOpsScmTest.java
```diff
@@ -0,0 +1,51 @@
...
+            assertThat(parsed).isInstanceOf(AzureDevopsCloneUrl.class);
+            AzureDevopsCloneUrl azureDevopsCloneUrl = (AzureDevopsCloneUrl) parsed;
+            assertThat(azureDevopsCloneUrl.getOrigin()).isEqualTo(expectedOrigin);
+            assertThat(azureDevopsCloneUrl.getPath()).isEqualTo(expectedPath);
+            assertThat(azureDevopsCloneUrl.getProject()).isEqualTo(expectedProject);
+            assertThat(azureDevopsCloneUrl.getOrganization()).isEqualTo(expectedOrganization);
+        }
+    }
+
+}
```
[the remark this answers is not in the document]
timtebeek: If you see this message it wants you to put a newline at the end of the file 🤷🏻

timtebeek:
Resolved these issues with a quick recipe run from the IntelliJ IDEA plugin through
```yaml
name: Scratch rewrite.yml
recipeList:
  - org.openrewrite.java.format.EmptyNewlineAtEndOfFile
```
The full list of checks on every pull request is maintained here:
https://github.com/openrewrite/rewrite/blob/45685773db9d4996dab9833ef963039bd997d3ae/.github/workflows/receive-pr.yml#L21-L84

I'm considering adding that to a root `rewrite.yml` to make it easier to run these recipes on your local machine. Thoughts?

pstreef:
> Resolved these issues with a quick recipe run from the IntelliJ IDEA plugin through
> 
> ```yaml
> type: specs.openrewrite.org/v1beta/recipe
> name: Scratch rewrite.yml
> recipeList:
>   - org.openrewrite.java.format.EmptyNewlineAtEndOfFile
> ```
> 
> The full list of checks on every pull request is maintained here:
> 
> https://github.com/openrewrite/rewrite/blob/45685773db9d4996dab9833ef963039bd997d3ae/.github/workflows/receive-pr.yml#L21-L84
> 
> I'm considering adding that to a root `rewrite.yml` to make it easier to run these recipes on your local machine. Thoughts?

Ah that's what it was trying to do. The suggestions were not showing that and not applyable.. Yeah sound like a good plan!

okundzich:
just talked to Bryce about azuredevops and this came up. We require configuration of baseurl in moderne CLI/agent. curious about this.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -67,14 +75,40 @@ public class GitProvenance implements Marker {
...
+                         @Nullable EOL eol,
+                         @Nullable List<Committer> committers) {
+        this.id = id;
+        this.origin = origin;
+        this.branch = branch;
+        this.change = change;
+        this.autocrlf = autocrlf;
+        this.eol = eol;
+        this.committers = committers;
+        cloneUrl = parseCloneUrl(origin);
```
jkschneider, line 96: We should probably only calculate this when it is first used, otherwise this slows down deserialization.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -67,14 +74,44 @@ public class GitProvenance implements Marker {
...
+
     /**
      * Extract the organization name, including sub-organizations for git hosting services which support such a concept,
      * from the origin URL. Needs to be supplied with the
      *
      * @param baseUrl the portion of the URL which precedes the organization
      * @return the portion of the git origin URL which corresponds to the organization the git repository is organized under
+     * @deprecated use {@link #getCloneUrl().getOrganization()} instead }
      */
+    @Deprecated
```
jkschneider, line 109: I'm not sure it's necessary to deprecate since it is in widespread use. Fine to have a convenience method for organization on `CloneUrl`

pstreef: actually I think we should point back to the "old" method that does not take an argument, and deprecate this one which (again) takes the origin (with scheme).

pstreef:
replaced by https://github.com/openrewrite/rewrite/pull/4381

PR #4372, opened 2024-08-02
Isolate JavaTemplate issue with generic type parameters

timtebeek:
## What's changed?
Replicate generic type issue affecting a downstream PR.

```diff
diff --git a/Foo.java b/Foo.java
index c811849..2466f30 100644
--- a/Foo.java
+++ b/Foo.java
@@ -1,7 +1,8 @@ 
 import batch.StepBuilder;
 class Foo {
     void test() {
-        new StepBuilder()
+        new StepBuilder()/*__TEMPLATE_STOP__*/
+                ./*__TEMPLATE_STOP__*/<String>method()
             .<String>method();
     }
 }
```

## What's your motivation?
This is holding up
- https://github.com/openrewrite/rewrite-spring/pull/284

Not yet explored the fix; figured log the issue first to document the problem in isolation. Any help welcome.

@knutwannheden or @sambsnyd

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/TreeVisitingPrinter.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/TreeVisitingPrinter.java
@@ -17,14 +17,12 @@ package org.openrewrite.java;
 
 import org.openrewrite.*;
 import org.openrewrite.internal.lang.Nullable;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JLeftPadded;
-import org.openrewrite.java.tree.JRightPadded;
-import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.*;
 
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.joining;
 import static java.util.stream.StreamSupport.stream;
 
 
@@ -174,7 +172,8 @@ public class TreeVisitingPrinter extends TreeVisitor<Tree, ExecutionContext> {
             return s != null ? s : "";
         }
 
-        String[] lines = tree.toString().split("\n");
+        String precedingComments = tree instanceof J ? printComments(((J) tree).getPrefix().getComments()) : "";
+        String[] lines = (precedingComments + tree).split("\n");
         StringBuilder output = new StringBuilder();
         for (int i = 0; i < lines.length; i++) {
             output.append(lines[i].trim());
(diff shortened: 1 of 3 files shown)


timtebeek:
For context: the generated template in this case is
```java
import batch.StepBuilder;
import org.openrewrite.java.internal.template.__M__;
import org.openrewrite.java.internal.template.__P__;
class Foo{void test(){
/*__TEMPLATE__*/new StepBuilder()/*__TEMPLATE_STOP__*/
./*__TEMPLATE_STOP__*/<String>method();}}
```

rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -803,6 +806,13 @@ public J visitMethodInvocation(J.MethodInvocation method, Integer integer) {
                     //noinspection ConstantConditions
                     return mi.getSelect();
                 }
+                if (method.getTypeParameters() != null) {
+                    // For method chains return `select` if `STOP_COMMENT` is found before `typeParameters`
+                    if (stopCommentExists(mi.getPadding().getTypeParameters().getBefore().getComments())) {
+                        //noinspection ConstantConditions
+                        return mi.getSelect();
+                    }
+                }
```
timtebeek, line 815: This is what fixed the unit test added here, for the minimal possible impact change we could do here.

A larger change could reduce remove the `__TEMPLATE_STOP__` added twice, but it's unclear what unintended side effects that might have downstream, so I went for this safer option.

rewrite-java/src/main/java/org/openrewrite/java/internal/template/BlockStatementTemplateGenerator.java
```diff
@@ -193,14 +202,6 @@ private boolean isTemplateStopComment(Comment comment) {
         return js;
     }
 
-    private void template(Cursor cursor, J prior, StringBuilder before, StringBuilder after, J insertionPoint, JavaCoordinates.Mode mode) {
-        if (contextSensitive) {
-            contextTemplate(cursor, prior, before, after, insertionPoint, mode);
-        } else {
-            contextFreeTemplate(cursor, prior, before, after);
-        }
-    }
```
timtebeek, line 202: Inlined above for simplicity, as this was only used once and gave a confusion indirection with unused arguments.

rewrite-java/src/main/java/org/openrewrite/java/TreeVisitingPrinter.java
```diff
@@ -174,7 +172,8 @@ private static String printTreeElement(Tree tree) {
             return s != null ? s : "";
         }
 
-        String[] lines = tree.toString().split("\n");
+        String precedingComments = tree instanceof J ? printComments(((J) tree).getPrefix().getComments()) : "";
```
timtebeek, line 175: Discovered while debugging the below; figured helpful for folks to see preceding comments too.

timtebeek:
Merging already given how this is guarded to only affect cases that would have previously failed.

PR #4376, opened 2024-08-02
Expand GradleDependency Trait to change dependencies in jvm-test-suite DSL & buildscript

sihutch:
## What's changed?
Dependencies [configured within a test suite](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#configure_dependencies_of_a_test_suite) are not updated by the org.openrewrite.gradle.ChangeDependency recipe. 

This PR has only a failing test case. I would be happy to try to contribute a fix but I would need a lot of pointers to get me started. 

## What's your motivation?
We are heavy users of the Test Suites and would like to be able to upgrade their dependencies using OpenRewrite.


rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -15,8 +15,10 @@
  */
 package org.openrewrite.gradle;
 
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
+import org.openrewrite.gradle.util.GradleWrapper;
```
timtebeek, line 21:
```suggestion
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -15,8 +15,10 @@
  */
 package org.openrewrite.gradle;
 
+import org.junit.jupiter.api.DisplayName;
```
timtebeek, line 18: I think these might be left over from an earlier change perhaps?
```suggestion
```

timtebeek:
Great to see this replicated, thanks! Did you already step through using this test as well?

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
sihutch: Sorry, fixed

sihutch:
No problem. I haven't stepped through yet. I am about to go on vacation for a week so will hopefully pick it up on my return unless someone else does.

timtebeek:
Ah thanks a lot for logging the issue before logging off; enjoy your time off! We'll see if some of the usual suspects gets to it before you return. ;)

shanman190:
@sihutch and @ryanwalker, I just noticed that both of you are working on the same update to the same recipe. Would one of you like to perform the updates on ChangeDependency while the other does the same for UpgradeDependency? This way you'd both get the experience and it'd help out 2 of the many Gradle recipes needing this particular update.

This is Ryan's PR that he's working on presently: https://github.com/openrewrite/rewrite/pull/4354

sihutch:
@shanman190 @ryanwalker - I am happy to wait for Ryan to complete his PR on ChangeDependency and then work for UpgradeDependency. Once we have a working solution, it should be straightforward to extract any commonality.

sihutch:
@shanman190 Here's a first pass at a solution

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -68,6 +68,10 @@ public Matcher artifactId(String artifactId) {
                     return null;
                 }
 
+                if (!withinDependenciesBlock(cursor)) {
+                    return null;
+                }
+
                 GradleDependencyConfiguration configuration = gradleProject.getConfiguration(methodInvocation.getSimpleName());
                 if (configuration == null) {
```
shanman190, line 76: Let's here instead also allow `"classpath".equals(methodInvocation.getSimpleName())`. By doing this we can completely eliminate the `GradleDependencyPredicate` code altogether and only use the `GradleDependency.Matcher`.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyClassifierTest.java
```diff
@@ -411,4 +412,84 @@ void noNewClassifier_2(String group, String artifact) {
           )
         );
     }
+
+    @Test
+    void worksWithDependencyDefinedInJvmTestSuite() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeDependencyClassifier("org.openrewrite", "*", "classified", ""))
+            .beforeRecipe(withToolingApi()),
```
shanman190, line 420: The `withToolingApi()` will need to be put in the defaults `beforeRecipe` and will no longer be needed in each method case across all of the unit tests.

Presently, the other test cases are all passing because the code is falling back to the `MethodMatcher`. With my other comments, this won't be the case anymore and we will need the `GradleProject` marker for the all of these other tests for the recipes being updated in this PR.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
shanman190: This is a good question. I decided to go ahead and try to correctly capture the `classpath` configuration via https://github.com/openrewrite/rewrite/pull/4459. If that PR gets merged, then we'll be able to check for the configuration within the buildscript's own configurations map and then get the `ResolvedDependency` just the same way as with a project dependency.

sihutch: Perfect. I will watch for your PR

shanman190: PRs are merged, so now you should be able to update the `GradleDependency` trait to look at the new `GradleBuildscript` in order to find the `classpath` configuration and from that the `ResolvedDependency` instance that will be available in that configuration.

sihutch:
@shanman190 @timtebeek 

I have removed `DependencyMatchPredicate` now so only the trait is being used. I have updated all of the tests to work with the `toolingApi` - I have not been able to get two tests to pass so maybe you can either fix these or give me some pointers. 

The failing tests are

1. `DependencyUseStringNotationTest#basicMapLiteral`
2. `ChangeDependencyConfigurationTest#worksForProjectDependencies`

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependencyArtifactId.java
```diff
@@ -112,7 +113,10 @@ public G visitCompilationUnit(G.CompilationUnit compilationUnit, ExecutionContex
             @Override
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (!dependencyDsl.matches(m) || !(StringUtils.isBlank(configuration) || m.getSimpleName().equals(configuration))) {
+
+                GradleDependency.Matcher gradleDependencyMatcher = new GradleDependency.Matcher();
+                
+                if (!gradleDependencyMatcher.get(getCursor()).isPresent() || !(StringUtils.isBlank(configuration) || m.getSimpleName().equals(configuration))) {
```
shanman190, line 119: An additional enhancement here would be to expose `configuration` as another selector for the matcher. This isn't strictly necessary as part of this PR though, but I wanted to mention it while I was here.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -119,6 +123,30 @@ public Matcher artifactId(String artifactId) {
             return null;
         }
 
+        private static @org.jetbrains.annotations.Nullable GradleDependencyConfiguration getConfiguration(GradleProject gradleProject, J.MethodInvocation methodInvocation) {
```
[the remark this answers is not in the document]
shanman190: @timtebeek, should this particular update get migrated over to the JSpecify implementation?

shanman190:
@sihutch,

For tests 1-3, I'd expect for these to work. I'll try to debug these a little bit locally.
For tests 4 & 5, these feel like they are no longer valid, so I'd be all for removing them.

shanman190:
@timtebeek, do you have any ideas or opinions on how one might address test cases 2 or 3 from my analysis above?

For the project dependency, we could build enough of a `ResolvedDependency` to make it work or what Gradle does internally is they have another type to represent a project dependency that is separate from the external dependency.

The subprojects block dependencies is a harder problem as we've gotten by so far by abusing the current project, but that's not really sustainable. Gradle's configuration cache/project isolation feature will help as teams adopt it as it explicitly disallows this from happening on modern projects.

sihutch:
@shanman190 

> 1. `DependencyUseStringNotationTest#basicMapLiteral`: This is failing because the `GradleDependency.Matcher` doesn't handle the `G.MapLiteral` type. This should be an easy problem to resolve as it's mostly just delegating to the `G.MapEntry` path.

I've pushed a fix for this.

I have also removed tests 4 and 5

I have resolved merge conflicts, but this has resulted in a new failing test case

`UpgradeDependencyVersionTest#dependenciesBlockInFreestandingScript`

shanman190:
Yeah, I was just looking at that new test last night. For the standalone scripts, they aren't carrying a `GradleProject` marker at all at the moment. @sambsnyd was going to see about collecting project markers during the scan phase, but that is still a work in progress.

shanman190:
I'm wondering if we won't have to make a concession for these 3 failing tests by just retaining the old path as a fallback. This would ultimately only effect the `UpgradeDependencyVersion` and `ChangeDependencyConfiguration` recipes.

sihutch:
I would definitely do this. It seems a shame to hold back the otherwise useful functionality.

shanman190:
@sihutch, rather than the predicate, I think I'd probably update all of the references to first check with the `GradleDependency.Matcher`, then if that fails to use the `MethodMatcher("DependencyHandlerSpec *(..)")`. Calling this out specifically, but the `UpgradeDependencyVersuon#isLikelyDependencyConfiguration(..)` could likely be removed entirely as well. The `GradleDependency.Matcher` and `MethodMatcher` should provide stronger guarantees across the board.

As it relates to the predicate, I don't think the increase in API surface area is worth the indirection at the moment given the above.

sihutch:
@shanman190 I have implemented the fallback for the `UpgradeDependencyVersion` and `ChangeDependencyConfiguration recipes`/ All tests are passing now.

shanman190:
Awesome! I'll probably do a small cleanup commit tomorrow, but then we'll get this merged.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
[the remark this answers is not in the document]
timtebeek: Yes thanks! Updated the enforcement in https://github.com/openrewrite/rewrite/commit/83150ba5bef389f9cd28b4edf4b914351eb090f4

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyClassifierTest.java
timtebeek: Do we expect any issues with this for tests downstream looking at rewrite-java-dependencies and beyond? Just wondering what work might lie ahead when we merge this; it's no problem, but good to be aware.

timtebeek (approved):
Thanks a lot for the detailed work here! I've adapted the PR title before that becomes the squashed commit message; I hope I covered the essential well there.

Sure nice to see complexity concentrated into that GradleDependency Trait, and reused from each of the individual recipes.

Thanks as well to @shanman190 for the continued guidance on this PR; I'll let you merge since you mentioned you were planning a slight polish as well.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyClassifierTest.java
shanman190: So that's one benefit of retaining the `MethodMatcher` as a fallback throughout these recipes (partly what I intend to polish tomorrow for the others). In this way, we will use the `GradleDependency` trait when we have the `GradleProject` marker -- which will guarantee the configuration is registered with Gradle, including previously unknown ones -- and will fallback to our internalized list of configurations found in `DependencyHandlerSpec` when such a marker cannot be found.

sihutch:
Thanks @timtebeek. It was a good learning experience, and I appreciate the detailed guidance from @shanman190. 

I am a platform team member managing change for hundreds of projects. This is my first time using OpenRewrite for migration, and it is proving very useful. Thanks to the team. I hope to make more contributions in the future.

shanman190:
@sihutch, thanks again for all of the work put in on this!

sihutch:
@shanman190, we are heavy users of test suites so this improvement is great for us. Thanks for your patience and detailed responses. I really appreciate it and have learned a lot through the process.

PR #3517, opened 2023-08-30
RewriteTest / YamlResourceLoader Enhancements

nmck257:
## What's changed?
- YamlResourceLoader now uses jackson to try to instantiate imperative recipes with zero args if basic reflection fails. And, reverted change to RecipeSpec which made it scanRuntimeClasspath unnecessarily for yaml recipes. Fixes #3414
- adding RewriteTest/RecipeSpec recipeFromResources option to scan all rewrite yamls on classpath (ignoring classes. Fixes #3096

## Anything in particular you'd like reviewers to focus on?
- In `ClasspathScanningLoader`, the constructor overloads were starting to get diverse in functionality, so I opted for a factory method for this new scenario. Open to changing it
- In `ClasspathScanningLoader`, there's this comment, which I'm not sure is accurate today; I think the new case `RecipeLifecycleTest::declarativeRecipeChainFromResourcesIncludesImperativeRecipesInDescriptors` disproves it https://github.com/openrewrite/rewrite/blob/1834d4e3fd3253a6f604f7b34cfc14d02ee66662/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java#L102-L106
- I considered removing the reflective attempt to instantiate the recipe and only using jackson, but that led to a regression with recipes like this, because jackson (for some reason) picked the 1-arg constructor (passing a null value) instead of the annotated zero-arg constructor (which would have set the required field by default) https://github.com/openrewrite/rewrite-migrate-java/blame/main/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java

## Anyone you would like to review specifically?
@sambsnyd 

## Have you considered any alternatives or workarounds?
- Scanning the full runtime classpath when you're just looking for Yaml files is slow and doesn't seem beneficial
- You can also explicitly list the yamls you care about when defining your `RecipeSpec`, but with recursive validation, if you're testing a recipe which chains to many recipes across many files, then that becomes painful to maintain

### Checklist
- [X] I've added unit tests to cover both positive and negative cases
- [X] I've added the license header to any new files through `./gradlew licenseFormat`
- [X] I've used the IntelliJ IDEA auto-formatter on affected files
- [ ] I've updated the documentation (if applicable)

PR #4377, opened 2024-08-02
Added test property for FAIL_ON_UNKNOWN_PROPERTIES configuration

ajturnerora:
## What's changed?
I have changed the `new Properties()` used in test `RecipeSpec` to be a class field. This allows customization of logic deeper in rewrite-core. In this case I have changed the `JsonMapper` in `YamlResourceLoader` to allow `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to be customized. The default is that `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` is disabled as that was the existing behavior.

Users can now make their tests fail if there are unknown properties present in the recipe using
`spec.failOnUnknownProperties(true)`

I also had to update the `RECIPE_CACHE` key as the properties set can change the recipe outcome - although typically the new part of the key will be `{}` as no existing code could be configuring any properties.

## What's your motivation?
This allows bugs for incorrectly specified recipe options to be more easily caught. E.g., the in below example both are valid recipes and will run but the second one is the true intent and it can be difficult to catch during code reviews:

Incorrect:
```
  - org.openrewrite.maven.RemoveRedundantDependencyVersions:
      groupId: org.apache.commons
```

Correct:
```
  - org.openrewrite.maven.RemoveRedundantDependencyVersions:
      groupPattern: org.apache.commons
```

If there is a better/alternate approach - although it would seem that using the `Properties` object is the correct way to configure object.

- As discussed here: https://github.com/openrewrite/rewrite/discussions/4375
@timtebeek 

## Any additional context
I made this change as we hit the exact issue describe above and it wasn't caught during PR review. If we had the option to fail tests when unnecessary arguments are specified we would likely enable it for every test case.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
@@ -146,8 +146,12 @@ public class YamlResourceLoader implements ResourceLoader {
                 .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
                 .constructorDetector(ConstructorDetector.USE_PROPERTIES_BASED)
                 .build()
-                .registerModule(new ParameterNamesModule())
-                .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+                .registerModule(new ParameterNamesModule());
+
+        if (Boolean.FALSE.equals(properties.getOrDefault(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false))) {
+            mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+        }
+
         maybeAddKotlinModule(mapper);
 
         this.classLoader = classLoader;
--- a/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
@@ -131,6 +131,7 @@ public class YamlResourceLoader implements ResourceLoader {
      * @param source      Declarative recipe source
      * @param properties  Placeholder properties
      * @param classLoader Optional classloader to use with jackson. If not specified, the runtime classloader will be used.
+     * @param dependencyResourceLoaders Optional resource loaders for recipes from dependencies
      * @throws UncheckedIOException On unexpected IOException
      */
     public YamlResourceLoader(InputStream yamlInput,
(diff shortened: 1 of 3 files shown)


timtebeek:
Nice and measured approach here! I'd maybe even go as far as enabling this by default when running tests in `rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java` , to flag issues early through or testing framework. We have similar checks that are more strict when running tests only, such as our type validation checks. Would be interesting to see how widespread that is in downstream modules. I'd hope not of course, but certainly interesting.

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -146,8 +146,12 @@ public YamlResourceLoader(InputStream yamlInput,
                 .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
                 .constructorDetector(ConstructorDetector.USE_PROPERTIES_BASED)
                 .build()
-                .registerModule(new ParameterNamesModule())
-                .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+                .registerModule(new ParameterNamesModule());
+
+        if (Boolean.FALSE.equals(properties.getOrDefault(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false))) {
```
timtebeek, line 151: I'm not quite sure about the reuse of `Properties properties` for configuration; we at present use that to optionally replace properties in Yaml recipe files with looked up values.
https://github.com/openrewrite/rewrite/blob/83501012e8abaab57b9c006205dea7014db943c3/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java#L171-L172

I feel the dual use here might cause confusion in the future, but I'm also not yet sure about a better alternative. 

The prevailing style in `YamlResourceLoader` seems to be new overloaded constructors. Perhaps another one that takes a `Consumer<JsonMapper>` could offer the same functionality in a different way?

timtebeek: Or we juse remove the line that says `disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)` and see what catches fire. 🤷🏻

timtebeek: One of the tests that fails when I try that quickly was added here
- https://github.com/openrewrite/rewrite/pull/3517

timtebeek: Shannon rightfully pointed out that we might want to only and always enable `FAIL_ON_UNKNOWN_PROPERTIES` in tests, but keep that disabled when loading and running recipes in practice. Otherwise we get some unfortunate error handling we have to do in odd places. Just checking this in tests strikes the right balance we think, and ought to incentivize tests even more, similar to the type checks we do there.

shanman190:
Tim and I chatted a little bit and we're going to take what's here as a half step. This'll specifically cover directly loaded YAML, but misses YAML loaded from dism via `recipesFromResources`. That'll get fixed in a followup PR.

shanman190 (approved):

ajturnerora:
Thanks for getting to this so quickly. I feel bad that I suggest and enhancement and you did all the work 😂 

Were you thinking of taking a similar approach for the other loading path? (`Consumer<ObjectMapper>` which I think is a good solution btw.). The only code paths I see with `new YamlResourceLoader` are `Environment.scanX` -> `ClasspathScanningLoader` -> `YamlResourceLoader` which is a short call chain to modify. I could attempt the follow up PR if you're okay with the approach of overloading the `Environment.Builder` constructor with
```
public Builder(Properties properties, Consumer<ObjectMapper>)
```

There aren't any other `ResourceLoader`s other than `ClasspathScanningLoader` and `YamlResourceLoader` so that should cover all use cases.

timtebeek:
I'd briefly explored that option, but didn't arrive at a solution I'd liked just yet. And since this is mostly/only used in tests for now I'd abandoned it for now. You're welcome to try your hand at a non-intrusive change if you can find one. 😅

PR #4378, opened 2024-08-02
refactor: Azure devops refactoring

bryceatmoderne:
- built org name from organization and project a la gitlab
- use `Pattern`s to parse Azure DevOps urls
- additional tests
- additional changes to run with `moderne-cli`

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scm/AzureDevOpsCloneUrl.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scm/AzureDevOpsCloneUrl.java
@@ -25,15 +25,16 @@ public class AzureDevOpsCloneUrl implements CloneUrl {
     String organization;
     String repositoryName;
 
-    public AzureDevOpsCloneUrl(String cloneUrl,
-                               String origin,
-                               String organization,
-                               String project,
-                               String repositoryName) {
+    public AzureDevOpsCloneUrl(String cloneUrl, String origin, String path) {
         this.cloneUrl = cloneUrl;
         this.origin = origin;
-        this.organization = organization + '/' + project;
-        this.repositoryName = repositoryName;
-        this.path = this.organization + '/' + repositoryName;
+        this.path = path;
+        if (!this.path.contains("/")) {
+            throw new IllegalArgumentException("Azure DevOps clone url path must contain organization, project and repository");
+        }
+        String azureDevOpsOrganization = path.substring(0, path.indexOf("/"));
+        String azureDevOpsProject = path.substring(path.indexOf("/") + 1, path.lastIndexOf("/"));
+        this.repositoryName = path.substring(path.lastIndexOf("/") + 1);
+        this.organization = azureDevOpsOrganization + '/' + azureDevOpsProject;
     }
 }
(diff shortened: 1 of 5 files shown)


rewrite-core/src/main/java/org/openrewrite/scm/AzureDevOpsCloneUrl.java
```diff
@@ -18,23 +18,22 @@
...
-    public AzureDevopsCloneUrl(String cloneUrl, String origin, String path) {
+    public AzureDevOpsCloneUrl(String cloneUrl,
+                               String origin,
+                               String organization,
+                               String project,
+                               String repositoryName) {
         this.cloneUrl = cloneUrl;
         this.origin = origin;
-        this.path = path;
-        if (!path.contains("/")) {
```
pstreef, line 33: I think we should still do this check. Otherwise you get IndexOutOfBounds on a bad url

rewrite-core/src/main/java/org/openrewrite/scm/AzureDevOpsScm.java
```diff
@@ -15,7 +15,16 @@
  */
 package org.openrewrite.scm;
 
+import org.openrewrite.internal.lang.Nullable;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class AzureDevOpsScm implements Scm {
+    private static final Pattern AZURE_DEVOPS_HTTP_REMOTE =
```
pstreef, line 24: I dont think these patterns cover every possible url.
for example, this url is technically also valid: 
`ssh://ssh.dev.azure.com:22/v3/org/project/repo`
 (and such test cases are there for `UrlComponents`)
 
 I like the other changed, but using matchers here, and not for other SCM's means we have 2 ways of doing things and we should avoid that. Since we cannot use matchers for all Scm's I think we should stick with parsing.

rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
```diff
@@ -53,30 +55,47 @@
...
         );
     }
 
     @SuppressWarnings("deprecation")
     @ParameterizedTest
     @MethodSource("remotes")
-    void getOrganizationName(String remote) {
-        assertThat(new GitProvenance(randomId(), remote, "main", "123", null, null, emptyList()).getOrganizationName())
-          .isEqualTo("openrewrite");
+    void getRepositoryPath(String origin, String expectedOrg, String expectedRepo) {
```
pstreef, line 82: Unless it is reused we should make this a `@CsvSource`

rewrite-core/src/test/java/org/openrewrite/scm/AzureDevOpsScmTest.java
```diff
@@ -17,35 +17,41 @@
...
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {
+      "https://github.com/org/repo",
+      "https://gitlab.com/org/repo",
+      "https://scm.company.com/scm/project/repo.git",
+      "git@scm.company.com:context/path/scm/project/repo.git"
+    })
+    void splitOriginPathWithInvalidUrls(String cloneUrl) {
```
pstreef, line 54: Splitting these off makes sense, but this test could use a different name

pstreef (commented):
While most changes make sense, I would like to get rid of the matchers in AD handling for consistency.

rewrite-core/src/test/java/org/openrewrite/marker/GitProvenanceTest.java
pstreef: Looks like it is reused

PR #4379, opened 2024-08-03
Only add a Gradle wrapper if there is a Gradle project

shanman190:
## What's changed?
Modify `UpdateGradleWrapper` to only generate a wrapper when it has deduced that there also exists a Gradle project within the source set.

## What's your motivation?
When running the recipe with either the default or an explicit `true` value for the `addIfMissing` parameter will result in a new Gradle wrapper being generated at the root of the project, even if one doesn't make sense for the repository at hand.

Are there use cases where a Gradle wrapper may be desired and a Gradle project doesn't exist? 🤔

Anyone

* Use a precondition to gate the generation within upstream recipes
* Be more selective when using the default or explicit `true` value for the `addIfMissing` parameter

## Any additional context
N/A

### Checklist


sjungling:
Thank you for this improvement!

timtebeek (approved):
Great to see, thanks! Now we no longer have to remember not to add the wrapper to [rewrite-maven-plugin](https://github.com/openrewrite/rewrite-maven-plugin) 😁

PR #4381, opened 2024-08-04
feat: Add `GitRemote` detection logic

pstreef:
Add logic to determine the `GitRemote` from a list of registered (default and custom) origin/service mappings.

This can be used to determine the correct repository path, name and organization.


rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -0,0 +1,178 @@
...
+                    if (hostAndPath.concat().startsWith(maybeOrigin)) {
+                        service = origins.get(maybeOrigin);
+                        origin = maybeOrigin;
+                        break;
+                    }
+                }
+            }
+
+            if (service == null) {
+                // If we cannot find a service, we assume the last 2 path segments are the organization and repository name
```
bryceatmoderne, line 94: I don't think this will work with on-prem gitlab repos. You can name the host anything and have an arbitrary number of nested subgroups.

pstreef: correct, so for an on-prem gitlab we do require to register the base url

bryceatmoderne (approved):

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -0,0 +1,178 @@
...
+            switch (service) {
+                case AzureDevOps:
+                    if (origin.equals("ssh.dev.azure.com")) {
+                        origin = "dev.azure.com";
+                        repositoryPath = repositoryPath.replaceFirst("v3/", "");
+                    } else {
+                        repositoryPath = repositoryPath.replaceFirst("/_git/", "/");
+                    }
+
+                case Bitbucket:
```
jkschneider, line 116: missing a break statement

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -0,0 +1,178 @@
...
+                    origin = null;
+                } else {
+                    origin = Arrays.stream(segments, 0, segments.length - 2).collect(Collectors.joining("/"));
+                }
+            }
+
+            String repositoryPath = hostAndPath.repositoryPath(origin);
+
+            switch (service) {
+                case AzureDevOps:
```
jkschneider, line 108: missing a break statement. is that intentional?

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -67,13 +75,27 @@ public class GitProvenance implements Marker {
...
+
     /**
      * Extract the organization name, including sub-organizations for git hosting services which support such a concept,
      * from the origin URL. Needs to be supplied with the
      *
      * @param baseUrl the portion of the URL which precedes the organization
      * @return the portion of the git origin URL which corresponds to the organization the git repository is organized under
+     * @deprecated use {@link #getOrganizationName()} instead }
      */
+    @Deprecated
```
jkschneider, line 98: Don't deprecate please

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -85,54 +107,52 @@ public class GitProvenance implements Marker {
...
+    public @Nullable String getRepositoryName() {
+        if (getGitRemote() == null) {
             return null;
         }
+        return getGitRemote().getRepositoryName();
     }
 
-    public @Nullable String getRepositoryName() {
-        if (origin == null) {
+    public @Nullable String getRepositoryPath() {
```
jkschneider, line 130: I'd prefer we didn't change this name. I think you are probably trying to be more descriptive for the GitLab scenario, but I don't think that rises to the level of making a breaking change here.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
```diff
@@ -85,54 +107,52 @@ public class GitProvenance implements Marker {
...
+    }
+
+    public @Nullable String getRepositoryOrigin() {
+        if (getGitRemote() == null) {
             return null;
         }
+        return getGitRemote().getOrigin();
+    }
+
+    public static String getRepositoryPath(String origin) {
```
jkschneider, line 144: These two static methods feel low-use enough that I would remove.

rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
pstreef: The name is not changed, this is a new method (used to retrieve the path, or organization + "/" + reponame)

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: no, I missed it. I initially returned in the switch..

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -0,0 +1,178 @@
...
+                    scheme = uri.getScheme();
+                    host = uri.getHost();
+                    if (host == null && !"file".equals(scheme)) {
+                        throw new IllegalStateException("No host in url: " + url);
+                    }
+                    path = uri.getPath().replaceFirst("/$", "")
+                            .replaceFirst(".git$", "")
+                            .replaceFirst("^/", "");
+                } catch (URISyntaxException e) {
+                    throw new IllegalStateException("Unable to parse origin", e);
```
bryceatmoderne, line 150: Can you include `url` in the exception message?

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -0,0 +1,177 @@
...
+            if (service == Service.Unknown) {
+                // Do not override a known with an unknown service
+                origins.putIfAbsent(origin, service);
+            } else {
+                origins.put(origin, service);
+            }
+            return this;
+        }
+
+        public GitRemote parse(String url) {
```
bryceatmoderne, line 77: Should this be a class method?

bryceatmoderne: Nevermind... I thought it was an instance method of `GtiRemote`

rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
```diff
@@ -0,0 +1,93 @@
...
+      git@gitlab.com:group/subgroup/subergroup/subestgroup/repo.git, gitlab.com, group/subgroup/subergroup/subestgroup/repo, group/subgroup/subergroup/subestgroup, repo
+      ssh://git@gitlab.com:22/group/subgroup/subergroup/subestgroup/repo.git, gitlab.com, group/subgroup/subergroup/subestgroup/repo, group/subgroup/subergroup/subestgroup, repo
+                
+      https://bitbucket.org/PRJ/repo, bitbucket.org, PRJ/repo, PRJ, repo
+      git@bitbucket.org:PRJ/repo.git, bitbucket.org, PRJ/repo, PRJ, repo
+      ssh://bitbucket.org/PRJ/repo.git, bitbucket.org, PRJ/repo, PRJ, repo
+                
+      https://dev.azure.com/org/project/_git/repo, dev.azure.com, org/project/repo, org/project, repo
+      git@ssh.dev.azure.com:v3/org/project/repo, dev.azure.com, org/project/repo, org/project, repo
+      ssh://ssh.dev.azure.com:22/v3/org/project/repo, dev.azure.com, org/project/repo, org/project, repo
```
bryceatmoderne, line 42: The following URLs fail to parse:
```
ssh://ssh.dev.azure.com:v3/org/project/repo
ssh://git@ssh.dev.azure.com:v3/org/project/repo
```

pstreef: Correct, do you think they should? To me it looks like a combination of scp style (: between host and path)and normal url (with scheme)

bryceatmoderne: I dunno. If you use the AD UI to copy a repo ssh URL, you'll get `git@ssh.dev.azure.com:v3/org/project/repo` but that's not a valid ssh URL. ssh URLs don't include paths but we're parsing URLs with paths. 

Something like `ssh://ssh.dev.azure.com:22/v3/org/project/repo` or ``ssh://git@ssh.dev.azure.com:22/v3/org/project/repo` wouldn't be returned by AD.

Personally, I think we only need to support the following forms:
```
https://dev.azure.com/org/project/_git/repo
https://user@dev.azure.com/org/project/_git/repo
git@ssh.dev.azure.com:v3/org/project/repo
```

pstreef: but we're not only accepting things that AD is returning. We accept valid cloneUrls, including the SCP style.

If a user builds their own url, it should at least be something valid (so either `ssh://user@host/path` or `user@host:path` but not `ssh://user@host:path`

bryceatmoderne: Sorry, I'm just commenting in the context of AD URLs...

I mentioned the following URLs failed to parse:
```
ssh://ssh.dev.azure.com:v3/org/project/repo
ssh://git@ssh.dev.azure.com:v3/org/project/repo
```

But after you asked if they _should_ be parseable, I would say no since these URLs would never be returned by AD. I'm not suggesting we change what we accept for other SCMs.

pstreef: ahh ok, yes, I misunderstood then.
So no change required, right?

bryceatmoderne: The only change I might suggest is removing `ssh://ssh.dev.azure.com:22/v3/org/project/repo` as a test input since it imples we officially support it. You can make that call...

PR #4387, opened 2024-08-05
Retain dependencyManagement entries with exclusions

timtebeek:
## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/discussions/4386

I had doubts about including scope or classifier here as well; any thoughts there?

@crankydillo

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
@@ -1553,7 +1553,6 @@ class RemoveRedundantDependencyVersionsTest implements RewriteTest {
                             <dependency>
                                 <groupId>org.springframework.boot</groupId>
                                 <artifactId>spring-boot-starter-web</artifactId>
-                                <!-- The version number should be dropped -->
                                 <version>2.2.13.RELEASE</version>
                                 <!-- This exclusion should be kept -->
                                 <exclusions>
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
@@ -501,7 +501,7 @@ class RemoveRedundantDependencyVersionsTest implements RewriteTest {
                             <artifactId>myApiApp</artifactId>
                             <version>1.0.0</version>
                         </parent>
-                        
+
                         <dependencies>
                             <dependency>
                                 <groupId>com.google.guava</groupId>
@@ -550,7 +550,7 @@ class RemoveRedundantDependencyVersionsTest implements RewriteTest {
                             <artifactId>myApiApp</artifactId>
                             <version>1.0.0</version>
                         </parent>
-                        
+
                         <dependencies>
                             <dependency>
                                 <groupId>com.google.guava</groupId>
@@ -1063,7 +1063,7 @@ class RemoveRedundantDependencyVersionsTest implements RewriteTest {
                       </dependency>
                   </dependencies>
               </project>
-                """
+              """
           )
         );
     }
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -1530,4 +1530,50 @@ void keepVersionIfManagedIsNonExact() {
...
+                    <groupId>com.example</groupId>
+                    <artifactId>acme</artifactId>
+                    <version>0.0.1-SNAPSHOT</version>
+                    <dependencyManagement>
+                        <dependencies>
+                            <dependency>
+                                <groupId>org.springframework.boot</groupId>
+                                <artifactId>spring-boot-starter-web</artifactId>
+                                <!-- The version number should be dropped -->
+                                <version>2.2.13.RELEASE</version>
```
crankydillo, line 1557: I was thinking remove the version, but leave exclusions, scope, etc.  My colleagues agree, but I wouldn't say we've given it a ton of thought.  There's no clear-cut right answer in my opinion.  I can see an argument for making this configurable.  But leaving the managed version is likely more problematic than removing it altogether (current behavior).

crankydillo: I'd remove the managed dep if completely redundant, and it met the conditions (GTE, LTE, etc).  I guess you could test for equality, but I'd probably just remove if met the precondition and only had GAV coordinates.

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -1530,4 +1530,50 @@ void keepVersionIfManagedIsNonExact() {
...
+                    </parent>
+                    <groupId>com.example</groupId>
+                    <artifactId>acme</artifactId>
+                    <version>0.0.1-SNAPSHOT</version>
+                    <dependencyManagement>
+                        <dependencies>
+                            <dependency>
+                                <groupId>org.springframework.boot</groupId>
+                                <artifactId>spring-boot-starter-web</artifactId>
+                                <!-- The version number should be dropped -->
```
timtebeek, line 1556: This line was a silly left over; turns out we can't (for as much as I know) leave out a version on a managed dependency. At least our internals complain. How's that for you @crankydillo ?
```suggestion
```

crankydillo: You are right!

`mvn validate` failed complaining about missing version.  I guess I'm misreading the XSD or something.. 

Hmm, so what to do.  Functionality wise, my guess is we'd prefer the dependency to have a version that matches any newly changed upstream dependencyMgmt section.

timtebeek: Then I'd for now lean towards doing the safe thing, which is what I've implemented here, as to not make any change at all just yet. That's already an improvement over removing a known wrong case. I propose we merge this as-is and iterate in a follow up PR on a better approach if we can find one that's feasible. Would you agree?

crankydillo: My initial reaction was I'd prefer to leave it (opt 3), but I am starting to change my mind:). I'd be OK merging.

I understand this conversation needs to happen under new-approach PR, but just to set the stage..  I can think of 3 states:

1. Remove (current behavior)
2. Leave dependency with version set to old version (your current proposal).
3. Leave but set version to new managed dependency version (what I proposed above)

I don't think is 'the answer'.  I lean towards configuration.

timtebeek: I do agree that option 3 would be best; but perhaps also best tackled separately as it needs to be worked in. Option 2 is at least better than option 1, so then let's merge and iterate. :) Thanks again for the provided tests & helping think through the options!

crankydillo: I am back to say that we really need option 3;)   Leaving the old version is causing me a lot of pain.  I'll create a bug for this.

timtebeek: Thanks! I'll keep an eye out. :)

PR #4389, opened 2024-08-06
fix: Add test to reproduce `UpgradeDependencyVersion` picks wrong property

pstreef:
## What's changed?
Adding a test to try reproduce issue with `UpgradeDependencyVersion` picking the wrong property:

> We were able to reproduce the issue in sample app now:
> 
> Parent Pom
>     Dependency Management
>         artifact A
>         artifact B (Version is a Property)
>  
> Child Pom
>     Dependency
>         artifact B without version as its managed by Parent POM
> artifact A also has a BOM which has artifact B in Dependency Management where version is also a property
> 
> While Running UpgradeDependency on artifact B, the version name is picked up from Artifact A BOM, instead of nearest dependency

## What's your motivation?
Trying to figure out what exactly the issue is I'm trying to build a test for it

### Checklist


pstreef:
closed as already solved

PR #4392, opened 2024-08-07
fix: CommentOutProperty to work with single keys

jonesbusy:
Just demonstrate it doesn't works as expected (or I didn't understood the recipe).

Instead of commenting out the property it's removed.

issue

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/CommentOutProperty.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/CommentOutProperty.java
@@ -63,11 +63,16 @@ public class CommentOutProperty extends Recipe {
             private String indentation = "";
 
             @Override
-            public Yaml.Document visitDocument(Yaml.Document document, ExecutionContext executionContext) {
-                Yaml.Document doc = super.visitDocument(document, executionContext);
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the runnable example here @jonesbusy ! Had a brief look, and I suspect there's an logical error in the recipe.

As a bit of background: comments are not floating elements, but attached to the next LST element. That allows them to move with the code, as opposed to having to always check of and move both with any restructuring. In this case, it seems the recipe has made the assumption that we can remove the comment, and add it to the "next" element. But in your case there is no next element, so the comment is accidentally lost.

What we should be doing here (I think) is to add any left over comment to the `eof` of the matching document / documents (TBD). That should then make sure we don't accidentally lose any commented out elements at the end of a doc. I hope that's clear!

jonesbusy:
Thanks for the hint. Really helpful. I will try to take a look at it and perform also more test when the key is not the last element of the document

timtebeek (approved):
Thanks for taking this on! Exactly what I had in mind indeed; I've added & fixed another test case where there might be multiple documents, and with that I think we're good to merge once the tests pass.

jonesbusy:
Thanks for the fixed and merge! Much appreciated

Issue #4405, opened 2024-08-12
NoMissingTypes not work if the missingType is return value of method and the method has param


ckcd:
NoMissingTypes should find files with unknown imports, but it cannot work with this case:

```java
// this is an unknown import with invalid path
import java.util.GenClass1;
                  
public abstract class Foo {
      public abstract GenClass1 m(String a);
}
```

If use `RemoveUnusedImports` which uses `NoMissingTypes` as checker, it will wrongly remove the import.

But for below case, it will work for doing nothing.
```java
// this is an unknown import with invalid path
import java.util.GenClass1;
                  
public abstract class Foo {
      public abstract GenClass1 m();
}
```

The only difference is the first one has param (`String a`).

## What version of OpenRewrite are you using?

I am using
- Maven plugin 5.36.0

## How are you running OpenRewrite?

`mvn -X -B org.openrewrite.maven:rewrite-maven-plugin:5.36.0:runNoFork`

## What is the smallest, simplest way to reproduce the problem?

```java
import org.junit.jupiter.api.Test;
import org.openrewrite.java.RemoveUnusedImports;
import org.openrewrite.test.RecipeSpec;
import org.openrewrite.test.RewriteTest;
import org.openrewrite.test.TypeValidation;

import static org.openrewrite.java.Assertions.java;

public class RemoveUnuseImportsTest implements RewriteTest {
    @Override
    public void defaults(RecipeSpec spec) {
        spec.recipe(new RemoveUnusedImports());
    }

    // This test will fail, add it will remove all imports which is incorrect
    // The correct behavior is to remove none of the imports because we should not process files with missing types
    @Test
    void testWithGenClass1() {
        // language=java
        rewriteRun(
          spec -> spec.typeValidationOptions(TypeValidation.none()),
          java(
            """
                  import java.util.List;
                  import java.util.GenClass1;
                  
                  public abstract class Foo {
                      public abstract GenClass1 m(String a);
                  }
              """, """
                  import java.util.GenClass1;
                  
                  public abstract class Foo {
                      public abstract GenClass1 m(String a);
                  }
              """)
        );
    }

    // This test will fail, but it won't remove any import which is expected because we should not process files with missing types
    @Test
    void testWithGenClass2() {
        // language=java
        rewriteRun(
          spec -> spec.typeValidationOptions(TypeValidation.none()),
          java(
            """
                  import java.util.List;
                  import java.util.GenClass1;
                  
                  public abstract class Foo {
[...]

PR #4406, opened 2024-08-12
`NoMissingTypes` does not work if the missing type is a return value of a method and the method has a parameter

ckcd:
- Draft PR for https://github.com/openrewrite/rewrite/issues/4405, add tests to reproduce the issue.

@timtebeek Please help to perfect this fix if you are available, thanks.

Add some troubleshooting findings: it may related with this line:
https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java#L874
```java
m = m.getPadding().withParameters(visitContainer(m.getPadding().getParameters(), JContainer.Location.METHOD_DECLARATION_PARAMETERS, p));
```

This code will be called in `org.openrewrite.java.search.FindMissingTypes.FindMissingTypesVisitor#visitMethodDeclaration` because it will call `super.visitMethodDeclaration` which is `org.openrewrite.java.JavaVisitor#visitMethodDeclaration`. And the latter will in turn affect the `seenTypes` list of the former which causes the issue.


rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
```diff
@@ -1769,4 +1770,52 @@ public LocalDateTime toLocalDateTime(XMLGregorianCalendar cal) {
...
+              """
+          ),
+          java(
+            """
+              import java.util.GenClass1;
+
+              public abstract class Foo2 {
+                  public abstract GenClass1 m(String a);
+              }
+              """
```
timtebeek, line 1800: I'd assumed you don't want to see any change here; which is achieved by only having a single text block argument.

timtebeek:
This is a tricky one @ckcd ; I've found a way to get the tests going in https://github.com/openrewrite/rewrite/pull/4406/commits/27fc614f9e51132d73b1f71534234c55d079774a , but I'm not quite sure I like the code change just yet. Could you help think through from your earlier analysis if this is in line what you were thinking?

ckcd:
@timtebeek Thanks for the fix!! And yes it's align with my analysis I think. I will try with the snapshot version once we have it.

PR #4411, opened 2024-08-13
Build `GitRemote` url from it's parts

pstreef:
## What's changed?

Adds a static method on `GitRemote` that can be used to create a `URI` from the required parts:
```
URI buildRemoteUrl(Service service, boolean ssh, String origin, String path)
```

## What's your motivation?
A single place to put this logic, which is the reverse of the URL parser that is already in `GitRemote`

Are all these URI's (in the test) correct

@bryceatmoderne 

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -47,6 +47,10 @@ public class GitRemote {
         Unknown
     }
 
+    public URI toUri(boolean ssh) {
+        return buildRemoteUrl(service, ssh, origin, path);
+    }
+
     public static URI buildRemoteUrl(Service service, boolean ssh, String origin, String path) {
         StringBuilder url = new StringBuilder();
         path = path.replaceFirst("^/", "").replaceFirst("/$", "");
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -61,6 +61,7 @@ public class GitRemote {
         if (service == Service.AzureDevOps) {
             if (ssh) {
                 origin = "ssh." + origin;
+                path = "v3/" + path;
             } else {
                 path = path.replaceFirst("([^/]+)/([^/]+)/(.*)", "$1/$2/_git/$3");
             }
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -19,6 +19,7 @@ import lombok.Value;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.jgit.transport.URIish;
 
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
@@ -44,6 +45,34 @@ public class GitRemote {
         BitbucketCloud,
         AzureDevOps,
         Unknown
+    }
+
+    public static URI buildRemoteUrl(Service service, boolean ssh, String origin, String path) {
+        StringBuilder url = new StringBuilder();
+        path = path.replaceFirst("^/", "").replaceFirst("/$", "");
+        origin = origin.replaceFirst("/$", "");
+        if (service == Service.Bitbucket && !ssh) {
+            path = "scm/" + path;
+        }
+        if (service == Service.AzureDevOps) {
+            if (ssh) {
+                origin = "ssh." + origin;
+            } else {
+                path = path.replaceFirst("([^/]+)/([^/]+)/(.*)", "$1/$2/_git/$3");
+            }
+        } else if (service == Service.GitLab || ssh) {
+            path += ".git";
+        }
+        if (origin.startsWith("https://") || origin.startsWith("http://") || origin.startsWith("ssh://")) {
+            origin = origin.substring(origin.indexOf("://") + 3);
+        }
+        if (ssh) {
+            url.append("ssh://git@");
+        } else {
+            url.append("https://");
+        }
+        url.append(origin).append("/").append(path);
+        return URI.create(url.toString());
     }
 
     public static class Parser {
--- a/rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
@@ -122,7 +122,7 @@ public class GitRemoteTest {
       Bitbucket, false, scm.company.com/context/path, org/repo, https://scm.company.com/context/path/scm/org/repo
       Bitbucket, true, scm.company.com/context/path, org/repo, ssh://git@scm.company.com/context/path/org/repo.git
       AzureDevOps, false, dev.azure.com, org/repo, https://dev.azure.com/org/repo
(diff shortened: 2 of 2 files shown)


rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
```diff
@@ -102,4 +104,28 @@ void parseRegisteredRemote(String cloneUrl, String origin, GitRemote.Service ser
...
+      GitHub, false, scm.company.com/context/path, org/repo, https://scm.company.com/context/path/org/repo
+      GitHub, true, scm.company.com/context/path, org/repo, ssh://git@scm.company.com/context/path/org/repo.git
+      GitLab, false, gitlab.com, group/subgroup/repo, https://gitlab.com/group/subgroup/repo.git
+      GitLab, true, gitlab.com, group/subgroup/repo, ssh://git@gitlab.com/group/subgroup/repo.git
+      GitLab, false, scm.company.com/context/path, group/subgroup/repo, https://scm.company.com/context/path/group/subgroup/repo.git
+      GitLab, true, scm.company.com/context/path, group/subgroup/repo, ssh://git@scm.company.com/context/path/group/subgroup/repo.git
+      BitbucketCloud, false, bitbucket.org, org/repo, https://bitbucket.org/org/repo
+      BitbucketCloud, true, bitbucket.org, org/repo, ssh://git@bitbucket.org/org/repo.git
+      Bitbucket, false, scm.company.com/context/path, org/repo, https://scm.company.com/context/path/scm/org/repo
+      Bitbucket, true, scm.company.com/context/path, org/repo, ssh://git@scm.company.com/context/path/org/repo.git
```
shanman190, line 123: This one doesn't seem right by default for Bitbucket Data Center. It's default ssh clone port is 7999 and I'm not sure if it supports a context path for the ssh server.

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -46,6 +47,34 @@ public enum Service {
         Unknown
     }
 
+    public static URI buildRemoteUrl(Service service, boolean ssh, String origin, String path) {
```
shanman190, line 50: Now all of the components are here at the moment, but another alternative could be an instance method for `toURI(boolean)` which could convert the cloneUrl if it were present OR take the remaining components and build up the correct URI.

rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
pstreef: While 7999 is the standard port it is trivial to change it, same goes for a context path (reverse proxy)

I think we cannot assume 7999 and should have that as part of the origin.

But then the https and ssh origins are no longer the same.

It's a tricky situation either way, and I'm not sure what is the best way to handle it.

However, I know on the Moderne side we do assume the 7999 default so perhaps we need to just pick that as the lesser evil?

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: Yeah I was doubting if we need that, but might be useful down the line.

pstreef: I dont think we want to convert the clone url to be honest as it can come in in many forms (including scp style) and we want the output here to be a URI. Since the parser is already there, might as well just use the result directly

shanman190: For the SCP style URL, git internally just prefixes it with `ssh://`, so in theory we could do the same when we see the naked scp-style clone URL, but just always using the `buildRemoteUrl` works equally as well.

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -46,6 +47,38 @@ public enum Service {
...
+                path = path.replaceFirst("([^/]+)/([^/]+)/(.*)", "$1/$2/_git/$3");
+            }
+        } else if (service == Service.GitLab || ssh) {
+            path += ".git";
+        }
+        if (origin.startsWith("https://") || origin.startsWith("http://") || origin.startsWith("ssh://")) {
+            origin = origin.substring(origin.indexOf("://") + 3);
+        }
+        if (ssh) {
+            url.append("ssh://git@");
```
bryceatmoderne, line 74: Can you add ADO specific logic here, similar to 
```
         if (ssh) {
                if (scmType == RepositoryScmType.AzureDevOpsRepository) {
                    builder.append(":v3");
                } else if (scmType == RepositoryScmType.BitbucketRepository) {
                    int slashIndex = builder.indexOf("/", originIndex);
                    if (slashIndex != -1) {
                        builder.insert(slashIndex, ":7999");
                    } else {
                        builder.append(":7999");
                    }
                }
            }
```

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: I dont think `ssh://git@github.com:openrewrite/rewrite.git` is a valid url though
At least most parsers do not like it.

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: sure, however I do not want to build scp style urls. Even though that is what most SCM's give us I think it is best to build "real" URIs

bryceatmoderne: But the URI should be what is expected specific to the SCM no?

bryceatmoderne: I did some testing and the SCP syntax works, but for ADO you'll need to add a `v3/` segment. In other words

`ssh://git@ssh.dev.azure.com/moderneinc/moderne/git-test` does not work but `ssh://git@ssh.dev.azure.com/v3/moderneinc/moderne/git-test` does

```
[johnsmith@localhost ~/repos/azure-devops/clone-test2]$ git clone ssh://git@ssh.dev.azure.com/moderneinc/moderne/git-test
Cloning into 'git-test'...
remote: Command git-upload-pack '/moderneinc/moderne/git-test' is not in expected format.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
[johnsmith@localhost ~/repos/azure-devops/clone-test2]$ git clone ssh://git@ssh.dev.azure.com/v3/moderneinc/moderne/git-test
Cloning into 'git-test'...
remote: Azure Repos
remote: Found 31 objects to send. (125 ms)
Receiving objects: 100% (31/31), done.
```

bryceatmoderne: Ah, I was using an old diff then....

rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
bryceatmoderne, line 123: Have we agreed on _not_ adding the port 7999 to the ssh URL for Bitbucket?

pstreef: that is in a separate PR, but we probably need to decide that first: https://github.com/openrewrite/rewrite/pull/4414

PR #4414, opened 2024-08-14
Use 7999 as Bitbucket DC default port

pstreef:
## What's changed?
Assume 7999 as a default SSH port for Bitbucket and strip it off. This way we can match the origin to the https origin (without port)

## What's your motivation?
Currently when using the default port, we need to register 2 origins to make them both work. This can cause conflicts when switching between protocols.

@bryceatmoderne  @kmccarp 

Not doing this

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
@@ -79,6 +79,8 @@ public class GitRemoteTest {
       git@scm.company.com:stash/org/repo.git, scm.company.com/stash, Bitbucket, org/repo, org, repo
       ssh://scm.company.com/stash/org/repo, scm.company.com/stash, Bitbucket, org/repo, org, repo
       ssh://scm.company.com:22/stash/org/repo, scm.company.com/stash, Bitbucket, org/repo, org, repo
+      ssh://scm.company.com:22/stash/org/repo, scm.company.com:22/stash, Bitbucket, org/repo, org, repo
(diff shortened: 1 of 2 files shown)


rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
```diff
@@ -79,6 +79,8 @@ void parseUnknownRemote(String cloneUrl, String expectedOrigin, String expectedP
       git@scm.company.com:stash/org/repo.git, scm.company.com/stash, Bitbucket, org/repo, org, repo
       ssh://scm.company.com/stash/org/repo, scm.company.com/stash, Bitbucket, org/repo, org, repo
       ssh://scm.company.com:22/stash/org/repo, scm.company.com/stash, Bitbucket, org/repo, org, repo
+      ssh://scm.company.com:22/stash/org/repo, scm.company.com:22/stash, Bitbucket, org/repo, org, repo
+      ssh://scm.company.com:7999/stash/org/repo, scm.company.com/stash, Bitbucket, org/repo, org, repo
       ssh://scm.company.com:7999/stash/org/repo, scm.company.com:7999/stash, Bitbucket, org/repo, org, repo
```
bryceatmoderne, line 85: aren't these the same inputs? expected different results?

pstreef: no, the origin is both input and expected result

pstreef:
replaced by https://github.com/openrewrite/rewrite/pull/4429

PR #4418, opened 2024-08-15
Migrate to JSpecify from OpenRewrite JSR-305 meta-annotations

jkschneider:
Use this link to re-run the recipe: https://app.moderne.io/builder/gW3xaSLWV?organizationId=T3BlblJld3JpdGU%3D


jkschneider:
JSpecify `@NonNull` is enforced at runtime by Lombok generated constructors based on this annotation's presence on fields. This was not true for OpenRewrite's `@NonNull` so there is a bit of behavior change here. For now, we'll keep references to OpenRewrite `@NonNull` on fields and deal with this in another iteration.

PR #4429, opened 2024-08-19
Support multiple protocols and ports for a single `GitRemote` server and build URIs

pstreef:
## What's changed?
In stead of manipulating URIs and matching on a simple origin, we now register multiple URIs for a remote git server. This way we can indicate the multiple protocols/ports indeed point to the same server.

## What's your motivation?
We can currently get multiple origins for the same server which means we could fail to match.

the `normalizeUri` method

@bryceatmoderne

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -57,9 +57,9 @@ public class GitRemote {
             servers.add(new RemoteServer(Service.AzureDevOps, "dev.azure.com", URI.create("https://dev.azure.com"), URI.create("ssh://ssh.dev.azure.com")));
         }
 
-        // todo test
-
         /**
+         * Register a remote git server with multiple protocols and ports all matching the same server/origin.
+         *
          * @param service       the type of SCM service
          * @param remoteUri     the (main) origin of the server
          * @param alternateUris the alternate origins of the server
@@ -77,12 +77,13 @@ public class GitRemote {
         }
 
         /**
+         * Register a remote git server with a single origin with a single (supplied or guessed) protocol.
+         * If multiple protocols and/or ports should be supported use {@link #registerRemote(Service, URI, Collection)}
+         *
          * @param service the type of SCM service
          * @param origin  the origin of the server
          * @return this
-         * @deprecated Use {@link #registerRemote(Service, URI, Collection)} instead to configure multiple uris for a single remote server
          */
-        @Deprecated
         public Parser registerRemote(Service service, String origin) {
             add(new RemoteServer(service, origin, normalize(origin)));
             return this;
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -62,6 +62,9 @@ public class GitRemote {
             URI selectedBaseUrl;
 
             if (remote.service == Service.Unknown) {
+                if (PORT_PATTERN.matcher(remote.origin).find()) {
+                    throw new IllegalArgumentException("Unable to guess protocol port combination for an unregistered origin with a port: " + remote.origin);
+                }
                 selectedBaseUrl = URI.create((ssh ? "ssh://" : "https://") + stripProtocol(remote.origin));
             } else {
                 selectedBaseUrl = servers.stream()
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -58,18 +58,19 @@ public class GitRemote {
         }
 
         // todo test
+
         /**
          * @param service       the type of SCM service
-         * @param uri           the (main) origin of the server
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -47,69 +48,116 @@ public enum Service {
...
 
-        public Parser registerRemote(Service service, String origin) {
-            if (origin.startsWith("https://") || origin.startsWith("http://") || origin.startsWith("ssh://")) {
-                origin = new Parser.HostAndPath(origin).concat();
+        public URI toUri(GitRemote remote, boolean ssh) {
+            URI selectedBaseUrl;
+
+            if (remote.service == Service.Unknown) {
+                if (PORT_PATTERN.matcher(remote.origin).find()) {
+                    throw new IllegalArgumentException("Unable to guess protocol port combination for an unregistered origin with a port: " + remote.origin);
```
bryceatmoderne, line 66: nit: I'd prefer less casual phrasing than "guess". Perhaps extract or ascertain.... 

`"Unable to ascertain protocol/port combination for unregistered origin " + remote.origin`

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -47,69 +48,116 @@ public enum Service {
...
+            servers.add(new RemoteServer(Service.GitHub, "github.com", URI.create("https://github.com"), URI.create("ssh://git@github.com")));
+            servers.add(new RemoteServer(Service.GitLab, "gitlab.com", URI.create("https://gitlab.com"), URI.create("ssh://git@gitlab.com")));
+            servers.add(new RemoteServer(Service.BitbucketCloud, "bitbucket.org", URI.create("https://bitbucket.org"), URI.create("ssh://git@bitbucket.org")));
+            servers.add(new RemoteServer(Service.AzureDevOps, "dev.azure.com", URI.create("https://dev.azure.com"), URI.create("ssh://git@ssh.dev.azure.com")));
         }
 
-        public Parser registerRemote(Service service, String origin) {
-            if (origin.startsWith("https://") || origin.startsWith("http://") || origin.startsWith("ssh://")) {
-                origin = new Parser.HostAndPath(origin).concat();
+        public URI toUri(GitRemote remote, boolean ssh) {
```
bryceatmoderne, line 61: Would it make sense to change the `ssh` parameter to `protocol` to allow for `http` or `https`?

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -47,69 +48,116 @@ public enum Service {
...
-            if (service == Service.Unknown) {
-                // Do not override a known with an unknown service
-                origins.putIfAbsent(origin, service);
-            } else {
-                origins.put(origin, service);
+            String maybeSlash = selectedBaseUrl.toString().endsWith("/") ? "" : "/";
+            return URI.create(selectedBaseUrl + maybeSlash + path);
+        }
+
+        private static String stripProtocol(String origin) {
```
bryceatmoderne, line 106: nit: can replace with a one-liner using `origin.replaceFirst("\\w+://", "")`

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -129,66 +177,134 @@ public GitRemote parse(String url) {
...
+                        if (url.contains(":")) {
+                            // Looks like SCP style url
+                            scheme = "ssh";
+                        } else {
+                            scheme = "https";
+                        }
+                        uri = new URIish(scheme + "://" + url);
+                        host = uri.getHost();
+                    } else if (!"file".equals(scheme)) {
+                        throw new IllegalStateException("No host in url: " + url);
```
bryceatmoderne, line 228: nit: can you uppercase the `url` in the message? `No host found in URL " + url`

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -129,66 +177,134 @@ public GitRemote parse(String url) {
...
+                    }
                 }
-                if (!path.isEmpty()) {
-                    if (builder.length() != 0) {
-                        builder.append('/');
+                if (scheme == null) {
+                    if (PORT_PATTERN.matcher(url).find()) {
+                        throw new IllegalArgumentException("Unable to normalize URI. Port without a scheme is not supported: " + url);
```
bryceatmoderne, line 233: nit: `"Unable to normalize URL: Specifying a port without a scheme is not supported for URL " + url`

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: Updated to use "determine". which I like a bit better.

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
pstreef: I did not think it was very useful, but it's easy to add now, harder to change later. So lets!

PR #4446, opened 2024-08-26
Added arrays to AddOrUpdateAnnotationAttribute

marcel-gepardec:
Added the option to add, replace or remove string arrays in annotation attributes.

## What's changed?
The Implementation of the Recipe AddOrUpdateAnnotationAttribute is now able to add, replace or remove a string array in a annotation attribute.

Example:
```yml
- org.openrewrite.java.AddOrUpdateAnnotationAttribute:
     annotationType: org.example.Foo
     attributeName: array
     attributeValue: newTest
     addOnly: false
```

Interface:
```java
package org.example;
public @interface Foo {
    String[] array() default {};
}
```

Before:
```java
import org.example.Foo;

@Foo(array = {"oldTest"})
public class A {
}
```

After:
```java
import org.example.Foo;

@Foo(array = {"newTest"})
public class A {
}
```

## What's your motivation?
A work colleague was using this part of the project and discovered the missing implementation, so he handed it over to me to implement it.


sambsnyd:
Thanks @marcel-gepardec !

PR #4447, opened 2024-08-26
Do not change package when non-recursive

timtebeek:
Add unit test as reported via Slack by @BhavanaPidapa; update precondition to evaluate recursive boolean for package declaration too.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
@@ -88,14 +88,14 @@ public class ChangePackage extends Recipe {
             public @Nullable J preVisit(J tree, ExecutionContext ctx) {
                 if (tree instanceof JavaSourceFile) {
                     JavaSourceFile cu = (JavaSourceFile) requireNonNull(tree);
+                    boolean recursive = ChangePackage.this.recursive == null || ChangePackage.this.recursive;
                     if (cu.getPackageDeclaration() != null) {
                         String original = cu.getPackageDeclaration().getExpression()
                                 .printTrimmed(getCursor()).replaceAll("\\s", "");
-                        if (original.startsWith(oldPackageName)) {
+                        if (original.equals(oldPackageName) || recursive && original.startsWith(oldPackageName)) {
                             return SearchResult.found(cu);
                         }
                     }
-                    boolean recursive = Boolean.TRUE.equals(ChangePackage.this.recursive);
                     String recursivePackageNamePrefix = oldPackageName + ".";
                     for (J.Import anImport : cu.getImports()) {
                         String importedPackage = anImport.getPackageName();
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
@@ -71,6 +71,15 @@ class ChangePackageTest implements RewriteTest {
               class Test {
               }
               """
+          ),
+          java(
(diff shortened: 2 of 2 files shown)


anandfresh (commented):
LGTM

BhavanaPidapa:
Hey Tim,

Below is the test case where changes to `com.sun.net.ssl.internal.*` are not expected. Please refer to the attached screenshot.

@Test
void excludePackageTest() {
    //language=java
    rewriteRun(
      spec -> spec.recipe(new ChangePackage(
        "com.sun.net.ssl",
        "javax.net.ssl",
        false
      )),
      java(
        """
          package com.test;
          import com.sun.net.ssl.internal.*;
          import com.sun.net.ssl.HostnameVerifier;
          class Test {
           public com.sun.net.ssl.HostnameVerifier hv;
              void exampleMethod() {
              }
          }
          """,
        """
          package com.test;
          import com.sun.net.ssl.internal.*;
          import javax.net.ssl.HostnameVerifier;
          class Test {
           public javax.net.ssl.HostnameVerifier hv;
              void exampleMethod() {
              }
          }
          """
      )
    );
}

image

timtebeek:
This is an interesting case: we first and separate evaluate a precondition to determine if we need further changes; that precondition resolves to true due to the presence of `com.sun.net.ssl.HostnameVerifier`; then we unconditionally change any package references in the actual visitor, whereas we should probably still take into account there whether we should do so recursively.

knutwannheden:
Maybe fixed by https://github.com/openrewrite/rewrite/pull/4942?

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -91,14 +91,14 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                    boolean recursive = ChangePackage.this.recursive == null || ChangePackage.this.recursive;
                     if (cu.getPackageDeclaration() != null) {
                         String original = cu.getPackageDeclaration().getExpression()
                                 .printTrimmed(getCursor()).replaceAll("\\s", "");
-                        if (original.startsWith(oldPackageName)) {
+                        if (original.equals(oldPackageName) || recursive && original.startsWith(oldPackageName)) {
                             return SearchResult.found(cu);
                         }
                     }
-                    boolean recursive = Boolean.TRUE.equals(ChangePackage.this.recursive);
```
timtebeek, line 101: This is where things get tricky; we didn't document the default behaviour, but line 101 indicates it would have been non-recursive by default.

The bug report here though is we were seeing recursive updates when explicitly disabled. It's a question whether folks would have become accustomed to that unintended behavior downstream in the changes they (now) expect. 

Wondering if we should make further adjustments based on those expectations, or partially revert these changes and only keep the `visitImport` change for instance.

rewrite-java/src/main/java/org/openrewrite/java/internal/FindFullyQualifiedTypeReferences.java
```diff
@@ -0,0 +1,78 @@
...
+import org.openrewrite.java.tree.JavaSourceFile;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Finds fully qualified type references in the AST that may not be captured in TypesInUse
+ * when the types are not on the classpath (and thus marked as Unknown).
+ */
+public class FindFullyQualifiedTypeReferences extends JavaIsoVisitor<Integer> {
```
timtebeek, line 31: Claude went off the rails here a bit; no too pleased seeing this file added.

PR #4454, opened 2024-08-29
Upgrade dependencies in custom gradle scripts

pstreef:
Some users split their `build.gradle` using custom scripts which are applied using `apply from: custom.gradle`
This works in a similar way as when the custom script would be inserted into the `build.gradle` (with some differences)

When parsing a custom script, the parser does not know the root type and this makes it fail to match `dependencies` blocks. If we are more lenient on the matching in known custom scripts we can still allow these upgrades to succeed

## What's changed?
Custom scripts now can have their dependency blocks get upgrades

## What's your motivation?
Users would like to upgrade dependencies in such custom scripts

is there a better way to match a method when it's type is `null`?

-

Maybe we can fix the type attribution

## Any additional context
Currently it seems the parser does not yet support this and sees the custom gradle script as a standalone script. Meaning in the case where it just has a dependencies block it will fail because it thinks the java plugin is not applied.

### Checklist


pstreef:
requires: https://github.com/openrewrite/rewrite-gradle-tooling-model/pull/22

pstreef:
we decided to do a proper fix on the LST generation level.
https://moderneinc.slack.com/archives/C01VADFPJQZ/p1724926907451669

PR #4458, opened 2024-08-29
`ChangeType` should fully qualify type usage in the case of conflicting imports

### LINKED ISSUE #4452: ChangeType is not addressing import conflicts

I am using

- OpenRewrite v8.33.8
- Gradle plugin v8.4
- rewrite-java v8.33.8

```java
public class ChangeTypeTest implements RewriteTest {

    @Override
        spec.recipe(
                new ChangeType("javax.annotation.Nonnull", "org.checkerframework.checker.nullness.qual.NonNull", null));
    }

    @Test
    void conflictingImports() {

        // language=java
        rewriteRun(
                // dissabling validation to avoid the 'Identifier type is missing or malformed' error. I couldn't add
                // the lombok dependency to the test.
                spec -> spec.typeValidationOptions(TypeValidation.none()),
                java(
                        """
                        import lombok.NonNull;
                        import javax.annotation.Nonnull;
                        import org.immutables.value.Value;

                        @Value.Immutable
                        @Value.Style(passAnnotations = Nonnull.class)
                        public interface ConflictingImports {
                                private void lombokMethod(@NonNull final String lombokNonNull){}
                        }
                        """,
                        """
                        import lombok.NonNull;
                        import org.immutables.value.Value;

                        @Value.Immutable
                        @Value.Style(passAnnotations = org.checkerframework.checker.nullness.qual.NonNull.class)
                        public interface ConflictingImports {
                                private void lombokMethod(@NonNull final String lombokNonNull){}
                        }
                        """));
    }
}
```

If where is already a class imported with the same name in the original code (lombok.NonNull in the example), the recipe should fully-qualified the new Type included (org.checkerframework.checker.nullness.qual.NonNull). 

## What did you see instead?
I'm seeing two imports for the same name type, from different packages.

```java
import lombok.NonNull;
import org.checkerframework.checker.nullness.qual.NonNull;
```

```diff
diff --git a/ConflictingImports.java b/ConflictingImports.java
index 4172665..31c1781 100644
--- a/ConflictingImports.java
+++ b/ConflictingImports.java
@@ -1,8 +1,9 @@ 
 import lombok.NonNull;
+import org.checkerframework.checker.nullness.qual.NonNull;
 import org.immutables.value.Value;
 
 @Value.Immutable
-@Value.Style(passAnnotations = org.checkerframework.checker.nullness.qual.NonNull.class)
+@Value.Style(passAnnotations = NonNull.class)
 public interface ConflictingImports {
         private void lombokMethod(@NonNull final String lombokNonNull){}
 }
\ No newline at end of file
```
expected: 
```java
  import lombok.NonNull;
  import org.immutables.value.Value;
  
  @Value.Immutable
  @Value.Style(passAnnotations = org.checkerframework.checker.nullness.qual.NonNull.class)
  public interface ConflictingImports {
          private void lombokMethod(@NonNull final String lombokNonNull){}
  }
```

 but was: 
```java
  import lombok.NonNull;
  import org.checkerframework.checker.nullness.qual.NonNull;
  import org.immutables.value.Value;
  
  @Value.Immutable
  @Value.Style(passAnnotations = NonNull.class)
  public interface ConflictingImports {
          private void lombokMethod(@NonNull final String lombokNonNull){}
  }
```

## Extra information, slack thread about this issue:
https://rewriteoss.slack.com/archives/C01A843MWG5/p1723733741135269?thread_ts=1723658700.585459&cid=C01A843MWG5

Not at this moment. Maybe in the future.

Laurens-W:
## What's changed?

Detect when an import becomes ambiguous and once it does start fully qualifying references to the new type

## What's your motivation?

- Fixes #4452 

@timtebeek 

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -505,6 +511,18 @@ private Expression updateOuterClassTypes(Expression typeTree) {
         private boolean isTargetFullyQualifiedType(JavaType.@Nullable FullyQualified fq) {
             return fq != null && TypeUtils.isOfClassType(fq, originalType.getFullyQualifiedName()) && targetType instanceof JavaType.FullyQualified;
         }
+
+        private boolean becomesAmbiguous(JavaSourceFile cu) {
```
timtebeek, line 515: This once again gave me warnings when opened in the IDE; please keep that `F2` before commit in mind, and 
image

Also be sure to enable this option for yet another nudge to not push any code with new warnings; that helps me review & improve code quality.
image

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -505,6 +511,18 @@ private Expression updateOuterClassTypes(Expression typeTree) {
...
+            JavaType.FullyQualified newType = TypeUtils.asFullyQualified(targetType);
+            JavaType.FullyQualified oldType = TypeUtils.asFullyQualified(originalType);
+
+            return newType != null && oldType != null && cu.getImports().stream().anyMatch(imprt -> {
```
timtebeek, line 519: Try to avoid using the streams API, even when convenient, but especially on paths that might get evaluated often; Adding imports is one of those hot paths where we really want to avoid these object allocations where we can.

timtebeek (approved):
Great to see this case fixed! Thanks!

timtebeek:
We now see some failures downstream related to imports: 
https://ge.openrewrite.org/s/fyq3wpbm56qgk/tests/task/:test/details/org.openrewrite.apache.httpclient4.CookieConstantsTest/cookieConstantsMapping()?top-execution=1

knutwannheden:
While working on #4460 I noticed that `TypeTree#build()` (which is used by `ChangeType`) produces `J.FieldAccess` (or `J.Identifier`) trees, where the type attribution isn't entirely correct. For the last segment the corresponding `J.Identifier` doesn't have a type. As a consequence of that the `SimplifyQualifiedTypeNames` visitor didn't work correctly. I tried to fix this behavior but then some `ChangeType` tests started failing, so I ended up reverting those changes and do that locally as part of the `AddMethodParameter` visitor instead.

I just wanted to mention this here, as the PR seems somewhat related and I think it would be great if we could fix that `TypeTree#build()` bug at some point. Possibly you have some insights regarding this.

Laurens-W:
> We now see some failures downstream related to imports: https://ge.openrewrite.org/s/fyq3wpbm56qgk/tests/task/:test/details/org.openrewrite.apache.httpclient4.CookieConstantsTest/cookieConstantsMapping()?top-execution=1

A situation where `newType` was already being used wasn't correctly covered, [this](https://github.com/openrewrite/rewrite/pull/4461) fixes it

PR #4463, opened 2024-09-03
Also exclude omitted dependencies

timtebeek:
## What's your motivation?
When conflicts occur dependencies are omitted; when we then exclude from the conflict, the omitted dependency can resurface. The goal of ExcludeDependency is to fully exclude a dependency, not to see the omitted conflict being used. We should then also add `<exclusion>` where dependencies are omitted.

### Checklist


timtebeek:
- Replaced by https://github.com/openrewrite/rewrite/pull/4563

PR #4471, opened 2024-09-04
Add DoesNotUseType recipe for use in preconditions to reject source files

Laurens-W:
## What's your motivation?

This recipe is required to be used as a precondition for JUnit 4 to 5 migration, which we don't want to do if any TestNG type is used

Test coverage

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/NotUsesType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/NotUsesType.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.search;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+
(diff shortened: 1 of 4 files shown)


rewrite-java/src/test/java/org/openrewrite/java/search/NotUsesTypeTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new NotUsesType("java.lang.String", true));
+    }
+
+    @DocumentExample
+    @Test
+    void doesNotUseType() {
+        rewriteRun(
+          spec -> spec.expectedCyclesThatMakeChanges(1),
```
timtebeek, line 36:
```suggestion
```

rewrite-java/src/test/java/org/openrewrite/java/search/NotUsesTypeTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class NotUsesTypeTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new NotUsesType("java.lang.String", true));
```
timtebeek, line 29: Perhaps it's better to show the intended usage here through `.recipeFromYaml`, containing a declarative recipe that uses `NotUsesType` as a precondition, before calling `OrderImports` in the `recipeList:`. Then each of the tests below can show that unused imports are (not) removed.

rewrite-java/src/main/java/org/openrewrite/java/search/NotUsesType.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Inverse of `UsesType`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.not(new UsesType<>(fullyQualifiedType, includeImplicit));
+    }
```
timtebeek, line 51: 1. The fields here are likely not nullable, so remove the nullable annotations; that then also removes these IDE warnings.
image

2. Also remove the `private final` as indicated by my IDE.

3. Let's rename class to `DoesNotUseType`, as `NotUsesType` sounds a bit awkward.

4. "Inverse of `UsesType`" isn't very descriptive in and of itself; instead let's describe the intended use as a precondition to invalidate classes using a particular type. e.g. do not run recipe X when type Y is used in a class.

timtebeek (approved):

PR #4482, opened 2024-09-10
`RemoveUnusedImports` should retain explicit import when there are conflicting classes in the same package

Laurens-W:

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
@@ -94,7 +94,7 @@ class RemoveUnusedImportsTest implements RewriteTest {
           java(
             """
               import java.util.Map.Entry;
-
+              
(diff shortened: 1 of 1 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
```diff
@@ -1983,4 +1983,56 @@ public void foo() {
...
+          ),
+          java(
+            """
+              package com.a;
+
+              import com.b.ConflictingClass;
+              import com.c.ImplProvider;
+
+              class CImpl {
+                 ConflictingClass impl = ImplProvider.getImpl();
```
timtebeek, line 2032:
```suggestion
                  ConflictingClass impl = ImplProvider.getImpl();
```

timtebeek (approved):
Great to see this case was now already handled; Approved as a safe guard going forward with one last small suggestion to go from 3 to 4 spaces in test class.

PR #4483, opened 2024-09-11
Improve handling of inner classes

Laurens-W:
## What's changed?

Improved handling of inner classes

## What's your motivation?

https://rewriteoss.slack.com/archives/C01A843MWG5/p1725947336140549

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
@@ -724,38 +724,4 @@ class OrderImportsTest implements RewriteTest {
           )
         );
     }
-
-    @Test
-    void staticNestedInnerClass() {
-        // language=java
-        rewriteRun(
-          spec -> spec.recipe(new OrderImports(true)),
-          java(
-            """
-              package a;
-              
-              public class A {
-                  public final class B {}
-                  public static class C {}
-              }
-              """,
-            SourceSpec::skip),
-          java(
-            """
-              import a.*;
-              
-              public class Foo {
-                  A method(A.B ab, A.C ac) {}
-              }
-              """,
-            """
-              import a.A;
-              
-              public class Foo {
-                  A method(A.B ab, A.C ac) {}
-              }
-              """
-          )
-        );
-    }
 }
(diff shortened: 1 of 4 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
```diff
@@ -686,4 +684,78 @@ void foldPackageWithExistingImports() {
           )
         );
     }
+
+    @Test
+    void nestedInnerClass() {
+        // language=java
+        rewriteRun(
+          spec -> spec.recipe(new OrderImports(true)),
```
timtebeek, line 692: I only see code changes in `ImportLayoutStyle` and `OrderImports`. Is it necessary to test through `new OrderImports(true)` here, or can we move these tests to `OrderImportsTest` ?

Laurens-W: I think we can move it!

rewrite-java/src/main/java/org/openrewrite/java/style/ImportLayoutStyle.java
```diff
@@ -788,11 +788,8 @@ private static String packageOrOuterClassName(JRightPadded<J.Import> anImport) {
         if (anImport.getElement().isStatic()) {
             return typeName;
         } else {
-            String className = anImport.getElement().getClassName();
-            if (className.contains("$")) {
-                return anImport.getElement().getPackageName() + "." +
-                        className.substring(0, className.lastIndexOf('$'))
-                                .replace('$', '.');
+            if (typeName.contains("$")) {
+                return typeName.substring(0, typeName.lastIndexOf('$'));
```
timtebeek, line 792: Was this a necessary change, or just a simplification you'd thought to apply?
The way it's written I wonder if it had meant to handled twice-nested inner classes.

Laurens-W: This was a necessary change as it never reached inside the if because `anImport.getElement().getClassName();` returns an already sanitized reference, you were right about having to handle multiple inner classes though, So I've updated it

timtebeek (approved):

PR #4489, opened 2024-09-12
Order imports should retain nested enum field imports

### LINKED ISSUE #4165: OrderImports breaks the build by collapsing enum import + whitespace inconsistencies

I am using

- Maven/Gradle plugin v5.29.0

I am using the Maven plugin, and my project is a single module project.
```xml
  <profiles>
    <profile>
      <id>openrewrite</id>
      <!-- `mvn -P openrewrite org.openrewrite.maven:rewrite-maven-plugin:run` -->
      <build>
        <plugins>
          <plugin>
            <artifactId>rewrite-maven-plugin</artifactId>
            <version>5.29.0</version>
            <configuration>
              <activeRecipes>
                <recipe>org.openrewrite.java.OrderImports</recipe>
              </activeRecipes>
              <activeStyles>
                <style>com.yourorg.JacksonImportStyle</style>
              </activeStyles>
              <failOnDryRunResults>true</failOnDryRunResults>
            </configuration>
            <dependencies>
              <dependency>
                <groupId>com.yourorg</groupId>
                <artifactId>rewrite-recipe-starter</artifactId>
                <version>0.1.0-SNAPSHOT</version>
              </dependency>
            </dependencies>
          </plugin>
        </plugins>
      </build>
    </profile>
  </profiles>
```

## What is the ~smallest, simplest~ way to reproduce the problem?

This example is not minimal, but it is reproducible.

1. Checkout https://github.com/timo-a/rewrite-recipe-starter/commit/297e72f259f17cae8d62dd7c29b8481f426a37ca which adds a custom style:

```java
    public static ImportLayoutStyle importLayout() {
        return ImportLayoutStyle.builder()
                .importAllOthers()
                .blankLine()
                .importPackage("com.fasterxml.jackson.core.*")
                .importPackage("tools.jackson.*")
                .blankLine()
                .importStaticAllOthers()
                .build();
    }
```

2. publish this style locally with `./gradlew publishToMavenLocal`
3. checkout https://github.com/timo-a/jackson-core/commit/4e44f2d2ca9b98f8fd2d094d9982b793ddf60121 
4. run `mvn -P openrewrite org.openrewrite.maven:rewrite-maven-plugin:run`

## What did you see?

this change set: https://github.com/timo-a/jackson-core/commit/c2970c0192090c3cb35f7bd32332d5dd49cde76f 

1. Collapsing enum imports into a wildcard breaks the build.
```java
import com.fasterxml.jackson.core.util.DefaultIndenter;
import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
import com.fasterxml.jackson.core.util.MinimalPrettyPrinter;
import com.fasterxml.jackson.core.util.Separators;
import com.fasterxml.jackson.core.util.Separators.Spacing;
```
are replaced with 
```java
import com.fasterxml.jackson.core.util.*;
```
 in [PrettyPrinterTest](https://github.com/timo-a/jackson-core/commit/c2970c0192090c3cb35f7bd32332d5dd49cde76f#diff-63b3384ea7a7587219e99d47918dfddac5216004ba7ca8528540d05cfedbb928). As a consequence `Spacing.NONE` in line 271 can no longer be resolved. I don't know what issue the compiler has here, but apparently there is an edge case concerning enum imports. Other files are affected as well, but it is the same pattern there.

2. `extends` on new line is indented 
Example: https://github.com/timo-a/jackson-core/commit/c2970c0192090c3cb35f7bd32332d5dd49cde76f#diff-88fe5e2a29aa76eab8d3ec510f7aeb1ed3505a08c183f74e01543ff398ef7d0d
The docs only speak of ordering imports:

> Groups and orders import statements. If a [style has been defined](https://docs.openrewrite.org/concepts-explanations/styles), this recipe will order the imports according to that style. If no style is detected, this recipe will default to ordering imports in the same way that IntelliJ IDEA does.

so this is unexpected.

3. Unwanted blank lines
Based on my style specification, I expect imports to be split in at most three groups: other, my explicitly defined package and static imports. However there is always (?) a blank line after java imports while not necessarily above it,
example: https://github.com/timo-a/jackson-core/commit/c2970c0192090c3cb35f7bd32332d5dd49cde76f#diff-fc3028f1ae776b0c8a46f8cacd73ecad36c879df456fb575d69ff48f5a2c3291

no, just reporting

timo-a:
## What's your motivation?
- Fixes #4165 observation 1.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/style/ImportLayoutStyle.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/style/ImportLayoutStyle.java
@@ -233,7 +233,7 @@ public class ImportLayoutStyle implements JavaStyle {
         }
 
         if (isFoldable && (((paddedToAdd.getElement().isStatic() && nameCountToUseStarImport <= sameCount) ||
-                (!paddedToAdd.getElement().isStatic() && classCountToUseStarImport <= sameCount)) || isPackageAlwaysFolded(packagesToFold, paddedToAdd.getElement()))) {
+                            (!paddedToAdd.getElement().isStatic() && classCountToUseStarImport <= sameCount)) || isPackageAlwaysFolded(packagesToFold, paddedToAdd.getElement()))) {
             starFold.set(true);
             if (insertPosition != starFoldFrom.get()) {
                 // if we're adding to the middle of a group of imports that are getting star folded,
@@ -258,7 +258,7 @@ public class ImportLayoutStyle implements JavaStyle {
             if (block(after) == addToBlock) {
                 after = after.withElement(after.getElement().withPrefix(Space.format("\n")));
             } else if (!isNewBlock.get() && after.getElement().getPrefix().getLastWhitespace().chars()
-                    .filter(c -> c == '\n').count() < 2) {
+                                                    .filter(c -> c == '\n').count() < 2) {
                 after = after.withElement(after.getElement().withPrefix(Space.format("\n\n")));
             }
         }
@@ -350,8 +350,8 @@ public class ImportLayoutStyle implements JavaStyle {
                         // Preserve the existing newline character type of either CRLF or LF.
                         // Classic Mac OS new line return '\r' is replaced by '\n'.
                         String newLineCharacters = whitespaceContainsCRLF ||
-                                StringUtils.isNullOrEmpty(orderedImport.getElement().getPrefix().getWhitespace()) &&
-                                        "\r\n".equals(prevWhitespace) ? "\r\n" : "\n";
+                                                   StringUtils.isNullOrEmpty(orderedImport.getElement().getPrefix().getWhitespace()) &&
+                                                   "\r\n".equals(prevWhitespace) ? "\r\n" : "\n";
 
                         StringBuilder newWhitespace = new StringBuilder(newLineCharacters);
                         for (int i = 0; i < extraLineSpaceCount; i++) {
@@ -399,7 +399,7 @@ public class ImportLayoutStyle implements JavaStyle {
 
         public Builder blankLine() {
             if (!blocks.isEmpty() &&
-                    blocks.get(blocks.size() - 1) instanceof Block.BlankLines) {
+                blocks.get(blocks.size() - 1) instanceof Block.BlankLines) {
                 ((Block.BlankLines) blocks.get(blocks.size() - 1)).count++;
             } else {
                 blocks.add(new Block.BlankLines());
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the clear runnable test @timo-a and thanks for diving in for the fix @Laurens-W ! Great to see this come together to fix a long standing issue.

rewrite-java/src/main/java/org/openrewrite/java/style/ImportLayoutStyle.java
```diff
@@ -785,14 +785,15 @@ public String toString() {
 
     private static String packageOrOuterClassName(JRightPadded<J.Import> anImport) {
         String typeName = anImport.getElement().getTypeName();
-        if (anImport.getElement().isStatic()) {
+        JavaType.FullyQualified type = TypeUtils.asFullyQualified(anImport.getElement().getQualid().getType());
+        if (anImport.getElement().isStatic() || (type != null && type.getKind() == JavaType.FullyQualified.Kind.Enum)) {
```
timtebeek, line 789: I don't think the exceptional handling of enums here is quite right; if I alter the above test slightly you can see an import that should be folded, but isn't.

```java
@Issue("https://github.com/openrewrite/rewrite/issues/4165")
@Test
void doNotPreserveOuterEnum() {
    rewriteRun(
      java(
        """
          package com.fasterxml.jackson.core.util;
          
          public enum Spacing { NONE, BEFORE, AFTER, BOTH }
          """,
        SourceSpec::skip
      ),
      java(
        """
          import com.fasterxml.jackson.core.util.DefaultIndenter;
          import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
          import com.fasterxml.jackson.core.util.JacksonFeature;
          import com.fasterxml.jackson.core.util.MinimalPrettyPrinter;
          import com.fasterxml.jackson.core.util.Separators;
          import com.fasterxml.jackson.core.util.Spacing;
          
          class PrettyPrinterTest {
              void print() {
                  System.out.println(Spacing.NONE);
              }
          }
          """,
        """
          import com.fasterxml.jackson.core.util.*;
          
          class PrettyPrinterTest {
              void print() {
                  System.out.println(Spacing.NONE);
              }
          }
          """
      )
    );
}
```

As this fails with:
```diff
diff --git a/PrettyPrinterTest.java b/PrettyPrinterTest.java
index 3267855..c62317b 100644
--- a/PrettyPrinterTest.java
+++ b/PrettyPrinterTest.java
@@ -1,4 +1,5 @@ 
 import com.fasterxml.jackson.core.util.*;
+import com.fasterxml.jackson.core.util.Spacing;
 
 class PrettyPrinterTest {
     void print() {
```

timtebeek (changes requested):
Let's first see through
- https://github.com/openrewrite/rewrite/pull/4483

Then see if we can rework this to not be specific to Enums, but specifc to nested classes or enums, if it's even still needed.

rewrite-java/src/main/java/org/openrewrite/java/style/ImportLayoutStyle.java
Laurens-W: Apparently in this case the enum does not get marked as static, checking the flags instead of the kind solves this

timtebeek:
I've reverted the functional changes here after merging
- https://github.com/openrewrite/rewrite/pull/4483

Locally it seems like we don't need any further changes to make these tests pass; which then makes me doubt if we should even merge this PR, as we wouldn't want any duplicate test coverage. Thoughts?

timtebeek:
The failures are unrelated Let's close this one. Thanks both for looking into this!

PR #4491, opened 2024-09-13
Adding gradle wrapper through distribution is now possible

### LINKED ISSUE #4490: Update gradle wrapper adds wrong wrapper when customDistributionUrl is requested.
In the event of a gradle wrapper that needs to be added to a project that does not contain one AND the distributionUrl is mentioned, this code will no add the correct wrapper. 

If the outside world is reachable, this try block does not fail -> the distribution url is not used and the wrong wrapper gets added. 

https://github.com/openrewrite/rewrite/blob/0b6914b82be589bb6924f273c7ee2cb6c873b853/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java#L124-L160

The Gradle wrapper will be created using the outside world services gradle and we will never reach the distributionUrl wrapper additions. 

https://github.com/openrewrite/rewrite/blob/0b6914b82be589bb6924f273c7ee2cb6c873b853/rewrite-gradle/src/main/java/org/openrewrite/gradle/util/GradleWrapper.java#L60-L90

The recipe from the list: 
Screenshot 2024-09-13 at 12 04 19

The output: 
Screenshot 2024-09-13 at 12 04 45

I think if we rework the getWrapper method we could fix this. 

```java 
    private GradleWrapper getGradleWrapper(ExecutionContext ctx) {
        if (gradleWrapper != null) {
            return gradleWrapper;
        }
        if (wrapperUri != null) {
            if (wrapperUri.contains("${version})")) {
                if (version == null) {
                    throw new IllegalArgumentException(
                            "wrapperUri contains a ${version} interpolation specifier but no version parameter was specified.");
                }
                if (!version.matches("[0-9.]+")) {
                    throw new IllegalArgumentException(
                            "Version selectors like \"" + version + "\" are unavailable when custom distribution url is used. " +
                            "Specify an exact, literal version number.");
                }
            }
            String effectiveWrapperUri = wrapperUri
                    .replace("${version}", version == null ? "" : version)
                    .replace("${distribution}", distribution == null ? "bin" : distribution);
            gradleWrapper = GradleWrapper.create(URI.create(effectiveWrapperUri), ctx);
        } else {
            try {
                gradleWrapper = GradleWrapper.create(distribution, version, null, ctx);
            } catch (Exception e) {
                // services.gradle.org is unreachable, possibly because of a firewall
                // But if the user specified a wrapperUri to an internal repository things might still be workable
                // If the user didn't specify a wrapperUri, but they did provide a specific version we assume they know this version
                // is available from whichever distribution url they were previously using and update the version
                if (!StringUtils.isBlank(version) && Semver.validate(version, null).getValue() instanceof ExactVersion) {
                    gradleWrapper = new GradleWrapper(version, new DistributionInfos("", null, null));
                } else {
                    throw new IllegalArgumentException(
                            "Could not reach services.gradle.org, no alternative wrapper URI is provided and no exact version is provided. " +
                            "To use this recipe in environments where services.gradle.org is unavailable specify a wrapperUri or exact version.", e);
                }
            }
        }
        return gradleWrapper;
    }
```

Jenson3210:
- Fixes #4490

By allowing only distributionUri OR version/distribution, we also no longer use gradle services to determine the version which could not be available on the distributionUrl

Follow up from
- https://github.com/openrewrite/rewrite/pull/4445

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
@@ -22,16 +22,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
 import org.jspecify.annotations.Nullable;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.FileAttributes;
-import org.openrewrite.Option;
-import org.openrewrite.PathUtils;
-import org.openrewrite.Preconditions;
-import org.openrewrite.ScanningRecipe;
-import org.openrewrite.SourceFile;
-import org.openrewrite.Tree;
-import org.openrewrite.TreeVisitor;
-import org.openrewrite.Validated;
+import org.openrewrite.*;
 import org.openrewrite.gradle.search.FindGradleProject;
 import org.openrewrite.gradle.util.DistributionInfos;
 import org.openrewrite.gradle.util.GradleWrapper;
@@ -52,25 +43,11 @@ import org.openrewrite.text.PlainText;
 
 import java.net.URI;
 import java.time.ZonedDateTime;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 
 import static java.util.Objects.requireNonNull;
 import static org.openrewrite.PathUtils.equalIgnoringSeparators;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_BATCH_LOCATION;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_BATCH_LOCATION_RELATIVE_PATH;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_JAR_LOCATION;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_JAR_LOCATION_RELATIVE_PATH;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_PROPERTIES_LOCATION;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_PROPERTIES_LOCATION_RELATIVE_PATH;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_SCRIPT_LOCATION;
-import static org.openrewrite.gradle.util.GradleWrapper.WRAPPER_SCRIPT_LOCATION_RELATIVE_PATH;
+import static org.openrewrite.gradle.util.GradleWrapper.*;
 import static org.openrewrite.internal.StringUtils.isBlank;
 
 @RequiredArgsConstructor
@@ -247,19 +224,19 @@ public class UpdateGradleWrapper extends ScanningRecipe<UpdateGradleWrapper.Grad
 
                         GradleWrapper gradleWrpr = getGradleWrapper(ctx);
                         if (StringUtils.isBlank(gradleWrpr.getDistributionUrl()) && !StringUtils.isBlank(version) && Semver.validate(version, null)
-                                                                                                                           .getValue() instanceof ExactVersion) {
+                                .getValue() instanceof ExactVersion) {
                             String newDownloadUrl = currentDistributionUrl.replace("\\", "")
-                                                                          .replaceAll("(.*gradle-)(\\d+\\.\\d+(?:\\.\\d+)?)(.*-(?:bin|all).zip)",
-                                                                                      "$1" + gradleWrapper.getVersion() + "$3");
+                                    .replaceAll("(.*gradle-)(\\d+\\.\\d+(?:\\.\\d+)?)(.*-(?:bin|all).zip)",
+                                            "$1" + gradleWrapper.getVersion() + "$3");
                             gradleWrapper = new GradleWrapper(version, new DistributionInfos(newDownloadUrl, null, null));
                         }
                         String wrapperHost = currentDistributionUrl.substring(0, currentDistributionUrl.lastIndexOf("/")) + "/gradle-";
                         if (StringUtils.isBlank(wrapperUri) && !StringUtils.isBlank(gradleWrpr.getDistributionUrl()) && !gradleWrpr.getPropertiesFormattedUrl().startsWith(wrapperHost)) {
 
                             String newDownloadUrl = gradleWrpr.getDistributionUrl()
-                                                              .replace("\\", "")
-                                                              .replaceAll("(.*gradle-)(\\d+\\.\\d+(?:\\.\\d+)?)(.*-(?:bin|all).zip)",
-                                                                          wrapperHost + gradleWrapper.getVersion() + "$3");
+                                    .replace("\\", "")
+                                    .replaceAll("(.*gradle-)(\\d+\\.\\d+(?:\\.\\d+)?)(.*-(?:bin|all).zip)",
+                                            wrapperHost + gradleWrapper.getVersion() + "$3");
                             gradleWrapper = new GradleWrapper(gradleWrpr.getVersion(), new DistributionInfos(newDownloadUrl, null, null));
                         }
 
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
@@ -22,7 +22,16 @@ import lombok.RequiredArgsConstructor;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
 import org.jspecify.annotations.Nullable;
-import org.openrewrite.*;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.FileAttributes;
+import org.openrewrite.Option;
+import org.openrewrite.PathUtils;
+import org.openrewrite.Preconditions;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.SourceFile;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.Validated;
 import org.openrewrite.gradle.search.FindGradleProject;
 import org.openrewrite.gradle.util.DistributionInfos;
 import org.openrewrite.gradle.util.GradleWrapper;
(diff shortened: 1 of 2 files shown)


Jenson3210:
Strange... this failing test succeeds locally 🤷‍♂️

shanman190:
So the intent of the parameters `${version}` and `${distribution}` for the wrapper URI, was to make reasoning about the recipe a little easier when being used within an enterprise environment that does choose to install distributions locally. These parameters have always been meant to be previously known, exact versions and not allowing dynamic resolution as is the case with the default path against services.gradle.org. It's pretty equal though in just specifying the full wrapperUri as well.

Jenson3210:
I could always alter the validation to allow for fixed versions and distributions + do the replacements in distributionurl if that sounds better for you? 

But at this moment it's using services.gradle for the resolution if mentioned.

I also believe the distributionurl is more important in deciding the impact than the fact if services.gradle is available or not. At our site we have access to services.gradle, but have custom wrappers in local repo. 
It is impossible to get there with current implementation as we receive a version from gradle services repo. If you mention a url, you already know you want specific behaviour.

shanman190:
I don't think resolving the dynamic versions should be in scope right now. The reasons that it's been avoided so far as that each repository hosting tool has a different schema for versions, so it would be difficult to make a fully workable solution that covers all cases including within an environment that completely disallows access to services.gradle.org.

Personally, I don't see much of a different between the wrapperUri being fully defined versus the exact version parameterized replacements because they both are right there in the same spot and must be configured by the user at the time of defining the recipe. Maybe @timtebeek or @sambsnyd have some more opinions, but I've been more inclined to just use the wrapperUri as is with no replacements like you've done here in this PR.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
```diff
@@ -664,7 +710,7 @@ void servicesGradleOrgUnavailable() {
           .setHttpSender(unhelpfulSender)
           .setLargeFileHttpSender(unhelpfulSender);
         rewriteRun(
-          spec -> spec.recipe(new UpdateGradleWrapper("8.6", null, null, "https://artifactory.moderne.ninja/artifactory/gradle-distributions/gradle-${version}-bin.zip"))
```
Laurens-W, line 667: I think we can leave this test in-tact just move the version into the wrapperUri

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
```diff
@@ -111,6 +111,9 @@ public String getDescription() {
     @Override
     public Validated<Object> validate() {
         Validated<Object> validated = super.validate();
+        if (wrapperUri != null && (version != null || distribution != null)) {
+            return Validated.invalid("wrapperUri", wrapperUri, "WrapperUri cannot be used with other parameters");
```
Laurens-W, line 115: Perhaps we can extract the version from the wrapperUri here and validate it's an exact version

Laurens-W:
The only slight concern I see is that with this change we completely disregard the checksum when using the wrapperUri. Then again I agree with @shanman190 that we shouldn't be looking at implementing full support for every hosting tool here.

I think we stand by the comment I left during my work on #4445, if the user specifies a wrapperUri we assume they know it's valid.

shanman190:
I think the checksum part is a non-issue because if the URL is anything except a services.gradle.org, it's most likely that the wrapper is being updated per some organization policy. At that point, the checksum would have been invalid anyway.

As @Jenson3210 was pointing out on the linked issue, in their case services.gradle.org is available, but they were also setting an internal wrapperUri. As a result, the custom wrapperUri wasn't being used.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
shanman190: I'm not sure that would benefit anything. If it's a custom uri, it wouldn't necessarily need to have a version in it at all and it's on the user to ensure that it is a valid distribution URL in the first place.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
Jenson3210: To me this test case matters most if using the version lookup from services.gradle. 

If I provide a wrapperURI the fact if the service is available or not should not matter. 
In this case the service is unavailable, we are using services.gradle version and want to bump that version.

With proposed changes, there is no importance of the availability if wrapperUri is provided.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
Jenson3210: I tend to agree with @shanman190  here. 
We have different distributions (other than bin and all), or at least use that part of the distribution url to set some differentiation between different flavours for our wrappers. 

I think if the users mentions a wrapperUri, he/she should be aware of the correct path of the wrapper and no replacements need to be done anymore.

Jenson3210:
@shanman190 , @Laurens-W I added the checksum property like you mentioned above.

timtebeek:
> Strange... this failing test succeeds locally 🤷‍♂️ Would you have any idea @timtebeek ?

Interesting; I saw the following locally:

1. Run all tests the first time: `UpdateGradleWrapperTest#servicesGradleOrgUnavailable` fails
2. Run `servicesGradleOrgUnavailable` in isolation: test passes
3. Run all tests a second time: all tests pass
4. `rm -rf ~/.rewrite/remote` (via [DEFAULT_ARTIFACT_CACHE](https://github.com/openrewrite/rewrite/blob/6d78b3c8699afbab6c0e2cac959d843e61f95a27/rewrite-core/src/main/java/org/openrewrite/remote/RemoteExecutionContextView.java#L24-L25))
5. goto 1. 

Hope that's enough for you to have another go at fixing the test failure definitively.

Laurens-W:
> > Strange... this failing test succeeds locally 🤷‍♂️ Would you have any idea @timtebeek ?
> 
> Interesting; I saw the following locally:
> 
> 1. Run all tests the first time: `UpdateGradleWrapperTest#servicesGradleOrgUnavailable` fails
> 2. Run `servicesGradleOrgUnavailable` in isolation: test passes
> 3. Run all tests a second time: all tests pass
> 4. `rm -rf ~/.rewrite/remote` (via [DEFAULT_ARTIFACT_CACHE](https://github.com/openrewrite/rewrite/blob/6d78b3c8699afbab6c0e2cac959d843e61f95a27/rewrite-core/src/main/java/org/openrewrite/remote/RemoteExecutionContextView.java#L24-L25))
> 5. goto 1.
> 
> I suspect it's to do with tests using overlapping remote files, and one filling/polluting the cache for another. Hope that's enough for you to have another go at fixing the test failure definitively.

Remote files are stored in `~/.rewrite/remote`, try emptying that and running the tests again :)

Jenson3210:
Adapted the test with the following reasoning: 

The addition of a wrapper downloads the necessary scripts/jar from the downloadURI. If that service is unavailable, it makes sense that the result is empty. 

As that test before was only bumping/updating wrapper version in url, I added the wrapper files and expect that they will not be touched if downloaduri is not available. With this before state, it is better representing the current situation if the bat/sh/jar files are present already.

timtebeek (approved):
Looks good to me! I especially like the `validate()` change to clear out conditionals in `getGradleWrapper()`. Thanks for picking this up! 

Since @shanman190 had some feedback earlier I'll leave a decision to merge up to him.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
```diff
@@ -222,16 +219,26 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
...
                         }
+                        String wrapperHost = currentDistributionUrl.substring(0, currentDistributionUrl.lastIndexOf("/")) + "/gradle-";
+                        if (StringUtils.isBlank(wrapperUri) && !StringUtils.isBlank(gradleWrpr.getDistributionUrl()) &&
+                            !gradleWrpr.getPropertiesFormattedUrl().startsWith(wrapperHost)) {
+                            String newDownloadUrl = gradleWrpr.getDistributionUrl()
+                                    .replace("\\", "")
+                                    .replaceAll("(.*gradle-)(\\d+\\.\\d+(?:\\.\\d+)?)(.*-(?:bin|all).zip)",
+                                            wrapperHost + gradleWrapper.getVersion() + "$3");
+                            gradleWrapper = new GradleWrapper(gradleWrpr.getVersion(), new DistributionInfos(newDownloadUrl, null, null));
+                        }
```
shanman190, line 241: I don't think this section is necessary as the wrapperUri would already be what's used in the `GradleWrapper` object from the `getGradleWrapper()` invocation.

Jenson3210: That's not necessarily true i think. 

Imagine calling the updatewrapper("8.2", null...) with a custom url already present in the properties file. Getwrapper would return the gradle services hosted wrapper (as no custom url is given) with sha and jarsha. We want to early on replace that with the custom distribution url wrapper without shas so that later replacement could be removed in the properties entry visitor (as that was where the patch of url happened before) + would add a sha in the properties visitor (previous bug I discovered during adding tests)

shanman190: Ok, I see. The previous case is we weren't able to resolve while this case is we were. 

Any reason not to just merge these two of blocks together then? So that if the distribution host is null or doesn't match?

There's an additional edge case of going from internal to external hosting that could be another possibility.

Jenson3210: Well... I thought about it, but there's a small difference in the `replaceAll`. Merging would require to extract the `$1` from first if-block through custom code in order to end up with a single `newDownloadUrl` assignment. 
I thought that's not worth it.

Jenson3210: @shanman190 do you want me to try to merge the 2 blocks? Or can we close this thread and merge?

shanman190: I've thought about this as well and it's probably fine to leave it like you've got it presently.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
```diff
@@ -532,18 +528,16 @@ void skipWorkIfUpdatedEarlier() {
         );
     }
 
-    @Issue("https://github.com/openrewrite/rewrite/issues/2651")
```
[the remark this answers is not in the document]
Jenson3210: Do you mean [this test](https://github.com/openrewrite/rewrite/blob/eba054e9726fecf376aa3fb55472d03f011d7a9c/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java#L773-L809)?
It was added as a separate test.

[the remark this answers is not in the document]
shanman190: I see, so this test and that test are kind of doing the same thing.

I'm kinda wondering one of two things:

* For the test case here, it's no longer testing anything different and the test name no longer matches up with the behavior. It's also nice that it was tagged with an issue.
* The other test that you linked, isn't making use of its http sender.

What would you think about dropping your test case and reverting this one back to it's mostly original state with the issue tag? The test originally was trying to make sure that an existing internal distribution URL was retained anyway.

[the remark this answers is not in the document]
shanman190: That's fair. Let's do what you've proposed here.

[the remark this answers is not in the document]
Jenson3210: Quoting on: `The other test that you linked, isn't making use of its http sender.`

Is [this ctx](https://github.com/openrewrite/rewrite/blob/eba054e9726fecf376aa3fb55472d03f011d7a9c/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java#L787) not being used which has the [senders](https://github.com/openrewrite/rewrite/blob/eba054e9726fecf376aa3fb55472d03f011d7a9c/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java#L781-L783) set to mock? 

Or did I misinterprete this remark?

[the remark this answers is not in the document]
shanman190: Nah, you're good. I was forgetting that we actually have to download the distribution to generate the unix and Windows scripts during the edit phase. So you can disregard that part.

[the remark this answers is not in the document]
Jenson3210: applied those in the commit just made!

Jenson3210:
> Looks good to me! I especially like the `validate()` change to clear out conditionals in `getGradleWrapper()`. Thanks for picking this up!
> 
> Since @shanman190 had some feedback earlier I'll leave a decision to merge up to him.

Now that all comments have been resolved. @shanman190, do you agree we can merge?

shanman190 (approved):

shanman190:
@Jenson3210, merged. Thanks for the help on this one!

PR #4492, opened 2024-09-14
Remove redundant versions for Gradle

shanman190:
## What's changed?
Add recipe to remove redundant versions.

## What's your motivation?
A similar feature is available in Maven.

Nothing in particular

Anybody

Forcibly removing versions would result in broken builds.

## Any additional context
* https://github.com/openrewrite/rewrite/discussions/4485

### Checklist


shanman190:
The implementation here is now handling everything except for freestanding scripts which like the other recipes is also spotty for support anyway.

PR #4493, opened 2024-09-14
RemoveRedundantDependencyVersions handles pluginManagement incorrectly

DidierLoiseau:
I noticed several issues with `RemoveRedundantDependencyVersions` and plugins

- version shouldn't be removed from plugin in `pluginManagement` when no parent already defines it – probably a regression from #4156, the current logic finds the plugin in `pluginManagement` and removes the version since it is managed!
- version should be removed from `pluginManagement` when it matches the parent’s `pluginManagement`
- plugin should be removed from `pluginManagement` when nothing else is left (like for `dependencyManagement`) – need to keep it if there is configuration, executions or anything other than `groupId`, `artifactId` and `version`.

## What's changed?
For now these are just failing unit tests to show the problems.

## What's your motivation?
I would like proper management of plugins by `RemoveRedundantDependencyVersions` :)

In #4156, the definition of `isPluginTag()` was changed to match plugins in `pluginManagement` as well as in profiles. As mentioned there, this breaks any recipe that is not handling that properly.
Moreover, it is inconsistent with `isDependencyTag()` which only matches `/project/dependencies/dependency` (no `dependencyManagement` or `profiles`)

Maybe @timtebeek as he is the one who initiated that change of the base method. It seems the change was much broader than #3962 which inspired it.

I’m not sure what’s the best approach here. We certainly need the `isManagedPluginTag()` that was removed by #4156, and an `isPluginTag()` that does not match managed ones, like `isDependencyTag()`.

#4156 may need to be reverted or altered quite a bit.

## Any additional context
There seems to be more inconsistency between methods in `MavenVisitor`. For example, `isDependencyTag()` does not match managed or profile dependencies, but `isDependencyTag(String, String)` can match profile dependencies.

- #3962
- #4156

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -42,9 +42,8 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
     static final XPathMatcher PROFILE_MANAGED_DEPENDENCY_MATCHER = new XPathMatcher("/project/profiles/profile/dependencyManagement/dependencies/dependency");
     static final XPathMatcher PROPERTY_MATCHER = new XPathMatcher("/project/properties/*");
     static final XPathMatcher PLUGIN_MATCHER = new XPathMatcher("//plugins/plugin");
-    static final XPathMatcher MANAGED_PLUGIN_MATCHER = new XPathMatcher("///pluginManagement/plugins/plugin");
+    static final XPathMatcher MANAGED_PLUGIN_MATCHER = new XPathMatcher("//pluginManagement/plugins/plugin");
     static final XPathMatcher PARENT_MATCHER = new XPathMatcher("/project/parent");
-
 
     private transient Xml.@Nullable Document document;
 
@@ -408,10 +407,9 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
     public @Nullable Plugin findManagedPlugin(Xml.Tag tag) {
         for (Plugin resolvedPlugin : getResolutionResult().getPom().getPluginManagement()) {
             String reqGroup = resolvedPlugin.getGroupId();
-            String artifactId = resolvedPlugin.getArtifactId();
             String reqVersion = resolvedPlugin.getVersion();
             if ((reqGroup == null || reqGroup.equals(tag.getChildValue("groupId").orElse(null))) &&
-                (artifactId == null || artifactId.equals(tag.getChildValue("artifactId").orElse(null)) )&&
+                resolvedPlugin.getArtifactId().equals(tag.getChildValue("artifactId").orElse(null)) &&
                 (reqVersion == null || reqVersion.equals(tag.getChildValue("version").orElse(null)))) {
                 return resolvedPlugin;
             }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -399,8 +399,8 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
         return findPlugin(tag, getResolutionResult().getPom().getPluginManagement());
     }
 
-    private static @Nullable Plugin findPlugin(Xml.Tag tag, List<Plugin> pluginManagement) {
-        for (Plugin resolvedPlugin : pluginManagement) {
+    private static @Nullable Plugin findPlugin(Xml.Tag tag, List<Plugin> plugins) {
+        for (Plugin resolvedPlugin : plugins) {
             String reqGroup = resolvedPlugin.getGroupId();
             String reqVersion = resolvedPlugin.getVersion();
             if ((reqGroup == null || reqGroup.equals(tag.getChildValue("groupId").orElse(null))) &&
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -42,6 +42,7 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
     static final XPathMatcher PROFILE_MANAGED_DEPENDENCY_MATCHER = new XPathMatcher("/project/profiles/profile/dependencyManagement/dependencies/dependency");
     static final XPathMatcher PROPERTY_MATCHER = new XPathMatcher("/project/properties/*");
     static final XPathMatcher PLUGIN_MATCHER = new XPathMatcher("//plugins/plugin");
+    static final XPathMatcher MANAGED_PLUGIN_MATCHER = new XPathMatcher("///pluginManagement/plugins/plugin");
     static final XPathMatcher PARENT_MATCHER = new XPathMatcher("/project/parent");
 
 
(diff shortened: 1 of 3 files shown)


timtebeek:
Really appreciate you calling out these issues in the form of runnable tests, and calling out earlier work; surprised these cases weren't tested before, as indeed otherwise we would not have made those earlier changes. I've pushed up a first few steps at remedying these issues; welcome to tag team on the remaining work if you see a clear next step.

DidierLoiseau:
Thanks @timtebeek, I have completed the remaining implementation.

Some notes
* Full disclosure: there was a case issue with the `groupid`/`artifactid` in my original tests. They were still failing after fixing it, but I amended the original commit to keep history clean. Don’t know how it happened as it was copy/paste from other tests.
* I had to duplicate `matchesVersion(ResolvedManagedDependency d, ExecutionContext ctx)` for managed plugins, with two main differences:
  * there is no `requested` on `Plugin` (and there is no `ResolvedManagedPlugin` type either) so I had to check the `<version>` element in `visitTag()`
  * apparently, the `version` of a `Plugin` does not have the variables replaced, so I had to call `parentPom.getValue(versionAccordingToParent)`
* I’m worrying about the `download().resolve()` performance. I think the download goes to the cache, but the whole resolution happens every time, doesn’t it?

Side note: I noticed `onlyIfVersionsMatch` is deprecated, but the (online) documentation does not say so.

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -1170,6 +1229,116 @@ void removeRedundantVersionsFromPluginsManagedByParent() {
...
+                      <version>2.7.18</version>
+                      <relativePath/>
+                  </parent>
+                  <build>
+                      <pluginManagement>
+                          <plugins>
+                              <plugin>
+                                  <groupId>pl.project13.maven</groupId>
+                                  <artifactId>git-commit-id-plugin</artifactId>
+                                  <configuration><whatever /></configuration>
```
timtebeek, line 1281: I had earlier wondered if this is allowed in Maven: to have a pluginManagement plugin without a version. I think I've seen this fail in practice, but it could have been that the circumstances were ever so slightly different.

timtebeek (approved):
Nice and measured improvement here @DidierLoiseau ! Thanks for taking this on. Also appreciate the notes in https://github.com/openrewrite/rewrite/pull/4493#issuecomment-2351854510; I think we're fine to continue as it seeing how we already had that pattern for non-plugins, and it's a (hopefully) seldom hit code path with caching.

I'll add a deprecation note to `onlyIfVersionsMatch` separately, thanks!

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
timtebeek: Let me know your thoughts on this one and then we'll merge right after.

DidierLoiseau: I added a test for non-inherited plugins (`keepPluginIfParentInheritanceDisabled()`)… and I was surprised that it succeeded out-of-the-box.

So I added another test for non-inherited within the same pom (`removeRedundantVersionsFromManagedButNonInheritedPlugins()`) and it fails: apparently [`ResolvedPom.mergePlugins()`][1] completely ignores non-inherited plugin, so the recipe does not see the managed plugin!

It’s a different issue though, so it should maybe be handled separately (might need deeper refactoring of the `mergePlugins()` logic).

Another side note: I noticed that IntelliJ reformats some parts of `RemoveRedundantDependencyVersionsTest` that I didn’t touch, so I don’t know if my setup is correct. I see there is an `.editorconfig` file there, but it does not contain much (and I see these files mostly only exist in test folders? shouldn’t there be one at the root of the repository?)

[1]: https://github.com/openrewrite/rewrite/blob/c39a0836c7073104c465ce2c592533948b8fbd3c/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java#L702-L704

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -388,16 +392,21 @@ public boolean isDependencyLikeTag() {
...
-                    return resolvedPlugin;
-                }
+        return findPlugin(tag, getResolutionResult().getPom().getPlugins());
+    }
+
+    public @Nullable Plugin findManagedPlugin(Xml.Tag tag) {
+        return findPlugin(tag, getResolutionResult().getPom().getPluginManagement());
+    }
+
+    private static @Nullable Plugin findPlugin(Xml.Tag tag, List<Plugin> pluginManagement) {
```
DidierLoiseau, line 402: Should maybe rename the `pluginManagement` parameter here.

rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
```diff
@@ -346,4 +387,16 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
             return t;
         }
     }
+
+    private static class RemoveEmptyPluginsTags extends MavenIsoVisitor<ExecutionContext> {
+        @Override
+        public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
+            Xml.Tag t = super.visitTag(tag, ctx);
+            if (("pluginManagement".equals(t.getName()) || "plugins".equals(t.getName())) && (t.getContent() == null || t.getContent().isEmpty())) {
```
DidierLoiseau, line 395: It seems `t.getContent().isEmpty()` returns false if there is just a comment left, do we want to keep it then? (but I guess it’s the same for `<dependencyManagement>`)

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
timtebeek: Thanks! Changed just now.

rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
timtebeek: Comments are typically attached to the next element, such that when the `<plugin>` is removed, so is any comment attached to it. As such I don't think we have to handle the case of a dangling comment here, although I could be wrong. Thanks for pointing it out though! I think it's rare enough that we don't need to handle this just yet.

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
timtebeek: Ah I had for now worked around this, but perhaps there's a better fix indeed in that mergePlugins; would love to hear as I'm switching between a few too many PRs in a day to fully dive in. :)

DidierLoiseau: I don’t know the codebase well enough, but I suppose a distinction could be made between plugin management and inherited plugin management. In this particular case we want to see the full plugin management for the current pom, but only the inherited one from the parent pom.

The same might be needed for non-managed plugins too…

OTOH maybe it’s good enough like this? It could possibly be more important to document what `getPlugins()`/`getPluginManagement()` return, as they aren’t very clear about inheritance.

timtebeek:
Thanks for all the help here @DidierLoiseau ! Great to be more predictable in what we change when, and these tests & fixes help a good deal towards that.

PR #4494, opened 2024-09-15
`SemanticallyEqual` should allow for different lambda parameter names

knutwannheden:
While comparing two lambdas for semantical equality, the lambda parameter names should be ignored. This actually also applies to other variables (e.g. variables declared in a block such as a lambda body), but that is out of scope for this PR.

This will allow Refaster recipes to match a lambda against other lambdas using different parameter names.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/search/SemanticallyEqualTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/search/SemanticallyEqualTest.java
@@ -299,6 +299,28 @@ class SemanticallyEqualTest {
         );
     }
 
+    @Test
+    void lambdaParameterNames() {
+        assertExpressionsEqual(
+          """
+            import java.util.Comparator;
+            class T {
+                Comparator<Integer> a = (x1, y1) -> x1 - y1;
+                Comparator<Integer> b = (x2, y2) -> x2 - y2;
+            }
+            """
+        );
+        assertExpressionsNotEqual(
+          """
+            import java.util.Comparator;
+            class T {
+                Comparator<Integer> a = (x1, y1) -> x1 - y1;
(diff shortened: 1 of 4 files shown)


knutwannheden:
@timtebeek I was looking into supporting lambdas in Refaster recipes. While we still can't support templates directly returning lambdas, we can support templates "using" lambdas. So while the following can't be supported, because `JavaTemplate` doesn't allow us to provide the type of the lambda:
```java
        @BeforeTemplate
        public Comparator<Integer> before(List<Integer> l) {
            return (x, y) -> x - y;
        }

        @AfterTemplate
        public Comparator<Integer> after() {
            return Comparator.comparingInt(x -> x);
        }
```
something like this we could support:
```java
        @BeforeTemplate
        public void before(List<Integer> l) {
            l.sort((x, y) -> x - y);
        }

        @AfterTemplate
        public void after(List<Integer> l) {
            l.sort(Comparator.comparingInt(x -> x));
        }
```
In this particular example that would of course be very unfortunate, as replacing the `(x, y) -> x - y` lambda (with type `Comparator<Integer>`) with `Comparator.comparingInt(x -> x)` would always make sense, regardless of where it is used. But I think there are still enough other use cases, when supporting lambdas when part of a method invocation would already be a very useful addition.

This PR then only makes sure that we can then use `JavaTemplate.match()` to match a lambda which uses different parameter names. In the example above we also want to match `l.sort((a, b) -> a - b);` for example.

It would then be nice if we at some point can also extend this to also support templates which directly return a lambda.

knutwannheden:
See also https://github.com/openrewrite/rewrite-templating/pull/111.

timtebeek (approved):
Looks good! Thanks for taking the time to expand what we can cover with Refaster.

PR #4498, opened 2024-09-17
Add test for relocating a dependency in an subproject

sihutch:
## What's changed?
Add test for relocating a dependency in an subproject. This test passes however when I do the same in an actual project (see [here](https://github.com/sihutch/rewrite-subproject-test)), the dependency in the sub-module is not updated. Strangely if you copy the dependency section from the sub-module to the root and run the recipe then the dependency is updated in the root and the sub-module.

- https://github.com/openrewrite/rewrite-gradle-plugin/issues/334

## What's your motivation?
To be able to relocate dependencies defined in sub-modules.

@timtebeek @shanman190

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
@@ -21,6 +21,7 @@ import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.gradle.Assertions.buildGradle;
+import static org.openrewrite.gradle.Assertions.settingsGradle;
 import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
 
 class ChangeDependencyTest implements RewriteTest {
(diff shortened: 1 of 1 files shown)


shanman190:
@sihutch , I think the root issue has now been fixed and this can be closed. Would you agree?

sihutch:
@shanman190, yeah, I will close. Regarding the fix to the Gradle plugin, would it be possible to release this soon? It affects numerous recipes (i.e., all that use the Trait dependency matcher and the Gradle plugin) we plan to use for an upcoming migration.

sihutch:
Closing as fixed by https://github.com/openrewrite/rewrite-gradle-plugin/issues/334

timtebeek:
> Regarding the fix to the Gradle plugin, would it be possible to release this soon? It affects numerous recipes (i.e., all that use the Trait dependency matcher and the Gradle plugin) we plan to use for an upcoming migration.

Thanks for helping clear out open issues & PRs! Makes catching up on the road just that bit easier. We plan to do a release this Wednesday; folding in all improvements over the past two weeks as per our release schedule. Would that be soon enough for you?

sihutch:
@timtebeek Perfect, thanks. And thanks for your help @shanman190.

PR #4500, opened 2024-09-17
Use a `HEAD` request to check if a maven server is reachable

pstreef:
## What's changed?
Use `HEAD` in stead of `GET` request to determine if a maven server exists

## What's your motivation?
If a maven server does not block directory listing the listing itself might be very slow causing read timeouts.

The removed check looks redundant to me

@timtebeek 

Do nothing

## Any additional context
no

### Checklist
~- [ ] I've added unit tests to cover both positive and negative cases~ (already exist)

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -750,7 +750,7 @@ public class MavenPomDownloader {
                     httpsUri += "/";
                 }
 
-                HttpSender.Request.Builder request = applyAuthenticationToRequest(repository, httpSender.get(httpsUri));
+                HttpSender.Request.Builder request = applyAuthenticationToRequest(repository, httpSender.head(httpsUri));
                 MavenRepository normalized = null;
                 try {
                     sendRequest(request.build());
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -792,8 +792,7 @@ Collection<MavenRepository> distinctNormalizedRepositories(
                             }
                         }
                     }
-                    if (normalized == null && !(t instanceof HttpSenderResponseException &&
-                                                ((HttpSenderResponseException) t).getBody().contains("Directory listing forbidden"))) {
+                    if (normalized == null) {
```
timtebeek, line 795: We implement `repositoryAccessFailed` in a couple places, including the CLI, to warn about a failure to access a repository. Not being allowed to list directories seems like a valid setting that should not trigger that logged statement; can we determine those cases without a body?

pstreef: As far as I can tell normalized is never going to be `null` in the case that the body would contain `Directory listing forbidden` which returns with a 403 and means `e.isServerReached()` returns `true`

```
if (t instanceof HttpSenderResponseException) {
    HttpSenderResponseException e = (HttpSenderResponseException) t;
    // response was returned from the server, but it was not a 200 OK. The server therefore exists.
    if (e.isServerReached()) {
        normalized = repository.withUri(httpsUri);
    }
}
```

timtebeek (approved):

Issue #4194, opened 2024-05-16
Gradle wrapper upgrade recipe to 7.x+ is not updating "compile" dependencies to "implementation" in build.gradle files


aakash-pinc:
## How are you running OpenRewrite?
moderne.io

## Problem statement
Gradle wrapper upgrade recipe to 7.x is not updating the "compile" to "implementation" string in build.gradle files. 

## What is the smallest, simplest way to reproduce the problem?
Create a Gradle (4.x) project with dependencies and then run the "Gradle wrapper upgrade" recipe to 7.x.  You will notice the "compile" in build.gradle is not getting replaced with "implementation"

## Are you interested in [contributing a fix to OpenRewrite]
No

timtebeek:
Thanks for logging the issue @aakash-pinc ; I'll link this doc page here for some more context:
https://docs.gradle.org/8.7/userguide/java_library_plugin.html#sec:java_library_separation

> The compile and runtime configurations have been removed with Gradle 7.0. Please refer to the [upgrade guide](https://docs.gradle.org/8.7/userguide/upgrading_version_6.html#sec:configuration_removal) how to migrate to implementation and api configurations`.

shanman190:
So the base recipe for this is available in the [ChangeDependency](https://docs.openrewrite.org/recipes/gradle/changedependencyconfiguration) recipe. Unfortunately, it's not as easy as to migrate from `compile` to `implementation`. This works only for Gradle projects that fit the category of applications and falls apart for libraries. For libraries, it's more nuanced in that it needs to migrate to either `api` or `implementation`. 

While migrating from `compile` to `implementation` would result in a successful build for the library code, it's dependency descriptor would state that all of these dependencies should be in the `runtime` scope for projects that then consume them, thus often resulting in compilation errors in those external projects.

Note: There are also other configurations that would need to be migrated as part of this migration as well.

PR #4502, opened 2024-09-19
Add declarative Gradle upgrade recipes

timtebeek:
## What's your motivation?
When upgrading Java versions folks also need to upgrade older Gradle versions. While upgrading the wrapper is one part of that upgrade, at times there's required changes as well, in particular going from 4.x to 5.0. Some of these changes we can already automate, others might still follow now that the framework for these migrations is there.

- For the `RemoveEnableFeaturePreview` record I went by when those keys were dropped from the enum, going by latest.patch; seemed ok, but could be double checked as documentation was hard to find.
- Intentionally left out `compile` -> `api/implementation`, although that'll be a bigger pain point. Anything we can do for reasonable coverage here? Perhaps look if `java-library` plugin is defined and base our behavior on that? Or at worst merely post a comment on the first case?

The components here were already available. We could document the steps and considerations for `compile` -> `api/implementation`, and how they could be achieved with recipes.

## Any additional context
- https://github.com/openrewrite/rewrite/issues/4194

### DIFF
--- a/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-6.yml
+++ b/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-6.yml
@@ -15,10 +15,11 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/recipe
-name: org.openrewrite.gradle.UpgradeGradle6
-displayName: Upgrade Gradle to 6.x
-description: Upgrade to version 6.x. See the Gradle upgrade guide from [version 5.x to 6.0](https://docs.gradle.org/current/userguide/upgrading_version_5.html) for more information.
+name: org.openrewrite.gradle.MigrateToGradle6
+displayName: Migrate to Gradle 6 from Gradle 5
+description: Migrate to version 6.x. See the Gradle upgrade guide from [version 5.x to 6.0](https://docs.gradle.org/current/userguide/upgrading_version_5.html) for more information.
 recipeList:
+  - org.openrewrite.gradle.MigrateToGradle5
   - org.openrewrite.gradle.UpdateGradleWrapper:
       version: 6.x
       addIfMissing: false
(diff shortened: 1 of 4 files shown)


shanman190:
General question, but are these more migration recipes and as such should follow the [migration ADR](https://github.com/openrewrite/rewrite/blob/main/doc%2Fadr%2F0004-library-migration-conventions.md)?

timtebeek:
Good call! Wasn't aware of that doc, but helps to standardize where possible.

Any other steps that come to mind? Things like required common plugin upgrades, or removal of jcenter?

Of course we can expand the steps later as well; just let me know when you'd consider this ready for a first merge.

shanman190:
So the `jcenter()` method is still available as of Gradle 8.10.1, but with that said JFrog has redirected https://jcenter.bintray.com/ to https://repo1.maven.org/maven2/. So in practice it's now the same as `mavenCentral()`. So I could go either way there.

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-6.yml
```diff
@@ -0,0 +1,51 @@
...
+  - org.openrewrite.gradle.ChangeDependencyConfiguration:
+      groupId: "*"
+      artifactId: "*"
+      configuration: testCompile
+      newConfiguration: testImplementation
+  - org.openrewrite.gradle.ChangeDependencyConfiguration:
+      groupId: "*"
+      artifactId: "*"
+      configuration: runtime
+      newConfiguration: runtimeOnly
```
shanman190, line 37: supernit: In practice this doesn't change the meaning, but generally when I'm thinking about these sort of changes, I'm thinking in terms of `main` and then `test`.

```suggestion
  - org.openrewrite.gradle.ChangeDependencyConfiguration:
      groupId: "*"
      artifactId: "*"
      configuration: runtime
      newConfiguration: runtimeOnly
  - org.openrewrite.gradle.ChangeDependencyConfiguration:
      groupId: "*"
      artifactId: "*"
      configuration: testCompile
      newConfiguration: testImplementation
```

shanman190 (approved):

PR #4504, opened 2024-09-19
Handle bom import in RemoveDuplicateDependencies

DidierLoiseau:
## What's changed?
Added support for `<scope>import</scope>` in `RemoveDuplicateDependencies`

## What's your motivation?
It seems we have duplicate imports in our poms 😏

I don’t really see any.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveDuplicateDependencies.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveDuplicateDependencies.java
@@ -125,6 +125,9 @@ public class RemoveDuplicateDependencies extends Recipe {
             }
 
             private @Nullable DependencyKey getManagedDependencyKey(Xml.Tag tag) {
+                if (tag.getChildValue("scope").map("import"::equalsIgnoreCase).orElse(false)) {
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/RemoveDuplicateDependencies.java
```diff
@@ -125,6 +125,9 @@ private boolean isManagedDependenciesTag() {
             }
 
             private @Nullable DependencyKey getManagedDependencyKey(Xml.Tag tag) {
+                if (tag.getChildValue("scope").map("import"::equalsIgnoreCase).orElse(false)) {
+                    return DependencyKey.from(tag);
+                }
                 ResolvedManagedDependency resolvedDependency = findManagedDependency(tag);
```
DidierLoiseau, line 131: `findManagedDependency()` does not actually find imported boms. It seems they are not included in dependency management by the resolver.

timtebeek (approved):
Thanks a lot for yet another improvement @DidierLoiseau ; Must say I really appreciate your thorough work! :)

Issue #4003, opened 2024-02-14
Declarative Preconditions - `or` association


nmck257:
## What problem are you trying to solve?
Declarative preconditions currently associate via `AND` if you declare multiple on the same recipe.
If you want to achieve an `OR` relationship, you can do it, but through a much more verbose "diamond" pattern.

Sample:
```yml
---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething
displayName: Do Something
recipeList:
  - org.sample.DoSomething.condition.1
  - org.sample.DoSomething.condition.2
  - org.sample.DoSomething.condition.3

---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething.condition.1
displayName: Do Something (condition 1)
preconditions:
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/my.json"
recipeList:
  - org.sample.DoSomething.for.real

---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething.condition.2
displayName: Do Something (condition 2)
preconditions:
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/your.json"
recipeList:
  - org.sample.DoSomething.for.real  

---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething.condition.3
displayName: Do Something (condition 3)
preconditions:
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/our.json"
recipeList:
  - org.sample.DoSomething.for.real

---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething.for.real
displayName: Do Something (for real)
recipeList:
  - org.openrewrite.json.ChangeKey:
      qwe: qwe

```
(if you imagine graphing how those 5 recipes relate to each other, it kinda looks like a diamond)

## Describe the solution you'd like

```yml
---
type: specs.openrewrite.org/v1beta/recipe
name: org.sample.DoSomething
displayName: Do Something
preconditionAssociation: OR
preconditions:
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/my.json"
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/your.json"
  - org.openrewrite.FindSourceFiles:
      filePattern: "**/our.json"
recipeList:
  - org.openrewrite.json.ChangeKey:
      qwe: qwe

```

## Have you considered any alternatives or workarounds?
[...]

PR #4505, opened 2024-09-20
Creating "OR" preconditions instead of "AND" (#4003) does not actually work

DidierLoiseau:
## What's changed?
Adds 2 failing tests using declarative recipes as preconditions.

Based on [Creating "OR" preconditions instead of "AND"](https://docs.openrewrite.org/reference/yaml-format-reference#creating-or-preconditions-instead-of-and), you should be able to use a declarative recipe as a precondition to combine recipes into an “or” clause, but it does not actually work.

Upon debugging, it appears that `DeclarativeRecipe` does not actually override `getVisitor()`, so when [`DeclarativeRecipe.getRecipeList()` calls `getVisitor()`](https://github.com/openrewrite/rewrite/blob/956fa7a820f1a520499165d5a2d1126033d12791/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java#L273) on the declarative recipe used as precondition, it just returns `noop()`.

## What's your motivation?
Be able to re-use a configured recipe as precondition for other recipes.

In #4003 this use-case was documented and later reported as not working.

#4003 involved a discussion mostly between @timtebeek and @nmck257 who might be interested.

Currently the only alternative is to repeat the predicate configuration, using only imperative recipes. If you need a disjunction of conditions, you will have to duplicate the whole recipe, but it may not always work.

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/config/DeclarativeRecipeTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/config/DeclarativeRecipeTest.java
@@ -175,6 +175,62 @@ class DeclarativeRecipeTest implements RewriteTest {
     }
 
     @Test
+    void yamlDeclarativeRecipeAsPrecondition() {
+        rewriteRun(
+          spec -> spec.recipeFromYaml("""
+            ---
+            type: specs.openrewrite.org/v1beta/recipe
+            name: org.openrewrite.PreconditionTest
+            description: Test.
+            preconditions:
+              - org.openrewrite.DeclarativePrecondition
+            recipeList:
+              - org.openrewrite.text.ChangeText:
+                 toText: 2
+              - org.openrewrite.text.ChangeText:
+                 toText: 3
+            ---
+            type: specs.openrewrite.org/v1beta/recipe
+            name: org.openrewrite.DeclarativePrecondition
+            recipeList:
+              - org.openrewrite.text.Find:
+                  find: 1
+            """, "org.openrewrite.PreconditionTest"),
+          text("1", "3"),
+          text("2")
+        );
+    }
+
+    @Test
+    void orPreconditions() {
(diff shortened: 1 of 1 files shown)


timtebeek:
Very much appreciated! Had this in the back of my mind still, but didn't quite get to writing the tests and giving them a try. Indeed this ought to work as described, as that's a nice simplification where possible. Did you already explore what might be the cause? I wonder if the `getRecipeList()` is evaluated at all, or whether it's only looking at `getVisitor()`... 🤔

DidierLoiseau:
> Did you already explore what might be the cause? I wonder if the `getRecipeList()` is evaluated at all, or whether it's only looking at `getVisitor()`... 🤔

I did some debugging which I described in the "what changed" section. I don't know what would be an appropriate solution though. Did it ever work?

timtebeek:
Thanks a lot @DidierLoiseau ; I've pushed up a change that fixes the tests you've provided. 🙏🏻 
Could you let me know if that aligns with what you were thinking?

DidierLoiseau:
Amazing! Thanks for your reactivity as well!

I should be able to try it out next week. However, as I am using the Maven plugin from the command-line, and it seems I can’t override the version of the rewrite modules used by the plugin without adding it to the pom (I’m using it with `-Drewrite.recipeArtifactCoordinates` and my own library, which depends on the newer version). Is there a way to do that? Alternatively: would you mind releasing the plugin as well? 🙏

timtebeek:
I've just kicked off a build (not yet release) of the rewrite-maven-plugin: That should produce a [snapshot version](https://docs.openrewrite.org/reference/snapshot-instructions) that uses the latest from openrewrite/rewrite. Ideally that's enough to try this out before the scheduled release next Wednesday. 🤞🏻

PR #4506, opened 2024-09-20
Use `OPTIONS` request before falling back to `HEAD` request when trying to connect to repositories during normalization.

pstreef:
## What's changed?

To test if a repository is reachable we currently do a HEAD request. Unfortunately a HEAD request is just as slow as a GET on artifactory. `OPTIONS` is much faster, but it does not take authentication into account and is not always supported (thinking about reverse-proxies)

With this PR we now try an OPTIONS request first, if that returns 200 we accept that the server is reachable. If an error is returned we first try a HEAD request to see if that returns a 200 (in case of maven central OPTIONS gives an error, while HEAD returns 200)
If that is OK we set `knownToExist` to true.

We still first try https before (maybe) trying the original url.

## What's your motivation?
Some artifactory repositories are VERY slow at directory listing. 

`knownToExist` is now true in most cases, even if the authentication would fail. This might not work as expected, but I'm not entirely sure what that is used for. Technically a 200 response from OPTIONS would mean it exists.

@sambsnyd @timtebeek 

Increasing the timeout (which seems to not yet be used in this part of the code) but that leaves this to be very slow.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -779,21 +779,20 @@ public class MavenPomDownloader {
 
         ReachabilityResult reachability = reachable(repository, applyAuthenticationToRequest(repository, request));
         if (reachability.isSuccess()) {
-            return repository.withUri(httpsUri).withKnownToExist(true);
+            return repository.withUri(httpsUri);
         }
         reachability = reachable(repository, applyAuthenticationToRequest(repository, request.withMethod(HttpSender.Method.HEAD).url(httpsUri)));
         if (reachability.isReachable()) {
-            return repository.withUri(httpsUri).withKnownToExist(reachability.isSuccess());
+            return repository.withUri(httpsUri);
         }
-
         if (!originalUrl.equals(httpsUri)) {
             reachability = reachable(repository, applyAuthenticationToRequest(repository, request.withMethod(HttpSender.Method.OPTIONS).url(originalUrl)));
             if (reachability.isSuccess()) {
-                return MavenRepository.from(repository, originalUrl).withKnownToExist(true);
+                return MavenRepository.from(repository, originalUrl);
             }
             reachability = reachable(repository, applyAuthenticationToRequest(repository, request.withMethod(HttpSender.Method.HEAD).url(originalUrl)));
             if (reachability.isReachable()) {
-                return MavenRepository.from(repository, originalUrl).withKnownToExist(reachability.isSuccess());
+                return MavenRepository.from(repository, originalUrl);
             }
         }
         // Won't be null if server is unreachable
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java
```diff
@@ -133,6 +133,17 @@ public MavenRepository(
...
+        return new MavenRepository(
+                repository.getId(),
+                url,
+                repository.getReleases(),
+                repository.getSnapshots(),
+                repository.getUsername(),
+                repository.getPassword(),
+                repository.getTimeout());
+    }
+
```
sambsnyd, line 146: Use the existing `MavenRepository.withUri()` instead of creating this method.

pstreef: will do. I added this because there are 2 nonFinal fields which are then copied. but since we dont care about `knownToExist` it's only `deriveMetadataIfMissing` and I dont think that one matters either?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -808,6 +758,90 @@ Collection<MavenRepository> distinctNormalizedRepositories(
...
+            httpsUri += "/";
+        }
+
+        HttpSender.Request.Builder request = httpSender.options(httpsUri);
+        if (repository.getTimeout() != null) {
+            request = request.withConnectTimeout(repository.getTimeout())
+                    .withReadTimeout(repository.getTimeout());
+        }
+
+        ReachabilityResult reachability = reachable(repository, applyAuthenticationToRequest(repository, request));
```
timtebeek, line 780: I thought OPTIONS requests should not include any credentials, but it seems these are added here; should we remove those just to be sure?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -808,6 +758,90 @@ Collection<MavenRepository> distinctNormalizedRepositories(
...
+    enum Reachability {
+        SUCCESS,
+        ERROR,
+        UNREACHABLE;
+    }
+
+    private ReachabilityResult reachable(MavenRepository repository, HttpSender.Request.Builder requestBuilder) {
+        HttpSender.Request.Builder request = applyAuthenticationToRequest(repository, requestBuilder);
+        try {
+            sendRequest(request.build());
```
timtebeek, line 830: Calls into `reachable` also seem to call `applyAuthenticationToRequest` themselves before passing in their `requestBuilder`. Should we leave that up to the callers given the OPTIONS without auth headers above?
```suggestion
        try {
            sendRequest(requestBuilder.build());
```

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
pstreef: I dont think sending auth is a problem, but options should indeed not require it. However, you never know with custom proxy setups so I think it's OK to just leave it?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
pstreef: good point. Updated

timtebeek (approved):

PR #4507, opened 2024-09-20
Handle plugins without groupId

DidierLoiseau:
## What's changed?
Handle the default value of plugin `groupId` as per https://maven.apache.org/xsd/maven-4.0.0.xsd

- Found after #4493 was merged and released 🙄

See also [is groupId required for plugins in Maven pom.xml?](https://stackoverflow.com/questions/65527291/is-groupid-required-for-plugins-in-maven-pom-xml) on Stack Overflow.

## What's your motivation?
Someone declared a plugin without groupId…

It would be possible to fix just `RemoveRedundantDependencyVersions`, but since this default value is defined in the xsd, I thought it would be better to fix the parsing/resolution.

## Any additional context
It was causing an invisible NPE in `RemoveRedundantDependencyVersions` for the `pluginManagement` case. I’m not sure how such exceptions should be handled.

Theoretically the `artifactId` is optional as well, but I don’t see why anyone wouldn’t provide it. Moreover there isn’t a default for it.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
@@ -40,6 +40,7 @@ import java.util.Objects;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
+import static org.openrewrite.maven.tree.Plugin.PLUGIN_DEFAULT_GROUPID;
 
 /**
  * A value object deserialized directly from POM XML
@@ -249,6 +250,7 @@ public class RawPom {
     @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
     @Data
     public static class Plugin {
+        @Nullable
         String groupId;
         String artifactId;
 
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
```diff
@@ -510,9 +512,9 @@ private List<org.openrewrite.maven.tree.Plugin> mapPlugins(@Nullable List<Plugin
         if (rawPlugins != null) {
             plugins = new ArrayList<>(rawPlugins.size());
             for (Plugin rawPlugin : rawPlugins) {
-
+                String pluginGroupId = rawPlugin.getGroupId();
                 plugins.add(new org.openrewrite.maven.tree.Plugin(
-                        rawPlugin.getGroupId(),
+                        pluginGroupId == null ? PLUGIN_DEFAULT_GROUPID : pluginGroupId,
```
DidierLoiseau, line 517: I don’t see a way to set it anywhere else because the field is final `final` in the raw plugin, and the constructors are generated by Lombok.

timtebeek (approved):
Thanks for the quick fix!

timtebeek:
Thanks for the many fixes! I've just put out a new patch release:
- https://github.com/openrewrite/rewrite/releases/tag/v8.35.4

PR #4509, opened 2024-09-21
Drop `OperatorWrap` as there's a copy maintained in rewrite-static-analysis

timtebeek:
## What's changed?
Drop `OperatorWrap` here, after we split off similar recipes to rewrite-static-analysis. 
## What's your motivation?
Only maintain a single copy, after adding an option:
- https://github.com/openrewrite/rewrite-static-analysis/pull/341

The copy there is looped into `org.openrewrite.staticanalysis.CommonStaticAnalysis`, which has likely been copied and modified since, so is harder to remove at this point.


knutwannheden (approved):
Should we make it more convenient to supply style config when running recipes? Just so that we don't en up sprinkling convenience style options all over the place.

timtebeek:
> Should we make it more convenient to supply style config when running recipes? Just so that we don't en up sprinkling convenience style options all over the place.

That might be a good idea, as I'm indeed not a fan of sprinkling options, even if the alternative is cumbersome. Not quite sure what making it easier would like though.

I did notice we also have a copy of this recipe in rewrite-static-analysis, so I'll move this change there for now, and remove the recipe here, as there's no apparent references elsewhere at this point in time.
https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/OperatorWrap.java

Issue #4325, opened 2024-07-12
Capture `MavenMetadata.lastUpdated` for a recipe that shows outdated dependencies


timtebeek:
## What problem are you trying to solve?
Apart from the most recent version, we might also be interested in capturing when that dependency was last updated.

This then allows for a recipe that tells out where you might be using dependencies that have not seen a release in a long time,
perhaps because they have been relocated: https://docs.openrewrite.org/recipes/java/dependencies/relocateddependencycheck

## Describe the solution you'd like
Add a `lastUpdated` field to `MavenMetadata.Versioning`
https://github.com/openrewrite/rewrite/blob/56986647f0c8b8a162d548d52c70572b939f1d0a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenMetadata.java#L55-L63

Based on the metadata pulblished in Maven repositories.
```xml
<metadata>
  <groupId>org.openrewrite.recipe</groupId>
  <artifactId>rewrite-recipe-bom</artifactId>
  <versioning>
    <latest>2.15.0-SNAPSHOT</latest>
    <versions>
      <version>0.1.0-SNAPSHOT</version>
      <!-- truncated -->
      <version>2.15.0-SNAPSHOT</version>
    </versions>
    <lastUpdated>20240712162001</lastUpdated>
  </versioning>
</metadata>
```
https://oss.sonatype.org/service/local/repositories/snapshots/content/org/openrewrite/recipe/rewrite-recipe-bom/maven-metadata.xml
https://repo1.maven.org/maven2/org/openrewrite/recipe/rewrite-recipe-bom/maven-metadata.xml

## Additional context
Inspired by https://github.com/Giovds/outdated-maven-plugin

PR #4511, opened 2024-09-21
Parse `MavenMetadata.Versioning.lastUpdated` as `ZonedDateTime`

timtebeek:
## What's changed?
Added `MavenMetadata.Versioning.lastUpdated` and parse that using `@JsonFormat` and `JavaTimeModule`.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/4325

Unlocks a new type of analysis, looking at when dependencies were last maintained.

Ok to add that `JavaTimeModule`?
Ok to drop that `MavenMetadata.xmlParser`?

Could have kept the time as String, but then that would have just passed the burden onto consumers.

## Any additional context
Inspired by https://github.com/Giovds/outdated-maven-plugin

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenMetadata.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenMetadata.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.maven.tree;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import lombok.AccessLevel;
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenMetadata.java
```diff
@@ -62,10 +55,15 @@ public static class Versioning {
         @Nullable
         Snapshot snapshot;
 
+        @Nullable
+        ZonedDateTime lastUpdated;
```
timtebeek, line 59: Adding a nullable field should be fine right? Since I also do not see MavenMetadata persisted anywhere.

knutwannheden (approved):

Issue #4508, opened 2024-09-12
rewrite:run (warns about) failing to download SNAPSHOT parent from local repository


DidierLoiseau:
## What version of OpenRewrite are you using?
I am using

- Maven plugin v5.39.2
- rewrite-migrate-java v2.23.0 (shouldn’t be relevant?)

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project with a parent in SHAPSHOT version.

## What is the smallest, simplest way to reproduce the problem?
**Parent pom**
Just install it in your local repository with `mvn install`.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.example</groupId>
    <artifactId>parent-pom</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>pom</packaging>

</project>
```
**Child pom**
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>com.example</groupId>
        <artifactId>parent-pom</artifactId>
        <version>1.0-SNAPSHOT</version>
    </parent>

    <artifactId>child-pom</artifactId>

    <build>
        <plugins>
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>5.39.2</version>
                <configuration>
                    <activeRecipes>
                        <recipe>org.openrewrite.java.migrate.UpgradeBuildToJava17</recipe>
                    </activeRecipes>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-migrate-java</artifactId>
                        <version>2.23.0</version>
                    </dependency>
                </dependencies>
            </plugin>
[...]

PR #4512, opened 2024-09-21
Use `maven-metadata-local.xml` for `file://` paths

timtebeek:
## What's changed?
Use `maven-metadata-local.xml` when accessing `file://` URIs, as documented on
- https://maven.apache.org/ref/3.9.9/maven-repository-metadata/

## What's your motivation?
Improve odds of picking up an existing local file, as opposed to downloading metadata remotely.

Presumably the local metadata might be outdated; should we guard against that somehow?

@DidierLoiseau

Locally I also see `maven-metadata-central.xml` as well as a few other repo-ids; Should we account for those?

## Any additional context
- https://github.com/openrewrite/rewrite/issues/4508

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -256,15 +256,14 @@ public class MavenPomDownloader {
                 boolean cacheEmptyResult = false;
                 try {
                     String scheme = URI.create(repo.getUri()).getScheme();
-                    String uri = repo.getUri() + (repo.getUri().endsWith("/") ? "" : "/") +
+                    String baseUri = repo.getUri() + (repo.getUri().endsWith("/") ? "" : "/") +
                                  requireNonNull(gav.getGroupId()).replace('.', '/') + '/' +
                                  gav.getArtifactId() + '/' +
-                                 (gav.getVersion() == null ? "" : gav.getVersion() + '/') +
-                                 "maven-metadata.xml";
+                                 (gav.getVersion() == null ? "" : gav.getVersion() + '/');
 
                     if ("file".equals(scheme)) {
                         // A maven repository can be expressed as a URI with a file scheme
-                        Path path = Paths.get(URI.create(uri));
+                        Path path = Paths.get(URI.create(baseUri + "maven-metadata-local.xml"));
                         if (Files.exists(path)) {
                             MavenMetadata parsed = MavenMetadata.parse(Files.readAllBytes(path));
                             if (parsed != null) {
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -256,23 +256,22 @@ public MavenMetadata downloadMetadata(GroupArtifactVersion gav, @Nullable Resolv
                 boolean cacheEmptyResult = false;
                 try {
                     String scheme = URI.create(repo.getUri()).getScheme();
-                    String uri = repo.getUri() + (repo.getUri().endsWith("/") ? "" : "/") +
+                    String baseUri = repo.getUri() + (repo.getUri().endsWith("/") ? "" : "/") +
                                  requireNonNull(gav.getGroupId()).replace('.', '/') + '/' +
                                  gav.getArtifactId() + '/' +
```
DidierLoiseau, line 261: This bit is duplicated at lines 259-261 and 538-540. I wonder if the URI building logic (and possibly more) wouldn’t be better suited in the `MavenRepository` class itself.

DidierLoiseau:
> Presumably the local metadata might be outdated; should we guard against that somehow?

I didn’t check the whole thing, but the metadata is per-repository. The local repository keeps track of some remote repository metadata which might be outdated indeed, but I think the code doesn’t use it since it always fetches the metadata from the remote repository. Since it downloads the metadata from all repositories and then merges it, I guess it’s fine? (assuming the merge is correct)

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -256,23 +256,22 @@ public MavenMetadata downloadMetadata(GroupArtifactVersion gav, @Nullable Resolv
...
-                                 (gav.getVersion() == null ? "" : gav.getVersion() + '/') +
-                                 "maven-metadata.xml";
+                                 (gav.getVersion() == null ? "" : gav.getVersion() + '/');
 
                     if ("file".equals(scheme)) {
                         // A maven repository can be expressed as a URI with a file scheme
-                        Path path = Paths.get(URI.create(uri));
+                        Path path = Paths.get(URI.create(baseUri + "maven-metadata-local.xml"));
```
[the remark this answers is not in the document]
timtebeek: Yes my main goal here was to have a better chance of matching an existing file, while not yet perfect. I'd say this is better than falling through to deriving that information needlessly further down.

timtebeek:
Thanks for helping think through the options here! Indeed we merge the metadata as also seen on this PR
- https://github.com/openrewrite/rewrite/pull/4511

With that I think we should be safe to merge this as an improvement for a commonly expected case.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
DidierLoiseau: I wanted to mean that it was duplicated 3 times. The third time is actually at lines 356-358.

Issue #4514, opened 2024-09-23
ChangeDependencyGroupIdAndArtifactId can duplicate new dependency


ashakirin:
In case if new dependency already exists, it will be doubled:
```java
@Test
void shouldNotAddNewIfDependencyAlreadyExistsManaged() {
    rewriteRun(
      spec -> spec.recipe(new ChangeDependencyGroupIdAndArtifactId(
        "javax.activation",
        "javax.activation-api",
        "jakarta.activation",
        "jakarta.activation-api",
        null,
        null
      )),
      pomXml(
        """
          <project>
              <modelVersion>4.0.0</modelVersion>
              <groupId>com.mycompany.app</groupId>
              <artifactId>my-app</artifactId>
              <version>1</version>
              <dependencies>
                  <dependency>
                      <groupId>javax.activation</groupId>
                      <artifactId>javax.activation-api</artifactId>
                  </dependency>
                  <dependency>
                      <groupId>jakarta.activation</groupId>
                      <artifactId>jakarta.activation-api</artifactId>
                  </dependency>
              </dependencies>
              <dependencyManagement>
                  <dependencies>
                      <dependency>
                          <groupId>javax.activation</groupId>
                          <artifactId>javax.activation-api</artifactId>
                          <version>1.2.0</version>
                      </dependency>
                      <dependency>
                          <groupId>jakarta.activation</groupId>
                          <artifactId>jakarta.activation-api</artifactId>
                          <version>1.2.1</version>
                      </dependency>
                  </dependencies>
              </dependencyManagement>
          </project>
          """,
        """
          <project>
              <modelVersion>4.0.0</modelVersion>
              <groupId>com.mycompany.app</groupId>
              <artifactId>my-app</artifactId>
              <version>1</version>
              <dependencies>
                  <dependency>
                      <groupId>jakarta.activation</groupId>
                      <artifactId>jakarta.activation-api</artifactId>
                  </dependency>
              </dependencies>
              <dependencyManagement>
                  <dependencies>
                      <dependency>
                          <groupId>javax.activation</groupId>
                          <artifactId>javax.activation-api</artifactId>
                          <version>1.2.0</version>
                      </dependency>
                      <dependency>
                          <groupId>jakarta.activation</groupId>
                          <artifactId>jakarta.activation-api</artifactId>
                          <version>1.2.1</version>
                      </dependency>
                  </dependencies>
              </dependencyManagement>
          </project>
          """
      )
    );
}
```
[...]

PR #4515, opened 2024-09-23
feat: added check for duplicated dependency

ashakirin:
- Refs: #4514


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -98,6 +98,316 @@ void changeDependencyGroupIdAndArtifactId() {
...
+                      <dependency>
+                          <groupId>jakarta.activation</groupId>
+                          <artifactId>jakarta.activation-api</artifactId>
+                          <version>1.2.2</version>
+                      </dependency>
+                      <dependency>
+                          <groupId>jakarta.activation</groupId>
+                          <artifactId>jakarta.activation-api</artifactId>
+                          <version>1.2.1</version>
+                      </dependency>
```
timtebeek, line 202: I wonder if in this case we should remove the old one instead. What are your thoughts on that?
```suggestion
```

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -98,6 +98,316 @@ void changeDependencyGroupIdAndArtifactId() {
...
+                          <artifactId>jakarta.activation-api</artifactId>
+                      </dependency>
+                  </dependencies>
+                  <dependencyManagement>
+                      <dependencies>
+                          <dependency>
+                              <groupId>javax.activation</groupId>
+                              <artifactId>javax.activation-api</artifactId>
+                              <version>1.2.0</version>
+                          </dependency>
```
timtebeek, line 325: Same here; I think ideally we remove the duplicate, since two managed versions of the same dependency don't make much sense (to me).
```suggestion
```

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
ashakirin: @timtebeek: in theory - yes, on practice: perhaps user has a good reason to keep old dependency in pom, we do not know that. I do not think that it is responsibility of the CahngeDependencyGroupIdAndArtifact to cleanup old unused dependencies. Perhaps separate recipe would be more clean solution for this case.

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
ashakirin: Do you think it should be responsibility of ChangeDependency recipe? I think it can be a bit confusing for user, if existing dependencies disappearing

sambsnyd:
Thanks for the contribution! I agree with Tim, duplicates and old versions should not be left behind. Going to modify the recipe to that effect myself

Issue #4513, opened 2024-09-21
OpenRewrite fails to parse java files when using Java 23


talios:
When setting `<release>23</release>` on my Maven project, OpenRewrite fails to parse Javadoc, most likely due to the updates to support Markdown Javadoc.

## What version of OpenRewrite are you using?

I am using the following versions:

```
locked org.openrewrite.maven:rewrite-maven-plugin:5.40.2;
locked org.openrewrite.recipe:rewrite-cucumber-jvm:1.5.7;
locked org.openrewrite.recipe:rewrite-java-dependencies:1.18.1;
locked org.openrewrite.recipe:rewrite-java-security:2.12.0;
locked org.openrewrite.recipe:rewrite-logging-frameworks:2.13.0;
locked org.openrewrite.recipe:rewrite-migrate-java:2.25.1;
locked org.openrewrite.recipe:rewrite-static-analysis:1.16.0;
locked org.openrewrite.recipe:rewrite-testing-frameworks:2.18.0;
locked org.openrewrite:rewrite-java:7.40.8;
```

## How are you running OpenRewrite?

```
mvn clean install org.openrewrite.maven:rewrite-maven-plugin:runNoFork
```

## What did you see instead?

```
[WARNING] There were problems parsing Foo.java
[WARNING] java.lang.NoSuchMethodError: 'com.sun.tools.javac.tree.DCTree$DCDocComment com.sun.tools.javac.tree.DocCommentTable.getCommentTree(com.sun.tools.javac.tree.JCTree)'
  org.openrewrite.java.isolated.ReloadableJava21ParserVisitor.convert(ReloadableJava21ParserVisitor.java:1666)
  org.openrewrite.java.isolated.ReloadableJava21ParserVisitor.visitCompilationUnit(ReloadableJava21ParserVisitor.java:537)
  org.openrewrite.java.isolated.ReloadableJava21ParserVisitor.visitCompilationUnit(ReloadableJava21ParserVisitor.java:73)
  com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:625)
  com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:92)
```

timtebeek:
Thanks for the report @talios ! Good to track this here. I'd already explored this briefly after getting a report through
- https://github.com/PicnicSupermarket/error-prone-support/pull/1329

In short we're having to depend on an internal API, which seems to have changed.
[...]

PR #4516, opened 2024-09-23
Accommodate Java 23 in ReloadableJava21ParserVisitor

timtebeek:
## What's changed?

`com.sun.tools.javac.tree.DocCommentTable#getCommentTree` used to return `com.sun.tools.javac.tree.DCTree.DCDocComment` in Java 21.
As of Java 23 it returns `com.sun.source.doctree.DocCommentTree`, an interface that was already present and implemented by `DCDocComment`.

https://github.com/openrewrite/rewrite/blob/a98d83d698dda9971c93a0767b7f85d5866f306d/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java#L1666

This PR changes the usage downstream in `formatWithCommentTree` to use the interface instead.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/4513

A separate Java 23 parser; decided against for now as we still have work to do on our Java 21 parser, and will likely only target LST releases going forward. That said if support is easy to add, as seen here, we will try to do so.

## Any additional context
- https://github.com/openrewrite/rewrite/issues/4513#issuecomment-2367966854

### DIFF
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
@@ -16,6 +16,7 @@
 package org.openrewrite.java.isolated;
 
 
+import com.sun.source.doctree.DocCommentTree;
 import com.sun.source.tree.*;
 import com.sun.source.util.TreePathScanner;
 import com.sun.tools.javac.code.Flags;
@@ -2136,7 +2137,7 @@ public class ReloadableJava21ParserVisitor extends TreePathScanner<J, Space> {
         return annotations;
     }
 
-    Space formatWithCommentTree(String prefix, JCTree tree, DCTree.@Nullable DCDocComment commentTree) {
+    Space formatWithCommentTree(String prefix, JCTree tree, @Nullable DocCommentTree commentTree) {
         Space fmt = format(prefix);
         if (commentTree != null) {
             List<Comment> comments = fmt.getComments();


Stephan202:
Hey @timtebeek! Have you considered backporting this fix to `rewrite-java-17`, as alluded to [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1329#issuecomment-2350934389)? :pray:

timtebeek:
> Hey @timtebeek! Have you considered backporting this fix to `rewrite-java-17`, as alluded to [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1329#issuecomment-2350934389)? 🙏

I didn't think that'd be necessary if rewrite-java-21 is available, but it looks like [EPS only uses rewrite-java-17](https://github.com/PicnicSupermarket/error-prone-support/blob/21437a43aa2e082d19652c0537f959df3350ef5a/error-prone-contrib/pom.xml#L209): would it be possible to change that in EPS only? Or would you still prefer the backport to 17?

timtebeek:
Pending your response I've already pushed up a PR here in case this is indeed needed; tell me it is and we'll merge:
- https://github.com/openrewrite/rewrite/pull/4516#issuecomment-2440037849

Stephan202:
> > Hey @timtebeek! Have you considered backporting this fix to `rewrite-java-17`, as alluded to [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1329#issuecomment-2350934389)? 🙏
> 
> I didn't think that'd be necessary if rewrite-java-21 is available, but it looks like [EPS only uses rewrite-java-17](https://github.com/PicnicSupermarket/error-prone-support/blob/21437a43aa2e082d19652c0537f959df3350ef5a/error-prone-contrib/pom.xml#L209): would it be possible to change that in EPS only? Or would you still prefer the backport to 17?

Fair question. When I upgrade EPS to `rewrite-java-21` and then build with JDK 17.0.13 (which we currently still support), the build fails for two reasons:
1. A complaint by the Maven Enforcer plugin (can perhaps be worked around, would have to check):
    ```
    [WARNING] Rule 6: org.codehaus.mojo.extraenforcer.dependencies.EnforceBytecodeVersion failed with message:
    Found Banned Dependency: org.openrewrite:rewrite-java-21:jar:8.38.0
    Use 'mvn dependency:tree' to locate the source of the banned dependencies.
    ```
2. A test failure in `StringRulesRecipesTest.stringValueOf:28 ExceptionInInitializer`:
    ```
    Caused by: java.lang.IllegalStateException: Unable to create a Java parser instance. `rewrite-java-8`, `rewrite-java-11`, `rewrite-java-17`, or `rewrite-java-21` must be on the classpath.
    	at org.openrewrite.java.JavaParser.fromJavaVersion(JavaParser.java:250)
    	at org.openrewrite.java.Assertions.<clinit>(Assertions.java:104)
	... 66 more
    ```

(That `IllegalStateException` message is a bit misleading :sweat_smile:.)

I could look into having our build require JDK 21, while still targeting and testing against JDK 17 using toolchain support, but a backport would be easier :sweat_smile:.

PR #4518, opened 2024-09-23
Parse .NET .vbproj and .fsproj files as XML

bryceatmoderne:
## What's changed?
`.vbproj` and `.fsproj` are .NET project XML files and need to be parsed as such.

## What's your motivation?
Prior to these changes these file types were being parsed as `PlainText`.

No.

@timtebeek 

No.

## Any additional context
No.

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
@@ -111,7 +111,7 @@ public class XmlParser implements Parser {
     public boolean accept(Path path) {
         String p = path.toString();
         int dot = p.lastIndexOf('.');
-        if (dot > 0 && dot < (p.length() - 1)) {
+        if (0 < dot && dot < (p.length() - 1)) {
             if (ACCEPTED_FILE_EXTENSIONS.contains(p.substring(dot + 1))) {
                 return true;
             }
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

rewrite-xml/src/main/java/org/openrewrite/xml/XmlParser.java
```diff
@@ -32,9 +32,43 @@
...
+            "tbd",
+            "tfm",
+            "dqs",
+            "stp",
+            "dcn",
+            "pst",
+            // .NET project files
+            "csproj",
+            "vbproj",
+            "fsproj"));
```
knutwannheden, line 70: I wonder if we also should add `"targets"` and `"props"` from the .NET ecosystem here or not. Thoughts?

timtebeek: I think those were intentionally kept as plaintext, given any ambiguity in those file extensions being used elsewhere:
https://github.com/moderneinc/moderne-cli/pull/2117/files

knutwannheden: I see. So allowing these mappings to be more dynamic (e.g. parse `.props` files as XML in .NET projects) seems like it will at some point become necessary.

timtebeek:
Already merging this now such that it makes it into today's release. Thanks both for your input!

Issue #4521, opened 2024-09-24
Gradle ChangeDependency behaves incorrectly, if new dependency exists


ashakirin:
Gradle ChangeDependency shows following result, if new dependency was already defined:
```
        rewriteRun(
          spec -> spec.recipe(new ChangeDependency("commons-lang", "commons-lang", "org.apache.commons", "commons-lang3", "3.11.x", null, null)),
          buildGradle(
            """
              plugins {
                  id "java-library"
              }
              
              repositories {
                  mavenCentral()
              }
              
              dependencies {
                 implementation "org.apache.commons:commons-lang3:2.6"
                 implementation group: "org.apache.commons", name: "commons-lang3", version: "2.6"
                 implementation "org.apache.commons:commons-lang3:3.11"
              }
              """,
            """
              plugins {
                  id "java-library"
              }
              
              repositories {
                  mavenCentral()
              }
              
              dependencies {
                  implementation "commons-lang:commons-lang:2.6"
                  implementation group: "org.apache.commons", name: "commons-lang3", version: "2.6"
                  implementation "org.apache.commons:commons-lang3:3.11"
              }
              """
```
Recipe replaced commons-lang:commons-lang with new groupId and artifactId, but kept the version from old dependency.
Expected behavior will be just remove old dependency in this case:
```
                 implementation "org.apache.commons:commons-lang3:3.11"
```

timtebeek:
Sounds like you're uncovering quite some edge cases over there! Appreciate you reporting these here and already tackling the Maven side of things in
- https://github.com/openrewrite/rewrite/pull/4515

timtebeek:
Fixed by commit 3a848abe3 (December 10, 2025) - "Avoid duplicate dependencies following one or more ChangeDependency recipes (#6406)". Added `maybeRemoveDuplicateTargetDependency()` method that detects when the new dependency already exists and removes the old one instead of creating duplicates. Test: `removeIfExists()`.

PR #4523, opened 2024-09-25
feat: Added functionality to Gradle ChangeDependency to avoid duplica…

ashakirin:
Refs: #4521

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -18,6 +18,7 @@ package org.openrewrite.gradle;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
@@ -36,9 +37,11 @@ import org.openrewrite.java.tree.J;
 import org.openrewrite.maven.MavenDownloadingException;
 import org.openrewrite.maven.tree.GroupArtifact;
 import org.openrewrite.maven.tree.GroupArtifactVersion;
+import org.openrewrite.maven.tree.ResolvedDependency;
 import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
 import org.openrewrite.semver.DependencyMatcher;
 import org.openrewrite.semver.Semver;
+import org.openrewrite.semver.VersionComparator;
 
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -198,13 +219,17 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+            @NullMarked
             private J.MethodInvocation updateDependency(J.MethodInvocation m, ExecutionContext ctx) {
                 List<Expression> depArgs = m.getArguments();
                 if (depArgs.get(0) instanceof J.Literal) {
                     String gav = (String) ((J.Literal) depArgs.get(0)).getValue();
                     if (gav != null) {
                         Dependency original = DependencyStringNotationConverter.parse(gav);
-                        if (original != null && depMatcher.matches(original.getGroupId(), original.getArtifactId())) {
+                        if (original != null && oldDepMatcher.matches(original.getGroupId(), original.getArtifactId())) {
+                            if (isNewDependencyPresent) {
```
shanman190, line 230: This either needs to be evaluated when we have found the dependency OR precomputed by configurations that contain it as a first order dependency.

With the current implementation, if the new dependency was in `testImplementation`, but the old is in `implementation`, we would incorrectly remove the necessary dependency.

The converse side of this, where the new dependency has now been added to a configuration upstream of the current one, I think right now we should leave as out of scope and for a future cleanup duplicate dependencies recipe instead.

ashakirin: @shanman190: do you generally interpret this as consistent case to have dependency in testImplementation and the same dependency with older version in implementation?

shanman190: Gradle configurations are a foundational component of Gradle. I was merely using `implementation` and `testImplementation` as examples. The more specific issue is when the old dependency is in a configuration upstream from the new dependency. (Ex: `api` -> `implementation`; `implementation` -> `testImplementation`).

The present implementation doesn't account for the Gradle configuration that the existing dependency is in or if the new dependency is strictly upstream from itself and as a result, illustrated by my example, shows how the current implementation would become broken and can lead directly to code that no longer compiles in these situations. A recipe that would result in code that fails to compile would violate OpenRewrite's do no harm principle. 

Just to be clear, what you're wanting to achieve is possible, the decisioning just needs to happen where I'm indicating and taking into account the current configuration (gotten by looking up the configuration using `MethodInvocation#getSimpleName`), then to be able to remove more often (ie. The new dependency is not in the same configuration as the old) looking at the configurations upstream to see if the new dependency is present.

timtebeek:
This likely got some conflicts as a result of https://github.com/openrewrite/rewrite/commit/9936a7a70a4792aa73e7f2d1f1378c05b3507123 ; is there anything left to fix after that change? Or perhaps a different direction you'd now recommend @shanman190 ?

shanman190:
@timtebeek, this PR still needs some rework based upon my last comment here:
https://github.com/openrewrite/rewrite/pull/4523#discussion_r1779500203

This also applies probably to more of the recipes than just `ChangeDependency`, but that one is a good starting place.

PR #4529, opened 2024-09-26
Make ParentPomInsight recursive

DidierLoiseau:
## What's changed?
Made `ParentPomInsight` recursive, with a parameter to disable recursivity.

## What's your motivation?
We needed to check the parents more deeply before performing some cleanup. @sambsnyd suggested updating this recipe instead of creating a new one. @timtebeek suggested to make it the default.

This could also be used to conditionally perform some upgrades, e.g. only perform Spring Boot 3 migration if we are not already in Spring Boot 3 (see also openrewrite/rewrite-spring#361) – although you would also need to check the imported poms in this case.

Note how the change affects `matchNonSnapshot()` which must explicitly disable recursivity to keep the same behavior (otherwise all child modules would match as well).

I also took the opportunity to document `Semver.validate()`, although the documentation is mostly copied from existing recipes and may need some clarification.

The recipe does not support a `versionPattern` – I’m not sure how to handle it when `version` isn’t provided, which could be useful for example if someone wanted to check if any ancestry is a snapshot or release candidate.

The current `ParentPomInsight` only allows checking the direct parent, so you would have to list all possible values instead.

### Checklist
- [x] I've used the IntelliJ IDEA auto-formatter on affected files – yes, except I reverted the change of indentation the `version` field multi-line description.


sambsnyd:
This sounds a lot like the existing [ParentPomInsight ](https://github.com/openrewrite/rewrite/blob/a98d83d698dda9971c93a0767b7f85d5866f306d/rewrite-maven/src/main/java/org/openrewrite/maven/search/ParentPomInsight.java#L33)recipe. Can that cover this use-case?

DidierLoiseau:
@sambsnyd this one is recursive, but maybe I could make `ParentPomInsight` optionally recursive as well instead?

(also using `SearchResult.found` seems simpler than what I did here)

timtebeek:
Still catch up from last week; was great to meet briefly at Devoxx! Indeed seems best to make the existing ParentPomInsight recursive as well; judging by [how that's used](https://github.com/search?q=org%3Aopenrewrite%20ParentPomInsight&type=code) we should be fine to just change that by default, or at least until someone comes along with a convincing case to be selective.

Feel free to use this same pull request; we'll squash merge anyway, and I imagine it's mostly moving over your test case and introducing a while loop. Thanks again!

DidierLoiseau:
@timtebeek & @sambsnyd, I have updated the PR and its description accordingly. I made it recursive by default as suggested.

rewrite-maven/src/main/java/org/openrewrite/maven/search/ParentPomInsight.java
```diff
@@ -79,26 +91,42 @@ public Validated<Object> validate() {
...
+                            return t;
+                        }
+                        try {
+                            ResolvedPom ancestorPom = mpd.download(ancestor.getGav(), null, null, mrr.getPom().getRepositories())
+                                    .resolve(emptyList(), mpd, ctx);
+                            ancestor = ancestorPom.getRequested().getParent();
+                            relativePath = null;
+                        } catch (MavenDownloadingException e) {
+                            return e.warn(t);
+                        }
```
timtebeek, line 129: I wonder if we should do anything here to break the while loop, just to ensure we don't get stuck when there's persisted failures to download a particular parent; for instance the following, but open to other suggestions.
```suggestion
                        } catch (MavenDownloadingException e) {
                            ancestor = null;
                            return e.warn(t);
                        }
```

timtebeek (approved):
Approved already, but I before we merge I think we should break that while loop when there are downloading errors ; would you agree there?

Great to see how thorough your are with your tests when adapting an existing recipe!

rewrite-maven/src/main/java/org/openrewrite/maven/search/ParentPomInsight.java
DidierLoiseau: It’s a return statement already :sweat_smile: 

Note that I’m not familiar with this marker mechanism, I just noticed that’s how exceptions were handled in other places.

rewrite-maven/src/main/java/org/openrewrite/maven/search/ParentPomInsight.java
```diff
@@ -52,6 +58,12 @@ public class ParentPomInsight extends Recipe {
     @Nullable
     String version;
 
+    @Option(displayName = "Recursive",
+            description = "Whether to search recursively through the parents. True by default.",
+            required = false)
+    @Nullable
+    Boolean recursive;
+
```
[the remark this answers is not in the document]
DidierLoiseau: Yes, since this is breaking change I though it does not cost much to still support it. Maybe some people would search for direct usage of a specific parent?

rewrite-maven/src/main/java/org/openrewrite/maven/search/ParentPomInsight.java
timtebeek: You're absolutely right. I blame the fever I got from hiking through the rain. 😅

PR #4532, opened 2024-09-28
Fix all XPathMatcherTest TODO's + handle nested elements with the same name

DidierLoiseau:
## What's changed?
* Fixed all existing TODO’s in `XPathMatcherTest
* Handle XML documents that contain nested elements with the same name for XPath using `//`
* Code cleanup

## What's your motivation?
Needed to use such XPaths.

I will put comments on the bits of code I cleaned up.

There are still limitations with XPath that do not start with `/` or that start with `//`, in particular, they don’t support `//` in the middle and there were not tests for that.

I think the implementation would really still benefit from a lot of refactoring, or even a full rewrite:
* There is no validation of the provided XPath
* The XPath parsing is mixed with the XPath maching (e.g. extracting conditions, determining whether we are facing an attribute or a function etc.)
* The implementation of XPaths starting with `/` is completely dissociated from those starting with `//` or no `/`, which causes different limitations for the two.

I think it would make sense to represent the XPath as a structured object instead of a `String[]`…

## Any additional context
As a user, it’s not really clear what is supported and what is not without looking at the unit test…

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
@@ -92,18 +92,8 @@ public class XPathMatcher {
                     if (index < 0) {
                         return false;
                     }
-                    if (part.startsWith("@")) { // is attribute selector
-                        partWithCondition = part;
-                        tagForCondition = i > 0 ? path.get(i - 1) : path.get(i);
-                    } else { // is element selector
-                        if (part.charAt(index + 1) == '@') { // is Attribute condition
-                            partWithCondition = part;
-                            tagForCondition = path.get(i);
-                        } else if (part.contains("(") && part.contains(")")) { // is function condition
-                            partWithCondition = part;
-                            tagForCondition = path.get(i);
-                        }
-                    }
+                    partWithCondition = part;
+                    tagForCondition = path.get(pathIndex);
                 } else if (i < path.size() && i > 0 && parts[i - 1].endsWith("]")) {
                     String partBefore = parts[i - 1];
                     int index = partBefore.indexOf("[");
@@ -117,6 +107,7 @@ public class XPathMatcher {
                     }
                 } else if (part.endsWith(")")) { // is xpath method
                     // TODO: implement other xpath methods
+                    throw new UnsupportedOperationException("XPath methods are not supported");
                 }
 
                 String partName;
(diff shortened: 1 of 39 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
```diff
@@ -92,18 +92,8 @@ public boolean matches(Cursor cursor) {
                     if (index < 0) {
                         return false;
                     }
-                    if (part.startsWith("@")) { // is attribute selector
-                        partWithCondition = part;
-                        tagForCondition = i > 0 ? path.get(i - 1) : path.get(i);
```
DidierLoiseau, line 97: Using `i` straight up in `path.get()` was completely wrong since `path` is reverted while `parts` is not – `pathIndex` should always be used.

Note that if a part starts with `@`, it MUST be the last part – I wanted to check it here, but then I realized all `if`s actually result in the same code (and conditions matching sub-elements were not handled…).

rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
```diff
@@ -92,18 +92,8 @@ public boolean matches(Cursor cursor) {
...
-                            partWithCondition = part;
-                            tagForCondition = path.get(i);
-                        } else if (part.contains("(") && part.contains(")")) { // is function condition
-                            partWithCondition = part;
-                            tagForCondition = path.get(i);
-                        }
-                    }
+                    partWithCondition = part;
+                    tagForCondition = path.get(pathIndex);
                 } else if (i < path.size() && i > 0 && parts[i - 1].endsWith("]")) {
```
DidierLoiseau, line 97: I don’t understand what’s the purpose of this `if`. All tests still pass if I remove it but I’m not sure whether I’m missing an edge case or it’s really unneeded.

rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
```diff
@@ -164,36 +157,42 @@ public boolean matches(Cursor cursor) {
...
             }
 
-            return startsWithSlash || path.size() - pathIndex <= 1;
+            // we have matched the whole XPath, and it does not start with the root
+            return true;
         } else {
             Collections.reverse(path);
 
             // Deal with the two forward slashes in the expression; works, but I'm not proud of it.
-            if (expression.contains("//") && !expression.contains("://") && Arrays.stream(parts).anyMatch(StringUtils::isBlank)) {
```
DidierLoiseau, line 172: Removing `://` does not seem to break anything, but fixes XPaths with both `//` and a namespace URL condition.

I don’t think the first condition is still relevant either, however the `isBlank` check is inconsistent with the `indexOf("")` on the next line. I think it should use `isEmpty()` like I did to compute `tagMatchingParts`.

Note that the first condition is insufficient on its own because of URLs again – or actually any value that could contain double slashes in a condition.

sambsnyd:
Great stuff thanks @DidierLoiseau !

PR #4533, opened 2024-09-28
Expose AddOrUpdateChild as a recipe using XPath + child XML

DidierLoiseau:
## What's changed?
Simple XML recipe that can add or update a child below the element(s) selected by XPath.

## What's your motivation?
Provide more flexible ways of updating XML/pom files. In our specific case we need to configure `openapi-generator-maven-plugin` with `<useJakartaEe>true</useJakartaEe>` when it matches `<generatorName>spring</generatorName>`, following `JavaxMigrationToJakarta` (once this PR is merged, maybe it could be included there).

I had considered offering the option to add the child even when it already exists, but that wouldn’t be idempotent. I implemented an option to not replace the child in such a case, disabled by default.

I based this branch on top of #4532 so the former should be merged first and it’s not needed to review the changes on XPath implementation. I just wanted to use the XPath fixes in the tests of this new recipe.

I don’t think any other equivalent recipe exists. `ChangePluginConfiguration` only allows to override the whole configuration.

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/AddOrUpdateChildTag.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/AddOrUpdateChildTag.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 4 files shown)


sambsnyd:
Thanks @DidierLoiseau !

DidierLoiseau:
Thanks for merging @sambsnyd 

> I based this branch on top of #4532 so the former should be merged first

Sorry it seems I wrote _former_ instead of _latter_ there, so when you merged #4532 afterwards it created an empty commit, and c307b31 includes the changes of both :confused:

PR #4547, opened 2024-10-03
Nullability cleanups in `TreeVisitor`

knutwannheden:
Adjust some of the `@Nullable` and `@NonNull` annotations in `TreeVisitor` to reduce the number of warnings in code using the API as well as in `TreeVisitor` itself.


sambsnyd:
Thanks @knutwannheden !

PR #4548, opened 2024-10-03
add test case to illustrate issue with unsolicited indentation.

timo-a:
## What's changed?

This PR adds a test to illustrate a bug.

## What's your motivation?

I want to enforce a custom order of imports on new PRs through a custom style.
As I report in "observation 2" of https://github.com/openrewrite/rewrite/issues/4165 custom styles also change indentation. I don't think they should do that.

## Any additional context

I've marked this PR ready for review in he hope of triggering a build, but it seems a manual approval is necessary.
I won't be available to work on a fix.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
@@ -19,9 +19,12 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.Issue;
 import org.openrewrite.java.style.ImportLayoutStyle;
+import org.openrewrite.java.style.TabsAndIndentsStyle;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
+
+import java.util.Arrays;
 
 import static java.util.Collections.emptySet;
 import static java.util.Collections.singletonList;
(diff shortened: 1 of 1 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/OrderImportsTest.java
```diff
@@ -709,4 +709,52 @@ public class Foo {
...
+        rewriteRun(
+          spec -> spec.parser(JavaParser.fromJavaVersion().styles(
+            singletonList(
+              new NamedStyles(
+                randomId(),
+                "it.does.more.than.JustOrderImports",
+                "custom style",
+                null,
+                emptySet(),
+                singletonList(
```
Laurens-W, line 729: Hey @timo-a, I think in your case you also want to influence the tabs and indents like we do [here](https://github.com/openrewrite/rewrite/blob/v8.8.4/rewrite-java/src/main/java/org/openrewrite/java/style/IntelliJ.java#L42) in our IntelliJ style.
Specifically the 4th parameter of the `TabsAndIndentsStyle` [here](https://github.com/openrewrite/rewrite/blob/v8.8.4/rewrite-java/src/main/java/org/openrewrite/java/style/IntelliJ.java#L71) influences the indent spacing

knutwannheden:
Adding something like this to the `TabsAndIndentsVisitor` should fix this specific case, although the same should then probably also be done for `JContainer` and `JRightPadded`:

```java
    @Override
    public <T> JLeftPadded<T> visitLeftPadded(@Nullable JLeftPadded<T> left, JLeftPadded.Location loc, P p) {
        if (getCursor().getNearestMessage("stop") != null) {
            return left;
        }
        return super.visitLeftPadded(left, loc, p);
    }
```

PR #4554, opened 2024-10-06
Remove maven properties should remove comment

jonesbusy:
I saw in many project maven properties with an attached comment.

When running some recipes to remove properties the comment is left behind

I don't know how to fix it but I've implemented a test that demonstrates the issue

Example when cleanup property override (that are defined on a parent pom)

I would expect the TODO comment is also gone

Screenshot from 2024-10-06 07-43-46

Some help to implement it

Not really

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/RemoveContentVisitor.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/RemoveContentVisitor.java
@@ -27,11 +27,6 @@ public class RemoveContentVisitor<P> extends XmlVisitor<P> {
     private final boolean removeEmptyAncestors;
     private final boolean removePrecedingComment;
 
-    @Deprecated
-    public RemoveContentVisitor(Content tag, boolean removeEmptyAncestors) {
-        this(tag, removeEmptyAncestors, false);
-    }
-
     public RemoveContentVisitor(Content tag, boolean removeEmptyAncestors, boolean removePrecedingComment) {
         this.scope = tag;
         this.removeEmptyAncestors = removeEmptyAncestors;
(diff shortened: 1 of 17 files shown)


timtebeek:
Thanks for the report! Makes sense to also want to remove any preceding comments; I've made it optional in `RemoveContentVisitor` same as we did with removing empty ancestors. I'm on the fence a bit if we should make this the new default or not across the other usages. That might surprise folks, although in practice I think it's rare that this would lead to issues. Any insights there?

image

rewrite-xml/src/main/java/org/openrewrite/xml/RemoveContentVisitor.java
```diff
@@ -25,10 +25,17 @@ public class RemoveContentVisitor<P> extends XmlVisitor<P> {
 
     private final Content scope;
     private final boolean removeEmptyAncestors;
+    private final boolean removePrecedingComment;
 
+    @Deprecated
     public RemoveContentVisitor(Content tag, boolean removeEmptyAncestors) {
+        this(tag, removeEmptyAncestors, false);
```
timtebeek, line 32: I'm leaning towards making the new behavior `true` by default, then inlining and removing the old constructor such that folks consciously have to choose to adopt the new default or not.
```suggestion
        this(tag, removeEmptyAncestors, true);
```

jonesbusy:
My opinion it that is should be the default. I don't see any reason to keep existing comment when removing such tag.

I see that we filter `Xml.Comment` to the risk is close to zero

knutwannheden:
Since we in the end produce a diff, the developer can still decide to restore the comment, in the case that it also pertains to any subsequent elements. I would probably remove them by default.

In OpenRewrite we don't have any concept of trailing comments, but I think it could be interesting to build a trait or something similar which would allow recipes that remove elements to also remove the "associated" comments, which would typically be any preceding multi-line comment and trailing single-line comments.

timtebeek (approved):
Thanks for kicking this off and the quick feedback (both) on changing the default. Applied this everywhere it made sense, except where we were replacing with a similar element.

PR #4561, opened 2024-10-08
adding gradle and groovy auto detection

lkerford:
Updating the autodetect class to pick up styles from the gradle files

Previously no style would be detected for groovy files, this is because we weren't parsing the block element within lambdas in method declarations


rewrite-gradle/src/test/java/org/openrewrite/gradle/style/AutodetectTest.java
```diff
@@ -0,0 +1,59 @@
...
+              
+              dependencies {
+                implementation 'io.freefair.gradle:lombok-plugin:8.4'
+                implementation 'com.github.spotbugs.snom:spotbugs-gradle-plugin:5.0.13'
+              }
+              """
+          );
+        var detector = Autodetect.detector();
+        parse.forEach(detector::sample);
+        var styles = detector.build();
```
sambsnyd, line 52: Is the type of `styles` `org.openrewrite.java.style.Autodetect` or `org.openrewrite.gradle.style.Autodetect` ?  🙂

lkerford: you are 100% correct, I have pushed an update which will fix this issue

sambsnyd (approved):

rewrite-java/src/main/java/org/openrewrite/java/style/Autodetect.java
```diff
@@ -65,8 +65,7 @@ public static class Detector {
         private final FindLineFormatJavaVisitor findLineFormat = new FindLineFormatJavaVisitor();
 
         public void sample(SourceFile cu) {
-            // only sample Java sources (extending languages need their own auto-detection)
-            if (cu instanceof J.CompilationUnit) {
+            if (cu instanceof JavaSourceFile) {
```
knutwannheden, line 68: @sambsnyd It seems to me that we shouldn't have to change this now. What am I missing?

PR #4562, opened 2024-10-09
Add withModifier to base java visitors

macsux:
## What's changed?
Allows `Modifier` LST to be visited

## What's your motivation?
Consistency and resolve issues with remoting to other langugages

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
@@ -449,8 +449,6 @@ public class JavaVisitor<P> extends TreeVisitor<J, P> {
             c = (J.ClassDeclaration) temp;
         }
         c = c.withLeadingAnnotations(ListUtils.map(c.getLeadingAnnotations(), a -> visitAndCast(a, p)));
-        c = c.withModifiers(ListUtils.map(c.getModifiers(),
-                mod -> mod.withPrefix(visitSpace(mod.getPrefix(), Space.Location.MODIFIER_PREFIX, p))));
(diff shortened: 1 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -920,6 +916,14 @@ public J visitMethodInvocation(J.MethodInvocation method, P p) {
         return m;
     }
 
+    public J visitModifier(J.Modifier modifer, P p) {
```
knutwannheden, line 919: I haven't seen the corresponding override of acceptJava() on J.Modifier. is that file missing here?

rewrite-java/src/main/java/org/openrewrite/java/UpdateSourcePositions.java
```diff
@@ -66,14 +66,15 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
             }
 
             @Override
-            protected void visitModifier(J.Modifier modifier, PrintOutputCapture<ExecutionContext> p) {
+            public J visitModifier(J.Modifier modifier, PrintOutputCapture<ExecutionContext> p) {
```
knutwannheden, line 69: I suspect that this method can now probably be entirely removed, as it will be covered by the `visit()` method now.

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
macsux: Added. Now failing tests passing too...

sambsnyd:
Thanks @macsux !

PR #4563, opened 2024-10-09
Also exclude omitted dependencies

ckcd:
Try to complete https://github.com/openrewrite/rewrite/pull/4463

The root cause should be in this line we ignore the omitted dependencies.
https://github.com/openrewrite/rewrite/blob/ed13f4a599c904258b5a29bcb8eb8677eee10dc6/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java#L913

So I try to add a new `omitDependencies` for `ResolvedDependency` to record these omitted information. And uses this `omitDependencies` when search for potential relation in `ExcludeDependency`. To avoid affecting other logic, I added a new method `findDependencyWithOmit`.

## What's changed?

`ExcludeDependency` also exclude omitted dependencies.

## What's your motivation?

When conflicts occur dependencies are omitted; when we then exclude from the conflict, the omitted dependency can resurface. The goal of ExcludeDependency is to fully exclude a dependency, not to see the omitted conflict being used. We should then also add `<exclusion>` where dependencies are omitted.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedDependency.java
@@ -52,6 +52,10 @@ public class ResolvedDependency implements Serializable {
     @EqualsAndHashCode.Exclude
     List<ResolvedDependency> dependencies;
 
+    @NonFinal
+    @EqualsAndHashCode.Exclude
+    List<GroupArtifactVersion> omitDependencies;
+
     List<License> licenses;
 
     @Nullable
@@ -79,6 +83,10 @@ public class ResolvedDependency implements Serializable {
      */
     void unsafeSetDependencies(List<ResolvedDependency> dependencies) {
         this.dependencies = dependencies;
+    }
+
+    void unsafeSetOmitDependencies(List<GroupArtifactVersion> omitDependencies) {
+        this.omitDependencies = omitDependencies;
     }
 
     void unsafeSetEffectiveExclusions(List<GroupArtifact> effectiveExclusions) {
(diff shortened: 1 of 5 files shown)


ckcd:
@timtebeek Thanks for your reply!

Sure for the `boolean` related I can refine the logic to make it more concise.

And just confirm, do you mean we can't add new field to `ResolvedDependency`? Generally add a field does not affect backward compatibility I think.

ckcd:
+ @SiBorea

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedDependency.java
```diff
@@ -142,6 +150,40 @@ public boolean isTransitive() {
...
+                                matchesGlob(found.getArtifactId(), exclusion.getArtifactId())) {
+                            continue outer;
+                        }
+                    }
+                }
+                return found;
+            }
+        }
+
+        for (GroupArtifactVersion omit : omitDependencies) {
```
timtebeek, line 179: This is where we'd have to handle a possible `null` `omitDependencies` field.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -916,6 +916,13 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
                         } else if (contains(dependencies, ga, d.getClassifier())) {
                             // we've already resolved this previously and the requirement didn't change,
                             // so just skip and continue on
+                            ResolvedDependency includedBy = dd.getDependent();
+                            if (includedBy != null) {
+                                if (includedBy.getOmitDependencies().isEmpty()) {
+                                    includedBy.unsafeSetOmitDependencies(new ArrayList<>());
+                                }
+                                includedBy.getOmitDependencies().add(new GroupArtifactVersion(d.getGroupId(), d.getArtifactId(), d.getVersion()));
```
timtebeek, line 924: Also here the generated getter might return null; as an alternative we could create an explicit getter that always returns a collection.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedDependency.java
```diff
@@ -52,6 +52,10 @@ public class ResolvedDependency implements Serializable {
     @EqualsAndHashCode.Exclude
     List<ResolvedDependency> dependencies;
 
+    @NonFinal
+    @EqualsAndHashCode.Exclude
+    List<GroupArtifactVersion> omitDependencies;
```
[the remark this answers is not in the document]
SiBorea: I have did a little defensive programming for it

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
SiBorea: `getOmitDependencies()` never return null now.

SiBorea: `getOmitDependencies()` never return null now

timtebeek:
Thanks for the work here @ckcd & @SiBorea ! It took me a while to think through the options, and from my other reviews you'll know you're not forgotten; it's just at times hard to get to everything.

For the exact problem here with ExcludeDependency I found a way to make that work without changing the model, which I think is preferred for now.
- https://github.com/openrewrite/rewrite/pull/4724

Thanks a again for the work done here, and prompting me to have another look. I look forward to having the underlying fix available.

PR #4565, opened 2024-10-10
Added semicolon test for Java parser

rlsanders4:
## What's changed?
This adds a new unit test to `JavaParserTest` that validates that the parser is working correctly when a simicolon is specified directly after the final bracket in a method declaration.

## What's your motivation?
A semicolon specified in this manner is unnecessary, but it is still valid Java code. The parser is currently unable to parse this, and the unit test gives the following error:
```
Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser. 
public class Test {
  public static void myMethod() {
  }~~(non-whitespace)~~>;
<~~}
```

### Checklist


timtebeek:
Thanks again for reporting this issue @rlsanders4 ; logging some useful context from Slack here as well:
> I believe mapping these semicolons to J.Empty statements would be the correct course of action.

timtebeek:
So to give a bit of context: we want to be able to losslessly parse and reproduce any source file. As a part of that we have validations that we appropriately model all source elements, and check that we not for instance incorrectly model non-whitespace characters as whitespace, as that could interfere with visiting & modifying those elements. If there are any such model and reproduce errors then we decide not to make _any_ changes to the affected file, as to do no harm.

In this exact case there does not appear to be a reliable way to model that `;` after a method invocation, if I understand you correctly. Then we ought to either introduce a structure that could model that `;`, or tolerate that `;` in this specific case in the whitespace, and allow any other changes to the same file, as opposed to blocking any such changes.

nielsdebruin:
Sidenote: The latter will also happen when you do things like this `import java.util.List;;`. But it won't when you do ` 
  /**/;;;`. In that case `;;;` become part of the `suffix` field on the `Comment` class which are stored inside a field on `Space`.

timtebeek:
Funny that you point out those semicolons after imports; we actually have a recipe for those
- https://docs.openrewrite.org/recipes/java/migrate/removeillegalsemicolons
- https://bugs.openjdk.org/browse/JDK-8027682

timtebeek:
hi @rlsanders4 ; We've done a couple quick tests here and it seems this is only an issue when those semicolons are present in tests using RewriteTest. That's where we have additional validations to help improve our parsers.

In practice a project that has these extra semi colons following a method declaration can still be parsed, modified and printed according to our tests. Modeling these semicolons any different would be slightly more complicated for little added value, we feel. Recipes can still target these semicolons to remove them [as seen in this example](https://docs.openrewrite.org/recipes/java/migrate/removeillegalsemicolons).

I hope that's satisfactory for you as well; do let us know if you see any other behavior that would warrant another look.

PR #4566, opened 2024-10-10
Add `ResolutionEventListener#downloadMetadataSuccess` to communicate long download times to consumers of `MavenPomDownloader`

bryceatmoderne:
## What's changed?
Long download times for small files such as `maven-metadata.xml` can indicate network latency issues to internal artifactories. This PR introduces `ResolutionEventListener#downloadMetadataSuccess` as a method to relay download times to consumers of `MavenPomDownloader` where they optional log warnings to assist troubleshooting long recipe runs and time outs.

## What's your motivation?
These changes are intended to assist in troubleshooting recipe run execution time outs.

No.

@pstreef @timtebeek 

Currently recorded metrics do not provide file level granularity and doing so would be prohibitively expensive WRT cardinality.

## Any additional context
No.

### Checklist


pstreef:
you should also include `MavenArtifactDownloader`

bryceatmoderne:
> you should also include `MavenArtifactDownloader`

After looking into this, `MavenExecutionContextView.resolutionListener` may already provide the desired functionality of this PR...

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -602,8 +602,9 @@ public Pom download(GroupArtifactVersion gav,
                             pom = pom.withGav(pom.getGav().withDatedSnapshotVersion(versionMaybeDatedSnapshot));
                         }
                         mavenCache.putPom(resolvedGav, pom);
-                        ctx.getResolutionListener().downloadSuccess(resolvedGav, containingPom);
-                        sample.stop(timer.tags("outcome", "downloaded").register(Metrics.globalRegistry));
+                        long nanos = sample.stop(timer.tags("outcome", "downloaded").register(Metrics.globalRegistry));
+                        ctx.getResolutionListener()
+                                .downloadSuccess(resolvedGav, containingPom, Duration.ofNanos(nanos));
```
sambsnyd, line 607: So the old, durationless event type still exists but will now never be called? That isn't great. If the old event is still going to exist it should continue to be called.

sambsnyd (changes requested):

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
bryceatmoderne: I would prefer to remove the old event under the assumption only Moderne is consumers of it.

pstreef:
@bryceatmoderne should we maybe close this?

PR #4567, opened 2024-10-10
Handle case-insensitive document declaration in the XML parser

lkerford:
Current we are strict on the document declaration for files which are parsed by the XML parser, the document declaration must be in upper case. Going forward support document declarations regardless of their case.

We want to do this to allow the XML to parse HTML files, which aren't as strict in their formatting


rewrite-xml/src/main/antlr/XMLLexer.g4
```diff
@@ -65,7 +65,7 @@ DTD_CLOSE        :  '>'    -> popMode ;
 DTD_SUBSET_OPEN  :  '['    -> pushMode(INSIDE_DTD_SUBSET) ;
 DTD_S            :   S     -> skip ;
 
-DOCTYPE          :  'DOCTYPE' ;
+DOCTYPE options { caseInsensitive = true; }         :  'DOCTYPE' ;
```
lkerford, line 68: We do have the option to apply this to the entire file instead of specifically `DOCTYPE`

rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
```diff
@@ -669,6 +674,7 @@ public String getPrefix() {
 
         Markers markers;
         Ident name;
+        String documentDeclaration;
```
sambsnyd, line 677:
```suggestion
        String documentDeclaration;
        // Override lombok default getter to avoid backwards compatibility problems with old LSTs
        public getDocumentDeclaration() {
            //noinspection ConstantValue
            if ( documentDeclaration == null) {
                return "DOCTYPE"
            }
            return documentDeclaration;
        }
```

PR #4568, opened 2024-10-10
cleanup temp remote artifact

natedanner:
## What's changed?
Move cleanup of temporary artifact in the put method of `LocalRemoteArtifactCache.java` to the finally block. 

## What's your motivation?
Having disk fill up in the Moderne SaaS on a large run that is updating gradle wrappers.
Related to: https://github.com/moderneinc/customer-requests/issues/620

@sambsnyd

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/remote/LocalRemoteArtifactCache.java
+++ b/rewrite-core/src/main/java/org/openrewrite/remote/LocalRemoteArtifactCache.java
@@ -17,6 +17,7 @@ package org.openrewrite.remote;
 
 import org.jspecify.annotations.Nullable;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


sjungling (approved):

rewrite-core/src/main/java/org/openrewrite/remote/LocalRemoteArtifactCache.java
```diff
@@ -46,8 +46,9 @@ public LocalRemoteArtifactCache(Path cacheDir) {
 
     @Override
     public @Nullable Path put(URI uri, InputStream artifactInputStream, Consumer<Throwable> onError) {
+        Path artifact = null;
```
sambsnyd, line 49: I suggest something like this:
```java
    @Override
    public @Nullable Path put(URI uri, InputStream artifactInputStream, Consumer<Throwable> onError) {
        synchronized (this) {
            Path artifact = cacheDir.resolve(UUID.randomUUID() + ".tmp");
            try (InputStream is = artifactInputStream) {
                Files.copy(is, artifact, StandardCopyOption.REPLACE_EXISTING);
                Path cachedArtifact = cacheDir.resolve(hashUri(uri));
                if (!Files.exists(cachedArtifact)) {
                    Files.move(artifact, cachedArtifact, StandardCopyOption.ATOMIC_MOVE,
                            StandardCopyOption.REPLACE_EXISTING);
                }
                return cachedArtifact;
            } catch (Exception e) {
                onError.accept(e);
                return null;
            } finally {
                if (Files.exists(artifact)) {
                    try {
                        Files.delete(artifact);
                    } catch (IOException ignored) {
                        // Suppress
                    }
                }
            }
        }
    }
```

PR #4573, opened 2024-10-11
Support parsing the Metadata field in html using the XML parser

lkerford:
We want to be able to parse HTML using the XML parse, this change will enable us to parse the meta table tags


sambsnyd (approved):

sambsnyd:
There are a number of "void" elements in HTML which do not have a closing element. We should probably be able to handle them all alike:
https://html.spec.whatwg.org/multipage/syntax.html#void-elements

lkerford:
> There are a number of "void" elements in HTML which do not have a closing element. We should probably be able to handle them all alike:

If we decide to go forward and support all HTML, let make this more tag agnostic. For now lets just support the "meta" tag

PR #4574, opened 2024-10-11
Support parsing the Metadata field in html using the XML parser

lkerford:
We want to be able to parse HTML using the XML parse, this change will enable us to parse the meta table tags


lkerford:
Making this a PR for ease of finding if we decide to go forward and fully support HTML as a language. 

Couple other notes: supporting the "&" and "<" characters in strings needs to be added and we need to support all void elements

PR #4579, opened 2024-10-14
Do not override version property placeholder in UpgradeDependencyVersion

dimijdriven:
## What's changed?
bug-474: added relevant test, in some cases the provided requested value contains the version, I added an extra check getting the value directly from the pom

## What's your motivation?
The provided requested versions was not always correct

- Fixes https://github.com/openrewrite/rewrite-spring/issues/474

@Laurens-W 

## Any additional context

This solves half the issue, it stops the placeholder from being replaced from the version.
The spring boot version is still not upgraded, but this is because of the way that the upgrade call is done 
with org.springframework and not with org.springframework.boot

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
@@ -39,82 +39,6 @@ import static org.openrewrite.maven.Assertions.pomXml;
 class UpgradeDependencyVersionTest implements RewriteTest {
 
     @Test
-    @Issue("https://github.com/openrewrite/rewrite-spring/issues/474")
-    void upgradePropertyAdditionalCheck() {
-        rewriteRun(
-          spec -> spec.recipe(new UpgradeDependencyVersion("org.springframework", "*", "5.2.x", "",
-            false, null)).expectedCyclesThatMakeChanges(2),
-          mavenProject("project",
-            //language=xml
-            pomXml(
-              """
-                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-                	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
-                	<modelVersion>4.0.0</modelVersion>
-                	<groupId>com.openrewrite.example</groupId>
-                	<artifactId>example</artifactId>
-                	<version>0.0.1-SNAPSHOT</version>
-                	<name>example</name>
-                	<description>Demo project for Spring Boot</description>
-                	<properties>
-                		<spring.boot.version>2.2.13.RELEASE</spring.boot.version>
-                		<spring.jms.version>5.2.22.RELEASE</spring.jms.version>
-                	</properties>
-                	<dependencyManagement>
-                		<dependencies>
-                			<dependency>
-                				<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -402,6 +407,37 @@ private String resolveVersion(String version) {
...
             }
+
+            /**
+             * Gets the requested group id for the provided artifact id from the pom
+             *
+             * @return The groupd if for the requested artifact id or null
+             */
+            @Nullable
+            private String getRequestedGroupIdForArtifact(String artifactId) {
+                return getResolutionResult().getPom().getRequested().getDependencyManagement().stream()
```
Laurens-W, line 418: Probably best not to use streams here as this recipe is potentially ran quite a few times in a single run and the overhead of the intermediate objects could be too high

rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -402,6 +407,28 @@ private String resolveVersion(String version) {
                 return MavenDependency.findNewerVersion(groupId, artifactId, version, getResolutionResult(), metadataFailures,
                         versionComparator, ctx);
             }
+
+            /**
+             * Gets the requested group id for the provided artifact id from the pom
+             *
+             * @return The groupd if for the requested artifact id or null
+             */
```
timtebeek, line 415: The comments here seem to just echo the method name & types. Lets drop rather than maintain those.
```suggestion
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -38,6 +38,82 @@
 
 class UpgradeDependencyVersionTest implements RewriteTest {
 
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-spring/issues/474")
+    void upgradePropertyAdditionalCheck() {
+        rewriteRun(
+          spec -> spec.recipe(new UpgradeDependencyVersion("org.springframework", "*", "5.2.x", "",
+            false, null)).expectedCyclesThatMakeChanges(2),
```
Laurens-W, line 46: Can we check once more whether we can reduce the amount of cycles to 1? 
Enforcing that all recipes complete in 1 cycle is something that's being looked at

rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -38,6 +38,82 @@
...
+                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+                	<modelVersion>4.0.0</modelVersion>
+                	<groupId>com.openrewrite.example</groupId>
+                	<artifactId>example</artifactId>
+                	<version>0.0.1-SNAPSHOT</version>
+                	<name>example</name>
+                	<description>Demo project for Spring Boot</description>
+                	<properties>
+                		<spring.boot.version>2.2.13.RELEASE</spring.boot.version>
```
Laurens-W, line 91: This version property should be updated too, right?

Laurens-W: My bad, didn't read the group id properly, you are right :)

PR #4581, opened 2024-10-14
Added recipe for changing Groovy params

ThomasKianek-Gepardec:
## What's changed?

Added recipe for changing Groovy parameter value in method invocation.

## What's your motivation?

To be used in Jenkins files (e.g. to update the JDK version).

Preconditions for selecting a given method in Groovy did not work as expected (had to use if-statement instead).

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeGroovyMethodInvocationParameter.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeGroovyMethodInvocationParameter.java
@@ -1,0 +1,58 @@
+package org.openrewrite.groovy;
+
+
+import org.openrewrite.*;
+import org.openrewrite.groovy.tree.G;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.Markers;
+
+import java.util.UUID;
+
+public class ChangeGroovyMethodInvocationParameter extends Recipe {
+    @Option
+    public String key;
+
+    @Option
+    public String value;
+
+    public ChangeGroovyMethodInvocationParameter() {
+    }
+
+    public ChangeGroovyMethodInvocationParameter(String key, String value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    @Override
+    public @NlsRewrite.DisplayName String getDisplayName() {
+        return "Change groovy a  parameter";
+    }
+
+    @Override
+    public @NlsRewrite.Description String getDescription() {
+        return "It updates the value of a given parameter on all groovy method invocations.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new GroovyIsoVisitor<>() {
+            @Override
+            public J visitMapEntry(G.MapEntry mapEntry, ExecutionContext executionContext) {
+                mapEntry = (G.MapEntry) super.visitMapEntry(mapEntry, executionContext);
+
+                if (mapEntry.getValue().toString().equals("'" + value + "'")) {
+                    return mapEntry;
+                }
+
+                if (mapEntry.getKey().toString().equals(key)) {
+                    final J.Literal valueLiteral = new J.Literal(UUID.randomUUID(), Space.SINGLE_SPACE, Markers.EMPTY, "'" + value + "'", "'" + value + "'", null, JavaType.Primitive.String);
+                    return mapEntry.withValue(valueLiteral);
+                }
+
+                return mapEntry;
+            }
+        };
+    }
+}
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeGroovyMethodInvocationParameter.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeGroovyMethodInvocationParameter.java
@@ -15,7 +15,7 @@
  */
 package org.openrewrite.groovy;
 
-
+import org.jetbrains.annotations.NotNull;
 import org.openrewrite.*;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.java.tree.J;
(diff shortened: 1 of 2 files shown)


sgartner03:
Hi @timtebeek, I incorporated the suggested changes from the GH Action. I think this contribution matches your code styling now.

sgartner03:
Hi @timtebeek, yes exactly that is what this recipe does. Thank you for your, feedback, I'll incorporate it now.

sgartner03:
Hi, I've added 3 more tests. This tests:
- Multiple params
- Empty Methods
- Parameters of other type.

As of now, this recipe only supports Strings and GStrings

sgartner03:
@timtebeek

rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeStringValueOfNamedParameterInMethodInvocation.java
```diff
@@ -0,0 +1,92 @@
...
+
+                if (mapEntry.getValue().toString().equals(value)) {
+                    return mapEntry;
+                }
+
+                return replaceValue(mapEntry, quote);
+            }
+
+            private boolean isInTargetMethod() {
+                return getCursor().firstEnclosingOrThrow(J.MethodInvocation.class).getSimpleName().equals(methodName);
```
timtebeek, line 80: What's the intended use case here? Because we'd much rather match using a MethodMatcher as opposed to a String comparision, as this would match _any_ method anywhere of the same name, which is not as type safe as we prefer to be. For Gradle build script type information isn't always available to match reliably, but even then there's established patterns to be a little more sure we're not just matching any method by simple name, hence why I ask here.

rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeStringValueOfNamedParameterInMethodInvocation.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            private G. MapEntry replaceValue(final G.MapEntry mapEntry, char quote) {
+                return mapEntry.withValue(new J.Literal(Tree.randomId(), Space.SINGLE_SPACE, Markers.EMPTY, value, String.format("%c%s%c", quote, value, quote), null, JavaType.Primitive.String));
+            }
+
+            private char extractQuoting(final G.MapEntry mapEntry) {
+                return mapEntry.getValue().printTrimmed(getCursor()).charAt(0);
```
timtebeek, line 88: This is the only time I've seen `printTrimmed` used in all of `rewrite-groovy`; is there no safer alternative? It seems you're looking to skip over anything that's not a J.Literal or GString; perhaps that can be checked here?

rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeStringValueOfNamedParameterInMethodInvocation.java
```diff
@@ -0,0 +1,92 @@
...
+            @Override
+            public J visitMapEntry(G.MapEntry mapEntry, ExecutionContext ctx) {
+                mapEntry = (G.MapEntry) super.visitMapEntry(mapEntry, ctx);
+
+                if (!isInTargetMethod()) {
+                    return mapEntry;
+                }
+
+                char quote = extractQuoting(mapEntry);
+                if (quote != '\'' && extractQuoting(mapEntry) != '"') {
```
timtebeek, line 64: See below as well for a potential alternative looking at the type of the `mapEntry` instead, but calling the same method twice with the same argument is a bit suspect.
```suggestion
                char quote = extractQuoting(mapEntry);
                if (quote != '\'' && quote != '"') {
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeStringValueOfNamedParameterInMethodInvocation.java
```diff
@@ -0,0 +1,92 @@
...
+
+                return replaceValue(mapEntry, quote);
+            }
+
+            private boolean isInTargetMethod() {
+                return getCursor().firstEnclosingOrThrow(J.MethodInvocation.class).getSimpleName().equals(methodName);
+            }
+
+            private G. MapEntry replaceValue(final G.MapEntry mapEntry, char quote) {
+                return mapEntry.withValue(new J.Literal(Tree.randomId(), Space.SINGLE_SPACE, Markers.EMPTY, value, String.format("%c%s%c", quote, value, quote), null, JavaType.Primitive.String));
```
timtebeek, line 84: Would we not want to retain any original formatting and comments by using `withValue` and `withValueSource` instead?

timtebeek (commented):
hi @sgartner03 ; I was out hiking a few days, hence the delayed response. I've pushed up a small formatting for the tests, and posted a number of questions on the implementation that I think we should address before a merge. ~A wider context of intended use might help for this otherwise generic recipe.~ I see now that this is intended to be used for Jenkins projects. Perhaps we can start this as a visitor in `rewrite-jenkins` first and let it mature from there?

rewrite-groovy/src/main/java/org/openrewrite/groovy/ChangeStringValueOfNamedParameterInMethodInvocation.java
sgartner03: Yeah, you're right. I've removed extractQuoting

sgartner03:
Hey @timtebeek ! Thank you for your extensive feedback. This is really valuable for me - so far I've only developed InHouse-Recipes for one Repo, where I didn't focus on clean code a lot.

Yes, you're right, this is intended for Jenkinsfiles on our end. We thought, it could be used in other Groovy Use-Cases too.

Initially, we thought we could decide between GString / J.String by using instanceof on mapEntry#getValue. However, we did notice that through all our tests, only J.Strings exist. Meaning, ", as well as ' quotes are represented by J.String. But I've removed extractQuoting() and replaced it by filtering through J.Literal#getType and just replacing the old value inside of J.Literal#getValueSource with the new one. This also replaced the constructor-call with just ".with*"-Operations.

 However, I'm not that sure about using a MethodMatcher. I have to point out, that my groovy knowledge is close to zero. But how can types even be used in terms of stuff like a Jenkinsfile or other groovy stuff? You've mentioned other established ways to limit the scope, what would these be? And can't limiting the scope be implemented with a precondition on the user's end?
 
 (PS I didn't mean to stress - I just forgot to ping you on the initial message)

PR #4583, opened 2024-10-16
Add `addToRootPom` option to `AddPlugin` recipe

Laurens-W:
## What's your motivation?

Generally to avoid conflicts in a multimodule project you want to manage plugins from a parent/root pom.
However the current behavior of the `AddPlugin` is to add it to every single pom.

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
@@ -75,9 +75,9 @@ public class AddPlugin extends Recipe {
 
     @Option(displayName = "File pattern",
             description = "A glob expression that can be used to constrain which directories or source files should be searched. " +
-                                "Multiple patterns may be specified, separated by a semicolon `;`. " +
-                                "If multiple patterns are supplied any of the patterns matching will be interpreted as a match. " +
-                                "When not set, all source files are searched. ",
+                    "Multiple patterns may be specified, separated by a semicolon `;`. " +
+                    "If multiple patterns are supplied any of the patterns matching will be interpreted as a match. " +
+                    "When not set, all source files are searched. ",
             required = false,
             example = "**/*-parent/grpc-*/pom.xml")
     @Nullable
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
@@ -20,6 +20,7 @@ import lombok.Value;
 import org.intellij.lang.annotations.Language;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
+import org.openrewrite.maven.tree.MavenResolutionResult;
 import org.openrewrite.xml.AddToTagVisitor;
 import org.openrewrite.xml.ChangeTagValueVisitor;
 import org.openrewrite.xml.XPathMatcher;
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
```diff
@@ -154,13 +169,13 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
-                            (configuration != null ? configuration.trim() + "\n" : "") +
-                            (dependencies != null ? dependencies.trim() + "\n" : "") +
-                            "</plugin>");
+                                                      "<groupId>" + groupId + "</groupId>\n" +
+                                                      "<artifactId>" + artifactId + "</artifactId>\n" +
+                                                      (version != null ? "<version>" + version + "</version>\n" : "") +
+                                                      (executions != null ? executions.trim() + "\n" : "") +
+                                                      (configuration != null ? configuration.trim() + "\n" : "") +
+                                                      (dependencies != null ? dependencies.trim() + "\n" : "") +
+                                                      "</plugin>");
```
timtebeek, line 178: Let's minimize these types of changes.

Laurens-W: I'll revert the indent changes!

rewrite-maven/src/test/java/org/openrewrite/maven/AddPluginTest.java
```diff
@@ -361,4 +362,135 @@ void addPluginWithExistingPlugin() {
...
+    @Test
+    void addPluginOnlyToRootPom() {
+        rewriteRun(
+          spec -> spec.recipe(new AddPlugin("org.springframework.boot", "spring-boot-maven-plugin", "3.1.5", null, null, null, null, true)),
+          pomXml(
+            """
+              <project>
+                <groupId>com.mycompany.app</groupId>
+                <artifactId>my-app</artifactId>
+                <version>1</version>
```
timtebeek, line 375: Can we add another test here where the parent in the repository has an external parent like Spring Boot? I'd want to make sure we still add a plugin in such common cases.

Laurens-W: I completely missed your comment, sorry! Added the testcase

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
```diff
@@ -109,6 +116,18 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
...
             }
+            if (!Boolean.TRUE.equals(addToRootPom)) {
+                return super.isAcceptable(sourceFile, ctx);
+            } else {
+                Optional<MavenResolutionResult> mvnResult = sourceFile.getMarkers().findFirst(MavenResolutionResult.class);
+                if (mvnResult.isPresent()) {
+                    if (mvnResult.get().getParent() == null) {
+                        return super.isAcceptable(sourceFile, ctx);
+                    } else {
+                        return false;
```
timtebeek, line 127: The if block here is the same as of we were to fall through. Let's convince the two if blocks, negate the second condition, and just return false in that case to simplify the nesting here

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
timtebeek: I think if you put the initial `"<plugin>"` on a new line as well you stop IntelliJ from butchering the indentation.

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
```diff
@@ -82,6 +83,12 @@ public class AddPlugin extends Recipe {
     @Nullable
     String filePattern;
 
+    @Option(displayName = "Add to the root pom",
+            description = "Add to the root pom where root is the eldest parent of the pom within the source set.",
+            required = false)
+    @Nullable
+    Boolean addToRootPom;
```
jkschneider, line 90: I don't agree that this should be an option. In the presence of a multimodule project, always add it to the root pom. 

I don't think it would ever make sense to run this recipe with this being set to false as running the recipe in that configuration on a mixture of single and multi-module projects would always do the wrong thing on some set of repositories.

jkschneider (changes requested):

rewrite-maven/src/main/java/org/openrewrite/maven/AddPlugin.java
sambsnyd: I've pushed a commit to fix this

PR #4587, opened 2024-10-17
Add `TypeReference` trait to identify Java type references in XML

Laurens-W:
## What's changed?
A new `TypeReference` trait interface is added which is used by any source file types implementing the new `SourceFileWithTypeReferences` interface and allows the source file to report any type references that the source file may contain (for example Java type references in Spring configuration files), so that they can be reported by search recipes and ultimately also be refactored by recipes like `ChangeType`.

The `TypeReference.Provider` interface is a service interface which uses the `ServiceLoader` mechanism to find implementations. Thus, a recipe jar can contribute implementations which then can return type references in source files (e.g. `application.properties` files).

## What's your motivation?
Configuration files often contain type references, which currently don't get reported or updated by recipes like `FindTypes` and `ChangeType`. The `TypeReference` trait and `Provider` service mechanism allow registering providers for specific source files (e.g. matching a given source path or XML schema) to locate and return any type references in the source file and thus transparently be included by recipes with support for this mechanism.


rewrite-xml/src/main/java/org/openrewrite/xml/trait/ContainsPackageReference.java
```diff
@@ -0,0 +1,76 @@
...
+                return tag.getValue().get();
+            }
+        }
+        return null;
+    }
+
+
+    public static class Matcher extends SimpleTraitMatcher<ContainsPackageReference> {
+        private final Pattern PACKAGE_OR_TYPE_REFERENCE = Pattern.compile("^([a-zA-Z_][a-zA-Z0-9_]*)(\\.[a-zA-Z_][a-zA-Z0-9_]*)*(\\.[A-Z][a-zA-Z0-9_]*|\\$[A-Z][a-zA-Z0-9_]*)*(\\.\\*)?$");
+        private final List<String> ATTRIBUTES_THAT_REFERENCE_PACKAGE_OR_TYPE = Arrays.asList("class", "type");
```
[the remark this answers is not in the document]
timtebeek: I think regular expressions are an unsafe way to go about it, but I'm not exactly sure what a better approach would be.  XPath would be a little more predictable, but also feel out of place here to have Spring/Micronaut/Quarkus/EE expressions all in here to be picked up when changing types.

Perhaps @knut has some suggestions on using service discovery where frameworks specific recipe modules can provide a mechanism to provide additional paths to check and convert.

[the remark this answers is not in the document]
knutwannheden: I think it would probably be best if we could register some kind of "extensions" with parsers like those for XML, YAML, Properties, etc. So we could then already while ingesting the source file attach some kind of `JavaTypeReference` marker to the LST elements. Then the `ChangeType` recipe could check if the `SourceFile` if it has some kind of `HasJavaTypeReferences` marker or possibly it even uses the `TypesInUse` we use for `JavaSourceFile`.

An additional complication may be that these "parser extensions" might have to be registered dynamically (e.g. depending on whether the project is a Spring project or not).

This could quickly end up being quite a bit of work. But I get the impression that this would long-term be the better and more sustainable solution. Maybe we can still do something like this, but start out small?

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringJavaTypeReference.java
Laurens-W, line 1: This should be moved to rewrite-spring

rewrite-xml/src/main/java/org/openrewrite/xml/internal/JavaTypeReferences.java
```diff
@@ -0,0 +1,42 @@
...
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter
+public class JavaTypeReferences {
+    private final Xml.Document document;
+    private final Set<JavaTypeReference> typeReferences;
+
+    public static JavaTypeReferences build(Xml.Document doc) {
+        Set<JavaTypeReference> typeReferences = new HashSet<>();
+        new SpringJavaTypeReference.Matcher().asVisitor(reference -> {
```
Laurens-W, line 36: Need to figure out a dynamic way to pass a trait to use here

rewrite-core/src/main/java/org/openrewrite/trait/JavaTypeReference.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.trait;
+
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Tree;
+
+public interface JavaTypeReference extends Trait<Tree> {
```
Laurens-W, line 21: Should I remove the `extends` here and place it back on `SpringJavaTypeReference.java` as to not force a `JavaTypeReference` to be a trait?

rewrite-core/src/main/java/org/openrewrite/internal/JavaTypeReferences.java
```diff
@@ -0,0 +1,45 @@
...
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter
+public class JavaTypeReferences {
+    private final SourceFile document;
+    private final Set<JavaTypeReference> typeReferences;
+
+    public static JavaTypeReferences build(SourceFile doc) {
+        Set<JavaTypeReference> typeReferences = new HashSet<>();
+        ServiceLoader<JavaTypeReferenceProvider> loader = ServiceLoader.load(JavaTypeReferenceProvider.class);
```
Laurens-W, line 37: We might have to use a different classloader here, which is possible because this method has an overloaded variant that takes a classloader as argument

rewrite-core/src/main/java/org/openrewrite/SourceFileWithTypeReferences.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
+
+import org.openrewrite.internal.JavaTypeReferences;
+
+public interface SourceFileWithTypeReferences {
```
Laurens-W, line 20: Not sure if this is over interfacing it, currently it's only about XML documents. Do we see any future use cases for other types of SourceFiles

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringJavaTypeReference.java
Laurens-W, line 1: This should be moved to rewrite-spring, together with it's test and the SpringJavaTypeReferenceProvider

rewrite-core/src/main/java/org/openrewrite/SourceFileWithTypeReferences.java
Laurens-W: After talking to Knut: Both Spring and Quarkus have an application config file that could be written in YAML and could contain type references, so there are definitely use cases outside of Xml.

rewrite-core/src/main/java/org/openrewrite/trait/JavaTypeReference.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.openrewrite.trait;
+
+import org.jspecify.annotations.Nullable;
+
+public interface JavaTypeReference {
+
+    @Nullable
+    String getValue();
```
knutwannheden, line 23: I think if the XML element or attribute has a null or empty value, there should be no reference.
```suggestion
    String getName();
```

rewrite-xml/src/main/java/org/openrewrite/xml/SpringJavaTypeReferenceProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.xml;
```
knutwannheden, line 16: I was going to suggest to move this to an internal package, but we might even have to move it to the `tree` package for the class loader issues we've discussed. I think we will have to test this and also make sure that the class loaders can properly load the `META-INF/services` files.

rewrite-xml/src/main/java/org/openrewrite/xml/SpringJavaTypeReferenceProvider.java
```diff
@@ -0,0 +1,51 @@
...
+import org.openrewrite.JavaTypeReferenceProvider;
+import org.openrewrite.SourceFile;
+import org.openrewrite.trait.JavaTypeReference;
+import org.openrewrite.xml.trait.SpringJavaTypeReference;
+import org.openrewrite.xml.tree.Xml;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class SpringJavaTypeReferenceProvider implements JavaTypeReferenceProvider {
```
knutwannheden, line 27: AFAIK the service implementation classes are allowed to be package private.

```suggestion
class SpringJavaTypeReferenceProvider implements JavaTypeReferenceProvider {
```

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringJavaTypeReference.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openrewrite.trait.JavaTypeReference;
+import org.openrewrite.trait.SimpleTraitMatcher;
+import org.openrewrite.trait.Trait;
+import org.openrewrite.xml.XPathMatcher;
+import org.openrewrite.xml.tree.Xml;
+
+import java.util.regex.Pattern;
+
+@Value
+public class SpringJavaTypeReference implements JavaTypeReference, Trait<Tree> {
```
knutwannheden, line 31: Also this class can be package private I think.

```suggestion
class SpringJavaTypeReference implements JavaTypeReference, Trait<Tree> {
```

rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
```diff
@@ -162,6 +167,28 @@ public <S, T extends S> T service(Class<S> service) {
             }
             return SourceFile.super.service(service);
         }
+
+        @Nullable
+        @NonFinal
+        SoftReference<JavaTypeReferences> javaTypeReferences;
```
knutwannheden, line 173: Not sure how important that `@Transient` on the getter is, but we should definitely have a `transient` modifier here.

```suggestion
        transient SoftReference<JavaTypeReferences> javaTypeReferences;
```

rewrite-core/src/main/java/org/openrewrite/internal/JavaTypeReferences.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.HashSet;
+import java.util.ServiceLoader;
+import java.util.Set;
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter
+public class JavaTypeReferences {
+    private final SourceFile document;
+    private final Set<JavaTypeReference> typeReferences;
```
knutwannheden, line 33: I know this is basically how it is done for `TypesInUse`, but if we in a first step can instead declare a query method (something like `findExactMatch(String)`) that would be better.

Also, we should already think one step further here: What will `ChangePackage` and `ChangeType` require once we know that there is a reference that needs to be updated? How can we satisfy that requirement?

knutwannheden (commented):
@Laurens-W Sorry about some of the very vague comments here.

@timtebeek What are your thoughts about this?

rewrite-core/src/main/java/org/openrewrite/TypeReferenceProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
+
+import org.openrewrite.trait.TypeReference;
+
+import java.util.Set;
+
+public interface TypeReferenceProvider {
```
knutwannheden, line 22: Should we nest this interface as `Provider` inside `TypeReference` or is it too awkward with a file name like `META-INF/services/org.openrewrite.TypeReference$Provider`?

rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.trait;
+
+public interface TypeReference {
+
+    String getName();
```
knutwannheden, line 20: We might have to add that `getTree()` method from `Trait` or some other way so we can correlate it with the LST and later also update the LST.

rewrite-xml/src/main/java/org/openrewrite/xml/SpringJavaTypeReferenceProvider.java
Laurens-W: Seems the implementation has to be public: 
```Caused by: java.lang.IllegalAccessException: class java.util.ServiceLoader$ProviderImpl (in module java.base) cannot access a member of class org.openrewrite.xml.trait.SpringTypeReferenceProvider with modifiers "public"```

File: rewrite-core/src/main/java/org/openrewrite/TypeReferenceProvider.java (hunk 12 above)
Comment (Laurens-W) [reply rewrite-core/src/main/java/org/openrewrite/TypeReferenceProvider.java:22]: It does feel slightly awkward but we'll have an implementation as an example that we can point to right away, and it's also a convention so it can be explained

File: rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java (hunk 13 above)
Comment (Laurens-W) [reply rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java:20]: I didn't want to tightly couple `TypeReference` to `Trait` but at that point it might make sense to have `TypeReference` extend `Trait<Tree>` perhaps?

File: rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java [hunk 14]
```diff
@@ -0,0 +1,51 @@
...
+
+    default boolean supportsRename() {
+        return false;
+    }
+
+    default TreeVisitor<Tree, ExecutionContext> renameTo(String name) {
+        throw new UnsupportedOperationException();
+    }
+
```
timtebeek, line 36: I'd noticed these two are unused for now; would we still want those to be part of the initial spec?
If so I think we should implement these from SpringTypeReference already, as that's the use case here.

knutwannheden: We can also remove them for now. I added them as I wanted to validate the design. Implementing ChangeType felt like it would be too much work, so I stopped there.

timtebeek: @Laurens-W I think we started down this route specifically to support change type right? Would it make sense to add this in before we merge? Otherwise we'd immediately have to change this again, across two repositories. Or would you want to not merge the SpringTypeReference here and have that in rewrite-spring only?

Laurens-W: I think merging this in the current state allows Knut to do some testing regarding the CLI while I can simultaneously make changes to `ChangeType`, I do agree that the Spring specific stuff should be moved to rewrite-spring before merging this though

knutwannheden: I think it would be nice if we can leave the SpringTypeReference in here until we can verify it works in CLI. I actually suspect it won't work, but I find it quite hard to test the CLI against unrelease OR changes. So I propose we merge this PR (with or without the methods above) and then I look into getting it working in CLI and SaaS. In a second step we can then move it. This also allows us to fix ChangeType without having to modify two repos.

knutwannheden: I would leave the Spring stuff in here for now. It doesn't really hurt and allows us to better validate the implementation before moving it. Also, it doesn't expose any public API, as SpringTypeReference is package private.

knutwannheden (approved):
I propose we apply whatever cleanups we still had in mind and then merge it. After that we make sure it works in CLI and extend ChangeType (and ChangePackage). In a last step we can then move SpringTypeReference to rewrite-spring.

timtebeek (approved):

PR #4590, opened 2024-10-18
Apply add dependency correctly, taking aggregating poms into account

Laurens-W:
## What's changed?

Introduce a mechanism to know which modules an aggregating pom applies to in order to prevent unnecessarily adding dependencies to aggregating poms

@timtebeek @sambsnyd 

## Any additional context

An aggregating pom could be a parent, when it's not there's no point in adding a dependency to it

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -235,7 +238,14 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
                     }
                 }
 
-                if(onlyIfUsing == null && getResolutionResult().getParent() != null && acc.pomsDefinedInCurrentRepository.contains(getResolutionResult().getParent().getPom().getGav())) {
+                if (onlyIfUsing == null && getResolutionResult().getParent() != null && acc.pomsDefinedInCurrentRepository.contains(getResolutionResult().getParent().getPom().getGav())) {
+                    return maven;
+                }
+
+                if (!getResolutionResult().getPom().getModules().isEmpty()
```
Laurens-W, line 245: Haven't fully got this logic sorted, will continue on Monday

Laurens-W:
The naming between `getResolutionResult().getModules()` and `getResolutionResult().getPom().getModules()` is a bit ambiguous. 
`getResolutionResult().getModules()` are child poms that we've determined within the project
`getResolutionResult().getPom().getModules()` refers to the modules element inside a pom

Not sure which one to refactor is any

rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
```diff
@@ -115,6 +115,9 @@ public static int getModelVersion() {
     @Builder.Default
     List<Plugin> pluginManagement = emptyList();
 
+    @Builder.Default
+    List<String> modules = emptyList();
```
timtebeek, line 119: To avoid confusion & future changes, perhaps it helps to already adopt the Maven 4 convention of [calling these subprojects](https://github.com/openrewrite/rewrite/discussions/4553#discussioncomment-10996072)?

Laurens-W: Sounds good, that takes away the ambiguity!

Laurens-W: I've taken both `modules` and `subProjects` into account for the `RawPom` and named it `subProjects` anywhere beyond there

Laurens-W:
Perhaps some additions to the `RawPomTest` would be justified here

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -235,7 +238,14 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
                 }
 
-                if(onlyIfUsing == null && getResolutionResult().getParent() != null && acc.pomsDefinedInCurrentRepository.contains(getResolutionResult().getParent().getPom().getGav())) {
+                if (onlyIfUsing == null && getResolutionResult().getParent() != null && acc.pomsDefinedInCurrentRepository.contains(getResolutionResult().getParent().getPom().getGav())) {
+                    return maven;
+                }
+
+                if (!getResolutionResult().getPom().getSubprojects().isEmpty()
+                    && (getResolutionResult().getModules().isEmpty()
+                        || getResolutionResult().getModules().stream().map(MavenResolutionResult::getPom).map(ResolvedPom::getGav).map(ResolvedGroupArtifactVersion::getArtifactId).noneMatch(art -> getResolutionResult().getPom().getSubprojects().contains(art))
```
timtebeek, line 247: The bot wants you to place the `&&` and `||` on the previous line, in case that wasn't clear before.

Any why the use of Streams API in one of our most used recipes? Does it help to extract the condition into a properly named method such that you can do an early return from a loop there?

rewrite-maven/src/test/java/org/openrewrite/maven/internal/RawPomTest.java
```diff
@@ -81,6 +81,41 @@ void repositoriesSerializationAndDeserialization() {
...
+            
+                  <modules>
+                    <module>my-module</module>
+                    <module>my-other-module</module>
+                  </modules>
+            
+                  <subprojects>
+                    <subproject>my-subproject</subproject>
+                    <subproject>my-other-subproject</subproject>
+                  </subprojects>
```
sambsnyd, line 104: Maven's concept of a "subproject" is called "module". If you actually put this in a Maven pom it refuses to proceed with a fatal error:
```
[ERROR]   The project com.mycompany.app:parent:1.0-SNAPSHOT (F:\Projects\maven-scratch\pom.xml) has 1 error
[ERROR]     Malformed POM F:\Projects\maven-scratch\pom.xml: Unrecognised tag: 'subprojects' (position: START_TAG seen ...</modules>\n\n  <subprojects>... @23:16)  @ F:\Projects\maven-scratch\pom.xml, line 23, column 16 -> [Help 2]
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
```diff
@@ -209,6 +209,10 @@ public boolean parentPomIsProjectPom() {
                 .anyMatch(gav -> gav.equals(parentGav));
     }
 
+    public boolean isAggregatingPom() {
+        return !getPom().getSubprojects().isEmpty();
+    }
```
sambsnyd, line 214: "Aggregating pom" is not a term a Maven user would necessarily understand to mean "multi-module"

sambsnyd (changes requested):

sambsnyd (approved):

timtebeek (approved):
Great to see; I've reworded the conditionals to make it easier to grasp what was added and why; hope you agree! :)

PR #4594, opened 2024-10-21
ChangePropertyValue in list yaml entries

### LINKED ISSUE #4593: ChangePropertyValue could improve array value behaviour
In our yamls we want to update an entry of a list in the yaml with a certain text value by another text value. If we would fo this with a simple text Find/Replace, the other occurences outside the propertyKey binding would also be updated. 

```yaml
job-name4:
  script:
    - replaceme
    - replaceme
    - do not replaceme
    - replaceme should not be done
  rules:
    - replaceme
```

When running with regex false for replaceme -> replaced
```yaml
job-name4:
  script:
    - replaced
    - replaced
    - do not replaceme
    - replaceme should not be done
  rules:
    - replaceme
```

When running with regex true for (.\*)replaceme(.\*) -> $1replaced$2
```yaml
job-name4:
  script:
    - replaced
    - replaced
    - do not replaced
    - replaced should not be done
  rules:
    - replaceme
```

The branch is present

Jenson3210:
- Fixes #4593

## What's changed?
Support Yaml.Sequence for ChangePropertyValue

## What's your motivation?
In our yamls we want to update an entry of a list in the yaml with a certain text value by another text value. If we would fo this with a simple text Find/Replace, the other occurences outside the propertyKey binding would also be updated.

Thinking about backwards compatibilty, I did not add a boolean flag that enables this behaviour. 

@timtebeek anyone you think of in particular?

Custom recipe for arrays?

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/ChangePropertyValue.java
@@ -116,17 +116,18 @@ public class ChangePropertyValue extends Recipe {
                     scalar.getValue().replaceAll(Objects.requireNonNull(oldValue), newValue) :
                     newValue);
             return scalar.getValue().equals(newScalar.getValue()) ? null : newScalar;
-        } else if (value instanceof Yaml.Sequence) {
+        }
+        if (value instanceof Yaml.Sequence) {
             Yaml.Sequence sequence = (Yaml.Sequence) value;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see! Good improvement indeed to support more cases with the same API. I've pushed up a very small polish but otherwise good to merge.

And do know you don't have to create branches on your fork anymore, although you're welcome to continue to do so of course.

timtebeek:
Hmm; as far as I can see you should be able to push branches as a member of the GitHub org.

I've merged just now; when I posted my polish commit was still running. 😇

Jenson3210:
Just verified, could create branch now 🥳

PR #4595, opened 2024-10-21
Setup for adding JavaType marker after parsing XML

Laurens-W:
## What's your motivation?

- https://github.com/openrewrite/rewrite/pull/4587#discussion_r1808474695

@knutwannheden @timtebeek 

- https://github.com/openrewrite/rewrite/pull/4587

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/marker/JavaType.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/marker/JavaType.java
@@ -1,0 +1,13 @@
+package org.openrewrite.xml.marker;
+
+import lombok.Value;
+import lombok.With;
+import org.openrewrite.marker.Marker;
+
+import java.util.UUID;
+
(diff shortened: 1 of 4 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/marker/JavaType.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Value;
+import lombok.With;
+import org.openrewrite.marker.Marker;
+
+import java.util.UUID;
+
+@Value
+@With
+public class JavaType implements Marker {
```
knutwannheden, line 11: I think we should probably move this marker to `rewrite-core` and name it something like `TypeReference` and then possibly add a `language` property to it. This way `FindType`, `ChangeType` and other recipes can support it and can at the same time be independent of the language that the markers occur in.

Issue #1297, opened 2021-12-28
Support Lombok Annotations


hetzijzo:
`org.openrewrite.java.cleanup.ExplicitInitialization` and `org.openrewrite.java.cleanup.UseDiamondOperator` removes the argument type when it's a lombok.val variable.

`val products = new ArrayList<Product>();`

`val products = new ArrayList<>();`

which is incorrect and doesn't compile anymore.

pway99:
Hi @hetzijzo, Thanks for reporting the issue we'll take a look.

pway99:
Unfortunately, we cannot fix this issue until support for Lombok annotations and types are added to the rewrite framework.

josemariavillar:
Good morning @pway99, is there any estimate to have the problem with the Lombok library solved?

pway99:
Good Morning @josemariavillar,  It's a tricky problem that will require a significant effort; at this time, we do not have an estimate.

pway99:
Hello @hetzijzo, That's a good question, I am not aware of a good workaround for Lombok issues at this time.  That said in the spirit of _do no harm_ each recipe should do its best to guard against making a breaking change due to invalid type information.  

I took another look at the `UseDiamondOperators` recipe and was able to guard it against NewClass initializers associated with VariableDeclarations having a null or unknown type.

Are you able to provide some more information or create an issue for the `CleanupMockitoImports` problems your encountering?

pway99:
revert due to accidental committed to the wrong branch

hetzijzo:
I have the following two Java files in a project

```java
import lombok.data;

@Data
class MyObject {
  String someField;
}
```

```java
import static org.mockito.Mockito.when;

import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.mockito.InjectMocks;
import org.mockito.Mock;
import org.mockito.junit.jupiter.MockitoExtension;

@ExtendWith(MockitoExtension.class)
class MyObjectTest {
  @Mock
  MyObject myObject;

  @Test
  void test() {
    when(myObject.getSomeField()).thenReturn("testValue");
    ......
  }
}
```

Now when I run the following command it will lead to a unwanted deleted static import of `import static org.mockito.Mockito.when`. The project will no longer compile. 
The reason for the delete is that the Lombok annotation processor will normally add the getter & setter of the `someField` which is not detected or taken care of by the AST of openrewrite. 


```
[...]

PR #4596, opened 2024-10-21
enhancing Java parsers to recover method invocation type attribution when arg types are missing

nmck257:
(if there's only one valid signature matching the method name and arg count)

## What's changed?

see subject

## What's your motivation?

Lombok support (#1297) is presumably still far away. In the interim, this PR attempts to "patch over" one hole in type attribution which Lombok repos can introduce.

For example, I frequently work with ( / refactor) code like this:

```java
MyCoolBuilder.builder()
.propA(...)
.propB(...)
.propC(...)
.fancyConfig(...)
```

Sometimes developers will supply those props from a separate Properties class with getters from Lombok, eg, `.propA(lombokPropertyClass.getPropA())`. OpenRewrite won't know the type of the getter, because Lombok. That cascades to OpenRewrite not knowing the type of the `propA(...)` method invocation. That cascades to OpenRewrite not knowing the type of the `select` for the next method, and so on. If I want to refactor `fancyConfig(...)`, then even if those arg types *are* known, the `select` is unknown, and method matchers miss.

- Conceptually, if we add support for annotation processors like Lombok in the future, then I believe this *should* become dead code as we *should* have "perfect" type attribution in the first place. But, I think it's worthwhile until then?
- Known gap: if the relevant methodInvocation is being called without a select, then I don't see a sturdy way to identify the implicit receiver, so basically this new code just doesn't apply
- I haven't touched the Java Parsers before; I expect by the time I'm done, I'd just need to copy-paste this new source bit (in JDK 8-friendly syntax) to each of the 4 parsers?
- Not sure if we'd want to do any preemptive integration test for LST generation before merging something like this?

maybe @sambsnyd , @knutwannheden 

- Wait for Lombok / annotation processor / code gen support
- Do nothing

## Any additional context

It's a draft because I haven't ported to the other JDK versions yet (or tested much beyond the unit tests), and figured I'd solicit feedback first in case there were directional concerns.

### Checklist

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -21,6 +21,7 @@ import com.sun.source.tree.*;
 import com.sun.source.util.TreePathScanner;
 import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.tree.DocCommentTable;
 import com.sun.tools.javac.tree.EndPosTable;
 import com.sun.tools.javac.tree.JCTree;
@@ -933,8 +934,27 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
         Symbol methodSymbol = (jcSelect instanceof JCFieldAccess) ? ((JCFieldAccess) jcSelect).sym :
                 ((JCIdent) jcSelect).sym;
 
+        Type jcSelectType = jcSelect.type;
+        if (jcSelectType == null && jcSelect instanceof JCFieldAccess &&
+                hasOneMatchingMethod(((JCFieldAccess) jcSelect).selected.type, name.getSimpleName(), node.getArguments().size())) {
+            jcSelectType = methodSymbol.type;
+        }
+
         return new J.MethodInvocation(randomId(), fmt, Markers.EMPTY, select, typeParams, name, args,
-                typeMapping.methodInvocationType(jcSelect.type, methodSymbol));
+                typeMapping.methodInvocationType(jcSelectType, methodSymbol));
+    }
+
+    private boolean hasOneMatchingMethod(Type methodSelectType, String name, int argCount) {
+        Symbol methodSymbol = null;
+        Type classType = methodSelectType;
+        while (!classType.equals(Type.noType)) {
+            for (var sym : classType.tsym.members().getSymbols(sym -> sym.name.toString().equals(name) && ((Symbol.MethodSymbol) sym).params().size() == argCount)) {
+                if (methodSymbol != null) return false;
+                methodSymbol = sym;
+            }
+            classType = ((Type.ClassType) classType).supertype_field;
+        }
+        return methodSymbol != null;
     }
 
     @Override
--- a/rewrite-java-tck/build.gradle.kts
+++ b/rewrite-java-tck/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
 
 dependencies {
     implementation("org.assertj:assertj-core:latest.release")
+    implementation("org.junit-pioneer:junit-pioneer:2.0.0")
     implementation(project(":rewrite-java"))
     implementation(project(":rewrite-java-test"))
     implementation(project(":rewrite-test"))
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodInvocationTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/MethodInvocationTest.java
@@ -15,10 +15,13 @@
  */
 package org.openrewrite.java.tree;
 
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.test.RewriteTest;
+import org.openrewrite.test.TypeValidation;
 
 import java.util.Optional;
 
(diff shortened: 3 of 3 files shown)


traceyyoshima:
Hi Nick, I doubt processors will be introduced any time soon since annotation processing is triggered in `JavaCompiler#compile` instead of `JavaCompiler#parse`. It might be simpler to consider saving `RetentionPolicy.SOURCE` annotations during type attribution or making an exception for `lombok` types.

Saving the `SOURCE` annotations would preserve the `lombok` annotations directly on the types, which could be handled by recipes. Technically, recipes could be used to generate LST elements based on the annotations, and the `JavaTemplate` could be used to set the appropriate types as well. It's also likely faster and more efficient since recipes are so lightweight compared to compilation.

timtebeek:
Thanks @nmck257 ! Not sure if you've spoken to Sam about this effort already, but you've at least inspired 
https://github.com/openrewrite/rewrite/compare/main...lombok-spike . Let's see what the outcome of those efforts is as well.

nmck257:
yeah, we've had a few pings across mismatched working hours. the `lombok-spike` branch does seem promising, and if I understand it right, would remove the need for this PR :)

knutwannheden:
The work there has hardly been tested at all yet. There is basically only a simple test for `@Getter` and one for `@Builder`. Adding the `@Builder` test revealed a few missing pieces, so I suspect there could be more of that.

I have checked that I can run `./gradlew pTML`, so if anyone wants to give this a try (only implemented for `rewrite-java-17` for now), that would be very appreciated.

traceyyoshima:
Or Sam will classload the annotation processor and call the relevant methods, instead, haha. I won't have time this week, but I can throw most of the parsing together over the weekend.

nmck257:
> I have checked that I can run `./gradlew pTML`, so if anyone wants to give this a try (only implemented for `rewrite-java-17` for now), that would be very appreciated.

Getting this exception:

> [ERROR] Failed to execute goal org.openrewrite.maven:rewrite-maven-plugin:5.43.0-SNAPSHOT:run (default-cli) on project grt-gimfm-security-utils: Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:5.43.0-SNAPSHOT:run failed: Unable to find runtime dependencies beginning with: 'lombok', classpath: [...]

I assume lombok is just a compile-only dependency for `rewrite-java`, so either that would need to change, or use the `classpathFromResources` pattern.

knutwannheden:
Thanks for taking a look! @sambsnyd might have some idea or insights.

nmck257:
if it's useful with minimal context, here's an error I got when parsing a file which uses `@Slfj` but doesn't have any other obvious lombok interaction (with the `lombok-spike` branch).

The field it was parsing is the first occurrence of `Map` in the file (outside imports).

```
[DEBUG]
org.openrewrite.java.JavaParsingException: Failed to convert for the following cursor stack:--- BEGIN PATH ---
JCCompilationUnit(sourceFile = /redacted/Foobar.java)
JCClassDecl(name = Foobar, line = 19)
JCVariableDecl(name = foobarVariable, line = 30)
JCTypeApply(line = 31)
--- END PATH ---

    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1705)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1711)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertAll (ReloadableJava17ParserVisitor.java:1749)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertTypeParameters (ReloadableJava17ParserVisitor.java:1766)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitParameterizedType (ReloadableJava17ParserVisitor.java:1134)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitParameterizedType (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCTypeApply.accept (JCTree.java:2728)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariables (ReloadableJava17ParserVisitor.java:1568)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariable (ReloadableJava17ParserVisitor.java:1528)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariable (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCVariableDecl.accept (JCTree.java:1045)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1711)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertStatements (ReloadableJava17ParserVisitor.java:1822)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertStatements (ReloadableJava17ParserVisitor.java:1799)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitClass (ReloadableJava17ParserVisitor.java:507)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitClass (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCClassDecl.accept (JCTree.java:860)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertAll (ReloadableJava17ParserVisitor.java:1735)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitCompilationUnit (ReloadableJava17ParserVisitor.java:552)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitCompilationUnit (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept (JCTree.java:614)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17Parser.lambda$parseInputs$1 (ReloadableJava17Parser.java:189)
[..]
Caused by: java.lang.StringIndexOutOfBoundsException: begin 2485, end 2485, length 2480
    at java.lang.String.checkBoundsBeginEnd (String.java:4608)
    at java.lang.String.substring (String.java:2711)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1673)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1711)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertAll (ReloadableJava17ParserVisitor.java:1749)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertTypeParameters (ReloadableJava17ParserVisitor.java:1766)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitParameterizedType (ReloadableJava17ParserVisitor.java:1134)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitParameterizedType (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCTypeApply.accept (JCTree.java:2728)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariables (ReloadableJava17ParserVisitor.java:1568)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariable (ReloadableJava17ParserVisitor.java:1528)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitVariable (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCVariableDecl.accept (JCTree.java:1045)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1711)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertStatements (ReloadableJava17ParserVisitor.java:1822)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertStatements (ReloadableJava17ParserVisitor.java:1799)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitClass (ReloadableJava17ParserVisitor.java:507)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitClass (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCClassDecl.accept (JCTree.java:860)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert (ReloadableJava17ParserVisitor.java:1675)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convertAll (ReloadableJava17ParserVisitor.java:1735)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitCompilationUnit (ReloadableJava17ParserVisitor.java:552)
    at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.visitCompilationUnit (ReloadableJava17ParserVisitor.java:73)
    at com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept (JCTree.java:614)
    at com.sun.source.util.TreePathScanner.scan (TreePathScanner.java:86)
    at org.openrewrite.java.isolated.ReloadableJava17Parser.lambda$parseInputs$1 (ReloadableJava17Parser.java:189)
[...]
```

traceyyoshima:
TLDR: I quickly looked, and some Lombok annotations get pretty ugly. There are ways to work around the complexity by preserving the original tree, but I don't want to touch it. Jon, Knut, or Sam will find a clean solution.

Info:

Annotations that generate statements such as `@Synchronized`, `@SneakyThrows`, `@Cleanup` introduce code that isn't detectable from the JCTree via Symbols.

Before:
```java
import lombok.Cleanup;
import java.io.*;
            
public class CleanupExample {
    public static void main(String[] args) throws IOException {
        @Cleanup InputStream in = new FileInputStream(args[0]);
        @Cleanup OutputStream out = new FileOutputStream(args[1]);
        byte[] b = new byte[10000];
        while (true) {
            int r = in.read(b);
            if (r == -1) break;
            out.write(b, 0, r);
        }
    }
}
```

After:
```java
import lombok.Cleanup;
import java.io.*;

public class CleanupExample {
    
    public CleanupExample() {
        super();
    }
    
    public static void main(String[] args) throws IOException {
        @Cleanup
        InputStream in = new FileInputStream(args[0]);
        try {
            @Cleanup
            OutputStream out = new FileOutputStream(args[1]);
            try {
                byte[] b = new byte[10000];
                while (true) {
                    int r = in.read(b);
                    if (r == -1) break;
                    out.write(b, 0, r);
                }
            } finally {
                if (java.util.Collections.singletonList(out).get(0) != null) {
                    out.close();
                }
            }
        } finally {
            if (java.util.Collections.singletonList(in).get(0) != null) {
                in.close();
            }
        }
    }
}
```

Lombok doesn't account for unnecessary annotations, which adds additional complexity:

Before:
```java
import lombok.Cleanup;
import java.io.*;

public class CleanupExample {

    public CleanupExample() {
        super();
    }

    public static void main(String[] args) throws IOException {
        @Cleanup
        InputStream in = new FileInputStream(args[0]);
        try {
            @Cleanup
            OutputStream out = new FileOutputStream(args[1]);
            try {
                byte[] b = new byte[10000];
                while (true) {
                    int r = in.read(b);
                    if (r == -1) break;
                    out.write(b, 0, r);
                }
            } finally {
                if (java.util.Collections.singletonList(out).get(0) != null) {
                    out.close();
                }
            }
        } finally {
            if (java.util.Collections.singletonList(in).get(0) != null) {
                in.close();
            }
        }
    }
}
```

After:
```java

import lombok.Cleanup;
import java.io.*;

public class CleanupExample {
    
    public CleanupExample() {
        super();
    }
    
    public static void main(String[] args) throws IOException {
        @Cleanup
        InputStream in = new FileInputStream(args[0]);
        try {
            try {
                @Cleanup
                OutputStream out = new FileOutputStream(args[1]);
                try {
                    try {
                        byte[] b = new byte[10000];
                        while (true) {
                            int r = in.read(b);
                            if (r == -1) break;
                            out.write(b, 0, r);
                        }
                    } finally {
                        if (java.util.Collections.singletonList(out).get(0) != null) {
                            out.close();
                        }
                    }
                } finally {
                    if (java.util.Collections.singletonList(out).get(0) != null) {
                        out.close();
                    }
                }
            } finally {
                if (java.util.Collections.singletonList(in).get(0) != null) {
                    in.close();
                }
            }
        } finally {
            if (java.util.Collections.singletonList(in).get(0) != null) {
                in.close();
            }
        }
    }
}
```

nmck257:
yeah, that Cleanup example is pretty grizzly -- I wonder if an interim option could be configuring the lombok annotation processor to simply ignore some of its more-troublesome annotations?

nmck257:
an "elegant" (if not performant) option might be compiling the files twice -- once without annotation processors and once with -- and visiting the non-annotation-processed outputs to produce the trees, but, using the processed version as a parallel reference point to re-populate missing types

traceyyoshima:
> an "elegant" (if not performant) option might be compiling the files twice -- once without annotation processors and once with -- and visiting the non-annotation-processed outputs to produce the trees, but, using the processed version as a parallel reference point to re-populate missing types

IIRC, the majority of parsing time is spent in the `enterAll`, which is impactful on large sources. I think there is a timeout on parsing sources, so I suspect calling `enterAll` twice will result in losing source files.

There's probably a relatively simple solution to parsing Lombok -- I'll think about it over the weekend, but I'm pretty sure Jon, Sam, or Knut will have a solution beforehand.

knutwannheden:
I investigated the Lombok code a bit and it looks like Lombok supports registering custom handlers for annotations, so the idea would be try to supply handlers for the problematic annotations when using Lombok programmatically. I will try this out next week if I find some time.

knutwannheden:
I found a way to disable the handlers for the problematic annotations. Luckily these typically don't add any new declarations, as we would then at least have to run them partially. But as we can now override handlers, that should also be feasible if necessary.

I didn't yet test all of Lombok's experimental annotations, but it looks pretty good now, AFAICT.

nmck257:
closing this in favor of the now-merged #4602; we can dig this back up in the future if we encounter another scenario which causes missing types where we don't see a path to fix it at the source

PR #4472, opened 2024-09-04
Add processing for nested modules with a parent version property

marcel-gepardec:
## What's changed?
Added the ability to process projects with modules in modules where the parent version is a property.

If the project's version does not match the GroupArtifactVersion GAV or cannot be resolved with getValue() because there is no property such as the version string, each parent POM should now be checked to see if the property is set.
If no parent, including the last parent, the root POM, has a property like this, then it can proceed as usual.
If the property is in the root POM and corresponds to the version of the project, this project can be returned and further processed.

## What's your motivation?
@ErhardSiegl noticed an incorrect behavior regarding the usage of openrewrite in a project with modules in modules where the parent version is a property. He always got this error message:
```
[ERROR] Failed to execute goal org.openrewrite.maven:rewrite-maven-plugin:5.39.2:run (default-cli) on project test-common-shared: Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:5.39.2:run failed: Error while visiting pom.xml: java.lang.IllegalArgumentException: Illegal character in path at index 66: file:///Users/test/.m2/repository/com/example/exatest/test-common/${revision}/test-common-${revision}.pom
[ERROR]   java.base/java.net.URI.create(URI.java:883)
[ERROR]   org.openrewrite.maven.internal.MavenPomDownloader.download(MavenPomDownloader.java:539)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom$Resolver.resolveParentPom(ResolvedPom.java:500)
[ERROR]  org.openrewrite.maven.tree.ResolvedPom$Resolver.resolveParentPropertiesAndRepositoriesRecursively(ResolvedPom.java:407)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom$Resolver.resolveParentsRecursively(ResolvedPom.java:365)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom$Resolver.resolve(ResolvedPom.java:354)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom.resolve(ResolvedPom.java:176)
[ERROR]   org.openrewrite.maven.UpdateMavenModel.updateResult(UpdateMavenModel.java:162)
[ERROR]   org.openrewrite.maven.UpdateMavenModel.lambda$updateResult$8(UpdateMavenModel.java:167)
[ERROR]   org.openrewrite.internal.ListUtils.map(ListUtils.java:176)
[...]

PR #4598, opened 2024-10-22
Fix infinite loop in MavenPomDownloader

philippe-granet:
## What's changed?
Fix an infinite loop in MavenPomDownloader

## What's your motivation?
Since this MR https://github.com/openrewrite/rewrite/pull/4472 , there is an infinite loop in `MavenPomDownloader.download()` when root pom have a parent.

- https://github.com/openrewrite/rewrite/pull/4472

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -501,7 +501,7 @@ public class MavenPomDownloader {
                     return projectPom;
                 }
                 Parent parent = projectPom.getParent();
-                while (parent != null){
+                if (parent != null){
                     for (Pom project : projectPoms.values()) {
                         if (parent.getGroupId().equals(project.getGroupId()) && parent.getArtifactId().equals(project.getArtifactId())){
                             if (projectPom.getVersion().equals(project.getValue(gav.getVersion()))){


ErhardSiegl:
I think this fix would break projects that use a revision property and where the module structure is more than one level deep. I assume @marcel-gepardec has an example for that. The code wants to traverse to the root-pom, but apparently the condition for reaching the root is not good enough.

timtebeek:
Thanks for the quick work here @philippe-granet ; I'll give @marcel-gepardec a little time to chime in, but ideally we work this in before tomorrow's release.

nmck257:
> Thanks for the quick work here @philippe-granet ; I'll give @marcel-gepardec a little time to chime in, but ideally we work this in before tomorrow's release.

I also ran into this infinite loop with some local testing :)
imo, if we don't land on a root fix in time for that release, then #4472 should just be reverted for the release until a thorough fix is available.
(but I *am* grateful for the intended enhancement; I've tried to fix the `${revision}` pattern in the past and it keeps coming back to haunt me)

ErhardSiegl:
@marcel-gepardec wont be available today before noon CET. Its unlikely that he can provide a fix today especially since the code is really complicated imho. In my opinion the code would need a more in depth rework.

timtebeek:
Thanks for that context @ErhardSiegl ; then I propose we merge the change already, given that the same tests still pass, it fixes an issue for two folks (at least), and might only break when the module structure is deeper. That last case can then be picked up separately with a matching test.

A test for the scenarios @philippe-granet & @nmck257 encountered would also be helpful to avoid regressions at that point, but let's not hold up this fix and the release for that. Thanks all!

nmck257:
well this is odd -- one of the tests from that PR (`shouldNotThrowExceptionForModulesInModulesWithRightProperty`) is *already* failing on my machine at `main`, but I see CI is happy.

I'm on windows; Java URI syntax isn't OS-specific?...

```
org.opentest4j.AssertionFailedError: Unexpected exception thrown: java.lang.IllegalArgumentException: Illegal character in path at index 52: file:///C:/Users/me/.m2/repository/test/test2/${test}/test2-${test}.pom

	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)
	at org.junit.jupiter.api.AssertDoesNotThrow.createAssertionFailedError(AssertDoesNotThrow.java:84)
	at org.junit.jupiter.api.AssertDoesNotThrow.assertDoesNotThrow(AssertDoesNotThrow.java:75)
	at org.junit.jupiter.api.AssertDoesNotThrow.assertDoesNotThrow(AssertDoesNotThrow.java:58)
	at org.junit.jupiter.api.Assertions.assertDoesNotThrow(Assertions.java:3259)
	at org.openrewrite.maven.internal.MavenPomDownloaderTest$WithNativeHttpURLConnectionAndTLS.shouldNotThrowExceptionForModulesInModulesWithRightProperty(MavenPomDownloaderTest.java:675)
Caused by: java.lang.IllegalArgumentException: Illegal character in path at index 52: file:///C:/Users/me/.m2/repository/test/test2/${test}/test2-${test}.pom
	at java.base/java.net.URI.create(URI.java:906)
	at org.openrewrite.maven.internal.MavenPomDownloader.download(MavenPomDownloader.java:560)
	at org.openrewrite.maven.internal.MavenPomDownloaderTest$WithNativeHttpURLConnectionAndTLS.lambda$shouldNotThrowExceptionForModulesInModulesWithRightProperty$5(MavenPomDownloaderTest.java:675)
	at org.junit.jupiter.api.AssertDoesNotThrow.assertDoesNotThrow(AssertDoesNotThrow.java:71)
	... 7 more
Caused by: java.net.URISyntaxException: Illegal character in path at index 52: file:///C:/Users/me/.m2/repository/test/test2/${test}/test2-${test}.pom
	at java.base/java.net.URI$Parser.fail(URI.java:2974)
	at java.base/java.net.URI$Parser.checkChars(URI.java:3145)
	at java.base/java.net.URI$Parser.parseHierarchical(URI.java:3227)
	at java.base/java.net.URI$Parser.parse(URI.java:3175)
	at java.base/java.net.URI.<init>(URI.java:623)
	at java.base/java.net.URI.create(URI.java:904)
	... 10 more
```

marcel-gepardec:
@nmck257 @timtebeek I fixed the issue with the infinity loop and the failing test in #4609

PR #4609, opened 2024-10-24
Fix failing test and infinity loop in MavenPomDownloader

marcel-gepardec:
I'm really sorry for solving the problem so late. @philippe-granet 
- https://github.com/openrewrite/rewrite/pull/4598
tried to fix the infinity loop with an if statement. That's also probably why one of the tests mentioned by @nmck257 failed.

## What's changed?
I fixed the problem by implementing a method called resolveGav(). This uses the already available mergeProperties method to see if the property for the version is located in one of the modules.

## What's your motivation?
Since it was my own fault 
- https://github.com/openrewrite/rewrite/pull/4472

and the attempt to fix this problem led to a failing test 
- https://github.com/openrewrite/rewrite/pull/4598.

My sincere apologies for producing such a catastrophe.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -172,10 +172,10 @@ public class MavenPomDownloader {
 
     private Map<GroupArtifactVersion, Pom> projectPomsByGav(Map<Path, Pom> projectPoms) {
         Map<GroupArtifactVersion, Pom> result = new HashMap<>();
-        for (final Pom projectPom : projectPoms.values()) {
-            final List<Pom> ancestryWithinProject = getAncestryWithinProject(projectPom, projectPoms);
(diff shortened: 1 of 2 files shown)


timtebeek:
Hi @marcel-gepardec ; No worries about the earlier failures! :) Thanks for coming back here with a proposed fix. I do notice that there's still some failures in the tests; did you already explore those?

```
MavenParserTest > multipleCiFriendlyVersionPlaceholders() FAILED
    java.lang.IllegalArgumentException: Illegal character in path at index 57: file:///home/runner/.m2/repository/bogus/example/parent/${revision}${changelist}/parent-${revision}${changelist}.pom
        at java.base/java.net.URI.create(URI.java:906)
        at org.openrewrite.maven.internal.MavenPomDownloader.download(MavenPomDownloader.java:554)
```

timtebeek:
I've applied a few more simplifications; I hope you all agree! Now we immediately return if there's an exact match without replacements, or when there's a match after replacements. Only when neither of those cases is true do we continue to download.

PR #4610, opened 2024-10-25
Add Recipe that allows for the removal/negation of gitignore rules.

Jenson3210:
## What's changed?
Added a recipe

## What's your motivation?
First iteration of this seems to work well.

@timtebeek as discussed

FindAndReplaceText is very restrictive on this. This really uses gitignore rules/logic to determine if a change to the gitignore is necessary.

## Any additional context
I've added a workaround for https://github.com/openrewrite/jgit/issues/3
In a next iteration, I want to remove this workaround again once jgit is fixed but guessing we will have some longer checks there. 
Planning on 2 later iterations: 
- removeEmptyGitignoreFile boolean option that removes a file if the last gitignore entry was removed or it only contains comments/empty lines after changes.
- support wildcard filepaths or directories with wildcard as input: This should match all files matching and for each of these files calculate same negations. Instead of iteration over paths variable, we iterate over all files matching the paths wildcard variables.

### Checklist
- [X] I've added unit tests to cover both positive and negative cases (A LOT 😄 )

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
+++ b/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
@@ -63,7 +63,6 @@ public class ExcludeFileFromGitignore extends ScanningRecipe<Repository> {
             @Override
             public PlainText visitText(PlainText text, ExecutionContext ctx) {
                 try {
-                    text.getSourcePath().getParent()
                     String gitignoreFileName = text.getSourcePath().toString();
                     gitignoreFileName = gitignoreFileName.startsWith("/") ? gitignoreFileName : "/" + gitignoreFileName;
                     IgnoreNode ignoreNode = new IgnoreNode();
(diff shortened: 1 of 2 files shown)


Jenson3210:
Thanks @timtebeek, did not get the time for the suggestions yet today. Would've done it otherwise :) 

About this build failure... 
Looks like it got introduced by this suggestion?
Screenshot 2024-10-25 at 14 08 15

Anyhow pushed a change that uses ctx

Jenson3210:
@timtebeek, I am done with final polishing. 
I've refactored the accumulator to no longer need the workaround. I now decide myself per rule if I need to do pathMatch as true or false and no longer rely on the IngoreNode's internal true default value which is not always correct. 

As I identified that the jgit lib is working fine when doing a TreeWalk, I considered probably in there, they do a similar thing bypassing ignoreRule. 
As jgit is not really meant for recipe usage, I would not considere it a bug perse, rather tried fixing it cleaner internally. 

I think i did a pretty good job (pat on the self-shoulder 😉 )

next to that I also added some logic to insert the new negating lines as close to possible with the ignoring lines. 
It's still possible if comments are there that the line is splitted from the other one by a comment, but at least not all added at eof.

timtebeek (approved):
Great to see you've worked this out; hope this helps you unignore those .jar files ahead of further changes.

PR #4627, opened 2024-10-29
Check that we have downloaded the file successfully

lkerford:
When we are downloading files we should check that the http call was successful instead of assuming that because no exception was thrown that the http status was successful. In the case were we get a 408, we don't throw an exception. This results in a InputStream of null to be returned. When reading from a InputStream of null, it infinitely returns a list of null bytes. In our use case we end up writing null to a temporary file until we run out of disk space

We can also make this change in the `HttpSender` and throw when we get a none successful http code

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java
+++ b/rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java
@@ -93,7 +93,11 @@ public class RemoteArchive implements Remote {
             Path localArchive = cache.compute(uri, () -> {
                 //noinspection resource
                 HttpSender.Response response = httpSender.send(httpSender.get(uri.toString()).build());
-                return response.getBody();
+                if (response.isSuccessful()) {
+                    return response.getBody();
+                } else {
+                    throw new IllegalStateException("Failed to download " + uri + " to artifact cache got an " + response.getCode());
+                }
             }, ctx.getOnError());
 
             if (localArchive == null) {
--- a/rewrite-core/src/test/java/org/openrewrite/remote/RemoteArchiveTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/remote/RemoteArchiveTest.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.remote;
 
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.ExecutionContext;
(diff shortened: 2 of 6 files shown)


natedanner (approved):

rewrite-core/src/test/java/org/openrewrite/remote/RemoteArchiveTest.java
```diff
@@ -52,6 +54,27 @@ void gradleWrapper(String version) throws Exception {
...
+            Paths.get("gradle/wrapper/gradle-wrapper.jar"),
+            distributionUrl.toURI()
+          )
+          .build("gradle-[^\\/]+\\/(?:.*\\/)+gradle-wrapper-(?!shared).*\\.jar");
+
+
+        assertThat(
+          assertThrows(IllegalStateException.class, () -> getInputStreamSize(remoteArchive.getInputStream(ctx)))
+            .getMessage()
+        ).contains("Failed to download");
```
timtebeek, line 75: This nested JUnit inside AssertJ reads a bit odd to me; can't we use AssertJ's `assertThatThrownBy` to directly assert the message?

sambsnyd (commented):
I agree with Tim's feedback about asserting exceptions but otherwise this looks good

rewrite-core/src/test/java/org/openrewrite/remote/RemoteArchiveTest.java
lkerford: `assertThatThrownBy` is much better fit I will update

lkerford:
I'm also going to update the test to each have their own cache, because at the moment only the 1st test running the expect logic while the others are falling back to the cache and aren't testing the expected logic

rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java
```diff
@@ -93,7 +93,11 @@ public InputStream getInputStream(ExecutionContext ctx) {
             Path localArchive = cache.compute(uri, () -> {
                 //noinspection resource
                 HttpSender.Response response = httpSender.send(httpSender.get(uri.toString()).build());
-                return response.getBody();
+                if (response.isSuccessful()) {
```
zieka, line 96: question to team:
should we be doing some type of retry here to improve reliability with transient errors like network glitch or delays?

lkerford: If want to implement retires, I think we might want to move this change to the `HttpSender`, I suspect other endpoints could benefit from it

sjungling (approved):

timtebeek (approved):

PR #4629, opened 2024-10-30
Subprojects is potentially null

Laurens-W:
## What's changed?

Add testcase that disproves this.

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/internal/RawPomTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/internal/RawPomTest.java
@@ -117,7 +117,7 @@ class RawPomTest {
     }
 
     @Test
-    void modulesAndSubProjectsEmpty() {
+    void modulesAndSubProjectsNull() {
         RawPom pom = RawPom.parse(
           //language=xml
           new ByteArrayInputStream("""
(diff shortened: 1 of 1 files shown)


Laurens-W:
https://github.com/openrewrite/rewrite/blob/ed1a481737fe7526aaabab3759613bc9f8580828/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java#L584-L595
This method which is used when `toPom` is called takes care of any null conditions, every later field definition has and `emptyList` default:
* https://github.com/openrewrite/rewrite/blob/d030c63c15bba6ddbf03f9f10f73f74d2e3c1e7e/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java#L118-L119
* https://github.com/openrewrite/rewrite/blob/d030c63c15bba6ddbf03f9f10f73f74d2e3c1e7e/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java#L117-L119

Constructors either use `emptyList` or the value passed all the way from the `RawPom` through the `toPom` into `Pom` and then `ResolvedPom`

Laurens-W:
Resolved in 
* https://github.com/openrewrite/rewrite/pull/4630

PR #4639, opened 2024-11-01
Upstream RemoveMethodInvocationsVisitor

nmck257:
## What's changed?

This visitor existed in rewrite-spring; I'm moving it over here (with some minor bugfixes)
https://github.com/openrewrite/rewrite-spring/blob/b0a95c850c521ebf6104c946f044ef9c1c43bd5b/src/main/java/org/openrewrite/java/spring/RemoveMethodInvocationsVisitor.java#L39

## What's your motivation?

This visitor had a bug with long invocation chains which include args which are themselves long invocation chains. Also some NPE risks. I wanted to fix those, and I figure there's nothing specific to rewrite-spring in here so it makes sense to give broader reach.
It could probably also be wrapped into a recipe for convenient consumption in the future.

I added two new test cases:
- `removeFromWithinArguments` (verifies the bugfix I cared about)
- `removeWithoutSelect` (annotated as expected to fail; something I noticed along the way. my src changes convert this test case from uncaught NPE to no-changes)

~Next week I'll probably deprecate all the original code in rewrite-spring and adjust the impls to pass through to this class.

I could apply the same fixes in-place in rewrite-spring, if preferred

## Any additional context

I took care to preserve git history from rewrite-spring for these files, so I'd love to not squash the PR on merge

### Checklist

### DIFF
--- a/rewrite-java/build.gradle.kts
+++ b/rewrite-java/build.gradle.kts
@@ -64,6 +64,7 @@ dependencies {
     testRuntimeOnly(project(":rewrite-java-17"))
     testImplementation("com.tngtech.archunit:archunit:1.0.1")
     testImplementation("com.tngtech.archunit:archunit-junit5:1.0.1")
+    testImplementation("org.junit-pioneer:junit-pioneer:2.0.0")
 
     // For use in ClassGraphTypeMappingTest
     testRuntimeOnly("org.eclipse.persistence:org.eclipse.persistence.core:3.0.2")
--- a/rewrite-java/src/main/java/org/openrewrite/java/RemoveMethodInvocationsVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/RemoveMethodInvocationsVisitor.java
@@ -99,7 +99,7 @@ public class RemoveMethodInvocationsVisitor extends JavaVisitor<ExecutionContext
                     } else if (isLambdaBody) {
                         return ToBeRemoved.withMarker(J.Block.createEmptyBlock());
                     } else {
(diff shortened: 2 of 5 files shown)


rewrite-java/build.gradle.kts
```diff
@@ -64,6 +64,7 @@ dependencies {
     testRuntimeOnly(project(":rewrite-java-17"))
     testImplementation("com.tngtech.archunit:archunit:1.0.1")
     testImplementation("com.tngtech.archunit:archunit-junit5:1.0.1")
+    testImplementation("org.junit-pioneer:junit-pioneer:2.0.0")
```
sambsnyd, line 67: I would rather use JUnit 5's built in assertThrows() unless there is some really compelling advantage to `@ExpectedToFail`

nmck257: I've seen it in other modules, and I like the semantics of it for documenting a not-yet-implemented feature with a test.
But yeah, it does require another dep, and we can get similar functionality with assertThrows. I can swap it out when I'm at my desk later.

knutwannheden: I also quite like `@ExpectedToFail` for tests which test something that is supposed to work but doesn't due to a bug. I use `assertThrows()` when I want to test that something is supposed to throw an exception.

When the bug is fixed, all I have to do is to remove the `@ExpectedToFail` annotation.

nmck257: sounds like @sambsnyd is outvoted here, unless I hear otherwise :)
any other review concerns for this PR, before we rebase-not-squash and merge?
@timtebeek , @knutwannheden

PR #4640, opened 2024-11-02
Optimize `AdaptiveRadixTree`

knutwannheden:
Benchmark results:

```
Benchmark                                                      Mode  Cnt         Score   Error   Units
JavaTypeCacheBenchmark.readAdaptiveRadix                      thrpt    2       674.861           ops/s
JavaTypeCacheBenchmark.readAdaptiveRadix:gc.alloc.rate        thrpt    2      2009.346          MB/sec
JavaTypeCacheBenchmark.readAdaptiveRadix:gc.alloc.rate.norm   thrpt    2   3122809.227            B/op
JavaTypeCacheBenchmark.readAdaptiveRadix:gc.count             thrpt    2       112.000          counts
JavaTypeCacheBenchmark.readAdaptiveRadix:gc.time              thrpt    2        49.000              ms
JavaTypeCacheBenchmark.readSnappy                             thrpt    2       332.215           ops/s
JavaTypeCacheBenchmark.readSnappy:gc.alloc.rate               thrpt    2      3192.919          MB/sec
JavaTypeCacheBenchmark.readSnappy:gc.alloc.rate.norm          thrpt    2  10082058.492            B/op
JavaTypeCacheBenchmark.readSnappy:gc.count                    thrpt    2       179.000          counts
JavaTypeCacheBenchmark.readSnappy:gc.time                     thrpt    2        78.000              ms
JavaTypeCacheBenchmark.writeAdaptiveRadix                     thrpt    2       535.267           ops/s
JavaTypeCacheBenchmark.writeAdaptiveRadix:gc.alloc.rate       thrpt    2      4515.371          MB/sec
JavaTypeCacheBenchmark.writeAdaptiveRadix:gc.alloc.rate.norm  thrpt    2   8848657.549            B/op
JavaTypeCacheBenchmark.writeAdaptiveRadix:gc.count            thrpt    2       176.000          counts
JavaTypeCacheBenchmark.writeAdaptiveRadix:gc.time             thrpt    2       235.000              ms
JavaTypeCacheBenchmark.writeSnappy                            thrpt    2       313.278           ops/s
JavaTypeCacheBenchmark.writeSnappy:gc.alloc.rate              thrpt    2      3414.677          MB/sec
JavaTypeCacheBenchmark.writeSnappy:gc.alloc.rate.norm         thrpt    2  11435420.340            B/op
JavaTypeCacheBenchmark.writeSnappy:gc.count                   thrpt    2       133.000          counts
JavaTypeCacheBenchmark.writeSnappy:gc.time                    thrpt    2       174.000              ms
```

Additionally, here is the memory footprint for the two implementations:

```
Retained AdaptiveRadixTree size:    8425616 bytes
Retained Snappy size:               8613760 bytes
```


knutwannheden:
As can be seen reads are over 80% faster and writes about 50% faster. At the same time the allocations for reads drop by around 70%. For reads only around 5%. Finally, the memory footprint is now almost exactly the same (even slightly lower), so the implementation looks like it is better across the board. An additional bonus is that the implementation doesn't depend on Snappy.

PR #4643, opened 2024-11-04
Merge `SpringTypeReferenceProvider` into `SpringTypeReference`

knutwannheden:
The PR also includes a few more renames and also adds a prototype for `FindTypes`.


knutwannheden:
@Laurens-W I renamed `TypeReference.TypeReferenceProvider` to `TypeReference.Provider` and the same for `SpringTypeReferenceProvider`. What do you think?

Laurens-W:
> @Laurens-W I renamed `TypeReference.TypeReferenceProvider` to `TypeReference.Provider` and the same for `SpringTypeReferenceProvider`. What do you think?

Makes sense, no need to duplicate the name since it's an inner now

PR #4648, opened 2024-11-05
Update `ChangeType` and `ChangePackage` to work with `SourceFileWithReference`

Laurens-W:
## What's changed?

Added implementation for `SourceFileWithTypeReference` in `ChangeType` and `ChangePackage` recipes

Test coverage (edge cases)

@knutwannheden @timtebeek 

## Any additional context

- https://github.com/openrewrite/rewrite/pull/4587

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -349,4 +382,24 @@ private boolean isTargetRecursivePackageName(String packageName) {
...
+        Set<TypeReference> matches;
+        String oldPackageName;
+        String newPackageName;
+
+        @Override
+        public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+            Tree tree1 = super.visit(tree, ctx);
+            for (TypeReference ref : matches) {
+                if (ref.getTree().equals(tree) && ref.supportsRename()) {
+                    return ref.renameTo(ref.getName().replace(oldPackageName, newPackageName)).visit(tree, ctx);
```
Laurens-W, line 398: Not sure if we should rely on replacing the entire reference or whether we actually need a `renamePackageTo` method on the `TypeReference` interface

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -113,14 +116,45 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (tree instanceof JavaSourceFile) {
+                    return new JavaChangePackageVisitor().visit(tree, ctx);
+                } else if (tree instanceof SourceFileWithTypeReferences) {
+                    SourceFileWithTypeReferences sourceFile = (SourceFileWithTypeReferences) tree;
+                    SourceFileWithTypeReferences.TypeReferences typeReferences = sourceFile.getTypeReferences();
+                    Set<TypeReference> matches = new HashSet<>();
+                    for (TypeReference ref : typeReferences.getTypeReferences()) {
+                        if (ref.getName().startsWith(oldPackageName)) {
```
knutwannheden, line 146: Do we need to already here check for `recursive` and make sure the ref also has a dot after the package name?

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeReference.java
```diff
@@ -52,6 +50,27 @@ public String getName() {
...
+                if (tree instanceof Xml.Tag) {
+                    return ((Xml.Tag) tree).withValue(name);
+                }
+                return super.visit(tree, executionContext);
+            }
+        };
+    }
+
     static class Matcher extends SimpleTraitMatcher<SpringTypeReference> {
         private final Pattern typeReference = Pattern.compile("(?:[a-zA-Z_][a-zA-Z0-9_]*\\.)+[A-Z*][a-zA-Z0-9_]*(?:<[a-zA-Z0-9_,?<> ]*>)?");
```
Laurens-W, line 75: The regex isnot fully covering just yet: 
image

knutwannheden: Yes, we might indeed have to cover generics, too. Otherwise this would also work: `"\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(?:\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*"`.

knutwannheden (commented):
Looking good already!

I added some more review comments. Most of them probably apply to both recipes.

rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java
```diff
@@ -26,6 +24,14 @@ public interface TypeReference extends Trait<Tree> {
 
     String getName();
 
+    default boolean supportsRename() {
+        return false;
+    }
+
+    default TreeVisitor<Tree, ExecutionContext> renameTo(String name) {
+        throw new UnsupportedOperationException();
+    }
```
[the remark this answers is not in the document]
timtebeek: Then lets keep it like this; with the Annotated trait I was not able to use that in all cases as I lacked access to the context, if I recall correctly. Better then not to limit ourselves here.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -113,14 +116,48 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                     }
                     stopAfterPreVisit();
                 }
+                if (tree instanceof SourceFileWithTypeReferences) {
+                    SourceFileWithTypeReferences cu = (SourceFileWithTypeReferences) tree;
+                    boolean recursive = Boolean.TRUE.equals(ChangePackage.this.recursive);
+                    String recursivePackageNamePrefix = oldPackageName + ".";
+                    for (TypeReference ref : cu.getTypeReferences().getTypeReferences()) {
+                        if (ref.getName().equals(oldPackageName) || recursive && ref.getName().startsWith(recursivePackageNamePrefix)) {
```
knutwannheden, line 124: Ideally I think we should only be using the `matches()` and `renameTo()` methods instead of the `getName()` method, but coming up with a good API for those two methods feels pretty difficult, so maybe this will have to wait. I was thinking something along the lines of Java's `Matcher` for regex:

image

knutwannheden: One example I keep thinking about here is when we have `META-INF/services` files as we have them for the `TypeReference.Provider` interface. In this case the name of a nested type looks like `foo.Bar$Baz`. Now `ChangeType` can't possibly know that it needs to call `renameTo("foo.Bar$Baz")` and not `renameTo("foo.Bar.Baz")`. The latter is probably correct for some other config files. Currently, the `ChangeType` visitor would have to make sure to always pass in names like `"foo.Bar$Baz"` and then the `TypeReference` implementation would have to know if it should apply a heuristic (although `Bar$Baz` could even be the name of the top-level class, but that would be highly unlikely). The question is just what should the API look like to cater for this?

rewrite-core/src/main/java/org/openrewrite/trait/TypeReference.java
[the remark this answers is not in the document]
knutwannheden: Agreed. It would just be good if we could get the API finalized before people try to start using it.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -113,14 +113,52 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+        return Preconditions.check(condition, new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                return sourceFile instanceof JavaSourceFile || sourceFile instanceof SourceFileWithTypeReferences;
+            }
+
+            @Override
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (tree instanceof JavaSourceFile) {
+                    return new JavaChangePackageVisitor().visit(tree, ctx);
```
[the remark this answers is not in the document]
knutwannheden: Ideally, yes.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeReference.java
Laurens-W: I had never seen `\p{javaJavaIdentifierStart}` before but it seems that matches what we want to match here much better!

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    interface Provider {
+
+        Set<Reference> getTypeReferences(SourceFile sourceFile);
+
+        boolean isAcceptable(SourceFile sourceFile);
+    }
+
+    interface MatcherMutator extends Matcher, Renamer {
```
Laurens-W, line 54: Not sure how I feel about this aggregating interface

rewrite-core/src/main/java/org/openrewrite/TypeReferenceProvider.java
```diff
@@ -15,13 +15,13 @@
  */
 package org.openrewrite;
 
-import org.openrewrite.trait.TypeReference;
+import org.openrewrite.trait.reference.Reference;
```
knutwannheden, line 18: Do we typically have subpackages for traits? Unless we add more types to this package, I think we can keep it all in the `trait` package.

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
```diff
@@ -0,0 +1,70 @@
...
+        return matcher.matchesReference(this);
+    }
+
+    default TreeVisitor<Tree, ExecutionContext>  rename(Renamer renamer, String replacement) {
+        return renamer.rename(replacement);
+    }
+
+    interface Provider {
+
+        Set<Reference> getTypeReferences(SourceFile sourceFile);
```
knutwannheden, line 49: Now that the references are typed, should we rename this method?
```suggestion
        Set<Reference> getReferences(SourceFile sourceFile);
```

rewrite-core/src/main/java/org/openrewrite/TypeReferenceProvider.java
knutwannheden, line 1: Isn't this a "relic"? AFAICT we replaced this with `TypeReference.Provider` (now `Reference.Provider`).

knutwannheden (commented):
First batch of comments.

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
timtebeek: Wouldn't it suffice to have `TypeMatcher` and `PackageMatcher` each `implements Reference.Matcher, Reference.Renamer` and drop the separate `MatcherMutator`?

Looks like the usage sites could then just use `PackageMatcher` and `TypeMatcher` directly without the need for further changes or an intermediate interface (at least for now).

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -113,14 +116,45 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                     }
                     stopAfterPreVisit();
                 }
+                if (tree instanceof SourceFileWithReferences) {
```
timtebeek, line 119:
```suggestion
                } else if (tree instanceof SourceFileWithReferences) {
```

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -113,14 +116,45 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                    SourceFileWithReferences cu = (SourceFileWithReferences) tree;
+                    boolean recursive = Boolean.TRUE.equals(ChangePackage.this.recursive);
+                    String recursivePackageNamePrefix = oldPackageName + ".";
+                    for (Reference ref : cu.getReferences().getReferences()) {
+                        if (ref.getKind().equals(Reference.Kind.PACKAGE) && ref.getValue().equals(oldPackageName) || recursive && ref.getValue().startsWith(recursivePackageNamePrefix)) {
```
timtebeek, line 124: Did you mean? Without these parentheses it looks like we might match Kind.TYPE against a recursive prefix.
```suggestion
                        if (ref.getKind().equals(Reference.Kind.PACKAGE) && 
                           (ref.getValue().equals(oldPackageName) || recursive && ref.getValue().startsWith(recursivePackageNamePrefix))) {
```

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -81,22 +82,45 @@ public String getDescription() {
...
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
                 if (tree instanceof JavaSourceFile) {
-                    JavaSourceFile cu = (JavaSourceFile) requireNonNull(tree);
+                    JavaSourceFile cu = (JavaSourceFile) tree;
                     if (!Boolean.TRUE.equals(ignoreDefinition) && containsClassDefinition(cu, oldFullyQualifiedTypeName)) {
                         return SearchResult.found(cu);
                     }
                     return new UsesType<>(oldFullyQualifiedTypeName, true).visitNonNull(cu, ctx);
                 }
+                if (tree instanceof SourceFileWithReferences) {
```
timtebeek, line 93:
```suggestion
                } else if (tree instanceof SourceFileWithReferences) {
```

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+    private String targetPackage;
+
+    @Getter
+    private final Boolean recursive;
+
+    public PackageMatcher(@Nullable String fieldType) {
+        this(fieldType, false);
+    }
+
+    public PackageMatcher(@Nullable String fieldType, @Nullable boolean recursive) {
```
timtebeek, line 40: Try to resolve any warnings you see.
image
```suggestion
    public PackageMatcher(@Nullable String fieldType, @Nullable Boolean recursive) {
```

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.trait.reference.Reference;
+import org.openrewrite.xml.tree.Xml;
+
+@Getter
+public class PackageMatcher implements Reference.MatcherMutator {
+
+    @Nullable
+    private String targetPackage;
```
timtebeek, line 31:
```suggestion
    private final @Nullable String targetPackage;
```

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Getter
+    private final Boolean recursive;
+
+    public PackageMatcher(@Nullable String fieldType) {
+        this(fieldType, false);
+    }
+
+    public PackageMatcher(@Nullable String fieldType, @Nullable boolean recursive) {
+        this.targetPackage = fieldType;
```
timtebeek, line 41: Let's rename the the parameter to match the field, or vice versa, for clarity.

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (StringUtils.isNotEmpty(newValue)) {
+                    if (tree instanceof Xml.Attribute) {
+                        return ((Xml.Attribute) tree).withValue(((Xml.Attribute) tree).getValue().withValue(getReplacement(((Xml.Attribute) tree).getValueAsString(), targetPackage, newValue)));
+                    }
+                    if (tree instanceof Xml.Tag) {
+                        if (((Xml.Tag) tree).getValue().isPresent()) {
+                            return ((Xml.Tag) tree).withValue(getReplacement(((Xml.Tag) tree).getValue().get(), targetPackage, newValue));
```
timtebeek, line 65: There's a nullability warning here on `targetPackage`; Please resolve as the getReplacement method does not allow a nullable argument here.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
```diff
@@ -30,16 +28,22 @@
 import java.util.regex.Pattern;
 
 @Value
-class SpringTypeReference implements TypeReference {
+class SpringTypeOrPackageReference implements Reference {
```
timtebeek, line 31: With `Kind` being a property of `Reference` I wonder if we should have `TypeOrPackage` in the name here, or opt for `SpringReference` instead.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
```diff
@@ -52,45 +56,56 @@ public String getName() {
...
+                    if (tag.getValue().isPresent() && typeReference.matcher(stringVal).matches()) {
+                        return new SpringTypeOrPackageReference(cursor, determineKind(stringVal));
                     }
                 }
             }
             return null;
         }
+
+        Kind determineKind(String value) {
+            return Character.isUpperCase(value.charAt(value.lastIndexOf('.')+1)) ? Kind.TYPE : Kind.PACKAGE;
```
timtebeek, line 94:
```suggestion
            return Character.isUpperCase(value.charAt(value.lastIndexOf('.') + 1)) ? Kind.TYPE : Kind.PACKAGE;
```

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
```diff
@@ -52,45 +56,56 @@ public String getName() {
...
 
-    static class Matcher extends SimpleTraitMatcher<SpringTypeReference> {
-        private final Pattern typeReference = Pattern.compile("(?:[a-zA-Z_][a-zA-Z0-9_]*\\.)+[A-Z*][a-zA-Z0-9_]*(?:<[a-zA-Z0-9_,?<> ]*>)?");
+    @Override
+    public boolean supportsRename() {
+        return true;
+    }
+
+    static class Matcher extends SimpleTraitMatcher<SpringTypeOrPackageReference> {
+        private final Pattern typeReference = Pattern.compile("\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(?:\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*");
```
timtebeek, line 65: Should we rename this field to just `reference`? (also further down).
```suggestion
        private final Pattern reference = Pattern.compile("\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(?:\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*");
```

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
knutwannheden: Let's also handle the primitive types case (no dot). While it wouldn't throw any exception, it would return `TYPE`.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
knutwannheden: Not related to Tim's comment: Looking at https://www.springframework.org/schema/beans/spring-beans-4.3.xsd, I think we should also match the `key-type` and `value-type` attributes in some context in any case.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
```diff
@@ -52,45 +56,56 @@ public String getName() {
...
                     }
                 }
             } else if (value instanceof Xml.Tag) {
                 Xml.Tag tag = (Xml.Tag) value;
                 if (tags.matches(cursor)) {
-                    if (tag.getValue().isPresent() && typeReference.matcher(tag.getValue().get()).matches()) {
-                        return new SpringTypeReference(cursor);
+                    String stringVal = tag.getValue().get();
+                    if (tag.getValue().isPresent() && typeReference.matcher(stringVal).matches()) {
+                        return new SpringTypeOrPackageReference(cursor, determineKind(stringVal));
```
knutwannheden:
```suggestion
                    Optional<String> stringVal = tag.getValue();
                    if (stringVal.isPresent() && typeReference.matcher(stringVal.get()).matches()) {
                        return new SpringTypeOrPackageReference(cursor, determineKind(stringVal.get()));
```

rewrite-java/src/main/java/org/openrewrite/java/TypeMatcher.java
```diff
@@ -112,7 +116,26 @@ private static boolean isPlainIdentifier(MethodSignatureParser.TargetTypePattern
     }
 
     @Override
-    public boolean matchesName(String name) {
-        return matchesTargetTypeName(name);
+    public boolean matchesReference(Reference reference) {
+        return matchesTargetTypeName(reference.getValue());
```
knutwannheden, line 120: Shouldn't we check the kind here?

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+        this(fieldType, false);
+    }
+
+    public PackageMatcher(@Nullable String fieldType, @Nullable boolean recursive) {
+        this.targetPackage = fieldType;
+        this.recursive = recursive;
+    }
+
+    @Override
+    public boolean matchesReference(Reference reference) {
```
knutwannheden, line 46: Shouldn't we check the kind here?

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
```diff
@@ -0,0 +1,70 @@
...
+    interface Matcher {
+        boolean matchesReference(Reference value);
+    }
+
+    interface Renamer {
+        default TreeVisitor<Tree, ExecutionContext> rename(String replacement) {
+            return rename(replacement, false);
+        }
+
+        default TreeVisitor<Tree, ExecutionContext> rename(String replacement, boolean recursive) {
```
knutwannheden: Currently we don't use this overload nor the `recursive` parameter anywhere.

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+package org.openrewrite.trait.reference;
+
+import org.openrewrite.*;
+import org.openrewrite.trait.Trait;
+
+import java.util.Set;
+
+@Incubating(since = "8.39.0")
+public interface Reference extends Trait<Tree> {
```
knutwannheden: I get a warning on the `Tree` type parameter here, as we don't have a `pacakge-info.java` with Jspecify annotations.

rewrite-core/src/main/java/org/openrewrite/SourceFileWithReferences.java
```diff
@@ -0,0 +1,66 @@
...
+        private final SourceFile sourceFile;
+        private final Set<Reference> references;
+
+        public Collection<Reference> findMatches(Reference.Matcher matcher) {
+            return findMatchesInternal(matcher, null);
+        }
+
+        public Collection<Reference> findMatches(Reference.Matcher matcher, Reference.Kind kind) {
+            return findMatchesInternal(matcher, kind);
+        }
```
knutwannheden, line 43: I think we might as well simplify the API by removing this method. As the caller supplies the matcher that matcher can filter on the kind.
```suggestion
```

Instead, we might want to provide something like a `containsMatches()` returning a `boolean`. But this isn't high priority.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -527,6 +551,25 @@ private boolean hasNoConflictingImport(@Nullable JavaSourceFile sf) {
...
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class ReferenceChangeTypeVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Set<Reference> matches;
+        Reference.Renamer renamer;
+        String newFullyQualifiedName;
+
+        @Override
+        public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
+            Tree tree1 = super.visit(tree, ctx);
```
knutwannheden, line 563: This call looks suspicious to me, as it calls `visit()` rather than `preVisit()`. I don't think you need to call `super.preVisit()` anyway. You can just return `tree` in the else case.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -349,4 +383,24 @@ private boolean isTargetRecursivePackageName(String packageName) {
...
+
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class ReferenceChangePackageVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Set<Reference> matches;
+        Reference.Renamer renamer;
+        String newPackageName;
+
+        @Override
+        public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 395: In `ReferenceChangeTypeVisitor` we override `preVisit()`. That (or `postVisit()`) is generally preferred when applicable.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -349,4 +383,24 @@ private boolean isTargetRecursivePackageName(String packageName) {
         }
 
     }
+
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class ReferenceChangePackageVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Set<Reference> matches;
```
knutwannheden, line 390: I would instead declare a field like this:
```suggestion
        Map<Tree, Reference> matches;
```
The way that `Tree` is required to override `equals()` and `hashCode()` you can then call `get()` on the map.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
Laurens-W: I made a mistake in my train of thought, as `Type` also contains a package, which should be changed here, so I'm removing the `ref.getKind().equals(Reference.Kind.PACKAGE)` statement I added

rewrite-java/src/main/java/org/openrewrite/java/PackageMatcher.java
Laurens-W: I'm not sure if this field will stay final in the long run, but for now it can be!

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringTypeOrPackageReference.java
Laurens-W: In Springs case this is handled implicitly, see https://docs.spring.io/spring-framework/reference/core/validation/convert.html and https://docs.spring.io/spring-framework/reference/core/validation/beans-beans.html#beans-beans-conversion

rewrite-core/src/main/java/org/openrewrite/trait/reference/Reference.java
Laurens-W: No longer an issue after moving the reference trait up a package

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
Laurens-W: I don't really know what happened here but the tests were failing because of mismatching super calls. Removed the super calls now!

rewrite-xml/src/test/java/org/openrewrite/xml/trait/SpringReferenceTest.java
```diff
@@ -67,13 +70,16 @@ void xmlConfiguration() {
                   <property name="age" value="10"/>
                   <property name="sibling">
                       <bean <!--~~(org.springframework.beans.TestBean)~~>-->class="org.springframework.beans.TestBean">
-                          <property name="age" value="11" <!--~~(java.lang.Integer)~~>-->class="java.lang.Integer"/>
+                          <property name="age" value="11" <!--~~(java.lang.int)~~>-->class="java.lang.int"/>
```
timtebeek, line 73: This one's a bit odd; was that intentional?

Laurens-W: Nope, accidentally left that in, good find!

knutwannheden: What is interesting is that we can't really add comments in front of attributes like this. It renders the XML invalid. But I suspect that is acceptable.

Laurens-W: The only purpose of the test is to verify the right elements are identified as type reference

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -112,15 +114,48 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                         }
                     }
                     stopAfterPreVisit();
```
knutwannheden, line 116: The stopAfterPreVisit() method should be called before the return statement, so that the tree visit is aborted.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -112,15 +114,48 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
 
-        return Preconditions.check(condition, new ChangePackageVisitor());
+        return Preconditions.check(condition, new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                return sourceFile instanceof JavaSourceFile || sourceFile instanceof SourceFileWithReferences;
+            }
+
+            @Override
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 139: Also here we should make sure to call stopAfterPreVisit() once we have found everything we are after. Otherwise the visitor will continue to traverse the entire tree.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -349,4 +384,22 @@ private boolean isTargetRecursivePackageName(String packageName) {
...
+
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class ReferenceChangePackageVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Map<Tree, Reference> matches;
+        Reference.Renamer renamer;
+        String newPackageName;
+
+        @Override
+        public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 396: Same here again.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -81,22 +82,47 @@ public String getDescription() {
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         TreeVisitor<?, ExecutionContext> condition = new TreeVisitor<Tree, ExecutionContext>() {
             @Override
-            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 85: And here.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -81,22 +82,47 @@ public String getDescription() {
...
 
-        return Preconditions.check(condition, new ChangeTypeVisitor(oldFullyQualifiedTypeName, newFullyQualifiedTypeName, ignoreDefinition));
+        return Preconditions.check(condition, new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                return sourceFile instanceof JavaSourceFile || sourceFile instanceof SourceFileWithReferences;
+            }
+
+            @Override
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 107: And here.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -527,6 +553,23 @@ private boolean hasNoConflictingImport(@Nullable JavaSourceFile sf) {
...
 
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class ReferenceChangeTypeVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Map<Tree, Reference> matches;
+        Reference.Renamer renamer;
+        String newFullyQualifiedName;
+
+        @Override
+        public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 564: One more.

knutwannheden (approved):
I might have missed some, but I added a comment regarding the preVisit() methods. I suggest you take a look at TreeVisitor#visit() so you understand how this mechanism works. When all you want to do is extract some information from the top-level node (or also some other node) using the preVisit() method, you should call stopAfterPreVisit() before returning.

knutwannheden:
If we can go through all visitors and make sure preVisit() is implemented as appropriate, I think we should merge the PR.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
knutwannheden: In this case I probably didn't look carefully enough. I suspect here we need to visit the entire tree.

knutwannheden (approved):

PR #4649, opened 2024-11-05
reworked ExcludeFileFromGitignore to support leading and trailing wildcards

Jenson3210:
## What's changed?
There's a bug in the currect version with `**/path/` entries. This should be fixed by adding the support + tests.

## What's your motivation?
Fix the bug + add wildcard support. We do not support wildcards in the middle part of a path, nor entries with multiple wildcarded parts. Also only the * wildcard for now is supported. Not the `?` yet. 

@lkerford , I thought I said I am DONE with gitignore and NEVER want to see them again... Yet here we go ;)

### Checklist

Note: when I run the best practices, I also receive changes in lots of other files, so only kept these applicable on my 2 changed files.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
+++ b/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
@@ -152,48 +152,47 @@ public class ExcludeFileFromGitignore extends ScanningRecipe<Repository> {
                     .sorted(comparingInt(String::length).reversed())
                     .collect(toList());
 
-            IgnoreNode.MatchResult isIgnored;
             for (String impactingFile : impactingFiles) {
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
```diff
@@ -268,4 +220,190 @@ private IgnoreNode.MatchResult isIgnored(IgnoreNode ignoreNode, String path) {
             return CHECK_PARENT;
         }
     }
+
+    private static class IgnoreRule {
```
lkerford, line 224: I do wonder if we should be applying these changes in our fork of jGit instead, but I think that given we are also making changes to how we want the changes written this is fine

lkerford (approved):

PR #4651, opened 2024-11-06
Update `GitRemote#normalize` to support organizations prefixed with a number

bryceatmoderne:
## What's changed?
`GitRemote#normalize` was incorrectly identifying an organization  starting with a number, such `1org`, as a port number. These changes update the regex `Pattern` to match a numerical string up to a forward slash.

## What's your motivation?
Some repositories such as https://github.com/1Remote/1Remote were failing to build using the Moderne CLI.

No

@pstreef 

No

## Any additional context
No

### Checklist


rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -279,7 +279,7 @@ private String repositoryPath(RemoteServerMatch match, URI normalizedUri) {
                     .replaceFirst("^/", "");
         }
 
-        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+");
+        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+/");
```
pstreef, line 282: what if the org is a number 🤔

bryceatmoderne: Well at least github supports number-only orgs. (https://github.com/1234)

timtebeek (approved):

timtebeek (approved):

PR #4653, opened 2024-11-07
Use `indexOfNextNonWhitespace()` rather than `getStartPosition()`

knutwannheden:
Lombok messes with the positions of the AST elements and as a result they don't always align with the source text.


jevanlingen:
Superseded by:
- https://github.com/openrewrite/rewrite/pull/5010

Issue #4654, opened 2024-11-07
OpenRewrite fails with "Illegal state while comparing versions" when using internal dependencies with non-SemVer versions


kadirbekov:
**Hello,**

I'm encountering an issue with OpenRewrite when attempting to upgrade to Spring Boot 3.3 using the UpgradeSpringBoot_3_3 recipe. Our project includes internal company dependencies that use version numbers not following Semantic Versioning (SemVer), such as **ORC-246-1-SNAPSHOT**. This causes OpenRewrite to fail with an exception when comparing versions.

**Environment:**
OpenRewrite Gradle Plugin Version: 6.26.0
OpenRewrite Spring Recipe Version: 5.22.0
Build Tool: Gradle
Gradle Version: gradle-7.6
Java Version: JDK 17
Operating System: (macOS Sequoia)

**Build Script Snippets:**
build.gradle file:
```

plugins {
    id 'java'
    id 'org.springframework.boot' version '3.0.1'
    id 'io.spring.dependency-management' version '1.1.0'
    id 'org.hibernate.orm' version '6.1.6.Final'
    id 'org.openrewrite.rewrite' version '6.26.0'
}

dependencies {
    // Other dependencies...

    implementation "com.veon.eurasia.oraculum:jira-api:ORC-246-1-SNAPSHOT"
    rewrite("org.openrewrite.recipe:rewrite-spring:5.22.0")
}

rewrite {
    activeRecipe("org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_3")
}
```

**Problem Description:**
When running the rewriteRun task, we receive the following error:

```

Caused by: java.lang.IllegalStateException: Illegal state while comparing versions: [ORC.0.0.0-246.0.0.0] and [9.0.0]. Metadata = [9.0.0]
    at org.openrewrite.semver.LatestRelease.compare(LatestRelease.java:164)
    at org.openrewrite.semver.VersionComparator.upgrade(VersionComparator.java:45)
    at org.openrewrite.gradle.DependencyVersionSelector.select(DependencyVersionSelector.java:142)
    at org.openrewrite.gradle.DependencyVersionSelector.select(DependencyVersionSelector.java:83)
    ...
Caused by: java.lang.IllegalStateException: No match found
```
[logs.txt](https://github.com/user-attachments/files/17660966/logs.txt)

**Analysis:**
[...]

PR #4656, opened 2024-11-08
Catch non-semver compatible exception in case of version update

Jenson3210:
See the slack exception logs -> https://rewriteoss.slack.com/archives/C01A843MWG5/p1730974519581749

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -262,7 +262,7 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
                                         acc.versionPropNameToGA.put(versionVariableName, ga);
                                         acc.gaToNewVersion.put(ga, resolvedVersion);
                                     }
-                                } catch (IllegalStateException | MavenDownloadingException e) {
+                                } catch (MavenDownloadingException e) {
                                     acc.gaToNewVersion.put(ga, e);
                                 }
                             }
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -262,7 +262,7 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                                         acc.versionPropNameToGA.put(versionVariableName, ga);
                                         acc.gaToNewVersion.put(ga, resolvedVersion);
                                     }
-                                } catch (MavenDownloadingException e) {
+                                } catch (IllegalStateException | MavenDownloadingException e) {
```
shanman190, line 265: This probably should instead be fixed here:
https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FDependencyVersionSelector.java#L168

The rationale is that the version is being compared and should really be skipped as exotic. The `DependencyVersionSelector` is used in several spots all of which _could_ encounter this exact same issue. By moving this fix to the location in the link, we'll fix all of the recipes in one fell swoop.

Jenson3210: I did consider that also. However, I thought "If you use the versionSelector, you **know** that  a version is selectable".

What would you return if no version could be selected?

Jenson3210: @shanman190 moved the check to the proposed position!

Jenson3210:
Fixes #4654

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
shanman190: Sweet!

Yeah, the version selector is simply trying to answer the question of: "is there a version to which we can upgrade to?" At which point, the version selector should return only the value to be upgraded to or null to indicate that there was no upgrade available or possible. In particular, with versions like the one from the issue, there's less of a natural order (outside of ASCII) in order to ensure that we are moving up the version strictly. This appears as well with the old Spring Cloud release trains where they were using the alpha ordering of train stations across Europe to order their releases (ie. Angel, Brixton, Camden, etc).

PR #4657, opened 2024-11-08
Feature - YAML search - find properties with a specific scalar value

aamotharald:
## What's changed?
I have changed the org.openrewrite.yaml.search.FindProperty search recipe to allow searching for yaml properties which have a specific scalar value. The new reciped parameter propertyValue is OPTIONAL which ensures downwards-compatibility.

## What's your motivation?
 I need this search functionality for a specific use case I am working on.
 I want to alter a yamle file if the specified yaml key does have a certain value. The search recipe can act as a Precondition here.

I thried to create a dedicated org.openrewrite.yaml.search.FindPropertyWithValue recipe inheriting from org.openrewrite.yaml.search.FindProperty, but recipes are final and do not allow for inheritance. This is why I added an optional parameter to the existing recipe instead.
Also I did not find out how to test a negative search case. I marked hence the negative test as `@Disabled`.

Alternative: Use SourceGraph to find the repos having the specified Yaml condition. Well then I can also run the change with Sourcegraph.

## Any additional context
I love the generic yaml stuff you have

### Checklist
- [x] I've added unit tests to cover both positive and negative cases (well comment about the negative test case)

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/search/FindProperty.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/search/FindProperty.java
@@ -48,6 +48,7 @@ public class FindProperty extends Recipe {
     Boolean relaxedBinding;
 
     @Option(displayName = "Property value",
+            example = "false",
             description = "If provided, only properties specified in propertyKey having this value will be found. Works only for scalar values",
             required = false)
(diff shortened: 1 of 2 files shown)


sambsnyd:
Thanks @aamotharald !

PR #4658, opened 2024-11-08
Analyse escaping for local vars

MBoegers:
## What's changed?
I added two visitors that help decide if variables escape and where.

## What's your motivation?
When migrating existing code, the question "Is this instance ever leaking its defining scope?" raises from time to time.
Take the migration from `StringBuffer` to `StringBuilder`.
`StringBuilder` is generally in favor, if no synchronization is needed (see [JavaDoc StringBuilder](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html)).
Synchronization is not needed if the instance never escapes its defining scope.
These helping visitors enable such migration by answering the question of escaping vars.

Additional ways to leak an instance.

I considered the Dataflow API, but I am not fluent enough to archive the same..

## Any additional context
This visitor and its helper methods will help @mkarg to realize his migrations for the OpenJDK project.

### Checklist


shanman190:
Not specifically related to this PR, but thought that I'd share a bit simpler data flow API example given I too found rewrite-security-frameworks a bit difficult to follow originally. I ended up using the data flow API over in rewrite-feature-flags to great effect.

https://github.com/openrewrite/rewrite-feature-flags/blob/main/src/main/java/org/openrewrite/featureflags/search/FindFeatureFlag.java#L99-L112

MBoegers:
Thanks @shanman190 for the reference, I'll definitely have a look. I think the Dataflow API needs some introduction material. After I had a breakthrough I'll share some ;)

mkarg:
> Thanks @shanman190 for the reference, I'll definitely have a look. I think the Dataflow API needs some introduction material. After I had a breakthrough I'll share some ;)

Does that imply that this PR will change or that it will get revoked? I am actively waiting for this visitors to be finally usable. Thanks for updating.

mkarg:
I personally have no strong feelings *how* the visitors are implemented as long as there is a stable API to call them.

rewrite-java/src/test/java/org/openrewrite/java/search/FindEscapingVariablesTest.java
```diff
@@ -0,0 +1,453 @@
...
+                        void test() {
+                            StringBuilder sb = new StringBuilder();
+                            StringBuilder other = new StringBuilder(sb != null ? sb : new StringBuilder());
+                        }
+                    }
+                    """, """
+                    package com.sample;
+                    public class Foo{
+                        void test() {
+                            StringBuilder /*~~>*/sb = new StringBuilder();
```
mkarg, line 259: As `other` is itself *non-escaping*, `sb` essentially is *non-escpaing*, too.

rewrite-java/src/test/java/org/openrewrite/java/search/FindEscapingVariablesTest.java
```diff
@@ -0,0 +1,453 @@
...
+                        void test() {
+                            StringBuilder sb = new StringBuilder();
+                            StringBuilder other = new StringBuilder(sb);
+                        }
+                    }
+                    """, """
+                    package com.sample;
+                    public class Foo{
+                        void test() {
+                            StringBuilder /*~~>*/sb = new StringBuilder();
```
mkarg, line 143: As `other` is itself *non-escaping*, `sb` essentially is *non-escpaing*, too.

mkarg:
Any progress?

mkarg:
@MBoegers Too bad that now that your are employed with Moderne you did not find the time to finish this really essential contribution.

PR #4659, opened 2024-11-11
Keep multiple whitelines and same comments from the gitignore file

Jenson3210:
As I originally used a Set, the empty lines and comments with same text where being filtered to a single occurence (the last one).

By changing the implementation to a LinkedList into which we add conditionally entries, we can save a new List instantiation + keep comments and whitelines non-distinct, maintaining dinstinction of the gitgnore lines (same line later in file makes sure we only keep that one)

See the changed test for the bug I was facing while first time running on a gitignore file with multiple empty lines.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
+++ b/rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
@@ -103,7 +103,7 @@ public class ExcludeFileFromGitignore extends ScanningRecipe<Repository> {
             private List<String> sortRules(String[] originalRules, List<String> newRules) {
                 LinkedList<String> results = new LinkedList<>();
                 Arrays.stream(originalRules).filter(line -> {
-                    if (line.startsWith("#") || StringUtils.isBlank(line)) {
+                    if (StringUtils.isBlank(line) || line.startsWith("#")) {
                         return true;
                     }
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

rewrite-core/src/main/java/org/openrewrite/ExcludeFileFromGitignore.java
```diff
@@ -127,17 +127,18 @@ private List<String> sortRules(String[] originalRules, List<String> newRules) {
                 return distinctValuesStartingReversed(results);
             }
 
-            private <T> List<T> distinctValuesStartingReversed(List<T> list) {
-                Set<T> set = new LinkedHashSet<>();
-                ListIterator<T> iterator = list.listIterator(list.size());
+            private List<String> distinctValuesStartingReversed(List<String> list) {
+                LinkedList<String> filteredList = new LinkedList<>();
```
timtebeek, line 131: Is there any benefit to using LinkedList here over a regular ArrayList? Or can we move to ArrayList instead?
```suggestion
                List<String> filteredList = new ArrayList<>();
```

Jenson3210: I believe there is. 

`addFirst` is the main thing I believe. 
1. It’s easier to read VS adding on index 0
2. When inserting I believe internally not all objects need to be shifted 1 index. Might be wrong on that benefit ☺️

Issue #4660, opened 2024-11-11
Support JEP-441: Pattern Matching for switch


nielsdebruin:
## What problem are you trying to solve?
To support Java 21 recipes the parser need to be extended to ensure compatibility with JEP 440 (Record Patterns) and JEP 441 (Pattern Matching for switch).

## Describe the solution you'd like
Parsing logic should recognize and handle record patterns and switch pattern matching, for different situations and edge cases, which have already been outlined in the proposal.

## Additional context
* https://openjdk.org/jeps/441

timtebeek:
I've split off a separate
- https://github.com/openrewrite/rewrite/issues/4893

PR #4661, opened 2024-11-11
Support JEP-441: Pattern Matching for switch for Java 21+

nielsdebruin:
## What's changed?
Initial test cases for verifying parser support for JEP 440 and 441. The goal of these tests is to demonstrate that code containing JEP 440/441 structures remains unchanged. The tests very closely resemble the code snippets used in the JEPs themselves to describe their changes. They can be summarized as follows:

**Pattern Switch Statement**
Verifies that switch expressions using patterns for different object types (e.g., Integer, Long, Double, String) maintain structure after parsing.

**Pattern Matching for Records**
Tests parsing of instanceof with record deconstruction (e.g., if (obj instanceof Point(int x, int y))), ensuring that records can be deconstructed and match syntax remains unchanged.

**Nested Pattern Matching for Records and Enums**
Ensures that nested pattern matching with records containing other records and enums (e.g., Rectangle(ColoredPoint(Point p, Color c), ColoredPoint lr)) is correctly handled and maintains structure.
	
**null Case Handling in Switch Statements**
Checks that switch cases with null handling (e.g., case null ->)

**Improved Enum Support with Sealed Types**
Tests switch handling for sealed types and enums (e.g., case Foo.CLUBS, case Bar b) to ensure that sealed interfaces with enums and final classes maintain parsing integrity.

**Relaxed Type Restrictions in Switch Statements**
Verifies that switch cases with mixed types (e.g., String, Color, Point, int[]) correctly serialize and deserialize while preserving type information.

**Special Integer Cases in Switch**
Tests specific integer cases and when clauses (e.g., case -1, 1, case Integer j when (j - 1) > -1) in switch statements, ensuring all conditional logic remains intact after parsing.

## What's your motivation?
Provide support for JEP 440/441
- see #4660 for details 

### Checklist


knutwannheden:
Aren't some of the tests (like record pattern matching) for preview only features? So far we haven't supported any preview features and people haven't really asked for it either.

nielsdebruin:
As far as I understand it, both 440/441 have been delivered/closed (see status in [440](https://openjdk.org/jeps/440) and [441](https://openjdk.org/jeps/441)) and have been included in the release of Java 21 as **permanent** JEPs (source: https://www.jrebel.com/blog/java-21).

I double-checked by running some of the code snippets in these using Java `Termurin-21.0.4`, which worked fine.

knutwannheden:
I think you are right. I missed that!

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1067,8 +1067,19 @@ public Case withBody(J body) {
             return getPadding().withBody(JRightPadded.withElement(this.body, body));
         }
 
+        @Nullable
+        JContainer<CaseLabel> labels;
```
Laurens-W, line 1071: Changing this to a list of `JRightPadded` was mentioned but that brings along a load of changes. The expressions and statements fields are also JContainer so I feel like it's best to keep things in that line

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/SwitchPatternMatchingTest.java
```diff
@@ -0,0 +1,156 @@
...
+    void shouldParseJava21EnumSupportInSwitch() {
+        rewriteRun(
+          java(
+            //language=java
+            """
+              sealed interface Currency permits Coin {}
+              enum Coin implements Currency { HEADS, TAILS }
+              
+              class Test {
+                  void switchEnum(Coin c) {
```
timtebeek, line 77: Should we perhaps switch on `Currency` here? Or why was that sealed interface included above?
```suggestion
                  void switchEnum(Currency c) {
```

Laurens-W: The example was taken from here https://openjdk.org/jeps/441, not entirely sure why the sealed interface was used. I'm assuming defining just interface coin with it's 2 values would suffice?

timtebeek (commented):
Looks good to me, but it's hard for me to say if the `J.VariableDeclarations` now-also-implements `Expression` could have adverse effects downstream. Guess that's something to monitor as it's not something we can easily work around I suppose.

knutwannheden (commented):
I don't see any mention of that this PR doesn't yet support record patterns. I think we should make sure that this won't cause trouble down the line.

timtebeek:
> I don't see any mention of that this PR doesn't yet support record patterns. I think we should make sure that this won't cause trouble down the line.

Ah I'd split off pattern matching for records into a separate issue, which Laurens is already looking into; I'd also updated the title above, but not the description of this PR.
- https://github.com/openrewrite/rewrite/issues/4893

knutwannheden:
> Ah I'd split off pattern matching for records into a separate issue, which Laurens is already looking into; I'd also updated the title above, but not the description of this PR.

Makes sense. I missed that!

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -342,19 +342,25 @@ public J visitBreak(BreakTree node, Space fmt) {
...
         J.Case.Type type = node.getCaseKind() == CaseTree.CaseKind.RULE ? J.Case.Type.Rule : J.Case.Type.Statement;
+
+        JContainer<J> caselabels = JContainer.build(
+                node.getLabels().isEmpty() ? EMPTY : sourceBefore("case"),
+                node.getLabels().isEmpty() || node.getLabels().getFirst() instanceof DefaultCaseLabelTree ?
+                        List.of(JRightPadded.build(new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), skip("default"), null, null))) :
+                        convertAll(node.getLabels(), commaDelim, t -> EMPTY),
+                Markers.EMPTY
+        );
+        JContainer<Expression> expressions = JContainer.build(caselabels.getBefore(), caselabels.getElements().stream().filter(Expression.class::isInstance).map(Expression.class::cast).map(JRightPadded::build).toList(), caselabels.getMarkers());
```
Laurens-W, line 353: I don't like this part a single bit, but expressions are always part of the list of labels
```
        @Override @DefinedBy(Api.COMPILER_TREE)
        public List<JCExpression> getExpressions() {
            return labels.stream()
                         .filter(p -> p.hasTag(CONSTANTCASELABEL))
                         .map(p -> ((JCConstantCaseLabel) p).expr)
                         .collect(List.collector());
        }
```
so I think there is no other way (we can't reset the cursor and build the JContainer as we do for labels as we might run into elements we cannot parse) to compose the list of expressions

Laurens-W: The Java 17 implementations is even clearer:
```
        @Override @DefinedBy(Api.COMPILER_TREE)
        public List<JCExpression> getExpressions() { return labels.stream().filter(p -> p instanceof JCExpression).map(p -> (JCExpression) p).collect(List.collector()); }
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -779,12 +785,24 @@ public J visitInstanceOf(InstanceOfTree node, Space fmt) {
...
+            return new J.Identifier(randomId(), sourceBefore(b.getVariable().getName().toString()), Markers.EMPTY, emptyList(), b.getVariable().getName().toString(),
+                    type, typeMapping.variableType(b.var.sym));
+        } else if (pattern instanceof JCRecordPattern r) {
+            int saveCursor = cursor;
+            int endCursor = max(endPos(r), cursor);
+            cursor = endCursor;
+            return new J.Unknown(randomId(), whitespace(), Markers.EMPTY, new J.Unknown.Source(randomId(), whitespace(), Markers.EMPTY, source.substring(saveCursor, endCursor)));
+        } else {
+            return null;
+        }
```
knutwannheden, line 803: I can't find it now anymore, but I think I added a comment somewhere that we might be able to merge these two branches into one and only return `null` if the `pattern` is `null`.

What I find curious is that the `ParenthesizedPatternTree` only exists in the Java 17 compiler. Does that mean that parenthesized guards are not allowed in Java 21? Was that maybe only ever allowed in previews of this feature?

It seems like Java 21 instead has the `AnyPatternTree`: `if (r instanceof R(_)) {}`. See also https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-14.30.1

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -342,19 +342,25 @@ public J visitBreak(BreakTree node, Space fmt) {
     @Override
     public J visitCase(CaseTree node, Space fmt) {
         J.Case.Type type = node.getCaseKind() == CaseTree.CaseKind.RULE ? J.Case.Type.Rule : J.Case.Type.Statement;
+
+        JContainer<J> caselabels = JContainer.build(
+                node.getLabels().isEmpty() ? EMPTY : sourceBefore("case"),
+                node.getLabels().isEmpty() || node.getLabels().getFirst() instanceof DefaultCaseLabelTree ?
+                        List.of(JRightPadded.build(new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), skip("default"), null, null))) :
+                        convertAll(node.getLabels(), commaDelim, t -> EMPTY),
```
knutwannheden, line 350: I suppose the override for `visitDeconstructionPattern()` will still be required. Is the plan to add that in this PR?

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -342,19 +342,25 @@ public J visitBreak(BreakTree node, Space fmt) {
...
-                JContainer.build(
-                        node.getExpressions().isEmpty() ? EMPTY : sourceBefore("case"),
-                        node.getExpressions().isEmpty() ?
-                                List.of(JRightPadded.build(new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), skip("default"), null, null))) :
-                                convertAll(node.getExpressions(), commaDelim, t -> EMPTY),
-                        Markers.EMPTY
-                ),
+                expressions,
+                caselabels,
+                null,
```
knutwannheden, line 363: Why don't we have a `convert(node.getGuard())` here?

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
knutwannheden: I am wondering if we can't align the structure of `J.Case` so that we don't require a `expressions` field anymore and instead only store the `labels`.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
Laurens-W: Because Java 17 doesn't have a guard

knutwannheden: Makes sense! I thought I saw a `getGuard()` method, but I must have been looking at the Java 21 version of `CaseTree`.

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
Laurens-W: Not yet, this PR aims to provide correct modelling of switch pattern matching and print idempotency for Record pattern matching (by mapping that to J.Unknown)

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
Laurens-W: https://github.com/openrewrite/rewrite/pull/4900#discussion_r1914858017

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1027,14 +1027,27 @@ public Case withPattern(@Nullable Expression pattern) {
...
-            return expressions.getElements();
+            return caseLabels != null ? caseLabels.getElements().stream().filter(Expression.class::isInstance).map(Expression.class::cast).collect(toList()) : emptyList();
         }
 
         public Case withExpressions(List<Expression> expressions) {
-            return getPadding().withExpressions(requireNonNull(JContainer.withElementsNullable(this.expressions, expressions)));
+            if (caseLabels != null && caseLabels.getElements().stream().allMatch(Expression.class::isInstance)) {
+                return getPadding().withCaseLabels(requireNonNull(JContainer.withElementsNullable(this.caseLabels, expressions.stream().map(J.class::cast).collect(toList()))));
+            } else {
+                throw new IllegalStateException("caseLabels contains an entry that is not an Expression, use withCaseLabels instead.");
```
Laurens-W, line 1038: This exception will only occur when Java 21+ is involved, we do however need to find a way to convey to recipe authors that they should use `withCaseLabels` instead of  `withExpressions` when their recipe applies to Java 21.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1127,23 +1148,36 @@ public static class Padding {
...
             }
 
             public Case withExpressions(JContainer<Expression> expressions) {
-                return t.expressions == expressions ? t : new Case(t.id, t.prefix, t.markers, t.type, null, expressions, t.statements, t.body);
+                if (t.getExpressions() == expressions) {
+                    return t;
+                } else if (t.caseLabels != null && t.caseLabels.getElements().stream().allMatch(Expression.class::isInstance)) {
+                    return new Case(t.id, t.prefix, t.markers, t.type, null, expressions, null, t.guard, t.statements, t.body);
+                }
+                throw new IllegalStateException("caseLabels contains an entry that is not an Expression, use withCaseLabels instead.");
```
Laurens-W, line 1172: Same applies here https://github.com/openrewrite/rewrite/pull/4661/files#r1918837253

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1027,14 +1027,27 @@ public Case withPattern(@Nullable Expression pattern) {
...
-
         public List<Expression> getExpressions() {
-            return expressions.getElements();
+            return caseLabels != null ? caseLabels.getElements().stream().filter(Expression.class::isInstance).map(Expression.class::cast).collect(toList()) : emptyList();
         }
 
         public Case withExpressions(List<Expression> expressions) {
-            return getPadding().withExpressions(requireNonNull(JContainer.withElementsNullable(this.expressions, expressions)));
+            if (caseLabels != null && caseLabels.getElements().stream().allMatch(Expression.class::isInstance)) {
+                return getPadding().withCaseLabels(requireNonNull(JContainer.withElementsNullable(this.caseLabels, expressions.stream().map(J.class::cast).collect(toList()))));
```
knutwannheden, line 1036:
```suggestion
            if (caseLabels != null && caseLabels.getElements().stream().allMatch(Expression.class::isInstance)) {
                //noinspection unchecked
                return getPadding().withCaseLabels(requireNonNull(JContainer.withElementsNullable(this.caseLabels, (List<J>) (List<?>) expressions)));
```

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1067,17 +1080,25 @@ public Case withBody(J body) {
...
-        public Case(UUID id, Space prefix, Markers markers, Type type, @Deprecated @Nullable Expression pattern, JContainer<Expression> expressions, JContainer<Statement> statements, @Nullable JRightPadded<J> body) {
+        public Case(UUID id, Space prefix, Markers markers, Type type, @Deprecated @Nullable Expression pattern, @Nullable JContainer<Expression> expressions, @Nullable JContainer<J> caseLabels, @Nullable Expression guard, JContainer<Statement> statements, @Nullable JRightPadded<J> body) {
             this.id = id;
             this.prefix = prefix;
             this.markers = markers;
             this.type = type;
             if (pattern != null) {
-                this.expressions = requireNonNull(JContainer.withElementsNullable(null, singletonList(pattern)));
+                this.caseLabels = requireNonNull(JContainer.withElementsNullable(null, singletonList(pattern)));
+            } else if (expressions != null && !expressions.getElements().isEmpty()) {
```
knutwannheden, line 1096: I think we better leave out the second part so that we know that `caseLabels` will be set to a non-null value.
```suggestion
            } else if (expressions != null) {
```

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1027,14 +1027,27 @@ public Case withPattern(@Nullable Expression pattern) {
...
-            return getPadding().withExpressions(requireNonNull(JContainer.withElementsNullable(this.expressions, expressions)));
+            if (caseLabels != null && caseLabels.getElements().stream().allMatch(Expression.class::isInstance)) {
+                return getPadding().withCaseLabels(requireNonNull(JContainer.withElementsNullable(this.caseLabels, expressions.stream().map(J.class::cast).collect(toList()))));
+            } else {
+                throw new IllegalStateException("caseLabels contains an entry that is not an Expression, use withCaseLabels instead.");
+            }
+        }
+
+        @Nullable
+        JContainer<J> caseLabels;
```
knutwannheden, line 1043: In which cases is this field going to be `null`?

rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java
```diff
@@ -366,7 +366,7 @@ public J.Case visitCase(J.Case _case, J j) {
                 J.Case compareTo = (J.Case) j;
                 this.visitList(_case.getStatements(), compareTo.getStatements());
                 visit(_case.getBody(), compareTo.getBody());
-                this.visitList(_case.getExpressions(), compareTo.getExpressions());
+                this.visitList(_case.getCaseLabels(), compareTo.getCaseLabels());
```
knutwannheden, line 369: Here we should also compare the new guard.

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -423,7 +423,7 @@ public J visitCase(J.Case case_, P p) {
         } else {
             c = (J.Case) temp;
         }
-        c = c.getPadding().withExpressions(visitContainer(c.getPadding().getExpressions(), JContainer.Location.CASE_EXPRESSION, p));
+        c = c.getPadding().withCaseLabels(visitContainer(c.getPadding().getCaseLabels(), JContainer.Location.CASE_LABEL, p));
```
knutwannheden, line 426: Also here a visit of the guard is missing.

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -423,7 +423,8 @@ public J visitCase(J.Case case_, P p) {
...
+        c = c.withGuard((Expression) visitExpression(c.getGuard(), p));
```
knutwannheden, line 427: Please note that `visitExpression()` would be wrong to call here. This is a very special visitor method, which is only there to allow subclasses to implement a visit method that ends up getting called for every LST type implementing the `Expression` interface. By default it does nothing.

```suggestion
        c = c.withGuard(visitAndCast(c.getGuard(), p));
```

rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java
```diff
@@ -366,7 +366,8 @@ public J.Case visitCase(J.Case _case, J j) {
...
+                this.visitExpression(_case.getGuard(), compareTo.getGuard());
```
knutwannheden, line 370:
```suggestion
                if (_case.getGuard() != null && compareTo.getGuard() != null) {
                    visit(_case.getGuard(), compareTo.getGuard());
                } else if (nullMissMatch(_case.getGuard(), compareTo.getGuard()) {
                    isEqual.set(false);
                    return _case;
                }
                
```

knutwannheden (approved):
LGTM!

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -348,13 +348,15 @@ public J visitCase(CaseTree node, Space fmt) {
...
                 null,
+                null,
                 JContainer.build(
-                        node.getExpressions().isEmpty() ? EMPTY : sourceBefore("case"),
-                        node.getExpressions().isEmpty() ?
+                        node.getLabels().isEmpty() ? EMPTY : sourceBefore("case"),
+                        node.getLabels().isEmpty() || node.getLabels().getFirst() instanceof DefaultCaseLabelTree ?
                                 List.of(JRightPadded.build(new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), skip("default"), null, null))) :
-                                convertAll(node.getExpressions(), commaDelim, t -> EMPTY),
+                                convertAll(node.getLabels(), commaDelim, ignored -> sourceBefore("when", '-')),
```
knutwannheden, line 356: I know I suggested this, but thinking about it I think this is cleaner and safer (untested):

```suggestion
                                convertAll(node.getLabels(), commaDelim, ignored -> node.getGuard() != null ? sourceBefore("when") : EMPTY),
```

knutwannheden (approved):

PR #4663, opened 2024-11-11
Update model for interfaces extending interfaces

rlsanders4:
## What's changed?
Update the parser model to correctly represent an interface extending another interface.

## What's your motivation?
Currently, when a Java interface extends another interface it is represented in the resulting `J.ClassDeclaration` as an 'implements' rather than an 'extends'. This can cause confusion when authoring custom recipes, because instead of accessing the extended interface using `J.ClassDeclaration.getExtends()` it must instead be accesed using `J.ClassDeclaration.getImplements()`.

For example:

Parent interface
```
package parent;
public interface parentInterface {}
```

Child interface
```
package child;
public interface childInterface extends parentInterface {}
```

Once the child interface has been parsed, in a `visitClassDeclaration()` visitor we would expect to be able to access the extended parent interface with `cd.getExtends()`. However, that currently returns null, and you must access it instead with `cd.getImplements()`.

### Checklist


rewrite-java/src/test/java/org/openrewrite/java/ExtendInterfaceTest.java
```diff
@@ -0,0 +1,54 @@
...
+        @Language("java")
+        String parentInterface = """
+          package parent;
+          public interface parentInterface {}
+          """;
+
+        @Language("java")
+        String childInterface = """
+          package child;
+          public interface childInterface extends parentInterface {}
```
timtebeek, line 38:
```suggestion
          package child;
          import parent.parentInterface;
          public interface childInterface extends parentInterface {}
```

timtebeek:
Saving others a click: we [fail on line 50](https://github.com/openrewrite/rewrite/pull/4663/files#diff-4245575319ef095eebf414e0777b3f7615fe273a8098f94a9449ec18bfbe7087R50) with
```
        // Check that childInterface does not implement any interfaces
        assertNull(cd.getImplements());
```
```
ExtendInterfaceTest > interfaceExtendsInterface() FAILED
    org.opentest4j.AssertionFailedError: expected: <null> but was: <[parentInterface]>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertNull.failNotNull(AssertNull.java:50)
        at app//org.junit.jupiter.api.AssertNull.assertNull(AssertNull.java:35)
        at app//org.junit.jupiter.api.AssertNull.assertNull(AssertNull.java:30)
        at app//org.junit.jupiter.api.Assertions.assertNull(Assertions.java:279)
        at app//org.openrewrite.java.ExtendInterfaceTest.interfaceExtendsInterface(ExtendInterfaceTest.java:50)
```

timtebeek:
Thanks for diving in here @rlsanders4 ! Surprised no one else had stumbled upon this before; Now I'm wondering if anyone had worked around this, although recipes for interfaces that extend interfaces might be rare.

As a note to myself we should likely add the same changes in the other Java parsers, just so this is consistent when parsed and targeted there on Java 8, 11 and 21 as well.

timtebeek (changes requested):
Capturing some of what was discussed internally here as well: Yes it's confusing that extends/implements are intuitively the wrong way round, but it matches how these are modeled in the Java compiler as well. In part this might be due to interfaces being able to extend multiple other interfaces, whereas classes can only extend one class, but can implement many interfaces. Changing any part of that now it likely to be painful given that we serialize LSTs. Perhaps the best way forward is to document this seeming inversion for interfaces on the getExtends and getImplements methods for class declarations.

timtebeek (approved):
Thanks a lot for working through this with us to improve the experience for folks going forward!

PR #4664, opened 2024-11-11
add gradle recipe: RemoveSettingsPluginManagement

sullis:
## What's changed?

new recipe:  RemoveSettingsPluginManagement

This recipe removes "pluginManagement" from Gradle settings.gradle

## What's your motivation?

At work, I have hundreds of Gradle projects. I need to remove "pluginManagement" from all of my Gradle projects.

n/a

n/a

n/a

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveSettingsPluginManagement.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveSettingsPluginManagement.java
@@ -24,11 +24,8 @@ import org.openrewrite.TreeVisitor;
 import org.openrewrite.gradle.IsSettingsGradle;
 import org.openrewrite.groovy.GroovyIsoVisitor;
 import org.openrewrite.groovy.tree.G;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.Statement;
-
-import java.util.ArrayList;
-import java.util.List;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveSettingsPluginManagement.java
```diff
@@ -0,0 +1,57 @@
...
+public class RemoveSettingsPluginManagement extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Remove Gradle settings pluginManagement";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Remove Gradle settings pluginManagement from `settings.gradle(.kts)`.";
```
timtebeek, line 41:
```suggestion
        return "Remove Gradle settings pluginManagement from `settings.gradle`.";
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveSettingsPluginManagement.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.tree.J;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class RemoveSettingsPluginManagement extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Remove Gradle settings pluginManagement";
```
timtebeek, line 36:
```suggestion
        return "Remove Gradle settings `pluginManagement`";
```

sullis:
CI build in progress.

shanman190:
@sullis, can you provide some insight on the deeper context that you're trying to solve for?

At first glance, the deeper context is not apparent and the recipe itself feels a bit heavy at the moment from a removal scope by itself alone.

sullis:
> @sullis, can you provide some insight on the deeper context that you're trying to solve for?
> 
> At first glance, the deeper context is not apparent and the recipe itself feels a bit heavy at the moment from a removal scope by itself alone.

I just updated the PR description.

shanman190:
@sullis, so there is a lot of configuration that can happen within the `pluginManagement` block, so a blanket remove the entire block feels a bit odd/one off to me coming at it from an OSS recipe catalog viewpoint.

Some of the things that do come to mind are maybe we are failing to cleanup a now empty block after having say removed the last plugin repository. In this case, I'd almost expect for the remove repository recipe to cleanup empty blocks that it's left behind.

With that said, we have seen users create their own internal recipe libraries where they have put similar recipes where they know within their strict context that they want to be destructive even though a more general purpose recipe can't really be described or arrived at. Maybe that would work better here?

sambsnyd:
I share Shannon's questioning about how general-use this recipe is. It seems quite specific to your scenario.

In your situation you know that this change is broadly desirable, so it isn't necessary for the recipe itself to be any smarter than this. For most other gradle builds most of the time removing the plugin management is going to break the build. If this were a recipe which could tell when plugin management is unnecessary/redundant and remove it only in those cases I would be more enthusiastic. If you're not interested in pursuing that maybe this is a better fit for an internal recipe module rather than the OSS catalogue.

sullis:
> If you're not interested in pursuing that maybe this is a better fit for an internal recipe module rather than the OSS catalogue.

Sounds reasonable.  I'll close this PR.  I am going to make this an internal recipe.

Issue #4157, opened 2024-04-29
Allow `AddOrUpdateAnnotationAttribute` to add values to an array


timtebeek:
## What problem are you trying to solve?
`AddOrUpdateAnnotationAttribute` can not currently add values to an array. When you try that you get
```
ClassCastException: class org.openrewrite.java.tree.J$NewArray cannot be cast to class org.openrewrite.java.tree.J$Literal
```

## Describe the solution you'd like
To be able to add a value as seen here

**Before**
```java
@SpringBootApplication(scanBasePackages = {"com.abc"})
```
**After**
```java
@SpringBootApplication(scanBasePackages = {"com.abc", "com.xyz"})
```

## Have you considered any alternatives or workarounds?
We could create a new recipe, or explicitly not allow this.
Not clear yet if we should append to existing array, or overwrite existing value.

timtebeek:
From further discussion duplicates should be prevented; append when element is not already present.
For removal of elements a similar change can be made to the remove attribute from annotation recipe.
https://github.com/openrewrite/rewrite/blob/bf4dcfe8a4656f089b8de75a56263f7669950b74/rewrite-java/src/main/java/org/openrewrite/java/RemoveAnnotationAttribute.java#L29

PR #4667, opened 2024-11-13
Append items to annotation attribute array

nielsdebruin:
## What's changed?

This PR enhances the functionality of the `AddOrUpdateAnnotationAttribute` recipe, allowing not only to override but also to append to array values. A new option, `arrayAppend`, has been introduced, enabling users to append to attributes with an `array` type. When used in conjunction with the `addOnly` option, users can control whether a value will be appended only if it is not already present, thereby avoiding duplication.

## What's your motivation?
- See #4157 

@timtebeek @Laurens-W 

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
@@ -70,6 +70,14 @@ public class AddOrUpdateAnnotationAttribute extends Recipe {
     @Nullable
     Boolean addOnly;
 
+    @Option(displayName = "Append array",
+            description = "If the attribute is an array, setting this option to `true` will append the value(s). " +
+                          "In conjunction with `addOnly`, it is possible to control duplicates:" +
+                          "`addOnly=true`, always append. " +
+                          "`addOnly=false`, only append if the value is not already present.")
+    @Nullable
+    Boolean appendArray;
+
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return Preconditions.check(new UsesType<>(annotationType, false), new JavaIsoVisitor<ExecutionContext>() {
@@ -127,6 +135,17 @@ public class AddOrUpdateAnnotationAttribute extends Recipe {
                                 List<Expression> jLiteralList = ((J.NewArray) as.getAssignment()).getInitializer();
                                 String attributeValueCleanedUp = attributeValue.replaceAll("\\s+","").replaceAll("[\\s+{}\"]","");
                                 List<String> attributeList = Arrays.asList(attributeValueCleanedUp.contains(",") ? attributeValueCleanedUp.split(",") : new String[]{attributeValueCleanedUp});
+                                if (Boolean.TRUE.equals(appendArray)) {
+                                    for (int i = 0, j = jLiteralList.size(); i < attributeList.size(); j++, i++) {
+                                        String newAttributeListValue = maybeQuoteStringArgument(attributeName, attributeList.get(i), finalA);
+                                        if (Boolean.FALSE.equals(addOnly) && attributeValIsAlreadyPresentOrNull(jLiteralList, newAttributeListValue)) {
+                                            j--;
+                                            continue;
+                                        }
+                                        jLiteralList.add(j, new J.Literal(Tree.randomId(), jLiteralList.get(j - 1).getPrefix(), Markers.EMPTY, newAttributeListValue, newAttributeListValue, null, JavaType.Primitive.String));
+                                    }
+                                    return as.withAssignment(((J.NewArray) as.getAssignment()).withInitializer(jLiteralList));
+                                }
                                 int m = 0;
                                 for (int i = 0; i< Objects.requireNonNull(jLiteralList).size(); i++){
                                     if (i >= attributeList.size()){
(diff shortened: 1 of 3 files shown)


Laurens-W:
I think you first proposed option is the way to go. You'd want to separately be able to specify whether you only want to add a property value when it's missing but also to specify whether you want to replace or append the new value(s)

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -123,10 +136,28 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
                                 List<String> attributeList = Arrays.asList(attributeValueCleanedUp.contains(",") ? attributeValueCleanedUp.split(",") : new String[]{attributeValueCleanedUp});
+
+                                for (Marker m : as.getMarkers().getMarkers()) {
+                                    if (m instanceof AlreadyAppended && ((AlreadyAppended) m).getValues().equals(newAttributeValue)) {
+                                        return as;
+                                    }
+                                }
+
+                                if (Boolean.TRUE.equals(appendArray)) {
+                                    for (int i = 0, j = jLiteralList.size(); i < attributeList.size(); j++, i++) {
```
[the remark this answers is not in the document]
Laurens-W: Or at the very least giving it a descriptive name would :)

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -70,6 +72,14 @@ public String getDescription() {
     @Nullable
     Boolean addOnly;
 
+    @Option(displayName = "Append array",
+            description = "If the attribute is an array, setting this option to `true` will append the value(s). " +
+                          "In conjunction with `addOnly`, it is possible to control duplicates:" +
```
timtebeek, line 77: Note that String concatenation means these lines would be directly appended. Best to have a space after the colon then.
```suggestion
                          "In conjunction with `addOnly`, it is possible to control duplicates: " +
```

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -255,4 +282,23 @@ private static boolean attributeIsString(@Nullable String attributeName, J.Annot
         }
         return false;
     }
+
+    private static boolean attributeValIsAlreadyPresentOrNull(List<Expression> expression, @Nullable String attributeValue) {
```
timtebeek, line 286: What does the `OrNull` mean in the method name suffix here? Because I do not see this return `true` if the new or old attribute value were to be `null`.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
nielsdebruin: Fixed

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -123,10 +133,28 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
+                                    for (String attrListValues : attributeList) {
+                                        String newAttributeListValue = maybeQuoteStringArgument(attributeName, attrListValues, finalA);
+                                        if (Boolean.FALSE.equals(addOnly) && attributeValIsAlreadyPresent(jLiteralList, newAttributeListValue)) {
+                                            continue;
+                                        }
+                                        unchanged = false;
+                                        jLiteralList.add(new J.Literal(randomId(), Space.EMPTY, Markers.EMPTY, newAttributeListValue, newAttributeListValue, null, JavaType.Primitive.String));
+                                    }
+                                    return unchanged ? as : as.withAssignment(((J.NewArray) as.getAssignment()).withInitializer(jLiteralList))
+                                            .withMarkers(as.getMarkers().add(new AlreadyAppended(randomId(), newAttributeValue)));
```
timtebeek, line 156: Might just be me, but I prefer to name my booleans in a way that does not contain negation, as the alternative leads to double negations which are harder to read when using in conditionals.
```suggestion
                                    boolean changed = false;
                                    for (String attrListValues : attributeList) {
                                        String newAttributeListValue = maybeQuoteStringArgument(attributeName, attrListValues, finalA);
                                        if (Boolean.FALSE.equals(addOnly) && attributeValIsAlreadyPresent(jLiteralList, newAttributeListValue)) {
                                            continue;
                                        }
                                        changed = true;
                                        jLiteralList.add(new J.Literal(randomId(), Space.EMPTY, Markers.EMPTY, newAttributeListValue, newAttributeListValue, null, JavaType.Primitive.String));
                                    }
                                    return changed ? as.withAssignment(((J.NewArray) as.getAssignment()).withInitializer(jLiteralList))
                                            .withMarkers(as.getMarkers().add(new AlreadyAppended(randomId(), newAttributeValue))) : as;
```

timtebeek (approved):
Approved aside from the last suggestion; let me know then we can merge right after.

PR #4671, opened 2024-11-14
When merging yaml files comments on the last element of existing block  are placed at the wrong line

jevanlingen:
## What's changed?
After merge of two yam files, the comments will be placed at the right position.

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/2218

### Checklist


rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -22,16 +22,27 @@
...
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static org.openrewrite.internal.ListUtils.*;
+import static org.openrewrite.yaml.MergeYaml.REMOVE_PREFIX;
+
```
jevanlingen, line 38:
```suggestion
/**
 * Visitor class to merge two yaml files.
 *
 * @implNote Loops recursively through the documents, for every part a new MergeYamlVisitor instance will be created.
 * As inline comments are put on the prefix of the next element (which can be an item very much higher in the tree),
 * the following solutions are chosen to merge the comments as well:
 * <ul>
 * <li>when an element has new items, the comment of the next element is copied to the previous element
 * <li>the original comment will be removed (either by traversing the children or by using cursor messages)
 *
 * @param <P> An input object that is passed to every visit method.
 */
```

Possible JavaDoc.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
```diff
@@ -110,9 +111,11 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
             public Yaml.Document visitDocument(Yaml.Document document, ExecutionContext ctx) {
                 if ("$".equals(key)) {
-                    return document.withBlock((Yaml.Block) new MergeYamlVisitor<>(document.getBlock(), yaml,
-                            Boolean.TRUE.equals(acceptTheirs), objectIdentifyingProperty).visitNonNull(document.getBlock(),
-                            ctx, getCursor()));
+                    Yaml.Document d = document.withBlock((Yaml.Block)
+                            new MergeYamlVisitor<>(document.getBlock(), yaml, Boolean.TRUE.equals(acceptTheirs), objectIdentifyingProperty)
+                                    .visitNonNull(document.getBlock(), ctx, getCursor())
+                    );
+                    return getCursor().getMessage(REMOVE_PREFIX, false) ? d.withEnd(d.getEnd().withPrefix("")) : d;
```
jevanlingen, line 118: I actually don't like very much that I needed to put this here; I would rather keep the cursor messages locked in the `MergeYamlVisitor` class. I couldn't make it work though, because the ` visitDocument` will not be called in that class. 

If I overlooked something please let me know! It would be really nice to let this MergeYaml visitor file untouched.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -89,96 +100,160 @@ public Yaml visitSequence(Yaml.Sequence existingSeq, P p) {
...
-    public Yaml visitMapping(Yaml.Mapping existingMapping, P p) {
-        if (scope.isScope(existingMapping) && incoming instanceof Yaml.Mapping) {
-            return mergeMapping(existingMapping, (Yaml.Mapping) incoming, p, getCursor());
+    public Yaml visitMapping(Mapping existingMapping, P p) {
+        if (existing.isScope(existingMapping) && incoming instanceof Mapping) {
+            Mapping mapping = mergeMapping(existingMapping, (Mapping) incoming, p, getCursor());
+
+            if (getCursor().getMessage(REMOVE_PREFIX, false)) {
+                List<Entry> entries = ((Mapping) getCursor().getValue()).getEntries();
+                return mapping.withEntries(mapLast(mapping.getEntries(), it -> it.withPrefix("\n" + grabPartLineBreak(entries.get(entries.size() - 1), 1))));
```
jevanlingen, line 109: Doing `entries.get(entries.size() - 1),` over and over again is not my cup of tea. Maybe we should consider a `last` function in the ListUtils class...

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -89,96 +101,160 @@ public Yaml visitSequence(Yaml.Sequence existingSeq, P p) {
...
-                if (keyMatches(existingEntry, incomingEntry)) {
+        List<Entry> mutatedEntries = concatAll(mergedEntries, map(m2.getEntries(), it -> {
+            for (Entry existingEntry : m1.getEntries()) {
+                if (keyMatches(existingEntry, it)) {
                     return null;
                 }
             }
-            if (shouldAutoFormat) {
-                incomingEntry = autoFormat(incomingEntry, p, cursor);
+            // workaround: autoFormat put sometimes extra spaces before elements
```
jevanlingen, line 153: Workaround fixes only so many cases; should be removed in the future and done in the parser:
- https://github.com/openrewrite/rewrite/pull/4358

rewrite-core/src/main/java/org/openrewrite/Cursor.java
```diff
@@ -57,7 +57,7 @@ public Cursor getRoot() {
     /**
      * @return true if this cursor is the root of the tree, false otherwise
      */
-    final boolean isRoot() {
+    final public boolean isRoot() {
```
Laurens-W, line 60: Not sure whether this is a desired change

jevanlingen: You can already check this via a different route: `getCursor().getRoot().equals(getCursor())`. As the `getRoot()` is already public, I don't see why the `isRoot()` should not be public as well.

rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
```diff
@@ -1010,6 +1010,197 @@ void mergeSequenceMapChangeComplexMapping() {
...
+                      1: something else
+                      2: old desc # Some comment 5
+                    D2:
+                      1: a
+                      2: b # Some comment 10
+                      3: c
+                    D3: # Some comment 6
+                      1: old description # Some comment 7
+                    D4: # Some comment 8
+                      1: old text # Some comment 9
```
timtebeek, line 1144: Right now there's a mix of comments I suspect are not even visited, but they are indistinguishable from comments that are relevant to the recipe. I feel there's a bit of a missed opportunity here in the comments to clarify what the recipe should be doing. How about instead of `Some comment` we describe what we expect for that line & comment with comments like
- `Comment untouched`,
- `Comment moved from next block to previous element`

rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -142,7 +144,7 @@ private Yaml.Documents parseFromInput(Path sourceFile, EncodingDetectingInputStr
                         documents.add(document.withEnd(new Yaml.Document.End(
                                 randomId(),
                                 fmt,
-                                Markers.EMPTY,
+                                EMPTY,
```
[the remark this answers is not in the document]
jevanlingen: Ah the forever discussion between conciseness vs completeness. I will revert the changes if that's the way to go.

rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
```diff
@@ -1010,6 +1010,197 @@ void mergeSequenceMapChangeComplexMapping() {
...
+              spring:
+                application:
+                  name: main # Some comment
+                  name2: main
+                  description: a description
+              """
+          )
+        );
+    }
+
```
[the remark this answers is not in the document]
jevanlingen: Very good, I expanded the `existingEntryBlockWithCommentsAllOverThePlace` test with all those nasty cases.

PR #4672, opened 2024-11-14
Fix bug using `AddOrUpdateAnnotationAttribute` on annotation with empty brackets

nielsdebruin:
## What's changed?
Fixed bug that caused incorrect code when `AddOrUpdateAnnotationAttribute` on annotation with empty bracket e.g., `@Foo()`.

## What's your motivation?

The recipe `AddOrUpdateAnnotationAttribute` can be used to modify the attribute values on annotation. When no attribute values have been specified, the recipe will work fine if the syntax `@Foo` is used. However, if `@Foo()` is used, which is also valid java. Then, the recipe breaks.

**Example**
```java
public @interface Foo {
    String[] array() default {};
}
```
If we run `AddOrUpdateAnnotationAttribute` on `A` annotated with `@Foo`, configured to set the value of `Foo.array` to "newTest1,newTest2" we **correctly** get:
```java
@Foo
public class A {}
@Foo(array = {"newTest1", "newTest2"})
public class A {
}
```
Yet applying it to `@Foo()`, **incorrectly** yields
```java
@Foo()
public class A {}
@Foo(array = "newTest1,newTest2", )
public class A {
}
```

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
@@ -82,7 +82,7 @@ public class AddOrUpdateAnnotationAttribute extends Recipe {
 
                 String newAttributeValue = maybeQuoteStringArgument(attributeName, attributeValue, a);
                 List<Expression> currentArgs = a.getArguments();
-                if (currentArgs == null || currentArgs.isEmpty()) {
+                if (currentArgs == null || currentArgs.isEmpty() || currentArgs.get(0) instanceof J.Empty) {
                     if (newAttributeValue == null) {
                         return a;
                     }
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -82,7 +82,7 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
 
                 String newAttributeValue = maybeQuoteStringArgument(attributeName, attributeValue, a);
                 List<Expression> currentArgs = a.getArguments();
-                if (currentArgs == null || currentArgs.isEmpty()) {
+                if (currentArgs == null || currentArgs.isEmpty() || currentArgs.get(0) instanceof J.Empty) {
```
Laurens-W, line 85: Should this perhaps be changed in the parser instead of here?
I'd assume when you use `@Foo` currentArgs = null and when you use `@Foo()` currentArgs = empty

timtebeek (approved):

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
timtebeek: We use that J.Empty when we want to be explicit about the presence of those parentheses; I think this change is good to merge. Thanks both!

nielsdebruin: Thanks for review!

Issue #4676, opened 2024-11-15
Redundant space for DeleteMethodArgument when argumentIndex=0


ckcd:
Below case will failed because it will change to `"public class A {{ B.foo( 1, 2); }}"` which contains a redundant space before the first argument.

```java
    @Test
    void deleteFirstArgument() {
        rewriteRun(
          spec -> spec.recipe(new DeleteMethodArgument("B foo(int, int, int)", 0)),
          java(b),
          java(
            "public class A {{ B.foo(0, 1, 2); }}",
            "public class A {{ B.foo(1, 2); }}"
          )
        );
    }
```

correct answer: `"public class A {{ B.foo(1, 2); }}"`, no space before the first `1` element.
actual answer:    `"public class A {{ B.foo( 1, 2); }}"` has a redundant space before the first `1` element.

## What version of OpenRewrite are you using?

I am using

- `main` branch for rewrite.

## How are you running OpenRewrite?

I am using the `rewrite-maven-plugin` with latest version.

## What is the smallest, simplest way to reproduce the problem?

```java
    @Test
    void deleteFirstArgument() {
        rewriteRun(
          spec -> spec.recipe(new DeleteMethodArgument("B foo(int, int, int)", 0)),
          java(b),
          java(
            "public class A {{ B.foo(0, 1, 2); }}",
            "public class A {{ B.foo(1, 2); }}"
          )
        );
    }
```

## What did you expect to see?

correct answer: `"public class A {{ B.foo(1, 2); }}"`, no space before the first `1` element.

## What did you see instead?

actual answer:    `"public class A {{ B.foo( 1, 2); }}"` has a redundant space before the first `1` element.

## What is the full stack trace of any errors you encountered?

```
[Unexpected result in "A.java":
diff --git a/A.java b/A.java
index b7c498e..fda8943 100644
--- a/A.java
+++ b/A.java
@@ -1 +1 @@ 
-public class A {{ B.foo(1, 2); }}
\ No newline at end of file
+public class A {{ B.foo( 1, 2); }}
\ No newline at end of file
] 
expected: "public class A {{ B.foo(1, 2); }}"
 but was: "public class A {{ B.foo( 1, 2); }}"
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Sure, will file a PR to fix this.

PR #4677, opened 2024-11-15
Remove redundant space for DeleteMethodArgument when argumentIndex=0

ckcd:
## What's changed?

- Fix https://github.com/openrewrite/rewrite/issues/4676

Below case will failed because it will change to `"public class A {{ B.foo( 1, 2); }}"` which contains a redundant space before the first argument.

```java
    @Test
    void deleteFirstArgument() {
        rewriteRun(
          spec -> spec.recipe(new DeleteMethodArgument("B foo(int, int, int)", 0)),
          java(b),
          java(
            "public class A {{ B.foo(0, 1, 2); }}",
            "public class A {{ B.foo(1, 2); }}"
          )
        );
    }
```

correct answer: `"public class A {{ B.foo(1, 2); }}"`, no space before the first `1` element.
actual answer:    `"public class A {{ B.foo( 1, 2); }}"` has a redundant space before the first `1` element.

This PR aims to fix the issue.

## What's your motivation?

Fix https://github.com/openrewrite/rewrite/issues/4676

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindPluginsTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindPluginsTest.java
@@ -69,7 +69,8 @@ class FindPluginsTest implements RewriteTest {
     void settingsResolutionStrategy() {
         rewriteRun(
           spec -> spec.beforeRecipe(withToolingApi()),
-          settingsGradle("""
+          settingsGradle(
+              """
             pluginManagement {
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks @ckcd ! Indeed looks better when there's no leading space after removing the first element. I've pushed up a very small change to assume the prefix used previously, with an additional test class that shows when that would help here. With that we're good to merge here.

Issue #4400, opened 2024-08-08
Support for changing versions in TOML files


michaelmcfadyensky:
## What problem are you trying to solve?
We specify all of our dependencies and their versions inside toml files for our gradle projects. We would like a away to automate the upgrading of these versions via open rewrite.

## Describe the solution you'd like
Recipe available for changing the version of a dependency specified in a toml file

## Have you considered any alternatives or workarounds?
Consider using the find and replace recipe but it leads to several work arounds having to be in place for it to work.

## Additional context
https://rewriteoss.slack.com/archives/C01A843MWG5/p1723123348552829

timtebeek:
@steplica thanks a lot for the offer to help!
1. I think in the early stages of getting the parser to work @KnutWannheden might be best to guide you, as he's been most active with recent parser developments. I'd expect `rewrite-toml` should still be part of `openrewrite/rewrite`, with an eye towards integration with Gradle
2. once the parser works we'd need to hook it into the [rewrite-maven-plugin](https://github.com/openrewrite/rewrite-maven-plugin) and [rewrite-gradle-plugin](https://github.com/openrewrite/rewrite-gradle-plugin), and [rewrite-polyglot](https://github.com/openrewrite/rewrite-polyglot)
3. then once files are parsed as Toml, we'd need to integrate with our Gradle support; at that point we can tag Shannon Pamperl (also here) for guidance, likely looking towards our Trait support to minimize recipe changes

tompson:
will this also support gradle version catalogs defined in `settings.gradle` (no toml files involved) ?

jizhangattw:
👍 sounds good. keep an eye on this issue.

timtebeek:
- One step closer after https://github.com/openrewrite/rewrite/pull/4845

shanman190:
Next steps:
- [ ] Update `gradle.UpgradeDependencyVersion` to handle dependencies from the Gradle Version Catalog `libraries` table with the version appearing in a String literal [1]
- [ ] Update `gradle.UpgradeDependencyVersion` to handle dependencies from the Gradle Version Catalog `libraries` table with the version appearing within the defined inline table [2]
- [ ] Update `gradle.UpgradeDependencyVersion` to handle dependencies from the Gradle Version Catalog `libraries` table with a version reference being used [3]
- [ ] Update `gradle.plugins.UpgradePluginVersion` to handle plugins from the Gradle Version Catalog `plugins` table with the version appearing in a String literal [4]
[...]

PR #4678, opened 2024-11-15
Add toml parser

steplica:
### Checklist


shanman190:
Dropping this here as well, since Steven and I are collaborating in parallel on the effort trying different things here and there. This PR will serve as the final source to be contributed for TOML support.

https://github.com/shanman190/rewrite-toml

shanman190:
Related to https://github.com/openrewrite/rewrite/issues/4400. A future PR will take the work done here and integrate it with the rewrite-gradle module.

rewrite-toml/src/main/java/org/openrewrite/toml/internal/grammar/TomlParserBaseVisitor.java
```diff
@@ -0,0 +1,190 @@
...
+	 * {@link #visitChildren} on {@code ctx}.</p>
+	 */
+	@Override public T visitArray_(TomlParser.Array_Context ctx) { return visitChildren(ctx); }
+	/**
+	 * {@inheritDoc}
+	 *
+	 * <p>The default implementation returns the result of calling
+	 * {@link #visitChildren} on {@code ctx}.</p>
+	 */
+	@Override public T visitArray_values(TomlParser.Array_valuesContext ctx) { return visitChildren(ctx); }
```
knutwannheden, line 133: Our ANTLR grammars typically use camel case as that then also yields nicer Java identifiers out-of-the-box.

shanman190: That's definitely fair. This was just pulled directly from https://github.com/antlr/grammars-v4/tree/master/toml. 

Do you recall if future syncs with the upstream just usually remake those particular kinds of adjustments for other language bindings? This was the thought running through my head as to why I hadn't mentioned the same thing to Steve.

knutwannheden: In my experience these grammars are hardly ever updated and quite often also contain small bugs, so I wouldn't worry too much. If we ever need to fix a bug we can still look at the history of the upstream grammar.

steplica: @knutwannheden Do you just want the methods in TomlParserBaseVisitor to be camel case? Or do you also want all the types/methods in TomlParser (also generated by ANTLR)

shanman190: @steplica, Knut is meaning to update the ANTLR grammar to remove the underscores there which then translates throughout the rest of the code as well.

timtebeek:
Thanks a lot for the initial push here @steplica !
- Superseded by #4845

PR #4682, opened 2024-11-18
Introduce scoped variable trait

Laurens-W:
## What's changed?

Added a trait to pull together the scope and identifier of `NamedVariable`s and added testcases to showcase its usage

@sambsnyd @timtebeek 

## Any additional context

Continued from 
- https://github.com/openrewrite/rewrite-migrate-java/pull/604

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/trait/ScopedVariable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/trait/ScopedVariable.java
@@ -46,7 +46,7 @@ public class ScopedVariable implements Trait<J.VariableDeclarations.NamedVariabl
             return null;
         }
 
-        public Cursor getDeclaringScope(Cursor cursor) {
+        private Cursor getDeclaringScope(Cursor cursor) {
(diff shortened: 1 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/trait/ScopedVariable.java
```diff
@@ -0,0 +1,57 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.trait.SimpleTraitMatcher;
+import org.openrewrite.trait.Trait;
+
+@Value
+public class ScopedVariable implements Trait<J.VariableDeclarations.NamedVariable> {
```
knutwannheden, line 27: If we want to support `instanceof` pattern matching, then we won't be able to use `NamedVariable` as the generic type parameter but would instead have to use `J` for this.

PR #4688, opened 2024-11-20
Make `FindMissingTypes` stricter for method invocations

knutwannheden:
Validate argument count with respect to parameter count.


knutwannheden:
@timtebeek What do you think? We could of course also validate the argument types against the parameter types.

timtebeek:
Yes makes sense to me! Might want to check the same for ParameterNames while we're at it, but otherwise good to add & monitor downstream.

knutwannheden:
Looks like this revealed a bug we have in the Javadoc type attribution.

timtebeek (approved):
Thanks a lot! Should help harden recipes downstream.

PR #4692, opened 2024-11-20
Add version tag for managed dependency with property

mrakshith21:
## What's changed?

When properties are used with managed dependencies, update property only for dependencies managed by parent. For those managed by a dependency under `dependencyManagement`, add version tag.

Checking if `requestedBom` and parent have the same gav

## Any additional context
https://rewriteoss.slack.com/archives/C01A843MWG5/p1731906971515379?thread_ts=1725855120.842389&cid=C01A843MWG5


timtebeek (approved):
Love the small change with big impact here @mrakshith21 ; Great to reliably actually override the version, as opposed to looking like we would but don't. Thanks a lot!

Issue #4681, opened 2024-11-18
Expand `Reference` to include `.properties` and `.yml`


timtebeek:
## What problem are you trying to solve?
Type references can occur in more places; for instance as value in a properties file, or as value or even key in yaml files.

## Describe the solution you'd like
Expand the recently added Reference and SourceFileWIthReferences to also cover properties and yaml. For Yaml specifically it would be nice to also cover cases where the type might be a key, and not just a value.
```yaml
recipeList:
  - org.....MyRecipe:
      with: arg
```
For properties only looking at values is likely enough.

## Additional context
- https://github.com/openrewrite/rewrite-hibernate/issues/44
- [SpringReference](https://github.com/openrewrite/rewrite/blob/main/rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringReference.java)
- [Xml.Document](https://github.com/openrewrite/rewrite/blob/main/rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java#L86)
- [Xml.Document reference cache](https://github.com/openrewrite/rewrite/blob/main/rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java#L170-L189)

PR #4697, opened 2024-11-21
Add initial support for java references in properties file

nielsdebruin:
## What's changed?

Add initial support for generating java type references from a `.properties` file, support has only been added for the values, not the keys. 

## What's your motivation?
Adding this feature allows existing recipes that support references to be applied to `.properties` files. For more details see:
- #4681 

The current implementation is still rather limited in its functionality and many design choices are still to be made. 

Currently, references of `Kind.Type` or `Kind.Package` are generated for either a Java package, e.g., `com.openrewrite`, or a fully qualified type, such as `java.lang.Integer`, but not `Integer`, using a regex derived from the current implementation of `SpringReference`. This matching is only done for the **_values_** in a properties file, not the keys. If this is not the case, nearly every key would likely qualify as a package since they usually contain at least a single ".". Is this as desired?

I have added support for renaming `Properties.Entry` in the `TypeMatcher` and `PackageMatcher` and added a test for `ChangeType` and `ChangePackage` recipes. It would perhaps be nice to have a list of all the classes which need to support properties, e.g. `AnnotationMatcher`?

@timtebeek @Laurens-W 

### Checklist

### DIFF
--- a/rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
+++ b/rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.properties.trait;
+
+import lombok.Value;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.SourceFile;
+import org.openrewrite.properties.tree.Properties;
+import org.openrewrite.trait.Reference;
+import org.openrewrite.trait.SimpleTraitMatcher;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+@Value
+public class PropertiesReference implements Reference {
+    Cursor cursor;
+    Kind kind;
+
+    @Override
+    public @NonNull Kind getKind() {
+        return kind;
+    }
+
+    @Override
+    public @NonNull String getValue() {
+        if (getTree() instanceof Properties.Entry) {
+            return ((Properties.Entry) getTree()).getValue().getText();
+        }
+        throw new IllegalArgumentException("getTree() must be an Properties.Entry: " + getTree().getClass());
+    }
+
+    @Override
+    public boolean supportsRename() {
+        return true;
+    }
+
+    public static class Matcher extends SimpleTraitMatcher<PropertiesReference> {
+        private static final Pattern javaFullyQualifiedTypePattern = Pattern.compile("\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(?:\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*");
+
+        @Override
+        protected @Nullable PropertiesReference test(Cursor cursor) {
+            Object value = cursor.getValue();
+            if (value instanceof Properties.Entry && javaFullyQualifiedTypePattern.matcher(((Properties.Entry) value).getValue().getText()).matches()) {
+                if (determineKind(((Properties.Entry) value).getValue().getText()) == Kind.PACKAGE) {
+                    return new PropertiesReference(cursor, Kind.PACKAGE);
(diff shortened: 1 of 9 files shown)


rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
```diff
@@ -0,0 +1,92 @@
...
+
+        @Override
+        protected @Nullable PropertiesReference test(Cursor cursor) {
+            Object value = cursor.getValue();
+            if (value instanceof Properties.Entry && javaFullyQualifiedTypePattern.matcher(((Properties.Entry) value).getValue().getText()).matches()) {
+                if (determineKind(((Properties.Entry) value).getValue().getText()) == Kind.PACKAGE) {
+                    return new PropertiesReference(cursor, Kind.PACKAGE);
+                } else if (determineKind(((Properties.Entry) value).getValue().getText()) == Kind.TYPE) {
+                    return new PropertiesReference(cursor, Kind.TYPE);
+                }
```
Laurens-W, line 65:
```suggestion
                return new PropertiesReference(cursor, determineKind(((Properties.Entry) value).getValue().getText()));
```

rewrite-properties/src/test/java/org/openrewrite/properties/trait/PropertiesReferenceTest.java
```diff
@@ -0,0 +1,50 @@
...
+              c.type=Integer
+              """,
+            """
+              ~~(java.lang.String)~~>a.fqt=java.lang.String
+              ~~(java.lang)~~>b.package=java.lang
+              c.type=Integer
+              """,
+            spec -> spec.afterRecipe(doc -> {
+                assertThat(doc.getReferences().getReferences().stream().filter(typeRef -> typeRef.getValue().equals("java.lang.String") && typeRef.getKind().equals(Reference.Kind.TYPE)).count()).isEqualTo(1);
+                assertThat(doc.getReferences().getReferences().stream().filter(typeRef -> typeRef.getValue().equals("java.lang") && typeRef.getKind().equals(Reference.Kind.PACKAGE)).count()).isEqualTo(1);
```
Laurens-W, line 45: How relevant is it that there is only one match?

rewrite-properties/src/main/java/org/openrewrite/properties/tree/Properties.java
```diff
@@ -55,7 +55,9 @@ default <P> boolean isAcceptable(TreeVisitor<?, P> v, P p) {
 
     @lombok.Value
     @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
-    class File implements Properties, SourceFile {
+    @RequiredArgsConstructor
+    @AllArgsConstructor
```
timtebeek, line 59: Xml.Document uses 
```java
```
What's the reason behind not copying that approach here?

rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
```diff
@@ -0,0 +1,92 @@
...
+    public static class Provider implements Reference.Provider {
+        @Override
+        public @NonNull Set<Reference> getReferences(SourceFile sourceFile) {
+            Set<Reference> references = new HashSet<>();
+            new Matcher().asVisitor(reference -> {
+                references.add(reference);
+                return reference.getTree();
+            }).visit(sourceFile, 0);
+            return references;
+        }
```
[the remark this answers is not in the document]
nielsdebruin: The implementation still needs a reference to the `Matcher`, or a visitor...perhaps, getTraitMatcher method?

rewrite-properties/src/test/java/org/openrewrite/properties/trait/PropertiesReferenceTest.java
nielsdebruin: I think one can be resolved after the change Tim commited?

rewrite-properties/src/main/java/org/openrewrite/properties/tree/Properties.java
nielsdebruin: Fixed

rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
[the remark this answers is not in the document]
Laurens-W: Hmm didn't realize that when I commented, thinking about it again I don't think there's a clean way to make a default implementation without making lots of changes elsewhere

Laurens-W (approved):

timtebeek (approved):
Great to see! Approved already, but it's still marked as draft?

A quick glance from @knutwannheden would be appreciated, but not a hard requirement. Should we mark this as Incubating still?

nielsdebruin:
> Great to see! Approved already, but it's still marked as draft?
> 
> A quick glance from @knutwannheden would be appreciated, but not a hard requirement. Should we mark this as Incubating still?

That might be wise, the code be considered incubating, as most of it implements interfaces that are marked as incubation. I added the incubation annotation.

Laurens-W:
I still think since the `Reference` interface is marked as incubating inherently the implementations are also subject to change and therefore might not have to specifically be annotated with incubating

nielsdebruin:
@Laurens-W I have no preference either way, and it's unclear what the convention is within this project. The `SimpleTraitMatcher`, extends 'TraitMatcher` and implements `Trait`. While the latter two are already marked incubating `SimpleTraitMatcher` is also marked incubating...

timtebeek (approved):

PR #4710, opened 2024-11-24
RFC: LST contains missing or invalid type information reproducer

signed:
## What's changed?

I'm working on a Recipe to replace constructor calls with a static factory method.
My test is failing with
````
LST contains missing or invalid type information
MethodInvocation->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(MethodInvocation type is missing or malformed)~~>*/createHello("Alice", 42L)
https://docs.openrewrite.org/reference/faq#im-seeing-lst-contains-missing-or-invalid-type-information-in-my-recipe-unit-tests-how-to-resolve
MethodInvocation->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(MethodInvocation type is missing or malformed)~~>*/createHello("Alice", 42L)
````
When I turn of Type Validation, everything works as expected.

## Any additional context

To me it looks like a Bug in the JavaTemplate/TestRunner, but it is my first Recipe. Maybe I'm making a stupid mistake.
I debugged into the Template code but got lost. If it is indeed a bug I'd appreciate a pointer to the code that needs adjusting and I'l give it another try.

### CODE REVIEW

File: rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest.java [hunk 1]
```diff
@@ -428,6 +428,93 @@ void foo() {
...
+                        final J.ClassDeclaration classDeclaration = parent.firstEnclosing(J.ClassDeclaration.class);
+                        final var isInstantiationInFactoryMethod = FactoryMethod.matches(methodDeclaration, classDeclaration);
+                        if (isInstantiationInFactoryMethod) {
+                            return result;
+                        }
+
+                        final var template = JavaTemplate.builder("createHello(#{any()}, #{any()})")
+                          .doBeforeParseTemplate(System.out::println)
+                          .staticImports("f.HelloFactory.createHello")
+                          .build();
```
timtebeek, line 453: Hi! Any JavaTemplate is compiled in isolation, without any references to existing classpath entries; usually these are provided in the form of `.classpath` or `.classpathFromResources` method calls, but in other cases you can also use a minimized `.javaParser` using `.dependsOn`. In your case that would look something like this:
```suggestion
                        final var template = JavaTemplate.builder("createHello(#{any()}, #{any()})")
                          .doBeforeParseTemplate(System.out::println)
                          .javaParser(JavaParser.fromJavaVersion().dependsOn(
                            """
                              package a;
                              public class Hello {
                                public Hello(String name, long value) {}
                              }
                              """,
                            """
                              package f;
                              import a.Hello;
                              public class HelloFactory {
                                public static Hello createHello(String name, long value) {
                                  return new Hello(name, value);
                                }
                              }
                              """
                          ))
                          .staticImports("f.HelloFactory.createHello")
                          .build();
```

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest.java
```diff
@@ -428,6 +428,93 @@ void foo() {
...
+
+                        maybeRemoveImport("a.Hello");
+                        maybeAddImport("f.HelloFactory", "createHello", false);
+
+                        final var arguments = result.getArguments();
+                        final var methodInvocation = template.apply(getCursor(), result.getCoordinates().replace(), arguments.get(0), arguments.get(1));
+                        if (null != ((J.MethodInvocation) methodInvocation).getMethodType()) {
+                            throw new RuntimeException("JavaTemplate does not attach the MethodType. IMO, it should.");
+                        }
```
timtebeek, line 462: The conditional there should likely be `==`; but even then this check is not necessary: there's an implied check for missing types that runs with every unit test through `rewriteRun`.
```suggestion
```

timtebeek (commented):
hi @signed ; I've added two comments with feedback on how to get your test going; I do not see any issues with OpenRewrite itself. With these pointers you'll be able to get your recipe going. There's some docs to look over here as well:
- https://docs.openrewrite.org/concepts-and-explanations/javatemplate#referencing-external-types
- https://docs.openrewrite.org/authoring-recipes/modifying-methods-with-javatemplate

I hope that helps! With that I think it's ok to close this PR; feel free to ask any further questions in [our Slack](https://join.slack.com/t/rewriteoss/shared_invite/zt-nj42n3ea-b~62rIHzb3Vo0E1APKCXEA)!

PR #4713, opened 2024-11-25
Refactor references updating

nielsdebruin:
## What's changed?

This PR introduces an initial attempt at refactoring the updating references for properties and YAML files. The effect is both improved performance and decoupling.

## What's your motivation?
A `Reference` has a "value" of type `String`. However, implementation-wise, the underlying `Reference` only has a `Cursor`, from which we retrieve the `Tree`. The specific implementation of the `Reference` interface will map this to a `String` depending on the type which becomes the value. 

For example, in the `Properties` reference, this would look like:

```java
@Override
public @NonNull String getValue() {
    if (getTree() instanceof Properties.Entry) {
        return ((Properties.Entry) getTree()).getValue().getText();
    }
    throw new IllegalArgumentException("getTree() must be a Properties.Entry: " + getTree().getClass());
}
```

Currently, the specific implementations for different types of references are handled in the TypeMatcher and PackageMatcher classes. This leads to code duplication and increased coupling. Moreover, both classes generate a new visitor for each reference update, which is inefficient. This refactor moves the specific implementation to the `Reference` interface by introducing a `withValueFromString(String newValue)` method in Reference. This method can be used when implementing the newly introduced `ReferenceUpdateFunction`.

## Anyone you would like to review specifically?

@timtebeek @Laurens-W

### Checklist
- [] I've added unit tests to cover both positive and negative cases
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

PR #4714, opened 2024-11-25
Refactor references updating

knutwannheden:
Alternative to #4713

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
+++ b/rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
@@ -31,16 +31,16 @@ public interface Reference extends Trait<Tree> {
 
     String getValue();
 
-    default boolean supportsRename() {
-        return false;
-    }
-
     default boolean matches(Matcher matcher) {
         return matcher.matchesReference(this);
     }
 
-    default TreeVisitor<Tree, ExecutionContext> rename(Renamer renamer, String replacement) {
-        return renamer.rename(replacement);
+    default boolean supportsRename() {
+        return false;
+    }
+
+    default Tree rename(Renamer rename, Cursor cursor, ExecutionContext ctx) {
+        return cursor.getValue();
     }
 
     interface Provider {
@@ -52,11 +52,13 @@ public interface Reference extends Trait<Tree> {
 
     interface Matcher {
         boolean matchesReference(Reference value);
-    }
-
-    interface Renamer {
-        default TreeVisitor<Tree, ExecutionContext> rename(String replacement) {
+        default Renamer renameTo(String newName) {
             throw new UnsupportedOperationException();
         }
+    }
+
+    @FunctionalInterface
+    interface Renamer {
+        String rename(String oldName);
     }
 }
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
@@ -148,7 +148,7 @@ public class ChangePackage extends Recipe {
                     for (Reference ref : references.findMatches(matcher)) {
                         matches.put(ref.getTree(), ref);
                     }
-                    return new ReferenceChangePackageVisitor(matches, matcher, newPackageName).visit(tree, ctx, requireNonNull(getCursor().getParent()));
+                    return new ReferenceChangePackageVisitor(matches, matcher.renameTo(newPackageName)).visit(tree, ctx, requireNonNull(getCursor().getParent()));
                 }
                 return tree;
             }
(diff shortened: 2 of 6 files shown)


rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
```diff
@@ -52,11 +52,13 @@ interface Provider {
...
+            throw new UnsupportedOperationException();
+        }
     }
 
+    @FunctionalInterface
     interface Renamer {
-        default TreeVisitor<Tree, ExecutionContext> rename(String replacement) {
-            throw new UnsupportedOperationException();
-        }
+        String rename(String oldName);
```
nielsdebruin, line 62: I think that perhaps extending this to `String rename(Reference reference)`  would be better here, currently we only need access to the String value, but a future Renamer might also need the value of `Kind`, or any future properties that will be added.

rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
```diff
@@ -52,11 +52,13 @@ interface Provider {
 
     interface Matcher {
         boolean matchesReference(Reference value);
+        default Renamer renameTo(String newName) {
```
nielsdebruin, line 55: This method return a new `Renamer` but does not rename things. I this a better name would be `getRenamer` or `getReferenceRenamer`.

rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
```diff
@@ -31,16 +31,16 @@ enum Kind {
...
     }
 
-    default TreeVisitor<Tree, ExecutionContext> rename(Renamer renamer, String replacement) {
-        return renamer.rename(replacement);
+    default boolean supportsRename() {
+        return false;
+    }
+
+    default Tree rename(Renamer rename, Cursor cursor, ExecutionContext ctx) {
+        return cursor.getValue();
```
nielsdebruin, line 43: Unlike the previous version, the variants will never work unless it's overridden. I think the default implementation should just contain throw a UnsupportedOperationException.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringReference.java
```diff
@@ -62,6 +65,21 @@ public boolean supportsRename() {
         return true;
     }
 
+    @Override
+    public Tree rename(Renamer rename, Cursor cursor, ExecutionContext ctx) {
```
nielsdebruin, line 69: I think the name of this function is still a bit confusing, as this function does not make any modification to the references itself, nor does it return a new reference, effectively, the method can be static, as the cursor is used.

nielsdebruin (changes requested):

rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
knutwannheden: Sure, let's use Reference instead. The implementation can then call getValue() on that.

knutwannheden:
@nielsdebruin Do you want to take over again?

rewrite-core/src/main/java/org/openrewrite/trait/Reference.java
nielsdebruin: My thoughts exactly

nielsdebruin:
@knutwannheden Sure!

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -390,13 +390,12 @@ private boolean isTargetRecursivePackageName(String packageName) {
     private static class ReferenceChangePackageVisitor extends TreeVisitor<Tree, ExecutionContext> {
         Map<Tree, Reference> matches;
         Reference.Renamer renamer;
-        String newPackageName;
 
         @Override
         public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 395: I think we should change this to be a `postVisit()` method instead. That way the cursor available to the `Reference` should still be valid while we call the `rename()` method.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -560,13 +560,12 @@ private boolean hasNoConflictingImport(@Nullable JavaSourceFile sf) {
     private static class ReferenceChangeTypeVisitor extends TreeVisitor<Tree, ExecutionContext> {
         Map<Tree, Reference> matches;
         Reference.Renamer renamer;
-        String newFullyQualifiedName;
 
         @Override
         public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
```
knutwannheden, line 565: Same here.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringReference.java
```diff
@@ -62,6 +65,24 @@ public boolean supportsRename() {
         return true;
     }
 
+    /**
+     * {@inheritDoc}
+     */
```
knutwannheden, line 70:
```suggestion
```

PR #4717, opened 2024-11-26
Skip `project` dependencies in `UpgradeDependencyVersion` recipe

knutwannheden:
Project dependencies don't take a GAV as an argument and could cause an exception to be thrown by `DependencyStringNotationConverter#assignValue()` if interpreted as GAVs.


knutwannheden:
@shanman190 I didn't know / look if there are other recipes which need to be modified in a similar way. I found this problem in the recipe results for a Spring project.

timtebeek (approved):

Issue #4703, opened 2024-11-23
Groovy parser does not correctly handle nested parenthesis


sambsnyd:
When we attempt to parse:
```groovy
def foo(Map map) {
    ((map.containsKey("foo"))
            && ((map.get("foo")).equals("bar")))
}

def timestamp(int hours, int minutes, int seconds) {
    (hours) * 60 * 60 + (minutes * 60) + seconds
}
```

What we actually get is an `IllegalStateException` out of the parser.

Reproduced in [BinaryTest.extraParensAroundInfixOperator()](https://github.com/openrewrite/rewrite/blob/652654e57c5387c93c0c7d844b957c29ced52871/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/BinaryTest.java#L207)

timtebeek:
Half of this has already been fixed through
- https://github.com/openrewrite/rewrite/pull/4751

jevanlingen:
Other half has been fixed today: 
- https://github.com/openrewrite/rewrite/pull/4801

Issue #4615, opened 2024-10-26
Gradle parser fails with extra parens around an expression


mccartney:
## What is the smallest, simplest way to reproduce the problem?
```
    @Test
    void parensAroundAnExpression() {
        rewriteRun(
          buildGradle(
            """
              plugins {
                  id 'java-library'
              }
              def version = (rootProject.jobName.startsWith('a')) ? "latest.release" : "3.0"
              """
          )
        );
    }
```

## What is the full stack trace of any errors you encountered?
```
java.lang.AssertionError: Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser. 
plugins {
    id 'java-library'
}
def version = ~~(non-whitespace)~~>(<~~rootProject.jobName.startsWith('a')~~(non-whitespace)~~>) <~~? "latest.release" : "3.0"
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:323)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:132)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:127)
	at org.openrewrite.gradle.GradleParserTest.parensAroundAnExpression(GradleParserTest.java:199)
```

## Context
It works fine when removing the admittedly excessive parentheses.

shanman190:
@mccartney, what version of OpenRewrite are you using? I could have sworn that this was fixed a while ago in the Groovy parser.

mccartney:
@shanman190 I have just reproduced the test failure with current main, i.e. 36efd20f640902d87926945e5748bb1f4039b7a6

knutwannheden:
Looks like the Groovy parser version we are using is somehow not adding that `_INSIDE_PARENTHESES_LEVEL` metadata around the condition. This is what we rely on to do the parentheses parsing correctly.

@shanman190 AFAIK Gradle uses Groovy 3 in build scripts and I assume it will stay that way. I think I remember that Groovy 4 should be mostly backwards compatible with Groovy 3 (some deprecated API was removed). Would we be able to use the Groovy 4 parser?

shanman190:
[...]

PR #4718, opened 2024-11-26
Make Groovy Parser correctly handle nested parenthesis

jevanlingen:
## What's changed?
Nested parentheses around method invocations are supported as well for the groovy parser.

## What's your motivation?

### Checklist


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1664,9 +1667,13 @@ public void visitMethodCallExpression(MethodCallExpression call) {
                     Expression selectExpr = visit(call.getObjectExpression());
```
jevanlingen, line 1667: I did some change below this line, but I have the feeling I am at the wrong track. For instance with following groovy code:

```groovy
(map.get("foo")).equals("bar")`
```

I can retrieve here an expression that looks printed like `(map.get("foo")`. I think you would expect it evenly distributed like `(map.get("foo"))`. So maybe that's what should be changed instead of the code here below.

jevanlingen: I stop working on this for now. I can't wrap my head around how the parentheses are set. I probably need a little bit more experience before I start working on parser code 😇.

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/BinaryTest.java
```diff
@@ -201,17 +215,12 @@ void stringMultipliedInParentheses() {
...
     }
 
-    @Disabled
     @Issue("https://github.com/openrewrite/rewrite/issues/4703")
     @Test
     void extraParensAroundInfixOperator() {
         rewriteRun(
           groovy(
             """
-              def foo(Map map) {
```
Laurens-W, line 211: is this not relevant context for this test case?

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodInvocationTest.java
```diff
@@ -262,4 +262,31 @@ static void main(String[] args) {
...
+              static def foo(Map map) {
+                  ((map.containsKey("foo"))
+                      && ((map.get("foo")).equals("bar")))
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/4703")
```
Laurens-W, line 281: Don't think this example came from this issue?

rewrite-java/src/main/java/org/openrewrite/java/tree/Space.java
```diff
@@ -286,13 +286,17 @@ public String toString() {
...
+                    // should never happen (probably a bug in the parser)
+                    printedWs.append(c);
                 }
             }
         }
+        String whitespaces = printedWs.toString();
 
         return "Space(" +
-               "comments=<" + (comments.size() == 1 ? "1 comment" : comments.size() + " comments") +
-               ">, whitespace='" + printedWs + "')";
```
Laurens-W, line 295: Do agree that looking at `whitespace='')"` can be very confusing
The first are two single quotes

Laurens-W (approved):
LGTM but since it's an extensive change let's let Tim, Sam or Knut take a look too!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyPrinter.java
```diff
@@ -24,6 +24,7 @@
 import org.openrewrite.groovy.tree.GContainer;
 import org.openrewrite.groovy.tree.GRightPadded;
 import org.openrewrite.groovy.tree.GSpace;
+import org.openrewrite.internal.StringUtils;
```
timtebeek, line 27:
```suggestion
```

timtebeek: The reminded me of [The Queen's Duck](https://bwiggs.com/notebook/queens-duck/); not sure if it was intended that way. ;)

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/RangeTest.java
```diff
@@ -39,9 +39,9 @@ void parenthesized() {
         rewriteRun(
           groovy(
             """
-              ( 8..19 ).each { majorVersion ->
+              (((( 8..19 ))).each { majorVersion ->
                 if (majorVersion == 9) return
-              }
+              })
```
timtebeek, line 44: I'm surprised we're changing the original here as well; would it make sense to add a second `groovy` block within the same `rewriteRun` instead?

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/CastTest.java
```diff
@@ -73,7 +73,7 @@ void groovyCastAndInvokeMethod() {
         rewriteRun(
           groovy(
             """
-              ( "" as String ).toString()
+              (((((( "" as String ))))).toString())
```
timtebeek, line 76: also here: perhaps best to keep both the original and the unexpected case you've modified for your purposes.

rewrite-groovy/src/test/java/org/openrewrite/groovy/GroovyParserParentheseDiscovererTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        assertEquals(1, result);
+    }
+
+    @Test
+    void invokeWithArguments() {
+        MethodCallExpression node = MockMethodCallExpression.of("invoke");
+
+        //language=groovy
+        String source = "(a.invoke(\"A\", \"\\$\", \"C?)\"))";
```
timtebeek, line 46: If you're having to escape the double quotes then a text block might make more sense.
```suggestion
        String source = """
          (a.invoke("A", "\\$", "C?)"))
          """;
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyPrinter.java
jevanlingen: No, it was a remnant of older try-outs. Should be remove indeed.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -604,50 +579,39 @@ null, emptyList(),
         }
 
         @Override
-        public void visitConstructor(ConstructorNode ctor) {
+        public void visitConstructor(ConstructorNode constructor_) {
```
timtebeek, line 582: Usually we reserve the `_` suffix when it concerns a Java keyword; constructor is not a keyword; what made you add that suffix here?

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodInvocationTest.java
jevanlingen: Nope, different issue but very same reason. So I marked the issue as duplicate and put given test in the PR.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2136,19 +2078,21 @@ public TypeTree visitVariableExpressionType(VariableExpression expression) {
...
-                type = typeMapping.type(staticType((org.codehaus.groovy.ast.expr.Expression) expression));
-            }
-            queue.add(new J.Identifier(randomId(),
-                    sourceBefore(expression.getName()),
-                    Markers.EMPTY,
-                    emptyList(),
-                    expression.getName(),
-                    type, null)
-            );
+            queue.add(insideParentheses(expression, fmt -> {
```
timtebeek, line 2081: Calling out this one for myself as the 'new' part of the changes in this class.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: I really don't know

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserParentheseDiscoverer.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Pattern PARENTHESES_GROUP = Pattern.compile(".*?(\\(+[^()]+\\)+).*", Pattern.DOTALL);
+    // Matches when string consist of multiple parenthese blocks
+    // Eg: `((a.invoke())` does not match, `((a.invoke() && b.invoke())` does match
+    private static final Pattern HAS_SUB_PARENTHESIS = Pattern.compile("\\(+[^)]+\\)+[^)]+\\).*", Pattern.DOTALL);
+
+    /**
+     * Calculates the insideParenthesesLevel for method calls, because the compiler does not set the _INSIDE_PARENTHESES_LEVEL flag for method calls.
+     */
+    public static @Nullable Integer getInsideParenthesesLevel(MethodCallExpression node, String source, int cursor) {
+        String sourceFromCursor = source.substring(cursor);
```
timtebeek, line 37: I'm somewhat surprised to see `source` and `cursor` as separate arguments here when they are merely used to do this substring. Would it make sense to change the method signature here to
```suggestion
    public static @Nullable Integer getInsideParenthesesLevel(MethodCallExpression node, String sourceFromCursor) {
```

jevanlingen: I had that before, but reverted it back to current state. I think it more clear if the _GroovyParserParentheseDiscoverer_ is a completely separate entity. So no former logic is needed to determine the wrapping. Just pass everything you have (thus the node, the source code of the entire file and the current place you are parsing right now) and let the function discover it all.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserParentheseDiscoverer.java
```diff
@@ -0,0 +1,90 @@
...
+package org.openrewrite.groovy;
+
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.internal.StringUtils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GroovyParserParentheseDiscoverer {
```
timtebeek, line 25: Nit:
```suggestion
class GroovyParserParentheseDiscoverer {
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserParentheseDiscoverer.java
```diff
@@ -0,0 +1,90 @@
...
+package org.openrewrite.groovy;
+
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.internal.StringUtils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class GroovyParserParentheseDiscoverer {
```
sambsnyd, line 25: I haven't reviewed in detail but at first glance it doesn't seem to be aware of comments so any parentheses in comments might throw it off

shanman190:
This is just my 2c, but the several regexes makes me feel like the solution here isn't quite right. For trivial source files found in the test cases, it works. But I fear that for actual source files that this would bring G-based parsing to a crawl.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserParentheseDiscoverer.java
knutwannheden: I wonder if this can be simplified. For Python I had the same basic problem to solve, but I didn't have any "parentheses level" from the AST or similar. Additionally, in Python the parentheses for tuples and generator expressions are sometimes optional, which doesn't make things simpler.

Anyway, rather than counting the "parentheses levels" up front, I added something like a `maybeParenthesized(node, delegate)` method which does something like this:
1. Parse whitespace (and comments) at cursor into `Space`
2. Check if the next character is a `(`
3. If not, reset cursor to before whitespace and return result from `delegate` and we are done
4. Otherwise push a lambda on a parentheses stack, which is a field on the parser visitor
5. Call the delegate
6. Consume whitespace and check if we are at a `)`
7. If yes, pop lambda from parentheses stack and apply it to result, which will wrap it in a `J.Parentheses`

For a language like Java or Python, where we can expect the source files to be reasonably large, I would definitely do something like this, as the substringing and regexp matching won't be free in terms of performance. But for Groovy that would probably not be a big concern, as scripts tend to be small and few.

I will let you think a bit about this and don't want to force you into a different solution.

sambsnyd: I second Knut's comments and like the approach he describes. I worry that these regexes will be slow and error prone.

knutwannheden: The one tricky situation I can think of are Java-style type casts, as the first token there is also a `(`. So if the AST node is a Java-style type cast (assuming they can be distinguished from Groovy-style type casts in the AST), the logic could for example pop that last element from the stack and reset the cursor before calling the corresponding visit method (or do it in the method itself). There are also other alternative solutions.

jevanlingen: I agree, your implementation seems like a better idea @knutwannheden. I would rather first merge this and then do a second round with improvements (or better said, go for the new/better implementation). Currently we don't support nested parenthesis, which just breaks scripts that use them. Having this sub-ideal implementation makes them at least runnable, though with a performance loss. 

Then I can start working on the better implementation.

knutwannheden: It isn't quite as easy as I made it out to be and I now remember the struggles I had with Python here. The difficulty arises because we don't know at what AST level the parentheses are. Here that parentheses level may come in handy.

timtebeek:
- Fixed in https://github.com/openrewrite/rewrite/pull/4801

PR #4720, opened 2024-11-26
Replace `UnaryOperator` with `Function` in `ListUtils` methods

knutwannheden:
The reason for this change is that the mapping functions should all expect a non-`null` input but are allowed to return `null` to signal that the corresponding result is to be removed from the list.

Additionally, the return type of the methods has been annotated with `@Nullable` where appropriate (typically when the input `List` is `null`). So that this doesn't result in new warnings getting reported, the methods have now additionally also been annotated with `@Contract`, using which we capture the details of when the method returns `null`. With powerful static analysis tools like IDEA, this should typically not result in any new unwanted warnings.

For example when a `with`-method is called by mapping the result of the corresponding `get`-method this is not an issue, as both the `get`-method return type and the `with`-method parameter are annotated as `@Nullable`.

> [!IMPORTANT]  
> This PR breaks binary compatibility and must therefore be part of a minor release.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.internal;
 
+import org.jetbrains.annotations.Contract;
 import org.jspecify.annotations.Nullable;
 
 import java.util.ArrayList;
@@ -22,7 +23,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-import java.util.function.UnaryOperator;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
@@ -117,7 +117,8 @@ public final class ListUtils {
      * @param <T>      The type of elements in the list.
      * @return A new list with the modified last element, or the original list if unchanged.
      */
-    public static <T> List<T> mapLast(@Nullable List<T> ls, UnaryOperator<@Nullable T> mapLast) {
+    @Contract("null, _ -> null; !null, _ -> !null")
+    public static <T> @Nullable List<T> mapLast(@Nullable List<T> ls, Function<T, @Nullable T> mapLast) {
         if (ls == null || ls.isEmpty()) {
             //noinspection ConstantConditions
             return ls;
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
```diff
@@ -355,6 +358,7 @@ public static <T> List<T> concat(@Nullable List<T> ls, @Nullable T t) {
      * @return A new list with the added element at the start, the original list if the element is null or a null
      * object if both element and list are null.
      */
+    @Contract("null, null -> null; !null, _ -> !null; _, !null -> !null")
     public static <T> @Nullable List<T> concat(@Nullable T t, @Nullable List<T> ls) {
```
jevanlingen, line 362: We could consider going one step further and also annotated the generic in the list being nullable:

```suggestion
    public static <T> @Nullable List<T> concat(@Nullable T t, @Nullable List<@Nullable T> ls) {
```

rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
```diff
@@ -378,12 +382,11 @@ public static <T> List<T> concat(@Nullable List<T> ls, @Nullable T t) {
      * @param <T>  The type of elements in the lists.
      * @return A new list containing both lists, or one of the lists if the other is null.
      */
+    @Contract("null, null -> null; !null, _ -> !null; _, !null -> !null")
     public static <T> @Nullable List<T> concatAll(@Nullable List<T> ls, @Nullable List<? extends T> t) {
```
jevanlingen, line 386: Same as above:

```suggestion
    public static <T> @Nullable List<T> concatAll(@Nullable List<@Nullable T> ls, @Nullable List<? extends @Nullable T> t) {
```

rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
knutwannheden: Not sure. We don't actually filter out any nulls and we typically don't want to have nulls in our lists.

jevanlingen: Agreed, let's keep it this way. Now I think about it, you are right. We don't want nulls in our lists, so let's keep it tight (so non null!).

rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
jevanlingen: Nope, leave it as it is.!

jevanlingen (approved):

PR #4724, opened 2024-11-27
Exclude omitted dependencies as well through a second cycle

timtebeek:
## What's changed?
Run a second cycle of ExcludeDependency when the document has changed, to also exclude previously omitted dependencies.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/pull/4563

Second cycle might require changes in the future, but at that point we're more free to change the model as well, as seen in:
 - https://github.com/openrewrite/rewrite/pull/4563

@siborea @ckcd 

Credits for the test here go to @ckcd 🙏🏻 

We can change the model as seen in this work
 - https://github.com/openrewrite/rewrite/pull/4563

But that would then need careful consideration to account for serialization, and we'd be doing so for a single use case here.


ckcd:
@timtebeek Thanks for your deep check and fixing! Nice to see the issue fixed.

Issue #4726, opened 2024-11-27
Git Provenance generates a warning when a GitHub organization happens to start with a number


tylervangorder-8451:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.38.0
- Maven plugin v5.43.4

## How are you running OpenRewrite?

We are running OpenRewrite via the maven plugin.
 
This happens across any of our repositories that have been checked out via SSH: 

Our organization name is `8451LLC` which it appears is tripping up Git Provenance parsing: 

```
java.lang.IllegalArgumentException: Unable to normalize URL: Specifying a port without a scheme is not supported for URL git@github.com:8451LLC/any-repo-in-our-org.git
	at org.openrewrite.GitRemote$Parser.normalize(GitRemote.java:305)
	at org.openrewrite.GitRemote$Parser.parse(GitRemote.java:223)
	at org.openrewrite.marker.GitProvenance.fromGitConfig(GitProvenance.java:259)
	at org.openrewrite.marker.GitProvenance.fromGitConfig(GitProvenance.java:241)
	at org.openrewrite.marker.GitProvenance.fromProjectDirectory(GitProvenance.java:224)
	at org.openrewrite.marker.GitProvenance.fromProjectDirectory(GitProvenance.java:177)
	at org.openrewrite.maven.MavenMojoProjectParser.lambda$gitProvenance$18(MavenMojoProjectParser.java:795)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1134)
	at org.openrewrite.maven.MavenMojoProjectParser.gitProvenance(MavenMojoProjectParser.java:795)
	at org.openrewrite.maven.MavenMojoProjectParser.generateProvenance(MavenMojoProjectParser.java:259)
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:178)
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
[...]

PR #4727, opened 2024-11-27
Allow orgs with leading numbers in name in GitRemote

timtebeek:
## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/4726

@tylervangorder-8451

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -279,7 +279,7 @@ public class GitRemote {
                     .replaceFirst("^/", "");
         }
 
-        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+");
+        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+[^-\\w]"); // Allow orgs with numbers in name
 
         static URI normalize(String url) {
             try {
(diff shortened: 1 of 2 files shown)


tylervangorder-8451 (commented):
Thanks for the quick turnaround

timtebeek:
You're welcome! You did _just_ miss the release we did today, but I imagine you're comfortable running snapshot versions if needed. All the best!

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -279,7 +279,7 @@ private String repositoryPath(RemoteServerMatch match, URI normalizedUri) {
                     .replaceFirst("^/", "");
         }
 
-        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+");
+        private static final Pattern PORT_PATTERN = Pattern.compile(":\\d+[^-\\w]"); // Allow orgs with numbers in name
```
bryceatmoderne, line 282: github at least supports org names with only numbers.

timtebeek: Guess those will still get the exception thrown then, but at least this works for partial numeric org names. Welcome suggestions for a complete fix!

bryceatmoderne: > Guess those will still get the exception thrown then, but at least this works for partial numeric org names. Welcome suggestions for a complete fix!

I had opened this previously https://github.com/openrewrite/rewrite/pull/4651

timtebeek: Ah, I had no idea. Approved just now & sorry about the conflict. Do let me know if there's ever something open for too long.

PR #4728, opened 2024-11-28
MergeYaml does not process multiline scalar value blocks properly when multiline block is located in a subkey

jevanlingen:
## What's changed?
Improve formatting of multiline scalar values.

## What's your motivation?
Since 
- https://github.com/openrewrite/rewrite/pull/4358

multiline scalar value blocks should work now, but it did not work properly when the key of the scalar is located deeper in the tree like:

```
deep:
  object:
    script: |
       #!/bin/bash
       echo "hello"
```

Secondly, there was no support for values with different levels of indentation, which is valid in Yaml is as well:
```
script: |
  #!/bin/bash
         echo "hello"
```

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
@@ -327,12 +327,12 @@ public class MergeYamlVisitor<P> extends YamlVisitor<P> {
         return !s1.equals(s2) && !acceptTheirs ? y1.withValue(s2) : y1;
     }
 
-    public String substringOfBeforeFirstLineBreak(String s) {
+    private String substringOfBeforeFirstLineBreak(String s) {
         String[] lines = LINE_BREAK.split(s);
         return lines.length > 0 ? lines[0] : "";
     }
 
-    public String substringOfAfterFirstLineBreak(String s) {
+    private String substringOfAfterFirstLineBreak(String s) {
         String[] lines = LINE_BREAK.split(s);
         return lines.length > 1 ? String.join(linebreak(), Arrays.copyOfRange(lines, 1, lines.length)) : "";
     }


rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
```diff
@@ -720,4 +720,19 @@ public static String formatUriForPropertiesFile(String uri) {
...
+
+    public static String substringOfAfterFirstLineBreak(String s) {
+        String[] lines = LINE_BREAK.split(s);
+        return lines.length > 1 ? String.join("\n", Arrays.copyOfRange(lines, 1, lines.length)) : "";
+    }
+
+    public static String substringOfAfterLastLineBreak(String s) {
+        String[] parts = LINE_BREAK.split(s);
+        return parts.length > 1 ? parts[parts.length - 1] : "";
+    }
```
timtebeek, line 737: Let's not yet pull up these single use methods to the more widely used StringUtils. Even if it's internal, it's still public and exposed, whereas these can be private to their single usage site.

jevanlingen: 👍🏻, I removed them!

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -317,19 +321,26 @@ private Yaml.Sequence mergeSequence(Yaml.Sequence s1, Yaml.Sequence s2, P p, Cur
...
+
+    public String substringOfBeforeFirstLineBreak(String s) {
+        String[] lines = LINE_BREAK.split(s);
+        return lines.length > 0 ? lines[0] : "";
+    }
+
+    public String substringOfAfterFirstLineBreak(String s) {
+        String[] lines = LINE_BREAK.split(s);
+        return lines.length > 1 ? String.join(linebreak(), Arrays.copyOfRange(lines, 1, lines.length)) : "";
+    }
```
timtebeek, line 338: Nitpicking here; but these do not have to be public right? And they are the same as the methods above? Can we minimize the diff and exposed API please?

PR #4729, opened 2024-11-28
Restore ListUtils map methods that accept UnaryOperator

timtebeek:
We're understandably getting reports that older recipe jars now can not run, following
- #4720

This restores those methods for now, but also locks us into new compilations again using the implied old methods. 🤔


pstreef (approved):
I assume you tested that cast? Looks like this will fix our problem!

PR #4735, opened 2024-12-01
Resove dependencies bom with properties on `ChangeParentPom`

jonesbusy:
~~This just demonstrate an issue we are facing with jenkins modernization.~~

The `ChangeParentPom` recipe wrongly add a dependency into dependency management when it should not

Such dependency is still managed by the bom.

```
    +            <dependency>
    +                <groupId>org.jenkins-ci.plugins</groupId>
    +                <artifactId>scm-api</artifactId>
    +                <version>698.v8e3b_c788f0a_6</version>
    +            </dependency>
```

I think it cause issue because the `${jenkins.baseline}` on the artifact id.

But I don't know how and the best way to fix it

Any help would be appreciated

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
@@ -68,11 +68,11 @@ public class PropertyPlaceholderHelper {
         return replacePlaceholders(value, properties::getProperty);
     }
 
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks for diving in here @jonesbusy ; in your commit message you've marked this as a tentative fix; should we hold off on a merge just yet?

jonesbusy:
Yes 2nd commit is the fix. I had some failure locally, but was probably related to my environment.

CI looks good so we can move forward

Issue #4740, opened 2024-12-04
Extend YAML CommentOutProperty with option to keep original property uncommented


ashakirin:
Currently YAML CommentOutProperty adds the comment and always comments out original property.
To be consistent with property AddPropertyComment recipe, introduce the option to keep original property and just add a comment.
In the future, it would be nice to align the names and parameters for YAML CommentOutProperty and property AddPropertyComment recipes

timtebeek:
- Fixed in https://github.com/openrewrite/rewrite/pull/4741
Thanks again!

PR #4741, opened 2024-12-04
feat: Added option to comment out property in yaml

ashakirin:
- Refs: #4740


timtebeek (approved):
Thanks for the addition and thorough tests! I've added some light polish to ensure any declarative usages do not fail with an NPE on the nullable field added.

ashakirin:
@timtebeek Thanks, Tim!
The only thing that worry me a bit, is that recipes for normal properties and for yaml doing the same things, but have two different names:
- yaml: CommentOutProperty
- properties: AddPropertyComment

Do you think would be possible to consolidate the names?
Andrei

Issue #4745, opened 2024-12-04
Preserve annotation values parsed from dependencies in the classpath


zexblue01:
## What problem are you trying to solve?
I’m working on a linter for detecting certain methods in our code base. The linting is based on method annotations. The annotation in question has a value attached which I need for making the linting decision. The method in question is only in the classpath, not part of the source files.

Dependency lib in the classpath:
```
@Target({ElementType.FIELD, ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
public @interface EssentialTracker {
  String value();
}

public class TestEssentialTracker {
 @EssentialTracker("aaa")
 public String getEssentialTracker() {
  return "test-essential-tracker";
 }
}
```

Source file input I'm linting:
```
String result = new TestEssentialTracker().getEssentialTracker();
```

In this case, I want to be able to access the value "aaa" in the recipe.

I did some experiments. It seems in the recipe I can access the annotation type but not the “value” because it was converted into `JavaType.FullyQualified` here without knowledge of the value: https://github.com/openrewrite/rewrite/blob/main/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17TypeMapping.java#L696

## Describe the solution you'd like
Preserve the annotation value pairs in the parsed LST.

## Have you considered any alternatives or workarounds?
I don't think there is a workaround with OpenRewrite. I worked around in my solution by loading all classes and use reflection to inspect the actual annotation values, which is a bit cumbersome.

## Additional context
n/a

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes, here is [the patch](https://github.com/openrewrite/rewrite/pull/4746) I worked on. I don't have good understanding of OpenRewrite implementation holistically though, would love your feedback.

knutwannheden:
Related to:
- #3504

PR #4746, opened 2024-12-04
Add annotation element values to type attribution model

zexblue01:
## What's changed?
This PR adds ability to preserve annotation values parsed by JavaParsers (I updated all existing Java parser versions: 8, 11, 17, 21).

It adds a new subclass `Annotation` of `FullyQualified` in `JavaType`. It delegates most functionality to annotation's `JavaType.FullyQualified` object but stores list of value attribute pairs as part of the structure in `ElementValue` objects (with the two implementations `SingleElementValue` and `ArrayElementValue`).

## What's your motivation?
In the recipe I'm working on I need to be able to access annotation values on the method invoked in the source file (method could be defined in a dependency in the classpath instead of being as part of the source file input).

- fix https://github.com/openrewrite/rewrite/issues/4745

Not sure if it is there already a model somewhere else for annotation I could use instead of defining a new subclass of FullyQualified.
Also the interface didn't change after this change. It is still returning `FullyQualified` objects when accessing annotations. I'm concerned that changing this will complicate this change but let me know if you feel differently.

@jkschneider 
@timtebeek

I don't think there is a workaround with OpenRewrite. I worked around in my solution by loading all classes and use reflection to inspect the actual annotation values, which is a bit cumbersome.

### Checklist
* [x]  I've added unit tests to cover both positive and negative cases
* [x]  I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
* [x]  I've used the IntelliJ IDEA auto-formatter on affected files


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17TypeMapping.java
```diff
@@ -705,7 +705,10 @@ private void completeClassSymbol(Symbol.ClassSymbol classSymbol) {
                 if (retention != null && retention.value() == RetentionPolicy.SOURCE) {
                     continue;
                 }
-                annotations.add(annotType);
+                List<JavaType.AnnotationValue> annotationValues = a.values.stream().map(attr -> new JavaType.AnnotationValue(
+                        methodDeclarationType(attr.fst, annotType), attr.snd.getValue().toString())).collect(Collectors.toList());;
```
timtebeek, line 709:
```suggestion
                        methodDeclarationType(attr.fst, annotType), attr.snd.getValue().toString())).collect(Collectors.toList());
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
```diff
@@ -716,7 +716,10 @@ private void completeClassSymbol(Symbol.ClassSymbol classSymbol) {
                 if (retention != null && retention.value() == RetentionPolicy.SOURCE) {
                     continue;
                 }
-                annotations.add(annotType);
+                List<JavaType.AnnotationValue> annotationValues = a.values.stream().map(attr -> new JavaType.AnnotationValue(
+                        methodDeclarationType(attr.fst, annotType), attr.snd.getValue().toString())).collect(Collectors.toList());;;
```
timtebeek, line 720:
```suggestion
                        methodDeclarationType(attr.fst, annotType), attr.snd.getValue().toString())).collect(Collectors.toList());
```

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -626,6 +628,83 @@ public static ShallowClass build(String fullyQualifiedName) {
...
+    @Data
+    class AnnotationValue {
+        private final Method method;
+        private final String value;
+    }
+
+    class Annotation extends FullyQualified {
+
+        public final List<AnnotationValue> values;
+        public final FullyQualified type;
```
timtebeek, line 640: I wonder if `@Delegate` might make sense here; did you explore that option?

timtebeek (commented):
Great to see already; some very minor feedback as a first pass through; I like how you've taken care to minimize your changes and immediately roll them out across the various parsers.

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
zexblue01: No, let me look into it.

zexblue01: https://projectlombok.org/features/experimental/Delegate

> Current status: negative - Currently we feel this feature will not move out of experimental status anytime soon, and support for this feature may be dropped if future versions of javac or ecj make it difficult to continue to maintain the feature.

Sounds like it is a experimental lombok feature that might get deprecated in future. Probably should avoid using that?

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -626,6 +628,83 @@ public static ShallowClass build(String fullyQualifiedName) {
         }
     }
 
+    @Data
+    class AnnotationValue {
+        private final Method method;
+        private final String value;
```
knutwannheden: Can we store the "actual" values rather than a string representation of it?

knutwannheden (commented):
Looking very good already! I have quite a few use cases in mind for this, actually. I think we have some issues I can link to this PR (I will try to dig them out).

I added some review comments and also took the liberty to push a commit to the PR to store the "actual" annotation values, rather than a stringified representation of them, as I suspect that could make them a lot more useful. This work definitely isn't complete yet. Also, feel free to revert if you disagree.

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
knutwannheden: Now I understand why we implement `FullyQualified`: It is to remain compatible with the existing API. That is indeed a good reason.

rewrite-java-test/src/main/java/org/openrewrite/java/JavaTypeMappingTest.java
```diff
@@ -223,12 +223,16 @@ default void enumTypeB() {
     @Test
     default void ignoreSourceRetentionAnnotations() {
```
timtebeek, line 224: With the logic changes above this test name no longer reflects what's asserted. These tests seem to have been added by @sambsnyd in https://github.com/openrewrite/rewrite/commit/8f6f78ca to
> "Exclude annotations with Source retention from Java TypeMapping so as to facilitate deduplication of types from different mapping implementations"

Might be good to double check that there's no harm in adding these `RetentionPolicy.SOURCE` annotations now, or restore the original behaviour of skipping those, unless you specifically need them.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17TypeMapping.java
```diff
@@ -692,22 +691,65 @@ private void completeClassSymbol(Symbol.ClassSymbol classSymbol) {
...
+            }
+            return list;
+        } else if (value instanceof String) {
+            return value;
+        } else if (value instanceof Boolean) {
+            return value;
+        } else if (value instanceof Integer) {
+            return value;
+        }
+        return value;
```
zexblue01, line 753: Thanks @knutwannheden! This looks much better with the actual annotation value!
Is there a reason this change is not applied to Java 8 and 11 parsers? I only need it for 17 (and possibly 21 later) but I'm curious.

knutwannheden: No reason other than I typically only do it for one parser first and then once I geel I am ready I update the other parsers 😇

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -629,6 +631,112 @@ public static ShallowClass build(String fullyQualifiedName) {
...
+        }
+
+        @Override
+        public String getFullyQualifiedName() {
+            return type.getFullyQualifiedName();
+        }
+
+        @Override
+        public FullyQualified withFullyQualifiedName(String fullyQualifiedName) {
+            return this;
```
Laurens-W, line 659: Should this not delegate to type?

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -629,6 +631,112 @@ public static ShallowClass build(String fullyQualifiedName) {
         }
     }
 
+    class Annotation extends FullyQualified {
```
Laurens-W, line 634:
```suggestion
    @Getter
    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
    class Annotation extends FullyQualified {
```

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -629,6 +631,112 @@ public static ShallowClass build(String fullyQualifiedName) {
...
+
+        private final FullyQualified type;
```
Laurens-W, line 636:
```suggestion
        @With
        private final FullyQualified type;
```

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -629,6 +631,112 @@ public static ShallowClass build(String fullyQualifiedName) {
...
+        private final ElementValue @Nullable [] values;
```
Laurens-W, line 637:
```suggestion
        @With
        private final ElementValue @Nullable [] values;
```

sambsnyd:
This broadly looks OK to me. Make sure that `JavaTypeVisitor` can access any types nested within these new annotation fields or it will mess up LST serialization/deserialization.

PR #4747, opened 2024-12-05
Fix AddImport match variable's name

SiBorea:
## What's changed?
The origin code is matching the `JavaType.Variable.getName()` with the static import member, which doesn't make any sense. And it also doesn't handle the Parameterized reference.


timtebeek (approved):
Thanks a lot for the fix here @SiBorea ; this will solve quite a few puzzling forced import issues I imagine.

PR #4750, opened 2024-12-05
Update get classpath by artifact name to support bazel

amishra-u:
## What's changed?
Added support to retrieve the classpath by artifact name for Bazel projects.

## What's your motivation?
In Bazel projects, creating an instance of JavaParser with artifact names as classpath can fail when artifact names include a "-". This issue arises because Bazel replaces "-" with "_" when generating JAR files.

Example usage
```
JavaParser.fromJavaVersion().classpath("aws-java-sdk-s3").build()
```
Above code fails with `IllegalStateException Unable to construct JavaParser`.

To address this, updated JarPattern to include the normalized artifact name, ensuring it can correctly locate the JAR file names generated using bazel convention.

Note: A minor refactor to make code testable for the part which was updated.

### Checklist


timtebeek (approved):
Thanks for the changes here @amishra-u ; good to improve compatibility indeed.

PR #4751, opened 2024-12-06
Support variable expressions in parenthesis

jevanlingen:
## What's changed?
Nested parentheses around variable expressions are supported as well for the groovy parser.
Also added the tests for method invocation with nested parentheses (but disabled). 

## What's your motivation?

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -36,7 +36,9 @@ import org.openrewrite.internal.EncodingDetectingInputStream;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.internal.JavaTypeCache;
 import org.openrewrite.java.marker.ImplicitReturn;
+import org.openrewrite.java.marker.OmitParentheses;
 import org.openrewrite.java.marker.Semicolon;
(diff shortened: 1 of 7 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2167,19 +2110,21 @@ public TypeTree visitVariableExpressionType(VariableExpression expression) {
...
-                type = typeMapping.type(staticType((org.codehaus.groovy.ast.expr.Expression) expression));
-            }
-            queue.add(new J.Identifier(randomId(),
-                    sourceBefore(expression.getName()),
-                    Markers.EMPTY,
-                    emptyList(),
-                    expression.getName(),
-                    type, null)
-            );
+            queue.add(insideParentheses(expression, fmt -> {
```
jevanlingen, line 2113: This was just a bug in the parser, the visitVariableExpression should also use the `visitVariableExpression` so things like:

```groovy
def timestamp(int hours, int minutes, int seconds) {
  (hours) * 60 * 60 + (minutes * 60) + seconds
}
```

starts working!

knutwannheden (approved):

timtebeek (approved):
Thanks a lot! Great to have this fix pulled in already, and the added tests to know what still needs work.

Issue #4748, opened 2024-12-05
Groovy parser does not support multiple assignments


jevanlingen:
#### A
```groovy
@Test
void multipleAssignments() {
    rewriteRun(
        groovy(
            """
            def (a, b) = ['1', '2']
            """
        )
     );
 }
```

#### B
```groovy
@Test
void multipleAssignmentsAtOneLine() {
    rewriteRun(
        groovy(
            """
             def a = '1', b = '2'
            """
        )
     );
 }
```

## What did you expect to see?
Working test

#### A
```
Failed to parse file.groovy at cursor position 0. The next 10 characters in the original source are `def (a, b)`
```

#### B
```
Expected :"def a = '1', b = '2'"
Actual   :"def a = '1', bb = '2'"
```

knutwannheden:
The B case seems like it is just a bug, that would probably be easy to fix. For the A case a new LST model element will be required.

jevanlingen:
I guess you are right, it's maybe better to split theses cases in two different issues now I think about it.

PR #4757, opened 2024-12-09
Fix groovy multivariable declaration

Laurens-W:
## What's changed?

When parsing multivariable declarations now takes into account that no traditional identifier is available for any variable after the first

## What's your motivation?

- Fix https://github.com/openrewrite/rewrite/issues/4748

@timtebeek @jevanlingen @sambsnyd @knutwannheden 

## Any additional context

The Groovy AST returns multivariable declarations as separate `DeclarationExpression`s within a block. Therefore we cannot easily map them to `J.VariableDeclarations`

### Checklist


timtebeek:
Saving others a click with the failures seen: https://ge.openrewrite.org/s/nr7naekjegaas/tests/overview?outcome=FAILED

[wildcardWithUpperBound()](https://ge.openrewrite.org/s/nr7naekjegaas/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.VariableDeclarationsTest/wildcardWithUpperBound())
[diamondOperator()](https://ge.openrewrite.org/s/nr7naekjegaas/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.VariableDeclarationsTest/diamondOperator())
[singleVariableDeclarationStaticallyTyped()](https://ge.openrewrite.org/s/nr7naekjegaas/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.VariableDeclarationsTest/singleVariableDeclarationStaticallyTyped())
[wildcardWithLowerBound()](https://ge.openrewrite.org/s/nr7naekjegaas/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.VariableDeclarationsTest/wildcardWithLowerBound())
[genericMaps()](https://ge.openrewrite.org/s/nr7naekjegaas/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.MapEntryTest/genericMaps())

First one fails with
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "file.groovy":	
```
```diff
diff --git a/file.groovy b/file.groovy	
index 3861186..ac7fbf9 100644	
--- a/file.groovy	
+++ b/file.groovy	
@@ -1 +1 @@ 	
-List<? extends String> l	
\ No newline at end of file	
+List<?< extends String> l	
\ No newline at end of file	
```
```
expected: "List<? extends String> l"	
 but was: "List<?< extends String> l"	
```

Laurens-W:
I simplified a little too far, reverted that partially and now they work (locally)

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2077,29 +2081,23 @@ public void visitPrefixExpression(PrefixExpression unary) {
...
 
-            if (expression.isDynamicTyped()) {
-                Space prefix = whitespace();
-                String keyword;
-                if (source.substring(cursor).startsWith("final")) {
-                    keyword = "final";
-                } else {
-                    keyword = source.substring(cursor, cursor + 3);
+            if (expression.isDynamicTyped() || source.charAt(cursor) == ',') {
+                while (!Character.isWhitespace(source.charAt(cursor))) {
```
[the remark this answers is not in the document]
knutwannheden: `def final` works too 😄

timtebeek:
One remaining failure it seems in [org.openrewrite.groovy.tree.AssignmentTest/simpleWithFinal()](https://ge.openrewrite.org/s/i2cuzsafsu3hm/tests/task/:rewrite-groovy:test/details/org.openrewrite.groovy.tree.AssignmentTest/simpleWithFinal()?top-execution=1)
```
final~~(non-whitespace)~~> def <~~x = "s"	
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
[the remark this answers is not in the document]
Laurens-W: Both work now as captured in a unit test :)

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2176,6 +2171,33 @@ private <T> T pollQueue() {
...
+            keyword.append("var");
+            cursor += 3;
+        } else if (leftover.startsWith("def")) {
+            keyword.append("def");
+            cursor += 3;
+        } else if (leftover.startsWith(",")) {
+            keyword.append(",");
+            cursor++;
+        } else {
+            throw new IllegalStateException("Ran into unknown or unimplemented identifier");
```
timtebeek, line 2189: I'm not a fan of throwing exception, but if we do, it might be nice to include `leftover` here to reduce the guess work.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2176,6 +2171,33 @@ private <T> T pollQueue() {
         }
     }
 
+    private void getKeyword(String variableName, StringBuilder keyword) {
```
jevanlingen, line 2174: I don't think I should use a StringBuilder here, but just return a String. The case someone uses multiple keywords is rather small, so for like > 95% of all scripts this function will only be run once per variable. So the performance gain is really small.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
Laurens-W: https://github.com/openrewrite/rewrite/blob/bbdd436fd665a8c20a3878297de313ce9c17ffb6/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L219-L222
Eventually the exception is caught and the 10 characters from where the cursor is are included in the checked exception.
Perhaps we should capture this in a test?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
timtebeek: That part seemed odd to me as well; clearer to just return a String indeed.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
timtebeek: I've seen that cursor be off quite a bit when there are parser issues; hence why I would include it in the thrown exception here when readily available. I don't think it's easy to trigger with a test, so let's not spend time on that.

Laurens-W: I did spend some time on it and it indeed isn't easy. Perhaps we could find a way to pass custom assertions with a `SourceSpec` but for now that's way out of scope and I'll just include a part of the remaining string as an indicator of where the exception occured!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2077,29 +2081,20 @@ public void visitPrefixExpression(PrefixExpression unary) {
 
         public TypeTree visitVariableExpressionType(VariableExpression expression) {
             JavaType type = typeMapping.type(staticType(((org.codehaus.groovy.ast.expr.Expression) expression)));
+            Space prefix = whitespace();
+            StringBuilder keyword = new StringBuilder();
```
jevanlingen, line 2085: This can now be a String as well.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2176,6 +2171,35 @@ private <T> T pollQueue() {
...
+            throw new IllegalStateException("Ran into unknown or unimplemented identifier at " + leftover.substring(0, 10));
+        }
+        int saveCursor = cursor;
+        Space space = whitespace();
+        if (source.substring(cursor).startsWith(variableName)) {
+            cursor = saveCursor;
+        } else {
+            retVal += space.getWhitespace();
+            retVal += getKeyword(variableName);
+        }
```
jevanlingen, line 2199: Can be simplified to:

```suggestion
        if (!source.substring(indexOfNextNonWhitespace(cursor, source)).startsWith(variableName)) {
            retVal += whitespace().getWhitespace();
            retVal += getKeyword(variableName);
        }
```

jevanlingen (approved):

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2176,6 +2171,31 @@ private <T> T pollQueue() {
         }
     }
 
+    private String getKeyword(String variableName) {
+        String retVal;
+        String leftover = source.substring(cursor);
```
knutwannheden, line 2176: We should avoid calling `String#substring()` as that can be very expensive operation for big source files when the parser does this over and over again. One alternative is to use `String#regionMatches()`.

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/AssignmentTest.java
```diff
@@ -16,14 +16,58 @@
...
+
+    @Test
+    void simpleWithFinal() {
+        rewriteRun(
+          groovy(
+            """
+              final def x = "x"
+              def final y = "y"
+              final z = "z"
+              """
```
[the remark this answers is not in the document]
knutwannheden: While this isn't perfect either, I think I would opt for mapping all of `final`, `def`, and `var` to modifiers (`J.Modifier`) and instead deprecating the `RedundantDef` marker, while still supporting it in the printer to remain backwards compatible.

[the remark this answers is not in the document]
Laurens-W: Implemented this. I'm not sure if we can actually mark the `RedundantDef` marker as deprecated though as that may also apply to declared methods, so I left that in.

As we discussed we may want to find a way to combine consecutive `DeclarationExpression`s into one `J.VariableDeclarations`. That could be a further improvement and I'm willing to work on that, but want to return focus to the parenthesis issue as well :)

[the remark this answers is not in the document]
knutwannheden: @sambsnyd Since you recently added that `RedundantDef`, do you also want to chime in here and tell us what your thoughts are about this?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1466,19 +1469,60 @@ public void visitDeclarationExpression(DeclarationExpression expression) {
...
+            cursor = saveCursor;
+            return Optional.empty();
+        }
+
+        private List<J.Modifier> getModifiers() {
+            List<J.Modifier> modifiers = new ArrayList<>();
+            int saveCursor = cursor;
+            Space prefix = whitespace();
+            while (source.startsWith("def", cursor) || source.startsWith("var", cursor) || source.startsWith("final", cursor)) {
+                if (source.regionMatches(cursor, "var", 0, 3)) {
```
knutwannheden, line 1507: Should also work here, I think:
```suggestion
                if (source.startsWith("var", cursor)) {
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependencyVisitor.java
```diff
@@ -290,22 +290,10 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                            addDependencyInvocation = addDependencyInvocation.withPrefix(currentStatement.getPrefix());
 
                             if (addDependencyInvocation.getSimpleName().equals(beforeDependency.getSimpleName())) {
-                                if (currentStatement instanceof J.VariableDeclarations) {
-                                    J.VariableDeclarations variableDeclarations = (J.VariableDeclarations) currentStatement;
-                                    if (variableDeclarations.getTypeExpression() != null && !variableDeclarations.getTypeExpression().getPrefix().equals(originalPrefix)) {
-                                        statements.set(i, variableDeclarations.withTypeExpression(variableDeclarations.getTypeExpression().withPrefix(originalPrefix)));
-                                    }
-                                } else if (!currentStatement.getPrefix().equals(originalPrefix)) {
+                                if (!currentStatement.getPrefix().equals(originalPrefix)) {
```
knutwannheden, line 296: It is very unclear to me what these changes here have to do with the parsing issue. Is a comment required here?

Laurens-W: Due to the way `DeclarationExpression`s were parsed into `J.VariableDeclarations` we needed a specific check here. After my changes the parsing is more in-line with other types and so we no longer need that exceptional case here!

knutwannheden (approved):
I think we should be able to proceed with this. Is it difficult to solve this without the `MultiVariable` marker?

Laurens-W:
> I think we should be able to proceed with this. Is it difficult to solve this without the `MultiVariable` marker?

I haven't looked into it too much yet but I'd imagine somewhere after this block: https://github.com/openrewrite/rewrite/blob/d4494909e99fbcad0e884a0f5b418bfac532d74f/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L203-L214 we have the required context to sweep consecutive `VariableDeclarations` together

PR #4758, opened 2024-12-09
Performance improvements of `Find` recipe

nielsdebruin:
## What's changed?

The `Find` has been modified to increase its performance. Nearly all these optimizations are related to how Strings are handled. 

#### Main changes
- Replace basic String concatenation with a `StringBuilder` where we set the capacity to the exact length of the final String. Apart from setting the capacity exactly, this will mostly only affect OpenRewrite users using old JVM, since most modern JVM will often already generate optimized bytecode using a StringBuilder.
- **Significantly** reduce the number of generated substrings, e.g., while looking up line ending on substrings.
- Avoid recomputing common values, byte precomputing them outside the loop.

## What's your motivation?

Make it faster.

@kmccarp @timtebeek 

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/text/FindTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/text/FindTest.java
@@ -144,6 +144,7 @@ class FindTest implements RewriteTest {
           )
         );
     }
+
     @Test
     void regexMatchingWhitespaceWithoutMultilineWithDotall() {
         rewriteRun(
@@ -153,11 +154,6 @@ class FindTest implements RewriteTest {
             """
               Zero
               One Two
-              Three
-              """,
-            """
-              Zero
-              ~~>One Two
               Three
               """
           )
--- a/rewrite-core/src/test/java/org/openrewrite/text/FindTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/text/FindTest.java
@@ -111,4 +111,106 @@ class FindTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void regexBasicMultiLine() {
+        rewriteRun(
+          spec -> spec.recipe(new Find("[T\\s]", true, true, true, null, null)),
+          text(
+            """
+              This is\ttext.
+              This is\ttext.
+              """,
+            """
+              ~~>This~~> is~~>\ttext.~~>
+              ~~>This~~> is~~>\ttext.
+              """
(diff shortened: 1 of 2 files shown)


kmccarp:
@lkerford I guess @nielsdebruin beat you to it :) please take a look

lkerford (approved):
I wanted to see how much of a difference this change would make with the performance, and I wasn't disappointed. The test I have shared below ran in 7753ms on the main branch vs 450ms on this branch. This is fantastic  

```
@Test
    void findPerformanceTest() {
        String intputLine = """
          This is a line above.
          This is text.
          This is a line below.
          """;
        String outputLine = """
          This is a line above.
          This is ~~>text.
          This is a line below.
          """;

        int i=0;
        String fileInput= intputLine;
        String fileOutput= outputLine;
        int numberOfEntries = 100000;
        while (i < numberOfEntries){
            fileInput += intputLine;
            fileOutput += outputLine;
            i++;
        }
        long startTime = System.currentTimeMillis();

        rewriteRun(
          spec -> spec.recipe(new Find("text", null, null, null, null, null))
            .dataTable(TextMatches.Row.class, rows -> {
                assertThat(rows).hasSize(numberOfEntries+1);
                assertThat(rows.get(0).getMatch()).isEqualTo("This is ~~>text.");
            }),
          text(
            fileInput,
            fileOutput
          )
        );
        long elapsedTime = System.currentTimeMillis() - startTime;
        System.out.println(elapsedTime + "ms");
    }
```

rewrite-core/src/test/java/org/openrewrite/text/FindTest.java
```diff
@@ -111,4 +111,106 @@ void caseInsensitive() {
...
+            """
+              ~~>This is text.
+              This is a line below.
+              This is a line above.
+              This is text.
+              This is a line below.
+              """
+          )
+        );
+    }
```
timtebeek, line 215: I find it quite hard to gauge from regex and output what we intent to match here and why; is it the first below? the second below? The first This? any of the others?

Can we use a clearer example with unique values across the lines, maybe something like
```
Zero
One
Two
Three
```
And then match on either `^One.Two$`, or `^One\\sTwo$` with variations of dot all and multilne enabled/disabled. I think that'll be clearer what exactly is matched as we only show the marker start.

timtebeek (approved):

PR #4769, opened 2024-12-11
Add lombok support for java-11

amishra-u:
## What's changed?
Added lombok support for java-11
## What's your motivation?
Enable type attribution for lombok generated methods for java 11.

### Checklist
        Lombok test already exist for java 17, enabled those test for java 11


timtebeek:
Thanks for the work done here to create parity between the Java 11 and 17 parsers @amishra-u ! As you've found there's still some issues we need to work out before we're comfortable repeating these patterns across parsers, and enable it by default:
- https://github.com/openrewrite/rewrite/issues/4781

Let's keep an eye on the discussion there and if we find a fix, apply that both for Java 17 and Java 11.

timtebeek (approved):
Thanks a lot for your help here & over on projectlombok/lombok#3800 @amishra-u!

amishra-u:
I was reading a reddit thread, found claim of lombok creator that lombok does't modify existing ones. 
https://www.reddit.com/r/java/comments/170gdha/comment/k3lq6p4/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

In this case, this issue is a design bug for Builder.Default annotation.

PR #4772, opened 2024-12-11
Groovy parser does not support multiple closure arguments without parentheses

jevanlingen:
## What's changed?
Support for multiple closure arguments without parentheses.

## What's your motivation?
If you have a piece of Gradle task like:

```groovy
copySpec {
    from { 'src/main/webapp' } { exclude "**/*.jpg" }
    rename '(.+)-staging(.+)', '$1$2'
}
```
 
it won't work because the Groovy parser could not parse multiple closure arguments without parentheses.

You can workaround the problem, by adding parentheses:
```groovy
copySpec {
    from({ 'src/main/webapp' }, { exclude "**/*.jpg" })
    rename '(.+)-staging(.+)', '$1$2'
}
```

### Checklist


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodInvocationTest.java
```diff
@@ -35,6 +35,11 @@ void gradle() {
               repositories {
                   mavenCentral()
               }
+              
+              copySpec {
+                  from { 'src/main/webapp' } { exclude "**/*.jpg" }
+                  rename '(.+)-staging(.+)', '$1$2'
+              }
```
timtebeek, line 42: The first test in our test classes usually serves as a minimal happy path test; to be copied for subsequent tests to test specific elements. As such I think it's not desirable to add this block to the topmost test, but rather add a second test, appropriately annotated with `@Issue` that merely adds this block and ensures we can still parse the block.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -883,6 +884,22 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
...
+                    if (foundNonClosure) {
+                        return false;
+                    }
+                } else {
+                    foundNonClosure = true;
+                }
+            }
+
+            return list.get(list.size() - 1) instanceof ClosureExpression;
+        }
```
timtebeek, line 901: Trying to mentally compute this method hurt my brain a little; perhaps I'm just switching too much. To me it'd make sense to for instance to the last check up front, to return early for methods that to not have a last Closure argument. Phrased like that I wonder if there's then even a need to loop over the other arguments: any method that ends with one closure would fit the `lastArgumentsAreAllClosures` above? Or is there more subtlety to it?

jevanlingen: > Or is there more subtlety to it?

Yeah, I fear there is. What I want to test here, is whether the list last item(s) contains a ClosureExpression and no 
ClosureExpression is used before. So for example:

```
1. [ClosureExpression, ClosureExpression] => true
2. [VariableExpression, ClosureExpression] => true
3. [VariableExpression, ClosureExpression, ClosureExpression] => true

4. [VariableExpression, VariableExpression] => false
5. [ClosureExpression, ClosureExpression, VariableExpression, ClosureExpression] => false
6. [ClosureExpression, VariableExpression, ClosureExpression, VariableExpression] => false
```

> Any method that ends with one closure would fit the lastArgumentsAreAllClosures above?

Nope, as you can see, that would _wrongly_ return `true` for case 5. So what I do here:
- Check reversely check if a ClosureExpression precedes a non-ClosureExpression. If so, it's either case 5 or 6, so return `false`.
- If we have gone through the whole list, it means there is no ClosureExpression preceding a non-ClosureExpression. But it could also be case 4. Thus check if the last is a ClosureExpression, which means it has to be case 1, 2 or 3.

--

You are right though, this reads maybe sub-optimal. Because you are right, the last item always has to be a ClosureExpression. So maybe this is better:

```java
public boolean endsWithClosures(List<org.codehaus.groovy.ast.expr.Expression> list) {
  if (!(list.get(list.size() - 1) instanceof ClosureExpression)) {
    return false;
  }
 
  boolean foundNonClosure = false;
  for (int i = list.size() - 1; i >= 0; i--) {
    if (list.get(i) instanceof ClosureExpression) {
      if (foundNonClosure) {
        return false;
      }
    } else {
      foundNonClosure = true;
    }
  }

  return false
}
```

jevanlingen: I pushed ⬆️. If you know a better way, please tell 😇.

timtebeek: Thanks for the explanation & adjustment; that makes more sense then indeed; I guess I was thrown off a bit by the naming of the method and variable, but understand them now that it's been explained to mean "all closures are at the end".

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -867,7 +868,7 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
                         if (omitParentheses == null) {
                             after = sourceBefore(")");
                         }
-                    } else {
+                    } else if (!(arg instanceof J.Lambda) || !(lastArgumentsAreAllClosures && omitParentheses != null)) {
```
timtebeek, line 871: The negations here were slightly confusing to me as well; All tests pass if I apply this refactor; does that still fit what you were after?
image
```suggestion
                    } else if (!(arg instanceof J.Lambda && lastArgumentsAreAllClosures && omitParentheses != null)) {
```

timtebeek (approved):
Great to see this fixed, thanks!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: It's logically the same, so it makes sense the test do succeed. In English this is what I want to convey:
>   `argument ` is a Closure where I know the method has parentheses and the method has trailing lambda's (that includes lambda's only), skip next part.
  
 Seems to me that's exactly what the refactor does also say, so let's go for it!

timtebeek (approved):

Issue #4687, opened 2024-11-19
openrewrite maven recipes shoudn't try to resolve every single pom


adastraperangusta:
Hello,

I have projects with external dependencies that have been installed manually into a local maven repository.
These artefacts don't have a pom in the repository.
On these projects 'mvn clean install' succeeds and there is only a warning about the missing pom :
```log
[WARNING] The POM for xxxxxxx is missing, no dependency information available
```

However when I try to run any openrewrite maven recipes on this kind of project, it fails with :
```log
[ERROR] </project>org.openrewrite.maven.MavenDownloadingException: Unable to download POM:xxxxx Tried repositories:
[ERROR] https://repo.maven.apache.org/maven2: HTTP 404
[ERROR]   org.openrewrite.maven.internal.MavenPomDownloader.download(MavenPomDownloader.java:636)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom.resolveDependencies(ResolvedPom.java:929)
[ERROR]   org.openrewrite.maven.tree.ResolvedPom.resolveDependencies(ResolvedPom.java:851)
[ERROR]   org.openrewrite.maven.tree.MavenResolutionResult.resolveDependencies(MavenResolutionResult.java:175)
```

Why openrewrite maven recipes would fail when a maven build succeeds with warnings ?
How to fix this ?

You can reproduce this bug with this minimal project : [project.zip](https://github.com/user-attachments/files/17821254/project.zip)

 Just install any jar locally with : 
```shell
mvn install:install-file -Dfile=<any jar file> -DgroupId=bug.report -DartifactId=somedep -Dversion=1.0 -Dpackaging=jar -DgeneratePom=false
```

Then 'mvn clean install' will give : 
```log
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------------< bug.report:somejar >-------------------------
[INFO] Building somejar 1.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for bug.report:somedep:jar:1.0 is missing, no dependency information available
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ somejar ---
[INFO] Deleting /home/stef/repos_github/bugreport/target
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ somejar ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[...]

PR #4738, opened 2024-12-03
Allow pom download failures

adastraperangusta:
## What's changed?
This PR adds a `org.openrewrite.allowPomDownloadFailure` property , inspired by the existing `org.openrewrite.test.readMavenSettingsFromDisk` property. When set to true, the MavenPomDowloader class doesn't throw an exception when it doesn't find a suitable pom to download for a dependency, and instead returns a minimal object.

## What's your motivation?
Some maven projects are correct (mave wise) but don't have all the pom's for their dependencies (see https://github.com/openrewrite/rewrite/issues/4687), which prevents maven recipes to work on these kind of project.
- Fix #4687

## Have you considered any alternatives or workarounds?
This is a workaround, I have the feeling that the real solution would be to use maven libraries to build the maven model instead of re-implementing it. But it would be quite an heavy change.

### Checklist
- [X] I've added unit tests to cover both positive and negative cases
- [X] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

PR #4777, opened 2024-12-12
Tolerate missing pom when jar is present

timtebeek:
## What's changed?
Catch a failure to download a pom in ResolvedPom, and if caught check to see if the corresponding jar can be found. If so, continue without issue.

## What's your motivation?
- Fix https://github.com/openrewrite/rewrite/issues/4687

- The `MavenArtifactDownloader` had previously been unused for quite a while; figured to bring that back here to check for jar availability. We could possibly move that to the same package as MavenPomDownloader, such that we don't need:
- the visibility of `MavenPomDownloader.distinctNormalizedRepositories` changed to `public`, as we need to look across all repositories for the `.jar` file. Not just the ones defined in the `ResolvedPom`.

There is an alternative implementation in
- https://github.com/openrewrite/rewrite/pull/4738

That chose the route of having changes mostly in MavenPomDownloader, which then also looks at jars and if so returns a dummy pom.xml;  I think it's better not to mix responsibilities there, nor to continue processing a known dummy pom.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -681,7 +681,7 @@ public class MavenPomDownloader {
         return gav.getVersion();
     }
 
-    Collection<MavenRepository> distinctNormalizedRepositories(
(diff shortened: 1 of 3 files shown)


timtebeek:
- Continued in https://github.com/openrewrite/rewrite/pull/4738

PR #4778, opened 2024-12-12
Drop unused GraphvizResolutionEventListener & dependency

timtebeek:
## What's your motivation?
This dragged in various dependencies, including slf4j-api, which then lead to warnings in the tests that no implementation was provided.
image

It looks like these scripts were developed years ago and not used or maintained since; we have other ways to visualize graphs in the platform now using Jupyter notebooks.


timtebeek:
Turns out something was using the transitive SLF4J; whereas we likely want to rip that out as we don't do logging much.

PR #4780, opened 2024-12-12
Normalize server URLs to avoid missing scheme

pstreef:
This changes the output slightly (no more user in the URL) but that is fine as the user is not required/used in matching anyway.

Also fixed the comment and exception text.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
+++ b/rewrite-core/src/main/java/org/openrewrite/GitRemote.java
@@ -118,11 +118,13 @@ public class GitRemote {
 
         /**
          * Build a {@link URI} clone url from components, if that protocol is supported (configured) by the matched server
+         *
          * @param service  the type of SCM service
          * @param origin   the origin of the SCM service, any protocol will be stripped (and not used for matching)
          * @param path     the path to the repository
          * @param protocol the protocol to use. Supported protocols: ssh, http, https
-         * @return
+         * @return the clone URL if it could be created.
+         * @throws IllegalArgumentException if the protocol is not supported by the server.
          */
         public URI buildUri(Service service, String origin, String path, String protocol) {
             if (!ALLOWED_PROTOCOLS.contains(protocol)) {
@@ -147,7 +149,7 @@ public class GitRemote {
                         .orElseGet(() -> {
                             URI normalizedUri = Parser.normalize(origin);
                             if (!normalizedUri.getScheme().equals(protocol)) {
-                                throw new IllegalStateException("No matching server found that supports ssh for origin: " + origin);
+                                throw new IllegalStateException("No matching server found that supports " + protocol + " for origin: " + origin);
                             }
                             return normalizedUri;
                         });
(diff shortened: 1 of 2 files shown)


rewrite-core/src/test/java/org/openrewrite/GitRemoteTest.java
```diff
@@ -180,25 +180,25 @@ void normalizePortWithoutSchemaNotSupported() {
...
       Bitbucket, scm.company.com/context/bitbucket, org/repo, http, http://scm.company.com/context/bitbucket/scm/org/repo.git
-      Bitbucket, scm.company.com/context/bitbucket, org/repo, ssh, ssh://git@scm.company.com:7999/context/bitbucket/org/repo.git
+      Bitbucket, scm.company.com/context/bitbucket, org/repo, ssh, ssh://scm.company.com:7999/context/bitbucket/org/repo.git
       GitHub, scm.company.com/context/github, org/repo, https, https://scm.company.com/context/github/org/repo.git
-      GitHub, scm.company.com/context/github, org/repo, ssh, ssh://git@scm.company.com/context/github/org/repo.git
+      GitHub, scm.company.com/context/github, org/repo, ssh, ssh://scm.company.com/context/github/org/repo.git
       GitLab, scm.company.com/context/gitlab, group/subgroup/repo, https, https://scm.company.com/context/gitlab/group/subgroup/repo.git
       GitLab, scm.company.com:8443/context/gitlab, group/subgroup/repo, https, https://scm.company.com:8443/context/gitlab/group/subgroup/repo.git
-      GitLab, scm.company.com/context/gitlab, group/subgroup/repo, ssh, ssh://git@scm.company.com:8022/context/gitlab/group/subgroup/repo.git
```
bryceatmoderne, line 201: Can you confirm this will not cause problem why trying to close using ssh?

pstreef: Hmm, thinking some more on it maybe this can cause problems if a local env requires it to match the right user/pw

bryceatmoderne (approved):

rewrite-core/src/main/java/org/openrewrite/GitRemote.java
```diff
@@ -142,12 +143,12 @@ public URI buildUri(Service service, String origin, String path, String protocol
                                 .anyMatch(o -> o.equalsIgnoreCase(stripProtocol(origin)))
                         )
                         .flatMap(server -> server.getUris().stream())
-                        .filter(uri -> uri.getScheme().equals(protocol))
+                        .filter(uri -> Parser.normalize(uri).getScheme().equals(protocol))
                         .findFirst()
                         .orElseGet(() -> {
```
bryceatmoderne, line 148: You could add a map to normalize the uri to remove duplicate calls to do so

pstreef: There is no duplicate call. I do not want the normalized URI as output, just in the filter. And if none are found then we normalize something else

bryceatmoderne: Ah, sorry, I didn't read the `orElseGet`

PR #4785, opened 2024-12-15
Add explicit imports to avoid conflicts with classes added to java.lang, like record

BramliAK:
## What's changed?
Add explicit imports to avoid conflicts with classes added to java.lang, like record

- For https://github.com/openrewrite/rewrite-migrate-java/issues/540

### Checklist


timtebeek (approved):
Thanks for the helpful change here @BramliAK ! Indeed seems like the right place to make this change, and we can now follow up with a recipe in rewrite-migrate-java that adds explicit imports as needed for fix https://github.com/openrewrite/rewrite-migrate-java/issues/540.

PR #4786, opened 2024-12-15
Add `testRuntimeOnly` dependency on `junit-platform-launcher` with Gradle 8.+

timtebeek:
## What's changed?
- Add recipe to add Gradle dependency on `org.junit.platform:junit-platform-launcher`
- Add transitive dependency when using `rewrite-test`

## What's your motivation?
Reduce the warnings seen in build logs when using Gradle 8.x, ahead of a migration to Gradle 9.
https://docs.gradle.org/8.11.1/userguide/upgrading_version_8.html#test_framework_implementation_dependencies

Would the `api` dependency already prevent that we have to add the testRuntimeOnly dependency for `rewrite-*`?

We could delay this until Gradle 9, but the warnings are already annoying right now.

### DIFF
--- a/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-8.yml
+++ b/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-8.yml
@@ -24,6 +24,9 @@ recipeList:
       version: 8.x
       addIfMissing: false
 
+  # https://docs.gradle.org/8.11.1/userguide/upgrading_version_8.html#test_framework_implementation_dependencies
+  - org.openrewrite.gradle.AddJUnitPlatformLauncher
+
   # https://github.com/gradle/gradle/blob/v7.6.4/subprojects/core/src/main/java/org/gradle/api/internal/FeaturePreviews.java
   # https://github.com/gradle/gradle/blob/v8.10.1/subprojects/core/src/main/java/org/gradle/api/internal/FeaturePreviews.java
   - org.openrewrite.gradle.RemoveEnableFeaturePreview:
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-8.yml
```diff
@@ -24,6 +24,9 @@ recipeList:
       version: 8.x
       addIfMissing: false
 
+  # https://docs.gradle.org/8.11.1/userguide/upgrading_version_8.html#test_framework_implementation_dependencies
+  - org.openrewrite.gradle.AddJUnitPlatformLauncher
```
shanman190, line 28: optional: Is there a benefit in the indirection here rather than the AddDependency being here directly?

shanman190 (approved):

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-8.yml
timtebeek: Only in the sense that it becomes a searchablle migration step, that folks can run separately as they are not likely to re-run `MigrateToGradle8` when already on Gradle 8. It is indeed optional.

shanman190:
> Would the api dependency already prevent that we have to add the testRuntimeOnly dependency for rewrite-*?

Yeah, based on my understanding once `rewrite-test` is exposing the dependency in the `api` configuration, then all second-level modules will get the benefit of it being available. The one part about this that we will want to keep an eye on is if there version conflicts introduced due to the `api` dependency with the imported `junit-platform-launcher` in transitive which are depending upon how Gradle interacts with it at runtime. 

On the flip side, the recommendation of placing the dependency into `testRuntimeOnly` means that it would be available for testing tasks, but would not be exported for others to use and consume.

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-8.yml
shanman190: Fair enough. One thing that I've wondered is how to approach these sort of utility recipes as searching for them in the recipe catalog can become pretty overwhelming, but that's an entirely separate problem.

timtebeek:
You're very right to point out that the `api("org.junit.platform:junit-platform-launcher")` now comes in with the wrong scope, making it available to sources as opposed to runtime only. But without this change we'd have to update ~50+ build.gradle.kts files which for now only depend on rewrite-test, and expect the other transitive junit api dependencies already. I'd wanted to avoid those updates, especially since .kts isn't supported yet. With that I think we can tolerate the potential downsides, at least for now.

PR #4787, opened 2024-12-15
Always add import for `Record` classes not from `java.lang` package, to avoid conflicts on Java 14+

BramliAK:
## What's changed?
Force Importing No Java Record 

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java
@@ -210,6 +210,36 @@ class AddImportTest implements RewriteTest {
           )
         );
     }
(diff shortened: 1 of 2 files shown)


timtebeek:
hi @BramliAK ! Thanks for the change; would you mind describing what additional cases this covers beyond what's seen in
- https://github.com/openrewrite/rewrite-migrate-java/pull/630 ?

I do know we had that disabled test there; wonder if this aims to fix that one as well.

timtebeek:
Ah I only just now saw your comment here; thanks!
- https://github.com/openrewrite/rewrite-migrate-java/pull/630#issuecomment-2543944899

timtebeek (approved):
Well spotted @BramliAK ; thanks a lot!

PR #4790, opened 2024-12-16
`RemoteArchive` `SearchResult` is not printed

pstreef:
This adds a test demonstrating that RemoteArchive printing does not add the search marker.


sambsnyd:
A `RemoteArchive` representing a binary file would have no sensible way to print a marker.
A `RemoteArchive` represents a test file I agree it makes sense to print the marker same as you would a PlainText.
It is straightforward assert the presence of the marker without printing.
So I don't see a lot of value in adding a printer for this, but I wont object so long as that printer is part of the test harness.

pstreef:
After some discussion we decided to leave this as is for now.

PR #4791, opened 2024-12-16
Allow file scheme in `RemoteArchive` to simplify testing

pstreef:
While it might look a bit controversial, the file scheme can also point to a remote (for instance a mounted network share) file. By allowing the `file://` scheme we can use `RemoteArchive` for those files.

As a useful side effect, this makes testing RemoteArchive handling a lot easier.

## What's changed?
`RemoteArchive` supports `file://` protocol

## What's your motivation?
Make it easier to test without having to manipulate the context and mocking

Do we want to allow `file` protocol for `RemoteArchive`

@sambsnyd 

Manipulate the context and mock the webserver

## Any additional context
less mocking, more better!

### Checklist


rewrite-core/src/test/java/org/openrewrite/remote/RemoteArchiveTest.java
```diff
@@ -58,10 +56,9 @@ void gradleWrapper(String version) throws Exception {
 
     @Test
     void gradleWrapperDownloadFails() throws Exception {
-        URL distributionUrl = requireNonNull(RemoteArchiveTest.class.getClassLoader().getResource("gradle-7.4.2-bin.zip"));
+        URL distributionUrl = new URL("http://example.com");
```
sambsnyd, line 59: That is a real website, however sparse, no need to have our tests ping someone's page.
```suggestion
        URL distributionUrl = new URL("http://example");
```

sambsnyd (approved):

PR #4793, opened 2024-12-17
Add Docker Image reference, to be used for gitlab, docker, etc

jevanlingen:
## What's changed?
Added a new `IMAGE` reference type, which points to a docker image. Could be used for gitlab-ci files, docker files, testcontainers etc.

Also did a little refactoring in the SourceFileWithReferences:
- Made the class abstract, so the `getReferences` cache logic can be once instead of every implementation.
- The `findMatches` method did also check for kind; as the `Reference.Matcher` does already do that, no need to do it here too.

## What's your motivation?
Run one find (and replace?) recipe to search for your docker images in one go, no matter how the image has been configured.

Be extra observant, as this is the first time I touch the Reference code. I put most of the code in the `rewrite-java` package, so maybe that could be improved as well.

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/ImageMatcher.java
```diff
@@ -0,0 +1,45 @@
...
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
timtebeek, line 30:
```suggestion
package org.openrewrite.java;
```

rewrite-core/src/main/java/org/openrewrite/text/DockerImageReference.java
```diff
@@ -0,0 +1,63 @@
...
+    Cursor cursor;
+    String value;
+
+    @Override
+    public Kind getKind() {
+        return Kind.IMAGE;
+    }
+
+    public static class Provider implements Reference.Provider {
+        private static final Pattern FROM = Pattern.compile("FROM\\s+([\\w:.-]*)", Pattern.CASE_INSENSITIVE);
```
knutwannheden, line 38: AFAIK the directives are case sensitive and I added some more allowed characters (possibly more exist).

```suggestion
        private static final Pattern FROM = Pattern.compile("\\bFROM\\s+([\\w/:.-]+)");
```

I also wonder what about the `--platform` option and other places images can occur as in the `COPY` directive's `--from` option.

rewrite-java/src/main/java/org/openrewrite/java/FindImage.java
```diff
@@ -0,0 +1,89 @@
...
+    private static class ReferenceFindSearchResultVisitor extends TreeVisitor<Tree, ExecutionContext> {
+        Map<Tree, List<Reference>> matches;
+
+        @Override
+        public Tree postVisit(Tree tree, ExecutionContext ctx) {
+            List<Reference> references = matches.get(tree);
+            if (references != null) {
+                if (tree instanceof PlainText) {
+                    String find = references.stream().map(Reference::getValue).sorted().collect(joining("|"));
+                    return new Find(find, true, null, null, null, null, true)
```
knutwannheden, line 80: I wonder if it is worth leveraging the Find recipe here.

rewrite-core/src/main/java/org/openrewrite/text/DockerImageReference.java
jevanlingen: > AFAIK the directives are case sensitive

No, there are not 😉. See also: https://docs.docker.com/reference/build-checks/from-as-casing/.

> I also wonder what about the --platform option and other places images can occur as in the COPY directive's --from option.

I will think if I can implement this as well. Or at least, I will give it a try! I was not even aware you could load images that way 😇.

rewrite-java/src/main/java/org/openrewrite/java/FindImage.java
jevanlingen: I think it does, because it's way nicer to point directly at the right place. Using _snippets_ here directly would not be a gain, because you would end up with kinda the same code as the Find recipe (but more tailored to docker-images stuff of course). 

In the end (but for a way later stage), I think it would be better to create a dockerfile-LST, so this kind of code is not needed anymore.

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringReference.java
```diff
@@ -26,13 +26,11 @@
...
 import org.openrewrite.xml.tree.Xml;
 
-import java.util.HashSet;
 import java.util.Optional;
-import java.util.Set;
 import java.util.regex.Pattern;
 
 @Value
-class SpringReference implements Reference {
+public class SpringReference implements Reference {
```
Laurens-W, line 33: As far as I'm aware only the Provider needs to be public

jevanlingen: I think you are right, but the inner `Provider` class will get a IntelliJ unused marker as soon as I remove the `public` keyword here. I think it is clearer to keep it public for that reason.

rewrite-core/src/main/java/org/openrewrite/text/DockerImageReference.java
jevanlingen: Updated the code to support `--platform` and `--from` options as well. In the end I ditched the regex, doing it manually by going through the entire file gave me more flexibility 💪🏻.

jevanlingen:
Moved implementation files to
- https://github.com/openrewrite/rewrite-docker/pull/9

So in the core it's just the Image reference + all needed things to set it up. In `rewrite-docker` we have the actual recipe.

rewrite-core/src/main/java/org/openrewrite/SourceFileWithReferences.java
```diff
@@ -18,34 +18,48 @@
...
+import lombok.experimental.NonFinal;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.trait.Reference;
 
+import java.lang.ref.SoftReference;
 import java.util.*;
 
 @Incubating(since = "8.39.0")
-public interface SourceFileWithReferences extends SourceFile {
+public abstract class SourceFileWithReferences implements SourceFile {
```
timtebeek, line 30: I wonder if this type of change to tree package classes might trip up the deserialization we do for stored LSTs. 🤔

jevanlingen: O boy, I don't know about this. Could you figure this out or should we just revert the changes, for safety reasons?

jevanlingen: I will kinda revert it, but still do some enhancement. As I was thinking about a different way to write this, but keep the interface intact, I saw that the `SourceFile` in Reference does not make a lot of sense. So by dropping that field, things become cleaner as well.

rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
```diff
@@ -64,28 +62,8 @@ public Tree rename(Renamer renamer, Cursor cursor, ExecutionContext ctx) {
         return tree;
     }
 
-    private static class Matcher extends SimpleTraitMatcher<PropertiesReference> {
```
timtebeek, line 67: What's the reason to inline the matcher here only? And do you expect any impact from creating a new anonymous class with Predicate versus having one static class with static field?

rewrite-yaml/src/main/java/org/openrewrite/yaml/trait/YamlApplicationConfigReference.java
```diff
@@ -0,0 +1,67 @@
...
+        private static final Predicate<String> applicationPropertiesMatcher = Pattern.compile("^application(-\\w+)?\\.(yaml|yml)$").asPredicate();
+
+        @Override
+        public boolean isAcceptable(SourceFile sourceFile) {
+            return super.isAcceptable(sourceFile) && applicationPropertiesMatcher.test(sourceFile.getSourcePath().getFileName().toString());
+        }
+
+        @Override
+        public SimpleTraitMatcher<YamlReference> getMatcher() {
+            return new SimpleTraitMatcher<YamlReference>() {
```
timtebeek, line 47: For consistency this could also be a static inner class.

rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java
jevanlingen: The reason I did it has nothing to do with performance, but more with the intention of the code. The implementation of the matcher should have meaning outside of the Provider class. 

But you are right of course, doing it this way could potentially impact performance! The right way would have been to move static class inside the Provider class. I will improve it, thx!

jevanlingen:
> Couple questions; other than that this is looking good already, thanks!

Sweet, updated the code a little bit 🙂

timtebeek (approved):

PR #4798, opened 2024-12-18
fix[maven]: update plugin dependencies

jonblack:
## What's changed?

Adds a unit test for the ChangeDependencyGroupIdAndArtifactId recipe to update plugin dependencies. The fix is not provided.

- fixes openrewrite/rewrite/#4779


timtebeek (approved):
Thanks again for the test @jonblack ; that made this easy to verify & fix.

PR #4799, opened 2024-12-19
Add `MarkerMode` to select a `MarkerPrinter`

pstreef:
Instead of having to redefine `MarkerPrinter` everywhere we have a few standard `MarkerPrinter`s that can be used. There is a new diff method that uses the enum to select the right marker printer.

## What's changed?
Add `MarkerMode` and define `MarkerPrinter`s

## What's your motivation?
Reuse

the set of marker printers

@knutwannheden 

Not doing this

## Any additional context
No

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/PrintOutputCapture.java
+++ b/rewrite-core/src/main/java/org/openrewrite/PrintOutputCapture.java
@@ -75,11 +75,31 @@ public class PrintOutputCapture<P> implements Cloneable {
         @Incubating(since = "8.41.4")
         @RequiredArgsConstructor
         enum MarkerMode {
+            /**
+             * Does not print any markers.
+             */
             NONE(MarkerPrinter.NONE),
+            /**
+             * Prints a squiggly line arrow in front of the marked element wrapped in a comment.
+             * /*~~>* /Thing thing
+             */
             DEFAULT(MarkerPrinter.DEFAULT),
+            /**
+             * Prints a squiggly line arrow in front of the marked element wrapped in a comment.
+             * Possibly adding verbose information, depending on the marker
+             * /*~~>(this is some verbose information)* /Thing thing
+             */
             VERBOSE(MarkerPrinter.VERBOSE),
-            FENCED(MarkerPrinter.FENCED),
-            SEARCH_ONLY(MarkerPrinter.SEARCH_ONLY);
+            /**
+             * Prints a squiggly arrow in front and after the marked element. It only includes {@link SearchResult}
+             * /*~~>* /Thing thing /**~~>* /
+             */
+            SEARCH_ONLY(MarkerPrinter.SEARCH_ONLY),
+            /**
+             * Prints a fenced marker ID in front and after the marked element. It only includes {@link Markup} and {@link SearchResult}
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/PrintOutputCapture.java
```diff
@@ -67,6 +71,22 @@ public PrintOutputCapture<P> clone() {
...
 
     public interface MarkerPrinter {
+
+        @Incubating(since = "8.41.4")
+        @RequiredArgsConstructor
+        enum MarkerMode {
+            NONE(MarkerPrinter.NONE),
+            DEFAULT(MarkerPrinter.DEFAULT),
+            VERBOSE(MarkerPrinter.VERBOSE),
+            FENCED(MarkerPrinter.FENCED),
```
knutwannheden, line 81: I think it would be nice if we can come up with slightly more self-explanatory names. NONE is clear, the others less so. Can we use MARKUP_ONLY (even if that includes SearchResult)?

pstreef: it would indeed be nice. Any more suggestions? :D
We could also add (javadoc) comments to explain the differences

knutwannheden: For FENCED we can maybe use something like MARKUP_IDS or similar.

knutwannheden: Not sure about the actual names. I suppose DEFAULT and VERBOSE are OK too.

knutwannheden (approved):

rewrite-core/src/main/java/org/openrewrite/PrintOutputCapture.java
```diff
@@ -67,6 +71,42 @@ public PrintOutputCapture<P> clone() {
     }
 
     public interface MarkerPrinter {
+
+        @Incubating(since = "8.41.4")
+        @RequiredArgsConstructor
+        enum MarkerMode {
```
jkschneider, line 77: I don't love the name. I think there is a similar design concept in `java.nio.charset`. There is a `Charset` abstract class and a separate top level class called `StandardCharsets`. I think a top level `StandardMarkerPrinters` might be a slightly better design.

jkschneider (changes requested):

rewrite-core/src/main/java/org/openrewrite/Result.java
```diff
@@ -185,13 +185,23 @@ public String diff() {
...
     public String diff(@Nullable Path relativeTo) {
-        return diff(relativeTo, null);
+        return diff(relativeTo, (PrintOutputCapture.MarkerPrinter) null);
     }
 
     public String diff(@Nullable Path relativeTo, PrintOutputCapture.@Nullable MarkerPrinter markerPrinter) {
         return diff(relativeTo, markerPrinter, false);
     }
 
+    @Incubating(since = "8.41.4")
```
jkschneider, line 195: If we move this `StandardMarkerPrinters` then we don't need to add these overloads. Indeed I would prefer we don't add more overload variations here unless absolutely necessary. Every new overload is one more item in the IDE autocomplete list.

PR #4801, opened 2024-12-19
Make Groovy Parser correctly handle nested parenthesis

Laurens-W:
## What's changed?

Nested parentheses around method invocations are supported as well for the groovy parser.

## What's your motivation?

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2539,16 +2539,21 @@ public class GroovyParserVisitor {
             return (Integer) rawIpl;
         } else if (node instanceof MethodCallExpression) {
             MethodCallExpression expr = (MethodCallExpression) node;
-            int saveCursor = cursor;
-            whitespace();
-            int count = determineParenthesisLevel(expr.getObjectExpression().getLineNumber(), expr.getLineNumber(), expr.getObjectExpression().getColumnNumber(), expr.getColumnNumber());
-            cursor = saveCursor;
-            return count;
+            return determineParenthesisLevel(expr.getObjectExpression().getLineNumber(), expr.getLineNumber(), expr.getObjectExpression().getColumnNumber(), expr.getColumnNumber());
         }
         return null;
     }
 
+    /**
+     * @param childLineNumber the beginning line number of the first sub node
+     * @param parentLineNumber the beginning line number of the parent node
+     * @param childColumn the column on the {@code childLineNumber} line where the sub node starts
+     * @param parentColumn the column on the {@code parentLineNumber} line where the parent node starts
+     * @return the level of parenthesis parsed from the source
+     */
     private int determineParenthesisLevel(int childLineNumber, int parentLineNumber, int childColumn, int parentColumn) {
+        int saveCursor = cursor;
+        whitespace();
         int childBeginCursor = cursor;
         if (childLineNumber > parentLineNumber) {
             for (int i = 0; i < (childColumn - parentLineNumber); i++) {
@@ -2564,6 +2569,7 @@ public class GroovyParserVisitor {
                 count++;
             }
         }
+        cursor = saveCursor;
         return count;
     }
 
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -763,10 +763,7 @@ public class GroovyParserVisitor {
         }
 
         private Expression insideParentheses(ASTNode node, Function<Space, Expression> parenthesizedTree) {
-            int saveCursor = cursor;
-            whitespace();
             Integer insideParenthesesLevel = getInsideParenthesesLevel(node);
-            cursor = saveCursor;
             if (insideParenthesesLevel != null && insideParenthesesLevel > 0) {
                 Stack<Space> openingParens = new Stack<>();
                 for (int i = 0; i < insideParenthesesLevel; i++) {
@@ -2536,6 +2533,8 @@ public class GroovyParserVisitor {
             return (Integer) rawIpl;
         } else if (node instanceof MethodCallExpression) {
             MethodCallExpression expr = (MethodCallExpression) node;
+            int saveCursor = cursor;
+            whitespace();
             int childBegin = cursor;
             if (expr.getObjectExpression().getLineNumber() != expr.getLineNumber()) {
                 for (int i = 0; i < (expr.getObjectExpression().getLineNumber() - expr.getLineNumber()) - 1; i++) {
@@ -2551,6 +2550,7 @@ public class GroovyParserVisitor {
                     count++;
                 }
             }
+            cursor = saveCursor;
             return count;
         } else {
             return null;
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -34,6 +34,7 @@ import org.openrewrite.groovy.marker.*;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.internal.EncodingDetectingInputStream;
 import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.java.internal.JavaTypeCache;
 import org.openrewrite.java.marker.ImplicitReturn;
 import org.openrewrite.java.marker.OmitParentheses;
@@ -762,8 +763,10 @@ public class GroovyParserVisitor {
         }
 
         private Expression insideParentheses(ASTNode node, Function<Space, Expression> parenthesizedTree) {
+            int saveCursor = cursor;whitespace();
             Integer insideParenthesesLevel = getInsideParenthesesLevel(node);
-            if (insideParenthesesLevel != null) {
+            cursor = saveCursor;
+            if (insideParenthesesLevel != null && insideParenthesesLevel > 0) {
                 Stack<Space> openingParens = new Stack<>();
                 for (int i = 0; i < insideParenthesesLevel; i++) {
                     openingParens.push(sourceBefore("("));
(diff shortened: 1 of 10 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -774,6 +778,8 @@ private Expression insideParentheses(ASTNode node, Function<Space, Expression> p
                             padRight(parenthesized, sourceBefore(")")));
                 }
                 return parenthesized;
+            } else {
+                cursor = saveCursor;
```
Laurens-W, line 782: This is unnecessary

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -762,8 +763,11 @@ private <T> List<JRightPadded<T>> visitRightPadded(ASTNode[] nodes, @Nullable St
         }
 
         private Expression insideParentheses(ASTNode node, Function<Space, Expression> parenthesizedTree) {
+            int saveCursor = cursor;
+            whitespace();
```
Laurens-W, line 767: grabbing whitespace and resetting the cursor is only to determine the parenthesis level correctly, perhaps it should be moved there

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -762,8 +763,11 @@ private <T> List<JRightPadded<T>> visitRightPadded(ASTNode[] nodes, @Nullable St
...
             Integer insideParenthesesLevel = getInsideParenthesesLevel(node);
-            if (insideParenthesesLevel != null) {
+            cursor = saveCursor;
+            if (insideParenthesesLevel != null && insideParenthesesLevel > 0) {
```
jevanlingen, line 770:
```suggestion
            if (insideParenthesesLevel != null) {
```

I think this is a premature optimisation. Only the Stack object will be created and removed from the heap (a for-loop with zero iterations do not do anything). As a rule of thumb, I think if code can be shorter, most of the time it is more readable. You could also choose to let the `getInsideParenthesesLevel` return a `int`. Maybe that's even better.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1350,20 +1356,20 @@ public void visitConstantExpression(ConstantExpression expression) {
...
                 } else if (type == ClassHelper.STRING_TYPE) {
                     jType = JavaType.Primitive.String;
-                    // String literals value returned by getValue()/getText() has already processed sequences like "\\" -> "\"
-                    int length = sourceLengthOfString(expression);
                     // this is an attribute selector
-                    if (source.startsWith("@" + value, cursor)) {
-                        length += 1;
+                    boolean attributeSelector = source.startsWith("@" + value, cursor);
+                    int length = lengthAccordingToAst(expression);
+                    Integer insideParenthesesLevel = getInsideParenthesesLevel(expression);
```
jevanlingen, line 1362: How is this even possible? The function is already wrapped in:
```
queue.add(insideParentheses(expression, fmt -> {
```
so I would imagine the parenthesis already being stripped?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
Laurens-W: This is actually a leftover of the mechanism to reset the cursor and can be removed

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1350,20 +1356,20 @@ public void visitConstantExpression(ConstantExpression expression) {
...
-                    if (text.startsWith("$/")) {
-                        delimiterLength = 2;
-                    } else if (text.startsWith("\"\"\"") || text.startsWith("'''")) {
-                        delimiterLength = 3;
-                    } else if (text.startsWith("/") || text.startsWith("\"") || text.startsWith("'")) {
-                        delimiterLength = 1;
+                    String valueAccordingToAST = source.substring(cursor, cursor + length + (attributeSelector ? 1 : 0));
+                    int delimiterLength = getDelimiterLength();
+                    if (StringUtils.containsWhitespace(valueAccordingToAST)) {
+                        length = delimiterLength + ((String) expression.getValue()).length() + delimiterLength;
```
jevanlingen, line 1369:
```suggestion
                        length = delimiterLength + expression.getValue().toString().length() + delimiterLength;
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2522,14 +2528,34 @@ private int sourceLengthOfString(ConstantExpression expr) {
...
         Object rawIpl = node.getNodeMetaData("_INSIDE_PARENTHESES_LEVEL");
         if (rawIpl instanceof AtomicInteger) {
             // On Java 11 and newer _INSIDE_PARENTHESES_LEVEL is an AtomicInteger
             return ((AtomicInteger) rawIpl).get();
-        } else {
+        } else if (rawIpl instanceof Integer) {
             // On Java 8 _INSIDE_PARENTHESES_LEVEL is a regular Integer
             return (Integer) rawIpl;
+        } else if (node instanceof MethodCallExpression) {
+            MethodCallExpression expr = (MethodCallExpression) node;
```
jevanlingen, line 2540: IntelliJ comment: 
> It's possible to extract method returning 'count' from a long surrounding method.

I think that's true it makes it more readable. Consider:

```java
private @Nullable Integer getInsideParenthesesLevel(ASTNode node) {
    Object rawIpl = node.getNodeMetaData("_INSIDE_PARENTHESES_LEVEL");
    if (rawIpl instanceof AtomicInteger) {
        // On Java 11 and newer _INSIDE_PARENTHESES_LEVEL is an AtomicInteger
        return ((AtomicInteger) rawIpl).get();
    } else if (rawIpl instanceof Integer) {
        // On Java 8 _INSIDE_PARENTHESES_LEVEL is a regular Integer
        return (Integer) rawIpl;
    } else if (node instanceof MethodCallExpression) {
        return getInsideParenthesesLevel((MethodCallExpression) node);
    }
    return null;
}

private int getInsideParenthesesLevel(MethodCallExpression node) {
    // impl 
}
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
Laurens-W: The `sourceLengthOfString` method previously used here did this under water as well. Notice we're visiting a `ConstantExpression` here, so this doesn't collide with the logic specifically for `MethodCallExpression`s, it merely grabs the value provided by the AST!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2522,14 +2528,34 @@ private int sourceLengthOfString(ConstantExpression expr) {
...
+            int childBegin = cursor;
+            if (expr.getObjectExpression().getLineNumber() != expr.getLineNumber()) {
```
jevanlingen, line 2539:
```suggestion
            if (expr.getObjectExpression().getLineNumber() >= expr.getLineNumber()) {
```
Just to catch some really weird/unexpected line numbering.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2522,14 +2523,37 @@ private int sourceLengthOfString(ConstantExpression expr) {
...
+            int count = 0;
+            for (int i = cursor; i < childBegin; i++) {
+                if (source.charAt(i) == '(') {
+                    count++;
+                }
+            }
+            cursor = saveCursor;
+            return count;
+        } else {
+            return null;
```
jevanlingen, line 2556:
```suggestion
        }
        return null;
```

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/CastTest.java
```diff
@@ -25,6 +24,18 @@
 @SuppressWarnings({"UnnecessaryQualifiedReference", "GroovyUnusedAssignment", "GrUnnecessarySemicolon"})
 class CastTest implements RewriteTest {
 
+    @Test
+    void javaStyleCastNew() {
```
jevanlingen, line 28:
```suggestion
    void cast() {
```

jevanlingen:
Very nice work, at last we have a working implementation!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2522,15 +2529,42 @@ private int sourceLengthOfString(ConstantExpression expr) {
...
+            int saveCursor = cursor;
+            whitespace();
+            int count = determineParenthesisLevel(expr.getObjectExpression().getLineNumber(), expr.getLineNumber(), expr.getObjectExpression().getColumnNumber(), expr.getColumnNumber());
+            cursor = saveCursor;
+            return count;
+        }
+        return null;
+    }
+
+    private int determineParenthesisLevel(int childLineNumber, int parentLineNumber, int childColumn, int parentColumn) {
```
jevanlingen, line 2551: To be honest, I don't think this abstraction is right, because the cursor changes are implied from outside of this function. With other words, this function does only work well if you use it exactly like it is meant to be.

I would go the `MethodCallExpression` as argument. That way it's more in line with the `getInsideParenthesesLevel(ASTNode node` (just an overloaded function with more explicit Type) and it's also works from everywhere because the cursor travel in all kept within this function. But for starters, I would include the `saveCursor` part within this function. Changing the argument to `MethodCallExpression` is up to you!

Laurens-W: I moved the `saveCursor` and related `whitespace` parts and added JavaDoc to the method. This setup makes it easier to reuse when another case is found where the AST falls short

knutwannheden (approved):

jevanlingen (approved):

PR #4804, opened 2024-12-20
suppress javax.json

pstreef:


suppressions.xml
```diff
@@ -39,4 +39,12 @@
         <packageUrl regex="true">^pkg:maven/org\.glassfish/javax\.json@.*$</packageUrl>
         <vulnerabilityName>CVE-2023-7272</vulnerabilityName>
     </suppress>
+    <suppress until="2025-12-20">
```
timtebeek, line 42: How long do we usually suppress these for?

pstreef, line 42:
```suggestion
    <suppress until="2025-03-20">
```

pstreef: not sure about the rewrite side, but I made it a little shorter.

PR #4805, opened 2024-12-20
Refactor SpringReference

Laurens-W:
## What's changed?

Distinguish between what is XmlReference and what is Spring specific (Matcher and Provider)

@timtebeek @jevanlingen 

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringXmlReference.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringXmlReference.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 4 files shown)


Laurens-W:
Merging this will probably cause conflicts here
- https://github.com/openrewrite/rewrite/pull/4793

timtebeek (approved):

rewrite-xml/src/main/java/org/openrewrite/xml/trait/SpringXmlReference.java
```diff
@@ -31,55 +28,9 @@
 import java.util.Set;
 import java.util.regex.Pattern;
 
-@Value
-class SpringReference implements Reference {
-    Cursor cursor;
-    Kind kind;
+public class SpringXmlReference {
```
jevanlingen, line 31: I think I should extend the `XmlReference` here.

Laurens-W: I was in doubt about this because in this case the XmlReference and SpringXmlReference perfectly fit eachother. When you run into a XmlReference that needs a different implementation of `getValue` it would make sense to introduce this inheritance. So I agree we might as well start now :)

PR #4807, opened 2024-12-20
Add fallback for determining parenthesis level for `BinaryExpression`

Laurens-W:
## What's changed?

Added fallback for determining parenthesis level for `BinaryExpression`

@timtebeek @jevanlingen @knutwannheden @sambsnyd 

## Any additional context

Lower versions of Gradle do not provide the `_INSIDE_PARENTHESES_LEVEL` flag we expect

### Checklist


jevanlingen (approved):

timtebeek (approved):
Approved already, but let's await the results you get locally with the jar I've just sent.

jevanlingen:
Maybe it's also a good idea to add an extra comment above the usage of `_INSIDE_PARENTHESES_LEVEL`. Something like:

```
private static @Nullable Integer getInsideParenthesesLevel(ASTNode node) {
  // Grab level from AST if provided (older groovy does not set this flag at all, newer versions omit this flag for MethodCallExpression nodes)
  Object rawIpl = node.getNodeMetaData("_INSIDE_PARENTHESES_LEVEL");
   ..
}
```

PR #4809, opened 2024-12-21
Add recipe to remove Gradle Enterprise and Develocity

timtebeek:
## What's your motivation?
We had recipes to add and migrate, but lacked recipes to remove. Some internal repositories for now can not use Gradle Enterprise/Develocity, so I'd like to remove these with a recipe.


shanman190:
Nice addition! It looks like the after assertion needs a small edit as the source file does not become 100% empty, but otherwise it looks good.

PR #4813, opened 2024-12-25
Support mapping of generic thrown exception types

jkschneider:
## What's changed?

Widened the type of `JavaType.Method#thrownExceptions` to `JavaType` to support the mapping of generic exceptions.

A method that throws a combination of a generic and non-generic exception looks like this:

```java
interface Test {
   <T extends FileNotFoundException> void throwsGenericException() throws T, InterruptedException;
}
```

Its [JVMS 4.3.4](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.4) signature is:

```
<T:Ljava/io/FileNotFoundException;>()V^TT;^Ljava/lang/InterruptedException;
```

## What's your motivation?

Preparing the way for compile verification as a post-recipe task.

Should be LST deserialization compatible, but this is an ABI change and we should release workers and CLI ASAP. Old serialized LSTs will not have mapped generic exceptions correctly. They instead previously mapped the class bound (in the example above `java.io.FileNotFoundException`.

### Checklist

### DIFF
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -511,7 +511,7 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
 
         JavaType returnType = null;
         List<JavaType> parameterTypes = null;
-        List<JavaType.FullyQualified> exceptionTypes = null;
+        List<JavaType> exceptionTypes = null;
 
         if (selectType instanceof Type.MethodType) {
             Type.MethodType methodType = (Type.MethodType) selectType;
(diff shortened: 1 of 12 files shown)


rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
```diff
@@ -634,20 +622,8 @@ public JavaType.Primitive primitive(TypeTag tag) {
                 if (!methodType.thrown.isEmpty()) {
                     exceptionTypes = new ArrayList<>(methodType.thrown.size());
                     for (Type exceptionType : methodType.thrown) {
-                        JavaType.FullyQualified javaType = TypeUtils.asFullyQualified(type(exceptionType));
```
jkschneider, line 637: This logic appears to be a holdover from a time when we had `JavaType.Cyclic`, so I believe is safe to simplify now.

PR #4819, opened 2024-12-27
Put parenthesis changes back

jevanlingen:
## What's changed?
- **Rereverted**: parenthesis changes 
- **Fix**: closing parentheses in `determineParenthesisLevel` not taking into account (like `new ObjectX().objectY.someClosure {}`
- **Fix**: for-loop could not handle modifiers like `for(def x : ["A", "B"])`
- **Improvement**: Support for varargs

## What's your motivation?
- https://github.com/openrewrite/rewrite/pull/4818

## Out of scope
When I added the real-live-examples, I also found out we do not support other things:
- [Pattern](https://groovy-lang.org/operators.html#_pattern_operator) operator
-  [anonymous inner classes](https://groovy-lang.org/objectorientation.html#_anonymous_inner_class )
- Java style [class](https://groovy-lang.org/style-guide.html#_classes_as_first_class_citizens) arguments

As these things are possible not easy fixes, I kept the unit tests, but marked them as not-supported-yet. Follow up PRs should fix these.

### Checklist
- [x] I've used the IntelliJ IDEA auto-formatter on affected file

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2543,7 +2543,6 @@ public class GroovyParserVisitor {
         } else if (node instanceof BinaryExpression) {
             BinaryExpression expr = (BinaryExpression) node;
             return determineParenthesisLevel(expr.getLeftExpression().getLineNumber(), expr.getLineNumber(), expr.getLeftExpression().getColumnNumber(), expr.getColumnNumber());
-
         }
         return null;
     }
@@ -2571,6 +2570,9 @@ public class GroovyParserVisitor {
         for (int i = cursor; i < childBeginCursor; i++) {
             if (source.charAt(i) == '(') {
                 count++;
+            }
+            if (source.charAt(i) == ')') {
+                count--;
             }
         }
         cursor = saveCursor;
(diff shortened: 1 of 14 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/RealWorldGroovyTest.java
```diff
@@ -0,0 +1,422 @@
...
+import static org.openrewrite.groovy.Assertions.groovy;
+
+/**
+ * A test with groovy examples picked from OSS repositories.
+ */
+class RealWorldGroovyTest implements RewriteTest {
+
+    @Test
+    @ExpectedToFail("Pattern operator is not yet supported") // https://groovy-lang.org/operators.html#_pattern_operator
+    @Issue("https://github.com/spring-projects/spring-boot/blob/main/settings.gradle")
```
timtebeek, line 32: Can we update these links to make them stable in light of any future changes?
```suggestion
    @Issue("https://github.com/spring-projects/spring-boot/blob/v3.4.1/settings.gradle")
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2595,6 +2653,7 @@ private TypeTree visitTypeTree(ClassNode classNode, boolean inferredType) {
 
         int saveCursor = cursor;
         Space fmt = whitespace();
+        // TODO: remove this? At least use `getModifiers` function
```
timtebeek, line 2656: Is this something you'd intended to pick up later? Or for this PR still?

jevanlingen: Later, I saw this line and I thought it was not good enough. At least just checking for `def` here is not enough.

timtebeek (approved):

PR #4820, opened 2024-12-30
Use `@org.jspecify.annotations.NonNull` for `@org.openrewrite.internal.lang.NonNull` instead of `@javax.annotation.Nonnull`

izeye:
## What's changed?

- This PR polishes the changes made in gh-4415 a bit.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/lang/NonNull.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/lang/NonNull.java
@@ -18,14 +18,13 @@ package org.openrewrite.internal.lang;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
-import javax.annotation.Nonnull;
 import javax.annotation.meta.TypeQualifierNickname;
 import java.lang.annotation.*;
 
 /**
  * A common annotation to declare that annotated elements cannot be {@code null}.
- * Leverages JSR 305 meta-annotations to indicate nullability in Java to common tools with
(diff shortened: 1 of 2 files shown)


timtebeek (commented):
Thanks for noticing and applying this change @izeye! I do recall seeing issues with 
https://github.com/openrewrite/rewrite/blob/0805fefab5526d9741597233a0433e30d33055b2/rewrite-java/src/main/java/org/openrewrite/java/style/EmptyBlockStyle.java#L28-L30
and
https://github.com/openrewrite/rewrite/blob/0805fefab5526d9741597233a0433e30d33055b2/rewrite-java/src/main/java/org/openrewrite/java/style/OperatorWrapStyle.java#L28-L33
which are the last two to use our internal `@NonNull` annotation changed here. Those changes had previously been reverted in https://github.com/openrewrite/rewrite/commit/0805fefa. That means we'll likely need to keep a closer eye on this after a merge, which is then perhaps best merged when everyone's back to work next week.

rewrite-core/src/main/java/org/openrewrite/internal/lang/NonNull.java
```diff
@@ -40,7 +39,7 @@
 @Target({ElementType.METHOD, ElementType.PARAMETER, ElementType.FIELD, ElementType.TYPE, ElementType.TYPE_USE})
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Nonnull
+@org.jspecify.annotations.NonNull
```
timtebeek, line 42: I'm seeing this warning when opening up the class in my IDE, so I don't think we can adopt this as is.
image

PR #4822, opened 2024-12-30
`AddAnnotationProcessor` should compare and update versions using a property

nielsdebruin:
## What's changed?
When adding a Maven annotation processor, handle a version specified as a property correctly.

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
@@ -178,4 +178,96 @@ class AddAnnotationProcessorTest implements RewriteTest {
           )
         );
     }
+
+
+    @Test
+    void addAnnotationWithVersionAsMavenProperty() {
+        rewriteRun(
+          spec -> spec.recipe(new AddAnnotationProcessor(
+            "org.projectlombok",
+            "lombok-mapstruct-binding",
+            "0.2.0"
+          )),
+          pomXml(
+            """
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
```diff
@@ -245,20 +230,50 @@ void addAnnotationWithVersionAsMavenProperty() {
...
-                                          <groupId>org.mapstruct</groupId>
-                                          <artifactId>mapstruct-processor</artifactId>
-                                          <version>${version.mapstruct}</version>
-                                      </path>
                                       <path>
                                           <groupId>org.projectlombok</groupId>
-                                          <artifactId>lombok</artifactId>
-                                          <version>${version.lombok}</version>
+                                          <artifactId>lombok-mapstruct-binding</artifactId>
+                                          <version>0.2.0</version>
```
nielsdebruin, line 236: This is not what we want, right?

timtebeek (approved):

rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
timtebeek: Indeed better to update the version tag; this was a first quick attempt at getting rid of the comparison failure. Improvement picked up in https://github.com/openrewrite/rewrite/pull/4822/commits/7c2fcd29d3a3a973e7b33a59263f4819d7807de0

PR #4823, opened 2024-12-31
Add support for lombok @Builder.Default Annotation

amishra-u:
## What's changed?
Added support for lombok Annotation @Builder.Default 

### Issues
The JavaParser fails to parse source files containing the @Builder.Default annotation.
1.  The Lombok handler for @Builder sets the [init expression to null](https://github.com/projectlombok/lombok/blob/master/src/core/lombok/javac/handlers/HandleBuilder.java#L814) for fields annotated with @Builder.Default. As a result, the AST-to-LST converter never see this init expression, since it was removed from the original code.
2. The Lombok handler uses a [hack](https://github.com/projectlombok/lombok/blob/master/src/core/lombok/javac/handlers/HandleBuilderDefault.java#L53) to convert @Builder.Default into a fully qualified name, which disrupts the AST positions.

### Fix
1. Added logic to Remove the @Builder.Default annotation before invoking the Lombok @Builder handler and restored the original value after processing.
2. Updated the @Builder.Default annotation handler to NoOp, as it doesn't produce any output necessary for type attribution.

### Checklist

### DIFF
--- a/rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/BuilderDefaultNoOpHandler.java
+++ b/rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/BuilderDefaultNoOpHandler.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks a lot @amishra-u ! Seems like a perfectly measured approach.

amishra-u:
I was reading a reddit thread, found claim of lombok creator that lombok does't modify existing ones.
https://www.reddit.com/r/java/comments/170gdha/comment/k3lq6p4/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

In this case, this issue is a design bug for Builder.Default annotation.

Issue #4704, opened 2024-11-23
Groovy parser fails to parse constructor declarations


sambsnyd:
I haven't categorized exactly _why_ this fails to parse yet in a way that is fully distinct from other issues, but this code should parse and instead throws an exception:

```groovy
class Pair {
    String foo
    String bar
    Pair(String foo, String bar) {
        this.foo = foo
        this.bar = bar
    }
}

class A {
    def foo(List l) {
        l.add(new Pair("foo", "bar"))
        l.add(new Pair("foo", "bar"))
        l.add(new Pair("foo", "bar"))
        l.add(new Pair("foo", "bar"))
    }
}
```

Reproduced in test [CompilationUnitTest.addingToMaps()](https://github.com/openrewrite/rewrite/blob/2b05fc22d1236e7b2ee51d8c6972d7f7c63e5bcf/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/CompilationUnitTest.java#L130)

Issue #4063, opened 2024-03-01
Support for inner classes in Groovy


Jnb1234:
## What version of OpenRewrite are you using?

I am using id 'org.openrewrite.rewrite' version '6.8.4'

- Gradle plugin 6.8.4

## How are you running OpenRewrite?

I am using the Gradle plugin, and my project is a single module project.
```xml
plugins {
    id 'maven-publish'
    id 'groovy'
    id 'idea'
    id 'net.researchgate.release' version '2.8.1'
    id "com.github.ben-manes.versions" version "0.36.0"
    id "com.gorylenko.gradle-git-properties" version "2.3.2"
    id 'java-library'
    id 'org.openrewrite.rewrite' version '6.8.4'
}
rewrite {
    activeRecipe("org.openrewrite.java.RemoveUnusedImports")
    }
```

## What is the smallest, simplest way to reproduce the problem?

```java
package com.gaic.bue.uwd.ra.common.dataaccess

class JnbInnerClassIssue
{

    class ShowParseIssue {
        String a

//        String getA() {
//            return a
//        }

        ShowParseIssue(String a) {
            this.a = a
        }

//Moving getA() before the constructor or after the setter setA(String a) seems to resolve the parsing issue
        String getA() {
            return a
        }

        void setA(String a) {
            this.a = a
        }

//        String getA() {
//            return a
//        }
    }
}
```

## What did you expect to see?

No parsing errors

## What did you see instead?

There were problems parsing src\test\java\com\gaic\bue\uwd\ra\common\dataaccess\JnbInnerClassIssue.groovy

## What is the full stack trace of any errors you encountered?

```
There were problems parsing some source files, run with --info to see full stack traces
There were problems parsing src\test\java\com\gaic\bue\uwd\ra\common\dataaccess\JnbInnerClassIssue.groovy
Error during rewrite dry run
No further details 

Note there are other groovy classes with parsing issues which I am trying to create small examples since the code is private
```

timtebeek:
Confirmed with a unit test in 
https://github.com/openrewrite/rewrite/blob/33d8f7d42f9bd6749ab93171fad31289b0e2bc97/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/ClassDeclarationTest.java#L31
```java
@Test
@Issue("https://github.com/openrewrite/rewrite/issues/4063")
void nestedClass() {
    rewriteRun(
      groovy(
        """
[...]

PR #4825, opened 2024-12-31
Support inner classes for groovy

jevanlingen:
## What's changed?
Inner classes our now totally supported (including anonymous classes).

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4063

Most work to support inner classes was already done, the new code just adds some missing pieces.

## Any additional context
To support
- https://github.com/openrewrite/rewrite/issues/4704

 a `visitConstructor` was added. The code was 95% the same as the `visitMethod`. So I merged the `visitMethod` and `visitConstructor` again (a _ConstructorNode_ extends a _MethodNode_, so _visitMethod_ is enough).

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -528,6 +528,20 @@ public class GroovyParserVisitor {
             Space namePrefix = whitespace();
             String methodName;
             if (isConstructor) {
+                /*
+                To support Java syntax for non-static inner classes, the groovy compiler uses an extra parameter with a reference to its parent class under the hood:
+                class A {                               class A {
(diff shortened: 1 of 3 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -599,6 +621,11 @@ varargs, emptyList(),
                     Markers.EMPTY
             );
 
+            if (isConstructorOfInnerNonStaticClass) {
+                ((BlockStatement) method.getCode()).getStatements().remove(0);
+                ((BlockStatement) method.getCode()).getStatements().remove(0);
```
jevanlingen, line 626: Should we do this neater by making a copy except the first two items? Or is this fine enough?

timtebeek (approved):

PR #4826, opened 2024-12-31
Don't skip type attribution on compilation error

amishra-u:
## What's changed?
Modified the compilation process to continue even when some steps encounter errors.

## What's your motivation?
While enabling Lombok annotation processing, custom logic was introduced to skip the processing of certain annotation arguments. This led to errors during the process step of the compilation. By default, the compiler [halts](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/main/JavaCompiler.java#L470) at any step encountering an error, which caused the subsequent type attribution step to be skipped entirely. As a result, a single error in the **process** step affected the entire project.

### Changes
1. Introduced a flag to allow compilation to continue despite errors in any step. (Applied changes to Java 11 and Java 17 parsers as Lombok is supported only for these versions.)
2. Added a new field in TypeValidation to enable finer control, allowing specific nodes to bypass type validation.
3. Added unit test to verify behavior for Lombok annotations with skipped arguments.

Before this change, a single skipped annotation argument in project would result in missing type attribution for the entire project.

### Checklist

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
@@ -118,6 +118,11 @@ public class ReloadableJava11Parser implements JavaParser {
         Options.instance(context).put("-g", "-g");
         Options.instance(context).put("-proc", "none");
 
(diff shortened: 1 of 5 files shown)


timtebeek (approved):
Nice and measured improvement once again @amishra-u !

PR #4831, opened 2025-01-02
Hcl - allow for expression on LHS in objectelem

greg-at-moderne:
### Checklist


timtebeek (approved):
Great to see, thanks! I've not looked at any of the internal classes beyond the HclParserVisitor, assuming those are all generated. If I should have a closer look there feel free to call out anything with a self-review. If not, then we're ready to merge. 🚀

greg-at-moderne:
Yep. All changes in `rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/grammar` are generated code based on the `.g4` lexer/parser.

PR #4549, opened 2024-10-04
Allow parsing nested Groovy annotations

knutwannheden:
See: https://github.com/openrewrite/rewrite/issues/4055#issuecomment-2392913541

PR #4832, opened 2025-01-02
Support nested annotations with properties and annotations with empty arguments

jevanlingen:
## What's changed?
Support for:
- nested annotations with properties, like: `@Foo(bar = @Bar(value = "1.0"))`
- annotations with empty arguments, like `@Foo()`

## What's your motivation?
- fix https://github.com/openrewrite/rewrite/issues/4055

## Any additional context
The [solution](https://github.com/openrewrite/rewrite/pull/4549) of using
```java
public void visitConstantExpression(ConstantExpression expression) {
  ..
  else if (expression instanceof AnnotationConstantExpression) {
    classVisitor.visitAnnotation((AnnotationNode) value);
    return ((Expression) classVisitor.pollQueue()).withPrefix(fmt); // pollQueue returns `J.Annotation`
  }

  return new J.Literal(randomId(), fmt, Markers.EMPTY, value, text, null, jType);
}
```
did not work properly, because it just fooled the compiler (both J.Annotation and J.Literal extend Expression). But in the end all ConstantExpressions are turned into a J.Literal, so the nested annotations were turned to Strings.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -60,6 +60,7 @@ import java.util.stream.Stream;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.toList;
 import static org.openrewrite.Tree.randomId;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see, thanks a lot!

PR #4834, opened 2025-01-02
Improve discovery of parentheses level for MethodCallExpression

jevanlingen:
## What's changed?
Parentheses discovery for method calls did not always work properly.

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4055

Failing parts:

```groovy
Micronaut.build(args)
    .packages('com.edalex')
    .mainClass(Application)
    .start()
```

## Any additional context
The discovery of parentheses is quite hard to grasp. This is the concept behind it. We defined two variables called `parent` and `child`. The parent and child are nodes, which could be visualised like: `<<child>  parent>`. Notice in source code, the nodes could span multiple lines.

To support this, we loop through all the lines of code of the parent and child element:

```
for (int i = 0; i < (childLineNumber - parentLineNumber); i++)
  // move `untilCursor` past next line break every time

// move `untilCursor` after childColumn of last line
```

### Checklist


timtebeek (approved):
Nice! Really hammering out those fixes today.

PR #4835, opened 2025-01-02
Add Support for Builder.Default annotation for final vars

amishra-u:
## What's changed?
Added support for the Lombok annotation `@Builder.Default` on final variables.

## What's your motivation?
- In #4823, the `@Builder.Default` annotation was temporarily removed for processing. However, Lombok's HandleBuilder  [skips](https://github.com/projectlombok/lombok/blob/9f5de53e0651142d1fa7a981f221f51a627f41d0/src/core/lombok/javac/handlers/HandleBuilder.java#L281) processing final variables with non-null initialization values unless they include the `@Builder.Default `annotation. To address this, added logic to temporarily remove the final modifier before invoking the annotation processor and restored it afterward.

### Checklist


timtebeek (approved):
Thanks again! Looks like the better fix indeed.

PR #4837, opened 2025-01-03
Allow to add Maven test scoped dependency

### LINKED ISSUE #4836: Maven `AddDependency` has an input `scope` that does not include `test`
I ran across this because we have a recipe that is adding a test scoped dependency. The recipe works correctly but the rewrite plugin in IntelliJ marks it as an error: 

image

This is a simple one line change here: 

See https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L81

timtebeek:
- Fixes #4836


timtebeek:
Seems silly that we had tests that use `"test"` already but did not allow the value before 🙃 ; thanks for noticing @tkvangorder !
https://github.com/openrewrite/rewrite/blob/8b44a1d343b264be2da571595c61268ffc776730/rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java#L1095-L1106

PR #4838, opened 2025-01-03
Cleanup groovy parser

jevanlingen:
## What's changed?
- Remove `def` check for visitTypeTree
- Remove unnecessary RedundantDef marker for DeclarationExpression
- Made RedundantDef function more specific for MethodNodes
    - Added support for `def` keyword for constructors (see [Def and type](https://groovy-lang.org/style-guide.html#_def_and_type))
- Use `skip` function as much as possible

## What's your motivation?
After the changes @Laurens-W made in
- https://github.com/openrewrite/rewrite/pull/4757

the parser did neither need the RedundantDef markers for DeclarationExpressions nor the `def` check for visitTypeTree. So I removed it and rewrote the RedundantDef stuff for methods (as it was the only usage of the `maybeRedundantDef` function, it could be written more specific targeted to method nodes only). 

When I was busy, I saw we could use the `skip` function a lot more than we do right now (instead of e.g. `cursor += 2`). So I also took some time to do that as well.

### Checklist


timtebeek (approved):
Neat readability improvement! That skip that takes a string is a lot better than cursor arithmetic all over the place. Thanks for taking this on, and doing so in a separate PR. 🙏🏻

PR #4843, opened 2025-01-03
Fix parsing string literal with line breaks and spaces

jevanlingen:
## What's changed?
- Fix: String with line breaks and spaces, like `"\n|  "` are supported as well.
- Cleanup: String literals and the String literals in the GString parser are written in the same way

### Checklist


jevanlingen:
I think it's possible to remove the `lengthAccordingToAst` function, the trick I did for String should also work at the GString part. So I'll give this a try before merging this part.

shanman190 (approved):

PR #4846, opened 2025-01-03
Adding option cutOffDate to RemoveOwaspSuppressions

lkerford:
There are cases were we want to only remove suppressions up to a specific date, this allows a cut off date to be provided when removed expired suppressions


rewrite-xml/src/main/java/org/openrewrite/xml/security/RemoveOwaspSuppressions.java
```diff
@@ -39,11 +37,18 @@ public String getDisplayName() {
         return "Remove out-of-date OWASP suppressions";
     }
 
+    @Option(displayName = "Until date",
+            required = false,
+            description = "Optional. Suppressions will be removed if they expired before the provided date. Default will be yesterday.",
```
timtebeek, line 42: There's no need to add the Optional. prefix here; that's added in the markdown generator.
https://github.com/openrewrite/rewrite-recipe-markdown-generator/blob/b0c0897c7f31d8d59f1e08b2d6757f443853c27b/src/main/kotlin/org/openrewrite/RecipeMarkdownGenerator.kt#L1151

```suggestion
            description = "Suppressions will be removed if they expired before the provided date. Default will be yesterday.",
```

timtebeek (approved):

Issue #4254, opened 2024-06-13
Groovy generated transform methods break parser


gtiwari333:
I'm seeing this on groovy class that uses groovy transform (`import groovy.transform.*`) such as @ToString, @EqualsAndHashCode.

## What version of OpenRewrite are you using?
 
I am using https://github.com/gtiwari333/openrewrite-groovy-bug/blob/main/init.gradle

- org.openrewrite.recipe:rewrite-recipe-bom  - 2.1.20
- org.openrewrite:rewrite-gradle-plugin - latest.release
- groovy 4.0.21
- jdk 17/21
- gradle 8.8

 
 

## How are you running OpenRewrite?
 `%./gradlew  clean rewriteRun --init-script init.gradle --info`

 

## What is the smallest, simplest way to reproduce the problem?

Checkout the following project and run `./gradlew  clean rewriteRun --init-script init.gradle --info`

Im seeing similar behavior with both jdk17 and jdk21 

https://github.com/gtiwari333/openrewrite-groovy-bug/blob/main/README.md 

## What did you expect to see?
 No failures. AutoFormat recipe should run on the code.

## What did you see instead?
 ```
Error during rewrite run
org.openrewrite.groovy.GroovyParsingException: Failed to parse src/main/groovy/com/example/demo/Test.groovy, cursor position likely inaccurate.
        at org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:170)
        at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$4(GroovyParser.java:154)
        at java.base@17.0.9/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base@17.0.9/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
 ...
Caused by: java.lang.StringIndexOutOfBoundsException: begin 137, end 137, length 134
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4606)
        at java.base/java.lang.String.substring(String.java:2709)
        at org.openrewrite.groovy.GroovyParserVisitor.whitespace(GroovyParserVisitor.java:2171)
        at org.openrewrite.groovy.GroovyParserVisitor.access$000(GroovyParserVisitor.java:66)
        at org.openrewrite.groovy.GroovyParserVisitor$RewriteGroovyVisitor.visitVariableExpressionType(GroovyParserVisitor.java:1955)
[...]

PR #4848, opened 2025-01-06
Skip parsing groovy generated transform methods

jevanlingen:
## What's changed?
Generated [transform](https://medium.com/@AlexanderObregon/exploring-groovys-powerful-ast-transformations-enhancing-your-code-at-compile-time-32bdcd8924a3) annotations are parsed.

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4254

## Any additional context
Notice with these fixes the parser just ignores the groovy AST transformations. So you cannot use the 'real' AST as the bytecodes dictate in the groovy recipes. If we want to support that, we would have to do a project like `rewrite-lombok`.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -643,18 +643,15 @@ public class GroovyParserVisitor {
 
             List<J.Annotation> paramAnnotations = new ArrayList<>(node.getAnnotations().size());
             for (AnnotationNode annotationNode : node.getAnnotations()) {
-                // The groovy compiler can add or remove additional annotations for AST transforms.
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2708,13 +2719,17 @@ private J.TypeParameter visitTypeParameter(GenericsType genericType) {
 
     /**
      * Sometimes the groovy compiler inserts phantom elements into argument lists and class bodies,
-     * presumably to pass type information around. These elements do not appear in source code and should not
-     * be represented in our AST.
+     * presumably to pass type information around. Other timers the groovy compiler add extra transform annotations.
```
timtebeek, line 2722:
```suggestion
     * presumably to pass type information around. Other times the groovy compiler adds extra transform annotations.
```

timtebeek (approved):

timtebeek (approved):

PR #4854, opened 2025-01-06
WIP: Fixing the estimated saved time in the gradle wrapper upgrade

lkerford:
## What's changed?
Do not take into consideration changes which are made is association with updating the markers.

## What's your motivation?
Currently the saved time in the Gradle wrapper recipe much larger than it should be

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapperFiles.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapperFiles.java
@@ -1,0 +1,498 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapperFiles.java
```diff
@@ -0,0 +1,498 @@
...
+@EqualsAndHashCode(callSuper = false)
+public class UpdateGradleWrapperFiles extends ScanningRecipe<UpdateGradleWrapperFiles.GradleWrapperState> {
+
+    @Override
+    public String getDisplayName() {
+        return "Update Gradle wrapper";
+    }
+
+    @Override
+    public String getDescription() {
```
lkerford, line 60: For the name and description of this class, I'm not certain how to different it in a meaningful way to users you aren't familiar with the implementation. Is there a way we can add the recipes to the jar but not have the available to be executed directly?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapperMarkers.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * This recipe only updates markers, so it does not correspond to human manual effort.
+     *
+     * @return Zero estimated time.
+     */
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(10);
```
sambsnyd, line 120: Comment says 0, code says 10 minutes

sambsnyd (commented):
My inclination is to try and make a more general fix for time estimates. Rather than refactor every marker-update into a zero-estimate recipe have the effort calculation omit marker-only changes.

lkerford:
I'm closing the change because I agree with Sam comment and think that we should solve this issue in a different location 

> My inclination is to try and make a more general fix for time estimates. Rather than refactor every marker-update into a zero-estimate recipe have the effort calculation omit marker-only changes.

PR #4856, opened 2025-01-07
Add recipe to enable Develocity build cache in xml configuration

nielsdebruin:
## What's changed?

A new recipe has been added that will add build-cache config to the develocity configuration xml file. If any previous configuration is present, no changes will be made.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/EnableBuildCache.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/EnableBuildCache.java
@@ -23,41 +23,48 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.xml.XmlVisitor;
 import org.openrewrite.xml.tree.Xml;
 
-
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class EnableBuildCache extends Recipe {
 
     @Override
     public String getDisplayName() {
-        return "Enable devolocity build cache";
+        return "Enable Develocity build cache";
     }
 
     @Override
     public String getDescription() {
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/EnableBuildCache.java
```diff
@@ -0,0 +1,103 @@
...
+                if ("develocity".equals(rootTag.getName()) && !rootTag.getChild("buildCache").isPresent()) {
+                    Xml.Tag tag = Xml.Tag.build(getBuildCacheConfig());
+                    rootTag = maybeAutoFormat(rootTag, rootTag.withContent(ListUtils.concat(rootTag.getChildren(), tag)), ctx);
+                    return document.withRoot(rootTag);
+                }
+                return document;
+            }
+        };
+
+        return Preconditions.check(new FindSourceFiles("*.xml"), visitor);
```
Laurens-W, line 78: Can we limit this further to a specific directory? Perhaps using `isAcceptable` or a more specific pattern passed to `findSourceFiles`?

nielsdebruin: I am not sure about this, docs state:

> Similarly to the extensions.xml file, the develocity.xml file **can** be placed in your Maven project’s .mvn directory. 

As it says _can_ not _must_ I am still unsure whether a further restriction to `.mvn` is safe.

timtebeek (approved):
Fixed a couple small issues; with that we're good to merge. Thanks @nielsdebruin !

PR #4859, opened 2025-01-07
Add recipe to enable Develocity build cache for Gradle

nielsdebruin:
## What's changed?

Recipe to enable Develocity build cache, functionality is identical to Maven equivalent merged in:
- https://github.com/openrewrite/rewrite/pull/4856

## Any additional context

- https://docs.gradle.com/develocity/gradle-plugin/current/

### Checklist


rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,71 @@
...
+public class EnableDevelocityBuildCache extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Enable Develocity build cache";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Add Develocity build cache configuration to any `.mvn/` Develocity configuration file that lack existing configuration.";
```
timtebeek, line 34:
```suggestion
        return "Add Develocity build cache configuration where missing.";
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/EnableDevelocityBuildCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+                      }
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void addBuildCacheLocalConfig() {
```
shanman190, line 58: The configuration example here isn't valid. Local doesn't accept the `develocity.buildCache` argument. Local is also default enabled with the caveat that the Gradle property or CLI switch must also be enabled for the feature to turn on.

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,130 @@
...
+    public Validated<Object> validate(ExecutionContext ctx) {
+        return super.validate(ctx)
+                .or(Validated.notBlank("remoteEnabled", remoteEnabled)
+                        .or(Validated.notBlank("remotePushEnabled", remotePushEnabled)));
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsSettingsGradle<>(),
+                new GroovyIsoVisitor<ExecutionContext>() {
```
jevanlingen, line 71:
```suggestion
        return Preconditions.check(new IsSettingsGradle<>(), new GroovyIsoVisitor<ExecutionContext>() {
```

which gives a nicer indenting (less deep)

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private J.@Nullable MethodInvocation develocityBuildCacheTemplate(String indent, ExecutionContext ctx) {
+        StringBuilder ge = new StringBuilder("\ndevelocity {\n");
+        ge.append(indent).append("buildCache {\n");
+
+        ge.append(indent).append(indent).append("remote(develocity.buildCache) {\n");
+        if (!StringUtils.isBlank(remoteEnabled)) {
+            ge.append(indent).append(indent).append(indent).append("enabled = ").append(remoteEnabled).append("\n");
+        }
```
jevanlingen, line 111:
```suggestion
        String ge =
            "\ndevelocity {\n" +
            "    buildCache {\n " +
            "        develocity.buildCache) {\n";
         if (!StringUtils.isBlank(remoteEnabled)) {
             ge += "            enabled = " + remoteEnabled + "\n";
         }    
```
etc...

is way easier to read, as you can see the template 'nearly' as code.

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,130 @@
...
+        if (!StringUtils.isBlank(remotePushEnabled)) {
+            ge.append(indent).append(indent).append(indent).append("push = ").append(remotePushEnabled).append("\n");
+        }
+        ge.append(indent).append(indent).append("}\n");
+        ge.append(indent).append("}\n");
+        ge.append("}\n");
+
+        G.CompilationUnit cu = GradleParser.builder().build()
+                .parseInputs(singletonList(
+                        Parser.Input.fromString(Paths.get("settings.gradle"), ge.toString())), null, ctx)
```
jevanlingen, line 122:
```suggestion
                .parse(ge.toString())
```
is probably enough.

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,120 @@
...
+import org.openrewrite.*;
+import org.openrewrite.groovy.GroovyIsoVisitor;
+import org.openrewrite.groovy.tree.G;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.java.tree.J;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class EnableDevelocityBuildCache extends Recipe {
```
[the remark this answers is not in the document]
nielsdebruin: The maven equivalent is also called Enable. Personally, I think this makes sense we already have develocity and its buildcache applied as a we are merely enabled in.

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private J.@Nullable MethodInvocation develocityBuildCacheTemplate() {
+        String ge = "\ndevelocity {\n" +
+                    "    buildCache {\n" +
+                    "        remote(develocity.buildCache) {\n";
+
+        if (!StringUtils.isBlank(remoteEnabled)) {
+            ge += "            enabled = " + remoteEnabled + "\n";
+        }
```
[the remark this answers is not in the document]
nielsdebruin: I am not an expert on this, but I see it like this: The option is not required, so if the user does not provide the enabled options, we leave it out, and it falls back to default behavior. If the user wants to set `enabled` based on a condition, e.g., environment variable, he has the ability to do so. However, if you believe that it should just be removed, I have no problem with that.

rewrite-gradle/src/test/java/org/openrewrite/gradle/EnableDevelocityBuildCacheTest.java
nielsdebruin: Fixed

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
```diff
@@ -0,0 +1,120 @@
...
+            description = "Value for `//develocity/buildCache/remote/storeEnabled`.",
+            example = "true",
+            required = false)
+    @Nullable
+    String remotePushEnabled;
+
+    @Override
+    public Validated<Object> validate(ExecutionContext ctx) {
+        return super.validate(ctx)
+                .or(Validated.notBlank("remoteEnabled", remoteEnabled)
```
shanman190, line 60:
```suggestion
                .and(Validated.notBlank("remoteEnabled", remoteEnabled)
```

callout: I think this was meant to be an `and` here to then combine the two internal options using `or`. I don't recall how much if any validation `Recipe` does though, so in practice it may not make a difference.

shanman190 (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/EnableDevelocityBuildCache.java
nielsdebruin: Yes, I think you are right. I fixed it.

timtebeek (approved):
Thanks both! I've applied some simplifications still and with that we're good to merge.

Issue #3997, opened 2024-02-13
Trailing comma in an array end up in white-space


dpozinen:
rewrite bom 2.6.4, wasn't an issue on 2.5.3

Test failures when there is a trailing comma in an annotation. If I remove it, tests pass.

`java.lang.AssertionError: Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser.`

```
package com.playtika.services.infra.rewrite.spring;

import org.springframework.boot.test.context.SpringBootTest;

@SpringBootTest(
        properties = {
                "feign.circuitbreaker.enabled=true",
                "feign.circuitbreaker.group.enabled=true",
                "eureka.client.enabled=false"~~(non-whitespace)~~>,
        <~~},
        webEnvironment = SpringBootTest.WebEnvironment.NONE
)
public abstract class A {
}
```

So the comma at the end of `properties` in a test like this
```
    @Test
    void shouldChangePropertyKeyInAnnotationProperties() {
        rewriteRun(
                spec -> spec.recipe(new ChangePropertyKeyAnnotationAttributeValue(
                        "org.springframework.boot.test.context.SpringBootTest",
                        "properties", "feign.circuitbreaker.enabled",
                        "spring.cloud.openfeign.circuitbreaker.enabled")),
                java(
                        """
                                        package com.playtika.services.infra.rewrite.spring;

                                        import org.springframework.boot.test.context.SpringBootTest;

                                        @SpringBootTest(
                                                properties = {
                                                        "feign.circuitbreaker.enabled=true",
                                                        "feign.circuitbreaker.group.enabled=true",
                                                        "eureka.client.enabled=false", <-- THIS COMMA HERE
                                                },
                                                webEnvironment = SpringBootTest.WebEnvironment.NONE
                                        )
                                        public abstract class A {
                                        }
                                """
...
        ));
    }

```

timtebeek:
Hi! We recently added validation indeed to flush out LST issues in our parsers. It's helpful to improve the parsers, but indeed might cause a test to fail until we circle back to the parser. I'd encountered this case as well and logged a test last week https://github.com/openrewrite/rewrite/commit/78a55a46056808b4b630e781f39bad112abd3d63

We can keep your issue open to see that parser issue resolved; in the meantime you can remove that trailing comma from the test. Hope that helps!

timtebeek:
[...]

Issue #3879, opened 2024-01-02
Parsing issue: Trailing comma on enums is parsed into whitespace


traceyyoshima:
```java
enum Scope {
    None,
    Compile,
}
```

traceyyoshima:
Additional parsing issue related to trailing commas in whitespace:

```java
              public class Test {
                  Integer[] integerArray = new Integer[] {
                      1,
                      2,
                      4,
                  };
              }
```

Issue #4614, opened 2024-10-26
Gradle parser fails with a trailing comma in method call


mccartney:
## What is the smallest, simplest way to reproduce the problem?
```
    @Test
    void trailingComma() {
        rewriteRun(
          buildGradle(
            """
              plugins {
                  id 'java-library'
              }
              dependencies {
                  implementation platform("commons-lang:commons-lang:2.6", )
              }
              """
          )
        );
    }
```

## What is the full stack trace of any errors you encountered?
```
java.lang.AssertionError: Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser. 
plugins {
    id 'java-library'
}
dependencies {
    implementation platform("commons-lang:commons-lang:2.6"~~(non-whitespace)~~>, <~~)
}
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:323)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:132)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:127)
	at org.openrewrite.gradle.GradleParserTest.trailingComma(GradleParserTest.java:199)
```

## Context 

- AFAIK this is a valid Groovy-flavor Gradle file.
- Obviously removing the comma makes the test pass.

shanman190:
I think this one needs to be verified with Gradle natively also before trying to fix the parser. I'd almost expect for Gradle to fail as well which is what I'd want to see verified.

mccartney:
@shanman190 Thanks for looking into this.
I've set up a dedicated Gradle project for reproducing the issue with simple `gradlew init`: https://github.com/mccartney/rewrite-repro-4614-gradle/commit/f5b959e3b386b12ce84df29d5a5456ab5ee31b56
The Gradle build passes with and without this commit.

Also - the example has been extracted from real-life Gradle configuration file.

mccartney:
Also checked Groovy program in [tio.run](https://tio.run/##Sy/Kzy@r/P8/OSexuFjBtSIxtyAnVaGaS0FBobgksSQzWaEsPzNFITcxM08juKQoMy89OlYhsSi9WBOiCAgKgKIlOXka6h6pOTn5CuH5RTkp6jqa1iDpWq7a//8B):
```
class Example {
   static void main(String[] args) {
      println('Hello World',);
   }
}
```

mccartney:
```
    @Test
[...]

PR #4869, opened 2025-01-08
Map trailing commas as `TrailingComma` `Marker`, to fix `J.Erroneous` issues seen

timtebeek:
- Fixes https://github.com/openrewrite/rewrite/issues/3997
- Fixes https://github.com/openrewrite/rewrite/issues/3879

- Related to https://github.com/openrewrite/rewrite/issues/4614


rewrite-test/src/main/java/org/openrewrite/test/TypeValidation.java
```diff
@@ -96,6 +96,13 @@ public class TypeValidation {
     @Builder.Default
     private Function<Object, Boolean> allowMissingType = o -> false;
 
+
+    /**
+     * Controls whether `J.Erroneous` nodes are allowed to be present in the LST.
```
knutwannheden, line 101: Is the documentation maybe inverted here? To be verified.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -1824,6 +1834,21 @@ private <J2 extends J> List<JRightPadded<J2>> convertAll(List<? extends Tree> tr
         return converted;
     }
 
+    private <J2 extends J> List<JRightPadded<J2>> convertAll(List<? extends Tree> trees,
```
Laurens-W, line 1856: Should the method above this not call this new overloaded variant?

knutwannheden: Yes, we could do that to remove some duplication.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -476,13 +476,31 @@ public J visitClass(ClassTree node, Space fmt) {
...
-
+            Tree lastConstant = jcEnums.get(jcEnums.size() - 1);
             List<JRightPadded<J.EnumValue>> enumValues = convertAll(jcEnums, commaDelim, t -> {
-                // this semicolon is required when there are non-value members, but can still
-                // be present when there are not
-                semicolonPresent.set(positionOfNext(";", '}') > 0);
-                return semicolonPresent.get() ? sourceBefore(";", '}') : EMPTY;
+                if (t != lastConstant) {
+                    return whitespace();
+                }
```
knutwannheden, line 483: This if-block is actually not necessary, as we are here inside the `suffix` function which is only ever applied to the last element.

Laurens-W: I agree but on the other hand it also guarantees that we don't accidentally do harm when it is called on an element that is not the last one

PR #4870, opened 2025-01-09
fix: create dependencyResourceLoaders in 2 passes

natedanner:
## What's changed?
In `scanJar` method in `rewrite-core/src/main/java/org/openrewrite/config/Environment.java`, we are now creating the list of dependencyResourceLoaders in 2 passes to ensure all recipes are resolved. 

## What's your motivation?
Recipe descriptors were missing due to transitive recipes failing to be resolved.
- Shown in this issue: https://github.com/moderneinc/customer-requests/issues/765

Are there any undesirable side effects from this change?

@timtebeek @sambsnyd @kmccarp

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -249,12 +249,25 @@ public class Environment {
          */
         @SuppressWarnings("unused")
         public Builder scanJar(Path jar, Collection<Path> dependencies, ClassLoader classLoader) {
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Thanks for diving in! Was it just by luck of order of dependencies that we didn't encounter this before? Great to have it resolved now.

kmccarp:
> Thanks for diving in! Was it just by luck of order of dependencies that we didn't encounter this before? Great to have it resolved now.

This is the first time we've had multiple levels of dependencies. Before, we had just first-order dependencies. Now, we have multiple depths of dependencies.

PR #4882, opened 2025-01-10
Lombok's generated @With method misses type information; fix for Java 17+

jevanlingen:
## What's changed?
Type information is added

## What's your motivation?
- I started improving/adding tests to the LombokTest class, then I saw `withX` method gave a `MethodInvocation type is missing or malformed` error.
- https://github.com/openrewrite/rewrite/pull/4602

### Checklist


Laurens-W:
@knutwannheden do you agree that we can merge these improvements already and look at the Java 8 and 11 issues in a separate PR?

knutwannheden:
AFAICT this represents a workaround. So what I am wondering is whether this is another bug in Lombok or if we initialize the parser and annotation processor incorrectly? I think it would be good if we could answer this question before we start looking at a workaround for Java 1.8 and Java 11. But we could merge this as a first step.

timtebeek (approved):
Great to see the incremental improvements here.

Issue #4765, opened 2024-12-10
`AddOrUpdateAnnotationAttribute` not working with nested annotations


danielshamis:
I am using

- Maven plugin v5.46.1

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
   <groupId>org.openrewrite.maven</groupId>
   <artifactId>rewrite-maven-plugin</artifactId>
   <version>5.46.1</version>
   <configuration>
      <exportDatatables>true</exportDatatables>
      <activeRecipes>
         <recipe>test.AddOrUpdateAnnotationAttributeExample</recipe>
      </activeRecipes>
   </configuration>
</plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
@Test({
        @TestFunction(name = "a"),
        @TestFunction(name = "b"),
})
```

## What did you expect to see?

```java
@Test({
        @TestFunction(name = "a", value = ""),
        @TestFunction(name = "b", value = ""),
})
```

## What did you see instead?
I am trying to add an attribute to each nested Annotation but no changes are being made. Updating an existing attribute does not work either.

timtebeek:
Hi @danielshamis ; Looks like the recipe was missing a call to `super.visitAnnotation(..)`. Would you mind trying [our snapshot version](https://docs.openrewrite.org/reference/snapshot-instructions) in ~15 minutes or so? Sorry that took a while to get to.

danielshamis:
Hi, thank you for your quick response! Unfortunately I can't use external Maven repositories. Furthermore I couldn't find the updated snapshot version in nexus.

To give more context to my issue, I am now using the latest available version of the maven plugin (5.46.2) but the problem persists. I am using declarative yaml format like this to add new attributes:

```yaml
recipeList:
-org.openrewrite.java.AddOrUpdateAnnotationAttribute:
annotationType:TestFunction
attributeName:value
attributeValue:'""'
addOnly:true
appendArray:false
```

`AddOrUpdateAnnotationAttribute` works for single annotations but has no effect on nested annotations. However both `ChangeAnnotationAttributeName` and `RemoveAnnotationAttribute` work fine with nested annotations.

Sorry for the trouble!

timtebeek:
If our snapshot versions aren't an option for you then we'll do another release tomorrow that should pick up this change. Hope that helps!

danielshamis:
I've tried the latest release 5.46.3 with this change, but the issue remains.

[...]

PR #4885, opened 2025-01-10
Add test for inner annotations and fix for 'AddOrUpdateAnnotationAttribute'

danielshamis:
- Resolves issue #4765 
where no changes were being made to inner annotations.


timtebeek (approved):
Thanks @danielshamis ! That'll do it indeed. :)

PR #4887, opened 2025-01-11
Groovy parser fail with single line comment

jonesbusy:
Just demonstrate a bug that we are facing parsing Jenkinsfile with specific multiline comment on one single line

Raise java.lang.StringIndexOutOfBoundsException: String index out of range: 287

No idea how to fix it but it just highlight the issue

### Checklist


jonesbusy:
Thanks @knutwannheden !

PR #4898, opened 2025-01-14
Fix AddOrUpdateAnnotationAttribute unable to handle FieldAccess

### LINKED ISSUE #4897: AddOrUpdateAnnotationAttribute unable to handle @Anno(key = ClassB.staticVar)
I am using

- OpenRewrite 8.42.0
- Mavenplugin 5.47.0

Unit Test, maven plugin

```java
              package com.example;

              public class Const {
                    public static final String PATH = "path";
              }
```

```java
              import jakarta.ws.rs.Path;
              import com.example.Const;
              
              @Path(value = Const.PATH)
              public class Example {
                  private void methodName() { }
              }
```

No errors.

## What did you see instead?

`Failed to parse sources or run recipe`

```
Caused by: java.lang.ClassCastException: class org.openrewrite.java.tree.J$FieldAccess cannot be cast to class org.openrewrite.java.tree.J$Literal (org.openrewrite.java.tree.J$FieldAccess and org.openrewrite.java.tree.J$Literal are in unnamed module of loader 'app')
	at org.openrewrite.java.AddOrUpdateAnnotationAttribute$1.lambda$visitAnnotation$2(AddOrUpdateAnnotationAttribute.java:189)
	at org.openrewrite.internal.ListUtils.map(ListUtils.java:243)
	at org.openrewrite.internal.ListUtils.map(ListUtils.java:265)
	at org.openrewrite.java.AddOrUpdateAnnotationAttribute$1.visitAnnotation(AddOrUpdateAnnotationAttribute.java:123)
	at org.openrewrite.java.AddOrUpdateAnnotationAttribute$1.visitAnnotation(AddOrUpdateAnnotationAttribute.java:85)
	at org.openrewrite.java.tree.J$Annotation.acceptJava(J.java:233)
	at org.openrewrite.java.tree.J.accept(J.java:59)
	at org.openrewrite.TreeVisitor.visit(TreeVisitor.java:250)
	... 44 more
```

## Are you interested in [contributing a fix to OpenRewrite]
Yes, I am trying to make a fix

SiBorea:
## What's changed?
- Fixes #4897 

### Checklist


rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
```diff
@@ -893,6 +892,47 @@ public class A {
         );
     }
 
+    @Test
+    void updateFieldAccessAttribute() {
+        rewriteRun(
```
MBoegers, line 897: Can you try to apply the recipe to change the value from `hello` to `Cost.HI`?

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -46,37 +42,37 @@ public String getDisplayName() {
     @Override
     public String getDescription() {
         return "Some annotations accept arguments. This recipe sets an existing argument to the specified value, " +
-               "or adds the argument if it is not already set.";
+          "or adds the argument if it is not already set.";
```
MBoegers, line 45: I think you have a custom style configured in your IDE. We use the IntelliJ internal formatter with standard configuration. 
Please reformat to keep the change as small as needed.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -174,23 +170,31 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
-                                    return it;
+                                Expression exp = as.getAssignment();
+                                if (exp instanceof J.Literal) {
+                                    J.Literal value = (J.Literal) exp;
+                                    if (newAttributeValue.equals(value.getValueSource()) || Boolean.TRUE.equals(addOnly)) {
+                                        return it;
+                                    }
+                                    return as.withAssignment(value.withValue(newAttributeValue).withValueSource(newAttributeValue));
+                                } else if (exp instanceof J.FieldAccess) {
+                                    as = JavaTemplate.builder("#{} = #{}")
```
MBoegers, line 193: You are omitting here to check if the `newAttributeValue` is already set or the recipes executes in the `addOnly` mode, see line 188. Please check to prevent unnecessary operations.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -174,23 +170,31 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
+                                      .build()
+                                      .apply(new Cursor(getCursor(), as), as.getCoordinates().replace(), var.getSimpleName(), newAttributeValue);
```
MBoegers, line 195: Why do you think you need a new `Cursor` here? `getCursor()` should do the trick

MBoegers:
Hi @SiBorea I also stumbled across this a few hours ago, thanks for the fix. If you have any questions regarding my comments, feel free to ask. :)

MBoegers (changes requested):

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
SiBorea: The `attributeValue` of the recipe is of type String, don't think we can change the value from `hello` to `Const.HI`

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
SiBorea: Using `getCursor()` here will return a `J.Annotation` object, rather than `J.FieldAccess` we need.

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
SiBorea: I tested it a little bit, the result shows that it can only change the value to `"Const.HI"`, not `Const.HI`

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
SiBorea: I omitted `newAttributeValue` since they will never be equal. Forgot to add `addOnly`.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -46,7 +46,7 @@ public String getDisplayName() {
     @Override
     public String getDescription() {
         return "Some annotations accept arguments. This recipe sets an existing argument to the specified value, " +
-               "or adds the argument if it is not already set.";
+                "or adds the argument if it is not already set.";
```
SiBorea, line 49: I restored my IntelliJ to default code style, this should be fine according to the default setting

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
MBoegers: A I see, thats right. Within this recipe the situation is solved by generating an new Annotation with one Assigment and then call `get(0)`, see https://github.com/openrewrite/rewrite/blob/ae8acf40c3faaea8ac16e7b752ddf17603ff7429/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java#L239

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
MBoegers: It _should_ be possible. The `attributeValue`'s value is interpreted as whatever it is. Did you try the FQN `Const.HI`?
If you don't mind, I would have a look myself.

SiBorea: Sure, please go ahead.

MBoegers: I will retest this later in my branch ;)

MBoegers:
Thanks @SiBorea, for your work here <3

PR #4899, opened 2025-01-14
HCL - support for legacy syntax for attribute expressions

greg-at-moderne:
## What's changed?
Amending the HCL parser with support for:
> The HCL native syntax also includes a legacy index operator that exists only for compatibility with the precursor language HIL:
> LegacyIndex = '.' digit+

## What's your motivation?

Fix a real-life parser failure in HCL code for one of the Moderne's customers.


greg-at-moderne:
Looking at the test failure.

greg-at-moderne:
Pursuing #4901 instead.

PR #4901, opened 2025-01-14
HCL - support for legacy syntax for attribute expressions

greg-at-moderne:
## What's changed?
Amending the HCL parser with support for:
> The HCL native syntax also includes a legacy index operator that exists only for compatibility with the precursor language HIL:
> LegacyIndex = '.' digit+

## What's your motivation?

Fix a real-life parser failure in HCL code for one of the Moderne's customers.


rewrite-hcl/src/main/antlr/HCLLexer.g4
```diff
@@ -81,7 +81,7 @@ fragment HexDigit
 // https://github.com/hashicorp/hcl2/blob/master/hcl/hclsyntax/spec.md#numeric-literals
 
 NumericLiteral
-    : [0-9]+ '.' [0-9]* ExponentPart?
+    : [0-9]+ '.' [0-9]+ ExponentPart?
```
greg-at-moderne, line 84: I've checked with Terraform and the no-decimal-dots syntax for numeric literals, e.g.:
```
two = 1. + 1.
```
is not practically supported.

The hcl2 spec agrees:
https://github.com/hashicorp/hcl2/blob/fb75b3253c80b3bc7ca99c4bfa2ad6743841b1af/hcl/hclsyntax/spec.md#L143.

We used to accept that, but we no longer do.

rewrite-hcl/src/main/java/org/openrewrite/hcl/HclVisitor.java
```diff
@@ -77,6 +77,22 @@ public Hcl visitAttributeAccess(Hcl.AttributeAccess attributeAccess, P p) {
         return a;
     }
 
+    public Hcl visitLegacyIndexAttribute(Hcl.LegacyIndexAttributeAccess legacyIndexAttributeAccess, P p) {
+        Hcl.LegacyIndexAttributeAccess li = legacyIndexAttributeAccess;
+        li = li.withPrefix(visitSpace(li.getPrefix(), Space.Location.ATTRIBUTE_ACCESS, p));
```
knutwannheden, line 82: We should define a dedicated new `Space.Location` for this location.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
```diff
@@ -236,6 +237,67 @@ public AttributeAccess withName(HclLeftPadded<Identifier> name) {
...
+        @Getter
+        Space prefix;
+
+        @With
+        @Getter
+        Markers markers;
+
+        @With
+        @Getter
+        Expression base;
```
knutwannheden, line 264: Similarly to `AttributeAccess` we need some way of capturing the whitespace that can occur before the `.`. I suggest we make this property of type `HclRightPadded<Expression>` or use `HclLeftPadded<String>` for the `index` field.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
```diff
@@ -236,6 +237,67 @@ public AttributeAccess withName(HclLeftPadded<Identifier> name) {
...
+
+        @With
+        @Getter
+        String index;
```
knutwannheden, line 268: I assume we are not using `int` here to allow for leading zeros (although I have no idea if the real HCL parsers would support something like `01`). Another alternative would have been to use `Literal` as the type, but that would maybe be a bit overkill.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
```diff
@@ -236,6 +237,67 @@ public AttributeAccess withName(HclLeftPadded<Identifier> name) {
...
+                    p = new Padding(this);
+                    this.padding = new WeakReference<>(p);
+                }
+            }
+            return p;
+        }
+
+        @RequiredArgsConstructor
+        public static class Padding {
+            private final LegacyIndexAttributeAccess t;
```
knutwannheden, line 297: If the type would not have any additional fields we should actually remove it. But we will in fact need it to support that whitespace before the `.` as noted above.

rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/HclParserVisitor.java
```diff
@@ -409,6 +409,17 @@ public Hcl visitIndexAccessExpression(HCLParser.IndexAccessExpressionContext ctx
...
+    @Override
+    public Hcl visitLegacyIndexAttributeExpression(HCLParser.LegacyIndexAttributeExpressionContext ctx) {
+        return convert(ctx, (c, prefix) -> new Hcl.LegacyIndexAttributeAccess(
+                randomId(),
+                Space.format(prefix),
+                Markers.EMPTY,
+                (Expression) visit(c.exprTerm()),
+                c.legacyIndexAttr().NumericLiteral().getText()
+        ));
+    }
```
knutwannheden, line 421: Looks like we are missing some code here to advance the cursor beyond the `.` character. That will be covered by the `HclLeftPadded` or `HclRightPadded` as described.

For HCL it looks like we don't have any verification in place to check that the `Space` doesn't contain any non-whitespace characters. That would otherwise have caught this issue.

rewrite-hcl/src/main/java/org/openrewrite/hcl/HclVisitor.java
greg-at-moderne: 3d99b0477

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
greg-at-moderne: TBH I thought there's some little value in parsing the int to String and then printing it back.
But maybe declaring it as `int` explicitly conveys a clearer message what it is.

=> 85a1ad8eb

knutwannheden: The alternative would be Literal, which both has a value and a string representation.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
greg-at-moderne: 81094c858

rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/HclParserVisitor.java
greg-at-moderne: 81094c858

knutwannheden (approved):
Nice work!

PR #4906, opened 2025-01-15
Performance issue parsing a larger TOML file

timtebeek:


timtebeek:
Verified locally this now runs in the millisecond range; with that there's no need to merge this PR anymore.

PR #4907, opened 2025-01-15
TOML model cleanups

knutwannheden:
Added a few missing `@Nullable` annotations and added accessors for `Toml.Table#name`.

### DIFF
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/tree/Toml.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/tree/Toml.java
@@ -128,7 +128,11 @@ public interface Toml extends Tree {
         String charsetName;
 
         boolean charsetBomMarked;
+
+        @Nullable
         Checksum checksum;
+
+        @Nullable
         FileAttributes fileAttributes;
 
         @Override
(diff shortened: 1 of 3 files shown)


rewrite-toml/src/main/java/org/openrewrite/toml/tree/Toml.java
```diff
@@ -309,9 +314,17 @@ class Table implements TomlValue {
         @With
         Markers markers;
 
-        @With
+        @Nullable
```
shanman190, line 317: The table name is a required element otherwise it would result in an invalid Toml document.

knutwannheden: In the parser I think we pass in `null` for inline tables.

knutwannheden: https://github.com/openrewrite/rewrite/blob/02411bd6e386f793b8c8af2c49f2154b528f6144/rewrite-toml/src/main/java/org/openrewrite/toml/internal/TomlParserVisitor.java#L370-L376

shanman190: Ahh, right, right. I forgot about that piece.

shanman190 (approved):

timtebeek (approved):

PR #4910, opened 2025-01-16
Correct TOML table parsing

knutwannheden:
Tables in TOML are not nested. I.e. the following should result in an AST with two top-level tables:

```toml
[table-1]
key1 = "some string"
key2 = 123

[table-2]
key1 = "another string"
key2 = 456
```

This has now been corrected in the parser.


timtebeek (approved):

rewrite-toml/src/main/antlr/TomlParser.g4
```diff
@@ -124,7 +123,7 @@ table
     ;
 
 standardTable
-    : L_BRACKET key R_BRACKET (commentOrNl* expression)*
+    : L_BRACKET key R_BRACKET (commentOrNl* keyValue)*
```
shanman190, line 126: callout: For the TOML language this isn't correct because TOML allows tables to also contain other nested tables and arrays of tables as well. So as a result of this change, we're temporarily artificially limiting the TOML files that we can parse.

knutwannheden: As discussed I think the nesting here is only "conceptual" and that the language structure and therefore LST model is indeed flat. We could of course provide a JSON mapping or similar if client code needs access to this conceptual model.

PR #4855, opened 2025-01-07
Partial support Lombok for java 8 (mostly done)

jevanlingen:
## What's changed?
- Java 8 parser can parse Lombok code as well.
- Other improvements
    - Lombok parser code optimised for all java parsers
    - Improved Lombok tests
    - Renamed all JavacAnnotationHandler without an action to `<LombokAnnotation>NoOpHandler` (half of the JavacAnnotationHandler classes that had this behavior were already named that way)

## What's does not work yet?
- `@AllArgsConstructorHandler`, `@NoArgsConstructorHandler` and `@NoArgsConstructorHandler` with the _onConstructor__ param.
- `@Jacksonized`

_Further research should be done in a follow up ticket. The rest of this code is good as it is and is mergeable. Especially since the [onX](https://projectlombok.org/features/experimental/onX) feature is still very much in the experimental status and @Jacksonized is not used that often._

## What's your motivation?
- https://github.com/openrewrite/rewrite/pull/4602
- https://github.com/openrewrite/rewrite/issues/4781

## Any additional context
I tried to bring the Java 8 and Java 17 code as close to each other as possible. That's why I added the `parseInputsToCompilerAst` method and removed the custom `Todo` class (just like the java 11, 17 and 21 parser).

Furthermore I squeezed the lombok code in all parsers as much as possible. By having just one `isLombokGenerated` method which works for whatever _com.sun.source.tree.Tree_ as input, the code is easier to follow.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #4914, opened 2025-01-16
Fix issue parsing string with Java8ReloadableParser

rlsanders4:
## What's changed?
Revert the `Java8ReloadableParser` to use `JavaCompiler.parse()` instead of `JavaCompiler.parseFiles()` to support parsing strings as input. This issue does not occur with other versions of the Java parser.

## What's your motivation?
Recent changes to `Java8ReloadableParser`  caused the following error in our custom recipes:
- #4855
```
SEVERE: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at com.sun.tools.javac.util.List.get(List.java:476)
	at org.openrewrite.java.ReloadableJava8Parser.parseInputsToCompilerAst(ReloadableJava8Parser.java:260)
	at org.openrewrite.java.ReloadableJava8Parser.parseInputs(ReloadableJava8Parser.java:211)
	at org.openrewrite.java.Java8Parser.parseInputs(Java8Parser.java:44)
	at org.openrewrite.java.JavaParser.parse(JavaParser.java:285)
	at org.openrewrite.java.JavaParser.parse(JavaParser.java:300)
```

This occurs when we attempt to parse a file from a string input:
```
                        Path targetFile = Paths.get("/path/to/file.java")

                        Stream<SourceFile> sourceJavaFile = JavaParser.fromJavaVersion()
                                .build()
                                .parse(new String(Files.readAllBytes(targetFile)));
```

@timtebeek 

## Any additional context
I would like to add a unit test for this, but I have not been able to figure out how to make calls to `JavaParser.fromJavaVersion().build().parse()` using Java 8 within the unit testing framework. Any assistance with this would be appreciated. I have confirmed this change fixes our recipes through end-to-end testing.

### Checklist


timtebeek (approved):
Looks ok from my end when compared to the [line removed previously](https://github.com/openrewrite/rewrite/pull/4855/files#diff-1a22c111a5b6210b56c84b9ca3639dc2592aafe80b5352b0b3cdc0666af36fd7L153). I wouldn't necessarily know how to best test this, but I would be fine merging this without another test.
/cc @knutwannheden in case I missed anything

knutwannheden:
I added a test case to `rewrite-java-tck` which I think should have been able to reproduce the issue as this will be executed for `rewrite-java-8` which uses a Java 1.8 VM. I was not able to reproduce the issue, but I don't see any harm with this PR either. Possibly we can get the test case working at some point.

PR #4916, opened 2025-01-17
Autodetect formatting style of JSON code

### LINKED ISSUE #4246: Json formatting/style doesn't seem to be respected when adding values/members

greg-at-moderne:
## What's changed?

Adding:
- support for autodetecting JSON formatting style
- and `AddKeyValue` recipe using it

The PR comes as patchwork of:
- the recipe implementation done in #4341 by @dpozinen 
- the style autodetection basically copied from respective classes for XML and YAML

## What's your motivation?

- Fixes #4246

- The style autodetection is a basic implementation focusing on just indents and tabs. There's no support for other qualities/features of code.
- No support for full code reformatting.


rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java
```diff
@@ -0,0 +1,142 @@
...
+    String keyPath;
+
+    @Option(displayName = "Key",
+            description = "The key to create.",
+            example = "myKey")
+    String key;
+
+    @Option(displayName = "Value",
+            description = "The value to add to the array at the specified key. Can be of any type." +
+                          " String values should be quoted to be inserted as Strings.",
```
knutwannheden, line 53: The description here appears to imply that a value will get added to an array, if the value is an array. Looking at the implementation that doesn't appear to be the case. Can we check how we handle this in YAML? Should this be supported or is that better handled by another recipe? In any case I think we ought to clarify that.

rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            private JsonRightPadded<JsonKey> rightPaddedKey() {
+                return new JsonRightPadded<>(
+                        new Json.Literal(randomId(), Space.EMPTY, Markers.EMPTY, "\"" + key + "\"", key),
+                        Space.EMPTY, Markers.EMPTY
+                );
+            }
+
+            private String unQuote(String value) {
```
knutwannheden, line 114: The value is only expected to be a string when it is quoted. Otherwise a Number or Boolean object would actually be expected. Or even the final result a `Json.JsonObject` or `Json.Array` instead of a `Json.Literal`. Possibly we could just use the `JsonParser`?

rewrite-json/src/main/java/org/openrewrite/json/format/Indents.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openrewrite.json.style.TabsAndIndentsStyle;
+import org.openrewrite.json.tree.Json;
+import org.openrewrite.style.NamedStyles;
+
+import static java.util.Collections.singletonList;
+
+public class Indents extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "YAML indent";
```
knutwannheden, line 32:
```suggestion
        return "JSON indent";
```

rewrite-json/src/main/java/org/openrewrite/json/format/Indents.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Format tabs and indents in YAML.";
```
knutwannheden, line 37:
```suggestion
        return "Format tabs and indents in JSON.";
```

rewrite-json/src/main/java/org/openrewrite/json/style/Autodetect.java
```diff
@@ -0,0 +1,237 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.json.style;
```
knutwannheden, line 16: As this package didn't exist until now, please add a `package-info.java` file as we have in all other packages.

rewrite-json/src/main/java/org/openrewrite/json/format/AutoFormatVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.json.format;
```
knutwannheden, line 16: As this package didn't exist until now, please add a `package-info.java` file as we have in all other packages.

rewrite-json/src/main/java/org/openrewrite/json/style/Autodetect.java
```diff
@@ -0,0 +1,237 @@
...
+
+            return new TabsAndIndentsStyle(
+                    useTabs,
+                    useTabs ? indent : 1,
+                    useTabs ? 1 : indent
+            );
+        }
+    }
+
+    private static class FindLineFormatJavaVisitor extends JsonVisitor<GeneralFormatStatistics> {
```
knutwannheden, line 126:
```suggestion
    private static class FindLineFormatJsonVisitor extends JsonVisitor<GeneralFormatStatistics> {
```

greg-at-moderne: 1e276396f instead.

rewrite-json/src/main/java/org/openrewrite/json/style/Autodetect.java
greg-at-moderne: 160d31050

rewrite-json/src/main/java/org/openrewrite/json/format/AutoFormatVisitor.java
greg-at-moderne: 160d31050

rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java
greg-at-moderne: 199f5c546

rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java
greg-at-moderne: eda9f61ae to clarify the current status.

greg-at-moderne: > Can we check how we handle this in YAML? Should this be supported or is that better handled by another recipe? In any case I think we ought to clarify that.

- In YAML we have a dedicated recipe to append to an array (sequence): https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/AppendToSequence.java
- Other than `MergeYaml` there's no direct counterpart of the `AddKeyValue` in YAML. There's a bunch of `Change*` recipes
- There's `CopyValue` recipe, but that doesn't seem to work with sequences/arrays either.

=> I suggest we just keep the `AddKeyValue` recipe as is currently. And then optionally at some point we might decide to implement `AppendToArray` recipe or some more generic array manipulation (`AppendOrUpdateArray`, etc.).

Issue #509, opened 2021-05-07
Accept YAML mapping values 


jkschneider:
Permit Spring-style variables in YAML (not actually valid YAML by spec):

```yaml
date: @build.timestamp@
version: @project.version@
```

A similar parsing issue has been reported to Spring in various places like https://github.com/spring-projects/spring-boot/issues/8438.

jkschneider:
The technique we employ is to pre-process the YAML string, replacing any variables we find with UUIDs. As the AST is built, we take scalar values observed by Snakeyaml and swap them out for the variable mapped to the UUID.

PR #4917, opened 2025-01-17
YAML Parser not to take @variables@ in the middle of scalar

greg-at-moderne:
## What's changed?
Changing the behavior of YAML parser with regards to `@variables@`. The special handling of these is now reduced to only the situations when there's space before `@`.

## What's your motivation?

One of the Moderne customers has reported an edge case of YAML parser corrupting values containing `@` in some cases.

## Context / Thought process on the solution

- These `@variables` are **not** part of the YAML standard
- It seems our support for them has been added in 2021 to address some specific Spring Boot issue - see #509
- As part of the spec, the `@` sign is not allowed to start a scalar value
- Thus I think it's a fine middle ground to continue to support these non-standard variables in a reduced scope, but fix the problem of `@` being used within the values.

- An alternative would be to drop the support for `@variables@` as non-complaint. But that would be hardly helpful.
- Another alternative would be to fine-tune the matching regexp to try to tell apart the variables use-case from just using `@` in the values. But that's hard to get it right and not cause other issues.

The down-side of the current solution is that the following (invalid) YAML:
```
date:@build.timestamp@
```
is **not** handled as a variable anymore (while it used to be). 

```
date: @build.timestamp@
```
continues to be handled the way it used to be.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -52,7 +52,7 @@ import static java.util.Collections.singletonList;
 import static org.openrewrite.Tree.randomId;
 
 public class YamlParser implements org.openrewrite.Parser {
-    private static final Pattern VARIABLE_PATTERN = Pattern.compile(":\\s*(@[^\n\r@]+@)");
+    private static final Pattern VARIABLE_PATTERN = Pattern.compile(":\\s+(@[^\n\r@]+@)");
 
     @Override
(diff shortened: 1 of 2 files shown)


knutwannheden:
To me this seems like an acceptable solution. The `@variable@` syntax is a very simple templating mechanism and since the `@` is not allowed by the YAML spec as the leading character of a scalar, some interoperability problems with other tools are to be expected when using this. Moreover, I think there is a very simple workaround for this when the replaced variable is a string (doesn't work when it is a number or boolean): Just put it in quotes. Also, I think files using this can be expected to be hand-written and will thus typically also have a leading whitespace.

knutwannheden (approved):

PR #4919, opened 2025-01-17
keep old constructor for backward compatibility

OlegDokuka:
followup to https://github.com/openrewrite/rewrite/pull/4661 to preserve backward compatibility


Laurens-W:
@knutwannheden what do you think? Should we introduce this overloading?

timtebeek:
As discussed on Slack
> I think I would not merge this. Historically we have not added backwards compatibility constructors like this and having two constructors annotated with `@JsonCreator` makes me a bit nervous, too.

PR #4923, opened 2025-01-20
Support Groovy methods declaring generic type parameters

knutwannheden:
Also, using `@ExpectedToFail` I found a bunch of tests which appear to have been fixed. I enabled those and replaced `@Disabled` with `@ExpectedToFail` on others.

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/ClassDeclarationTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/ClassDeclarationTest.java
@@ -353,7 +353,7 @@ class ClassDeclarationTest implements RewriteTest {
 
     @Issue("https://github.com/openrewrite/rewrite/pull/4346")
     @Test
-    @Disabled("Known issue; still need to explore a fix")
+    @ExpectedToFail("Known issue; still need to explore a fix")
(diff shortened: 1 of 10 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/ClassDeclarationTest.java
```diff
@@ -353,7 +353,7 @@ class A {
 
     @Issue("https://github.com/openrewrite/rewrite/pull/4346")
     @Test
-    @Disabled("Known issue; still need to explore a fix")
+    @ExpectedToFail("Known issue; still need to explore a fix")
```
shanman190, line 356: @knutwannheden, I think we could also drop this test out of https://github.com/openrewrite/rewrite/pull/4922 given you've added a new one there that mimics this exactly.

knutwannheden: Good point! I completely missed that. Which is why I like `@ExpectedToFail`.

knutwannheden: @timtebeek Thanks for fixing!

PR #4924, opened 2025-01-20
Fixing YAML handling of multiline literals in some cases

greg-at-moderne:
## What's changed?

Fixing the handling of an edge-case of a YAML with:
- multiline literals using the `|` syntax
- and quoted identifiers.

## What's your motivation?

In some scenarios the YAML parser tends to swallow the first letter of the literal:
```diff
 - "one": |
     two
-  "three": "four"
+  "hree": "four"
] 
```
which is clearly a bug.


knutwannheden (approved):
Nice find!

PR #4925, opened 2025-01-20
Support JEP-440: Pattern matching for Records

Laurens-W:
## What's changed?

Add full blown support for record pattern matching

## Any additional context

https://openjdk.org/jeps/440

### Checklist

### DIFF
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/SwitchPatternMatchingTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/SwitchPatternMatchingTest.java
@@ -152,4 +152,27 @@ class SwitchPatternMatchingTest implements RewriteTest {
           ));
     }
 
+    @Test
+    void shouldParseRecordPatternMatchingInSwitch() {
+        rewriteRun(
+          java(
+            //language=java
+            """
+              class Test {
+                  public interface Printable {}
+                  record A(String A) implements Printable {}
+                  record B(Integer B) implements Printable {}
+              
+                  void integerTester(Printable prt) {
+                      switch (prt) {
+                          case A(String a) -> System.out.println(a);
+                          case B(Integer b) -> System.out.println(b);
+                          default -> throw new IllegalStateException("Unexpected value: " + prt);
+                      }
+                  }
+              }
+              """
+          ));
+    }
+
 }


rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3048,6 +3048,81 @@ public InstanceOf withExpr(JRightPadded<Expression> expression) {
...
+
+            @Deprecated
+            public JContainer<Pattern> getNested() {
+                return t.nested;
+            }
+
+            @Deprecated
+            public DeconstructionPattern withNested(JContainer<Pattern> nested) {
+                return t.nested == nested ? t : new DeconstructionPattern(t.id, t.prefix, t.markers, t.deconstructor, nested, t.type);
+            }
```
timtebeek, line 3121: What's the reason to have these methods and have them start out deprecated?

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -785,10 +785,23 @@ public J visitInstanceOf(InstanceOfTree node, Space fmt) {
...
+
     private @Nullable J getNodePattern(@Nullable PatternTree pattern, JavaType type) {
         if (pattern instanceof JCBindingPattern b) {
             return new J.Identifier(randomId(), sourceBefore(b.getVariable().getName().toString()), Markers.EMPTY, emptyList(), b.getVariable().getName().toString(),
                     type, typeMapping.variableType(b.var.sym));
+        } else if (pattern instanceof DeconstructionPatternTree r) {
+            return visitDeconstructionPattern(r, whitespace());
         } else {
             if (pattern == null) {
                 return null;
```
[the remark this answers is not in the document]
Laurens-W: There's currently 3 types of patterns, binding which is mapped to variable, deconstruction which currently only implements records and `any` which is a preview feature. So in theory we no longer need it for Java 21, but it could help us with Java 24/25 print idempotency

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3048,6 +3048,79 @@ public InstanceOf withExpr(JRightPadded<Expression> expression) {
...
+        @Getter
+        Markers markers;
+
+        @With
+        @Getter
+        JRightPadded<Expression> deconstructor;
+
+        @With
+        @Getter
+        JContainer<Pattern> nested;
```
knutwannheden, line 3080: Both of these properties need custom getters and withers, which expose the unwrapped types. I.e. `Expression` for `deconstructor` and `List<Pattern>` for `nested.

rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java
```diff
@@ -745,6 +745,17 @@ public J visitInstanceOf(J.InstanceOf instanceOf, P p) {
         return i;
     }
 
+    public J visitDeconstructionPattern(J.DeconstructionPattern deconstructionPattern, P p) {
+        J.DeconstructionPattern d = deconstructionPattern;
+        d = d.withPrefix(visitSpace(d.getPrefix(), Space.Location.DECONSTRUCTOR_PREFIX, p));
+        d = d.withMarkers(visitMarkers(d.getMarkers(), p));
+        d = d.withDeconstructor(visitRightPadded(d.getDeconstructor(), JRightPadded.Location.DECONSTRUCTOR, p));
+        d = d.withNested(visitContainer(d.getNested(), JContainer.Location.DECONSTRUCTOR, p));
```
knutwannheden, line 753: These should work with the `getPadding()` indirection.

rewrite-java/src/main/java/org/openrewrite/java/JavaIsoVisitor.java
```diff
@@ -194,6 +194,11 @@ public J.InstanceOf visitInstanceOf(J.InstanceOf instanceOf, P p) {
         return (J.InstanceOf) super.visitInstanceOf(instanceOf, p);
     }
 
+    @Override
+    public J.DeconstructionPattern visitDeconstructionPattern(J.DeconstructionPattern deconstructionPattern, P p) {
+        return (J.DeconstructionPattern) super.visitDeconstructionPattern(deconstructionPattern, p);
+    }
+
```
knutwannheden, line 201: We also need to add an override to the `SemanticallyEqualVisitor`.

knutwannheden (commented):
I want to check a few more things, but here is a first set of comments.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
knutwannheden: Actually,. I think there is no need for `deconstructor` to use `JRightPadded`. The space after the type name will be part of the `JContainer`s `before`.

knutwannheden: Hmm. I suspect this only works correctly due to Java's type erasure. In a pattern like `Point(int x, int y)` the `int x` and `int y` part are mapped to `J.VariableDeclarations` which doesn't implement the `Pattern` interface.

Laurens-W: We run into the same issue here as with switch pattern matching, should it be a `JContainer<J>` then?

knutwannheden: Unsure. Having J.VariableDeclarations implement Pattern might be OK. It might cause issues with the inherited getCoordinates() method.

Laurens-W: Yeah it does cause that issue, so I'm leaning towards changing it to J

Laurens-W: Unless we want to introduce a `J.BindingPattern` which implements pattern and merely holds a `VariableDeclaration`

rewrite-java/src/main/java/org/openrewrite/java/tree/Pattern.java
```diff
@@ -0,0 +1,20 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.tree;
+
+public interface Pattern extends J {
```
knutwannheden, line 18: Since the only place we currently refer to this type is the `implements` clause of `DeconstructorPattern` I am thinking we might as well remove it for now and add it once we need it (e.g. as a parameter or return type somewhere).

knutwannheden (commented):
I added a few small comments. After merging this it would further be cool if we could also look into updating the formatter to support adding spaces before or after parentheses as configured in the style.

knutwannheden (approved):

PR #4927, opened 2025-01-20
Refactoring, introducing getStyleOrDefault and getStyleOrFromAutodetect

greg-at-moderne:
## What's changed?

Fairly minor refactoring around `Styles` for multiple languages.
Extracting `getStyleOrDefault` or `getStyleOrFromAutodetect` convenience methods.

## What's your motivation?

To DRY the code and align it more against multiple implementations for various languages.

Yes, please review if you consider it an improvement in code readability. I do, but it might be a matter of taste.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite;
 
-import org.jetbrains.annotations.NotNull;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.style.NamedStyles;
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite;
 
+import org.jetbrains.annotations.NotNull;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.style.NamedStyles;
@@ -25,6 +26,9 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+import static java.util.Collections.singletonList;
 
 public interface SourceFile extends Tree {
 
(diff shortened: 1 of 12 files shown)


rewrite-core/src/main/java/org/openrewrite/SourceFile.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite;
 
+import org.jetbrains.annotations.NotNull;
```
timtebeek, line 18:
```suggestion
```

rewrite-xml/src/main/java/org/openrewrite/xml/format/AutoFormatVisitor.java
```diff
@@ -56,16 +56,10 @@ public Xml visit(@Nullable Tree tree, P p, Cursor cursor) {
...
-                    return NamedStyles.merge(TabsAndIndentsStyle.class, singletonList(detector.build()));
-                });
-        assert tabsStyle != null;
-
-        t = new NormalizeTabsOrSpacesVisitor<>(tabsStyle, stopAfter)
-                .visit(t, p, cursor.fork());
+        TabsAndIndentsStyle tabsStyle =
+                doc.getStyleOrFromAutodetect(TabsAndIndentsStyle.class, () -> Autodetect.detector().sample(doc).build());
+        t = new NormalizeTabsOrSpacesVisitor<>(tabsStyle,
+                stopAfter).visit(t, p, cursor.fork());
```
timtebeek, line 62:
```suggestion
        t = new NormalizeTabsOrSpacesVisitor<>(tabsStyle, stopAfter)
                .visit(t, p, cursor.fork());
```

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/AutoFormatVisitor.java
```diff
@@ -45,12 +45,14 @@ public Yaml preVisit(Yaml tree, P p) {
 
         y = new MinimumViableSpacingVisitor<>(stopAfter).visitNonNull(y, p, cursor.fork());
 
-        y = new IndentsVisitor<>(Optional.ofNullable(docs.getStyle(IndentsStyle.class))
-                .orElse(Autodetect.tabsAndIndents(docs, YamlDefaultStyles.indents())), stopAfter)
+        y = new IndentsVisitor<>(
+                    docs.getStyle(IndentsStyle.class, Autodetect.tabsAndIndents(docs, YamlDefaultStyles.indents())),
```
timtebeek, line 49: Should we not use a supplier approach here and further down? Note that I'm only going off the diff in GitHub, so you might have a good reason not to that's out of view here.

timtebeek (approved):

rewrite-yaml/src/main/java/org/openrewrite/yaml/format/AutoFormatVisitor.java
greg-at-moderne: We cannot use the `Supplier` (at least with the current suggested content of the super class of `SourceFile`). Because in this line `Autodetect.tabsAndIndents` is not a `NamedStyles`, typically the type of `Autodetect` classes. 

The YAML `Autodetect` is implemented/structured a bit differently than (some of) the others. It has `Autodetect` and two static methods: `tabsAndIndents`, `generalFormat`.
While a few of the other `Autodetect` classes have a `Detector` inner class, whose `build()` method returns an object of `Autodetect extends NamedStyles` type.

Probably that's something that we could unify too for the sake of consistency.

rewrite-core/src/main/java/org/openrewrite/SourceFile.java
```diff
@@ -79,6 +82,21 @@ default <S extends Style> S getStyle(Class<S> style, S defaultStyle) {
         return s == null ? defaultStyle : s;
     }
 
+    default <S extends Style> S getStyleOrDefault(Class<S> style, Supplier<S> defaultStyle) {
```
knutwannheden, line 85: I JavaSourceFile we have overridden service() to support an AutoFormatService. Possibly we should pull that up here. That would then not require adding yet more methods to this API.

PR #4932, opened 2025-01-21
TypeTable as a substitute for packing whole jars into `META-INF/rewrite/classpath`

jkschneider:
## What's changed?

Introducing a new mechanism for recipes to have resource level dependencies on JARs when they need them for `JavaTemplate` classpaths. Type tables provide an inventory of essentially the ABI of whole classes and JARs, from which we can reconstitute enough bytecodes to be used in a later compilation step (such as a `JavaTemplate` use).

## What's your motivation?

In `rewrite-spring` we've been gradually increasing the size of the recipe JAR over time by including more and more versions . Additionally, we see security scanners sometimes blocking our recipe JARs because they have inlined "vulnerable" JARs into the META-INF directory, even though these JARs would never be executed (the scanners can't tell the difference).

`JavaParser#dependenciesFromResources` will, for a time, still look for whole JARs in the `/META-INF/rewrite/classpath` folder in addition to these type tables.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
@@ -1,0 +1,412 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.internal.parser;
+
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.objectweb.asm.*;
+import org.objectweb.asm.util.CheckClassAdapter;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Incubating;
+import org.openrewrite.java.JavaParserExecutionContextView;
+import org.xerial.snappy.SnappyInputStream;
+import org.xerial.snappy.SnappyOutputStream;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.regex.Pattern;
+
+import static org.objectweb.asm.ClassReader.SKIP_CODE;
+import static org.objectweb.asm.Opcodes.V1_8;
+import static org.openrewrite.java.internal.parser.JavaParserCaller.findCaller;
+
+/**
+ * Type tables are written as a TSV file with the following columns:
+ * <ul>
+ *     <li>groupId</li>
+ *     <li>artifactId</li>
+ *     <li>version</li>
+ *     <li>classAccess</li>
+ *     <li>className</li>
+ *     <li>classSignature</li>
+ *     <li>classSuperclassSignature</li>
+ *     <li>classSuperinterfaceSignatures[]</li>
+ *     <li>access</li>
+ *     <li>memberName</li>
+ *     <li>descriptor</li>
+ *     <li>signature</li>
+ *     <li>parameterNames</li>
+ *     <li>exceptions[]</li>
+ * </ul>
+ * <p>
+ * Descriptor and signature are in <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3">JVMS 4.3</a> format.
+ * Because these type tables could get fairly large, the format is optimized for fast record access. TSV was chosen over CSV for its
+ * simplicity and for the reasons cited <a href="https://github.com/eBay/tsv-utils/blob/master/docs/comparing-tsv-and-csv.md">here</a>.
+ * <p>
+ * There is of course a lot of duplication in the class and GAV columns, but compression cuts down on
+ * the disk impact of that and the value is an overall single table representation.
+ */
+@Incubating(since = "8.44.0")
+@Value
+public class TypeTable implements JavaParserClasspathLoader {
+    /**
+     * Verifies that the bytecodes written out for the types represented in a type table
+     * will not be invalid and therefore rejected by the JVM verifier when used in a compilation
+     * step.
+     */
+    public static final String VERIFY_CLASS_WRITING = "org.openrewrite.java.TypeTableClassWritingVerification";
+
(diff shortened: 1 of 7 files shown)


rewrite-java/src/test/java/org/openrewrite/java/internal/parser/TypeTableTest.java
```diff
@@ -0,0 +1,127 @@
...
+                jarsSize += writeJar(classpath, writer);
+            }
+
+            System.out.println("Total size of type table " + humanReadableByteCount(Files.size(tsv)));
+            System.out.println("Total size of jars " + humanReadableByteCount(jarsSize));
+        }
+    }
+
+    @Test
+    void writeReadMicrometer() throws IOException {
```
jkschneider, line 71: To best understand how `TypeTable` is used, start your review here.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -0,0 +1,412 @@
...
+    public static final String DEFAULT_RESOURCE_PATH = "META-INF/rewrite/classpath.tsv.snappy";
+
+    private static final Map<GroupArtifactVersion, Path> classesDirByArtifact = new LinkedHashMap<>();
+
+    public TypeTable(ExecutionContext ctx, Collection<String> artifactNames) {
+        this(ctx, findCaller().getClassLoader().getResourceAsStream(DEFAULT_RESOURCE_PATH), artifactNames);
+    }
+
+    public TypeTable(ExecutionContext ctx, InputStream is, Collection<String> artifactNames) {
+        try (InputStream snappy = new SnappyInputStream(is)) {
```
knutwannheden, line 86: I would recommend that we use LZ4, Zstd, or even just standard DEFLATE. I was hoping we could get rid of Snappy. Not because it is bad (although IMHO LZ4 and Zstd are typically better), but because it is one more dependency with native libs in it.

I understand that we at the moment already have Snappy on the classpath and not LZ4 nor Zstd. But I was working towards getting rid of Snappy in #4641. So how about using DEFLATE from the JDK or adding either of the other two to the classpath?

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -0,0 +1,412 @@
...
+ *     <li>signature</li>
+ *     <li>parameterNames</li>
+ *     <li>exceptions[]</li>
+ * </ul>
+ * <p>
+ * Descriptor and signature are in <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3">JVMS 4.3</a> format.
+ * Because these type tables could get fairly large, the format is optimized for fast record access. TSV was chosen over CSV for its
+ * simplicity and for the reasons cited <a href="https://github.com/eBay/tsv-utils/blob/master/docs/comparing-tsv-and-csv.md">here</a>.
+ * <p>
+ * There is of course a lot of duplication in the class and GAV columns, but compression cuts down on
```
knutwannheden, line 64: I suspect we will always read the whole thing rather than random access, so we could also just omit all repeated values. I.e. only write the GAV columns for every first entry of that artifact.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -0,0 +1,412 @@
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static Collection<String> artifactsNotYetWritten(Collection<String> artifactNames) {
+        Collection<String> notWritten = new ArrayList<>(artifactNames);
+        for (String artifactName : artifactNames) {
+            for (GroupArtifactVersion groupArtifactVersion : classesDirByArtifact.keySet()) {
+                if (Pattern.compile(artifactName + ".*")
```
knutwannheden, line 101: Pattern here seems like a good candidate to pull up outside the inner loop.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -0,0 +1,412 @@
...
+                    in.lines().skip(1).map(line -> line.split("\t", -1)).forEach(fields -> {
+                        GroupArtifactVersion rowGav = new GroupArtifactVersion(fields[0], fields[1], fields[2]);
+                        if (!rowGav.equals(gav)) {
+                            writeClassesDir();
+                        }
+
+                        String artifactVersion = fields[1] + "-" + fields[2];
+
+                        for (String artifactName : artifactNames) {
+                            if (Pattern.compile(artifactName + ".*").matcher(artifactVersion).matches()) {
```
knutwannheden, line 132: We could consider creating a map of precompiled patterns, so that we don't do that for every line in the file.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
jkschneider: I don't think this is worthwhile since there are probably low tens number of distinct iterations here.

knutwannheden: We can always later do some profiling and JMH benchmarking.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
jkschneider: Fixed in a23866b85faf9209880c1ebf17425b4c1feeda51.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
jkschneider: Done in ac31424840560e84d3c461c0cc9fd141d583f548.

jkschneider:
@shanman190 Interesting to learn about the `ct.sym` file. If I understand the content correctly, I think it actually may be less scalable for the solution space.

I see the type table eventually transitioning for a simple deflate TSV to eventually a columnar file format like parquet, with the eventual goal of building a type table for all of maven central, etc. For this same reason, I don't worry too much about column duplication because columnar file formats often _expect_ high levels of duplication in a column. 

I think our potential for large scale analytics on this type information is greater if we can pack this data into a tabular structure.

jkschneider:
Absolutely! And thanks for making me aware of `ct.sym`. I didn't know :)

jkschneider:
Just for curiosity, I "type tabled" the first 500 JARs in my m2 local cache:

shanman190:
Thats not too bad! The JDK signature table is 7.2 MB when Gunnar wrote that blog post (JDK 16 at the time).

I know you mentioned TypeTable was a kind of odd name, but the only other thing that I could think of was SignatureTable or SymbolTable (this one's not semantically accurate though).

jkschneider:
Yeah I have a hard time getting the JVMS 4.3 definition of "signature" out of my mind now. Signature is only non-null when there are generic types involved. "Descriptor" always exists but doesn't contain the generic info. Doesn't seem like there's an overarching term that encompasses both.

timtebeek:
Discovered a failure when attempting to use type tables in rewrite-hibernate; figured log an issue first
- https://github.com/openrewrite/rewrite/issues/4993

PR #4934, opened 2025-01-22
Amending YAML parser to add support for tags

greg-at-moderne:
## What's changed?
Adding support for [tags](https://yaml.org/spec/1.2.2/#24-tags) in YAML, e.g.
```
AttributeDefinitions: !Dynamo Title
```
`!Dynamo` is a tag here.

## What's your motivation?

Observed in real-life YAMLs that they sometimes use tags.
Currently they have been accidentally supported in some cases. E.g. for scalars we used to consider `!Dynamo Title` to be the value.

The LST model for YAML has been amended to include two new fields called `tag` in `Scalar` and `Sequence`. Which is not fully backward compatible. E.g. if someone has a recipe which does `new Yaml.Scalar(...)`, the code will no longer compile for 1 extra argument missing.
Not sure how do we provide coverage for these scenarios.


greg-at-moderne:
> tag's "name"

Renamed the field to `text` in 79f22f7cd

greg-at-moderne:
>  I wonder if these details should be part of the LST model rather than being implicitly encoded as the tag's "name".

Added in adf3fbb3b3e5fec468f792c8f4b154c760b1ba03

rewrite-yaml/src/main/java/org/openrewrite/yaml/CountLinesVisitor.java
```diff
@@ -85,6 +85,15 @@ public Yaml visitAnchor(Yaml.Anchor anchor, AtomicInteger count) {
         return super.visitAnchor(anchor, count);
     }
 
+    @Override
+    public Yaml visitTag(Yaml.Tag tag, AtomicInteger count) {
+        if(tag.getPrefix().contains("\n")) {
+            count.incrementAndGet();
+        }
+        return super.visitTag(tag, count);
```
knutwannheden, line 93: I get the impression that we should also check the suffix here. Maybe we even want to add a test.

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
```diff
@@ -506,4 +515,94 @@ public String toString() {
...
+        public Tag(UUID id, String prefix, String suffix, Markers markers, String name, Kind kind) {
+            this.id = id;
+            this.prefix = prefix;
+            this.suffix = suffix;
+            this.markers = markers;
+            this.name = name;
+            this.kind = kind;
+        }
+
+        public Tag(UUID id, String prefix, String suffix, Markers markers, String text) {
```
knutwannheden, line 573: I think we should push this logic into the parser instead or what is the use case you had in mind?

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
```diff
@@ -506,4 +515,94 @@ public String toString() {
...
+        }
+
+        @EqualsAndHashCode.Include
+        UUID id;
+
+        @With
+        String prefix;
+
+        @With
+        String suffix;
```
knutwannheden, line 553: Nit: With the exception of the id and markers fields, we try to order the other fields so that they correspond to the order in the syntax.

rewrite-yaml/src/main/java/org/openrewrite/yaml/CountLinesVisitor.java
greg-at-moderne: Good point.

BTW this logic is likely broken or incomplete. It doesn't take into account multiple line breaks.

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
greg-at-moderne: c80851a60

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
greg-at-moderne: e2d74f290

rewrite-yaml/src/main/java/org/openrewrite/yaml/CountLinesVisitor.java
greg-at-moderne: ea7f9ea08 + 58c8dabd2

timtebeek:
Hi @greg-at-moderne ; would you mind updating the downstream projects that are using these Yaml constructors? We saw failures in rewrite-spring already, and presumably a few others: https://github.com/openrewrite/rewrite-spring/pull/677/commits/2ed3511e279e8784dc67a4027809d2b0a890771f

Note that we have a recipe for this purpose specifically, which you can run against the OpenRewrite and Moderne Recipes orgs:
https://app.moderne.io/recipes/org.openrewrite.java.AddNullMethodArgument

greg-at-moderne:
3 runs:
- for Yaml.Scalar: https://app.moderne.io/results/IQVSmcmRC => has only `rewrite-spring` and `rewrite` 👍
- for Yaml.Mapping: https://app.moderne.io/results/6ZCiQ1cXu => same story 👍
- for Yaml.Sequence: https://app.moderne.io/results/I6LhVcNoU => only `rewrite` 👍

So it seems like `rewrite-spring` might have been the only one?
Will check for failures.

PR #4936, opened 2025-01-23
Json formatting arrays

greg-at-moderne:
### Checklist


rewrite-json/src/main/java/org/openrewrite/json/format/AutoFormat.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class AutoFormat extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Format JSON";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Indents JSON using the most common indentation size and tabs or space choice in use in the file.";
```
knutwannheden, line 29: As this will be about more than indenting, we might want to make the description a bit more general.

greg-at-moderne:
#4953 instead.

PR #4944, opened 2025-01-24
Adding ReplaceLegacyAttributeIndexSyntax

greg-at-moderne:
## What's changed?
Adding a new recipe for HCL called `ReplaceLegacyAttributeIndexSyntax`.
This is to migrate from this syntax which HCL considers legacy (which BTW we have recently added support for in the parser - see #4901):
```
aws_acm_certificate.google_dot_com.0.resource_record_name
```
to the "modern" one doing effectively the same:
```
aws_acm_certificate.google_dot_com[0].resource_record_name
```

## What's your motivation?
To enable HCL users of OpenRewrite to migrate their code to new syntax.


knutwannheden (approved):
Very nice!

PR #4931, opened 2025-01-21
Autodetect formatting style of JSON code - continued

greg-at-moderne:
## What's changed?
Continuation of #4916 - i.e. adding more coverage for auto-detecting JSON style. By more coverage I mean more cases in the original code are handled now.

## What's your motivation?
Another step towards full support of JSON autoformat.

PR #4953, opened 2025-01-27
Applying JSON formatting to members of objects and arrays

greg-at-moderne:
## What's changed?
Continuation of #4931.
Additional step in JSON autoformatting. Taking care of the indents, prefixes and suffixes around object and array members.

## What's your motivation?
To provide more complete support for JSON style autodetection and autoformatting.

As indicated in Github comments.

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
@@ -98,13 +98,11 @@ public class WrappingAndBracesVisitor<P> extends JsonIsoVisitor<P> {
     }
 
     private <J extends Json> List<JsonRightPadded<J>> ensureCollectionHasIndents(List<JsonRightPadded<J>> elements, LineWrapSetting wrapping) {
-        AtomicBoolean first = new AtomicBoolean(true);
-        return ListUtils.map(elements, elem -> {
-            boolean isFirst = first.getAndSet(false);
+        return ListUtils.map(elements, (Integer idx, JsonRightPadded<J> elem) -> {
             Space prefix = elem.getElement().getPrefix();
             final String newPrefixString;
             String currentAfterNewLine = prefix.getWhitespaceIndent();
-            if (wrapping == LineWrapSetting.DoNotWrap && isFirst) {
+            if (wrapping == LineWrapSetting.DoNotWrap && idx == 0) {
                 newPrefixString = "";
             } else if (wrapping == LineWrapSetting.DoNotWrap) {
                 newPrefixString = " ";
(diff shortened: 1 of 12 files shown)


rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
```diff
@@ -82,28 +97,67 @@ public Json.Array visitArray(Json.Array array, P p) {
...
 
-    private static <JS extends Json> List<JsonRightPadded<JS>> applyWrappingStyleToPrefixes(List<JsonRightPadded<JS>> elements, String possibleNewLine) {
+    private <J extends Json> List<JsonRightPadded<J>> ensureCollectionHasIndents(List<JsonRightPadded<J>> elements, LineWrapSetting wrapping) {
+        AtomicBoolean first = new AtomicBoolean(true);
         return ListUtils.map(elements, elem -> {
-            String oldAfterNewLine = elem.getElement().getPrefix().getWhitespaceIndent();
-            String newPrefix = possibleNewLine + oldAfterNewLine;
-            if (!newPrefix.equals(elem.getElement().getPrefix().getWhitespace())) {
-                return elem.withElement(elem.getElement().withPrefix(Space.build(newPrefix, emptyList())));
+            boolean isFirst = first.getAndSet(false);
```
greg-at-moderne, line 103: The `first` manual logic looks ugly to me.
But I fail to have a better idea on how express it.

I considered/tried:
# Implement an extra `ListUtils.mapWithSpecialFirstHandling` which would take: `List, Function<T, T> toBeAppliedToAllButFirst and Function<T, T> toBeAppliedToFirst`.
# Or similarly, implement `ListUtils.mapAllButFirst` with the same signature as `map()`.
# Or I tried adding some `mapWithIndex` which would zip the input list with index and only then apply a `Function<T, int>` to it.

Each of them turned out to be a nightmare in `ListUtils` code or basically copy&paste of the method.

Any advice?

rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
```diff
@@ -82,28 +97,67 @@ public Json.Array visitArray(Json.Array array, P p) {
...
-    private static <JS extends Json> List<JsonRightPadded<JS>> applyWrappingStyleToLastChildSuffix(List<JsonRightPadded<JS>> elements, String possibleNewLine) {
-        return ListUtils.mapLast(elements, last -> {
-            String currentAfterNewLine = last.getAfter().getWhitespaceIndent();
-            String currentAfter = last.getAfter().getWhitespace();
-            String newAfter = possibleNewLine + currentAfterNewLine;
+    private <JS extends Json> List<JsonRightPadded<JS>> applyWrappingStyleToSuffixes(List<JsonRightPadded<JS>> elements, LineWrapSetting wrapping, String relativeIndent) {
+        AtomicInteger i = new AtomicInteger(0);
+        return ListUtils.map(elements, elem -> {
+            boolean isLast = i.get() == elements.size() - 1;
+            i.incrementAndGet();
```
greg-at-moderne, line 128: Same issue with special logic for last element in the list.

rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
knutwannheden: Not sure I understand the issue here. I suppose you saw the map() overload where the lambda receives an index along with the element.

greg-at-moderne: 🤦 Now I see it :)

rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
greg-at-moderne: c90d9f75d

rewrite-json/src/main/java/org/openrewrite/json/format/WrappingAndBracesVisitor.java
greg-at-moderne: c90d9f75d

PR #4957, opened 2025-01-28
Gradle wrapper scripts packed as resources

jkschneider:
## What's changed?

`UpdateGradleWrapper` will now no longer download whole Gradle distributions just to retrieve the wrapper scripts from them, which causes slowness or timeouts in large scale situations where the artifact repository from which the wrapper is being downloaded from is in a degraded performance state.

After this change, `rewrite-gradle` will have a resources directory that contains a CSV with information about each Gradle version available and a checksum of the contents of the unix and windows wrapper scripts, which in turn can be retrieved from the unix/ or windows/ folders.

image

The checksums are Adler32, which is judged to be sufficiently collision immune given the size of the data set and are easily human readable for quick checks.

image

The main method on `GradleWrapperScriptDownloader` should be executed by CI periodically and followed up with a Git commit when the downloader finds new versions of Gradle to add to the list (TODO add the github actions workflow for this).

Because the `Remote` for a wrapper script is now coming from a classpath resource, a new implementation of `Remote` has been added called `RemoteResource`. `Remote#getUri`/`withUri` have been removed since those were specific to the implementation details of `RemoteFile` and `RemoteArchive` and only used in the update Gradle wrapper recipe anyway.

## What's your motivation?

Recipe timeouts on `UpdateGradleWrapper`.

@sambsnyd @shanman190


rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
```diff
@@ -0,0 +1,183 @@
+version,gradlew,gradlewBat
```
shanman190, line 1: question: Why a CSV with Adler32 versus a directory structure?

Ex:
```
/META-INF/rewrite/gradle-wrapper/
    4.10.3/
        gradlew
        gradlew.bat
    5.6.4/
        gradlew
        gradlew.bat
    ...
```

I'm guessing some of this is that having the CSV makes finding the versions a little bit easier.

rewrite-core/src/main/java/org/openrewrite/remote/RemoteResource.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.UUID;
+
+/**
+ * A remote resource that can be fetched from an {@link java.io.InputStream} which
+ * could be, for example, a classpath resource.
+ */
+@Value
+@EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+@With
+public class RemoteResource implements Remote {
```
shanman190, line 40: callout: This could potentially be represented via the same mechanism that was employed within `RemoteArchive` by accepting `file:` URIs. That would then allow us to reduce the API surface area introduced via this PR.

timtebeek:
In case it helps, there's a recent example of a [GitHub workflow to update a CSV periodically](https://github.com/openrewrite/rewrite-migrate-java/blob/main/.github/workflows/sdkman-candidates.yml).

rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/GradleWrapperScriptDownloader.java
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                }
+                System.out.printf("%03d: %s downloaded.%s%n", i.incrementAndGet(), v, threadName);
+            } catch (Throwable t) {
+                // FIXME There are some wrappers that are not downloading successfully. Why?
+                System.out.printf("%03d: %s failed to download: %s.%s%n", i.incrementAndGet(), v, t.getMessage(), threadName);
+                return null;
+            } finally {
+                lock.unlock();
+            }
```
shanman190, line 82: callout: I'd almost expect for this to be written out by the main thread once all of the downloads were completed. Given the use of a concurrent map, that would also alleviate as much contention on the lock which isn't needed -- especially with just the download, then the write of the CSV by the main thread.

shanman190:
callout: While we have encountered timeouts for the `UpdateGradleWrapper` recipe, we don't want to forget that the `UpdateMavenWrapper` recipe has nearly identical behavior.

https://github.com/openrewrite/rewrite/blob/main/rewrite-maven%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fmaven%2FUpdateMavenWrapper.java#L323-L341

Maven remote files:
* maven-wrapper.jar
* MavenWrapperDownloader.java
* mvnw
* mvnw.cmd

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
jkschneider: Because there is so many versions for which the wrapper script didn't change. So the proposed structure keeps only one copy of any given iteration of the wrapper script.

rewrite-core/src/main/java/org/openrewrite/remote/RemoteResource.java
jkschneider: I don't think `file:` URIs work for classpath resources in every situation.

lkerford (approved):

PR #4962, opened 2025-01-28
Trim descriptions in `YamlResourceLoader`

knutwannheden:
The YAML parser could return the description with a trailing line terminator, which in turn then causes the validation with the trailing period in the description to fail.


knutwannheden:
Close this for now. Not exactly sure why this popped up in rewrite-codemods and never before.

PR #4951, opened 2025-01-26
More robust Groovy parentheses handling

knutwannheden:

PR #4964, opened 2025-01-28
Exclude parenthese determination from comments and string literals

jevanlingen:
## What's changed?
Exclude parentheses handling for comments and string literals.

## What's your motivation?
Improvement to
- https://github.com/openrewrite/rewrite/pull/4951

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -62,8 +62,8 @@ import java.util.stream.Stream;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toMap;
 import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.groovy.GroovyParserVisitor.Delimiter.*;
 import static org.openrewrite.internal.StringUtils.indexOfNextNonWhitespace;
 import static org.openrewrite.java.tree.Space.EMPTY;
 import static org.openrewrite.java.tree.Space.format;
@@ -1337,13 +1337,11 @@ public class GroovyParserVisitor {
                         value = "@" + value;
                         text = "@" + text;
                     } else {
-                        String delimiter = getDelimiter();
+                        Delimiter delimiter = getDelimiter(cursor);
                         if (delimiter != null) {
-                            // handle the `pattern` operator
-                            String endDelimiter = "~/".equals(delimiter) ? "/" : delimiter;
                             // Get the string literal from the source, so escaping of newlines and the like works out of the box
-                            value = sourceSubstring(cursor + delimiter.length(), endDelimiter);
-                            text = delimiter + value + endDelimiter;
+                            value = sourceSubstring(cursor + delimiter.start.length(), delimiter.end);
+                            text = delimiter.start + value + delimiter.end;
                         }
                     }
                 } else if (expression.isNullExpression()) {
@@ -1597,7 +1595,7 @@ public class GroovyParserVisitor {
         @Override
         public void visitGStringExpression(GStringExpression gstring) {
             Space fmt = whitespace();
-            String delimiter = getDelimiter();
+            String delimiter = getDelimiter(cursor).start;
             skip(delimiter); // Opening delim for GString
 
             NavigableMap<LineColumn, org.codehaus.groovy.ast.expr.Expression> sortedByPosition = new TreeMap<>();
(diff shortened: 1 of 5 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2537,13 +2537,30 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
...
+        // Determine level of parentheses by going through the source, don't count parentheses in comments
         int count = 0;
+        boolean inSingleLineComment = false;
+        boolean inMultilineComment = false;
         for (int i = start; i < end; i++) {
-            if (source.charAt(i) == '(') {
-                count++;
-            } else if (source.charAt(i) == ')') {
-                count--;
+            if (source.charAt(i) == '/' && (end - 1) != i && source.charAt(i + 1) == '/') {
```
knutwannheden, line 2545: I would add a variable like this:
```java
char c = source.charAt(i);
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2537,38 +2535,67 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
...
+            return DOUBLE_QUOTE;
         } else if (maybeDelimiter.startsWith("'")) {
-            return "'";
+            return SINGLE_QUOTE;
         }
         return null;
     }
 
+    @RequiredArgsConstructor
+    enum Delimiter {
```
jevanlingen, line 2584: We could lift this field even to the `G` class. The GString class has also a delimiter field:

```java
final class GString implements G, Statement, Expression {
  UUID id;
  Space prefix;
  Markers markers;
  String delimiter;
  //...
```

Even the printer does already some magic to make it work:

```java
public J visitGString(G.GString gString, PrintOutputCapture<P> p) {
     //...   
     if ("$/".equals(delimiter)) {
       p.append("/$");   // <- which could then now be modeled like `delimiter.end`
     } else {
       p.append(delimiter);
     }
```

Problem would be we break current synchronised LSTs, so I did not implement this. Any ideas here @knutwannheden?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2537,38 +2535,69 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
...
+                delimiter = null;
             }
         }
         return Math.max(count, 0);
     }
 
-    private @Nullable String getDelimiter() {
+    private @Nullable Delimiter getDelimiter(int cursor) {
         String maybeDelimiter = source.substring(cursor, Math.min(cursor + 3, source.length()));
         if (maybeDelimiter.startsWith("$/")) {
```
knutwannheden, line 2563: The `String#startsWith(String, int)` overload transparently does the bounds checking. So you can just do this:
```suggestion
        if (source.startsWith("$/", cursor)) {
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2537,38 +2535,69 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
...
+                if (delimiter == null) {
+                    if (source.charAt(i) == '(') {
+                        count++;
+                    } else if (source.charAt(i) == ')') {
+                        count--;
+                    }
+                } else {
+                    i += delimiter.open.length() - 1; // skip the next chars for the rest of the delimiter
+                }
+            } else if (delimiter.close.equals(source.substring(i, Math.min(i + delimiter.close.length(), source.length())))) {
```
knutwannheden, line 2553: Same here:
```suggestion
            } else if (source.startsWith(delimiter.close, i)) {
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2537,38 +2535,69 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
         int start = sourceLineNumberOffsets[startingLineNumber - 1] + startingColumn - 1;
         int end = sourceLineNumberOffsets[endingLineNumber - 1] + endingColumn - 1;
 
-        // Determine level of parentheses by going through the source
+        // Determine level of parentheses by going through the source, don't count parentheses in comments, string literals and regexes
         int count = 0;
+        Delimiter delimiter = null;
```
knutwannheden, line 2540: Is it safe to assume that we can "ignore" the parens in G-strings like: `"Hello ${(to).you}!"`?

knutwannheden (commented):
It is a shame that we need to handcode a lexer here. Maybe we should at least explore the possibility of using `org.codehaus.groovy.antlr.parser.GroovyLexer` instead?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
knutwannheden: I don't think it is worth to change the model. We could instead provide a utility method to print a string using certain delimiters.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: Interesting question! Answer is "yes", because the placeholder expressions are parsed as new groovy elements. To prove my point, I added following test:

```java
@Test
void gStringWithStringLiteralsWithParentheses() {
    rewriteRun(
      groovy(
        """
         "Hello ${from(":-)").via(''':-|(''').via(":-)").to(':-(')}!"
         """
      )
    );
}
```

💪🏻.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: What about moving the `Delimiter` enum to a separate file? Then we could just transform above code to:

```java
public J visitGString(G.GString gString, PrintOutputCapture<P> p) {
  //...   
  Delimiter delimiter = Delimiter.of(gString.getDelimiter());
  p.append(delimiter.open);
  visit(gString.getStrings(), p);
  p.append(delimiter.close);
  //..
}
```

I think that's a nice middle ground of little improvement without altering too much.

knutwannheden: Yes, that sounds like an option. Let's add it to an `internal` package. I don't really see recipes needing this.

knutwannheden:
The idea with the lexer would have been to let the ANTLR lexer tokenize some given part of the source code. It would then return a stream of tokens, where each token is stuff like keywords, literals, identifiers, punctuation characters, comments, and of course also tokens representing open and close parentheses. So the idea would have been to just zip through this token stream and count the parentheses. The lexer would worry about literals, comments, etc.

Whether it is any easier or better in the end is hard to tell without trying it out. But an ANTLR lexer is definitely very well optimized for tokenization.

jevanlingen:
Ok, now I understand what you were getting to. My knowledge of lexers is zero to none, as you could see in the above post. I guess if we want to try that, you should take over. Still, I think it's an unnecessary action now that we have everything in place. Let's wrap up this PR; if we ever face the parenthesis question again, let's reconsider whether using the ANTLR lexer is a good idea.

PR #4965, opened 2025-01-28
--offline in some of the rewrite-gradle tests

greg-at-moderne:
## What's changed?
Changing how some of the `rewrite-gradle` tests are executed. (Roughly 20% of them)
Passing extra `--offline` parameter to Gradle as made available in https://github.com/openrewrite/rewrite-gradle-tooling-model/pull/31

## What's your motivation?
To speed up the tests in this repo, `rewrite-gradle` particularly.
This seems to reduce the total build time of `rewrite-gradle` from 17 min to 9 min in an unscientific test.

TBH, some of the tests fail even in clean `main` on my laptop, so I am not sure if we can use it in all tests as suggested by the current PR content.
```
406 tests completed, 16 failed, 1 skipped
```

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.openrewrite.gradle;
 
 import org.openrewrite.test.RewriteTest;
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
@@ -35,9 +35,9 @@ public class GradleSetup extends RewriteGradleTest {
               recipeSpec -> recipeSpec.beforeRecipe(withToolingApi()),
               buildGradle(
                 """
-                  repositories {
-                   mavenCentral()
-                  }
+                repositories {
+                 mavenCentral()
+                }
                 """
             ));
             alreadySetUp = true;
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
@@ -1,0 +1,31 @@
+package org.openrewrite.gradle;
+
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.gradle.Assertions.buildGradle;
+import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
+
+public class GradleSetup extends RewriteGradleTest {
+    private static Boolean alreadySetUp = false;
+
+    // This is a one-time initialization of Gradle dependencies for the tests. It wouldn't be needed
+    // if we didn't run (some of) the tests in offline mode.
+    public static final void makeSureArtifactsAreDownloaded() {
+        synchronized(GradleSetup.class) {
+            if (alreadySetUp) {
+                return;
+            }
+            RewriteTest fakeTest = new RewriteTest() {};
+            fakeTest.rewriteRun(
+              recipeSpec -> recipeSpec.beforeRecipe(withToolingApi()),
+              buildGradle(
+                """
+                  repositories {
+                   mavenCentral()
+                  }
+                """
+            ));
+            alreadySetUp = true;
+        }
+    }
+}
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
@@ -20,7 +20,7 @@ import org.openrewrite.test.RewriteTest;
 import static org.openrewrite.gradle.Assertions.buildGradle;
 import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
 
-public class GradleSetup extends RewriteGradleTest {
+public class GradleSetup extends RewriteGradleTest implements RewriteTest {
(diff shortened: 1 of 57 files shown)


shanman190:
@greg-at-moderne, I'm not sure it's possible to rely on offline mode in these contexts as what that does internally to Gradle is prevent dependency and plugin downloads from the Internet. If in the current execution context, those dependencies are not present within the Gradle cache, then any test case using the tooling API and needing to download a dependency or plugin would inherently fail with a result of not being able to create the `GradleProject` marker correctly.

greg-at-moderne:
> any test case using the tooling API and needing to download a dependency or plugin would inherently fail

And since you are the expert - would that be all tests in `rewrite-gradle` or some?

greg-at-moderne:
Also - would that be solvable with having a single run of Gradle in online mode first (before the whole suite begins), and then executing all (or some) of the tests in offline mode?

shanman190:
It's only going to be a partial scope for sure. Only those tests that make use of the tooling API AND are using an external dependency or plugin.

If those dependencies are downloaded upfront, then yes, the offline mode would work consistently.

knutwannheden:
Can we instead somehow get the tests to use the same cached downloaded artifacts and with a timeout (just for the execution of tests) that is high enough, that we don't need to download the Maven metadata again, if that is a problem? Would sure be nice to get faster builds.

shanman190:
For local, I'd expect the local Gradle cache to be populated on the first execution, then from there everything be pretty quick. In GitHub actions, it would be based upon if the Gradle cache is cached or not (I think it is). It's not like the original versions are changing between runs.

sambsnyd:
As Shannon has pointed out this breaks if anything is not already present in your local filesystem. You get a stack trace many layers deep, which does at the bottom clearly spell out what's wrong so that is nice, at least. With many frames omitted for brevity:
```
java.lang.RuntimeException: org.gradle.tooling.BuildException: Could not fetch model of type 'OpenRewriteModel' using connection to Gradle distribution 'https://services.gradle.org/distributions/gradle-8.12-bin.zip'.
	at org.openrewrite.test.UncheckedConsumer.accept(UncheckedConsumer.java:29)
	...
Caused by: org.gradle.tooling.BuildException: Could not fetch model of type 'OpenRewriteModel' using connection to Gradle distribution 'https://services.gradle.org/distributions/gradle-8.12-bin.zip'.
	at app//org.gradle.tooling.internal.consumer.ConnectionExceptionTransformer.transform(ConnectionExceptionTransformer.java:69)
	...
Caused by: org.gradle.internal.exceptions.LocationAwareException: A problem occurred configuring root project 'project7036922274535633441'.
	at org.gradle.initialization.exception.DefaultExceptionAnalyser.transform(DefaultExceptionAnalyser.java:97)
	...
Caused by: org.gradle.api.internal.artifacts.ivyservice.TypedResolveException: Could not resolve all artifacts for configuration 'classpath'.
	at org.gradle.api.internal.artifacts.ResolveExceptionMapper.mapFailures(ResolveExceptionMapper.java:56)
	...
Caused by: org.gradle.internal.resolve.ArtifactResolveException: Could not download rewrite-core-8.44.1.jar (org.openrewrite:rewrite-core:8.44.1): No cached version available for offline mode
	at org.gradle.api.internal.artifacts.ivyservice.ivyresolve.StartParameterResolutionOverride$FailedRemoteAccess.resolveArtifact(StartParameterResolutionOverride.java:166)
	...
```

I don't think we can accept this without some mechanism which ensures thorough per-population of the dependency caches. Perhaps something else needs to be done on CI

greg-at-moderne:
The PR has been updated to:
- add `GradleSetup.makeSureArtifactsAreDownloaded()` run to make sure the needed dependencies are in local store
- limit the scope of the change to selected ~20% of the tests in `rewrite-gradle` as the first step. These have been selected as the ones not having any extra dependencies beyond the ones baked into https://github.com/openrewrite/rewrite-gradle-tooling-model

The build duration drops from 17 min to 9min.

Kindly requesting @sambsnyd @shanman190 @knutwannheden to re-review.

rewrite-gradle/src/test/java/org/openrewrite/gradle/RewriteGradleTest.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.gradle;
+
+import org.junit.jupiter.api.BeforeAll;
+
+public class RewriteGradleTest {
```
greg-at-moderne, line 20: Maybe rename it for now if not all the tests are covered with this?

rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleSetup.java
```diff
@@ -0,0 +1,46 @@
...
+            RewriteTest fakeTest = new RewriteTest() {};
+            fakeTest.rewriteRun(
+              recipeSpec -> recipeSpec.beforeRecipe(withToolingApi()),
+              buildGradle(
+                """
+                repositories {
+                 mavenCentral()
+                }
+                """
+            ));
```
shanman190, line 42: This reducing the build time makes me _very suspicious_. With just this block and offline, makes me highly suspicious that there is something else entirely going on that's neither related to dependencies or offline mode and this is just happening to fix it.

From a Gradle standpoint, the following things happen on the first test execution and a clean workspace, for openrewrite/rewrite specifically:

* Download the specified Gradle distribution
* Start a Gradle daemon using that version
* Run the first test using the prepackaged initscript from rewrite-gradle-tooling-model
  * Initscript downloads it's model dependencies (rewrite-gradle-model-plugin and rewrite-maven snapshots)
* `GradleProject` marker is built and returned for the project directory (this step also involves downloading the dependencies and plugins for the project under test)


For every subsequent test, it should be using the warm daemon, the warm cache (with the exception of project dependencies), and only need to be performing the marker creation. Now one potentially sensitive part may be that the initscript uses the option to not cache changing versions -- of which `latest.integration` is one of -- so with each run it would also be inspecting the remote repositories declared (mavenLocal, Sonatype snapshot, and mavenCentral) for updated snapshot dependencies from the initscript itself. What I suspect is that this is instead where the random latency may be being introduced and as such it may be better to alter these caching rules back to their defaults (ie. By removing the rules all together; offline mode short circuits these as well, but much much more) for the initscript during standard executions. For developer executions, we can always modify the initscript or create a separate one.

greg-at-moderne: Thanks for the insights.

BTW this:

> it would also be inspecting the remote repositories declared [...] for updated snapshot dependencies from the initscript itself.

is what I observed to take most of the time for the most typical simple cases (roughly ~3s out of ~4s).

greg-at-moderne:
Closing as the solution didn't get popularity.
Will research https://github.com/openrewrite/rewrite/pull/4965#discussion_r1934095405 at some point.

PR #4977, opened 2025-01-31
Using withToolingApiUsingModulesCaching in rewrite-gradle tests

greg-at-moderne:
## What's changed?
Using `withToolingApiUsingModulesCaching` added in https://github.com/openrewrite/rewrite-gradle-tooling-model/pull/33.
Following the advice by @shanman190 at: https://github.com/openrewrite/rewrite/pull/4965#discussion_r1934073362

## What's your motivation?
To speed up CI in this project, specifically the duration of `rewrite-gradle` tests.

EDIT: Adding some non-scientific measurements:

| What | Without this change | With this change |
|------|--------------------|-----------------|
| `./gradlew rewrite-gradle:test` on my laptop | [13m 18s](https://ge.openrewrite.org/s/cvtnsj4koaljw) | [2m 6s](https://ge.openrewrite.org/s/xouyjhlxlyh7g) |
| CI on some other PR touching `rewrite-gradle` only | some PR a few days ago: [9m 43s](https://ge.openrewrite.org/s/27wxickd3r3cw) | this PR: [8m 21s](https://ge.openrewrite.org/s/xyk2hh2c6iwf2) |
| CI on a PR touching `rewrite-core` | [19m 17s](https://ge.openrewrite.org/s/ixitf4ctzeuvs) | [17m 59s](https://ge.openrewrite.org/s/rirclc3feblm4) |

(So it seems like the biggest win is for people living far away from Maven Central servers in terms of network latency)

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
@@ -33,7 +33,7 @@ import java.util.Optional;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.gradle.Assertions.buildGradle;
 import static org.openrewrite.gradle.Assertions.settingsGradle;
-import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
+import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApiUsingModulesCaching;
 import static org.openrewrite.groovy.Assertions.groovy;
 import static org.openrewrite.groovy.Assertions.srcMainGroovy;
 import static org.openrewrite.java.Assertions.java;
@@ -48,7 +48,7 @@ import static org.openrewrite.properties.Assertions.properties;
 class AddDependencyTest implements RewriteTest {
     @Override
     public void defaults(RecipeSpec spec) {
-        spec.beforeRecipe(withToolingApi())
+        spec.beforeRecipe(withToolingApiUsingModulesCaching())
           .parser(JavaParser.fromJavaVersion().classpath("junit-jupiter-api", "guava", "jackson-databind", "jackson-core", "lombok"));
     }
 
@@ -1124,7 +1124,7 @@ class AddDependencyTest implements RewriteTest {
     void addDynamicVersionDependency() {
         rewriteRun(
           spec -> spec
-            .beforeRecipe(withToolingApi())
+            .beforeRecipe(withToolingApiUsingModulesCaching())
             .recipe(addDependency("org.openrewrite:rewrite-core:7.39.X", "java.util.Date", "implementation")),
           mavenProject("project",
             srcMainGroovy(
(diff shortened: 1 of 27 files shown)


knutwannheden:
Looks promising. What difference does that make when running a clean build?

shanman190:
@greg-at-moderne, the reason that the options exist are because within the CI context and iterating quickly, you want to be able to pick up the latest dynamic version for `latest.integration` for rewrite-maven or the `org.openrewrite.gradle.tooling:plugin` quickly -- which by the way these two options _only_ come into play when pulling from remotely deployed artifacts. When dynamic (latest.release, latest.integration) or changing (1.0-SNAPSHOT) dependencies are in use by default, Gradle will cache these for 24 hours. In OpenRewrite, the snapshots are always used and integrated together rapidly with each other. You can see the same thing being done for all of the top-level OpenRewrite projects via the shared binary build plugin [openrewrite/rewrite-build-gradle-plugin](https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/main/src%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FRewriteJavaPlugin.java#L46-L49).

So whether you're developing on a core module, the Gradle tooling API, or one of the OSS build plugin, you're getting the latest snapshot versions for everything. It sounds like in your context, you're a little further from one of the Maven Central repositories, so you're experiencing increased latency as a result of these options under all of the above contexts. It may be that you're using the Intellij runner for most activities, which given it bypasses Gradle often, would result in you not feeling the impact of the core module variation conveniently. But when here in rewrite-gradle, it's directly in front of you resulting in being felt early and often.

@jkschneider & @sambsnyd, maybe we've reached a point where a value of `0.seconds` is now becoming a pain point and should be revaluated within the additional data?

greg-at-moderne:
>  from different contexts and physical locations.

IMHO, our CI running in Github Actions should be the top criterion. I've included my laptop just for the context.

shanman190:
Gotcha! Good context to know. I'll do some research into optimizing the build based upon the GHA build scans.

sambsnyd:
@shanman190 if we were to change the snapshot caching time from 0 to something higher I wonder what the right value is and if we should have that same value on CI and locally.

shanman190:
@sambsnyd, I think that 5-10 minutes would be a reasonable value for the default.

One thing I've been pondering is if we flip the script that Greg has started and make it so that the defaults are the 5-10 minutes, then add a `withDevelopmentToolingApi` method which uses an initscript with values of 0. In a local setting though, we're probably just publishing to mavenLocal in which case we get the updated immediately regardless. So having the alternative mechanism may not be worth the test API surface area at all.

But changing the defaults has huge, positive impact across anyone using the tooling API today.

timtebeek:
Went with the above approach of a five minute cache by default, and pushed a new release that should pick this up:
- https://github.com/openrewrite/rewrite-gradle-tooling-model/releases/tag/v2.13.0

Thanks all!

Issue #4068, opened 2024-03-04
Groovy parser fails when using duplicate static star imports


Jnb1234:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite 6.8.4
- Gradle plugin 7.4.2
- rewrite-_module_ v1.2.3

## How are you running OpenRewrite?

I am using the Gradle plugin, and my project is a single module project.
```xml
plugins {
    id 'maven-publish'
    id 'groovy'
    id 'idea'
    id 'net.researchgate.release' version '2.8.1'
    id "com.github.ben-manes.versions" version "0.36.0"
    id "com.gorylenko.gradle-git-properties" version "2.3.2"
    id 'java-library'
    id 'org.openrewrite.rewrite' version '6.8.4'
}

rewrite {
    activeRecipe("org.openrewrite.java.RemoveUnusedImports")
}
```

## What is the smallest, simplest way to reproduce the problem?

```java
@Test
void staticStarImport() {
  rewriteRun(
    groovy(
      """
        import static java.util.Collections.*
        import static java.util.Collections.*
      """
    )
  );
}
```

## What did you expect to see?

No parsing error

## What did you see instead?

parsing  error

## What is the full stack trace of any errors you encountered?

```
Error during rewrite dry run
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "file.groovy":
diff --git a/file.groovy b/file.groovy
index dff510f..56679bc 100644
--- a/file.groovy
+++ b/file.groovy
@@ -1,2 +1,2 @@ 
 import static java.util.Collections.*
-import static java.util.Collections.*
\ No newline at end of file
+import static java.util.Collections.//*
\ No newline at end of file
] 
expected: 
  "import static java.util.Collections.*
  import static java.util.Collections.*"
 but was: 
  "import static java.util.Collections.*
  import static java.util.Collections.//*"
```

timtebeek:
Thanks for the clear example @Jnb1234 ! It seems to fail on this particular line
https://github.com/openrewrite/rewrite/blob/06a1bc81d0c56a710f52133861b985ca196e1b13/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L232

Is there any reason why the types might be absent in your case above? Are those imports on the classpath?

[...]

PR #4982, opened 2025-02-03
Support duplicate static star imports for Groovy

jevanlingen:
## What's changed?
- Duplicate static star imports work as well.
- Shorten the parser code for ImportNode (we have now methods like `sourceStartsWith`)

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4068

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2786,7 +2786,7 @@ public class GroovyParserVisitor {
     /**
      * Duplicate imports do work out of the box for import, star-import and static-import.
      * For static-star-import, this does work though.
-     * The groovy compiler does only save the last duplicate import instead of all, so parse all static star imports by hand.
+     * The groovy compiler does only memoize the last duplicate import instead of all, so retrieve all static star imports by hand.
      */
     private List<ImportNode> getStaticStarImports(ModuleNode ast) {
         List<ImportNode> completeStaticStarImports = new ArrayList<>();
@@ -2822,8 +2822,7 @@ public class GroovyParserVisitor {
                     }
 
                     if (packageEnd < line.length() && line.charAt(packageEnd) == '*') {
-                        ImportNode orginalImportNode = staticStarImports.get(line.substring(packageBegin, packageEnd - 1));
-                        ImportNode node = new ImportNode(orginalImportNode.getType());
+                        ImportNode node = new ImportNode(staticStarImports.get(line.substring(packageBegin, packageEnd - 1)).getType());
                         node.setLineNumber(i + 1);
                         node.setColumnNumber(importIndex + 1);
                         completeStaticStarImports.add(node);
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2245,11 +2245,11 @@ public class GroovyParserVisitor {
         } else if (node instanceof ImportNode) {
             ImportNode importNode = (ImportNode) node;
             Space prefix = sourceBefore("import");
-            JLeftPadded<Boolean> static_;
+            JLeftPadded<Boolean> statik;
             if (importNode.isStatic()) {
-                static_ = padLeft(sourceBefore("static"), true);
+                statik = padLeft(sourceBefore("static"), true);
             } else {
-                static_ = padLeft(EMPTY, false);
+                statik = padLeft(EMPTY, false);
             }
             String packageName = importNode.getPackageName();
             J.FieldAccess qualid;
@@ -2276,7 +2276,7 @@ public class GroovyParserVisitor {
                 alias = padLeft(sourceBefore("as"), new J.Identifier(randomId(), sourceBefore(simpleName), Markers.EMPTY, emptyList(), simpleName, null, null));
             }
 
-            J.Import anImport = new J.Import(randomId(), prefix, Markers.EMPTY, static_, qualid, alias);
+            J.Import anImport = new J.Import(randomId(), prefix, Markers.EMPTY, statik, qualid, alias);
             return maybeSemicolon(anImport);
         }
 
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -162,7 +162,7 @@ public class GroovyParserVisitor {
         }
 
         // Duplicate imports do work out of the box for import, star-import and static-import.
-        // For static-star-import, this does work though.
+        // For static-star-import, this does not work though.
         // The groovy compiler does only save the last duplicate import instead of all, so parse all static star imports by hand.
         Map<String, ImportNode> staticStarImports = ast.getStaticStarImports();
         if (!staticStarImports.isEmpty()) {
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -159,8 +159,31 @@ public G.CompilationUnit visit(SourceUnit unit, ModuleNode ast) throws GroovyPar
...
+            // Take source code until last static star import for performance reasons
+            int lastLineNumber = -1;
+            for (ImportNode anImport : ast.getStaticStarImports().values()) {
+                lastLineNumber = Math.max(lastLineNumber, anImport.getLastLineNumber());
+            }
+            String importSource = sourceLineNumberOffsets.length <= lastLineNumber ? source : source.substring(0, sourceLineNumberOffsets[lastLineNumber]);
+
+            // Create a node for each `import static`
+            String[] lines = importSource.split("\n");
+            Pattern pattern = Pattern.compile("import\\s+static\\s+([\\w.]*)\\.\\*");
```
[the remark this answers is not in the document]
jevanlingen: Never mind, I'll change it anyways; I don't think the code will be that much harder to read without the regex.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -159,8 +160,51 @@ public G.CompilationUnit visit(SourceUnit unit, ModuleNode ast) throws GroovyPar
         for (ImportNode anImport : ast.getStaticImports().values()) {
             sortedByPosition.computeIfAbsent(pos(anImport), i -> new ArrayList<>()).add(anImport);
         }
-        for (ImportNode anImport : ast.getStaticStarImports().values()) {
-            sortedByPosition.computeIfAbsent(pos(anImport), i -> new ArrayList<>()).add(anImport);
+
+        // Duplicate imports do work out of the box for import, star-import and static-import.
```
Laurens-W, line 164: either of these comments needs to be negated

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -159,8 +160,51 @@ public G.CompilationUnit visit(SourceUnit unit, ModuleNode ast) throws GroovyPar
...
+        // For static-star-import, this does work though.
+        // The groovy compiler does only save the last duplicate import instead of all, so parse all static star imports by hand.
+        Map<String, ImportNode> staticStarImports = ast.getStaticStarImports();
+        if (!staticStarImports.isEmpty()) {
+            // Take source code until last static star import for performance reasons
+            int lastLineNumber = -1;
+            for (ImportNode anImport : ast.getStaticStarImports().values()) {
+                lastLineNumber = Math.max(lastLineNumber, anImport.getLastLineNumber());
+            }
+            String importSource = sourceLineNumberOffsets.length <= lastLineNumber ? source : source.substring(0, sourceLineNumberOffsets[lastLineNumber]);
```
Laurens-W, line 174: Do we have an idea of how often the false condition is hit here? I'm not sure if we can use something other than substring

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -159,8 +160,51 @@ public G.CompilationUnit visit(SourceUnit unit, ModuleNode ast) throws GroovyPar
...
+            String[] lines = importSource.split("\n");
+            for (int i = 0; i < lines.length; i++) {
+                String line = lines[i];
+                int index = 0;
+
+                while (index < line.length()) {
+                    int importIndex = line.indexOf("import", index);
+                    if (importIndex == -1) break;
+
+                    int maybeStaticIndex = indexOfNextNonWhitespace(importIndex + 6, line);
```
Laurens-W, line 186: Do we want to apply cursoring instead of incrementing indexes when they are used?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: > Do we have an idea of how often the false condition is hit here?

It will most likely never happen, because it happens when the last line of the source file is a static star import. For instance:

```groovy
package org.openrewrite.groovy.tree

import org.junit.jupiter.api.Test
import org.openrewrite.java.tree.Space
import org.openrewrite.test.RewriteTest

import static org.assertj.core.api.Assertions.assertThat
import static org.openrewrite.groovy.Assertions.groovy
<<EOF>>
```

You wouldn't find this code out there that much (who would ever push such a file 😆). But it's better to just support it instead of ignoring this case.

> I'm not sure if we can use something other than substring
I don't know, I created this kind of code to just only parse the imports instead of the whole file for performance reasons. If you know a faster way, I am glad to hear it!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: **No**, because this whole block of code is actually just a groovy-compiler-fix, by creating all the static-star-import-nodes ourselves (so the missing duplicate nodes are now present as well). Then we handle the ImportNodes at a later [stage](https://github.com/openrewrite/rewrite/blob/cc2c498c07adff86078d5b63c7edbe81c8d68646/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L2222).

**But** now you write this, I am thinking this through. We could go one step further. Above the import-node stuff, there is:

```java
if (ast.getPackage() != null) {
  prefix = whitespace();
  skip("package");
  pkg = maybeSemicolon(new J.Package(randomId(), EMPTY, Markers.EMPTY, typeTree(null), emptyList()));
}
```

We could just go through the import code ourselves and parse everything here. Then we don't need all this extra magic here, because we can cursor through the source file until there are no more imports left. The the parsen of the Groovy ImportNode stuff can go too, because then we are doing everything ourselves.

What do you think @Laurens-W? It would cost me a couple of more hours to get it right. Would it be worth it?

Laurens-W: Sounds like a worthwhile effort! Makes the parser flow more naturally

jevanlingen: Tried it out, but it's very hard to do, because imports can be everywhere in a Groovy [script](https://groovy-lang.org/structure.html#_scripts_versus_classes). For example, this is perfectly valid:

```groovy
import java.nio.file.Paths
println "hallo"
import java.io.File
println "hallo"
```

I will do some changes though, to improve the flow of the program.

Laurens-W (approved):
LGTM

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/ImportTest.java
```diff
@@ -90,4 +101,21 @@ void staticImportAlias() {
           )
         );
     }
+
+    @Test
+    void duplicateImports() {
+        rewriteRun(
+          groovy(
+            """
+                    import static      java.util.Collections.*    ;     import               static      java.util.Collections.*
```
knutwannheden, line 110: Note: Adding a commented out static import line (like `// import static foo.*`) causes an NPE.

jevanlingen: O boy, always those pesky comments. I'll create a follow up PR this fix this!

jevanlingen: Done ⬇️

PR #4985, opened 2025-02-05
Add null check for AddDependency

SiBorea:
## What's changed?
Add null check in `AddDependency`. The dependencies could be null in some cases, particularly in unit test.

### Checklist


timtebeek (approved):
Thanks @SiBorea ! Not sure in exactly which scenarios this would be null beyond tests, but I'll trust that you've seen this sufficiently to add the check here.

PR #4986, opened 2025-02-05
Replace `NavigableMap<LineColumn, List<ASTNode>>` to `NavigableMap<LineColumn, ASTNode>`, because it's impossible to have multiple elements at the very same line and column

jevanlingen:
## What's changed?
Little code refactoring; so no new features or bug fixes.

## What's your motivation?
When I worked on https://github.com/openrewrite/rewrite/issues/4068, I looked at the

```java
NavigableMap<LineColumn, List<ASTNode>> sortedByPosition = new TreeMap<>();` 
```

line. 

After thinking about it, I wondered how for any key there can be multiple nodes attached to it. Since the key is an object of line number and column (basically the x and y of the source code), there is absolutely no way multiple elements can be found, unless I am very much mistaken.

Thus I replaced the list in favour of one ASTNode object.


Laurens-W (approved):
LGTM

PR #4987, opened 2025-02-05
Use `sourceStartsWith` to simplify visitClass

jevanlingen:
## What's changed?
The `visitClass` method is written a little shorter

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -237,24 +237,22 @@ public class GroovyParserVisitor {
 
             Space kindPrefix = whitespace();
             J.ClassDeclaration.Kind.Type kindType = null;
-            if (source.startsWith("class", cursor)) {
+            if (sourceStartsWith("class")) {
                 kindType = J.ClassDeclaration.Kind.Type.Class;
                 skip("class");
-            } else if (source.startsWith("interface", cursor)) {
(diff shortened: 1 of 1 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -237,38 +237,31 @@ public void visitClass(ClassNode clazz) {
...
             JLeftPadded<TypeTree> extendings = null;
             if (kindType == J.ClassDeclaration.Kind.Type.Class) {
-                int saveCursor = cursor;
-                Space extendsPrefix = whitespace();
-                if (source.startsWith("extends", cursor)) {
-                    skip("extends");
-                    extendings = padLeft(extendsPrefix, visitTypeTree(clazz.getSuperClass()));
-                } else {
-                    cursor = saveCursor;
+                if (sourceStartsWith("extends")) {
```
Laurens-W, line 263: Does this take into account whitespace and potentially resetting the cursor? If yes this comment is a non issue

Laurens-W (approved):
Other then the 1 potential comment LGTM

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: It sure does 😄, from the JavaDoc:

> Tests if the source beginning at the current cursor starts with the specified delimiter. Whitespace characters are excluded, the cursor will not be moved.

PR #4988, opened 2025-02-05
Make `UpgradeDependencyVersion` recipe runnable with defined repositories in the buildscript only

jevanlingen:
## What's changed?
- The `UpgradeDependencyVersion` recipe can run with defined repositories in the buildscript only
- Refactoring of the `UpdateVariable` class in the `UpgradeDependencyVersion` recipe
- Update of the ` DefaultImportsCustomizer`

## What's your motivation?
The `UpgradeDependencyVersion` recipe did not work with a setup like:

```groovy
buildscript {
    ext {
        guavaVersion = "29.0-jre"
    }
    repositories {
        mavenCentral()
    }
    dependencies {
        classpath("com.google.guava:guava:${guavaVersion}")
    }
}
```

Why: [click](https://github.com/openrewrite/rewrite/pull/4988#issuecomment-2639161921).


rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/DefaultImportsCustomizer.java
jevanlingen, line 1: This file actually has nothing to do with the rest of my changes (when I was researching the problem, I thought it did). But I left these changes in because they are also worthwhile.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -373,18 +373,6 @@ void upgradesVariablesDefinedInExtraProperties() {
         rewriteRun(
           buildGradle(
             """
-              buildscript {
```
Laurens-W, line 376: I'd remove this part from the new test and potentially change the names of the test methods, so this diff becomes minimal

jevanlingen: No, I did this on purpose.
- One has an `ext` (that is Extra Properties) in the root of the Gradle file
- The other has the `ext` in the `buildscript`

The first case did already work, the second case did not.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -548,107 +548,75 @@ private class UpdateVariable extends GroovyIsoVisitor<ExecutionContext> {
...
+                    return null;
+                }
+                getCursor().dropParentUntil(p -> p instanceof SourceFile)
+                        .computeMessageIfAbsent(NEW_VERSION_KEY, m -> new HashMap<GroupArtifactVersion, Set<String>>())
+                        .computeIfAbsent(new GroupArtifactVersion(ga.getGroupId(), ga.getArtifactId(), selectedVersion), it -> new HashSet<>())
+                        .addAll(gaEntry.getValue());
+
+                newVersionLiteral = ChangeStringLiteral.withStringValue(literal, selectedVersion);
+            }
+            return newVersionLiteral;
```
timtebeek, line 619: Can we simplify this to the following? I suspect the default IntelliJ extract method was involved and made this slightly odd.
```suggestion
            for (Map.Entry<GroupArtifact, Set<String>> gaEntry : gaToConfigurations.entrySet()) {
                GroupArtifact ga = gaEntry.getKey();
                GroupArtifactVersion gav = new GroupArtifactVersion(ga.getGroupId(), ga.getArtifactId(), version);
                String selectedVersion = new DependencyVersionSelector(metadataFailures, gradleProject, null)
                        .select(gav, null, newVersion, versionPattern, ctx);
                if (selectedVersion == null) {
                    return null;
                }
                getCursor().dropParentUntil(p -> p instanceof SourceFile)
                        .computeMessageIfAbsent(NEW_VERSION_KEY, m -> new HashMap<GroupArtifactVersion, Set<String>>())
                        .computeIfAbsent(new GroupArtifactVersion(ga.getGroupId(), ga.getArtifactId(), selectedVersion), it -> new HashSet<>())
                        .addAll(gaEntry.getValue());

                return ChangeStringLiteral.withStringValue(literal, selectedVersion);
            }
            return null;
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyVersionSelector.java
```diff
@@ -188,8 +191,10 @@ private List<MavenRepository> determineRepos(@Nullable String configuration) {
...
             throw new IllegalStateException("Gradle project must be set to determine repositories."); // Caught by caller
         }
-        return "classpath".equals(configuration) ?
-                gradleProject.getBuildscript().getMavenRepositories() :
-                gradleProject.getMavenRepositories();
+        if ("classpath".equals(configuration)) {
+            return gradleProject.getBuildscript().getMavenRepositories();
+        }
+        List<MavenRepository> repos = gradleProject.getMavenRepositories();
+        return repos.isEmpty() ? gradleProject.getBuildscript().getMavenRepositories() : repos;
```
shanman190, line 198: The logic here is invalid from a Gradle standpoint.

Buildscript should be used for plugins and the `classpath` configuration while project repositories should be used for project configurations. The logic here that does the fallback to the buildscript repositories when project repositories is empty does not happen internally in Gradle, so I would recommend not doing it here either.

shanman190:
Overall, I'm a little puzzled with this PR... In the description it seems like an extension variable defined in the buildscript block fails to update the version contained within the variable, but there is explicitly the [UpgradeDependencyVersionTest#upgradesVariablesDefinedInExtraProperties](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Ftest%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FUpgradeDependencyVersionTest.java#L371-L435) test that verifies extension properties both on the buildscript as well as the project are updated.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
jevanlingen: No, I did write it on purpose like that, as the [original](https://github.com/openrewrite/rewrite/blob/22179f546227ba3c3a8202675aaeca627c324854/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java#L591) code did do this:

```java
try {
    for (Map.Entry<GroupArtifact, Set<String>> gaEntry : gaToConfigurations.entrySet()) {
        // ..
        if (selectedVersion == null) { return null }
        // ..
   
        J.Literal newVersionLiteral = ChangeStringLiteral.withStringValue(initializer, selectedVersion);
        v = v.withInitializer(newVersionLiteral);
    }
    return e.warn(v);
}

return v;
```

As you can see, it overrides the `v` variable every time a new entry can be applied. Maybe it's a bug, maybe it's not, but I did not want to change this behaviour.

jevanlingen:
> Overall, I'm a little puzzled with this PR... In the description it seems like an extension variable defined in the buildscript block fails to update the version contained within the variable, but there is explicitly the [UpgradeDependencyVersionTest#upgradesVariablesDefinedInExtraProperties](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Ftest%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FUpgradeDependencyVersionTest.java#L371-L435) test that verifies extension properties both on the buildscript as well as the project are updated (notice that two different variable names were used explicitly to guarantee one of them was not accidentally updated as an inadvertent side effect).

Ah, I should have given a little more context in the motivation part of this PR. Well, let's work it out here. Yes there was a test, but it should have been two tests. So the original test looked like:

```java
@Test
void upgradesVariablesDefinedInExtraProperties() {
    rewriteRun(
      buildGradle(
          """
          buildscript {
              ext { guavaVersion = "29.0-jre" }
              repositories { mavenCentral() } // <-- repositories in extension of build script
              dependencies { classpath("com.google.guava:guava:${guavaVersion}") }
          }

          plugins { id "java" }

          repositories { mavenCentral() } // <-- repositories in root

          ext { guavaVersion2 = "29.0-jre" }

          dependencies { implementation "com.google.guava:guava:${guavaVersion2}" }
          """,
          """
          <upgraded gradle script>
          """
      )
    );
}
```

The problem here is the `buildscript` extension is not totally tested. The `repositories` part is taken from the root. If you would define another test (which I did in this very PR) with a Gradle file containing **only** a buildscript:

```java
@Test
void upgradesVariablesDefinedInExtraPropertiesWithBuildscript() {
    rewriteRun(
      buildGradle(
          """
          buildscript {
              ext { guavaVersion = "29.0-jre" }
              repositories { mavenCentral() } // <-- repositories in extension of build script
              dependencies { classpath("com.google.guava:guava:${guavaVersion}") }
          }
          """,
          """
          <upgraded gradle script>
          """
      )
    );
}
```

You would end up with following error:

```groovy
"buildscript {
    ext { /*~~(com.google.guava:guava failed. Unable to download metadata.)~~>*/guavaVersion = "29.0-jre" }
    repositories { mavenCentral() } // <-- repositories in extension of build script
    dependencies { classpath("com.google.guava:guava:${guavaVersion}") }
}"
```

The reason this happens is because the retrieval of the repositories was done in two steps:

```java
try {
  // ..
  String selectedVersion = Optional.ofNullable(selector.select(gav, null, newVersion, versionPattern, ctx))
    .orElse(selector.select(gav, "classpath", newVersion, versionPattern, ctx));
  // ..
```

The idea behind it: first use a `null` configuration, if that lacks any results, try it with the "classpath" configuration. This does not work though, because the first `select` will already throw a `MavenDownloadingException`, so the next `select` will never be run.

That why I moved that idea to the `DependencyVersionSelector`:
```
return repos.isEmpty() ? gradleProject.getBuildscript().getMavenRepositories() : repos;
```

But as you mentioned a couple hours ago, this is invalid from a Gradle standpoint. So I'll revert my changes to the DependencyVersionSelector and change the recipe a little more (putting back the select(null) => select("classpath"), but without the bug).

Thanks for the review, my Gradle knowledge is not that big. So really nice to have a review from someone that knows Gradle very well!

rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyVersionSelector.java
jevanlingen: ⬇️

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
jevanlingen: After thinking this through, I think you are totally right. It did not make sense in the first place (if there were to be multiple items in that map, the latest would always be taken). So I rewrote it a little different by using a tuple.

shanman190:
Aha! That makes a lot more sense as to the where the bug in behavior comes from. Thanks for clarifying that context!

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -548,107 +548,78 @@ private class UpdateVariable extends GroovyIsoVisitor<ExecutionContext> {
...
+            DependencyVersionSelector dependencyVersionSelector = new DependencyVersionSelector(metadataFailures, gradleProject, null);
+            GroupArtifactVersion gav = new GroupArtifactVersion(ga.getGroupId(), ga.getArtifactId(), (String) literal.getValue());
+
+            String selectedVersion;
+            try {
+                selectedVersion = dependencyVersionSelector.select(gav, null, newVersion, versionPattern, ctx);
             } catch (MavenDownloadingException e) {
-                return e.warn(a);
+                // try again with "classpath" configuration; if this one fails as well, bubble error up so it can be handled
+                selectedVersion = dependencyVersionSelector.select(gav, "classpath", newVersion, versionPattern, ctx);
```
shanman190, line 611: We should probably check that the configurations list contain "classpath" prior to running this for all failing cases.

jevanlingen: Good one, will work on this a little more tomorrow 😄.

shanman190: Another little comment here would be wouldn't we want to continue adding an error marker when things truly fail to resolve? That way users can see and know what went wrong?

jevanlingen: > Another little comment here would be wouldn't we want to continue adding an error marker when things truly fail to resolve? That way users can see and know what went wrong?

It does continue how it did work 😉. The first `select` is catched, so the second `select` can be called. The second `select` is not catched, so in the end it works the same. I even made sure this work, by adding the new recipe `cannotDownloadMetaDataWhenNoRepositoriesAreDefined`:

```java
@Test
void cannotDownloadMetaDataWhenNoRepositoriesAreDefined() {
  rewriteRun(
    buildGradle(
      """
        plugins { id 'java-library' }
        dependencies { implementation "com.google.guava:guava:29.0-jre" }
      """,
      """
        plugins { id 'java-library' }
        dependencies { /*~~(com.google.guava:guava failed. Unable to download metadata.)~~>*/implementation "com.google.guava:guava:29.0-jre" }
      """
  ));
}
 ```   

There is no date table, we could opt to add that as well. If we want that, I'll suppose we should just create a separate PR.

shanman190: So to talk about some infrastructure setup for a moment, the Gradle default plugin repository `gradlePluginPortal()` will accept a request for an artifact, if it's found it'll return it. If it's not, then it forwards that request to `mavenCentral` behind the scenes.

When we talk about artifact resolution, you're well on the right path.

* Plugins
  * Plugin requests only use the `buildscript.repositories` (`pluginManagement.repositories` from the settings.gradle is pushed into this list by Gradle automatically for us as well).
* Project dependencies
  * External dependency requests only use the `repositories` (`dependencyResolutionManagement.repositories` from the settings.gradle is pushed into this list by Gradle automatically for us as well).

Why this little section is confusing us because we're updating variables outside of the scope of exactly where the dependency is declared or where the variable is declared. Part of this is extension properties (`ext`, `buildscript.ext`) and part of it is project properties (gradle.properties) -- and here in the near-term future even `libs.versions.toml`. When we are updating the dependency as a Literal or G.MapEntry, we always do make sure to take into account the call site, but using the "classpath" configuration as the indicator.

jevanlingen: Thx for your comment (had a few things to look after, so now I got time to come back to this issue)! So I was kinda right the recipe does not really follow the infrastructure setup of Gradle. I don't know if that's really a harm though. Do you think this PR is good as it is, or do you think it still needs more changes?

shanman190: Now with the contains classpath, we're at least minimizing our deviation away from Gradle's internal infrastructure to a minimum enough surface area that I think it's acceptable to proceed. Thanks for the changes that you've made to this effect!

shanman190 (approved):

PR #4990, opened 2025-02-05
Support duplicate star imports with comments

jevanlingen:
## What's changed?
Duplicate star imports can also have comments.

## What's your motivation?
- https://github.com/openrewrite/rewrite/pull/4982


Laurens-W:
Maybe we should add a testcase for regression purposes? Other than that I think it looks good

PR #4991, opened 2025-02-06
Disable parentheses level discovery for method calls for groovy 2

jevanlingen:
## What's changed?
Disable parentheses level discovery for methods for Groovy 2.

## What's your motivation?
Groovy 2 does not contain the information we need, it always return -1 for the method calls object-expression. In older version of the groovy-parser this did not cause any exceptions, but since
- https://github.com/openrewrite/rewrite/pull/4951

it does.

Notice before PR #4951, the parser did also not support parentheses level discovery for methods for Groovy 2, but I did not throw an errors by chance.

## Any additional context
Would  be nice to have a real working solution for Groovy 2 as well. But let's focus on that at a later stage.


timtebeek (approved):
Great to have this fixed ahead of the release, thanks!

PR #4994, opened 2025-02-06
Skip GitProvenance for bare repositories

timtebeek:
## What's changed?
Do not create `GitProvenance` for bare git repositories.

## What's your motivation?
We need a working tree to be able to process projects; a lack thereof should not proceed further.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/GitProvenance.java
@@ -247,7 +247,10 @@ public class GitProvenance implements Marker {
         }
     }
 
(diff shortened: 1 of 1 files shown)


okundzich:
what are bare git repositories? i worry about omitting git provenance

timtebeek:
Bare git repositories are repositories intended for folks to push and pull from, but that themselves lack a working tree. It would be rare for folks to encounter those at all, let alone attempt to run against them, but with the discovery we do with the CLI it's possible to stumble across these, which are then best skipped since no working tree means there are no sources to parse. Hence why I figured it best to not have git provenance for these cases, such that they are skipped automatically, as opposed to break further down the line.

bryceatmoderne (approved):

timtebeek:
Given that this is unlike to appear in practice, with the explanation above, I'd think we're good to add this now. If any late objections arrive we can revert again.

Issue #4995, opened 2025-02-06
ChangePackage doesn't handle classes with GWT JNSI methods


auke-:
## What version of OpenRewrite are you using?

I am using OpenRewrite v8.44.2 with the [ChangePackage](https://docs.openrewrite.org/recipes/java/changepackage) recipe.

## What is the smallest, simplest way to reproduce the problem?

```yaml
type: specs.openrewrite.org/v1beta/recipe
name: com.yourorg.ChangePackageExample
displayName: Rename package name example
recipeList:
  - org.openrewrite.java.ChangePackage:
      oldPackageName: old_package_name
      newPackageName: new_package_name
```

```java
package old_package_name;
class A {
  native void alert(String msg) /*-{ $wnd.alert(msg); }-*/;
}
```

## What did you expect to see?

```java
package new_package_name;
class A {
  native void alert(String msg) /*-{ $wnd.alert(msg); }-*/;
}
```

## What did you see instead?
OpenRewrite didn't touch the file. It seems like it ignores Java classes with `native` methods and/or trailing comment.

See [reproducer](https://github.com/openrewrite/rewrite/pull/4996)

timtebeek:
Thanks for the runnable report! Feel free to open up a draft PR; that makes it easier to pull into an IDEA and explore. We're not too strict on requiring an issue or working fix first.

auke-:
Hmm... looks i've made some errror on my side, the reproducer isn't correct.

auke-:
We did encounter issues with the ChangePackage recipe and it looked like is had to to with native methods. Will try to create a proper reproducer tomorrow. Do you have another channel for quick questions, besides `issues` and `discussions` in Github?

timtebeek:
Sure, always welcome on [our Slack](https://join.slack.com/t/rewriteoss/shared_invite/zt-nj42n3ea-b~62rIHzb3Vo0E1APKCXEA)! :)

auke-:
An even smaller reproducer shows that the trailing comment is causing the issue:
```java
import org.openrewrite.Foo;
interface Test {
  void foo() /* */;
}
```

timtebeek:
Thanks for the helpful reproducer! Indeed something to fix in the parser then it seems. Did you already explore that side?

auke-:
No, i didn't look into the parser side yet and i probably don't have time until after the weekend.

auke-:
[...]

PR #4412, opened 2024-08-13
Handle erroneous nodes in open rewrite

vudayani:
## What's changed?

The PR introduces handling for erroneous nodes in OpenRewrite. Currently, when there is an error in the LST, the parser may attempt to recover by ignoring or removing the erroneous node to produce a valid LST. The changes in this PR address this by adding a mechanism to handle these erroneous nodes instead of ignoring them, ensuring that they are preserved and correctly processed.

## What's your motivation?

The motivation behind this PR is to handle scenarios where errors in the source code could lead to discrepancies in the computed results. When any recipe is executed, the edits are computed on a parsed LST without the error statement. By handling erroneous nodes, we ensure that the error lines are retained and reflected correctly in the LST, leading to more accurate edits and transformations.

## Any additional context

This update is crucial for ensuring consistency in scenarios where error lines could affect the outcome of recipes and transformations.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #4996, opened 2025-02-07
fix: handle comment at the end of a method declaration

auke-:
Multi line comment is allowed at almost any part of the Java source code
where whitespace is allowed. The reloadable Java parser visitor however
swallows multi line comment at the end of a method declaration, which
happens to be a common pattern in GWT in order to embed JavaScript in a
Java class.

- Note that this effectively is a partial revert of https://github.com/openrewrite/rewrite/pull/4412
(commit https://github.com/openrewrite/rewrite/commit/ac32851e09c9b76aed56c898844e217fa9e61367).

- Fixes https://github.com/openrewrite/rewrite/issues/4995

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
@@ -1840,9 +1840,7 @@ public class ReloadableJava11ParserVisitor extends TreePathScanner<J, Space> {
         if (t instanceof JCMethodDecl) {
             JCMethodDecl m = (JCMethodDecl) t;
             if (m.body == null || m.defaultValue != null) {
-                String suffix = source.substring(cursor, positionOfNext(";", null));
-                int idx = findFirstNonWhitespaceChar(suffix);
-                return sourceBefore(idx >= 0 ? "" : ";");
+                return sourceBefore(";");
             } else {
                 return sourceBefore("");
             }
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
@@ -1920,9 +1920,7 @@ public class ReloadableJava17ParserVisitor extends TreePathScanner<J, Space> {
             case METHOD:
                 JCMethodDecl m = (JCMethodDecl) t;
                 if (m.body == null || m.defaultValue != null) {
-                    String suffix = source.substring(cursor, positionOfNext(";", null));
-                    int idx = findFirstNonWhitespaceChar(suffix);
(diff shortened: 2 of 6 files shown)


timtebeek (approved):
Appreciate the help and patience here @auke- ! This is looking good on our end. We'll wait for the tests to clear and then we can merge.

timtebeek:
Turns out this was indeed needed for those erroneous nodes you had helpfully linked above, thanks!

```
JavaParserTest > erroneousExpressionStatements(String) > [3] source=package com.example.demo;
class FooBar {
    public void test(int num string s, int b) {
      String a; this.ownerR
      System.out.println();
    }
}
 FAILED
    org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "com/example/demo/FooBar.java":
```
```diff
diff --git a/com/example/demo/FooBar.java b/com/example/demo/FooBar.java
index 5028cf3..7e634ec 100644
--- a/com/example/demo/FooBar.java
+++ b/com/example/demo/FooBar.java
@@ -1,7 +1,7 @@ 
 package com.example.demo;
 class FooBar {
     public void test(int num string s, int b) {
-      String a; this.ownerR
+      String a this.ownerR
       System.out.println();
     }
 }
\ No newline at end of file
```
```
    ] 
    expected: 
      "package com.example.demo;
      class FooBar {
          public void test(int num string s, int b) {
            String a; this.ownerR
            System.out.println();
          }
      }"
     but was: 
      "package com.example.demo;
      class FooBar {
          public void test(int num string s, int b) {
            String a this.ownerR
            System.out.println();
          }
      }"
        at app//org.openrewrite.test.RewriteTest.assertContentEquals(RewriteTest.java:618)
        at app//org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:308)
        at app//org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:129)
        at app//org.openrewrite.java.JavaParserTest.erroneousExpressionStatements(JavaParserTest.java:242)
```
/cc @knutwannheden

knutwannheden:
I wonder if we shouldn't remove that test case for now, which attempts to parse erroneous code like this:

```java
public void test(int num string s, int b) {
  String a; this.ownerR
  System.out.println();
}
```

While it would be nice to not fail to parse code like this, it seems like it would be higher priority to parse correct code correctly.

@vudayani, @BoykoAlex : Maybe you can comment on which erroneous input we need to be able to successfully parse into an LST?

timtebeek:
Had another look and adjusted the tests: turns out it's a parameterized test that tests a couple of different cases, but the failing case had two separate compilation issues: one in the arguments, and then another with a missing field reference and semicolon. I've separated those cases as I believe the goal here was to support completions at the point where the user has their cursor, thus most closely matching a single erroneous node at that position. With that I think we can merge this already, and revise if needed asynchronously without indeed breaking valid usage. Thanks all!

PR #4998, opened 2025-02-07
Use `pattern` operator for all possible string kinds

jevanlingen:
## What's changed?
The pattern operator is supported for all possible string kinds. 

## What's your motivation?
The parser did only support the pattern operator for slashy strings (`~/<pattern>/)`. Though by convention slashy strings are used most for the patten operator, you can use every kind of string: 

> The pattern operator (~) provides a simple way to create a java.util.regex.Pattern instance. While in general, you find the pattern operator with an expression in a slashy-string, it can be used with any kind of String in Groovy - [docs](https://groovy-lang.org/operators.html#_pattern_operator)

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/LiteralTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/LiteralTest.java
@@ -40,9 +40,46 @@ class LiteralTest implements RewriteTest {
     }
 
     @Test
-    void string() {
+    void singleQuoteString() {
         rewriteRun(
-          groovy("'hello'")
+          groovy(
+            """
+              'hello'
+              """
+          )
+        );
+    }
+
+    @Test
+    void regexPatternSingleQuoteString() {
+        rewriteRun(
+          groovy(
+            """
+              ~"hello"
+              """
+          )
+        );
+    }
+
+    @Test
+    void doubleQuoteString() {
+        rewriteRun(
+          groovy(
+            """
+              "hello"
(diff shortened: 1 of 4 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/LiteralTest.java
```diff
@@ -40,9 +40,46 @@ void insideParentheses() {
     }
 
     @Test
-    void string() {
+    void singleQuoteString() {
         rewriteRun(
-          groovy("'hello'")
+          groovy(
```
MBoegers, line 45: please add `//language=groovy` to enable hightlights for everyone

MBoegers (approved):
Just quickly scanned, I have no big parser knowledge

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/LiteralTest.java
jevanlingen: There should not be a need, because we use it in our source code:

```java
public static SourceSpecs groovy(@Language("groovy") @Nullable String before) {
 ..
}
```

Sometimes the hightlighing does work do and we also use `//language=groovy`. I think it's an IntelliJ bug if it does not work, for [both](https://www.jetbrains.com/help/idea/using-language-injections.html#use-language-injection-comments) options are supported. As we don't use `//language=groovy` in the rewrite-groovy module, I don't want to add it here.

PR #5000, opened 2025-02-07
Add dependencies in JVMTestSuite dependency block

MBoegers:
## What's changed?

If a dependency is added in a dependencies block inside a not standard sourceSet, the prefix mechanism in [org.openrewrite.gradle.AddDependency Line 180](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java#L180) produces unexpected namings.

### Checklist


rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1345,6 +1303,58 @@ void addUnconditionally() {
         );
     }
 
+    @Test
+    @Issue("https://github.com/moderneinc/customer-requests/issues/792")
+    void addScoped() {
```
shanman190, line 1308: Note that this is actually a combination of two distinct features. You've got a custom source set (which the configurations are created at the project level also), but the JVM Test Suites feature as well.

`AddDependency` hasn't yet been adapted to understand how to add dependencies to JVM Test Suite instances yet, so it's going to attempt to add the dependency like below which would still work, but doesn't match the test assertions.

```
dependencies {
    integrationTestImplementation "com.google.guava:guava:29.0-jre"
}
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1360,4 +1370,41 @@ private AddDependency addDependency(String gav, @Nullable String onlyIfUsing, @N
           (gavParts.length < 4) ? null : gavParts[3], null, null, null
         );
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"com.google.common.math.*", "com.google.common.math.IntMath"})
+    void onlyIfUsingTestScope(String onlyIfUsing) {
```
shanman190, line 1376: Was this test just being logically moved to make it easier to see with respect to the new test?

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
shanman190: I'd probably rename this test to something like `addToJVMTestSuite` which more clearly aligns to the assertion.

MBoegers: Thanks for clearing whats used here, that will help implementing. Its true `AddDependency` doesn't work as the expextation.
Is the nested dependency block a JVMTestSuite specific feature or possible in other combinations?

MBoegers: Thanks for clearing whats used here, that will help implementing. Its true `AddDependency` doesn't work as the expextation.
Is the nested dependency block a JVMTestSuite specific feature or possible in other combinations?

shanman190: There are four main forms of declaring dependencies which are nested at a deeper level than normal that I can recall.

* `buildscript`: to declare plugins; `plugins` blocks are preferred in modern versions
* `allprojects`: discouraged in modern versions and breaks configuration cache/project isolation
* `subprojects`: also discouraged in modern versions and breaks configuration cache/project isolation
* JVM Test Suites (as shown here)

Of course programmatically adding dependencies via the configuration is possible as well, but I'm leaving that off since we're really talking about DSL forms.

Since we handle plugins separately, so far we've ignored `buildscript` as well. Which leaves the last three above remaining, two of which are discouraged (and also complicated to deal with with respect to OpenRewrite's LST marker shape).

With JVM Test Suites, it's really wrapping up the historic logic that was necessary to create a source set for the tests to reside in -- which then inherently created the necessary configurations -- and tries to go a step further to encapsulate the test suite itself into a more use case specific and succinct form.

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
MBoegers: TBH idk

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1345,6 +1341,240 @@ void addUnconditionally() {
...
+                        id 'jvm-test-suite'
+                    }
+                    
+                    repositories {
+                        mavenCentral()
+                    }
+                    
+                    testing {
+                        suites {
+                            integrationTest(JvmTestSuite) {
```
shanman190, line 1546: A test suite that lacks dependencies originally is perfectly valid. I would think if we have a collision, then we'd probably add the dependencies block for the not present case.

shanman190 (commented):
Lots of good stuff here, @MBoegers! I've left a few comments that we can discuss about possibly implementing, but some are certainly more complex than others and we can of course handle those via future PRs should we want to take them on.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -258,7 +329,36 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
                     private boolean isTopLevel(Cursor cursor) {
                         return cursor.getParentOrThrow().firstEnclosing(J.MethodInvocation.class) == null;
                     }
+
+                    private Predicate<Cursor> isMatchingJVMTestSuite(String resolvedConfiguration) {
+                        return cursor -> {
+                            String sourceSet = resolvedConfiguration.substring(0, resolvedConfiguration.length() - 14);
```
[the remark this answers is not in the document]
MBoegers: Good point, we could make it a field and give it a name or rely on the compiler to infer that `"Implementation".lenght()` is constant.

Regarding the other way around, somthing like:
´´´java
J.MethodInvocation methodInvocation = cursor.getParentOrThrow().firstEnclosing(J.MethodInvocation.class);
return methodInvocation != null && resolvedConfiguration.startsWith(methodInvocation.getSimpleName())
´´´
This would work, but I like to stick with equals. lets precalcualte the sourceSet name and keep it. 
Suggestion:
```java
private Predicate<Cursor> isMatchingJVMTestSuite(String resolvedConfiguration) {
    String sourceSet = resolvedConfiguration.substring(0, resolvedConfiguration.length() - "Implementation".lenght());  
     return cursor -> {
         J.MethodInvocation methodInvocation = cursor.getParentOrThrow().firstEnclosing(J.MethodInvocation.class);
          return methodInvocation != null && sourceSet.equals(methodInvocation.getSimpleName());
      };
}
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1345,6 +1341,240 @@ void addUnconditionally() {
...
+                        mavenCentral()
+                    }
+                    
+                    dependencies {
+                        implementation "org.apache.logging.log4j:log4j-core:2.22.1"
+                    }
+                    
+                    testing {
+                        suites {
+                            test(JvmTestSuite) {
```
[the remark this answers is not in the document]
MBoegers: So the while test `actNormalForDefaultTestWithSuiteDefinition` is obsolete because not possible?

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
MBoegers: So you say it's more convenient to add dependencies to the `testing.suites.XYZ.dependencies` block than adding it to the global one?
If so, we have to expand the `AddDependencyVisitor`s logic.

MBoegers:
@shanman190 thanks for educating me here 🙏

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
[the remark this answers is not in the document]
shanman190: So there is a different method invocation that is called instead.
https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#sec:configuring_the_built_in_test_suite

The premise remains the same in that dependencies could be declared within the built-in `test` suite or at the root. We probably just need a little specialized logic to handle the `test` source set.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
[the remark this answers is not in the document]
shanman190: I feel like the problem is relying solely on `implementation` as the value/length. As you've seen with the default configuration list, Gradle follows a pattern of creating several configurations using a convention of the source set name being the prefix. With the defaults for the recipe, we will always drop the dependency into the `implementation` configuration, but the recipe does allow the ability to use a user-defined configuration instead. If the user happens to choose any of the non-`implementation` configurations, then this block will react incorrectly.

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
shanman190: This is the one where I was saying the issue is more complex and we may choose to solve this with a future PR.

In my mind, I'm trying to think about what would the user expect to happen and then we would align to that. I do agree that it's certainly easier to put the dependency at the root. And furthermore, Gradle doesn't provide any strong recommendation that I could find on using one versus the other.

For now though, I think the logic that is present is probably good enough to progress with, but I just wanted to mention it here as well.

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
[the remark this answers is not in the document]
MBoegers: We could solve it the same way as https://github.com/openrewrite/rewrite/pull/5000#discussion_r1952815276, add to the global dependencies block as `testImplementation` and wait if anybody requests this feature.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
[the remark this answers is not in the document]
MBoegers: In this case, what would be the expectation?

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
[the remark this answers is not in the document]
shanman190: I think that's probably acceptable for now as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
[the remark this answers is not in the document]
MBoegers: see, [new test](https://github.com/openrewrite/rewrite/pull/5000/files#diff-cae1d1ef7a9763533b28bd4c874f57d45a4256b044696efb6a378439731e9a1bR1521). This test is not possible at all because `AddDependency` ATM ignores the `SourceSet` if the configuration is set not `null`.
Its done [here](https://github.com/openrewrite/rewrite/pull/5000/files#diff-743cf2f048fdba39228e626dd60acbdf76b3f256061b826bffd26106e95338abL220).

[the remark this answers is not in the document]
MBoegers: It's not working in the original version, which seems like a more general problem here.
Let's create an issue to document and work on this as follow up.

```java
@Test
    void configurationAndSourceSetCombination() {
        rewriteRun(
          spec -> spec.recipe(addDependency("com.google.guava:guava:29.0-jre", null, "compileOnly")),
          mavenProject("project",
            srcTestJava(
              java(usingGuavaIntMath, spec -> sourceSet(spec, "anotherSet"))),
            buildGradle(
              """
                plugins {
                    id 'java'
                }
                
                repositories {
                    mavenCentral()
                }
                """,
              """
                plugins {
                    id 'java'
                }
                
                repositories {
                    mavenCentral()
                }
                
                dependencies {
                    anotherSetCompileOnly "com.google.guava:guava:29.0-jre"
                }
                """
            )
          )
        );
    }
```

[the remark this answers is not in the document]
MBoegers: I've created a follow-up to fix the problem more general.

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1345,6 +1342,284 @@ void addUnconditionally() {
...
+                )
+              )
+            );
+        }
+
+        @ParameterizedTest
+        @ValueSource(strings = {"implementation", "compileOnly", "runtimeOnly", "annotationProcessor"})
+        void jvmTestSuitePlugin(String configuration) {
+            rewriteRun(
+              spec -> spec.recipe(addDependency("com.google.guava:guava:29.0-jre", null, configuration)),
```
shanman190, line 1564: The `configuration` option here should be accepting a value that is prefixed with `integrationTest` (with the first letter of each of the values capitalized) to be correct. There won't be any special concatenation at all, but in the presence of a matching JVM Test Suite, we would remove the test suite name and lowercase the first letter to then add to the JVM Test Suite's configured test suite for that source set.

Ex:
```
configuration option : location to add dependency
integrationTestImplementation => testing.suites.integrationTest.dependencies.implementation
integrationTestCompileOnly => testing.suites.integrationTest.dependencied.compileOnly
integrationTestRuntimeOnly => testing.suites.integrationTest.dependencies.runtimeOnly
integrationTestAnnotationProcessor => testing.suites.integrationTest.dependencies.annotationProcessor
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1345,6 +1342,284 @@ void addUnconditionally() {
...
+                )));
+        }
+    }
+
+    @Disabled("ATM setting the configuration will overwrite the resolved configuration without condition")
+    @Nested
+    class WithDifferentConfigurations {
+        @ParameterizedTest
+        @ValueSource(strings = {"implementation", "compileOnly", "runtimeOnly", "annotationProcessor"})
+        void configurationAndSourceSetCombination(String configuration) {
```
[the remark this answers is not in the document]
MBoegers: Interesting, I read the [documentation](https://docs.openrewrite.org/recipes/gradle/adddependency) exactly the other way around. `Optional. A configuration to use when it is not what can be inferred from usage. Most of the time this will be left empty, but is used when adding a new as of yet unused dependency` in my understanding, the configuration value is used if no other configuration is inferable, not to overwrite. 
Do we need to change the Documentation here?

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
MBoegers: I've changed the test and mutated the implementation accordingly. see https://github.com/openrewrite/rewrite/pull/5000/files#diff-cae1d1ef7a9763533b28bd4c874f57d45a4256b044696efb6a378439731e9a1bR1526

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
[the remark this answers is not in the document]
shanman190: Ok, I see where you're going now.

So yes, when the `configuration` option is empty, the recipe will utilize the `onlyIfUsing` along with the `SourceSet` marker to generate a default named configuration. I think Tim pointed it out some where though that `srcMainJava` and `srcTestJava` both add the default for their respective source sets, so if you're wanting to test this permutation make sure to be a little careful around that aspect.

Now if the `configuration` option is not blank, then it takes precedence and the entire value is used only. In this permutation, the other filters still apply except the default named configuration will be ignored.

timtebeek (approved):
Can't say I've read all the earlier review threads; do let me know if I missed anything!

I like what you did with the separate Find recipe here. Did you intentionally leave out adding the data table row earlier?

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -258,6 +272,61 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
...
+                            return configuration.substring(0, configuration.length() - 11);
+                        } else if (configuration.endsWith("AnnotationProcessor")) {
+                            return configuration.substring(0, configuration.length() - 19);
+                        } else {
+                            return configuration;
+                        }
+                    }
+
+                    private String purgeSourceSet(@Nullable String configuration) {
+                        if (configuration == null || configuration.endsWith("Implementation")) {
```
shanman190, line 318: Historically, for SaaS we've needed to use `org.openrewrite.internal.StringUtils#isBlank` here. I don't recall if that's still the case or not.

@timtebeek, do you recall?

shanman190 (approved):
Just one comment, but great work @MBoegers!

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
timtebeek: Good to be consistent here indeed, just in case initialization is off.

MBoegers:
@timtebeek I can recall seeing an NPE, I think because using the `find`method (static context) does not initialize the data table properly, right?

timtebeek:
> @timtebeek I can recall seeing an NPE, I think because using the `find`method (static context) does not initialize the data table properly, right?

Ah yes that was solved in https://github.com/openrewrite/rewrite/pull/5000/commits/f4ecc974a5824cb0b8da8bd2ac4cc206dce936ce

PR #5002, opened 2025-02-08
Add SortYaml recipe to streamline yaml files (gitlab ci, spring boot properties...)

Jenson3210:
## What's changed?
Allowing yaml files to be sorted according to a list of input. Could help in creating a gitlab-ci.yaml order. 

Took some assumptions on the input variables -> please validate if this would be workable. Should we go for JsonPath instead? That would require us to support way more notations even. 

Not really, like mentioned I considered JsonPath but that's perhaps too elaborate. Maybe use jsonpath notations (with $) and in validation validate it's only going to be one of the 3 supported notations 

## Any additional context
The recipe supports several features in this version: 
- fixed property/path 
- Wildcards that match single property (`*`)
- Wildcards that match multiple properties (`**`)
- Property pickers (`[]`) to sort on their value instead of the mentioned property key

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/SortYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/SortYaml.java
@@ -1,0 +1,311 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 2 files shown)


jevanlingen:
⚠️ This recipe fails when the Yaml file contains comments. Though I do not suggest to add support for it (the recipe would be a lot harder to write, see the MergeYaml for example to get an idea), maybe it's a good idea to just skip yaml files with comments for now. Then at least you don't end up with files where the comments are all of a sudden on totally different lines.

Jenson3210:
I had worked locally on supporting comments, but it's really a pain with the prefixes containing all the comments. 
Perhaps a remove comments boolean input and then: 
- if true, remove the comments 
- if false, do not sort
But not a big fan of that myself (single responsability)

PR #5004, opened 2025-02-10
Groovy: Support property expression in annotation values

knutwannheden:
Support annotations like `@Foo(value = Test.CONST)` by using a custom `ResolveVisitor` which doesn't modify the AST by inlining the referenced constants in annotation attribute values.


rewrite-groovy/src/main/java/org/openrewrite/groovy/CustomResolveVisitor.java
```diff
@@ -0,0 +1,1659 @@
...
+        if (annotations.isEmpty()) return;
+        Map<String, AnnotationNode> tmpAnnotations = new HashMap<>();
+        for (AnnotationNode an : annotations) {
+            // skip built-in properties
+            if (an.isBuiltIn()) continue;
+            ClassNode annType = an.getClassNode();
+            resolveOrFail(annType, " for annotation", an);
+            for (Map.Entry<String, Expression> member : an.getMembers().entrySet()) {
+                Expression newValue = transform(member.getValue());
+                Expression adjusted = transformInlineConstants(newValue);
```
jevanlingen, line 1297: What about this one?

rewrite-groovy/src/main/java/org/openrewrite/groovy/CustomResolveVisitor.java
```diff
@@ -0,0 +1,1659 @@
...
+        ClassNode type = an.getClassNode();
+        resolveOrFail(type, " for annotation", an);
+        for (Map.Entry<String, Expression> member : an.getMembers().entrySet()) {
+            member.setValue(transform(member.getValue()));
+        }
+        return ace;
+    }
+
+    @Override
+    public void visitAnnotations(final AnnotatedNode node) {
```
jevanlingen, line 1286: I think it's better to not copy the whole file, but just only override this method. So remove all the methods and only keep this method in tact. And then change:
```diff
- Expression adjusted = transformInlineConstants(newValue);
+ // OR: Disable inlining, as it causes the AST and source code to be out of sync.
+ // Expression adjusted = transformInlineConstants(newValue);
```

This way it's less error prone to changes to this file + way easier to spot what we changed in the first place.

knutwannheden: That would be nice. I tried that first, but there are a lot of private methods called here, which in turn call other private methods, etc. So in the end a lot of code ends up getting pulled in.

knutwannheden: I fear that it will end up being harder to sync if there are upstream changes. At the moment the file is a copy with just two lines removed.

jevanlingen:
This PR is also the starting point from where we could work to stop other transformations happening. The [org.codehaus.groovy.transform](https://github.com/apache/groovy/tree/master/src/main/java/org/codehaus/groovy/transform) package contains a lot of transformations (most working with the annotations found in the [groovy.transform](https://github.com/apache/groovy/tree/master/src/main/java/groovy/transform) package).

PR #5007, opened 2025-02-10
Upgrading Antlr to 4.13.2

greg-at-moderne:
## What's changed?

Upgrading ANTLR to 4.13.2 (latest release as of now).

## What's your motivation?

Nothing specific strictly other than staying up-to-date.
I noticed we are using a >2y old version when debugging something and I thought we could upgrade.


greg-at-moderne:
The test failure of `UpgradeDependencyVersionTest`:
```
Caused by: org.openrewrite.maven.MavenDownloadingException: Unable to download POM: org.apache.hbase:hbase-annotations:1.1.1-mapr-1602. Tried repositories:	
https://repository.mapr.com/nexus/content/groups/mapr-public/: HTTP 503
```
is unrelated.

greg-at-moderne:
And now the failure is:
```
Error during rewrite run
java.lang.StringIndexOutOfBoundsException: Range [2725, 2705) out of bounds for length 11697
```
during the apply-openrewrite-suggestions. Also unrelated.

knutwannheden:
The job fails due to an OOME now.

timtebeek:
I think we've determine that the OOME is caused by the Lombok annotation parser issues in
- https://github.com/openrewrite/rewrite/pull/5013

Does that mean we can already merge this changeset? Figured helpful if we're eyeing a new .patch release.

knutwannheden:
I am afraid this PR alone won't help. When running in a build tool plugin, it would probably fail, as it would either not find Lombok at all on the classpath or the wrong version.

timtebeek:
> I am afraid this PR alone won't help. When running in a build tool plugin, it would probably fail, as it would either not find Lombok at all on the classpath or the wrong version.

Did you intend that comment for [this PR](https://github.com/openrewrite/rewrite/pull/5013) instead?

knutwannheden:
> Did you intend that comment for [this PR](https://github.com/openrewrite/rewrite/pull/5013) instead?

Yes.

knutwannheden:
I think we could merge this PR, but please be aware that we have some moderneinc repos which also have ANTLR dependencies and we should probably coordinate that effort, as I am not 100% clear on ANTLR's binary compatibility, especially wrt those ATNs which get serialized into the Java source code. I know this caused problems in one of the more recent releases (maybe 4.10 or 4.11).

PR #5012, opened 2025-02-11
Print issue with Object switch on Java 21+

timtebeek:
As discovered on
- https://github.com/openrewrite/rewrite-static-analysis/pull/462


timtebeek:
Weird how this is passing in CI, but was failing locally for me. We'll close this for now, but can reopen if needed.

PR #5025, opened 2025-02-12
Hardcode Lombok 1.18.37 version

knutwannheden:
Only use this version of Lombok.

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
@@ -124,13 +124,13 @@ public class ReloadableJava11Parser implements JavaParser {
         Options.instance(context).put("should-stop.ifError", "GENERATE");
 
         LOMBOK:
-        if (classpath != null && classpath.stream().anyMatch(it -> it.toString().contains("lombok"))) {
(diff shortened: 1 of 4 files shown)


rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
```diff
@@ -124,13 +124,13 @@ private ReloadableJava11Parser(
         Options.instance(context).put("should-stop.ifError", "GENERATE");
 
         LOMBOK:
-        if (classpath != null && classpath.stream().anyMatch(it -> it.toString().contains("lombok"))) {
+        if (classpath != null && classpath.stream().anyMatch(it -> it.toString().contains("lombok-1.18.37"))) {
```
timtebeek, line 127: I'm wondering if the below ` || part.toString().contains("rewrite-java-lombok"` will trigger now that this specifically looks for `-1.18.37`. Any insights there?

timtebeek (approved):
Approved already, but one question just above. ☝🏻

rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11Parser.java
knutwannheden: If Lombok itself isn't on the classpath, we can skip this entire code path.

PR #5029, opened 2025-02-12
Support for RELEASE metaversion in Maven

greg-at-moderne:
### Checklist


rewrite-core/src/main/java/org/openrewrite/semver/VersionComparator.java
```diff
@@ -16,28 +16,125 @@
...
 import java.util.Comparator;
 import java.util.Optional;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public interface VersionComparator extends Comparator<String> {
-    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?([-.+].*)?");
-    String[] RELEASE_SUFFIXES = new String[]{".final", ".ga", ".release"};
-    Pattern PRE_RELEASE_ENDING = Pattern.compile("[.-](alpha|a|beta|b|milestone|m|rc|cr|snapshot)[.-]?\\d*$", Pattern.CASE_INSENSITIVE);
+public abstract class VersionComparator implements Comparator<String> {
```
timtebeek, line 27: I have a slight worry that this type of change would make it hard to use newer plugins with older recipe libraries, if those had been compiled expecting `VersionComparator` to be an interface. This type of change might be source compatible, but I'm not sure it's binary compatible, and we can't force everyone to recompile & release their recipe modules at once. Would you agree that's a potential concern here?

shanman190: I absolutely agree with you, Tim. This change is source compatible, but not binary compatible.

timtebeek: That would be something to tack onto a list for Rewrite 9. For now I'd hold off on such a change here.

PR #5036, opened 2025-02-13
Improve MergeYaml recipe

jevanlingen:
## What's changed?
- Support comments for the `insert before` option
- Support `insert before` for sequences
- Add support to `insert after`.

## What's your motivation?
- https://github.com/openrewrite/rewrite/pull/4992

was written a little to fast. Lot's of edge cases were missing.

## Any additional context
There are still bugs left:
- Adding comments in the `YAML snippet` option does not really work; all cases where it does work are just by luck.
- Comments in [sequences](https://docs.openrewrite.org/concepts-and-explanations/yaml-lst-examples#sequence) do not work very well either. To fix this with current LST model, the same kind of code written inside the `mergeMapping` function should be written inside the `mergeSequence` too.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
@@ -23,6 +23,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 
 import static java.util.Collections.emptyList;
@@ -351,6 +352,37 @@ public final class ListUtils {
     @Contract("null, _ -> null; !null, _ -> !null")
     public static <T> @Nullable List<T> flatMap(@Nullable List<T> ls, Function<T, Object> flatMap) {
         return flatMap(ls, (i, t) -> flatMap.apply(t));
+    }
+
+    /**
+     * Filters a list based on the given predicate. If no elements are removed, the original list is returned.
+     *
+     * @param ls        The list to filter.
+     * @param predicate The predicate to test each element.
+     * @param <T>       The type of elements in the list.
+     * @return A new list with filtered elements, or the original list if unchanged.
+     */
+    @Contract("null, _ -> null; !null, _ -> !null")
+    public static <T> @Nullable List<T> filter(@Nullable List<T> ls, Predicate<T> predicate) {
+        if (ls == null || ls.isEmpty()) {
+            //noinspection ConstantConditions
+            return ls;
+        }
+
+        List<T> newLs = ls;
+        int j = 0;
+        for (int i = 0; i < ls.size(); i++, j++) {
+            T elem = ls.get(i);
+            if (!predicate.test(elem)) {
+                if (newLs == ls) {
+                    newLs = new ArrayList<>(ls);
+                }
+                newLs.remove(j);
+                j--;
+            }
+        }
+
+        return newLs;
     }
 
     /**
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.yaml;
 
+import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.intellij.lang.annotations.Language;
@@ -28,6 +29,7 @@ import static org.openrewrite.yaml.MergeYaml.InsertMode.*;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
 public class MergeYaml extends Recipe {
     @Option(displayName = "Key path",
             description = "A [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expression used to find matching keys.",
@@ -75,6 +77,14 @@ public class MergeYaml extends Recipe {
             example = "some-key")
     @Nullable
     String insertProperty;
+
+    /**
+     * @deprecated Use {@link #MergeYaml(String, String, Boolean, String, String, InsertMode, String)} instead.
+     */
+    @Deprecated
+    public MergeYaml(String key, @Language("yml") String yaml, @Nullable Boolean acceptTheirs, @Nullable String objectIdentifyingProperty, @Nullable String filePattern, @Nullable String insertProperty) {
+        this(key, yaml, acceptTheirs, objectIdentifyingProperty, filePattern, null, insertProperty);
+    }
 
     public enum InsertMode { Before, After, Last }
 
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -23,6 +23,8 @@ import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.yaml.tree.Yaml;
 
+import static org.openrewrite.internal.StringUtils.isBlank;
+
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class MergeYaml extends Recipe {
@@ -91,7 +93,6 @@ public class MergeYaml extends Recipe {
     }
 
     final static String FOUND_MATCHING_ELEMENT = "FOUND_MATCHING_ELEMENT";
-    final static String REMOVE_DOCUMENT_PREFIX = "REMOVE_DOCUMENT_PREFIX";
     final static String REMOVE_PREFIX = "REMOVE_PREFIX";
 
     @Override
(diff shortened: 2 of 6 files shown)


rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
```diff
@@ -353,6 +354,37 @@ public static <T> List<T> insert(@Nullable List<T> ls, @Nullable T t, int index)
...
+                    newLs = new ArrayList<>(ls);
+                }
+                newLs.remove(j);
+                j--;
+            }
+        }
+
+        return newLs;
+    }
+
```
greg-at-moderne, line 387: Out of curiosity - why would we need this over `.streams().filter().collect(toList)`?
Is it about making sure the original list is returned if no elements were filtered?

If so, wouldn't it be better if the implementation was:
- do `stream().filter().collect`
- check size on this
- if different than return the result of stream processing, if the same return the original list?

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
```diff
@@ -58,12 +61,22 @@ public class MergeYaml extends Recipe {
...
 
-    @Option(displayName = "Insert before property",
-            description = "Choose an insertion point when multiple mappings exist. Takes the `key` JsonPath into account.",
+    @Option(displayName = "Insert mode",
+            description = "Choose an insertion point when multiple mappings exist. Default is `Last`.",
+            valid = {"Before", "After", "Last"},
+            required = false,
+            example = "Last")
+    @Nullable
+    InsertMode insertMode;
```
greg-at-moderne, line 70: That's a change in the interface and I bet `MergeYaml` might be used in a few places (beyond these changed in this PR).
Not sure how do we typically deal with such interface changes?

greg-at-moderne (approved):
LGTM, didn't read the test cases.

rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
timtebeek: We indeed want to return the original collection when there are no changes, such that referential equality works. A lot of recipes use a pattern like `var abc_ = abc.withXyz(ListUtils.filter(abc.getXyz(), elem -> ...))`, which then means we can a very cheap comparison of `abc != abc_`.

We generally try to avoid the streams in what might be hot code paths, as there's quite some object allocation involved which could hurt performance. In this specific case we avoid instantiating a new arraylist whenever none of the elements fail the predicate.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -298,46 +308,42 @@ private Yaml.Sequence mergeSequence(Yaml.Sequence s1, Yaml.Sequence s2, P p, Cur
...
+                    return e1;
                 }
-                return s1;
             }
-        }
+            return entry;
+        });
+
+        List<Yaml.Sequence.Entry> entries = concatAll(
+                filter(s1.getEntries(), it -> !mutatedEntries.contains(it)),
```
timtebeek, line 339: Do we really need a new explicit `ListUtils.filter` operation used only once if we could also use
```suggestion
                map(s1.getEntries(), it -> mutatedEntries.contains(it) ? null : it),
```
I'm looking to minimize the API surface we're adding and maintaining going forward. A bit on the fence on this one.

rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
jevanlingen: > Is it about making sure the original list is returned if no elements were filtered?

Yeah, that's right. The goal of this ListUtils is to do as little changes to the list as possible; we don't want to alter the LST if we don't have to (less memory allocations => more performance). 

The Streaming API has a little overhead, if you traverse a lot of stuff, it starts to be noticeable. See also:
[Recipes must not mutate LSTs](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#recipes-must-not-mutate-lsts).

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
jevanlingen: I didn't even consider doing this 🧐. Technically you are right, but it really feels like a hack. But I can change it of course, if you are adement about this one!

---

Sidenote: That's my problem with the ListUtils right now, the contract of the functions are not confirm to the mathematical laws. A `map` actions to a list of x items should always return a [new] list with again x items. It's a little bit of nitpicking of course, but it does not make the code readable if you start adding these kind of side-effects to universal concepts like mapping/flat-mapping.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
jevanlingen: It's nullable, so the declarative recipes will work without a problem. I think the imperative recipes just have to update their constructors invocations... But what do you say @timtebeek? Should we add extra 'old' constructors to be binary compatible with older versions?

timtebeek: So there's also a bit of classloading fun involved here: the Yaml cosntructor classes are loaded on the workers we run already it seems, and those are deployed on a different schedule. Best then to temporarily add a deprecated backwards compatible constructor. That way we can deploy the workers without any time sensitivity, and folks do not have to instantly update their recipes.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
timtebeek: Mapping elements to `null` is quite common throughout our code base already, but perhaps indeed it helps to be explicit. Hence why I was a bit on the fence here. Ok to add this then I suppose and gradually adopt.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -62,14 +64,12 @@ public class MergeYamlVisitor<P> extends YamlVisitor<P> {
...
+    private final InsertMode insertMode;
 
-    private boolean shouldAutoFormat = true;
+    @Nullable
+    private final String insertProperty;
 
-    public MergeYamlVisitor(Yaml.Block block, Yaml incoming, boolean acceptTheirs, @Nullable String objectIdentifyingProperty, boolean shouldAutoFormat) {
-        this(block, incoming, acceptTheirs, objectIdentifyingProperty, null);
-        this.shouldAutoFormat = shouldAutoFormat;
-    }
```
timtebeek, line 72: Figured I'd call out I saw `MergeYamlVisitor` used directly downstream in rewrite-spring and rewrite-circleci. Be sure to update the usage there after this is merged. Same goes for direct usage of `MergeYaml` itself.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -346,20 +352,60 @@ private Yaml.Scalar mergeScalar(Yaml.Scalar y1, Yaml.Scalar y2) {
...
+        for (int i = 0; i < ls.size(); i++) {
+            T existingEntry = ls.get(i);
+            if (!hasInsertedBeforeOrAfterElements && insertMode == Before && insertProperty.equals(getValue.apply(existingEntry))) {
+                hasInsertedBeforeOrAfterElements = true;
+                mutatedEntries.addAll(t);
+                insertIndex = i;
+                closeIndex = i + t.size() - 1;
+            }
+            mutatedEntries.add(existingEntry);
+            if (!hasInsertedBeforeOrAfterElements && insertMode == After && insertProperty.equals(getValue.apply(existingEntry))) {
```
timtebeek, line 375: I'm seeing some warnings on these `.equals` calls because `insertProperty` might be null. Can we make the IDE not complain about any new lines of code? Typically press `F2` and `Alt + Enter` until done.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
```diff
@@ -58,12 +65,30 @@ public class MergeYaml extends Recipe {
     @Nullable
     String filePattern;
 
-    @Option(displayName = "Insert before property",
-            description = "Choose an insertion point when multiple mappings exist. Takes the `key` JsonPath into account.",
+    @Option(displayName = "Insert mode",
+            description = "Choose an insertion point when multiple mappings exist. Default is `Last`.",
+            valid = {"Before", "After", "Last"},
+            required = false,
+            example = "Last")
```
timtebeek, line 72: I think we can at least drop the `example` here, as that's already implied by `valid`.
```suggestion
            valid = {"Before", "After", "Last"},
            required = false)
```
For enums we _might_ be able to drop the `valid` as well, but I haven't checked that. Could be good not to repeat ourselves.

timtebeek (approved):
Approved apart from the earlier comments. Thanks again! Great to see the quick turnaround on this.

PR #5043, opened 2025-02-14
Read all TypeTable `classpath.tsv.zip` files

timtebeek:
## What's changed?
Read all TypeTable files on the classpath, not just the first.

## What's your motivation?
rewrite-spring depends on rewrite-migrate-java, and as such there's more than one. If the wrong one is picked up, then classes can not be found. If only the recipe modules' own is picked up, then chaining in recipes from other modules might not work.

## Any additional context
- As discovered on https://github.com/openrewrite/rewrite-spring/pull/676#issuecomment-2659270062


timtebeek:
Opting for async review given timezones and the relative small footprint of this change. We effectively do the same for jars already. Will merge as soon as the tests pass.

PR #5045, opened 2025-02-14
Support modifiers for Groovy method declaration parameters

jevanlingen:
## What's changed?
- **Bugfix**: Support for modifiers for method declarations parameters.
- **Improvement**: Merge `visitModifiers` and `getModifiers` methods.

## What's your motivation?
Code like

```groovy
def accept(final def Map m) {}
def accept2(def m) {}
```

could not be parsed.

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -526,13 +526,10 @@ public class GroovyParserVisitor {
                 Parameter param = unparsedParams[i];
 
                 List<J.Annotation> paramAnnotations = visitAndGetAnnotations(param, this);
-
-                TypeTree paramType;
-                if (param.isDynamicTyped()) {
-                    paramType = new J.Identifier(randomId(), EMPTY, Markers.EMPTY, emptyList(), "", JavaType.ShallowClass.build("java.lang.Object"), null);
-                } else {
-                    paramType = visitTypeTree(param.getOriginType());
-                }
+                List<J.Modifier> paramModifiers = getModifiers(param);
+                TypeTree paramType = param.isDynamicTyped() ?
+                        new J.Identifier(randomId(), EMPTY, Markers.EMPTY, emptyList(), "", JavaType.ShallowClass.build("java.lang.Object"), null) :
+                        visitTypeTree(param.getOriginType());
 
                 Space varargs = null;
                 if (paramType instanceof J.ArrayType && sourceStartsWith("...")) {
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2606,57 +2575,39 @@ private TypeTree visitTypeTree(ClassNode classNode, boolean inferredType) {
...
+            possibleModifiers.remove(currentModifier);
+            modifiers.add(new J.Modifier(randomId(), prefix, Markers.EMPTY, currentModifier, modifierNameToType.get(currentModifier), emptyList()));
+            skip(currentModifier);
+            saveCursor = cursor;
+            prefix = whitespace();
+            currentModifier = possibleModifiers.stream()
+                    .filter(modifierName ->
+                            // Try to avoid confusing a variable name with an incidentally similar modifier keyword
+                            (varName.length() < modifierName.length() || !source.startsWith(varName, cursor)) &&
+                                    source.startsWith(modifierName, cursor))
```
greg-at-moderne, line 2605: Not sure why is this extra check not done on the first iteration of the loop?

greg-at-moderne (approved):
One doubt/question, otherwise LGTM.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: I just copied this `getModifiers` method 😉. Let me take a look if I can do some improvements to this function.

jevanlingen: To answer your question, after reading more of the function. You don't have to do this check in the first iteration of the loop. Notice the very first check is actually done at line 2592:

```java
String currentModifier = possibleModifiers.stream().filter(modifierName -> source.startsWith(modifierName, cursor))
```

 Consider:
```groovy
def defaultPublicStaticFinal = 0  // first example
defaultPublicStaticFinal2 = 0 // second example
```

If there is no keyword before a variable at all, there are surely no modifiers (second example).  The other way, when there are modifiers (first example), you cannot be sure about the next part. It could be a variable where the name contains parts of a keyword, it also could be a next keyword like `def final publicStaticFinal = 0`.

jevanlingen: I did a small readability improvement 😄.

knutwannheden:
I am a bit too late here now, but note that a test like the following would fail:

```java
    @Test
    void parameterWithConflictingTypeName() {
        rewriteRun(
          groovy(
            """
              def accept(final def variable m) {
              }
              """,
                J.MethodDeclaration accept = (J.MethodDeclaration) cu.getStatements().get(0);
                J.VariableDeclarations m = (J.VariableDeclarations) accept.getParameters().get(0);
                assertThat(m.getModifiers()).satisfiesExactly(
                  mod -> assertThat(mod.getType()).isEqualTo(J.Modifier.Type.Final),
                  mod -> assertThat(mod.getKeyword()).isEqualTo("def")
                );
            })
          )
        );
    }
```

knutwannheden:
Also a test like this now fails, because `def` is parsed as the return type:

```java
    @Test
    void defIsNotReturnType() {
        rewriteRun(
          groovy(
            """
              final def accept(final def Object m) {
              }
              """,
                J.MethodDeclaration accept = (J.MethodDeclaration) cu.getStatements().get(0);
                assertThat(accept.getReturnTypeExpression()).isNull();
            })
          )
        );
    }
```

PR #5048, opened 2025-02-17
Groovy Parser - Improve retrieval of modifiers

jevanlingen:
## What's changed?
- Removed use of `RedundantDef` marker in parser
- Improved use of `getModifiers` method

## What's your motivation?
Improvement to:
- https://github.com/openrewrite/rewrite/pull/5045

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
@@ -281,4 +281,3 @@ class MethodDeclarationTest implements RewriteTest {
         );
     }
 }
-
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
@@ -71,7 +71,7 @@ class MethodDeclarationTest implements RewriteTest {
         rewriteRun(
           groovy(
             """
-              def accept(final def Map m) {
+              public final accept(final def Map m) {
               }
               """
           )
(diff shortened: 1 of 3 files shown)


timtebeek:
Looks like there's a duplicate newline at the end of this file. Can we clear that out to appease the bot?
```diff
diff --git a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
index 0601358..e637ca0 100644
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
@@ -269,4 +269,3 @@ final def accept(final def Object m) {
         );
     }
 }
-

```

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
```diff
@@ -71,7 +71,7 @@ void methodDeclarationWithModifiers() {
         rewriteRun(
           groovy(
             """
-              def accept(final def Map m) {
+              public final accept(final def Map m) {
```
knutwannheden, line 74: Not sure if you actually intended to modify a test instead of adding a new one.

jevanlingen: Yep! But now I think about it again, let's create a new test instead of modifying this one. That way it's more explicit we can have modifiers on both the return type and the arguments.

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodDeclarationTest.java
```diff
@@ -67,37 +67,37 @@ def accept(Map m) {
     }
 
     @Test
-    void methodDeclarationWithModifiers() {
+    void primitiveReturn() {
```
jevanlingen, line 70: All the test changes look a little ugly, but now all the kind of same tests are grouped together.

knutwannheden (approved):

PR #5049, opened 2025-02-17
Add a RemoveAfter annotation to mark elements to be deleted

timtebeek:
## What's changed?
Add a new annotation.
Possibly add recipe was well.

## What's your motivation?
Mark elements we want removed in the future, and automate their removal.

I went for a `date` field as opposed to `value`, to keep the door open to a future `version` field too (not planned).

@knutwannheden has suggested something like this in the past I believe; figured open a draft to discuss adding this, and a matching a recipe.

- Log issues for removal: that would clutter the backlog quite a bit, and not lead to a guaranteed removal.
- Add `@Deprecated` annotations: again not guaranteed to lead to removal; and on Java 8 we can not use `forRemoval = true`.

## Any additional context
- Prompted by discussion here: https://github.com/openrewrite/rewrite/pull/5036#discussion_r1958049146

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.internal;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * An annotation that indicates that the element should be removed after a certain point.
+ * Intended to be used with a matching recipe to clear out elements on a schedule.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+public @interface RemoveAfter {
+    /**
+     * Expects a date without a time-zone in the ISO-8601 calendar system, such as 2007-12-03.
+     * @return The date after which this element should be removed.
+     */
+    String date();
+}
--- a/rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
@@ -20,6 +20,7 @@ import java.lang.annotation.RetentionPolicy;
 
 /**
  * An annotation that indicates that the element should be removed after a certain point.
+ * Ideally applied with a matching `@Deprecated` annotation listing the reason & replacement.
  * Intended to be used with a matching recipe to clear out elements on a schedule.
  */
 @Retention(RetentionPolicy.SOURCE)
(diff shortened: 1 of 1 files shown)


jevanlingen:
Notice such an annotation will not work with temporary added annotations itself. For the linked issue I added following Lombok annotations to make it work:

```java
..
@NoArgsConstructor(force = true) // TODO: remove with @deprecated constructor
@AllArgsConstructor // TODO: remove with @deprecated constructor
public class MergeYaml extends Recipe {
  ..
}
```

As you can't annotate annotations itself (well anything is possible with annotation processors, but let's not go there), how would we handle such a situation automatically?

rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
```diff
@@ -0,0 +1,32 @@
...
+package org.openrewrite.internal;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * An annotation that indicates that the element should be removed after a certain point.
+ * Intended to be used with a matching recipe to clear out elements on a schedule.
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
knutwannheden, line 25: Since we only need recipes to operate on this, we might as well use SOURCE retention, right?

knutwannheden: Also we should allow the annotation to target anything, probably.

knutwannheden:
> As you can't annotate annotations itself (well anything is possible with annotation processors, but let's not go there), how would we handle such a situation automatically?

I think we can just place the annotation nearby and also allow it to have an optional description, where we can document what can be removed after the given date. We don't need to automate the removal necessarily. If we want that too, then I suppose we would possibly also need a boolean flag on it or so.

rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
timtebeek: My thought of having it be runtime would be that we can also evaluate and lock down at runtime if we have such a need; something of a kill switch.

knutwannheden: I think that if we in a first step just can get the best practices recipe or a similar mechanism to report that we have some code that can be removed, that would already be a very good start.

rewrite-core/src/main/java/org/openrewrite/internal/RemoveAfter.java
```diff
@@ -0,0 +1,32 @@
...
+ * An annotation that indicates that the element should be removed after a certain point.
+ * Intended to be used with a matching recipe to clear out elements on a schedule.
+ */
+@Retention(RetentionPolicy.SOURCE)
+public @interface RemoveAfter {
+    /**
+     * Expects a date without a time-zone in the ISO-8601 calendar system, such as 2007-12-03.
+     * @return The date after which this element should be removed.
+     */
+    String date();
```
knutwannheden, line 31: I think an optional `description` attribute would make sense to have.

Also, what do you think about some kind of boolean `processable` attribute (what to name this thing?) defaulting to `false`, which would allow a recipe to both remove the annotation and the thing it annotates? I think it sounds interesting, but I wonder how often it is actually actionable. Don't you typically also have to adjust other code too?

timtebeek: I'd briefly looked at adding a description, but then realized this is intended mostly to be used along side `@Deprecated` to signal that usage downstream should move to an alternative implementation, so i figured there would not be a need to  duplicate that message.

As to the `processable` I'm not sure that's necessary? Or how were you intending to use that when true vs false?

I'd mostly intended this annotation when we say add a field to a class, and want to add a backwards compatible constructor. It could then make sense to both deprecate and mark for future removal, to give folks some time to adapt.

Do you think we'd need anything more here? Or can we either merge or close if not deemed immediately usable?

timtebeek: That could work as well; then I'll close this PR and we can look at structured comments when we find time to build a recipe.

PR #5050, opened 2025-02-17
Rewrite RPC

jkschneider:
## What's changed?

Rewrite RPC is used to support non-JVM languages for both parsing and recipe writing/running, allowing LSTs to be passed between process boundaries.

We had an initial implementation that we've worked on over the last year and which was called `rewrite-remote`. We're now incorporating the lessons we learned from that initial work to perfect(ish) the design as we bring it into openrewrite/rewrite.

We'll also be bringing rewrite-javascript/python/csharp into this repository as a part of this work, though maybe not all in this single PR.

@knutwannheden

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.rpc;
+
+import io.moderne.jsonrpc.JsonRpc;
+import io.moderne.jsonrpc.JsonRpcRequest;
+import org.openrewrite.SourceFile;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.rpc.request.GetTreeDataRequest;
+import org.openrewrite.rpc.request.RecipeRpcRequest;
+import org.openrewrite.rpc.request.VisitRequest;
+import org.openrewrite.rpc.request.VisitResponse;
+
+import java.lang.reflect.Constructor;
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.*;
+
+import static io.moderne.jsonrpc.JsonRpcMethod.typed;
+import static org.openrewrite.rpc.TreeDatum.State.END_OF_TREE;
+
+public class RewriteRpc {
+    private final JsonRpc jsonRpc;
+    private final Duration timeout;
+
+    /**
+     * Keeps track of the local and remote state of trees that are used in
+     * visits.
+     */
+    private final Map<UUID, SourceFile> remoteTrees = new HashMap<>();
+    private final Map<UUID, SourceFile> localTrees = new HashMap<>();
+
+    /**
+     * Keeps track of objects that need to be referentially deduplicated, and
+     * the ref IDs to look them up by on the remote.
+     */
+    private final Map<Object, Integer> localRefs = new IdentityHashMap<>();
+    private final Map<Integer, Object> remoteRefs = new IdentityHashMap<>();
+
+    // TODO This should be keyed on both the visit (transaction) ID in addition to the tree ID
+    private final Map<UUID, BlockingQueue<TreeData>> inProgressGetTreeDatas = new ConcurrentHashMap<>();
+
+    private static final ExecutorService forkJoin = ForkJoinPool.commonPool();
+
+    public RewriteRpc(JsonRpc jsonRpc, Duration timeout) {
+        this.jsonRpc = jsonRpc;
+        this.timeout = timeout;
+
+        jsonRpc.method("visit", typed(VisitRequest.class, request -> {
+            Constructor<?> ctor = Class.forName(request.getVisitor()).getDeclaredConstructor();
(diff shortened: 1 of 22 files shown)


rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -0,0 +1,218 @@
...
+    public <P> VisitResponse scan(SourceFile sourceFile, String visitorName, P p) {
+        return scan(sourceFile, visitorName, p, null);
+    }
+
+    public <P> VisitResponse scan(SourceFile sourceFile, String visitorName, P p,
+                                  @Nullable Cursor cursor) {
+        // Set the local state of this tree, so that when the remote
+        // asks for it, we know what to send.
+        localObjects.put(sourceFile.getId().toString(), sourceFile);
+        String pId = Integer.toString(System.identityHashCode(p));
```
knutwannheden, line 117: I see we currently use `System.identityHashCode()` as IDs for non-`Tree` objects (like `ExecutionContext` and non-`Tree` objects in the cursor stack). I think that is a bad choice as there could be collisions. We could just use an `AtomicLong` as a generator.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public String print(Tree tree, Cursor parent) {
+        localObjects.put(tree.getId().toString(), tree);
+        return send("Print", new Print(tree.getId().toString(), getCursorIds(parent)), String.class);
+    }
+
+    private @Nullable List<String> getCursorIds(@Nullable Cursor cursor) {
+        List<String> cursorIds = null;
+        if (cursor != null) {
+            cursorIds = cursor.getPathAsStream().map(c -> {
```
knutwannheden, line 162: Just to confirm: I assume we are here deliberately not calling `getPathAsCursors()`, as we don't want to support cursor messaging over RPC.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -0,0 +1,218 @@
...
+                localObjects.put(id, c);
+                return id;
+            }).collect(Collectors.toList());
+        }
+        return cursorIds;
+    }
+
+    @VisibleForTesting
+    public <T> T getObject(String id) {
+        RpcReceiveQueue q = new RpcReceiveQueue(remoteRefs, () -> send("GetObject",
```
knutwannheden, line 175: Should we here check if `remoteObjects` already contains the object we are looking for? Not sure about this, but it seems like if we had a map of remote objects that were specific to the current "transaction", we shouldn't have to fetch the object.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -0,0 +1,218 @@
...
+        } catch (ExecutionException | TimeoutException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @VisibleForTesting
+    Cursor getCursor(@Nullable List<String> cursorIds) {
+        Cursor cursor = new Cursor(null, Cursor.ROOT_VALUE);
+        if (cursorIds != null) {
+            for (String cursorId : cursorIds) {
```
knutwannheden, line 210: To make the cursor align with what gets produced by `getCursorIds()` I propose that we here iterate over the IDs in reverse order:
```suggestion
            for (int i = cursorIds.size() - 1; i > 0; --i) {
                String cursorId = cursorIds.get(i);
```

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
knutwannheden: Or that `SnowflakeId` would certainly also work.

rewrite-core/src/main/java/org/openrewrite/rpc/request/Visit.java
```diff
@@ -0,0 +1,152 @@
...
+            return new VisitResponse(before != after);
+        }
+
+        private TreeVisitor<?, ?> instantiateVisitor(Visit request) {
+            String visitorName = request.getVisitor();
+
+            if (visitorName.startsWith("scan:")) {
+                //noinspection unchecked
+                ScanningRecipe<Object> recipe = (ScanningRecipe<Object>) preparedRecipes.get(visitorName.substring("scan:".length()));
+                return recipe.getScanner(0);
```
knutwannheden, line 97: Unless we already want to instantiate and register the accumulator as part of the `PrepareRecipe` requests, we need to do it lazily at this point at the latest. This means that we in `handle()` need to call `getVisitorP()` first and pass it in here and then check if the accumulator has already been registered in the execution context.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
jkschneider: Cursor messages only survive a single visit by design, and aren't supposed to survive further than that.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
jkschneider: No, even if the object is in `remoteObjects` it may not be the most recent iteration on that object. We always have to fetch to tell.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
jkschneider: I don't think collisions are a practical reality here given how few objects wind up in the cache.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
jkschneider: Good catch. Fixed in ac16cdc7240c11efb357ffa2fafab16233bc15db.

rewrite-core/src/main/java/org/openrewrite/rpc/request/Visit.java
jkschneider: I don't understand the concern here. ScanningRecipe already initializes the accumulator lazily at the point where it supplies it to the visitor/scanner.

image

Issue #4949, opened 2025-01-25
Removing managed dependency problem: Upgrade parent version or remove redundant version recipe (GTE)


crankydillo:
My ultimate goal is to upgrade the parent pom of a multi-module Maven project and to remove any project dependency versions that are lower that what the new parent manages. We initially started with UpgradeParentVersion, but doesn't seem to work well with modules. I'm suspicious that UpgradeMavenModel isn't propagating all changes to the 'POM tree', but I may be misunderstanding how that works. I have a few other issues/comments about this (apologies for not linking yet). Regardless, in later recipes we are messing with dependencies , so added a RemoveRedundantDependencyVersions recipe at the end. This covers multiple cases that you'd think would be covered by UpgradeParentVersion, but it doesn't seem to handle the case where if a.group:a.artifact:1.0 did not exist in the old parent, but does in the new. In that case, we expect all project usage of a.group:a.artifact:1.0 to be replaced with a.group:a.artifact (no version).

The following test, which I added to UpgradeParentVersionTest, is the shortest form of what we're experiencing. jakarta.annotation-api was added in spring-boot-dependencies:3. When you run, you will see that dependency version does not get removed from the child module pom. If you only run UpgradeParentVersion in the test (remove RemoveRedundant..), you'll see neither gets removed. If I bring that back and set onlyIfManagedVersionIs to ANY, it does remove it. However, we don't want to override project-level dependencies that are higher than what we manage.

Originally posted in [this discussion](https://github.com/openrewrite/rewrite/discussions/4938)

## What version of OpenRewrite are you using?

Test below was run on that `main` branch in the `rewrite-maven` module

## What is the smallest, simplest way to reproduce the problem?

```java
    @Test
    void sams() {
        rewriteRun(
          spec -> spec.recipe(
            new CompositeRecipe(
              Arrays.asList(
                new UpgradeParentVersion("org.springframework.boot", "spring-boot-dependencies", "3.4.1", null, null)
               , new RemoveRedundantDependencyVersions(null, null, RemoveRedundantDependencyVersions.Comparator.GTE, null)
              )
            )
          ),
            pomXml("""
              <project>
                <parent>
                  <groupId>org.springframework.boot</groupId>
                  <artifactId>spring-boot-dependencies</artifactId>
                  <version>2.4.0</version>
                </parent>
                <groupId>cool.stuff</groupId>
                <artifactId>parent</artifactId>
                <version>1.0-SNAPSHOT</version>
[...]

PR #5051, opened 2025-02-18
Fix updating of maven model when parent pom was updated previously in same cycle

mryan43:
## What's changed?
- Move projectPoms to the ExecutionContext instead of PomDownloader + MavenResolutionResult + UpdateMavenModel 
- in MavenVisitor, update the current maven resolution result if a parent pom from the project was updated
- Added minimal test to showcase issue caused by projectPoms not being shared between MavenResolutionResults

## What's your motivation?
- In multi-module maven projects where both the parent and child are in the same project, updates made to the parent  pom are not visible to recipes updating children poms, causing various bugs.
- The typical scenario that I want to fix is : 
1) A first recipe modifies a parent pom, 
2) Another recipe attempts to modify a child of this pom and requires the changes from step 1
3) Since the MavenResolutionResult is a Marker, the parent and the child each have their own and UpdateMavenModel does not update the marker on the child document when the parent model is updated.
4) Also, the PomDownloader uses its own projectPoms which can be stale, making the resolution process return out-of-date poms.
4) the recipe that should modify the child fails because the model is not up to date between documents

See added test for a very simple real-life scenario.
I also believe that this PR could fix a number of similar issues with many different combinations of recipes, such as this one for example : #4949

- I have considered using the ExecutionContext to share MavenResolutionResult created by UpdateMavenModel. It doesn't work because the PomDownloader uses projectPoms that are not shared between MavenResolutionResult during the resolution process.
- I have considered adding the ExecutionContext as a transient field in MavenResolutionResult but this solution was creating really bad coupling and would cause maintenance issues over time.

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
```diff
@@ -203,10 +204,7 @@ public boolean parentPomIsProjectPom() {
         if (getParent() == null) {
             return false;
         }
-        ResolvedGroupArtifactVersion parentGav = getParent().getPom().getGav();
-        return getProjectPoms().values().stream()
-                .map(Pom::getGav)
-                .anyMatch(gav -> gav.equals(parentGav));
+        return getParent().getPom().getRequested().getSourcePath() != null;
```
mryan43, line 207: I first considered adding the `ExecutionContext` as a parameter but that would break the public API of `getProjectPoms`. Is `getSourcePath() != null` a good enough invariant ? if not, how could I access the `projectPoms` from here ?

timtebeek:
Thanks a lot @mryan43 ! Tagging Knut for review as he proposed something similar for constructors just today.

MBoegers:
Hi @mryan43, looks neat! I will have a detailed look tomorrow in the morning 📓

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
MBoegers: Additional tests show its not a problem.

MBoegers (changes requested):
I've added a bit more coverage for some edgecases and found a problem with two layers of parent/child releationship, could you quickly elaborate if its a problem with `RemoveRedundantDependencyVersions` or you changes?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
MBoegers: the way to access the parent is aligned with other implementations, for the nullness see my comments above but it seems to be no problem.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -88,62 +85,96 @@ public class MavenPomDownloader {
     private boolean addLocalRepository;
 
     /**
-     * @param projectPoms    Other POMs in this project.
+     * @param ctx The execution context, which potentially contain Maven settings customization and {@link HttpSender} customization.
+     */
+    public MavenPomDownloader(ExecutionContext ctx) {
```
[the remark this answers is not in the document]
MBoegers: yes please make it a factory, atm we don't plan for breaking changes ;)

rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
```diff
@@ -97,9 +97,10 @@ public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path re
             }
         }
 
-        MavenPomDownloader downloader = new MavenPomDownloader(projectPomsByPath, ctx);
-
         MavenExecutionContextView mavenCtx = MavenExecutionContextView.view(ctx);
+        mavenCtx.setProjectPoms(projectPomsByPath);
```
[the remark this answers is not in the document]
MBoegers: You are right, I thought about the updating after a change, but there is no `MavenParser` involved.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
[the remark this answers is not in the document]
knutwannheden: I guess this is one caller that needs to be updated: https://github.com/openrewrite/rewrite-java-dependencies/blob/f43e93421f70b510c82a4b8e7461e4b15e310648/src/main/java/org/openrewrite/java/dependencies/DependencyResolutionDiagnostic.java#L147-L148

Let's also deprecate this method.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -88,62 +85,96 @@ public class MavenPomDownloader {
...
+        this(HttpSenderExecutionContextView.view(ctx).getHttpSender(), ctx);
+    }
+
+    /**
+     * A MavenPomDownloader for non-maven contexts where there are no project poms or assumption that maven central
+     * is implicitly added as a repository. In a Maven contexts, the regular constructor should be used instead
+     *
+     * @param ctx The execution context
+     */
+    public static MavenPomDownloader forNonMavenContext(ExecutionContext ctx) {
```
knutwannheden, line 100: I am wondering if we should maybe unconditionally set `addCentralRepository` and `addLocalRepository` to `false`.

I also think it would be nice to have two (or three?) factory methods instead of constructors. But this can be done in a separate step.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -170,22 +201,8 @@ byte[] sendRequest(HttpSender.Request request) throws IOException, HttpSenderRes
...
-                    projectPom.getArtifactId(),
-                    ResolvedPom.placeholderHelper.replacePlaceholders(projectPom.getVersion(), mergedProperties::get)
-            );
-            result.put(gav, projectPom);
-        }
-        return result;
-    }
 
-    private Map<String, String> mergeProperties(final List<Pom> pomAncestry) {
+    public static Map<String, String> mergeProperties(final List<Pom> pomAncestry) {
```
Laurens-W, line 205: Should these methods perhaps move to the MavenExecutionContextView? Then on line 507 you can cast the context and call it that way

jevanlingen: I think you are right. Let me move those.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
mryan43: thanks for checking and I'm glad it works out

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
mryan43: I think it makes sense to do it in scope of this task as we deprecate all the `projectPom` parameters in constructors and remove them in calls to factory methods

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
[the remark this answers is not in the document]
mryan43: I have deprecated all previously existing constructors without changing their behavior and created 3 new factory methods instead that don't require any `projectPoms`

mryan43:
PR is ready for another round of review.

PS: Huge thank you for all the work everyone has put in this PR today so far !

mryan43:
Merged origin/main into this branch, please approve again for CI. This PR is still awaiting a re-review after comments were addressed.

MBoegers:
Hi @mryan43 as you see from the number of participants, updating `MavenResolutionResults` after changing the Maven Pom content is a pain point for OpenRewrite, and we are very pleased by your contribution. 
You already noticed the public activity dropped, please let me explain. We noticed that your solution utilizes the `ExecutionContext` to store `Pom` instances. The `ExecutionContext` is designed to store meta information and configurations. Accumulators of `SearchRecipe`s, Maven Configurations and likewise are data to store in the `ExecutionContext`. `Pom`s in contrast, are LST elements and the main object of work for OpenRewrite, like a person for a CRM system. With this assumption, build in storing Poms in the `ExecutionContext` can cause major problems in downstream products like the Moderne SaaS or Moderne CLI.
Because of this fear, we came to the conclusion that we shouldn't merge this PR with the current implementation but seek for a solution that does utilize the `ExecutionContext` as designed to transport meta information.

I apologize for the delay and late intervention. I hope you can use your incredible skill and motivation to find a better solution with us.

mryan43:
Thanks for the answer and the call we had recently.

After investigating a bit more, the ExecutionContext seems to already contain those very same LST elements before those changes :
image
image
image

Could you share a bit more details about why the changes from this PR are dangerous for downstream products but those are ok ? That would help me find a better solution that fits everyone's needs

mryan43:
I would also like to challenge the idea that the Pom class represents a LST element.
I will argue that it is already meta-data:

- `MavenVisitor`  extends `TreeVisitor<XML,P>`, it does not have a `visitPom` or `visitDependency` method for example, only methods for visiting xml tags.
- `MavenVisitors` being an XML visitor, they do not mutate the fields of the Pom class when visiting. That is most likely why the UpdateMavenModel visitor was created in the first place and needs to run after modifications to the XML, to update the metadata (therefore MavenResolutionResult is also meta-data)
- Poms can be returned by the `MavenPomDownloader` class. This class can return poms downloaded from remote artifact repositories, how are those downloaded poms LST elements in OpenRewrite ? Could they be modified by visitors for example ? obviously not...

For all those reasons... I think only XML documents and tags can be considered LST elements in the current implementation.

knutwannheden:
You are right, Poms are not actually LST elements for the reasons you mention. There is also no technical reason why LST elements can't be stored in the ExecutionContext. Sometimes that can be the best way to achieve something. The reason we discourage storing LSTs in the ExecutionContext is that these can consume a lot of memory, especially if they have type attribution attached to them as in the case of Java. In the case of Maven the MavenResolutionResult can also consume quite a lot of memory due to all the transitive dependencies. I am not sure how much of that is attributed to the Pom objects, but I think it should probably be OK and I don't really see any other option anyway.

knutwannheden:
Also note that while using the InMemoryLargeSourceSet for recipe execution this isn't a problem, because all LST are already loaded into memory anyway and thua adding a reference to the ExecutionContext doesn't occupy any additional memory. With the Moderne SaaS and CLI the situation is different as they support processing huge repos by not loading all LSTs at once into memory. By carrying LSTs forward in the ExecutionContext the memory of those LSTs can never be released.

mryan43:
Thanks a lot for clarifying those concerns ! 

I was wondering if there exists any mechanism other than the `ExecutionContext` to propagate the update between parent and child pom that could be better suited ? I didn't find any in my research but this is my first contribution so I might have missed something

knutwannheden:
The only alternative I can come up with is if we would register something like a `UnaryOperator<Pom>` (probably a dedicated functional interface would be the better alternative) in the `ExecutionContext` as a function to transform the parent `Pom`. At least theoretically this seems like it should also be doable. This function could then be applied to transform the parent POM both for the parent project and for child projects. I am just putting this out there for discussion.

mryan43:
I'm not convinced that we can find a better solution without heavily refactoring the whole maven module... 

The question remains about what should happen to this PR.
I'm a bit sad that the fix seems to be blocked only because the architecture of the downstream closed source project.

mryan43:
I can add an interface with an implementation that contains the full Pom object and replaces it. Then downstream maintainers can replace it by whatever they want. Would that work for you ?

This would be similar to how the ExecutionContext vs InMemoryExecutionContext works

crankydillo:
How can we progress on this?  Ultimately, when a recipe is crawling over a child pom, we need all the parent metadata to be up to date.  I even tried multi-cycle as a quick hack to make things work, but it didn't.  I didn't really dig into why, but I seem to recall some caching.

Is the function approach mentioned [here](https://github.com/openrewrite/rewrite/pull/5051#issuecomment-2784489400) being discussed?

PR #5052, opened 2025-02-18
Add recepie `RemoveRedundantImports`

punkratz312:
Add new recepie `RemoveRedundantImports` to fix: https://checkstyle.sourceforge.io/checks/imports/redundantimport.html


punkratz312:
@timtebeek is this any good?

timtebeek:
Thanks for kicking this off @punkratz312 ! Couple quick immediate thoughts:
- we might need to add a precondition to skip languages where one might add an alias for an import, or restrict to known languages. We have such preconditions in rewrite-static-analysis; to be determined if we move those here, or move this recipe there
- it'd be interesting to see if our OrderImport recipe already supports this, or could be extended to support this. I think it might make sense to have this functionality there as opposed to a recipe that folks have to deliberately run separately. What are your thoughts on that?
- the recipe file header and class javadoc appear outdated copies. Perhaps good to clear that up if we do continue with this PR.
- In terms of implementation it might make sense to use ListUtils.map or ListUtils.filter instead of the usage of the Streams API

Let's discuss before making further changes just so we're on the same page :)

punkratz312:
Yes, indeed, it's actually working. I've double-checked it with both positive and negative tests.

Of course, it's very much related, so it's justifiable to have this here.

This dedicated approach would only make sense if there is a need or desire to address only the DRY violations and not to change the entire import tree, potentially due to ordering the imports as well.

image

Pankraz76:
feel free to take over and complete.

@iddeepak

timtebeek:
I think we'd rather nudge folks towards using `OrderImports` as opposed to having a separate `RemoveRedundantImports`, which folks would then need to find & run, as well as would need to be maintained going forward. Thanks for the suggestion though! In this case the existing recipes are preferred.

PR #5061, opened 2025-02-18
Declare root project name

gabrielfeo:
## What's changed?

Declare the root project's name. That's an [official Gradle recommendation][1]:

>Define the root project name in the settings file: The rootProject.name effectively assigns a name to the build, used in reports like Build Scans. If the root project name is not set, the name will be the container directory name, which can be unstable (i.e., you can check out your project in any directory). The name will be generated randomly if the root project name is not set and checked out to a file system’s root (e.g., / or C:\).

[1]: https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:naming_recommendations

## What's your motivation?

Ensuring Build Scans in both [OpenRewrite's Develocity instance][2] and [our own instances][3] have a stable project name.

[2]: https://ge.openrewrite.org/
[3]: https://ge.solutions-team.gradle.com/


timtebeek:
Thanks for fixing that here! Let me know if you'd want to collaborate on a recipe that sets this in bulk through [app.moderne.io](https://app.moderne.io) :)

timtebeek (approved):

Issue #4909, opened 2025-01-15
ScanningRecipes that don't add/modify/delete files should report the estimated time savings other than zero


bmuschko:
## What version of OpenRewrite are you using?

- Moderne Sass v10.181.0

## How are you running OpenRewrite?

From the Sass UI.

## What is the smallest, simplest way to reproduce the problem?

Run the recipe `org.openrewrite.FindGitProvenance` against one or many repositories. I am only using this recipe as an example in this case to demonstrate the issue. Say, I'd write a custom `ScanningRecipe` that only generates files but doesn't change any, I think I would see the same issue.

## What did you expect to see?

The "Estimated time savings" on the "Summary" tab should report a higher number than 0. Every recipe sets [5 minutes as the default for the estimated savings](https://github.com/openrewrite/rewrite/blob/f2b923248e2692a3d35bdff15285be256cfd69b6/rewrite-core/src/main/java/org/openrewrite/Recipe.java#L198-L203). Given that a `ScanningRecipe` doesn't override the method, I'd expect the same behavior.

Maybe this is expected behavior but it is not clearly documented as of now. Maybe you can clarify.

## What did you see instead?

The "Estimated time savings" on the "Summary" tab should reports 0 even though the data tables produced by the recipe have been populated. Work has been done even though no files were changed.

<img width="1709" alt="image" src="https://github.com/user-attachments/assets/4745ea4c-0bc0-4cb0-8c4f-0aad8e2acba4" />

timtebeek:
- https://github.com/openrewrite/rewrite/pull/5064

timtebeek:
As indicated on Slack we can further discuss the need to log effort when merely producing data table rows, or whether to include some effort for scanning files that do not result in changes or rows.

bmuschko:
Proven to be working, as per https://github.com/openrewrite/rewrite/pull/5563.

PR #5064, opened 2025-02-18
Record a SourcesFileResults.Row for generated sources

timtebeek:
## What's your motivation?
- Partial(?) fix for https://github.com/openrewrite/rewrite/issues/4909

Do we want to see additional effort logged for #4909 ? This seemed like an omission, but I'm not sure about further changes.

@bmuschko @sambsnyd 

We could also assign some effort for the scanning phase of a recipe, or when creating a data table row. That seems a bit further removed from how we've traditionally estimated and measured effort though, so not (yet) included here.


timtebeek:
Draft as I hadn't ran the tests yet, but otherwise ok to merge if you ask me. I'll look at this again with fresh eyes tomorrow.

bmuschko:
I think it would make sense to override `getEstimatedEffortPerOccurrence()` in `ScanningRecipe` and return a value of `Duration.ofMinutes(0)`. Now, that we calculate time savings for `ScanningRecipe`, "finder" recipes would provide a time saving which probably shouldn't be accounted.

PR #5068, opened 2025-02-19
Restore Yaml.* constructors without Tag for gradual adoption

timtebeek:
## What's changed?
Add constructors to Yaml.Scalar, Yaml.Sequence and Yaml.Mapping as they existed before
- https://github.com/openrewrite/rewrite/pull/4934

## What's your motivation?
We're getting reports about folks that are attempting to run older compiled recipes with newer versions of the CLI.
```
java.lang.NoSuchMethodError: 'void org.openrewrite.yaml.tree.Yaml$Scalar.<init>(java.util.UUID, java.lang.String, org.openrewrite.marker.Markers, org.openrewrite.yaml.tree.Yaml$Scalar$Style, org.openrewrite.yaml.tree.Yaml$Anchor, java.lang.String)'
```
By restoring these constructors we allow for a gradual adoption, as opposed to a hard required cut over.

I'd added the JsonCreator to avoid ambiguity with deserialization; anything else we'd need to think of?

Not possible to force everyone to switch over at once.
Not desirable to roll back the change that added tags.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.yaml.tree;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import lombok.*;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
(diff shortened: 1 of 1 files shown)


sambsnyd (approved):

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
```diff
@@ -243,6 +244,22 @@ public enum Style {
             PLAIN
         }
 
+        @Deprecated
+        public Scalar(UUID id, String prefix, Markers markers, Style style, @Nullable Anchor anchor, String value) {
+            this(id, prefix, markers, style, anchor, null, value);
+        }
+
+        @JsonCreator
```
knutwannheden, line 252: You could instead also have told Lombok to add this annotation to the constructor it generates.

timtebeek: Thanks! I'd considered that, but figured it'd be harder to clear out automatically as opposed to an effective all args constructor that we can remove when it's the only one left as part of Lombok best practices.

PR #5070, opened 2025-02-20
Parse properties from `maven.config` file

Laurens-W:
## What's changed?

Also parse `-D` style properties from `maven.config`.

Test coverage

@timtebeek @greg-at-moderne 

## Any additional context

For now we treat the `maven.config` as a read only file, so it is correct that we cannot write an updated version back to it.

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
@@ -3725,9 +3725,11 @@ class MavenParserTest implements RewriteTest {
                   assertThat(results.getPom().getProperties().get("revision")).isEqualTo("1.0.0");
               }
             ),
-            """
-            -Drevision=1.0.0
-            """
+            MavenParser.builder().mavenConfig(
+              """
+              -Drevision=1.0.0
+              """
+            )
           )
         );
     }
@@ -3748,9 +3750,11 @@ class MavenParserTest implements RewriteTest {
                   assertThat(results.getPom().getActiveProfiles()).contains("a", "b", "c");
               }
             ),
-            """
-            -P a,b,c
-            """
+            MavenParser.builder().mavenConfig(
+              """
+              -P a,b,c
+              """
+            )
           )
         );
     }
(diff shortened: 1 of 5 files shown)


timtebeek:
Thanks! I think it's more so a problem when it comes to the Maven plugin and building LSTs; did you try that for a sample project?

Laurens-W:
> Thanks! I think it's more so a problem when it comes to the Maven plugin and building LSTs; did you try that for a sample project?

Building a project locally with the minimal example works as intended, as well as an OS project I found

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -3707,4 +3708,98 @@ void systemPropertyTakesPrecedence() {
...
+              
+                  <groupId>com.example</groupId>
+                  <artifactId>my-app</artifactId>
+                  <version>${revision}</version>
+              </project>
+              """,
+            spec -> spec.afterRecipe(p -> {
+                var results = p.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow();
+                assertThat(results.getPom().getVersion()).isEqualTo("${revision}");
+                assertThat(results.getPom().getProperties().get("revision")).isEqualTo("1.0.0");
```
jevanlingen, line 3734: I am starting to think the assumption of having the `maven.config` as extra properties is not right. The information of the `text` block is another LST, we don't merge those together. Same applies for Gradle:

```java
@Test
void property() {
    rewriteRun(
      properties("junitVersion=4.13", spec -> spec.path("gradle.properties")),
      buildGradle(
        """
          plugins { id 'java-library' }
          dependencies { testImplementation "junit:junit:${junitVersion}" }
          """,
        spec -> spec.afterRecipe(cu -> cu.getMarkers().getMarkers().get(0))
      )
    );
}
```

That marker would show:

```
org.codehaus.groovy.syntax.SyntaxException: [Static type checking] - The variable [junitVersion] is undeclared.
 @ line 10, column 39.
  org.codehaus.groovy.control.ErrorCollector.addErrorAndContinue(ErrorCollector.java:84)
````

This is also in line with current recipe's. For example the _DependencyVulnerabilityCheck_ recipe, which I have now extended knowledge about, also just treats the properties LST as a different input.

File: rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java [hunk 2]
```diff
@@ -2008,4 +2009,45 @@ void exactVersionMissingInMavenMetadata() {
           )
         );
     }
+
+    @Test
+    void upgradeVersionWithGroupIdAndArtifactIdDefinedAsMavenConfigProperty() {
+        rewriteRun(
```
jevanlingen, line 2015: This test fails at the moment, because the properties in the text LST are not recognised:

image

MBoegers (changes requested):
Could you please add a test for an empty config file?

Laurens-W:
> Could you please add a test for an empty config file?

Like the one I added?

PR #5073, opened 2025-02-20
Correct handling of boolean literals in `JsonPath` grammar and matcher

knutwannheden:


greg-at-moderne:
Jsonpath grammar is also duplicated in `rewrite-hcl` we should copy your fix there too.

knutwannheden:
I fixed it for HCL independently of this as part of another PR. That's how I found this.

PR #5075, opened 2025-02-20
Make doAfterVisit() public

rlsanders4:
## What's changed?
The `TreeVisitor.doAfterVisit()` method was changed from `protected` to `public`.

## What's your motivation?

This will allow `doAfterVisit()` to be called from outside of the `TreeVisitor` class. This enables you to create reuseable helper methods which can be used across multiple recipes. Here is an example of such a use case:

```
public class MyRecipe extends Recipe {
    @Override
        return new JavaVisitor<ExecutionContext>() {
            @Override
            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration cd, ExecutionContext ctx) {
                    Utilities.changeTypeHelper(this);
                    return super.visitClassDeclaration(cd, ctx);
            }
        }
    }
}
```
```
public class Utilities {
    public static void changeTypeHelper(JavaVisitor<ExecutionContext> javaVisitor) {
        if (someCondition) {
            javaVisitor.doAfterVisit(new ChangeType(class1, class2, true).getVisitor());
        }
    }
}
```
 
@timtebeek

You can reimplement the helper method within each `TreeVisitor` class. However, this results in duplicated code.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/TreeVisitor.java
@@ -102,7 +102,7 @@ public abstract class TreeVisitor<T extends @Nullable Tree, P> {
      *
      * @param visitor The visitor to run.
      */
-    protected void doAfterVisit(TreeVisitor<?, P> visitor) {
+    public void doAfterVisit(TreeVisitor<?, P> visitor) {
         if (afterVisit == null) {
             afterVisit = new ArrayList<>(2);
         }


sambsnyd:
Hmm I'm a bit skeptical. There are quite a few ways to compose visitors together and we try to avoid the proliferation of utility classes. Do you have any other usecases or examples of functionality this would unlock for you?

rlsanders4:
Here is a bit more flushed out version of the above use case:

```
public class MyRecipe extends Recipe {
    @Override
        return new JavaVisitor<ExecutionContext>(CustomAccumulator acc) {
            @Override
            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration cd, ExecutionContext ctx) {
                    Utilities.changeTypeHelper(cd.getType().getFullyQualifiedName(), acc, this);
                    return super.visitClassDeclaration(cd, ctx);
            }
        }
    }
}
```
```
public class Utilities {
    public static void changeTypeHelper(String originalName, CustomAccumulator acc, JavaVisitor<ExecutionContext> javaVisitor) {
        boolean someCondition = acc.someConditionalLogic(); // Check the shared accumulator to see if class needs updating
        boolean newName = acc.getNewName(originalName); // Use the original name and saved data to generate a new name
        if (someCondition) {
            javaVisitor.doAfterVisit(new ChangeType(originalName, newName, true).getVisitor());
        }
    }
}
```

In this case `CustomAccumulator` is an accumulator object that is shared between all our recipes. Essentially, we found that we have to do the exact same logic (abreviated in this example) within multiple recipes, and we thought it would be cleaner for us to instead reference a utility method here. Is there a better way to handle this case?

jkschneider:
> is an accumulator object that is shared between all our recipes

How is it being shared between recipes?

rlsanders4:
@jkschneider In the first scanning recipe we set the value in the ExecutionContext:
```
    @Override
    public CustomAccumulator getInitialValue(ExecutionContext ctx) {
        CustomAccumulator acc = new CustomAccumulator();
        ctx.putMessage(CUSTOM_ACCUMULATOR, acc);
        return acc;
    }
```

And then we can access from other recipes like this:
```
    @Override
    public CustomAccumulator getInitialValue(CustomAccumulator ctx) {
        return ctx.getMessage(CUSTOM_ACCUMULATOR);
    }
```

timtebeek:
hi @rlsanders4 ; Based on the initial feedback above I'm inclined to close this PR for now, although we will monitor replies here or in our shared channel.

As for sharing state between recipes: we recently added validation to detect & discourage usage of the execution context to share state:
- https://github.com/openrewrite/rewrite/pull/4879

Perhaps it helps to have a broader understanding of what you're trying to achieve; feel free to use the shared channel if that helps.

PR #5076, opened 2025-02-21
Java parsing fix for empty comment as last line

greg-at-moderne:
## What's changed?

Fixing an edge case of Java parsing. Namely when a single-line empty comment is at the end of the file and there's no newline after that.
```
class A {
}
//
```

## What's your motivation?

Feedback from one of the OpenRewrite users.


knutwannheden:
We might want to check if any of the other Space classes of other grammars have the same issue.

PR #5080, opened 2025-02-21
`ChangeType`: Use unqualified name if possible

knutwannheden:
Changes `ChangeType` so that it replaces qualified type references with unqualified type references when possible.


knutwannheden:
@timtebeek For when you are back again: Do you think we should have an option for `ChangeType` to also shorten any fully qualified names? Now that I think about it, I am not sure if that part of the change was necessary or not. I will look into it again.

knutwannheden:
Since `ChangeType` does use `AddImport` I think it does make results more consistent if we also use `ShortenFullyQualifiedTypeReferences` as in this PR. Otherwise we could end up with unused imports.

Alternatively, we could change the `AddImport` implementation for the `onlyIfReferenced` option to check that any of the references which are found are unqualified. Would that be more consistent?

knutwannheden:
I think this is a good first step. We can refine later, if we want.

PR #5082, opened 2025-02-21
Adding latest and release fields to MavenMetadata, using them in AddDependency

greg-at-moderne:
## What's changed?
Two things:
1. Adding new fields to `MavenMetadata` "record" - latest and release
```xml
<metadata modelVersion="1.1.0">
  <groupId>commons-io</groupId>
  <artifactId>commons-io</artifactId>
  <versioning>
    <latest>2.18.0</latest>   <---------------
    <release>2.18.0</release> <---------------
    <versions>
      <version>2.17.0</version>
      <version>2.18.0</version>
    </versions>
    <lastUpdated>20241119205217</lastUpdated>
  </versioning>
</metadata>
```
2. Using them in `AddDependency` recipe, so that it honors input set to `latest.release` as taking the value from the top-level `release` field.

## What's your motivation?
Address feedback from one of the OpenRewrite users.

Yes!

1. Does the way `MavenMetadata` changes are handled provide compatibility?
2. The change in the version resolution logic is local to this single recipe (`AddDependency`) as I read this is the intent in the project. All recipes seem to have their own version matching logic.


rewrite-maven/src/main/java/org/openrewrite/maven/AddDependencyVisitor.java
```diff
@@ -187,24 +190,31 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
-            }
+    }
 
-            return resolvedVersion;
+    interface MavenMetadataSupplier {
+        MavenMetadata get() throws MavenDownloadingException;
+    }
+
+    // VisibleForTesting
+    static String findVersionToUse(@Nullable VersionComparator versionComparator, String defaultVersion, MavenMetadataSupplier mavenMetadataSupplier,
```
jevanlingen, line 200: To be honest, I don't like this architectural change (I also discussed this with @Laurens-W). I think it is better for both component isolation and your mental model to keep helper methods in its own class. So in this case the `findVersionToUse` should stay in the `InsertDependencyInOrder` visitor.

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependencyVisitor.java
```diff
@@ -16,9 +16,12 @@
 package org.openrewrite.maven;
 
 import lombok.RequiredArgsConstructor;
+import org.jetbrains.annotations.NotNull;
```
timtebeek, line 19: We try to avoid this annotation. Use JSpecify on the package instead where possible.

jevanlingen: W00ps good call! IntelliJ added at when refactoring 😢.

Laurens-W (approved):
LGTM

PR #5088, opened 2025-02-22
Fix YAML `JsonPathMatcher` for filter expressions on child mappings

knutwannheden:


knutwannheden:
When I compared the three JSONPath implementations we have (for JSON, YAML, and HCL) I noticed that they don't behave the same in this respect. From what I can tell, the YAML implementation is more correct here, but I didn't compare them in detail. It would probably be nice if we could align the implementations a bit more (not sure when and why they diverged).

It has been proposed to only have a single JSONPath parser. Likewise we could consider having only a single matching engine. But that clearly goes beyond the scope of this PR.

PR #5098, opened 2025-02-26
Prove once more maven config properties are resolved

Laurens-W:
## What's changed?

Added testcase to prove maven config properties work and are read only

@jevanlingen 

## Any additional context

Adjusted testcase proposed in #5070 that was removed

### Checklist


timtebeek:
I think the problem here might be multi module projects, where the child can not find the parent. Best explored on the Maven plugin project.

jevanlingen (approved):

Issue #5099, opened 2025-02-26
IndexOutOfBoundsException parsing valid YAML whose size is equal to StreamReader's buffer size


mkosmul:
## What version of OpenRewrite are you using?
OpenRewrite 8.40.3 but issue is also present in `main` as of 2025-02-26

## What is the smallest, simplest way to reproduce the problem?
The following test fails:

```java
    @Test
    void parseFlowSequenceAtBufferBoundary() throws IOException {
        // May change over time in SnakeYaml, rendering this test fragile
        var snakeYamlEffectiveStreamReaderBufferSize = 1024 - 1;

        @Language("yml")
        var yaml = "a: " + "x".repeat(1000) + "\n" + "b".repeat(16) + ": []";
        assertEquals(snakeYamlEffectiveStreamReaderBufferSize - 1, yaml.lastIndexOf('['));

        rewriteRun(
            spec -> spec.recipe(new DeleteKey(".nonexistent","*")),
            yaml(yaml)
        );
    }
```

The recipe used doesn't matter, it's there just to trigger the YAML parser.

This test fails with the following exception:
```
Failed to parse sources or run recipe
java.lang.AssertionError: Failed to parse sources or run recipe
	at org.openrewrite.test.RewriteTest.lambda$defaultExecutionContext$14(RewriteTest.java:638)
	at org.openrewrite.test.RewriteTest$$Lambda$509/0x0000020b8117b730.accept(Unknown Source)
	at org.openrewrite.yaml.YamlParser.lambda$parseInputs$0(YamlParser.java:76)
	at org.openrewrite.yaml.YamlParser$$Lambda$549/0x0000020b8125f340.apply(Unknown Source)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
[...]

PR #5100, opened 2025-02-26
#5099 | Fix issue 5099 - YAML parsing crashes with IndexOutOfBoundsEx…

mkosmul:
Fix for #5099 - `IndexOutOfBoundsException` when parsing YAML that has a flow-style sequence ending at the boundary of the internal buffer used by SnakeYaml StreamReader.

## What's changed?
`FormatPreservingReader.readStringFromBuffer()` tries to read one character past the end mark. This is OK as long as we have not reached the end of the buffer yet. If the token is at the end of the buffer, we need to handle that non-existent character as a special case.

## What's your motivation?
Being able to parse some valid YAML files which have caused `YamlParser` to crash so far.

Modifying the code in order to avoid checking the character at the end mark altogether - this would require some deeper changes and seems more risky, though.

### Checklist


timtebeek:
Appreciate the immediate fix here @mkosmul ! Tagging Jacob for review as he's made recent changes.

jevanlingen (approved):

PR #5107, opened 2025-02-28
reproduce-maven-config-issue

jevanlingen:
## What's changed?
Read-only maven properties (e.g. maven.config) are now separated from the normal configuration. A new field `constants` is added the to  Pom and ResolvedPom objects.

## What's your motivation?
If a recipe is run which uses the UpdateMavenModel visitor to update its Pom, the properties are wiped and recreated by retrieving them from the Xml.Document:

https://github.com/openrewrite/rewrite/blob/6f637caa7eae494852f116bffd9709d266eaf71f/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java#L42-L49

Currently, when a Pom is parsed, the `project.basedir`, `basedir` and external properties are merged together with the internal `<properties>..</properties>`:

https://github.com/openrewrite/rewrite/blob/6f637caa7eae494852f116bffd9709d266eaf71f/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java#L78-L87

These external properties will be lost once the UpdateMavenModel visitor removes all properties from the model and retrieves the current properties from the `<properties>..</properties>` part only. So I introduced a constants field, which should be an immutable map where external read-only properties are defined.

## Any additional context

Is this the right solution? Or should something else be considered? Another simpler solution would be to delete the removal of properties line:

https://github.com/openrewrite/rewrite/blob/6f637caa7eae494852f116bffd9709d266eaf71f/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java#L43

This could end up in having properties that do no longer make sense, but as these properties are obsolete, it wouldn't matter much. Notice new properties with the same key would still override the old values.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
@@ -81,10 +81,13 @@ public class MavenParser implements Parser {
                 if (pom.getProperties().isEmpty()) {
                     pom = pom.withProperties(new LinkedHashMap<>());
                 }
+
                 String baseDir = pomPath.toAbsolutePath().getParent().toString();
-                pom.getProperties().put("project.basedir", baseDir);
-                pom.getProperties().put("basedir", baseDir);
-                pom.getProperties().putAll(properties);
+                pom = pom.withConstants(new HashMap<String, String>() {{
+                    put("project.basedir", baseDir);
+                    put("basedir", baseDir);
+                    putAll(properties);
+                }});
 
                 SourceFile sourceFile = new MavenXmlParser()
                         .parseInputs(singletonList(source), relativeTo, ctx)
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
@@ -92,6 +92,9 @@ public class Pom {
     String packaging;
 
     @Builder.Default
+    Map<String, String> constants = emptyMap();
+
+    @Builder.Default
     Map<String, String> properties = emptyMap();
 
     @Builder.Default
@@ -205,6 +208,7 @@ public class Pom {
         return new ResolvedPom(
                 this,
                 activeProfiles,
+                constants,
                 properties,
                 emptyList(),
                 concatAll(initialRepositories, getEffectiveRepositories()),
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -70,26 +70,6 @@ public class ResolvedPom {
     @Builder.Default
     Iterable<String> activeProfiles = emptyList();
 
-    public ResolvedPom(Pom requested, Iterable<String> activeProfiles) {
-        this(requested, activeProfiles, requested.getConstants(), emptyMap(), emptyList(), null, emptyList(), emptyList(), emptyList(), emptyList(), emptyList(), emptyList());
-    }
-
-    @JsonCreator
-    ResolvedPom(Pom requested, Iterable<String> activeProfiles, Map<String, String> constants, Map<String, String> properties, List<ResolvedManagedDependency> dependencyManagement, @Nullable List<MavenRepository> initialRepositories, List<MavenRepository> repositories, List<MavenRepository> pluginRepositories, List<Dependency> requestedDependencies, List<Plugin> plugins, List<Plugin> pluginManagement, List<String> subprojects) {
-        this.requested = requested;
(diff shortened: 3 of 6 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -425,16 +435,7 @@ private void resolveParentPropertiesAndRepositoriesRecursively(List<Pom> pomAnce
 
             if (pom.getParent() != null) {
                 Pom parentPom = resolveParentPom(pom);
-
-                for (Pom ancestor : pomAncestry) {
```
MBoegers, line 429: Why do we have to re-resolve every parent now? To collect all values upstream?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -126,6 +131,12 @@ public ResolvedPom(Pom requested, Iterable<String> activeProfiles) {
     @Nullable // on older LSTs, this field is not yet present
     List<String> subprojects = emptyList();
 
+    public Map<String, String> getProperties() {
```
MBoegers, line 134: How often is this called? can we cache the result to prevent creating a new HashMap everytime?

rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
```diff
@@ -81,10 +81,13 @@ public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path re
...
                 }
+
                 String baseDir = pomPath.toAbsolutePath().getParent().toString();
-                pom.getProperties().put("project.basedir", baseDir);
-                pom.getProperties().put("basedir", baseDir);
-                pom.getProperties().putAll(properties);
+                pom = pom.withConstants(new HashMap<String, String>() {{
+                    put("project.basedir", baseDir);
+                    put("basedir", baseDir);
+                    putAll(properties);
```
MBoegers, line 89: Should the properties really be considered as constants? They can change from recipe to recipe.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
jevanlingen: Actually this is an another improvement I accidentally added in this PR. But no bother, because it is better. To answer your question we do already re-resolve every parent, I just remove the unnecessary List wrapper.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
jevanlingen: Yeah, this is the one part I am not happy with 😞. Maybe we can cache it, but I fear that would lead to another can of worms when we forget to update the cache. So that's why I just left it here like this.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
jevanlingen: Uhm I am not absolutely sure, but these 'external properties' should never be changed by a recipe, right?

timtebeek: Perhaps good to follow the Maven naming here; these are user properties coming from the MavenSession, and could be coming from .mvn/maven.config, or explicit arguments passed in. That's also why I'd want to get them from the session in the plugin, as opposed to from parsing the config file (same with profiles: use the session in the plugin instead there too).
https://maven.apache.org/ref/3.9.9/maven-core/apidocs/org/apache/maven/execution/MavenSession.html#getUserProperties()

rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
```diff
@@ -91,6 +91,9 @@ public static int getModelVersion() {
     @Nullable
     String packaging;
 
+    @Builder.Default
+    Map<String, String> constants = emptyMap();
```
timtebeek, line 95: I'll explore some more where I can, but I continue to be of the opinion that we should add these properties to the MavenResolutionResult Marker, not the Pom. The tree package classes are there to model what we saw in the source files; the markers can add information pulled in from the environment, like in this case or with active profiles.

timtebeek:
- Possible replacement: https://github.com/openrewrite/rewrite/pull/5114

jevanlingen:
> Possible replacement: https://github.com/openrewrite/rewrite/pull/5114

Yeah, looks better than this one 👌🏻. I copied the `mavenConstantsExistAfterUpdateMavenModel` into that branch (as it was part of the fix).

PR #5114, opened 2025-03-02
Store MavenSession.userProperties in MavenResolutionResult

timtebeek:
- https://github.com/openrewrite/rewrite-maven-plugin/pull/955

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindDependencyTest.java
@@ -17,9 +17,11 @@ package org.openrewrite.gradle.search;
 
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
+import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.gradle.Assertions.buildGradle;
+import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
 
 class FindDependencyTest implements RewriteTest {
 
@@ -39,7 +41,7 @@ class FindDependencyTest implements RewriteTest {
               }
               
               dependencies {
-                  api 'org.openrewrite:rewrite-core:latest.release'
+                  api "org.openrewrite:rewrite-core:latest.release"
               }
               """,
             """
@@ -52,7 +54,7 @@ class FindDependencyTest implements RewriteTest {
               }
               
               dependencies {
-                  /*~~>*/api 'org.openrewrite:rewrite-core:latest.release'
+                  /*~~>*/api "org.openrewrite:rewrite-core:latest.release"
               }
               """
           )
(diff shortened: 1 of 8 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
```diff
@@ -240,11 +248,6 @@ private Builder mavenConfig(@Nullable String mavenConfig) {
                             String[] profiles = matcher.group(1).split(",");
                             activeProfiles(Arrays.stream(profiles).map(String::trim).toArray(String[]::new));
                         }
-                    } else if (line.startsWith("-D")) {
-                        Matcher matcher = Pattern.compile("-D(.*?)=(.*)").matcher(line);
-                        if (matcher.find()) {
-                            property(matcher.group(1).trim(), matcher.group(2).trim());
-                        }
```
timtebeek, line 247: No need to (fragile) parse what Maven has already pulled from command line flags or maven.config file.

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java
```diff
@@ -38,15 +38,16 @@ public Xml visitDocument(Xml.Document document, P p) {
...
 
         Optional<Xml.Tag> properties = document.getRoot().getChild("properties");
-        requested.getProperties().clear();
         if (properties.isPresent()) {
             for (final Xml.Tag propertyTag : properties.get().getChildren()) {
                 requested.getProperties().put(propertyTag.getName(),
                         propertyTag.getValue().orElse(""));
             }
         }
+        requested.getProperties().putAll(resolutionResult.getUserProperties());
```
timtebeek, line 50: My thinking here is that user properties should override whatever is defined in pom files.

jkschneider, line 50: This will be `null` on old LSTs deserializing even though the field is non-null, so you should do a null check on this here.

jkschneider (changes requested):

rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java
jkschneider: Yes, agreed.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
```diff
@@ -69,12 +70,21 @@ public class MavenResolutionResult implements Marker {
...
+    @With
+    Map<String, String> userProperties;
+
     public List<String> getActiveProfiles() {
         // for backwards compatibility with ASTs that were serialized before activeProfiles was added
         //noinspection ConstantValue
         return activeProfiles == null ? emptyList() : activeProfiles;
     }
 
+    public Map<String, String> getUserProperties() {
```
knutwannheden, line 82: An alternative to this would be to add a package private @JsonCreator annotated constructor with a nullable field, which dies the defaulting.

Having a comment which we can use to later find and remove that constructor (or this method) would also be nice.

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java
jevanlingen: Well, I am in doubt... I think you can make the case for both this and the other way around. Let's talk about this!

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -3738,19 +3733,15 @@ void propertyFromMavenConfig() {
     @Test
     void propertyFromMavenConfigFromParentPomCanBeUsedInChild() {
         rewriteRun(
+          spec -> spec.parser(MavenParser.builder().property("revision", "1.0.0")),
```
jevanlingen, line 3736: Though I agree with this kind of change, I still want a test to prove the maven-parser can handle `maven.config` files. Can we still make such a test work?

timtebeek: That should be tested in the Maven plugin, if at all, as it's just a plain text file. No need to test that here.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
timtebeek: Yes I'd went for this approach as there was already a no args constructor with that creator annotation.
https://github.com/openrewrite/rewrite/pull/5114/files/72e88c7cc8b23ac8706a9fd353440fa73d13d3a7#diff-2cf9cbe4851aaf4690268bae848441a39ce782334d3e1350d1a5976076cc388eR39

Figured best then to copy the approach we already had for active profiles.

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java
timtebeek: This looks to confirm that command line overrides pom.
https://stackoverflow.com/a/13709976

Laurens-W:
https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java#L85-L86
these two properties are still getting lost upon re resolving the pom. Instead I suggest we move them to the maven plugin: https://github.com/openrewrite/rewrite-maven-plugin/pull/955/commits/cb3e3a64cb63bf56e5409deec18fc71eed2446c3

Laurens-W (approved):

rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindDependencyTest.java
timtebeek, line 1: This seems unrelated to this PR right?

Issue #5117, opened 2025-03-03
RemoveDependency not work when dependency has type=zip


ckcd:
## What version of OpenRewrite are you using?

I am using the latest rewrite (v8.47.3)

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```shell
mvn org.openrewrite.maven:rewrite-maven-plugin:run ...
```

## What is the smallest, simplest way to reproduce the problem?

Run this test case, it will failed. The `RemoveDependency` should remove `org.elasticsearch.distribution.integ-test-zip:elasticsearch` but actually it did not. Only because the dependency uses `type=zip`.

```java
@Test
void removeDependencyWithTypeZip() {
    rewriteRun(
      spec -> spec.recipe(new RemoveDependency("org.elasticsearch.distribution.integ-test-zip", "elasticsearch", null)),
      pomXml(
        """
          <project>
            <modelVersion>4.0.0</modelVersion>
            
            <groupId>com.mycompany.app</groupId>
            <artifactId>my-app</artifactId>
            <version>1</version>
            
            <dependencies>
              <dependency>
                <groupId>com.google.guava</groupId>
                <artifactId>guava</artifactId>
                <version>29.0-jre</version>
              </dependency>
              <dependency>
                <groupId>org.elasticsearch.distribution.integ-test-zip</groupId>
                <artifactId>elasticsearch</artifactId>
                <version>7.17.15</version>
                <type>zip</type>
              </dependency>
            </dependencies>
          </project>
          """,
        """
          <project>
            <modelVersion>4.0.0</modelVersion>
            
            <groupId>com.mycompany.app</groupId>
            <artifactId>my-app</artifactId>
            <version>1</version>
            
            <dependencies>
              <dependency>
                <groupId>com.google.guava</groupId>
                <artifactId>guava</artifactId>
                <version>29.0-jre</version>
              </dependency>
            </dependencies>
          </project>
          """
      )
    );
}
```

## What did you expect to see?

The `RemoveDependency` should remove `org.elasticsearch.distribution.integ-test-zip:elasticsearch` but actually it did not.

## What did you see instead?

```
java.lang.AssertionError: Recipe was expected to make a change but made no changes.
```

## What is the full stack trace of any errors you encountered?

```
java.lang.AssertionError: Recipe was expected to make a change but made no changes.

[...]

PR #5118, opened 2025-03-03
fix: ResolvedPom with type=zip

ckcd:
## What's changed?

- Fix https://github.com/openrewrite/rewrite/issues/5117

The reason for this bug should be we ignore parse pom when `type=zip`.
https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java#L893

I think currently the check is too restrict and for `type=zip` its ok to do after pom parse. For example for the dependency in my test case:
```xml
<dependency>
    <groupId>org.elasticsearch.distribution.integ-test-zip</groupId>
    <artifactId>elasticsearch</artifactId>
    <version>7.17.15</version>
    <type>zip</type>
</dependency>
```

Its `pom.xml` is:
https://repo1.maven.org/maven2/org/elasticsearch/distribution/integ-test-zip/elasticsearch/7.17.15/elasticsearch-7.17.15.pom

## What's your motivation?

Fix https://github.com/openrewrite/rewrite/issues/5117

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -890,7 +890,7 @@ public class ResolvedPom {
                         throw new MavenDownloadingException("No version provided", null, dd.getDependency().getGav());
                     }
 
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks again for the quick fix @ckcd !

PR #5120, opened 2025-03-03
MergeYaml recipe should not do anything when key does not match

jevanlingen:
## What's changed?
The MergeYaml recipe added new keys when there was no matching key:

```java
@Test
void keyHasNoMatch() {
    rewriteRun(spec -> spec
        .recipe(new MergeYaml("$.foo",
          """
            new-key: new-value
            """, false, null, null, null)),
      yaml("some: thing")
    );
}
``` 

## What's your motivation?
Although this add-new-key-when-key-does-not-match behavior is intentionally programmed this way, after more thought, we do not want it. The user of the recipe can achieve the very same result by using the root key (which is `$`).  Conversely, if another key is specified, the user might or might not mean to add the specified yaml. Therefore, we decided to remove this behavior.

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -130,7 +130,6 @@ public class MergeYaml extends Recipe {
         return "Merge a YAML snippet with an existing YAML document.";
     }
 
-    final static String FOUND_MATCHING_ELEMENT = "FOUND_MATCHING_ELEMENT";
     final static String REMOVE_PREFIX = "REMOVE_PREFIX";
 
     @Override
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
```diff
@@ -80,6 +80,11 @@ public class MergeYaml extends Recipe {
     @Nullable
     String insertProperty;
 
+    // Although the add-new-key-when-key-does-not-match behaviour was intentionally programmed, after more thought, we do not want it anymore,
+    // as the user of this recipe can achieve the very same result by using the root key (which is $).
+    // This boolean is intended for imperative recipes that rely on this old behavior.
+    boolean createNewKeys;
```
jevanlingen, line 86: I am starting to doubt this decision. Changing this behaviour has impact on other recipes (see my changes for the CopyValue recipe). It could very well be true customer recipes also use the MergeYaml recipe as a base. So I opt to go for a Boolean @Option instead, which by default uses the current behaviour.

Laurens-W: Perhaps this could be covered by a precondition rather than adding another option to this recipe? I don't know what precondition recipes we have for yaml though

jevanlingen: No, because whether you want to add a key or not is not something you can infer from the incoming/existing yaml (or any other object). What matters is the intent of the user's use of the recipe..

rewrite-yaml/src/main/java/org/openrewrite/yaml/CopyValue.java
```diff
@@ -113,25 +113,19 @@ public Yaml.Mapping.Entry visitMappingEntry(Yaml.Mapping.Entry entry, ExecutionC
...
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
-        if(acc.snippet == null) {
+        if (acc.snippet == null) {
             return TreeVisitor.noop();
         }
-        TreeVisitor<?, ExecutionContext> visitor = new MergeYaml(newKey, acc.snippet, false, null, null, null, null).getVisitor();
-        if(newFilePath == null) {
-            visitor = Preconditions.check(new FindSourceFiles(acc.path.toString()).getVisitor(), visitor);
```
jevanlingen, line 130: 💡 No need to do the precondition check here, as it's already in the MergeYaml visitor too:

```
@Override
  ..
  return Preconditions.check(new FindSourceFiles(filePattern), new YamlIsoVisitor<ExecutionContext>() {
}

MBoegers: But copy value is an independent recipe so it should have its own precondition, right?

jevanlingen: Ah you are right! At the moment, this works and is indeed "double the same work". But as soon as  the MergeYaml converges a bit, this change could cause problems. I'll revert it, good catch!

MBoegers (approved):
Please review your formatter settings, I see a lot `if(...)` to `if (...)` which is very noisy

Laurens-W:
@MBoegers we reviewed our formatter settings vs Tim's before and they are the same, so either IntelliJ changed their default or this was previously unformatted/formatted incorrectly :)

PR #5124, opened 2025-03-04
Fully qualified reference gets shortened

Laurens-W:
## What's changed?

Add reproducer

@MBoegers 

### Checklist


MBoegers:
Thanks for the moving the failing test here. @knutwannheden implemented it exactly this way here https://github.com/openrewrite/rewrite/commit/581546afbde93dc7559c3d74201d01b3cbfac31b
Let us double-check if the original test is not testing something complete different.

Laurens-W:
Actually these expectations are invalid since 
- https://github.com/openrewrite/rewrite/pull/5080

PR #5137, opened 2025-03-06
Fix rewrite-kotlin tests

Laurens-W:
## What's changed?

Fixed the build

## What's your motivation?

Failing build

@knutwannheden 

### Checklist

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -128,7 +128,7 @@ public class KotlinParser implements Parser {
 
     @Override
     public Stream<SourceFile> parse(@Language("kotlin") String... sources) {
-        Pattern packagePattern = Pattern.compile("\bpackage\\s+([`.\\w]+)");
+        Pattern packagePattern = Pattern.compile("\\bpackage\\s+([`.\\w]+)");
         Pattern classPattern = Pattern.compile("(class|interface|enum)\\s*(<[^>]*>)?\\s+(\\w+)");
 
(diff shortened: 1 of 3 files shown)


knutwannheden (approved):

knutwannheden (approved):

knutwannheden (approved):

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -128,7 +128,7 @@ public class KotlinParser implements Parser {
 
     @Override
     public Stream<SourceFile> parse(@Language("kotlin") String... sources) {
-        Pattern packagePattern = Pattern.compile("^package\\s+([^;]+);");
+        Pattern packagePattern = Pattern.compile("\\bpackage\\s+([`.\\w]+)");
```
shanman190, line 131: FYI, @Laurens-W, this same issue is present in the `GroovyParser` as well. I had fixed it by looking for `[^\r\n;]+`, but regardless we should apply this same fix to the `GroovyParser` as well.

I don't think we have a test case on the Groovy side that just so happens to trigger the error condition like with Kotlin.

If you don't grab it, I can make sure that it's included as part of my PR when I get it completed.

/cc @knutwannheden

Laurens-W: I'll make a separate PR and introduce a testcase as well!
Should we make the regex's equal, and if so which one do we prefer?

shanman190: Your regex would also account for a license header, so I think it's probably preferable.

Issue #5139, opened 2025-03-06
Failure while parsing switch expressions with combined clauses null and default


froque:
## What version of OpenRewrite are you using?

I am using

- Maven/Gradle plugin LATEST (v6.2.2)

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.

## What is the smallest, simplest way to reproduce the problem?

```xml
❯ cat pom.xml                                                                    
<project>
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.mycompany.app</groupId>
  <artifactId>my-app</artifactId>
  <version>1</version>

  <properties>
      <maven.compiler.target>21</maven.compiler.target>
      <maven.compiler.source>21</maven.compiler.source>
      <maven.compiler.release>21</maven.compiler.release>
  </properties>
</project> 
```
```java
❯ cat src/main/java/Foo.java
public class Foo {
	public static void main(String[] args) {
		Number number = null;
		switch (number) {
			case null, default -> System.out.println("something");
		}
	}
}
```

```
JAVA_HOME=/usr/lib/jvm/temurin-23-jdk-amd64 \
          /opt/maven/apache-maven-3.9.9/bin/mvn \
            -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN  \
              org.openrewrite.maven:rewrite-maven-plugin:LATEST:run \
                -Drewrite.activeRecipes=org.openrewrite.maven.OrderPomElements  
```

## What did you expect to see?

I expected rewrite to not throw an error. 

## What did you see instead?

An error parsing the Java code. 

## What is the full stack trace of any errors you encountered?

```
JAVA_HOME=/usr/lib/jvm/temurin-23-jdk-amd64 \
          /opt/maven/apache-maven-3.9.9/bin/mvn \
            -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN  \
              org.openrewrite.maven:rewrite-maven-plugin:LATEST:run \
                -Drewrite.activeRecipes=org.openrewrite.maven.OrderPomElements  
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[...]

PR #5140, opened 2025-03-06
Fixes parsing of switch expression with combined null and default

froque:
## What's changed?

Adds unit test and fix for parsing of switch expression.

## What's your motivation?

- provide a fix for #5139 

Changing the Java code, splitting the combined case. 

### Checklist


sambsnyd:
Thank you @froque !

PR #5141, opened 2025-03-07
Use GZIP rather than raw zlib

knutwannheden:
Using GZIP adds a small header and a checksum at the end and is a bit "friendlier" as it allows a developer to use `gzcat` to inspect the file contents.


Laurens-W (approved):
Perhaps we should also change the JavaDoc above this class to reflect or mention this change

timtebeek (approved):

PR #5142, opened 2025-03-07
Fix groovy parser parsing package definition

Laurens-W:
## What's changed?

Regex for parsing package definition in Groovy now correctly handles the optional semicolon

@knutwannheden @shanman190 

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
@@ -62,7 +62,7 @@ public class GroovyParser implements Parser {
 
     @Override
     public Stream<SourceFile> parse(@Language("groovy") String... sources) {
-        Pattern packagePattern = Pattern.compile("^package\\s+([^;]+);");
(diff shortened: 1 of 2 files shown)


shanman190 (approved):

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
```diff
@@ -62,7 +62,7 @@ public class GroovyParser implements Parser {
 
     @Override
     public Stream<SourceFile> parse(@Language("groovy") String... sources) {
-        Pattern packagePattern = Pattern.compile("^package\\s+([^;]+);");
+        Pattern packagePattern = Pattern.compile("\\bpackage\\s+([.\\w]+)");
         Pattern classPattern = Pattern.compile("(class|interface|enum)\\s*(<[^>]*>)?\\s+(\\w+)");
```
shanman190, line 66: This ironically makes me wonder if we don't support Groovy trait types in various capacities.
https://groovy-lang.org/objectorientation.html#_traits

PR #5143, opened 2025-03-07
There're no references to classes of `com.gradle:gradle-enterprise-gradle-plugin` in the Gradle module

bmuschko:
## What's changed?

The dependency on `com.gradle:gradle-enterprise-gradle-plugin` has been removed from the Gradle module.

## What's your motivation?

The Gradle module doesn't use any of the classes brought in by the dependency AFAIKT. We are using type tables here now so the dependency wouldn't be added as a transitive anymore though there's not point in having it's classes aren't used.

For reference, the two recipes that would potentially use the dependency:

- https://docs.openrewrite.org/recipes/gradle/plugins/adddevelocitygradleplugin
- https://docs.openrewrite.org/recipes/gradle/plugins/migrategradleenterprisetodevelocity

No

No

### DIFF
--- a/rewrite-gradle/build.gradle.kts
+++ b/rewrite-gradle/build.gradle.kts
@@ -32,7 +32,6 @@ recipeDependencies {
     parserClasspath("org.gradle:gradle-resources:latest.release")
     parserClasspath("org.gradle:gradle-testing-base:latest.release")
     parserClasspath("org.gradle:gradle-testing-jvm:latest.release")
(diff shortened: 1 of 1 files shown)


knutwannheden:
There is still a reference to the Gradle Enterprise classes in a Groovy script we have for Gradle type attribution support. But the dependency is a compile-only dependency. The important thing is that the gradle-enterprise.jar has been removed from the rewrite-gradle.jar.

The dependency you are removing is used to create the type table, so I think we should leave that in so that we can recreate the type table, when required.

Issue #5146, opened 2025-02-26
semicolon removed in enum w/o values (observed with upgrade rewrite-maven-plugin from 6.2.0 to 6.2.1)


pzygielo:
## What version of OpenRewrite are you using?
I am using

- Maven plugin v6.2.1

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a multi-module project.

## What is the smallest, simplest way to reproduce the problem?
Original code:
```java
     public enum A {
         ;
         public static final String X = "receipt-id";
     }
```

- https://github.com/pzrep/rewrite-maven-plugin-953 -> https://github.com/pzrep/rewrite-maven-plugin-953/pull/1 -> https://github.com/pzrep/rewrite-maven-plugin-953/actions/runs/13540330563/job/37839773851?pr=1#step:6:25

## What did you expect to see?
No patch created by openrewrite.

## What did you see instead?
`;` removed which caused `Compilation failure: enum constant expected here`

Seems to be a regression seen after
- https://github.com/openrewrite/rewrite/pull/5063

timtebeek:
@jevanlingen / @greg-at-moderne Could you have a look at this regression?

greg-at-moderne:
Thanks for reporting this.

- this should fix the Java parsing: #5147 
- and this adds a test case to rewrite-static-analysis: https://github.com/openrewrite/rewrite-static-analysis/pull/474

pzygielo:
As far as I am concerned, this works fine now in rewrite-maven-plugin:6.3.0.
Thank you.

PR #5147, opened 2025-03-10
Handling enums with no values as J.EnumValueSet

greg-at-moderne:
## What's changed?

Fixing how Java enums with no values are parsed. Now they get the `J.EnumValueSet` element with no values. Previously they were parsed as `J.Empty`.

## What's your motivation?

- addressing #5146

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
@@ -458,6 +458,11 @@ public class ReloadableJava11ParserVisitor extends TreePathScanner<J, Space> {
                     ),
                     EMPTY
             );
+        } else if (kind.getType() == J.ClassDeclaration.Kind.Type.Enum) {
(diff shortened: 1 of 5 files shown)


rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
```diff
@@ -458,6 +458,11 @@ public J visitClass(ClassTree node, Space fmt) {
                     ),
                     EMPTY
             );
+        } else if (kind.getType() == J.ClassDeclaration.Kind.Type.Enum) {
+            if (positionOfNext(";", null) >= 0) {
+                Space prefix = sourceBefore(";");
+                enumSet = padRight(new J.EnumValueSet(randomId(), prefix, Markers.EMPTY, emptyList(), true), EMPTY);
```
jevanlingen, line 464: Personally I would inline this:
```suggestion
                enumSet = padRight(new J.EnumValueSet(randomId(), sourceBefore(";"), Markers.EMPTY, emptyList(), true), EMPTY);
```

But I'll leave that up to you.

PR #5148, opened 2025-03-10
Pin AssertJ for `rewrite-java-8` to 3.x

knutwannheden:
For `rewrite-java-8`, `rewrite-java-11`, and `rewrite-java-tck` we need to use AssertJ 3.x, as AssertJ 4.x now requires Java 17.

### DIFF
--- a/rewrite-java-11/build.gradle.kts
+++ b/rewrite-java-11/build.gradle.kts
@@ -18,7 +18,6 @@ dependencies {
     implementation("org.ow2.asm:asm:latest.release")
 
     testImplementation(project(":rewrite-test"))
-    testImplementation("org.assertj:assertj-core:latest.release")
     "javaTck"(project(":rewrite-java-tck"))
 }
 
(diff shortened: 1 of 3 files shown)


greg-at-moderne (approved):
I understand resolutionStrategy is used, because we don't want to pin any specific version, just provide `3.+` constraint.

knutwannheden:
@sambsnyd Maybe you can also have a look at this and let me know how you think we should best be handling this. Especially with regards to `rewrite-test`.

knutwannheden:
> I understand resolutionStrategy is used, because we don't want to pin any specific version, just provide `3.+` constraint.

In Maven the closest dependency will control the version that actually gets used, whereas in Gradle it is the highest version of any of the dependencies. As always there are probably different ways to achieve the same result. I am not a Gradle expert and went for this approach without thinking much about if there are better alternatives. But I did not want to pin it to a fixed version, in case there are some criticial bug fixes to the 3.x branch.

PR #5153, opened 2025-03-11
Class files from type tables should not include all nested types

knutwannheden:
When we write the `.class` files for a `TypeTable`, the file should only contain information about the nested types which also belong to that type.

Before this change all `.class` files would contain information about all nested types, which could result in very large `.class` files.

For the `develocity-gradle-plugin` dependency the `TypeTable.Reader` would end up writing 2.7G worth of class files. With this change this drops down to 32M. At the same time this operation now also takes a lot less time to complete.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
@@ -17,6 +17,7 @@ package org.openrewrite.java.internal.parser;
 
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
+import lombok.experimental.NonFinal;
 import org.jspecify.annotations.Nullable;
 import org.objectweb.asm.*;
 import org.objectweb.asm.util.CheckClassAdapter;
@@ -39,6 +40,7 @@ import java.util.zip.GZIPOutputStream;
 import java.util.zip.InflaterInputStream;
 import java.util.zip.ZipException;
 
+import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 import static org.objectweb.asm.ClassReader.SKIP_CODE;
 import static org.objectweb.asm.Opcodes.V1_8;
@@ -144,6 +146,11 @@ public class TypeTable implements JavaParserClasspathLoader {
      */
     @RequiredArgsConstructor
     static class Reader {
+        private static final int NESTED_TYPE_ACCESS_MASK = Opcodes.ACC_PUBLIC | Opcodes.ACC_PRIVATE | Opcodes.ACC_PROTECTED |
+                                                           Opcodes.ACC_STATIC | Opcodes.ACC_FINAL | Opcodes.ACC_INTERFACE |
+                                                           Opcodes.ACC_ABSTRACT | Opcodes.ACC_SYNTHETIC | Opcodes.ACC_ANNOTATION |
+                                                           Opcodes.ACC_ENUM;
+
         private final ExecutionContext ctx;
 
         public void read(InputStream is, Collection<String> artifactNames) throws IOException {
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -144,6 +146,11 @@ private static Collection<String> artifactsNotYetWritten(Collection<String> arti
      */
     @RequiredArgsConstructor
     static class Reader {
+        private static final int NESTED_TYPE_ACCESS_MASK = Opcodes.ACC_PUBLIC | Opcodes.ACC_PRIVATE | Opcodes.ACC_PROTECTED |
+                                                           Opcodes.ACC_STATIC | Opcodes.ACC_FINAL | Opcodes.ACC_INTERFACE |
+                                                           Opcodes.ACC_ABSTRACT | Opcodes.ACC_SYNTHETIC | Opcodes.ACC_ANNOTATION |
+                                                           Opcodes.ACC_ENUM;
+
```
jevanlingen, line 153:
```suggestion
        private static final int NESTED_TYPE_ACCESS_MASK = ACC_PUBLIC | ACC_PRIVATE | ACC_PROTECTED | ACC_STATIC | ACC_FINAL | ACC_INTERFACE | ACC_ABSTRACT | ACC_SYNTHETIC | ACC_ANNOTATION | ACC_ENUM;
```

💡 Using static imports is a code style that some are against and some are into. I myself lean very much into the "pro" side (types don't matter much less than people think when you _read_ the code).

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -235,24 +252,17 @@ private void writeClassesDir(@Nullable GroupArtifactVersion gav, Map<ClassDefini
                         classDef.getSuperinterfaceSignatures()
                 );
 
-                Set<ClassDefinition> innerClasses = new HashSet<>();
-                for (ClassDefinition innerClassDef : membersByClassName.keySet()) {
```
knutwannheden, line 239: Note: This here was the problematic part. I could also only have made changes to this loop, but this quadratic overhead didn't feel right.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -182,17 +192,24 @@ public void read(InputStream is, Collection<String> artifactNames) throws IOExce
...
+                                new ClassDefinition(
+                                        Integer.parseInt(fields[3]),
+                                        name,
+                                        fields[5].isEmpty() ? null : fields[5],
+                                        fields[6].isEmpty() ? null : fields[6],
+                                        fields[7].isEmpty() ? null : fields[7].split("\\|")
+                                ));
+                        int lastIndexOf$ = classDefinition.getName().lastIndexOf('$');
+                        if (lastIndexOf$ != -1) {
+                            String ownerName = classDefinition.getName().substring(0, lastIndexOf$);
```
jevanlingen, line 206: Wouldn't `className.substring(0, lastIndexOf$)` work here as well?

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
knutwannheden: I didn't change this because in other places the `Opcodes` constants are referenced in a qualified form, with the exception of the `V1_8` constant.

jevanlingen: I know we tend to write static constants with its type in most of our code, I just don't like it. But as said, this is highly subjective,  so leave it as it is if that suits you better.

_Just wanted to comment on this, it's kinda the same discussion if you should use `var` or type out the type._

jevanlingen (approved):

PR #5160, opened 2025-03-12
Adding a missing return, without which a supposed optimization does nothing

mkosmul:
## What's changed?
Added a missing return statement in `ChangeType.updateOuterClassTypes()`.

## What's your motivation?
The comment in line 419 says "just a performance shortcut when this isn't an inner class", but without a return statement, the line only creates a modified object which is then discarded.

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases (nope, since this is just a minor performance improvement)


timtebeek:
Oh wow, that's been in there since 2021; thanks for noticing & immediately jumping to  a fix. Looks good to me, hope it helps performance from now on.

PR #5162, opened 2025-03-12
Origins information in `RecipeDescriptor`

MBoegers:
## What's changed?

The Enviroment loads the GitHub URL and LIcense information per Recipe from provided resources. The resouces are generated here https://github.com/openrewrite/rewrite-build-gradle-plugin/pull/83. 

## What's your motivation?

The Documentation uses a heuristic approach to determine which the GitHub URL and correct license information. 

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
@@ -46,7 +46,7 @@ public class ClasspathScanningLoader implements ResourceLoader {
 
     private final Map<String, List<Contributor>> recipeAttributions = new HashMap<>();
     private final Map<String, List<RecipeExample>> recipeExamples = new HashMap<>();
-    private final Map<String, License> recipeLicense = new HashMap<>();
+    private final Map<String, RecipeOrigin> recipeOrigins = new HashMap<>();
 
     /**
      * Construct a ClasspathScanningLoader scans the runtime classpath of the current java process for recipes
@@ -127,10 +127,10 @@ public class ClasspathScanningLoader implements ResourceLoader {
                 styles.addAll(resourceLoader.listStyles());
                 recipeAttributions.putAll(resourceLoader.listContributors());
                 recipeExamples.putAll(resourceLoader.listRecipeExamples());
-                recipeLicense.putAll(resourceLoader.listLicenses());
+                recipeOrigins.putAll(resourceLoader.listRecipeOrigins());
             }
             for (YamlResourceLoader resourceLoader : yamlResourceLoaders) {
-                recipeDescriptors.addAll(resourceLoader.listRecipeDescriptors(recipes, recipeAttributions, recipeExamples, recipeLicense));
+                recipeDescriptors.addAll(resourceLoader.listRecipeDescriptors(recipes, recipeAttributions, recipeExamples, recipeOrigins));
             }
         }
     }
@@ -210,8 +210,8 @@ public class ClasspathScanningLoader implements ResourceLoader {
     }
 
     @Override
-    public Map<String, License> listLicenses() {
-        return recipeLicense;
+    public Map<String, RecipeOrigin> listRecipeOrigins() {
+        return recipeOrigins;
     }
 
     @Override
--- a/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
@@ -46,6 +46,7 @@ public class ClasspathScanningLoader implements ResourceLoader {
 
     private final Map<String, List<Contributor>> recipeAttributions = new HashMap<>();
     private final Map<String, List<RecipeExample>> recipeExamples = new HashMap<>();
+    private final Map<String, License> recipeLicense = new HashMap<>();
 
     /**
      * Construct a ClasspathScanningLoader scans the runtime classpath of the current java process for recipes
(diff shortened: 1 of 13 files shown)


rewrite-core/src/main/java/org/openrewrite/config/License.java
```diff
@@ -0,0 +1,31 @@
...
+    private final String fullName;
+    private final String url;
+
+    public static License fromFullName(String name) {
+        for (License license : values()) {
+            if (license.fullName.equalsIgnoreCase(name)) {
+                return license;
+            }
+        }
+        throw new IllegalArgumentException("Invalid license name: " + name);
```
knutwannheden, line 20: Maybe users would like to specify a different license?

MBoegers: Yes, that need to be covered. Cheers

rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
```diff
@@ -63,5 +63,5 @@ public class RecipeDescriptor {
 
     List<RecipeExample> examples;
 
-    URI source;
+    RecipeOrigin origin;
```
timtebeek, line 66: With the new manifest approach is there any reason to logically group license and source location together? Or can we just add a new field here for the license?

MBoegers: But keep in mind, we should support License Name and URL to support custom (i.e. User Defined) licenses as 
well.

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -45,20 +39,25 @@
...
 
 public class YamlResourceLoader implements ResourceLoader {
     int refCount = 0;
 
     private static final PropertyPlaceholderHelper propertyPlaceholderHelper =
             new PropertyPlaceholderHelper("${", "}", ":");
 
+    @Nullable
+    private final Attributes attributes;
     private final URI source;
```
timtebeek, line 61:
```suggestion
    private final Attributes attributes;
    
    private final URI source;
```

rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -170,7 +175,7 @@ private void configureRecipes(ScanResult result, String className) {
             Timer.Builder builder = Timer.builder("rewrite.scan.configure.recipe");
             Timer.Sample sample = Timer.start();
             try {
-                Recipe recipe = constructRecipe(recipeClass);
+                Recipe recipe = constructRecipe(recipeClass).augmentRecipeDescriptor(artifactManifestAttributes.get(classInfo.getClasspathElementURI().toString()));
```
MBoegers, line 178: I see your Idea here for a fluent API. But ATM all other data is set by calling the respective `set` methode. We should stick to this pattern for now.

rewrite-core/src/main/java/org/openrewrite/config/License.java
```diff
@@ -0,0 +1,52 @@
...
+        if (url != null) {
+            switch (url) {
+                case APACHEV2_URL:
+                    return APACHE_V2;
+                case MSAL_URL:
+                    return MODERNE_SOURCE_AVAILABLE;
+                case MOD_PROPRIETARY_URL:
+                    return MODERNE_PROPRIETARY;
+            }
+            if (fullName != null) {
```
MBoegers, line 51: We have a prolem here. We depend on `url` for `equals` and `hashCode` and I am good with this. But ATM we are allowing `url` to be `null`. I suggest guarding `url` to be not null and add the contract `@Contract(value = "null, null -> null; !null, null -> null; null, !null -> new; !null, !null -> new", pure = true))`

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -45,20 +39,26 @@
...
+import static org.openrewrite.config.CategoryTree.CORE;
 
 public class YamlResourceLoader implements ResourceLoader {
     int refCount = 0;
 
     private static final PropertyPlaceholderHelper propertyPlaceholderHelper =
             new PropertyPlaceholderHelper("${", "}", ":");
 
+    @Nullable
+    private final Attributes attributes;
```
MBoegers: lets refer here to the manifest ;)

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
```diff
@@ -200,9 +223,9 @@ private Collection<Map<String, Object>> loadResources(ResourceType resourceType)
 
     @SuppressWarnings("unchecked")
     @Override
-    public Collection<Recipe> listRecipes() {
+    public Collection<DeclarativeRecipe> listRecipes() {
```
MBoegers, line 226: I keep wondering why we are keep changing the API

rewrite-core/src/main/java/org/openrewrite/config/License.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.config;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+
+@Value
```
MBoegers, line 25: With this we have two ways to create a License and lose our guards. 
We should relay on the `of` method as it uses the instances from L29-331

MBoegers, line 25: With `@Value` we crate an alternative way to create a licence and losing our guards.
`@AllArgsConstructor(access = AccessLevel.PRIVATE)`

rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
jevanlingen: It's a code improvement. I changed the API:
```diff
-  Collection<Recipe> listRecipes();
+  Collection<? extends Recipe> listRecipes();
```

This gives the benefit, as you see here, to use subclasses as result, which in turn can make your code the code a little better:
```diff
public Collection<RecipeDescriptor> listRecipeDescriptors(Collection<Recipe> externalRecipes,
                                                              Map<String, List<Contributor>> recipeNamesToContributors,
                                                              Map<String, List<RecipeExample>> recipeNamesToExamples) {
-  Collection<Recipe> internalRecipes = listRecipes();
+  Collection<DeclarativeRecipe> internalRecipes = listRecipes();

   // ..

-  for (DeclarativeRecipe recipe : internalRecipes) {
+  for (Recipe recipe : internalRecipes) {
-    DeclarativeRecipe declarativeRecipe = (DeclarativeRecipe) recipe;
     // ..
  }
```

jevanlingen (approved):

PR #5163, opened 2025-03-12
Find type field references in JavaDoc in interfaces and superclasses

jevanlingen:
## What's changed?
Field type references in superclasses and interfaces are now discovered as well.

## What's your motivation?
Having a setup like 
[Xhtml5BaseSink](https://github.com/apache/maven-doxia/blob/ede159d464cebf5fb17131794cac0b48d7f21f5a/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L372):

```java
import javax.swing.text.html.HTML.Tag;
                
public interface HtmlMarkup {
  Tag H1 = Tag.H1;
}

interface Test extends HtmlMarkup {
    /**
     * @see #H1
     */
    void onSectionTitle();
}
```

would not find the `H1` type.

### Checklist

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
@@ -748,8 +748,15 @@ public class ReloadableJava11JavadocVisitor extends DocTreeScanner<Tree, List<Ja
             }
         }
 
-        // a member reference, but not matching anything on type attribution
-        return null;
+        for (JavaType.FullyQualified interface_ : classType.getInterfaces()) {
+            for (JavaType.Variable member : interface_.getMembers()) {
+                if (member.getName().equals(ref.memberName.toString())) {
+                    return member;
+                }
+            }
+        }
+
+        return fieldReferenceType(ref, classType.getSupertype());
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


greg-at-moderne (approved):
My hunch is that this implementation of finding the member might misattribute the reference, but:

- I might be wrong
- Even if right, it's likely rare
- All in all, it's better than not supporting this references at all.

jevanlingen:
I think you are no totally wrong @greg-at-moderne 😎. I tested it out:

```java
public interface HtmlMarkupI {
    Tag H1 = Tag.H1;
}

public abstract class HtmlMarkup implements HtmlMarkupI {
    String H1 = "aa";
}

class Test extends HtmlMarkup {
    /**
     * @see #H1
     */
    void onSectionTitle() {}
}
```

If you write these three classes, you would see the H1 would automatically refer to the parent class (in this case the String). But when you would use:

```java
class Test extends HtmlMarkup implements HtmlMarkupI {
    /**
     * @see #H1
     */
    void onSectionTitle() {}
}
```
the current code would refer to the to the `Tag H1` in the interface. This is wrong, it should first try to discover it by it parent class and then check for the interfaces. I'll change the order, thx for the comments!

---

N.B. You can also use:

```java
* @see HtmlMarkupI#H1
```

But then, because you directly link the class, it will no longer user the "parent-route-discovery", so that will work as well. Or better said, that did already work.

greg-at-moderne (approved):

PR #5167, opened 2025-03-13
Support this-constructor-links in JavaDoc

jevanlingen:
## What's changed?
Direct links to constructors are now supported.

## What's your motivation?
Code like [Maven Doxia's](https://github.com/apache/maven-doxia/blob/ede159d464cebf5fb17131794cac0b48d7f21f5a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java#L41) IndexingSink class gave missing type reference errors:

```java
public class IndexingSink {
    /**
     * @deprecated legacy constructor, use {@link #IndexingSink(Sink)}.
     */
    @Deprecated
    public IndexingSink(IndexEntry rootEntry) {
        this(rootEntry, new SinkAdapter());
    }

    public IndexingSink(Sink delegate) {
        this(new IndexEntry("index"), delegate);
    }
}
```

The parser could not find the types for the constructor arguments.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
@@ -1390,8 +1390,8 @@ public interface JavaType {
                 return null;
             }
             String className = ((JavaType.Class) getReturnType()).getClassName();
-            String[] parts = className.split("\\.");
-            return parts[parts.length - 1];
+            int beginIndex = className.lastIndexOf(".");
+            return beginIndex == -1 ? className : className.substring(beginIndex);
         }
 
         public FullyQualified getDeclaringType() {
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
@@ -1385,6 +1385,15 @@ public interface JavaType {
             return "<constructor>".equals(name);
         }
 
+        public @Nullable String getConstructorName() {
+            if (!isConstructor()) {
+                return null;
+            }
+            String className = ((JavaType.Class) getReturnType()).getClassName();
+            String[] parts = className.split("\\.");
+            return parts[parts.length - 1];
+        }
+
         public FullyQualified getDeclaringType() {
             return declaringType;
         }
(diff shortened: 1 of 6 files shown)


rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -1385,6 +1385,15 @@ public boolean isConstructor() {
             return "<constructor>".equals(name);
         }
 
+        public @Nullable String getConstructorName() {
+            if (!isConstructor()) {
+                return null;
+            }
+            String className = ((JavaType.Class) getReturnType()).getClassName();
+            String[] parts = className.split("\\.");
+            return parts[parts.length - 1];
```
greg-at-moderne, line 1394: That's probably possible to simplify with
`className.lastIndexOf(".")` and then substring

greg-at-moderne (approved):

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
timtebeek: Would we need any special handling here for inner classes where the className might contain a `$` ?

greg-at-moderne: IMHO the `className` already has dollars `$` stripped away as per this line:
https://github.com/openrewrite/rewrite/blob/058101dd12a76e8054525515aba2864e99e4e30c/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L314

rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java
```diff
@@ -1385,6 +1385,15 @@ public boolean isConstructor() {
             return "<constructor>".equals(name);
         }
 
+        public @Nullable String getConstructorName() {
+            if (!isConstructor()) {
+                return null;
+            }
+            String className = ((JavaType.Class) getReturnType()).getClassName();
+            int beginIndex = className.lastIndexOf(".");
+            return beginIndex == -1 ? className : className.substring(beginIndex + 1);
```
greg-at-moderne, line 1394: The `-1` thing seems excessive. I think you could do:
```
return className.substring(className.lastIndexOf(".") + 1);
```

PR #5168, opened 2025-03-13
Fix handling of maven compiler plugin configuration for java version upgrade

Laurens-W:
## What's changed?

When the maven compiler plugin is configured with a property in the parent pom which is overridden by child poms the child pom is now correctly updated instead of adding the property to the parent

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
Laurens-W, line 1: Contemplating simply calling the `AddPropertyVisitor` visitor here so as to not mess with the  logic in the `AddProperty` visitor

Laurens-W: But then again I feel like the changes to the `AddProperty` are correct

PR #5170, opened 2025-03-13
`JavaTemplate`: Add support for match arrays

fmodesto:
## What's changed?
Added support for matching methods with array parameters.

## Motivation
While working on adding support for matching generics, I noticed that this specific scenario was missing.

I'm not entirely sure if this approach is the best way forward. The following templates seem logical but will fail:

```java
final JavaTemplate template1 = JavaTemplate
        .builder("SomeClass.someMethod(#{a:any(CharSequence)})")
        .build();

final JavaTemplate template2 = JavaTemplate
        .builder("SomeClass.someMethod(#{a:any(LocalDate)})")
        .imports("java.time.LocalDate")
        .build();
```
Instead of using `TemplateParameterParser` to create a `JavaType` and later converting it back to a `String` to generate `__P__.<$type>p()`, we could inject the string as-is. This should work more reliably.

The only downside is we lose the type.  Currently, in the matching step, the implementation generates a placeholder marked as `J.Empty`. Later, `JavaTemplateSemanticallyEqual` attempts to ensure that types match. However, I don't think this step is necessary. If the method signatures are identical, shouldn't the parameters inherently be correct?

I'm also considering how a template with generics should work. Something like this:
```java
final JavaTemplate before = JavaTemplate
        .builder("Optional.ofNullable(#{object:any(T)}).orElseGet(#{supplier:any(Supplier<S>))")
        .generics("T", "S extends T")
        .imports("java.util.Optional")
        .build();
```
Would generate the valid code.
```java
import java.util.Optional;
import org.openrewrite.java.internal.template.__M__;
import org.openrewrite.java.internal.template.__P__;
class Template<T, S extends T>{{
Object o =
/*__TEMPLATE__*/Optional.ofNullable(__P__.<T>/*__p0__*/p()).orElseGet(__P__.<Supplier<S>>/*__p1__*/p())/*__TEMPLATE_STOP__*/
;
}}
```

Most of the code is ANTLR-generated. Do we need to manually invoke `gradlew generateAntlrSources`? I did so and had to adjust some headers and minor formatting issues.

Additionally, `rewrite-java/src/main/antlr/TemplateParameterLexer.tokens` appears to be out of sync, as are some of the other `.tokens` files in that directory. It seems ANTLR generates `.tokens` files inside the `src/main/java` folder instead.

### Checklist


greg-at-moderne:
> Do we need to manually invoke gradlew generateAntlrSources?

Yes, we do.

greg-at-moderne:
> Additionally, rewrite-java/src/main/antlr/TemplateParameterLexer.tokens appears to be out of sync, as are some of the other .tokens files in that directory. It seems ANTLR generates .tokens files inside the src/main/java folder instead.

Thanks for noticing this. Dealing with it in #5176 (which BTW has now caused a merge conflict in this PR).

greg-at-moderne:
> I'm also considering how a template with generics should work. Something like this:

This we have already discussed internally in the project and that's the high-level plan. With details to be ironed out on the actual syntax of:
```
        .generics("T", "S extends T")
```

greg-at-moderne (approved):
Thanks for the high-quality contribution. I appreciate the effort put not only into addressing the issue, but the broader context too.

I am gonna wait if @knutwannheden would like to review the PR too.

knutwannheden:
I will take a look on Monday.

knutwannheden (approved):
While there is already an `anyArray()` match function (e.g. `#{anyArray(int)}`) to allow matching array typed expressions, I agree that it makes sense to allow the `TemplateParameterParser` to parse array types as part of the `type` production, so that these can then also be used as type parameters. As a result we can later also consider deprecating the `anyArray()` function.

Regarding your comments in the PR description:

> Instead of using `TemplateParameterParser` to create a `JavaType` and later converting it back to a `String` to generate `__P__.<$type>p()`, we could inject the string as-is. This should work more reliably.

I think at this point I would prefer leaving it the way it is. That will also validate the input, which isn't checked by the Java compiler and could contain typos. As you'll also have noticed, some minimal parsing is required anyway due to the parameter names.

> Later, `JavaTemplateSemanticallyEqual` attempts to ensure that types match. However, I don't think this step is necessary. If the method signatures are identical, shouldn't the parameters inherently be correct?

I am not quite sure I follow. If your template e.g. looks like `#{any()}.equals(#{any()})` then what you say is true. But if the template e.g. looks like `#{any()}.equals(42)` then you would still have to match the method invocation arguments to make sure it is a match.

> I'm also considering how a template with generics should work. Something like this:
> 
> ```java
> final JavaTemplate before = JavaTemplate
>         .builder("Optional.ofNullable(#{object:any(T)}).orElseGet(#{supplier:any(Supplier<S>))")
>         .generics("T", "S extends T")
>         .imports("java.util.Optional")
>         .build();
> ```

What you propose here is indeed very close to what we were considering ourselves. I think we should give this a try and see if we hit any problems. Ultimately, we want to be able to also use this from our Refaster support.

knutwannheden:
Thanks a lot for this contribution, @fmodesto !

PR #5171, opened 2025-03-13
Add new test validation that ScanningRecipe.getScanner() does not attempt edits.

sambsnyd:
By design edits during the scanning phase are ignored. This can be a source of confusion for new recipe authors who don't yet fully understand the difference between the scanning and editing phases of the recipe lifecycle.


timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
```diff
@@ -165,8 +164,8 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
                 if (acc.usingType) {
                     return SearchResult.found(document);
```
timtebeek, line 165: Wouldn't this line need to change too?

PR #5173, opened 2025-03-13
Assign response from `withTypeParameters` in `SpacesVisitor`

timtebeek:
I have a local recipe that still has some false positives, but already was able to find these two cases where the response to an `@With` annotated method call went unchecked.

- Inspired by https://github.com/openrewrite/rewrite/pull/5160

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/SpacesVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/SpacesVisitor.java
@@ -869,7 +869,7 @@ public class SpacesVisitor<P> extends JavaIsoVisitor<P> {
                 spaceAfter(m.getPadding().getContaining(), style.getAroundOperators().getMethodReferenceDoubleColon())
         );
         if (m.getPadding().getTypeParameters() != null) {
-            m.getPadding().withTypeParameters(spaceBefore(m.getPadding().getTypeParameters(), style.getAroundOperators().getMethodReferenceDoubleColon()));
+            m = m.getPadding().withTypeParameters(spaceBefore(m.getPadding().getTypeParameters(), style.getAroundOperators().getMethodReferenceDoubleColon()));
         } else {
             m = m.getPadding().withReference(
                     spaceBefore(m.getPadding().getReference(), style.getAroundOperators().getMethodReferenceDoubleColon())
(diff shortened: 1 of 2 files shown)


knutwannheden:
How did you find these cases? I would quite like to annotate all with-methods as `@Contract(pure=true)` and we could indeed do that using `@With(onMethod=...)`. I have also contemplated using the Lombok SPI to register our own handler for`@With` (or possibly a custom annotation) to make it aware of padding using `JLeftPadded` etc. That could then automatically declare the methods as pure.

timtebeek:
This was for now just a silly local recipe with some false positives, that I had thought to add to rewrite-rewrite & PR checks eventually:
```java
import org.openrewrite.Preconditions;
import org.openrewrite.TreeVisitor;
import org.openrewrite.java.search.UsesMethod;

public class UnusedWither extends Recipe {

    private static final MethodMatcher WITH_MATCHER = new MethodMatcher("*..*#with*(..)", true);

    @Override
        return "Find unused withers";
    }

    @Override
        return "Find wither methods called, but their return value is not used.";
    }

    @Override
        JavaIsoVisitor<ExecutionContext> visitor = new JavaIsoVisitor<ExecutionContext>() {
            @Override
            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                if (WITH_MATCHER.matches(method) &&
                        getCursor().getParentTreeCursor().getValue() instanceof J.Block) {
                    return SearchResult.found(method);
                }
                return super.visitMethodInvocation(method, ctx);
            }
        };
        return Preconditions.check(new UsesMethod<>(WITH_MATCHER), visitor);
    }
}
```
You'll notice that it picks up any with named method, including things like AssertJ assertion methods, as well as with-ers that `return this` instead of a new copy. Still it was helpful to identify these cases, and the one in rewrite-python, as it's an easy mistake to make:
- https://github.com/openrewrite/rewrite-python/pull/139

knutwannheden:
I would quite like to have the methods declared as pure, as then IDEA would show a warning when calling these methods as if they had a side effect.

knutwannheden (approved):

knutwannheden:
Definitely, yes. I didn't mean to stop the PR from getting merged.

sambsnyd:
Good catch Tim!

timtebeek:
Thanks! There's one more similar case in
- https://github.com/openrewrite/rewrite-python/pull/139

Issue #5174, opened 2025-03-14
ReplaceConstantWithAnotherConstant not work when use static constant within method


ckcd:
## What version of OpenRewrite are you using?

I am using the latest rewrite (8.48.0)

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```shell
mvn org.openrewrite.maven:rewrite-maven-plugin:run ...
```

## What is the smallest, simplest way to reproduce the problem?

Run this test, it will failed:
- Outside the method definition, the `String out = QUX1` will be changed correctly
- **But within the method definition, the `String in = QUX1` will NOT changed**

```java
@Test
void shouldUpdateWithinMethod() {
    rewriteRun(
      spec -> spec.recipe(new ReplaceConstantWithAnotherConstant("foo.Bar.QUX1", "foo.Bar.QUX2")),
      java(
        """
          package foo;

          public class Bar {
              public static final String QUX1 = "QUX1";
              public static final String QUX2 = "QUX2";
          }
          """
      ),
      java(
        """
          import static foo.Bar.QUX1;

          class Test {
              String out = QUX1;
              void a() {
                  String in = QUX1;
              }
          }
          """,
        """
          import static foo.Bar.QUX2;

          class Test {
              String out = QUX2;
              void a() {
                  String in = QUX2;
              }
          }
          """
      )
    );
}
```

As for the reason, I did some troubleshooting and I thinks it's because of the redundant check `!isVariableDeclaration`:
https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java#L82
https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java#L129

I really cannot find any useful case for this `!isVariableDeclaration`, and even i remove it, all test cases can be passed. And actually if we are concern about duplicated definition (the same constant name), it won't causes any issue becuase it will be defensed by `isConstant`:
https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceConstantWithAnotherConstant.java#L124

```java
@Test
void shouldUpdateWithinMethod3() {
    rewriteRun(
      spec -> spec.recipe(new ReplaceConstantWithAnotherConstant("foo.Bar.QUX1", "foo.Bar.QUX2")),
      java(
        """
[...]

PR #5175, opened 2025-03-14
fix: ReplaceConstantWithAnotherConstant not work when use static constant within method

ckcd:
- Fix https://github.com/openrewrite/rewrite/issues/5174

## What's changed?

Remove redundant `isVariableDeclaration`.

## What's your motivation?

Fix https://github.com/openrewrite/rewrite/issues/5174

### Checklist


timtebeek (approved):
Thanks for noticing @ckcd ; indeed seems like subsequent changes after this recipe was created have removed the need for these conditionals. Good to have these cleared out, thanks!

PR #5176, opened 2025-03-14
Removing obsolete ANTLR generated code

greg-at-moderne:
## What's changed?

Removing obsolete generated ANTLR code, which doesn't seem used by anything.
A newer version of the same artifacts are stored in `rewrite-java/src/main/java/org/openrewrite/java/internal/grammar`. So I guess this must be some forgotten outcome of ANTLR before the path got changed.

## What's your motivation?
Addressing this realisation from @fmodesto:
> Additionally, rewrite-java/src/main/antlr/TemplateParameterLexer.tokens appears to be out of sync, as are some of the other .tokens files in that directory. It seems ANTLR generates .tokens files inside the src/main/java folder instead.

in #5170 (thanks!)


fmodesto:
I think the g4 were necessary. Looking at `build.gradle`.
```kotlin
tasks.register<JavaExec>("generateAntlrSources") {
    mainClass.set("org.antlr.v4.Tool")

    args = listOf(
        "-o", "src/main/java/org/openrewrite/java/internal/grammar",
        "-package", "org.openrewrite.java.internal.grammar",
        "-visitor"
    ) + fileTree("src/main/antlr").matching { include("**/*.g4") }.map { it.path }

    classpath = antlrGeneration
}
```

greg-at-moderne:
> I think the g4 were necessary. Looking at build.gradle.

Ah. Right. 🤦 . I deleted too much.

PR #5182, opened 2025-03-17
Add Javascript

jkschneider:
## What's changed?

Bringing Javascript into the core repository. A lot is changing to make openrewrite/rewrite-javascript more idiomatic to JS/TS along the way.

## What's your motivation?

Native recipe development of Javascript recipes!

## TODOs

- [ ] Support unknown marker types
- [ ] Add shim JS parser API on Java side using RPC
- [ ] check if we need JavaType support in RPC
- [ ] Visitor adapter


rewrite-javascript/src/main/javascript/execution.ts
```diff
@@ -0,0 +1,5 @@
+export type ExecutionContext = Map<string | symbol, any> & { __brand: "ExecutionContext" }
```
zieka, line 1: 👍🏼 big fan of [opaque types](https://github.com/zieka/bonsai-components/blob/main/packages/utility-types/src/opaque-types.ts#L4)

rewrite-javascript/src/main/javascript/recipe.ts
```diff
@@ -0,0 +1,221 @@
...
+        RecipeRegistry.all.set(name, recipeClass);
+    }
+}
+
+/**
+ * @param recipeName A unique name for the recipe consisting of a dot separated sequence of category
+ * names in which the recipe should appear followed by a name. For example,
+ * "org.openrewrite.typescript.find-methods-by-pattern".
+ */
+export function Registered(recipeName: string) {
```
zieka, line 175: I like this

I could image a @Recipe decorator that accepts an object with recipe meta data and standardized across all languages supporting the decorator pattern to have the authorship experience converge slightly.  Would also reduce the code in the recipe down to just method implementations (javascript is already pretty slim but for other languages may reduce some boilerplate)

zieka: examples:

typescript
```ts
@Recipe({
  id: 'org.testkitchen.BigAwesome', 
  name: 'Big awesome', 
  description: 'A recipe ensuring big awesome things'
})
class BigAwesome extends Recipe {}
```

java
```java
@Recipe(
    id = "org.testkitchen.BigAwesome",
    name = "Big awesome",
    description = "A recipe ensuring big awesome things"
)
public class BigAwesome extends Recipe {}
```

python
```python
@recipe(
    id="org.testkitchen.BigAwesome",
    name="Big awesome",
    description="A recipe ensuring big awesome things"
)
class BigAwesome(Recipe):
```

c#
```csharp
[Recipe(
    Id = "org.testkitchen.BigAwesome",
    Name = "Big awesome",
    Description = "A recipe ensuring big awesome things"
)]
public class BigAwesome : Recipe {}
```

jkschneider: I don't think I'd like the decorator pattern here because there is no way to force compile time application of it.

PR #5183, opened 2025-03-17
Fix java version update

Laurens-W:
## What's changed?

Be more defensive when parsing potential java version properties

## What's your motivation?

Recipe failures after 
- https://github.com/openrewrite/rewrite/pull/5168

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -92,13 +92,19 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
                     for (Plugin plugin : getResolutionResult().getParent().getPom().getPlugins()) {
                         if (plugin.getGroupId().equals("org.apache.maven.plugins") && plugin.getArtifactId().equals("maven-compiler-plugin") && plugin.getConfiguration() != null) {
                             for (String property : JAVA_VERSION_PROPERTIES) {
-                                if (getResolutionResult().getPom().getProperties().get(property) != null && Float.parseFloat(getResolutionResult().getPom().getProperties().get(property)) < version &&
-                                    (plugin.getConfiguration().get("source") != null && plugin.getConfiguration().get("source").textValue().contains(property)) ||
-                                    (plugin.getConfiguration().get("target") != null && plugin.getConfiguration().get("target").textValue().contains(property)) ||
-                                    (plugin.getConfiguration().get("release") != null && plugin.getConfiguration().get("release").textValue().contains(property))) {
-                                    d = (Xml.Document) new AddPropertyVisitor(property, String.valueOf(version), null)
-                                            .visitNonNull(d, ctx);
-                                    maybeUpdateModel();
+                                if (getResolutionResult().getPom().getProperties().get(property) != null) {
+                                    try {
+                                        int parsed = Integer.parseInt(getResolutionResult().getPom().getProperties().get(property));
+                                        if (parsed < version &&
+                                            (plugin.getConfiguration().get("source") != null && plugin.getConfiguration().get("source").textValue().contains(property)) ||
+                                            (plugin.getConfiguration().get("target") != null && plugin.getConfiguration().get("target").textValue().contains(property)) ||
+                                            (plugin.getConfiguration().get("release") != null && plugin.getConfiguration().get("release").textValue().contains(property))) {
+                                            d = (Xml.Document) new AddPropertyVisitor(property, String.valueOf(version), null)
+                                                    .visitNonNull(d, ctx);
+                                            maybeUpdateModel();
+                                        }
+                                    } catch (NumberFormatException ignored) {
+                                    }
                                 }
                             }
                         }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -94,7 +94,7 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
                             for (String property : JAVA_VERSION_PROPERTIES) {
                                 if (getResolutionResult().getPom().getProperties().get(property) != null) {
                                     try {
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersionTest.java
```diff
@@ -249,6 +249,131 @@ void UpdateChildProperty() {
...
+                  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+                  ~ See the License for the specific language governing permissions and
+                  ~ limitations under the License.
+                  ~
+                  -->
+                
+                <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                         xmlns="http://maven.apache.org/POM/4.0.0"
+                         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
```
timtebeek, line 279:
```suggestion
                <project>
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersionTest.java
```diff
@@ -249,6 +249,131 @@ void UpdateChildProperty() {
...
+                    </organization>
+                
+                    <licenses>
+                        <license>
+                            <name>Apache License, Version 2.0</name>
+                            <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+                            <distribution>repo</distribution>
+                        </license>
+                    </licenses>
+
```
timtebeek, line 309:
```suggestion
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersionTest.java
```diff
@@ -249,6 +249,131 @@ void UpdateChildProperty() {
...
+                </project>""")
+          ),
+          mavenProject("spring-cloud-kubernetes-integration-tests",
+            pomXml(
+              """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                         xmlns="http://maven.apache.org/POM/4.0.0"
+                         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+                    <modelVersion>4.0.0</modelVersion>
```
timtebeek, line 336:
```suggestion
                <project>
                    <modelVersion>4.0.0</modelVersion>
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersionTest.java
```diff
@@ -249,6 +249,131 @@ void UpdateChildProperty() {
...
+                </project>""")
+          ),
+          mavenProject("spring-cloud-kubernetes-k8s-client-discovery-server",
+            //language=xml
+            pomXml(
+              """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns="http://maven.apache.org/POM/4.0.0"
+                         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
timtebeek, line 361:
```suggestion
                <project>
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersionTest.java
Laurens-W: I thought about removing these but thought we might get in trouble, then again, it's only a testcase

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
```diff
@@ -92,13 +92,19 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
...
-                                    (plugin.getConfiguration().get("source") != null && plugin.getConfiguration().get("source").textValue().contains(property)) ||
-                                    (plugin.getConfiguration().get("target") != null && plugin.getConfiguration().get("target").textValue().contains(property)) ||
-                                    (plugin.getConfiguration().get("release") != null && plugin.getConfiguration().get("release").textValue().contains(property))) {
-                                    d = (Xml.Document) new AddPropertyVisitor(property, String.valueOf(version), null)
-                                            .visitNonNull(d, ctx);
-                                    maybeUpdateModel();
+                                if (getResolutionResult().getPom().getProperties().get(property) != null) {
+                                    try {
+                                        int parsed = Integer.parseInt(getResolutionResult().getPom().getProperties().get(property));
```
knutwannheden, line 97: The idea behind using `Float.parseFloat()` is probably to also allow for e.g. `1.5`. I am however also wondering why the `${java.version}` didn't get resolved properly and why it attempted to parse that as-is.

timtebeek (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
knutwannheden: Should we try recursive resolution as long as we get back another property? Or should we allow setting the property's value rather than replacing the property reference?

knutwannheden: @Laurens-W Just so we don't forget about this just because the PR is merged: Do you see anything else here that needs to done? It seems like we probably have a utility somewhere to resolve dependencies recursively.

Laurens-W: @knutwannheden thanks for the reminder! I think we can and we should recursively check properties for both this new case and the main case of the recipe

PR #5184, opened 2025-03-17
Fix UseStaticImport adding conflicting imports when using *(..) in match pattern (#4776)

### LINKED ISSUE #4776: UseStaticImport should not import duplicate method
- org.openrewrite:rewrite-core:8.41.3
- 

```java
  @Test
  void foo() {
    rewriteRun(
        spec -> spec.recipe(new UseStaticImport("org.hamcrest.Matchers *(..)")),
        // language=java
        java(
            """
            package com.helloworld;

            import static org.hamcrest.core.IsEqual.equalTo;

            public class Foo {
              public static void foo() {
                  equalTo("foo");
                  org.hamcrest.Matchers.equalTo("bar");
              }
            }
            """));
  }
```

Results in:
```
1) foo() (com.spotify.bazel.rewrite.recipes.KnownBugsTest)
org.opentest4j.AssertionFailedError: [Expected recipe to complete in 0 cycles, but took at least one more cycle. Between the last two executed cycles there were changes to "com/helloworld/Foo.java"]
expected:
  "package com.helloworld;

  import static org.hamcrest.core.IsEqual.equalTo;

  public class Foo {
    public static void foo() {
        equalTo("foo");
        org.hamcrest.Matchers.equalTo("bar");
    }
  }"
 but was:
  "package com.helloworld;

  import static org.hamcrest.Matchers.equalTo;
  import static org.hamcrest.core.IsEqual.equalTo;

  public class Foo {
    public static void foo() {
        equalTo("foo");
        equalTo("bar");
    }
  }"
	at org.openrewrite.test.LargeSourceSetCheckingExpectedCycles.afterCycle(LargeSourceSetCheckingExpectedCycles.java:97)
	at org.openrewrite.Recipe.run(Recipe.java:378)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:376)
...
```

This then causes compilation to fail with `error: reference to equalTo is ambiguous`.

Since there is already an `equalTo` method imported and used, the recipe should not also try to statically import `org.hamcrest.Matchers.equalTo`. This is similar to
- https://github.com/openrewrite/rewrite/issues/4304.

SebastiaanvdMade:
## What's changed?

UseStaticImport was changed to also work with *(..) in the method pattern. It compares all the methods in the given class to the used imports to see if they conflict, to make sure that no conflicting imports are left in the code.
Added 2 unit tests (positive and negative case).

## What's your motivation?

- Fixes #4776 

The method names for the given class are found by instantiating the class with Class.forName(), I'm not sure if that is the best way to do this.
Also, I added an extra test-runtime dependency for hamcrest in order to use the exact test-case described in the issue. That might not be necessary if someone can think of another set of conflicting methods that are already included.

@timtebeek was already involved in the ticket.

\-

## Any additional context

\-

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java
```diff
@@ -61,16 +67,34 @@ public Validated<Object> validate() {
...
+        int indexSpace = methodPattern.indexOf(' ');
+        String typeNamePattern = methodPattern.substring(0, indexSpace);
+
+        if (methodPattern.contains(" *(")) {
+            try {
+                Class<?> cls = Class.forName(typeNamePattern);
+                Set<String> methodNames = Arrays.stream(cls.getMethods())
+                        .map(Method::getName)
+                        .collect(Collectors.toSet());
+                methodNameMatchers.addAll(methodNames);
```
timtebeek, line 82: Hmm; this class-for-name is not likely to work in all environments where we run recipes. There's the Maven and Gradle plugins, Moderne CLI & Platform, third party CLIs like Quarkus & Micronaut, all of which might have a different classpath when they run a recipe as opposed to what's available when you compile a project.

Is there any way we can fix this issue without this type of use?

timtebeek:
Thanks for the help here @SebastiaanvdMade ; went for a slightly different approach in
- https://github.com/openrewrite/rewrite/pull/5806

That didn't suffer from the need to know methods available as pointed out on
- https://github.com/openrewrite/rewrite/pull/5184#discussion_r2055897248

By only looking at named static imports we achieve a fix for the original issue reported.

PR #5185, opened 2025-03-17
When no version is present, skip resolving

Laurens-W:
## What's changed?

When trying to resolve a pom without a version simply skip it as that will not succeed.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -886,6 +886,9 @@ public class ResolvedPom {
                 // The dependency may be modified by the current pom's dependency management
                 d = getValues(d, depth);
                 try {
+                    if (depth == 0 && d.getVersion() == null) {
+                        throw new MavenDownloadingException("No version provided for direct dependency", null, dd.getDependency().getGav());
+                    }
                     if (d.getVersion() == null || (d.getType() != null && (!"jar".equals(d.getType()) && !"pom".equals(d.getType()) && !"zip".equals(d.getType())))) {
                         continue;
                     }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -886,11 +886,7 @@ public class ResolvedPom {
                 // The dependency may be modified by the current pom's dependency management
                 d = getValues(d, depth);
                 try {
-                    if (d.getVersion() == null) {
-                        throw new MavenDownloadingException("No version provided", null, dd.getDependency().getGav());
-                    }
-
-                    if (d.getType() != null && (!"jar".equals(d.getType()) && !"pom".equals(d.getType()) && !"zip".equals(d.getType()))) {
+                    if (d.getVersion() == null || (d.getType() != null && (!"jar".equals(d.getType()) && !"pom".equals(d.getType()) && !"zip".equals(d.getType())))) {
                         continue;
                     }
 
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -886,11 +886,7 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
                 // The dependency may be modified by the current pom's dependency management
                 d = getValues(d, depth);
                 try {
-                    if (d.getVersion() == null) {
-                        throw new MavenDownloadingException("No version provided", null, dd.getDependency().getGav());
-                    }
-
-                    if (d.getType() != null && (!"jar".equals(d.getType()) && !"pom".equals(d.getType()) && !"zip".equals(d.getType()))) {
+                    if (d.getVersion() == null || (d.getType() != null && (!"jar".equals(d.getType()) && !"pom".equals(d.getType()) && !"zip".equals(d.getType())))) {
```
Laurens-W, line 889: Do we have another way of leaving a warning for the user here?

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -1390,133 +1352,102 @@ void activeByDefaultWithoutPomLocalActiveProfile() {
...
               )
             );
-
-            AssertionFailedError err = assertThrows(AssertionFailedError.class, fn);
-            assertThat(err.getMessage()).contains("Problem parsing a/pom.xml");  // brittle:(, but class above is broad
         }
 
         @Test
-        @Disabled("Not yet implemented")
+        @Disabled("This context might be available if we go through the MavenMojoProjectParser in the rewrite-maven-plugin")
```
Laurens-W, line 1450: Same goes for this test

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -1390,133 +1352,102 @@ void activeByDefaultWithoutPomLocalActiveProfile() {
...
+              )
+            );
         }
 
         @Issue("https://github.com/openrewrite/rewrite/issues/4269")
         @DisplayName("activeByDefault=true profiles from a POM should not be active" +
                      " if deactivated from the command-line")
         @Test
-        @Disabled
+        @Disabled("This context might be available if we go through the MavenMojoProjectParser in the rewrite-maven-plugin")
```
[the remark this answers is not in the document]
timtebeek: Feel free to remove this test from here then; just to keep the focus on what works. Optional to test this in the plugin repository, but let's first get this issue fixed here.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
timtebeek: Not at this level I believe; we have that `MavenDownloadingException` and the try/catch wrapping `MavenDownloadingExceptions` to signify breaking issues; for non-breaking issues we sometimes add an informative marker that's not printed to the source, but that's not possible at this level. I think it's fine to tolerate the absence here and `continue`.

timtebeek (approved):
Approved already, but likely best to remove those disabled tests, and then possibly inline those `parent` and `child` text blocks again, especially if those are then only used once.

sambsnyd (commented):
Our behavior should match Maven's. 
If a version is missing in a pom you ask maven to build it's an error. If a version is missing in a transitive dependency's pom it's a warning. Not sure what the behavior is when a version is missing from a direct dependency's pom.

Please clarify what testing you've done and that we match Maven's own treatment of error vs warning in all cases.

Laurens-W:
> Approved already, but likely best to remove those disabled tests, and then possibly inline those parent and child text blocks again, especially if those are then only used once.

The parent and child are used in two tests currently, I threw away a test that made no sense looking at how maven handles active profiles and adjusted the MavenParser slightly to correctly handle profiles set in the settings.xml (those also override any activeByDefault profiles)

timtebeek (approved):
Thanks for making those changes to the tests too; looks good to me!

sambsnyd:
Thanks guys!

PR #5194, opened 2025-03-19
Improve UpdateMavenProjectPropertyJavaVersion for nested properties

Laurens-W:
## What's changed?

Nested properties are now properly resolved and the correct property is modified

@knutwannheden 

## Any additional context

https://github.com/openrewrite/rewrite/pull/5183#discussion_r1998863149

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -88,6 +88,19 @@ public class MavenVisitor<P> extends XmlVisitor<P> {
         return PROPERTY_MATCHER.matches(getCursor());
     }
 
+    public boolean isPropertyReference(String value) {
+        return value.startsWith("${") && value.endsWith("}");
+    }
+
+    public String resolvePropertyReference(String propertyReference) {
+        String propKey = propertyReference.replace("${", "").replace("}", "");
+        String propValue = getResolutionResult().getPom().getProperties().get(propKey);
+        if (propValue != null && isPropertyReference(propValue)) {
+            return resolvePropertyReference(propValue);
+        }
+        return propKey;
+    }
+
     public boolean isDependencyTag() {
         return isTag("dependency") && DEPENDENCY_MATCHER.matches(getCursor());
     }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -92,14 +92,20 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
                     for (Plugin plugin : getResolutionResult().getParent().getPom().getPlugins()) {
                         if (plugin.getGroupId().equals("org.apache.maven.plugins") && plugin.getArtifactId().equals("maven-compiler-plugin") && plugin.getConfiguration() != null) {
                             for (String property : JAVA_VERSION_PROPERTIES) {
-                                if (getResolutionResult().getPom().getProperties().get(property) != null) {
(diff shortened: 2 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
```diff
@@ -92,14 +92,20 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
                     for (Plugin plugin : getResolutionResult().getParent().getPom().getPlugins()) {
                         if (plugin.getGroupId().equals("org.apache.maven.plugins") && plugin.getArtifactId().equals("maven-compiler-plugin") && plugin.getConfiguration() != null) {
                             for (String property : JAVA_VERSION_PROPERTIES) {
-                                if (getResolutionResult().getPom().getProperties().get(property) != null) {
+                                String propValue = getResolutionResult().getPom().getProperties().get(property);
```
knutwannheden, line 95: I think this `getValue()` is what you were looking for:
```suggestion
                                String propValue = getResolutionResult().getPom().getValue(getResolutionResult().getPom().getProperties().get(property));
```

knutwannheden: Although that won't help with updating that "deepest" property. But I am also unsure if we really want that, as that property could also have other purposes. So I am not really sure what we should do here. How do other recipes handle this problem?

What are your thoughts here, @timtebeek ? Maybe it is after all best to leave such property references alone and then require the user to have a recipe to update those separately.

timtebeek: Not sure I have much more insights to share here; if indeed a property only refers to another one I think it could help if we upgrade the nested property, or risk not compiling because the source level is not set correctly in say a compiler plugin.

What would happen here without any changes? Do we override `java.version` discarding the old reference? Or do we not make any change at all?

knutwannheden: Without this PR no changes are made. To me it is unclear which approach is best. Following a consistent pattern across recipes would be best, but maybe we don't have any other cases like this one.

Laurens-W: Without the change it will check these conventional properties https://github.com/openrewrite/rewrite/blob/20dc764c508a7be8d1c5e197f39fe7508a510c2e/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java#L37-L45 but if any of them refer to another property (that is not coincidentally part of that list) nothing will change.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -100,6 +100,19 @@ public boolean isPropertyTag() {
         return PROPERTY_MATCHER.matches(getCursor());
     }
 
+    public boolean isPropertyReference(String value) {
+        return value.startsWith("${") && value.endsWith("}");
+    }
```
sambsnyd, line 105: While it is typical to have only one property reference in a string it is not required. This is valid:
```xml
<foo>${bar}.${baz}</foo>
```

Laurens-W:
Since there's no open issue regarding this improvement and there are some edge cases that would have to be closely looked at I'm going to close this PR here, we can always revisit later.

timtebeek:
- Possibly superseded by https://github.com/openrewrite/rewrite/pull/5634

Laurens-W:
I'll verify tomorrow by running the testcase in this PR on main

Laurens-W:
Test still fails on main

Issue #5072, opened 2025-02-20
GroovyParsingException when parsing Gradle build file


BrendanHart:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.45.0
- Gradle plugin v7.1.4
- rewrite-groovy v8.45.0
- rewrite-gradle v8.45.0
- Gradle 6.9

## How are you running OpenRewrite?

I am using the Gradle plugin, with an init script.

## What is the smallest, simplest way to reproduce the problem?

I have simplified my original `build.gradle` to something which produces a similar exception. I have been unable to reproduce this in a unit test since they use Groovy 3, which doesn't seem to produce the exception. I have tried to create a unit test with Groovy 2 dependencies, but it seemed the GroovyParserVisitor then sees a ClassNotFoundException for LamdaExpression, so I'm unsure how to configure a test to compile the Groovy code with Groovy 2 yet have Groovy 3 classes available. The only way I have been able to reproduce is by actually running `gradle --init-script init.gradle rewriteDryRun`.

init.gradle
```groovy
initscript {
    repositories {
        maven { url "https://plugins.gradle.org/m2" }
    }
    dependencies { classpath("org.openrewrite:plugin:latest.release") }
}
rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite:rewrite-java")
    }
    rewrite {
        activeRecipe("org.openrewrite.groovy.format.AutoFormat")
        setExportDatatables(true)
    }
    afterEvaluate {
        if (repositories.isEmpty()) {
            repositories {
                mavenCentral()
            }
        }
    }
}
```

build.gradle
```groovy
def x = (1 / 1 ) * 100
System.out.println("test")
```

## What did you expect to see?

No fail to parse errors.

## What did you see instead?
Failed to parse error.

## What is the full stack trace of any errors you encountered?

```
Error during rewrite dry run
org.openrewrite.groovy.GroovyParsingException: Failed to parse build.gradle, cursor position likely inaccurate.
	at org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:203)
	at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$3(GroovyParser.java:147)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Collections$2.tryAdvance(Collections.java:4717)
[...]

PR #5203, opened 2025-03-20
Ensure a divison is not interpreted as a `SLASHY_STRING` delimiter

BrendanHart:
## What's changed?

Added validation which, if a slash is found when determining parenthesis level, will traverse the node to find whether it is actually a division operator.

## What's your motivation?

- #5072 

Whether this is a valid solution, and if so, is narrowing down the traversal based on source offsets of each expression (like in `FindTokenVisitor.visitBinaryExpression`) necessary.

Also added some Groovy 2 tests to replicate for now, although maybe this would not be preferred as the "JVM Test Suite plugin is an [incubating](https://docs.gradle.org/current/userguide/feature_lifecycle.html#sec:incubating_state) API".

I considered whether we should skip the string parsing altogether and only find delimiters via traversing the node tree somehow, but validating the `SLASHY_STRING` seemed to require less change.

### Checklist

### DIFF
--- a/rewrite-groovy/build.gradle.kts
+++ b/rewrite-groovy/build.gradle.kts
@@ -1,6 +1,38 @@
 plugins {
     id("org.openrewrite.build.language-library")
+    id("jvm-test-suite")
 }
+
+testing {
+    suites {
+        val test by getting(JvmTestSuite::class) {
+            useJUnitJupiter()
+        }
+
+        register<JvmTestSuite>("groovy2Test") {
+            configurations.getByName("groovy2TestRuntimeClasspath") {
+                resolutionStrategy {
+                    force("org.codehaus.groovy:groovy:2.5.22")
+                }
+            }
+
+            dependencies {
+                implementation(project())
+                implementation(project(":rewrite-test"))
+                implementation("org.assertj:assertj-core:latest.release")
+            }
+
+            targets {
+                all {
+                    testTask.configure {
+                        shouldRunAfter(test)
+                    }
+                }
+            }
+        }
+    }
+}
+
 
 dependencies {
     api(project(":rewrite-java"))
@@ -24,12 +56,6 @@ dependencies {
     testRuntimeOnly(project(":rewrite-java-17"))
 }
 
-//val testJava8 = tasks.register<Test>("testJava8") {
-//    javaLauncher.set(javaToolchains.launcherFor {
-//        languageVersion.set(JavaLanguageVersion.of(8))
-//    })
-//}
-//tasks.named("check").configure {
-//    // Enable once the java8 tests are passing
-//    // dependsOn(testJava8)
-//}
+tasks.named("check") {
+    dependsOn(testing.suites.named("groovy2Test"))
+}
--- a/rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
+++ b/rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
@@ -27,9 +27,9 @@ class GroovyParserTest implements RewriteTest {
         rewriteRun(
                 groovy(
                         """
-                                def x = (1 / 1) * 2
-                                System.out.println("test")
-                                """
+                        def x = (1 / 1) * 2
+                        System.out.println("test")
+                        """
                 )
         );
     }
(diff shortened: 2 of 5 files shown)


rewrite-groovy/build.gradle.kts
```diff
@@ -24,12 +56,6 @@ dependencies {
...
 
-//val testJava8 = tasks.register<Test>("testJava8") {
-//    javaLauncher.set(javaToolchains.launcherFor {
-//        languageVersion.set(JavaLanguageVersion.of(8))
-//    })
-//}
-//tasks.named("check").configure {
-//    // Enable once the java8 tests are passing
-//    // dependsOn(testJava8)
-//}
```
greg-at-moderne, line 35: Thanks for cleaning up the obsolete code.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2548,7 +2549,7 @@ private int determineParenthesisLevel(int childLineNumber, int parentLineNumber,
             return SINGLE_LINE_COMMENT;
         } else if (source.startsWith("/*", c)) {
             return MULTILINE_COMMENT;
-        } else if (source.startsWith("/", c)) {
+        } else if (source.startsWith("/", c) && validateIsDelimiter(node, c)) {
```
shanman190, line 2552: So one thing that comes to mind is a `ConstantExpression` would be the only thing that could represent a "slashy" string. While a mathematical division should be a `BinaryExpression`. So wouldn't we be able to directly identify based on the type of the node whether it's one versus the other?

BrendanHart: My understanding is `BinaryExpression` could contain a `ConstantExpression` as the left or right expressions, meaning a slash can be found in `source` which belongs to a child node rather than the `BinaryExpression` we may be visiting now, hence the need to traverse the tree to find which node the `/` actually came from. Please let me know if I'm missing something.

Other thoughts were whether we should traverse just once and keep some kind of set of all `BinaryExpression` operators "skippable indexes" in the source for delimiters, but since I've only came across this issue for this particular case, I'm not sure if it's worth it.

shanman190: That's fair. I suppose we could encounter this -- albeit strange -- string concatenation:

```
/Hello, / + /world!/
```

Which would be a binary expression, but have both left and right being slashy strings.

rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.tree.ParseError;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class GroovyParserTest {
+
+    @Test
+    void souldNotTreatDivisionAsDelimiter() {
```
shanman190, line 26: Just curious, but was there a reason for this not to be a `rewriteRun(..)`? It should verify that everything parses correctly into the `G` type as well as print idempotency.

BrendanHart: Nope, changed to use `rewriteRun`! Also added a test case to test a mix of `SLASHY_STRING` and division.

BrendanHart:
I've also removed the `visitBlockExpression` override which was an exploration of whether we could prune the tree based on the position of the `/`, but if this is a rare case then maybe it would not be worth trying to do so initially. However, please let me know if we should do this!

rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
```diff
@@ -0,0 +1,48 @@
...
+class GroovyParserTest implements RewriteTest {
+
+    @Test
+    void shouldNotTreatDivisionAsDelimiter() {
+        rewriteRun(
+                groovy(
+                        """
+                                def x = (1 / 1) * 2
+                                System.out.println("test")
+                                """
```
greg-at-moderne, line 32:
```suggestion
                        """
                        def x = (1 / 1) * 2
                        System.out.println("test")
                        """
```

rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    void shouldHandleUsingSlashyStringsWithDivision() {
+        rewriteRun(
+                groovy(
+                        """
+                                def x = (Integer.parseInt(/3/) / Integer.parseInt(/2/)) * Integer.parseInt(/5/)
+                                System.out.println("test")
+                                """
```
greg-at-moderne, line 44:
```suggestion
                        """
                        def x = (Integer.parseInt(/3/) / Integer.parseInt(/2/)) * Integer.parseInt(/5/)
                        System.out.println("test")
                        """
```

PR #544, opened 2021-05-14
Test that parser can handle two classes with the same name and different packages

kmccarp:
In the `assert(Unc|C)hanged` test harness's `dependsOn`, if you have two files with the same simple name, the Java parser errors out with the below exception. This PR adds a disabled unit test that demonstrates the behavior.

```
java.lang.IllegalStateException: Call reset() on JavaParser before parsing anotherset of source files that have some of the same fully qualified names

	at org.openrewrite.java.Java11Parser.lambda$parseInputs$0(Java11Parser.java:179)
	at io.micrometer.core.instrument.composite.CompositeTimer.record(CompositeTimer.java:57)
	at org.openrewrite.java.Java11Parser.parseInputs(Java11Parser.java:174)
	at org.openrewrite.java.JavaParser.parse(JavaParser.java:98)
	at org.openrewrite.RecipeTest$DefaultImpls.assertChanged(RecipeTest.kt:66)
	at org.openrewrite.java.JavaRecipeTest$DefaultImpls.assertChanged(JavaRecipeTest.kt:122)
	at org.openrewrite.java.cleanup.UnnecessaryQualificationTest$DefaultImpls.assertChanged(UnnecessaryQualificationTest.kt:8)
	at org.openrewrite.java.Java11UnnecessaryQualificationTest.assertChanged(Java11VisitorDebugTests.kt)
	at org.openrewrite.RecipeTest$DefaultImpls.assertChanged$default(RecipeTest.kt:34)
	at org.openrewrite.java.cleanup.UnnecessaryQualificationTest$DefaultImpls.keepsTheMostUsedClass(UnnecessaryQualificationTest.kt:112)
	at org.openrewrite.java.Java11UnnecessaryQualificationTest.keepsTheMostUsedClass(Java11VisitorDebugTests.kt)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
[...]

PR #5204, opened 2025-03-20
Add recipe MoveLambdaArgumentParentheses

zakaria-shahen:
## What's changed?
- Add kotlin recipe in rewrite-kotlin module.
- The new recipe to Move lambda argument outside of method invocation parenthesis when they are the only argument

#### Before 
```kotlin
val foo = 1.let({ it + 1 })
```
#### After
```kotlin
val foo = 1.let { it + 1 }
```

## What's your motivation?
- The original idea coming from [#544](https://github.com/openrewrite/rewrite-kotlin/issues/544). 

@shauvik and anyone.

### Checklist

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/cleanup/MoveLambdaArgumentParentheses.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/cleanup/MoveLambdaArgumentParentheses.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.kotlin.cleanup;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.kotlin.KotlinTemplate;
+import org.openrewrite.kotlin.KotlinVisitor;
+import org.openrewrite.kotlin.tree.K;
+
+import java.time.Duration;
+import java.util.List;
+
+public class MoveLambdaArgumentParentheses extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Move lambda argument outside of method invocation parentheses";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Move lambda argument outside of method invocation parentheses when they are the only argument. " +
+               "For example, converts `1.let({ it + 1 })` to `1.let { it + 1 }`.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(1);
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new KotlinVisitor<ExecutionContext>() {
+            @Override
+            public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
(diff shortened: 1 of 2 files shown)


rewrite-kotlin/src/test/java/org/openrewrite/kotlin/cleanup/MoveLambdaArgumentParenthesesTest.java
```diff
@@ -0,0 +1,135 @@
...
+import static org.openrewrite.kotlin.Assertions.kotlin;
+
+class MoveLambdaArgumentParenthesesTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new MoveLambdaArgumentParentheses())
+                .typeValidationOptions(TypeValidation.none())
+                .expectedCyclesThatMakeChanges(1)
+                .cycles(1);
```
timtebeek, line 33: Let's try to see if we can make the recipe work without these escape hatches; each of these disable a check the proper functioning of the recipe, which could hurt the ability to chain recipes back to back.

For the cycles it helps to make no change at all if the desired state is already reached, either before starting, or after a second invocation.

For the type validation you might want to look at using template placeholders and arguments, as opposed to a string-ified template.

```suggestion
        spec.recipe(new MoveLambdaArgumentParentheses());
```

rewrite-kotlin/src/test/java/org/openrewrite/kotlin/cleanup/MoveLambdaArgumentParenthesesTest.java
```diff
@@ -0,0 +1,135 @@
...
+            kotlin(
+            """
+              fun method() {
+                  run(::print)
+              }
+              """
+            )
+        );
+    }
+}
```
timtebeek, line 135: It's probably good to add at least one test that only contains a before block, and already uses the desired `{ ... }` form, to show that we do not make any changes in those cases.

timtebeek (changes requested):
Thanks a lot for kicking this off! Neat to see how much you've already worked out for a first contribution here. Should make a nice addition to our Kotlin support to make the usage more idiomatic. I've posted two suggestions that hint at improvements we can still make before we merge. Let us know if you'd want to pick those up yourself or if at any point you'd like to hand over.

rewrite-kotlin/src/test/java/org/openrewrite/kotlin/cleanup/MoveLambdaArgumentParenthesesTest.java
zakaria-shahen: Thank you for review pull request. 
I edited my test cases to work without those escape hatches; however, one test case still needs skip-type validation. Until now, I haven't found a way to pass the validation (without escape hatches). What should I do about that?

timtebeek: Thanks for the changes! I'll see if I can help clear out that last hurdle before we merge.

zakaria-shahen: Please let me know if you think there is anything I can do to find a different way. Thanks.

timtebeek (approved):
Thanks again for the work here @zakaria-shahen ! To avoid the type issues we saw I've changed the implementation to instead add the marker that omits the parentheses when printing. With that we're ready for a merge, and you should see this in next week's release! 🙏🏻

rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
```diff
@@ -119,7 +119,7 @@ protected String parseStringValue(String value, Function<String, @Nullable Strin
                 }
 
                 // Proceed with unprocessed value.
-                startIndex = result.indexOf(placeholderPrefix, endIndex + placeholderSuffix.length());
+                startIndex = result.indexOf(placeholderPrefix, endIndex + placeholderSuffix.length() - 1);
```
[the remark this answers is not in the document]
zakaria-shahen: I accidentally included it to this PR, but yes, this changes to enable the next template pattern `${}${}` (without a space before the second $).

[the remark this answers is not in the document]
zakaria-shahen: I don't need it right now, but I can do a PR about it if you need it. 
Thank you for helping me.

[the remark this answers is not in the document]
timtebeek: Thanks! Maybe best to only open that if anyone needs it, but thanks for the offer to help! :)

PR #5208, opened 2025-03-21
Added deeper scanning into RemoveUnusedImports

opencirclesolutions-wove:
## What's changed?
Added deeper scanning into RemoveUnusedImports

## What's your motivation?
I’ve forked the class and made some improvements for them and wanted to provide feedback of those improvements.
These improvements were made because the recipe was throwing away a lot of imports that were being used for our use case.

I’m fairly new to OpenRewrite so I’m not 100% sure that my changes are in line with the best practices. But I tried. Feel free to improve on my code.

Use my own cloned class?

## Any additional context
I’ve disabled one unit test because it gave a parser error. The same unit test in my project using an older rewrite version works fine. It may help down track where the error may come from.

### Checklist


rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
```diff
@@ -2068,4 +2068,95 @@ A method(A.B ab, A.C ac) {}
...
+            }
+            """
+          )
+        );
+    }
+
+    @Test
+    void testImportAnnotation() {
+        rewriteRun(
+          java("""
```
timtebeek, line 2103:
```suggestion
        rewriteRun(
          java(
            """
```

rewrite-java/src/main/java/org/openrewrite/java/RemoveUnusedImports.java
```diff
@@ -315,6 +366,287 @@ public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionCon
...
+            cu.getTypesInUse().getTypesInUse().stream().forEach(type -> {
+                if (type instanceof JavaType.Class) {
+                    JavaType.Class aClass = (JavaType.Class) type;
+                    if (aClass.getKind() == JavaType.FullyQualified.Kind.Annotation) {
+                        FindAnnotations.find(cu, aClass.getFullyQualifiedName()).forEach(annotations::add);
+                    }
+                }
+            });
+            return annotations;
+        }
```
timtebeek, line 384: I'd have expected something like the following, instead of all the scan methods below; would that be possible here?
```java
Set<J.Annotation> annotations = new JavaIsoVisitor<Set<J.Annotation> >() {
    public J.Annotation visitAnnotations(J.Annotations ann, Set<J.Annotation> set) {
        set.add(ann);
        return ann;
    }
}.reduce(cu, new HashSet<>());
```

timtebeek:
Fixed a couple issues with the recipe yesterday
- https://github.com/openrewrite/rewrite/pull/5714

I've now reran the tests you added here without any of the recipe changes, and that means now only this test fails:
```java

    @Test
    void lombokValInLambda() {
        rewriteRun(
          java(
            """
            import lombok.val;
            import java.util.List;
            import java.util.ArrayList;
            
            public class Foo {
                List<String> names = new ArrayList<>();
                public String method() {
                    return names.stream()
                    .map(n -> {
                        val name = "";
                        return n + name;
                    })
                    .collect(Collectors.joining(""));
                }
            }
            """
          )
        );
    }
```
I don't think we'd need all of the changes you've added here just to support that case, and given the conflicts we see here I'm going to close this PR for now. You're welcome to propose a separate fix for that one failing case still. Thanks for helping out!

PR #5209, opened 2025-03-24
rewrite-xml not to depend on rewrite-maven

greg-at-moderne:
## What's changed?

Removing the dependency of `rewrite-xml` on `rewrite-maven`.

## What's your motivation?

To speed up the CI builds in this project.
`rewrite-xml` happened to be one of the modules I saw being built in what seemed an unrelated PR:
image
It turned out `rewrite-xml` depends on `rewrite-maven` for no strong reason, so I thought we could remove the dependency. And for the context `rewrite-maven` depends on `rewrite-core`, `rewrite-java` and more. Thus it's likely to be built in most of the PRs.

### DIFF
--- a/rewrite-xml/build.gradle.kts
+++ b/rewrite-xml/build.gradle.kts
@@ -31,7 +31,6 @@ dependencies {
     implementation("org.apache.commons:commons-text:1.11.+")
 
     testImplementation(project(":rewrite-test"))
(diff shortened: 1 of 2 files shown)


greg-at-moderne:
Confirmed it worked. The subsequent build of a similar code change didn't have `rewrite-xml`:
image

PR #5210, opened 2025-03-24
rewrite-json not to depend on rewrite-yaml

greg-at-moderne:
### Checklist


greg-at-moderne:
Confirmed it worked. The subsequent build of a similar code change didn't have `rewrite-json`:
image

PR #3460, opened 2023-08-04
Add Maven dependency before only comments in <dependencies>

timtebeek:
## What's changed?
Prevent NPE by checking whether the other tag we compare against is null.

## What's your motivation?
Fixes #3458

## Anything in particular you'd like reviewers to focus on?
Pointed out in review comments.

PR #5219, opened 2025-03-26
AddDependency recipe: Support only non-dependencies in Gradle's dependency block

jevanlingen:
## What's changed?
The AddDependency recipe does support gradle files with only non-dependencies in its dependency block.

## What's your motivation?
If you have a gradle file that looks for example like this:

```groovy
dependencies {
  if (project.hasProperty('x')) {
    implementation "commons-lang:commons-lang:1.0"
  }
}
```

The recipe would throw an error:

```bash
Caused by: java.lang.NullPointerException: Cannot read field "value" because "anotherFloat" is null
	at java.base/java.lang.Float.compareTo(Float.java:923)
	at org.openrewrite.gradle.internal.InsertDependencyComparator.compare(InsertDependencyComparator.java:85)
	at org.openrewrite.gradle.AddDependencyVisitor$InsertDependencyInOrder.visitMethodInvocation(AddDependencyVisitor.java:282)
```

## Any additional context
For the maven counterpart, this actually has already been addressed in https://github.com/openrewrite/rewrite/pull/3460. As I found this out after I created this PR, I had already written my code. To be honest, I find my code a little bit cleaner (as it fixes the problem instead of working around it with a null-check), I also updated the maven counterpart :innocent:.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/InsertDependencyComparator.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/InsertDependencyComparator.java
@@ -16,56 +16,66 @@
 package org.openrewrite.gradle.internal;
 
 import lombok.Getter;
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.Statement;
 
 import java.util.*;
-import java.util.stream.Collectors;
 
 public class InsertDependencyComparator implements Comparator<Statement> {
     private final Map<Statement, Float> positions = new LinkedHashMap<>();
 
     @Getter
+    @Nullable
     private Statement afterDependency;
 
     @Getter
+    @Nullable
     private Statement beforeDependency;
 
-    public InsertDependencyComparator(List<Statement> existingStatements, J.MethodInvocation dependencyToAdd) {
-        for (int i = 0, len = existingStatements.size(); i < len; i++) {
-            positions.put(existingStatements.get(i), (float) i);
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/InsertDependencyComparator.java
```diff
@@ -16,56 +16,66 @@
...
+        for (Statement s : statements) {
+            if (s instanceof J.MethodInvocation || (s instanceof J.Return && ((J.Return) s).getExpression() instanceof J.MethodInvocation)) {
+                dependencies.add(s);
+            }
+        }
 
-        ideallySortedDependencies.add(dependencyToAdd);
-        ideallySortedDependencies.sort(dependenciesComparator);
+        if (dependencies.isEmpty()) {
+            positions.put(dependencyToAdd, statements.size() + 0.5f);
```
Laurens-W, line 52: This 0.5f feels like magic?

jevanlingen: Agreed. The reason this works:

```
private final Map<Statement, Float> positions = new LinkedHashMap<>();
```

Which is filled with:
- First statement -> 1
- Second statement -> 2
- Third statement -> 3
- etc.

So it could also have been a List<Statement>. But with a map you can do a little trick! If you add a number between an existing number, like:
- Fouth statement -> 2.5

It is located in order between the second and third. Now, check the compare function:

```java
@Override
public int compare(Statement o1, Statement o2) {
  return positions.get(o1).compareTo(positions.get(o2));
}
```  
By using the `0.5f` trick, the comparison works kinda out-of-the-box.

I was even considering to change it. but I did not. Because I thought "It just works" and "It's a little bit out of scope to change this". Let me know what you think!

Laurens-W (approved):

PR #5220, opened 2025-03-26
Assume timeSavings=0 when no changes are made

greg-at-moderne:
## What's changed?

Making Estimated Time Savings count as 0 for source files which didn't have any change to their text.

## What's your motivation?

To fix the cases of overestimation of the time spent in some cases. E.g. this claims 23 hours saved on a purely mechanical change of 8 files:

I couldn't find a better way to exclude the remote source files than this ugly `instanceof Remote`.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Result.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Result.java
@@ -23,6 +23,7 @@ import org.openrewrite.jgit.lib.FileMode;
 import org.openrewrite.marker.DeserializationError;
 import org.openrewrite.marker.RecipesThatMadeChanges;
 import org.openrewrite.marker.SearchResult;
+import org.openrewrite.remote.Remote;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -63,7 +64,7 @@ public class Result {
         this.recipes = recipes;
 
         Duration timeSavings = null;
-        if ((before == after) || (before != null && after != null && before.printAllTrimmed().equals(after.printAllTrimmed()))) {
+        if (isLocalAndHasNoChanges(before, after)) {
             timeSavings = Duration.ZERO;
         } else {
             for (List<Recipe> recipesStack : recipes) {
@@ -258,5 +259,14 @@ public class Result {
     @Override
     public String toString() {
         return diff();
+    }
+
+    public static boolean isLocalAndHasNoChanges(@Nullable SourceFile before, @Nullable SourceFile after) {
+        return (before == after) ||
+                (before != null && after != null &&
+                        // Remote source files are fetched on `printAll`, let's avoid that cost.
+                        !(before instanceof Remote) &&
+                        !(after instanceof Remote) &&
+                        before.printAll().equals(after.printAll()));
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/Result.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Result.java
@@ -265,8 +265,7 @@ public class Result {
         return (before == after) ||
                 (before != null && after != null &&
                         // Remote source files are fetched on `printAll`, let's avoid that cost.
-                        !(before instanceof Remote) &&
-                        !(after instanceof Remote) &&
+                        !(before instanceof Remote) && !(after instanceof Remote) &&
                         before.printAll().equals(after.printAll()));
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/Result.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Result.java
@@ -63,12 +63,16 @@ public class Result {
         this.recipes = recipes;
 
         Duration timeSavings = null;
-        for (List<Recipe> recipesStack : recipes) {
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/Result.java
```diff
@@ -63,12 +64,16 @@ public Result(@Nullable SourceFile before, @Nullable SourceFile after, Collectio
...
 
         Duration timeSavings = null;
-        for (List<Recipe> recipesStack : recipes) {
-            if (recipesStack != null && !recipesStack.isEmpty()) {
-                Duration perOccurrence = recipesStack.get(recipesStack.size() - 1).getEstimatedEffortPerOccurrence();
-                if (perOccurrence != null) {
-                    timeSavings = perOccurrence;
-                    break;
+        if (isLocalAndHasNoChanges(before, after)) {
+            timeSavings = Duration.ZERO;
```
jevanlingen, line 68: Because `timeSavings` is a nullable field, you can just `return` here. Then you also don't need the `else` construct.

rewrite-core/src/main/java/org/openrewrite/Result.java
```diff
@@ -255,4 +260,13 @@ public String diff(@Nullable Path relativeTo, PrintOutputCapture.@Nullable Marke
     public String toString() {
         return diff();
     }
+
+    public static boolean isLocalAndHasNoChanges(@Nullable SourceFile before, @Nullable SourceFile after) {
+        return (before == after) ||
+                (before != null && after != null &&
+                        // Remote source files are fetched on `printAll`, let's avoid that cost.
+                        !(before instanceof Remote) &&
+                        !(after instanceof Remote) &&
```
jevanlingen, line 269: Would put this in one line:
```suggestion
                        !(before instanceof Remote) && !(after instanceof Remote) &&
```

If you read it, it's easier to follow:
```
BEFORE and AFTER are the same
<OR>
BEFORE and AFTER are not null <AND>
  BEFORE and AFTER are no instance of Remote <AND>
  BEFORE.printAll() is the same as AFTER.printAll()
```

rewrite-core/src/main/java/org/openrewrite/Result.java
```diff
@@ -255,4 +260,12 @@ public String diff(@Nullable Path relativeTo, PrintOutputCapture.@Nullable Marke
     public String toString() {
         return diff();
     }
+
+    public static boolean isLocalAndHasNoChanges(@Nullable SourceFile before, @Nullable SourceFile after) {
```
jevanlingen, line 264: I am wondering if this method should be moved to somewhere else. It feels weird the RecipeRunCycle calls a static method from a Result class. Maybe doing it the other way around makes more sense (so moving this method to the RecipeRunCycle class).

sambsnyd (approved):

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -225,7 +225,9 @@ private void recordSourceFileResult(@Nullable SourceFile before, @Nullable Sourc
         String beforePath = (before == null) ? "" : before.getSourcePath().toString();
         String afterPath = (after == null) ? "" : after.getSourcePath().toString();
         Recipe recipe = recipeStack.peek();
-        Long effortSeconds = (recipe.getEstimatedEffortPerOccurrence() == null) ? 0L : recipe.getEstimatedEffortPerOccurrence().getSeconds();
+        Long effortSeconds = (recipe.getEstimatedEffortPerOccurrence() == null || Result.isLocalAndHasNoChanges(before, after))
```
Laurens-W, line 228: Can we somehow get an instance of `Result` here and just ask that for it's time savings? That way `isLocalAndHasNoChanges` can become private

rewrite-core/src/main/java/org/openrewrite/Result.java
greg-at-moderne: 6f5403dc0

rewrite-core/src/main/java/org/openrewrite/Result.java
greg-at-moderne: This method is used in `Result` constructor.

`Result` is a more basic primitive than `RecipeRunCycle`, so I think it's better to have this logic in `Result`. Otherwise `Result` would be calling a static method from a more complex class of `RecipeRunCycle` and that would be weirder.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
greg-at-moderne: I think that's the whole point of calling this logic twice. These two (`Result` and `RecipeRunCycle`) are two separate pieces of logic to aggregate the results. Not sure why we have the similar thing done twice. When testing both were needed. And `RecipeRunCycle` doesn't seem to use any of `Result` logic (despite the name being similar), so it would add an additional dependency if we instantiated an artificial `Result` instance just to get the time savings data.

PR #5221, opened 2025-03-26
Handle the `instanceof final` pattern matching syntax in Java

### LINKED ISSUE #5200: Java idempotency issue with `instanceof final` pattern matching

Current main = 45ca3c98d074f7434f8d4aec781228619b21313d

```java
    @Test
    void instanceofFinal() {
        rewriteRun(
          java(
            """
            public class Main {
              public static void main(String[] argv) {
                String s = "a";
                if (s instanceof final String s2) {
                    System.out.println(s2);
                }
              }
            }
            """
          )
        );
    }
```

No change.

## What did you see instead?

```diff
-       if (s instanceof final String s2) {
+       if (s instanceof StringString s2) {
```

## Context

Inspired by real-life: https://github.com/PaperMC/Velocity/blob/d9f1016bd57d50f5052bb90222894adb40fd256c/api/src/main/java/com/velocitypowered/api/proxy/server/PingOptions.java#L84

sudouser777:
1. Updated the Java 17 and 21 Parser to handle `final` modifier in `instance of` patternmatching.
2. Added new field `modifier` to `J.InstanceOf` class.
3. Updated the `JavaPrinter`  to handle new `modifier` field.

- closes #5200 

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
@@ -3028,6 +3028,18 @@ public interface J extends Tree {
             return p;
         }
 
+        @Deprecated
+        public InstanceOf(UUID id, Space prefix, Markers markers, JRightPadded<Expression> expression, J clazz, @Nullable J pattern, @Nullable JavaType type) {
+            this.id = id;
+            this.prefix = prefix;
+            this.markers = markers;
+            this.expression = expression;
+            this.clazz = clazz;
+            this.pattern = pattern;
+            this.type = type;
+            this.modifier = null;
+        }
+
         @RequiredArgsConstructor
         public static class Padding {
             private final InstanceOf t;
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
@@ -2986,6 +2986,11 @@ public interface J extends Tree {
         @Getter
         JavaType type;
 
+        @With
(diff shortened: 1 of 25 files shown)


sudouser777:
closes #5200

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -2986,6 +2986,11 @@ public InstanceOf withExpression(Expression expression) {
         @Getter
         JavaType type;
 
+        @With
+        @Nullable
+        @Getter
+        Modifier modifier;
+
```
greg-at-moderne, line 2993: I think for backward compatibility reasons we would need another constructor which has the old signature. So that recipes code which uses the constructor continues to compile - without and with this change.

One example of such backward incompatible change is what you have noticed yourself in `rewrite-groovy`. But the problem might also affect other pieces of code, outside of this repository.

You can mark the "new old" constructor as deprecated already.

sudouser777: I made the change. Thank you for the suggestion.

greg-at-moderne: b9d44fa78f2deab60a96319011616e20c59399b2

greg-at-moderne (approved):
LGTM.
Thanks for this valuable contribution.

timtebeek (approved):
Great to see you immediately jump to a detailed contribution here @sudouser777 ! Thanks a lot!

PR #5222, opened 2025-03-26
add preconditions getter to Declarative Recipe

sdelamo:
## What's changed?
I added a getter for preconditions in the declarative recipe class. 

## What's your motivation?
I would like to be able to give a recipe examine the preconditions associated with the recipe. 

Yes, I don't want to access that via reflection. 

### Checklist


timtebeek:
Merging this as unlikely to cause issues; if @sambsnyd has any concerns after an async review we can go over those then. Thanks again!

Issue #5169, opened 2025-03-13
Java idempotency issue with single-line comments in the middle of a variable declaration


greg-at-moderne:
## What version of OpenRewrite are you using?
Current main = 22df20c0c138935a86960d986b9b594127baa2a2

## What is the smallest, simplest way to reproduce the problem?
```java
    @Test
    void trailingEmptySingleLineInTheMiddle() {
        rewriteRun(
          java(
            """
              public class Test {
                private//
                int i = 436;
              }
              """
          )
        );
    }
```

Inspired by real-life: https://github.com/spring-projects/spring-data-jpa/blob/1d4e301195fedb573a9647b6856dd8bb34d4d5b2/spring-data-jpa/src/test/java/org/springframework/data/jpa/domain/sample/EmbeddedIdExampleEmployee.java#L34

## What did you expect to see?

No change.

## What did you see instead?

```diff
  public class Test {
-   private//
    int i = 436;
  }
```

greg-at-moderne:
I haven't looked into this one, but I think there might be one challenge with fully supporting cases like this.
The comments might appear between various modifiers (e.g. between `private` and `static`), etc. And our LST model doesn't seem to support this.

traceyyoshima:
I have a personal Java parser that parses the entire language spec idempotent, but it’s modeled closely to the complied tree to enable analysis that OR can’t cover.

I’ve seen this issue before. IIRC, it’s caused by the modifier parsing conditions that detect the end of a modifier.

I don’t have time now, but I can dig up the changes and you can modify them as necessary.

traceyyoshima:
IIRC, the issue is [here](https://github.com/openrewrite/rewrite/blob/main/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java#L2238), and the modifier won't be matched.

I don't recall what I modified from the OR parser, but it should be close.

```java
    private List<Java.Modifier> mapModifiersAndAnnotations(Map<Integer, JCTree.JCAnnotation> annotationPosTable) {
        List<RightPadded<Java.Annotation>> currentAnnotations = null;
        List<Java.Modifier> sortedModifiers = null;

        boolean inComment = false;
        boolean inMultilineComment = false;
        int afterLastModifierPosition = cursor;
        int lastAnnotationPosition = cursor;

        int keywordStartIdx = -1;
        for (int i = cursor; i < source.length(); i++) {
[...]

PR #5226, opened 2025-03-27
Fix for Java parser when no space between modifier and annotation

### LINKED ISSUE #5199: Idempotency issue with Java parser when no space between modifier and annotation: `private@Annotation`

Current main = 45ca3c98d074f7434f8d4aec781228619b21313d

```java
    @Test
    void modifierNoSpaceThenAnnotation() {
        rewriteRun(
          java(
            """
            public class Main {
              public@jdk.jfr.Name("A") static void main(String[] argv) {
               System.out.println("A");
              }
            }
            """
          )
        );
    }
```

no change

## What did you see instead?
```diff
- public@jdk.jfr.Name("A") static void main(String[] argv) {
+ @jdklic@jdk.jfr.Name("A") void main(String[] argv) {
```

## Context

Inspired by real-life: https://github.com/hivemq/hivemq-community-edition/blob/f13006a4206ebaba44379911b22c6ff571e7b746/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L210

sudouser777:
1. Updated the code handle the no space between modifier and annotation.
2. Also handled the scenario if the multiple annotations followed by without space after modifer.
3. Updated unit tests for all the possible scenarios.
4. closes #5199


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -2342,7 +2344,8 @@ private List<J.Annotation> collectAnnotations(Map<Integer, JCAnnotation> annotat
                     continue;
                 }
                 annotations.add(convert(jcAnnotation));
-                i = cursor;
+                // Adjusting the index by subtracting 1 to account for the case where annotations are not separated by a space
+                i = cursor - 1;
```
greg-at-moderne, line 2348: I haven't executed the code, just reading it.

But the unconditional `-1` here looks off to me. Why would we subtract 1 regardless of the value of `noSpace`?

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -2209,6 +2209,7 @@ private ReloadableJava17ModifierResults sortedModifiersAndAnnotations(ModifiersT
         boolean inMultilineComment = false;
         int afterLastModifierPosition = cursor;
         int lastAnnotationPosition = cursor;
+        boolean noSpace = false;
```
greg-at-moderne, line 2212: I think it would be clearer if the `noSpace` was actually replaced with an int. `spaceCountBeforeModifier` or something with `0` as the default value. The code below would be more readable with just adding the `+spaceCountBeforeModifier` to some expressions, etc.

Also - one more thing to have a look at is other cases of spaces. The code seems to focus on 0 spaces or 1 space. What if there are two spaces? Can you add a test case to cover this?

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
sudouser777: we can add conditional  check. I thought it would be simpler this way. it would just make one more extra iteration. if there is a space.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
greg-at-moderne: Correct, it's about handling the case of no space between annotation and the modifier (or between annotations).
But the code also needs to handle other cases (the mainstream ones). And 99% of even more of the cases, it will be 1 space between annotation and modifier. The 0-space or 2-space examples are rare, but if we are adding support for 0-space, then I thought 2-space could be handled too. Thus my comments above.

Sorry for confusing you. I might have communicated my remarks better.

sudouser777: I got the test case part though you are right I should have added more testcases to check the regular secnarios. But changing variable type is kind confused. it's just to check if there is no space between modifer and annotation.  We don't need to keep track of that. Correct If I'm wrong.

I will add the testcases to cover other secnarios.

rewrite-java/src/test/java/org/openrewrite/java/ModifierTest.java
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class ModifierTest implements RewriteTest {
```
jevanlingen, line 23: Parser tests have to go in **rewrite-java-tck** (which is a module that's executed by the java-8, java-11, java-17 and java-21 modules). In this case these tests should be added to the [AnnotationTest](https://github.com/openrewrite/rewrite/blob/a1d887038c4c67d9f07801200f0a20afec0bb9a3/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/AnnotationTest.java#L39) class instead of creating a new test class.

sudouser777: After moving the tests to above module test were failing for java 8 and 11. I'm not able to debug the code as most of the code dependent on java 17 that causing issues in creating parser for 8 and 11. Any suggestions?

sudouser777: I have moved the tests to `AnnotationTest`.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
greg-at-moderne: Variable type - fine, if you find it better.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
greg-at-moderne: Fine. I think the `-1` is safe.
I was worried it would change the behavior (not performance).

greg-at-moderne (approved):

greg-at-moderne:
I am gonna wait a while if any of my colleagues want to review, otherwise I am gonna merge this one.

Once again, thanks for the contribution!

traceyyoshima:
> I am gonna wait a while if any of my colleagues want to review, otherwise I am gonna merge this one.
> 
> Once again, thanks for the contribution!

I think you may want to review the comments on https://github.com/openrewrite/rewrite/issues/5169 since it fixes both the annotations and the comments.

jevanlingen:
@traceyyoshima I am going to merge this one, as this one is ready to be used. Yours have to analysed and adjusted to make it work (looking good though, if I scan your code quickly 👌🏻). Could be done at a later stage!

PR #5230, opened 2025-03-28
WIP: Support expressions using type parameters in context-free templates

knutwannheden:


knutwannheden:
I also cleaned up our `JavaTemplate` tests to no longer call `contextSensitive()` unless necessary.

jkschneider (approved):

PR #5234, opened 2025-03-29
More efficient recipe loading

knutwannheden:
Adds a new `ResourceLoader#loadRecipe(String)` method to allow loading a single recipe. This allows the resource loader to only load a specific recipe. The `YamlResourceLoader` caches the parsed YAML so that any referenced recipes can be demand-loaded more quickly. Additional improvements:

* The `loadRecipe()` method allows specifying which "details" should get loaded (maintainers, contributors, and examples)
* Java recipes referenced from YAML are attempted to be loaded using reflection before reverting to the more powerful Jackson `ObjectMapper`

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/Environment.java
@@ -16,6 +16,7 @@
 package org.openrewrite.config;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.Contributor;
 import org.openrewrite.Recipe;
 import org.openrewrite.RecipeException;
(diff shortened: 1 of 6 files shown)


rewrite-core/src/main/java/org/openrewrite/config/Environment.java
```diff
@@ -141,8 +145,30 @@ public Collection<RecipeDescriptor> listRecipeDescriptors() {
...
     public Recipe activateRecipes(Iterable<String> activeRecipes) {
-        Map<String, Recipe> recipesByName = listRecipes().stream().collect(toMap(Recipe::getName, identity()));
+        List<String> asList = new ArrayList<>();
+        for (String activeRecipe : activeRecipes) {
+            asList.add(activeRecipe);
+        }
+        return activateRecipes(asList);
+    }
+
+    public Recipe activateRecipes(Collection<String> activeRecipes) {
```
jkschneider, line 157: Why does Iterable need to be narrowed to Collection?

jkschneider: I think I tend the other way -- make parameters `Iterable` unless there is a good reason for it to be a list or collection. Though clearly haven't done this consistently.

jkschneider (approved):

PR #5240, opened 2025-04-01
Optionally change String literals from `ChangePackage` and `ChangeType`

Laurens-W:
## What's changed?

Added a Reference implementation for finding type/package references in String literals

## What's your motivation?

- https://github.com/openrewrite/rewrite-migrate-java/pull/693#discussion_r2014430310

### Checklist


Laurens-W:
Tests will probably fail because I haven't added it properly to ChangeType and ChangePackage

Laurens-W:
I fear this change might be too broad as now it will by default always take String literals into account

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1477,7 +1477,7 @@ public ClassDeclaration withTypeParameters(@Nullable JContainer<TypeParameter> t
     @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
     @RequiredArgsConstructor
     @AllArgsConstructor(access = AccessLevel.PRIVATE)
-    final class CompilationUnit implements J, JavaSourceFile, SourceFile {
+    final class CompilationUnit implements J, JavaSourceFile, SourceFile, SourceFileWithReferences {
```
timtebeek, line 1480: Hmm; I've seen some conditionals checking first if something is a `JavaSourceFile`, and `else` if it is a `SourceFileWithReferences`. Those would have to be adjusted if we're to continue down this path.

Alternatively just have an option on ChangeType and ChangePackage to also include String literals (default false) could be the easier option.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -97,6 +105,16 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                     if (!Boolean.TRUE.equals(ignoreDefinition) && containsClassDefinition(cu, oldFullyQualifiedTypeName)) {
                         return SearchResult.found(cu);
                     }
+                    if (Boolean.TRUE.equals(visitStringLiterals)) {
+                        return Traits.literal().asVisitor(lit -> {
+                            stopAfterPreVisit();
+                            String string = lit.getString();
+                            if (string != null && string.contains(oldFullyQualifiedTypeName)) {
```
Laurens-W, line 112: Same here

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -97,6 +105,16 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                     if (!Boolean.TRUE.equals(ignoreDefinition) && containsClassDefinition(cu, oldFullyQualifiedTypeName)) {
                         return SearchResult.found(cu);
                     }
+                    if (Boolean.TRUE.equals(visitStringLiterals)) {
+                        return Traits.literal().asVisitor(lit -> {
```
timtebeek, line 109: I've pushed a new test that fails: when `visitStringLiterals = true`, but there is no String literal match, then we now also fail to replace type matches. I'd have expected this recipe to continue making type changes and not _only_ String literals. The same applies for ChangePackage, but I've not added a test there.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -91,6 +99,15 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                 stopAfterPreVisit();
                 if (tree instanceof JavaSourceFile) {
                     JavaSourceFile cu = (JavaSourceFile) tree;
+                    if (Boolean.TRUE.equals(visitStringLiterals)) {
+                        return Traits.literal().asVisitor(lit -> {
+                            String string = lit.getString();
+                            if (string != null && string.contains(oldPackageName)) {
```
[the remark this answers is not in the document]
timtebeek: Not sure what you mean by this comment; could you clarify?

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -91,6 +100,19 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                        Traits.literal().asVisitor(lit -> {
+                            String string = lit.getString();
+                            if (string != null && string.contains(oldPackageName)) {
+                                found.set(true);
+                            }
+                            return lit.getTree();
+                        }).visit(cu, ctx, getCursor().getParentTreeCursor());
+                        if (found.get()) {
+                            return SearchResult.found(cu);
+                        }
```
timtebeek, line 114: Personally I tend to prefer the reduce function where possible; in this case you're only interested in the boolean, and reduce makes that explicit
```suggestion
                        if (Traits.literal().asVisitor((Literal lit, AtomicBoolean bool) -> {
                                    String string = lit.getString();
                                    if (string != null && string.contains(oldPackageName)) {
                                        bool.set(true);
                                    }
                                    return lit.getTree();
                                }).reduce(cu, new AtomicBoolean(false), getCursor().getParentTreeCursor())
                                .get()) {
                            return SearchResult.found(cu);
                        }
```

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
timtebeek: The `stopAfterPreVisit();` is only here and not in `ChangePackage`; any specific reason for that?

And same as above `reduce` can likely be used here as well.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -132,11 +154,10 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
         });
     }
 
-    private static class JavaChangeTypeVisitor extends JavaVisitor<ExecutionContext> {
+    private class JavaChangeTypeVisitor extends JavaVisitor<ExecutionContext> {
```
timtebeek, line 157: Why the change to non static as opposed to expanding the constructor here? Now there's a bit of an awkward mix of passing through some outer class field values through the constructor, while directly accessing one field.

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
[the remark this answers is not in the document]
Laurens-W: The regex matcher used in `visitLiteral` is a little more specific on when something matches vs using contains. Using that here might save a little bit of overhead

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
Laurens-W: I noticed other classes in this package (including `changePackage`) don't define this kind of inner visitors as static, but I agree it's not very neat. I'll revert the removal of static and add the field and adjust the constructor

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -199,6 +224,19 @@ public J visitImport(J.Import import_, ExecutionContext ctx) {
             return updateType(javaType);
         }
 
+        @Override
+        public J visitLiteral(J.Literal literal, ExecutionContext ctx) {
+            J.Literal lit = literal;
+            if (Boolean.TRUE.equals(visitLiterals) && literal.getType() == JavaType.Primitive.String) {
+                Pattern pat = Pattern.compile("(?:\\A|\\s)" + oldFullyQualifiedTypeName + "(?:|\\s)");
```
timtebeek, line 231: Are we not needlessly complicating things here? I'm usually a fan of using [word boundaries with `\b`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) , as that matches a whole category:
```suggestion
                Pattern pat = Pattern.compile("\\b" + oldFullyQualifiedTypeName + "\\b");
```

rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java
```diff
@@ -235,6 +254,19 @@ public J visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ct
             return updateType(javaType);
         }
 
+        @Override
+        public J visitLiteral(J.Literal literal, ExecutionContext ctx) {
+            J.Literal lit = literal;
+            if (Boolean.TRUE.equals(visitStringLiterals) && literal.getType() == JavaType.Primitive.String) {
+                Pattern pat = Pattern.compile("(?:\\A|\\s)" + oldPackageName + "[.\\p{javaJavaIdentifierStart}]*(?:|\\s)");
```
timtebeek, line 261:
```suggestion
                Pattern pat = Pattern.compile("\\b" + oldPackageName + "\\b");
```

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -199,6 +224,19 @@ public J visitImport(J.Import import_, ExecutionContext ctx) {
             return updateType(javaType);
         }
 
+        @Override
+        public J visitLiteral(J.Literal literal, ExecutionContext ctx) {
+            J.Literal lit = literal;
+            if (Boolean.TRUE.equals(visitLiterals) && literal.getType() == JavaType.Primitive.String) {
```
jevanlingen, line 230:
```suggestion
            if (TRUE.equals(visitLiterals) && literal.getType() == String) {
```

⚠️ Code style change, but as I start using more and more static imports. I see other people follow.

PR #5242, opened 2025-04-01
Support for M2_HOME in Maven

greg-at-moderne:
**What**:
Adding support for `M2_HOME` environment variable, which can override the `~/.m2` default of where the Maven Local repository is stored.

**Why**:
Moderne customer request.

**Testing performed**:
Publishing a custom recipe JAR (`rewrite-install-test`) and then using Moderne CLI:

Part 1 - `mod config recipes jar install`
```
$ export M2_HOME=/tmp/customhome
$ mkdir -p ${M2_HOME}/repository/com/yourorg/rewrite-install-test
$ cp -r /Users/greg/.m2/repository/com/yourorg/rewrite-install-test/0.1.0-SNAPSHOT /Users/greg/.m2/repository/com/yourorg/rewrite-install-test/* ${M2_HOME}/repository/com/yourorg/rewrite-install-test/
$ rm -rf /Users/greg/.m2/repository/com/yourorg # so that it's not in Maven Local anymore
$ mod config recipes jar install com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT
[...]
com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT
Found 11 recipes
    + Installed
Finished installing recipes (1m 24s)

MOD SUCCEEDED in 1m 24s
$ cat /Users/greg/.moderne/cli/recipes/20250401185816-8u7DY/recipes.log | grep rewrite-install-test
Installing recipe: com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT
[...]
Resolving recipe jar com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT
  Downloaded pom for com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT from repository file:///tmp/customhome/repository/
  Downloading jar com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT from file:///tmp/customhome/repository/
  Downloaded jar com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT to /Users/greg/.moderne/cli/maven-cache/com/yourorg/rewrite-install-test/0.1.0-SNAPSHOT/rewrite-install-test-0.1.0-SNAPSHOT.jar
```

Part 2 - `mod build`

`mod build .` works as normally

Part 3 - recipe runs

Using https://github.com/openrewrite/rewrite-maven-plugin as an example of a Maven-based project.

All of these worked:
- `mod run . --recipe FindParseFailures`
- `mod run . --recipe org.openrewrite.maven.BestPractices`
- `mod run . --recipe org.openrewrite.maven.utilities.PrintMavenAsDot`

Part 4 - `mod config build maven settings`

With `export M2_HOME=/tmp/customhome` continued...

- `mod config build maven settings show` gives "There is no currently configured Maven settings"
- `mod config build maven settings edit /tmp/custom-settings.xml`  works with the following content:
```
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
     <localRepository>/tmp/newhome</localRepository>
</settings>
```
- subsequent `mod config recipes jar install com.yourorg:rewrite-install-test:0.1.0-SNAPSHOT` works as previously, i.e. the `M2_HOME` env variable is taken into account

**Anything in particular you'd like reviewers to focus on?**

Yes. I am not fully sure the environment should be called `M2_HOME`. This seems to be a bit confusing when compared to standard Maven landscape. In Maven the `M2_HOME` is a former (removed a few years ago, relevant to Maven 2.x) environment variable for specifying where Maven is installed - similar to `JAVA_HOME` or current `MAVEN_HOME`. It was never meant to mean the location of Maven Local repository location.


rewrite-maven/src/main/java/org/openrewrite/maven/MavenHomeDirectory.java
```diff
@@ -0,0 +1,33 @@
...
+        String m2Home = System.getenv("M2_HOME");
+        if (m2Home != null) {
+            return Paths.get(m2Home);
+        }
+        return Paths.get(System.getProperty("user.home"), ".m2");
+    }
+
+    public static Path getRepository() {
+        return get().resolve("repository");
+    }
```
jevanlingen, line 32: I think now you have introduce the `MavenHomeDirectory` I would make the `get()` method private and add the:
- userSecuritySettingsPath
- userSettingsPath

methods here. I think that's nice then letting the other files still figure it out.

okundzich:
https://docs.moderne.io/user-documentation/moderne-cli/cli-reference#mod-config-build-maven-settings 

You probably also saw this command and how it will interact with this env var.

greg-at-moderne:
> You probably also saw this command and how it will interact with this env var.

Thanks @okundzich, updated the PR description with more tests in Part 4.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenHomeDirectory.java
greg-at-moderne: 8a18cab61

rewrite-maven/src/main/java/org/openrewrite/maven/MavenHomeDirectory.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class MavenHomeDirectory {
+    public static Path get() {
```
jevanlingen, line 22: Approved, but maybe you want to keep this method private? I don't know for sure, so I'll let you decide for yourself.

jevanlingen (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/MavenHomeDirectory.java
greg-at-moderne: No strong feelings.
So following your suggestion in 16dcce183fc0f3dd1f728f10052cda8a31eab035

timtebeek:
I noticed there's a method here that expects a `conf/settings.xml`. Did you explore that as well?
https://github.com/openrewrite/rewrite/blob/8a18cab61e5ea912c734f5a2005a6d4ba4329bcb/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java#L150-L160

greg-at-moderne:
> I noticed there's two methods here that expect a conf/settings.xml. Did you explore those as well?

Thanks for noticing and bringing this up. After looking into this I think that the `findMavenHomeSettings()` and `findMavenHomeSettings()` methods you have found serve a slightly different purpose and should be go as they are (with this PR's changes). Maven seems to support two different sources of settings - one is from where it's installed (typically `$MAVEN_HOME/conf/settings.xml`) and user-specific settings (typically in `~/.m2/settings.xml`).

This PR changes the logic for the latter only - i.e. the user-specific ones. Both are eventually merged as observed in the code of these methods.

The only worrying part I can see is the same as I've raised before - the `M2_HOME` environment variable seems to be used for two different meanings if this PR is merged. One is the override for local user's `~/.m2` directory, the other one being alternative name for `MAVEN_HOME` - i.e. where Maven is installed.

But I guess we'll have to live with this confusion, as the request is explicitly for `M2_HOME` and nothing else.

timtebeek (approved):

okundzich:
> Yes. I am not fully sure the environment variable should be called M2_HOME. This seems to be a bit confusing when compared to standard Maven landscape. In Maven the M2_HOME is a former (removed a few years ago, relevant to Maven 2.x) environment variable for specifying where Maven is installed - similar to JAVA_HOME or current MAVEN_HOME. It was never meant to mean the location of Maven Local repository location. 

The request is just to follow the standard practices and locate maven home based on env variables set to identify it. So if the standard is to call this something else  (not M2_HOME) we should follow it and rename it. @jkschneider @greg-at-moderne

PR #5246, opened 2025-04-02
Introducing artifactChecksumCache in MavenWrapper

greg-at-moderne:
## What's changed?

Adding a cache for artifacts checksums to `MavenWrapper` code.

## What's your motivation?

The initial trigger for this was to speed up the CI builds of this project. In a non-scientific test locally, `UpdateMavenWrapperTest` (which happens to be one of the slowest tests in the whole suite) goes from ~44s to ~11s with this change.

When taking the whole context, I think it will have positive impact on running `UpdateMavenWrapper` recipe on multiple projects at once (or against a project with multiple `pom.xml` files) too.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenWrapper.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenWrapper.java
@@ -81,7 +81,7 @@ public class MavenWrapper {
     String distributionUri;
     Checksum distributionChecksum;
 
-    private final static Cache<String, Checksum> artifactChecksumCache = Caffeine.newBuilder().maximumSize(20).build();
+    private final static Cache<URI, Checksum> artifactChecksumCache = Caffeine.newBuilder().maximumSize(20).build();
 
     public static MavenWrapper create(
             @Nullable String wrapperVersion,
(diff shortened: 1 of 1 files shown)


shanman190:
IIRC, the MavenWrapper -- just like the GradleWrapper -- is cached on first access in the UpdateMavenWrapper recipe for the resolved version already.

greg-at-moderne:
> is cached on first access in the UpdateMavenWrapper recipe for the resolved version already.

IMHO it's cached only for the duration of one recipe run. This goes beyond.

shanman190:
Right, right. And Maven doesn't publish sha25y files, so as a result we have to create our own checksums from the remote or locally cached file again and again.

Would potentially updating `Checksum` to cache the result to the local cache make more sense?

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenWrapper.java
```diff
@@ -131,27 +135,33 @@ public static MavenWrapper create(
...
         }
     }
 
+    private static Checksum retrieveChecksumUsingCache(RemoteFile remote, ExecutionContext ctx) {
+        Checksum ret = artifactChecksumCache.getIfPresent(remote.getUri().toString());
+        if (ret != null) {
+            return ret;
+        }
+        ret = Checksum.sha256(remote, ctx).getChecksum();
+        artifactChecksumCache.put(remote.getUri().toString(), ret);
```
knutwannheden, line 161: I think URIs are safe to be used as cache keys (whereas URLs are problematic). The advantage would be that it doesn't require any normalization. E.g.

```java
        URI uri1 = URI.create("HTTP://localhost:8080");
        URI uri2 = URI.create("http://localhost:8080");
        System.out.println(uri1.equals(uri2)); // prints true
        System.out.println(uri1.toString().equals(uri2.toString())); // prints false
```

knutwannheden (approved):

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenWrapper.java
greg-at-moderne: 31a909ae

PR #5247, opened 2025-04-03
Fixes the YAML parser issue when unicode characters used.

### LINKED ISSUE #5177: YAML parser swallows a single character when Unicode is used

Current main = 27f7bd7172193860aed6dee3f807b3b687f1708a

The following unit test case:
```java
    @Test
    void unicodeCharacters() {
        rewriteRun(
          yaml(
              """
                - name: Autobuild
                # 📚COMMENT
                - name: Perform CodeQL Analysis
              """)
        );
    }
```
in `rewrite-yaml`

fails with:
```diff
  - name: Autobuild
  # 📚COMMENT
- - name: Perform CodeQL Analysis
+ - ame: Perform CodeQL Analysis
```

## Context

- Extracted from real-life: https://github.com/jenkinsci/dimensionsscm-plugin/blob/510dc415363a64fb6334ebba4f9a1cd318010efe/.github/workflows/codeql-analysis.yml
- Likely somehow related to: #4924 as it had similar symptoms

sudouser777:
1. Added code to adjust the `valueStart` when it contains unicode chracters.
2. closes #5177

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -245,8 +245,8 @@ public class YamlParser implements org.openrewrite.Parser {
                             tag = createTag(tagPrefix, Markers.EMPTY, tagName, tagSuffix);
                         }
 
-                        // Decrementing the value by 1 if fmt contains any Unicode character
-                        valueStart -= (int) fmt.codePoints().filter(Character::isSupplementaryCodePoint).count();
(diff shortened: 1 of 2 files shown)


greg-at-moderne (approved):
Thanks for fixing this.

I've applied the automatic code suggestions. I hope they didn't break any test cases. I think it didn't like the `"""` being indented differently than the first sequence item of YAML `- something`.

sudouser777:
> Thanks for fixing this.
> 
> I've applied the automatic code suggestions. I hope they didn't break any test cases. I think it didn't like the `"""` being indented differently than the first sequence item of YAML `- something`.

For somereason formatting not working properly in my IDE.

PR #5252, opened 2025-04-03
Add data table that records all download attempts of any Maven POM or metadata

kmccarp:
## What's changed?
I've added a data table for every attempt to download every Maven metadata and POM.

This data table records `start` and `end` for each download, to avoid a coordinated omission problem where a recipe run started, and then got forcibly killed or otherwise died before completion.

The following are example contents of this data table:

image

image

## What's your motivation?

I want to give users of Maven and Gradle recipes more detail around what downloads were attempted and how long they took, as well as if any requests never finished.

`MavenPomDownloader`.

@sambsnyd 

Yes, originally I had intended to just add `ResolutionEventListener` methods for this and leave this up to the client, but I believe it is a better experience for this to _always_ include these data table rows rather than have variance between clients.

## Any additional context

N/A

### Checklist


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -45,6 +47,8 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
 
     @EqualsAndHashCode.Exclude
     MavenMetadataFailures metadataFailures = new MavenMetadataFailures(this);
+    @EqualsAndHashCode.Exclude
```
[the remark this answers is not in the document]
kmccarp: Valid feedback, invalid comment. Thanks, bot!

```diff
diff --git a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
index f4f96c1..8771a6f 100644
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -47,6 +47,7 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
 
     @EqualsAndHashCode.Exclude
     MavenMetadataFailures metadataFailures = new MavenMetadataFailures(this);
+
     @EqualsAndHashCode.Exclude
     MavenDownloadEvents mavenDownloadEvents = new MavenDownloadEvents(this);
```

kmccarp, line 50:
```suggestion

    @EqualsAndHashCode.Exclude
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
```diff
@@ -151,6 +160,7 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
             private Optional<String> findAcceptableVersion(String groupId, String artifactId, ExecutionContext ctx)
                     throws MavenDownloadingException {
                 if (availableVersions == null) {
+                    ResolutionEventListener resolutionListener = new MavenExecutionContextView(ctx).getResolutionListener();
```
Laurens-W, line 163: this seems to be unused

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -148,22 +149,42 @@ public MavenPomDownloader(Map<Path, Pom> projectPoms, HttpSender httpSender, Exe
...
+                    Duration duration = Duration.ofNanos(System.nanoTime() - start);
+                    if (downloadEventsDataTable != null) {
+                        downloadEventsDataTable.insertFinishedRow(ctx, request.getUrl().toString(), Integer.toString(response.getCode()), duration);
                     }
                     return response.getBodyAsBytes();
                 }
             });
         } catch (FailsafeException failsafeException) {
             if (failsafeException.getCause() instanceof HttpSenderResponseException) {
+                Integer responseCode = ((HttpSenderResponseException) failsafeException.getCause()).getResponseCode();
```
Laurens-W, line 177: This seems unused

rewrite-maven/src/main/java/org/openrewrite/maven/ChangePluginGroupIdAndArtifactId.java
```diff
@@ -31,8 +32,6 @@
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class ChangePluginGroupIdAndArtifactId extends Recipe {
-    @EqualsAndHashCode.Exclude
```
Laurens-W, line 34: Somewhat unrelated to the contents of this PR: I think we should add a note to the description of this recipe that no validation is done, it is assumed the values passed are valid. Even the description of newVersion property is invalid as it doesn't resolvenode or semver style selectors

Laurens-W: - https://github.com/openrewrite/rewrite/pull/5257

Laurens-W (approved):
Aside from those small comments this LGTM!

rewrite-maven/src/main/java/org/openrewrite/maven/AddParentPom.java
kmccarp: Ah good catch, thank you! That was a holdover from a previous commit.

Issue #5236, opened 2025-03-31
Assigning parenthesised expressions fails with Groovy 2


BrendanHart:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.49.0
- Gradle plugin v7.3.0
- rewrite-groovy v8.49.0
- Gradle 6.9.1

## How are you running OpenRewrite?

Originally I found this bug whilst using the Gradle plugin with Gradle 6, but have since reproduced with a test.

## What is the smallest, simplest way to reproduce the problem?

Added the below to `GroovyParserTest` in the Groovy 2 test suite:
```java
    @Test
    void shouldBeAbleToAssignToParenthesisedExpressions() {
        rewriteRun(
            groovy(
                """
                x = (123)
                """
            )
        );
    }
```

## What did you expect to see?

The test pass.

## What did you see instead?

The test failed.

## What is the full stack trace of any errors you encountered?

```
expected: "x = (123)"
 but was: "x = 1233)"
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes

BrendanHart:
Updated as it seems to be more general than just assignment, and also covering other expression types.

PR #5253, opened 2025-04-03
Determine parenthesis level for `ConstructorCallExpression` and `ConstantExpression` when using Groovy 2

BrendanHart:
## What's changed?

Added logic to determine the parenthesis level for `ConstructorCallExpression`s and `ConstantExpression`s when using Groovy 2.

## What's your motivation?

- #5236 

Not sure if it's the best idea to make use of the `determineParenthesisLevel` method for `ConstantExpression` as I think we just need to count the opening brackets at the start, and we could break early if we find a character that's not an opening bracket.

I've only added this for Groovy 2 since I think `_INSIDE_PARENTHESES_LEVEL` is set on these expressions when using Groovy 3.

I also subtract one from the result of `determineParenthesisLevel` to account for the brackets in `new Class(...)`. Are there any cases this doesn't work?

### Checklist

### DIFF
--- a/rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
+++ b/rewrite-groovy/src/groovy2Test/java/org/openrewrite/groovy/GroovyParserTest.java
@@ -49,44 +49,44 @@ class GroovyParserTest implements RewriteTest {
     @Test
     void shouldBeAbleToParseParenthesisedConstructorCallExpressions() {
         rewriteRun(
-            groovy(
-                """
-                (new BigDecimal(10))
-                """
-            )
+                groovy(
+                        """
+                        (new BigDecimal(10))
+                        """
+                )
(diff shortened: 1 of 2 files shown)


shanman190 (approved):
I think this all makes sense to me.

One little performance thing that we may think about in the future is if we reorganize the if checks into more of a set of categories: always, not Groovy 3.x, and Groovy 3.x. With the currently limited set of conditions, it's not going to change much, but if we see more conditions come forward having the version be the top branch could reduce the total number of checks needed overall.

jevanlingen:
I agree @shanman190 and used this PR to improve it right now 😇. Notice I also moved the BinaryExpression into the groovy 2 part, as we actually wrote this to support groovy 2 (groovy 3+ provides the `_INSIDE_PARENTHESES_LEVEL` flag for BinaryExpressions).

PR #5254, opened 2025-04-04
refactor: OpenRewrite Recipe best practices

timtebeek:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.recipes.rewrite.OpenRewriteRecipeBestPractices?organizationId=T3BlblJld3JpdGU%3D


greg-at-moderne:
Some of it was already done in #5249.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -225,8 +225,8 @@ private void recordSourceFileResult(@Nullable SourceFile before, @Nullable Sourc
         String beforePath = (before == null) ? "" : before.getSourcePath().toString();
         String afterPath = (after == null) ? "" : after.getSourcePath().toString();
         Recipe recipe = recipeStack.peek();
-        Long effortSeconds = (recipe.getEstimatedEffortPerOccurrence() == null || Result.isLocalAndHasNoChanges(before, after))
-                ? 0L : recipe.getEstimatedEffortPerOccurrence().getSeconds();
+        Long effortSeconds = (recipe.getEstimatedEffortPerOccurrence() == null || Result.isLocalAndHasNoChanges(before, after)) ?
+                0L : recipe.getEstimatedEffortPerOccurrence().getSeconds();
```
greg-at-moderne, line 229: No strong feelings, but I find it less readable.

(I know this is coming from some recipe)

greg-at-moderne (approved):
I haven't looked at the ANTLR-generated code.
BTW I think we should exclude it from the best-practices.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
timtebeek: Before I added this to the best practices we had 1/3rd on the next lines, and 2/3rd on the previous. That's why I opted for the most common style already to minimize the diff at the time.

PR #5260, opened 2025-04-06
`DataTableStore` so that RewriteRpc does not need to push `ExecutionContext` or `DataTable` over the wire.

jkschneider:
## What's changed?

`ExecutionContext` should refer to `DataTableStore` rather than directly storing rows on itself in memory. By default, Java-only recipe execution can still use an `InMemoryDataTableStore` that essentially behaves as it does now.

However, this will enable us to inject a different implementation into recipe runs that use `RewriteRpc` so that each process can push rows to a single store without necessarily passing these rows over the wire as control passes between the processes.

This will also ultimately help solve large data table downloading in the SaaS.

`DataTableStore` has an `acceptRows` method on it that can be used by `Preconditions` implementations to essentially shut off the flow of rows into the store by recipes being used as preconditions.

The implementation of https://github.com/openrewrite/rewrite-maven-plugin/pull/751 would have to change accordingly.


shanman190:
It's linked from the Maven issue, but figured that I'd link it here directly too. Here's the parallel Gradle PR for exporting data tables that would need to be adjusted as well.

https://github.com/openrewrite/rewrite-gradle-plugin/issues/201

PR #5263, opened 2025-04-07
Parser error on multiline javadoc comment

rlsanders4:
## What's changed?
This PR provides a test case to demonstrate a parsing error that occurs on javadoc comments:
```
[When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "com/abc/Server.java":
diff --git a/com/abc/Server.java b/com/abc/Server.java
index 8023007..29f2d3d 100644
--- a/com/abc/Server.java
+++ b/com/abc/Server.java
@@ -2,8 +2,7 @@ 
 public class Server {
   /**
    * @see com.abc.Server#load(
-   * java.lang.
-   * String)
+   * java.lang.Stringng)
    */
   public void load(String str) {
   }
] 
expected: 
  "package com.abc;
  public class Server {
    /**
     * @see com.abc.Server#load(
     * java.lang.
     * String)
     */
    public void load(String str) {
    }
  }"
 but was: 
  "package com.abc;
  public class Server {
    /**
     * @see com.abc.Server#load(
     * java.lang.Stringng)
     */
    public void load(String str) {
    }
  }"
```

@timtebeek 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
@@ -393,4 +393,24 @@ class JavaParserTest implements RewriteTest {
             """));
     }
 
+    @Test
+    void mutliLinePackageComment() {
+        rewriteRun(
+          // language=java
+          java(
+            """
+              package com.abc;
+              public class Server {
+                /**
+                 * @see com.abc.Server#load(
+                 * java.lang.
+                 * String)
+                 */
+                public void load(String str) {
+                }
+              }
+              """
+          )
+        );
+    }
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for the runnable example! Odd how we're repeating those last two characters. 🤔

rlsanders4:
I have not looked into a fix for this yet.

timtebeek (approved):
Thanks again for the runnable example. Should help to now parse & modify files containing these newlines going forward.

greg-at-moderne (approved):

PR #5265, opened 2025-04-07
Add Groovy context-free template

shanman190:
## What's changed?
Adding a context-free template for Groovy.

## What's your motivation?
Learning, mimicking the KotlinTemplate, possible usage for template support (context-free template cache and parameterization) for `G` types with respect to Gradle more specifically

Nothing in particular

@knutwannheden 

N/A

## Any additional context
Started with mimicking the KotlinTemplate in it's current state. I did override the `KotlinTemplate#contextSensitive` method to throw an exception as discussed previously with Knut given Kotlin -- and Groovy via this PR -- only support context-free templates. So requesting a context-sensitive template is an end user error at this time.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
@@ -249,11 +249,12 @@ public class GroovyParser implements Parser {
         /**
          * This is an internal API which is subject to removal or change.
          */
-        public Builder addClasspathEntry(Path classpath) {
-            if (this.classpath.isEmpty()) {
-                this.classpath = Collections.singletonList(classpath);
-            } else {
-                this.classpath.add(classpath);
+        public Builder addClasspathEntry(Path entry) {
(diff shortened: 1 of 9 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java
```diff
@@ -246,6 +246,18 @@ public Builder classpathFromResource(ExecutionContext ctx, String... artifactNam
...
 
+        /**
+         * This is an internal API which is subject to removal or change.
+         */
+        public Builder addClasspathEntry(Path classpath) {
+            if (this.classpath.isEmpty()) {
+                this.classpath = Collections.singletonList(classpath);
+            } else {
+                this.classpath.add(classpath);
+            }
```
sambsnyd, line 257: `Collections.singletonList()` returns an immutable list. So if this were to be called twice it would fail. Perhaps use `ListUtils.concat()` in the `else` clause.

knutwannheden (approved):

PR #5267, opened 2025-04-08
Add constructor calls to RemoveMethodInvocations

davidh44:
Recipe `RemoveMethodInvocations` currently does not cover constructor calls. This PR adds functionality for that

## What's your motivation?

Would like to remove certain POJO instantiations.

As a workaround for my use case, I used a `JavaVisitor`, checked the `JavaType`, and returned `null`.


davidh44:
Added logic for constructor removal and additional tests.

Two issues came up, wondering if you could provide some pointers:

**a)** 
Given that almost all constructor calls will be assigned to a variable, how can we remove the entire line? Should we use `visitVariableDeclaration()`? Can we view the assignment method there?

**b)** 
As above, we want to remove the entire line, including chained method calls, e.g., `new Date().toInstant()`. What's the proper way to get rid of the chained method calls?

PR #5274, opened 2025-04-09
XML parser - Accepting input with just \r characters as newlines

greg-at-moderne:
## What's changed?

Making XML parser accept files which use `\r` (CR) characters only as newlines.

## What's your motivation?

It fails with:
```
Syntax error in file.xml at line 1:21 extraneous input 'n' expecting '<'.java.lang.AssertionError: Failed to parse sources or run recipe
```

## Hint for the reviewers

The only changed code is the `rewrite-xml/src/main/antlr/XMLLexer.g4` and the test code. The rest is generated code.


Laurens-W:
Devils advocate: Could this change cause issues with serialized LSTs?

PR #5277, opened 2025-04-09
Upgrading Develocity plugin to v4

greg-at-moderne:
### Checklist


shanman190:
So just a note, but this has _some_ breaking potential if a user is utilizing a `MethodMatcher` against the parent type. It's a low chance, but just calling it out.

At this point, most Gradle recipes are being written in a more landmark orientated fashion due to the low type attribution.

timtebeek:
Do we see any reason to keep this open still? I'd vote to either merge or close.

greg-at-moderne:
Let me close.
We still have the old version pinned, but I feel no expert to understand the consequences as pointed out by Shannon here. Someone more knowledgable can proceed with this or similar fix.

PR #5278, opened 2025-04-09
Add support for Groovy `!in` operator

scottfrederick:
## What's changed?

The Groovy parser currently recognizes the [membership operator](https://docs.groovy-lang.org/latest/html/documentation/core-operators.html#_membership_operator) `in`. Groovy supports a negated form of this operator `!in`, which the parser currently does not recognize. 

When `!in` is used in a `settings.gradle` or `build.gradle` file, OpenRewrite gets an error like this: 

```
Caused by: java.lang.IllegalStateException: Unknown binary expression BinaryExpression
```

This PR adds `!in` to the set of recognized binary operators. 

## What's your motivation?
Allow existing `settings.gradle` and `build.gradle` files to be parsed and processed without modification. 

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -935,6 +935,9 @@ public class GroovyParserVisitor {
                     case "in":
                         gBinaryOp = G.Binary.Type.In;
                         break;
+                    case "!in":
+                        gBinaryOp = G.Binary.Type.NotIn;
+                        break;
                     case "<=>":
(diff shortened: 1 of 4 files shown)


timtebeek (commented):
Great to see an immediate fix, thanks! Let me run a quick check by colleagues to see if when we (de)serialize enums we ought to add entries to the end, or whether this is good to go.

/cc @knutwannheden @greg-at-moderne @sambsnyd

greg-at-moderne:
I am not aware of how serialization of LST works with regards to enums. Let's wait for the experts to chime in.

knutwannheden (approved):

greg-at-moderne:
Glad to hear this.

@scottfrederick, thanks for the contribution!

PR #5280, opened 2025-04-10
Add generic types to JavaTemplate

fmodesto:
## What's changed?
Added generic types parameters to JavaTemplate.
```java
final JavaTemplate before = JavaTemplate
        .builder("java.util.Optional.ofNullable(#{object:any(T)}).orElseGet(#{supplier:any(java.util.function.Supplier<S>)})")
        .genericTypes("T", "S extends T")
        .build();
```

## What's your motivation?
Increase refaster support

I hope KotlinTemplate still works, I have consolidated the logic for `any` and `anyArray`. 

@timtebeek @knutwannheden 

### Checklist


knutwannheden (approved):
Apart from a problem I noticed with the parameter "unsubstitution", which I am not 100% sure yet how we should fix, I think this looks like very good work! Thanks a lot for working on this, @fmodesto !

PR #5282, opened 2025-04-10
Introduce recipes to rename Type and Package in Java String literals

Laurens-W:
## What's changed?

After the revert of 
- https://github.com/openrewrite/rewrite/pull/5240

This PR now introduces 2 new recipes with the isolated implementations of changing String literals

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -227,11 +217,10 @@ public J visitImport(J.Import import_, ExecutionContext ctx) {
...
         public J visitLiteral(J.Literal literal, ExecutionContext ctx) {
             J.Literal lit = literal;
-            if (Boolean.TRUE.equals(visitLiterals) && literal.getType() == JavaType.Primitive.String) {
-                Pattern pat = Pattern.compile("\\b" + oldFullyQualifiedTypeName + "\\b");
+            if (literal.getType() == JavaType.Primitive.String) {
+                Pattern pat = Pattern.compile("(?:\\A|\\s)" + oldFullyQualifiedTypeName + "(?:|\\s)");
                 if (lit.getValue() != null && pat.matcher((String) lit.getValue()).find()) {
-                    lit = lit.withValue(((String) lit.getValue()).replace(oldFullyQualifiedTypeName, newFullyQualifiedTypeName))
-                            .withValueSource(lit.getValueSource().replace(oldFullyQualifiedTypeName, newFullyQualifiedTypeName));
+                    lit = lit.withValue(((String) lit.getValue()).replace(oldFullyQualifiedTypeName, newFullyQualifiedTypeName)).withValueSource(lit.getValueSource().replace(oldFullyQualifiedTypeName, newFullyQualifiedTypeName));
```
timtebeek, line 223: Were these changes intentional? Or part of a revert?

Laurens-W:
After merging this we will also have to revert commits in several other places:
- https://github.com/openrewrite/rewrite-migrate-java/commit/27f3c9566e3f8d5ace058cc5d367a4df39aad9b3
- https://github.com/openrewrite/rewrite-migrate-java/commit/c8794d2a98374cb8d62e3d8c8a832b9c9ce3bcb5
- https://github.com/openrewrite/rewrite-static-analysis/pull/501
- https://github.com/openrewrite/rewrite-micrometer/commit/f791f2bb2ac887034b0ff9a3d796a463f4d75186
- https://github.com/openrewrite/rewrite-logging-frameworks/commit/97024e4363d91e3687ce15786f076456c0787cf1
- https://github.com/openrewrite/rewrite-feature-flags/commit/60115a049c3bf46ba9a96a28d4c6b107ba80037b
- https://github.com/openrewrite/rewrite-testing-frameworks/commit/eacc21f5b04a9b339b158748b96fd688e95a34fe
- https://github.com/openrewrite/rewrite-cucumber-jvm/commit/ad750d3800ecb307e984dcc8a34621194113c41d
- https://github.com/openrewrite/rewrite-spring/commit/8833d81fdb7faccd2284b34fcc06ef7bfcc08430
- https://github.com/openrewrite/rewrite-apache/commit/7f5bab3e55c578ac7a725e660df64d14e004efbb
- https://github.com/openrewrite/rewrite-openapi/commit/175ffbd2ba7d476e34d449255896976c43b0a819
- https://github.com/openrewrite/rewrite-quarkus/commit/0af817d4899fa6891f719be50e1b96d019acadf1

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
Laurens-W: Part of a revert, my bad

timtebeek (approved):

timtebeek:
We've reverted the original change in https://github.com/openrewrite/rewrite/commit/f87d1adfe7d41f6379cc5270ef036f2571b693d6 ; we'll reconsider, perhaps as a separate recipe the ability to change String literals, both out of concern for performance (precondition does a full visit) as well as unintended changes.

Laurens-W:
> precondition does a full visit
I've tried to reduce the impact of this by checking whether we've already made a match but it will still visit all literals. I couldn't see a mechanism in place like the `stopAfterPreVisit` to prevent this

Laurens-W:
After this PR we should add these new recipes as outlined by this previously merged and reverted PR
- https://github.com/openrewrite/rewrite-migrate-java/pull/693

knutwannheden:
Sometimes it is just better to not have any precondition. In this recipe I think there is not much of a point in having one.

Laurens-W:
> Sometimes it is just better to not have any precondition. In this recipe I think there is not much of a point in having one.

So should I move checking whether there are any matches in the preVisit? Or just not check at all?

timtebeek:
I think not check at all in the precondition; as we're going to have to visit every String literal anyway; might as well only do so once instead of twice.

knutwannheden:
Yes, what Tim said 😄 Visiting all at least once is required in any case, as the edit visitor needs to do this and whatever the precondition visitor is on top of that. Unless the precondition visitor can check more efficiently, there is no point in having it.

Laurens-W:
Currently the precondition stops visiting after it finds _any_ match. Perhaps I can change the implementation so that the precondition builds a list of matches and that get's used to specifically visit those elements.
But either way that means visiting some literals more than once.

I'll remove the precondition all together on monday!

rewrite-java/src/main/java/org/openrewrite/java/ChangeTypeInStringLiteral.java
```diff
@@ -0,0 +1,118 @@
...
+            if (literal.getType() == JavaType.Primitive.String && lit.getValue() != null) {
+                Matcher matcher = stringLiteralPattern.matcher((String) lit.getValue());
+                if (matcher.find()) {
+                    lit = lit.withValue(matcher.replaceAll(newFullyQualifiedTypeName))
+                            .withValueSource(stringLiteralPattern.matcher(lit.getValueSource()).replaceAll(newFullyQualifiedTypeName));
+                }
+            }
+            return super.visitLiteral(lit, ctx);
+        }
+    }
```
timtebeek, line 117: Why these levels of indirection? `JavaVisitor` already has a `isAcceptable` that checks `sourceFile instanceof JavaSourceFile`.
```suggestion
    public TreeVisitor<?, ExecutionContext> getVisitor() {
        final Pattern stringLiteralPattern = Pattern.compile("\\b" + oldFullyQualifiedTypeName + "\\b");
        return new JavaVisitor<ExecutionContext>() {
            @Override
            public J visitLiteral(J.Literal literal, ExecutionContext ctx) {
                J.Literal lit = literal;
                if (literal.getType() == JavaType.Primitive.String && lit.getValue() != null) {
                    Matcher matcher = stringLiteralPattern.matcher((String) lit.getValue());
                    if (matcher.find()) {
                        lit = lit.withValue(matcher.replaceAll(newFullyQualifiedTypeName))
                                .withValueSource(stringLiteralPattern.matcher(lit.getValueSource()).replaceAll(newFullyQualifiedTypeName));
                    }
                }
                return super.visitLiteral(lit, ctx);
            }
        };
    }
```

PR #5288, opened 2025-04-11
feat: merge categories which have matching roots

lkerford:
## What's changed?
In the event we "org.openrewrite.java.test" and "io.moderne.java.test", we want the subcategories "java" and "test" to be merged.
-https://github.com/moderneinc/moderne-recipe-execution/issues/1270

## What's your motivation?
We are seeing duplicate entries in the market place and this is confusing. 

## Any additional context
To achieve the removal of these duplicate categories we decided to merge them in the tree. 
```
The path to a category below a category root today (before any changes) is <categoryRoot>.path.to.sub.category, 
for instance, org.openrewrite.path.to.subcategory. We will instead make that path be path.to.sub.category. 
That means we will merge paths of categories such as io.moderne.java and org.openrewrite.java into java. 
If there are subcategories such as io.moderne.java.spring and org.openrewrite.java.spring, these will
 also be merged into java.spring.
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/CategoryDescriptor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/CategoryDescriptor.java
@@ -52,4 +52,6 @@ public class CategoryDescriptor {
     int priority;
 
     boolean synthetic;
+
+    Set<String> rootPackages;
 }
--- a/rewrite-core/src/main/java/org/openrewrite/config/CategoryTree.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/CategoryTree.java
@@ -66,7 +66,7 @@ public class CategoryTree<G> {
     public static class Root<G> extends CategoryTree<G> {
         private static final CategoryDescriptor ROOT_DESCRIPTOR = new CategoryDescriptor(
                 "ε", "", "", emptySet(), true,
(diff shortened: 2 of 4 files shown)


rewrite-core/src/main/java/org/openrewrite/config/CategoryDescriptor.java
```diff
@@ -52,4 +52,6 @@ public class CategoryDescriptor {
     int priority;
 
     boolean synthetic;
+
+    Set<String> rootPackages;
```
jkschneider, line 56: Something looks suspicious about this to me. We shouldn't have both `packageName` and `Set` of `rootPackages`. Which of the set of `rootPackages` does `packageName` represent?

lkerford: `packageName` does not include the root so that we can merge the packages together 
while the `rootPackages` contains all the know subset of roots. 

For example `org.openrewrite.path.to.subcategory` has a `packageName` of `path.to.subcategory` and a `rootPackages` of listof(`org.openrewrite`)

pstreef: I guess we should rename `packageName` then, to avoid this confusion.

PR #5293, opened 2025-04-14
Groovy support for multiple assignment/tuple unpacking

### LINKED ISSUE #5283: Groovy parser fails with parentheses on the left-hand side of the assignment

Current main = fc68c8dcebcaeff0846fc4381edbd296b966ec1e

The following test fails:
```java
    @Test
    void defParen() {
        rewriteRun(
                groovy(
                        """
                        def (key, value) = "a1:b2".split(":")
                        println(key)
                        println(value)
                        """
                )
        );
    }
```

with:
```
Caused by: java.lang.ClassCastException: class org.codehaus.groovy.ast.expr.ArgumentListExpression cannot be cast to class org.codehaus.groovy.ast.expr.VariableExpression (org.codehaus.groovy.ast.expr.ArgumentListExpression and org.codehaus.groovy.ast.expr.VariableExpression are in unnamed module of loader 'app')
	at org.codehaus.groovy.ast.expr.DeclarationExpression.getVariableExpression(DeclarationExpression.java:103)
	at org.openrewrite.groovy.GroovyParserVisitor$RewriteGroovyVisitor.visitDeclarationExpression(GroovyParserVisitor.java:1363)
	at org.codehaus.groovy.ast.expr.DeclarationExpression.visit(DeclarationExpression.java:89)
	at org.codehaus.groovy.ast.CodeVisitorSupport.visitExpressionStatement(CodeVisitorSupport.java:120)
```

which I think is a valid Groovy program.

## Context

I found the same pattern in an open-source: https://github.com/liquibase/liquibase-gradle-plugin/blob/966efedabb949b5eae661a1e0fed4461ece187ed/src/main/groovy/org/liquibase/gradle/ArgumentBuilder.groovy#L269

sudouser777:
1. Updated the groovy parser to support tuple unpacking.
2. closes #5283 .

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1363,29 +1363,30 @@ public class GroovyParserVisitor {
             List<J.Annotation> leadingAnnotations = visitAndGetAnnotations(expression, classVisitor);
             Optional<MultiVariable> multiVariable = maybeMultiVariable();
             List<J.Modifier> modifiers = getModifiers();
-            TypeTree typeExpr = visitVariableExpressionType(expression.getVariableExpression());
-
-            J.VariableDeclarations.NamedVariable namedVariable;
+            TypeTree typeExpr;
+            List<J.VariableDeclarations.NamedVariable> namedVariables;
             if (expression.isMultipleAssignmentDeclaration()) {
-                // def (a, b) = [1, 2]
-                throw new UnsupportedOperationException("Parsing multiple assignment (e.g.: def (a, b) = [1, 2]) is not implemented");
+                Space prefixBeforeOpenParentheses = whitespace();
+                typeExpr = visitTupleExpressionType(expression.getTupleExpression());
+                JContainer<J.Identifier> identifiers = visit(expression.getTupleExpression());
+                namedVariables = identifiers.getElements().stream()
+                        .map(this::createNamedVariable)
+                        .collect(toList());
+                int first = 0, last = namedVariables.size() - 1;
+                namedVariables.set(first, namedVariables.get(first).withPrefix(prefixBeforeOpenParentheses).withMarkers(Markers.build(singletonList(new OpenParentheses(randomId())))));
+                namedVariables.set(last, namedVariables.get(last).withMarkers(Markers.build(singletonList(new CloseParentheses(randomId())))));
             } else {
+                typeExpr = visitVariableExpressionType(expression.getVariableExpression());
                 J.Identifier name = visit(expression.getVariableExpression());
-                namedVariable = new J.VariableDeclarations.NamedVariable(
-                        randomId(),
-                        name.getPrefix(),
-                        Markers.EMPTY,
-                        name.withPrefix(EMPTY),
-                        emptyList(),
-                        null,
-                        typeMapping.variableType(name.getSimpleName(), typeExpr.getType())
-                );
+                namedVariables = new ArrayList<>(singletonList(createNamedVariable(name)));
             }
 
             if (!(expression.getRightExpression() instanceof EmptyExpression)) {
                 Space beforeAssign = sourceBefore("=");
                 Expression initializer = visit(expression.getRightExpression());
-                namedVariable = namedVariable.getPadding().withInitializer(padLeft(beforeAssign, initializer));
+
+                // For destructuring assignments, add the initializer to the last variable
+                namedVariables.set(namedVariables.size()-1, namedVariables.get(namedVariables.size()-1).getPadding().withInitializer(padLeft(beforeAssign, initializer)));
             }
 
             J.VariableDeclarations variableDeclarations = new J.VariableDeclarations(
@@ -1397,13 +1398,25 @@ public class GroovyParserVisitor {
                     typeExpr,
                     null,
                     emptyList(),
-                    singletonList(JRightPadded.build(namedVariable))
+                    namedVariables.stream().map(JRightPadded::build).collect(toList())
             );
             if (multiVariable.isPresent()) {
                 variableDeclarations = variableDeclarations.withMarkers(variableDeclarations.getMarkers().add(multiVariable.get()));
             }
 
             queue.add(variableDeclarations);
+        }
+
+        private J.VariableDeclarations.NamedVariable createNamedVariable(J.Identifier name) {
+            return new J.VariableDeclarations.NamedVariable(
+                    randomId(),
+                    name.getPrefix(),
+                    Markers.EMPTY,
+                    name.withPrefix(EMPTY),
+                    emptyList(),
+                    null,
+                    typeMapping.variableType(name.getSimpleName(), name.getType())
+            );
         }
 
         private Optional<MultiVariable> maybeMultiVariable() {
@@ -2014,6 +2027,12 @@ public class GroovyParserVisitor {
                     JLeftPadded.build(operator),
                     visit(unary.getExpression()),
                     null));
+        }
+
+        public TypeTree visitTupleExpressionType(TupleExpression expression) {
+            String typeName = "";
+            JavaType type = typeMapping.type(expression.getType());
+            return new J.Identifier(randomId(), EMPTY, Markers.EMPTY, emptyList(), typeName, type, null);
         }
 
         public TypeTree visitVariableExpressionType(VariableExpression expression) {
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/DestructuringTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/DestructuringTest.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.groovy.tree;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.groovy.Assertions.groovy;
+
+class DestructuringTest implements RewriteTest {
+
+    @Test
+    void basicDestructuring() {
+        rewriteRun(
+          groovy(
+            """
+              def (key, value) = "a1:b2".split(":")
+              println(key)
+              println(value)
+              """
+          )
+        );
+    }
+
+    @Disabled
+    @Test
+    void destructuringWithSpaces() {
+        rewriteRun(
+          groovy(
+            """
+              def ( key, value ) = "a1:b2".split(":")
+              println(key)
+              println(value)
+              """
+          )
+        );
+    }
+
+    @Disabled
+    @Test
+    void destructuringWithType() {
+        rewriteRun(
+          groovy(
+            """
+              def (String key, String value) = "a1:b2".split(":")
+              println(key)
+              println(value)
+              """
+          )
+        );
+    }
+
+    @Test
+    void destructuringWithList() {
+        rewriteRun(
+          groovy(
+            """
+              def (first, second, third) = [1, 2, 3]
+              println(first)
+              println(second)
+              println(third)
+              """
+          )
+        );
+    }
+
+    @Test
+    void destructuringWithMap() {
+        rewriteRun(
+          groovy(
+            """
+              def map = [a: 1, b: 2]
+              def (key, value) = map.entrySet().first()
+              println(key)
+              println(value)
+              """
+          )
+        );
+    }
+}
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyPrinter.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyPrinter.java
@@ -300,6 +300,27 @@ public class GroovyPrinter<P> extends GroovyVisitor<PrintOutputCapture<P>> {
         }
 
         @Override
+        public J visitVariable(J.VariableDeclarations.NamedVariable variable, PrintOutputCapture<P> p) {
+            beforeSyntax(variable, Space.Location.VARIABLE_PREFIX, p);
+            if(variable.getMarkers().findFirst(OpenParentheses.class).isPresent()) {
+                p.append('(');
+            }
+            visit(variable.getName(), p);
+            for (JLeftPadded<Space> dimension : variable.getDimensionsAfterName()) {
+                visitSpace(dimension.getBefore(), Space.Location.DIMENSION_PREFIX, p);
+                p.append('[');
+                visitSpace(dimension.getElement(), Space.Location.DIMENSION, p);
+                p.append(']');
+            }
+            if(variable.getMarkers().findFirst(CloseParentheses.class).isPresent()) {
+                p.append(')');
+            }
+            visitLeftPadded("=", variable.getPadding().getInitializer(), JLeftPadded.Location.VARIABLE_INITIALIZER, p);
+            afterSyntax(variable, p);
+            return variable;
+        }
+
+        @Override
         public J visitLambda(J.Lambda lambda, PrintOutputCapture<P> p) {
             beforeSyntax(lambda, Space.Location.LAMBDA_PREFIX, p);
             LambdaStyle ls = lambda.getMarkers().findFirst(LambdaStyle.class)
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/marker/CloseParentheses.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/marker/CloseParentheses.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.groovy.marker;
+
+import lombok.Value;
+import lombok.With;
+import org.openrewrite.marker.Marker;
+
+import java.util.UUID;
+
+/**
+ * Marker for multiple assignment variables.
+ * For example, in `def (a, b) = [1, 2]`, the variables are `a` and `b`.
+ */
+@Value
+@With
+public class CloseParentheses implements Marker {
+    UUID id;
+}
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/marker/OpenParentheses.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/marker/OpenParentheses.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.groovy.marker;
+
+import lombok.Value;
+import lombok.With;
+import org.openrewrite.marker.Marker;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Marker for multiple assignment variables.
+ * For example, in `def (a, b) = [1, 2]`, the variables are `a` and `b`.
+ */
+@Value
+@With
+public class OpenParentheses implements Marker {
+    UUID id;
+}


sudouser777:
I have tried to hanlde the spaces situation, but it's tricky spaces are not being handled properly
```
    @Test
    void destructuringWithSpaces() {
        rewriteRun(
          groovy(
            """
              def ( key, value ) = "a1:b2".split(":")
              println(key)
              println(value)
              """
          )
        );
    }
  ```
  Also static type also not working
  ```
    @Test
    void destructuringWithType() {
        rewriteRun(
          groovy(
            """
              def (String key, String value) = "a1:b2".split(":")
              println(key)
              println(value)
              """
          )
        );
    }
  ```
  
@greg-at-moderne I'm not sure if my approach is correct—open to any suggestions.

File: rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/DestructuringTest.java [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+              println(first)
+              println(second)
+              println(third)
+              """
+          )
+        );
+    }
+
+    @Test
+    void destructuringWithMap() {
```
shanman190, line 82: I don't think this would evaluate as valid Groovy since destructuring requires indexes which isn't the case here in this test.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1363,29 +1363,30 @@ public void visitDeclarationExpression(DeclarationExpression expression) {
...
-                throw new UnsupportedOperationException("Parsing multiple assignment (e.g.: def (a, b) = [1, 2]) is not implemented");
+                Space prefixBeforeOpenParentheses = whitespace();
+                typeExpr = visitTupleExpressionType(expression.getTupleExpression());
+                JContainer<J.Identifier> identifiers = visit(expression.getTupleExpression());
+                namedVariables = identifiers.getElements().stream()
+                        .map(this::createNamedVariable)
+                        .collect(toList());
+                int first = 0, last = namedVariables.size() - 1;
+                namedVariables.set(first, namedVariables.get(first).withPrefix(prefixBeforeOpenParentheses).withMarkers(Markers.build(singletonList(new OpenParentheses(randomId())))));
+                namedVariables.set(last, namedVariables.get(last).withMarkers(Markers.build(singletonList(new CloseParentheses(randomId())))));
```
shanman190, line 1377: This is usually a `JContainer` that holds the named variables, then wrapped inside of a `J.Parentheses` instead. This then means that the markers wouldn't be necessary.

sudouser777: But it's not including parenthesis, it's just returning `JRightPadded<J.Identifier>`

shanman190: Yes, originally I was talking in more abstract senses.

Specifically for this case, you're getting the `JContainer` back from visiting the identifiers. That is then needing to go directly into `J.NamedVariables`. So in this case, you'd then wrap the `J.NamedVariables` -- if parentheses were present -- into a `J.Parentheses` and return that. The parent element accepts any `Expression` type on the left hand side of the `J.Assignment` that is being built and `J.Parentheses` would meet that constraint.

shanman190: This was more in general. The way to add parentheses around an element is with `J.Parentheses` and when that element is a list it is typically wrapped in a `JContainer`.

I'm not sure what elements you currently have in scope from the Groovy AST, but I'd expect that something similar would be the case here.

sudouser777: It's using the same `J.VariableDeclarations` for normal variable declarations. This will require changes to the `J.VariableDeclarations` class. I'm not entirely sure about this approach, though.

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/DestructuringTest.java
greg-at-moderne: Indeed, it fails when I tried https://onecompiler.com/groovy/43ewkh679

sudouser777: I removed this test case.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
greg-at-moderne, line 1377: I think the typical way to do that in OpenRewrite code base is to use `ListUtils.mapFirst()` and `ListUtils.mapLast()`

greg-at-moderne (approved):

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
greg-at-moderne: Changed in 12f0d56da7aeda50bb2bf9b142c6e78991c52d9b using Github UI without running the code. I hope it will work.

timtebeek:
Thanks a lot for the continued improvements to the Groovy parser @sudouser777 ! Great to see steady progress being made.

Any reason you'd held off on a merge here @greg-at-moderne ?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1415,29 +1415,29 @@ public void visitDeclarationExpression(DeclarationExpression expression) {
...
-                // def (a, b) = [1, 2]
-                throw new UnsupportedOperationException("Parsing multiple assignment (e.g.: def (a, b) = [1, 2]) is not implemented");
+                Space prefixBeforeOpenParentheses = whitespace();
+                typeExpr = visitTupleExpressionType(expression.getTupleExpression());
+                JContainer<J.Identifier> identifiers = visit(expression.getTupleExpression());
+                namedVariables = identifiers.getElements().stream()
+                        .map(this::createNamedVariable)
+                        .collect(toList());
+                namedVariables = ListUtils.mapFirst(namedVariables, first -> first.withPrefix(prefixBeforeOpenParentheses).withMarkers(Markers.build(singletonList(new OpenParentheses(randomId())))));
+                namedVariables = ListUtils.mapLast(namedVariables, last -> last.withMarkers(Markers.build(singletonList(new CloseParentheses(randomId())))));
```
timtebeek, line 1428: Rather than adding new markers, is there any way we can make this work with regular `J.Parentheses`?

sudouser777: It's using `J.VariableDeclarations`, which is also used for normal variable declarations. If we add `J.Parentheses`, I'm not sure how many changes it would require in the existing code, and I'm also not sure whether it's possible to make the parentheses optional.

I'm open to suggestions—let me know if there's a better way we can handle this.

greg-at-moderne (approved):

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
timtebeek: Indeed looking in more detail now probably best not to make any structural changes here. Thanks for explaining!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
timtebeek: Perhaps indeed best not to make any structural changes there, as it would mean we'd have to handle that separately downstream when the left hand side can be an instance of J.Parentheses. I'd say a decent trade off to use markers here.

shanman190: I don't think this necessarily needs to be a part of this PR in particular, but this choice here does set a precedent into motion that has a pretty large ripple effect.

As a more specific example, Groovy, Kotlin, C#, Python, and JS (at least) all support destructing-like syntaxes. So by choosing markers here, we too are also likely choosing markers for all of those other languages as well.

To compound on this, the Java community periodically has discussions about adding destructuring support natively, but none of those discussions have yet to raise an accepted JEP.

@sambsnyd or @jkschneider, would either of you want to weigh in here as well?

- Groovy: this PR
- Kotlin (specific structural type): https://github.com/openrewrite/rewrite/blob/main/rewrite-kotlin%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fkotlin%2Ftree%2FK.java#L899
- C# (specific structural type): https://github.com/openrewrite/rewrite-csharp/blob/main/rewrite-csharp/src/main/java/org/openrewrite/csharp/tree/Cs.java#L4066
- Python: doesn't appear to yet be supported
- JS (specific structural type): supported via various binding assignment declarations

jevanlingen:
Merging it seems good to me, but I would create a new "Groovy parser does not support multiple assignment/tuple unpacking with whitespace" issue though.

timtebeek (changes requested):
Ok Shannon makes a valid point here, in that we're better off using a new specific structural type here to match the others
- https://github.com/openrewrite/rewrite/pull/5293#discussion_r2148383540

No need to pull that up into a generalized version just yet, but we should revise the approach taken here to remove the markers and introduce a new element.

Would fully understand if at this point you have other priorities @sudouser777 ; just let us know if you'd prefer someone takes over.

sudouser777:
> Ok Shannon makes a valid point here, in that we're better off using a new specific structural type here to match the others
> 
> * [Groovy support for multiple assignment/tuple unpacking #5293 (comment)](https://github.com/openrewrite/rewrite/pull/5293#discussion_r2148383540)
> 
> No need to pull that up into a generalized version just yet, but we should revise the approach taken here to remove the markers and introduce a new element.
> 
> Would fully understand if at this point you have other priorities @sudouser777 ; just let us know if you'd prefer someone takes over.

Sounds good. I’m happy to take a stab at the changes, though I’m not too familiar with the structural type part. If the changes are a bit involved, totally fine if someone else wants to take over—just let me know!

jevanlingen:
If you want to give it a try @sudouser777, that would be awesome! Don't worry, if after some time you decide you don't want to work on it anymore (because it's either too hard,, will take you too much time or whatever reason) that's totally fine. Just put a comment out that you would like someone else to take over.

The structural type part sounds a little more advanced, but it is "just" another LST element. I think reading the Kotlin counterpart will be a great starter:
https://github.com/openrewrite/rewrite/blob/1112346c651847e633e773c8d51769db52c2de9a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java#L2343-L2452

_In short, Kotlin has the `KtDestructuringDeclaration` AST object, where the Groovy parser uses a `TupleExpression` object._

sudouser777:
Thanks! @jevanlingen . I’ll check out the Kotlin example and let you know if I can take on the changes. If it’s urgent, I’m totally fine with someone else taking over.

rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
```diff
@@ -1020,4 +1020,109 @@ public G.Unary withOperator(JLeftPadded<G.Unary.Type> operator) {
             }
         }
     }
+
+    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    @RequiredArgsConstructor
+    @AllArgsConstructor(access = AccessLevel.PRIVATE)
+    final class MultipleAssignmentDeclaration implements G, Statement {
```
shanman190: I'm inclined to call it a `DestructuringDeclaration`, but open to options.

rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
```diff
@@ -1020,4 +1020,109 @@ public G.Unary withOperator(JLeftPadded<G.Unary.Type> operator) {
...
+        @Nullable
+        @Getter
+        TypeTree typeExpression;
+
+
+        @Getter
+        @With
+        JLeftPadded<Expression> initializer;
+
+        JContainer<VariableDeclarations.NamedVariable> elements;
```
shanman190: Should this be `variables` (and associated methods)?

rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
sudouser777: I can change it if we are ok with the one you suggested

rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
sudouser777: I will update it

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
```diff
@@ -186,4 +186,8 @@ public <J2 extends J> JContainer<J2> visitContainer(JContainer<J2> container,
                                                         GContainer.Location loc, P p) {
         return super.visitContainer(container, JContainer.Location.LANGUAGE_EXTENSION, p);
     }
+
+    public J visitMultipleAssignmentDeclaration(G.DestructuringDeclaration destructuringDeclaration, P p) {
```
shanman190: Missed one?

shanman190:
Looks good to me other than the one missed method rename. Thanks @sudouser777 !

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
sudouser777: Changed it

rewrite-groovy/src/main/java/org/openrewrite/groovy/tree/G.java
```diff
@@ -1020,4 +1020,109 @@ public G.Unary withOperator(JLeftPadded<G.Unary.Type> operator) {
             }
         }
     }
+
+    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    @RequiredArgsConstructor
+    @AllArgsConstructor(access = AccessLevel.PRIVATE)
+    final class DestructuringDeclaration implements G, Statement {
```
jevanlingen, line 1028: I am wondering if it would make sense that the DestructuringDeclaration class would extend the VariableDeclarations class. For the Kotlin variant this is also _**not**_ done, but maybe it would make sense to do so? What do you think @shanman190 and @greg-at-moderne?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
```diff
@@ -186,4 +186,8 @@ public <J2 extends J> JContainer<J2> visitContainer(JContainer<J2> container,
                                                         GContainer.Location loc, P p) {
         return super.visitContainer(container, JContainer.Location.LANGUAGE_EXTENSION, p);
     }
+
+    public J visitDestructuringDeclaration(G.DestructuringDeclaration destructuringDeclaration, P p) {
+        return destructuringDeclaration;
```
timtebeek: I'd perhaps have expected to at least call out to visit the initializer and variables here, much like we already do in the printer. Any particular reason those where left out? Mostly curious as recipes that aim to visit either destructuring declarations, or their subcomponents might miss those otherwise.

sudouser777: I'm just a bit confused about what's expected in this method — I thought only the visitor used in the printer would be needed.

jevanlingen: > I'm just a bit confused about what's expected in this method

This `visitDestructuringDeclaration` method is actually called when the groovy source tree is traversed, thus it's possible to use it when you create a recipe with a GroovyVisitor. But the subparts, for example when you use `visitContainer` in your recipe, will not be called if you "just" return `destructuringDeclaration` here. So this method needs to actually call the subparts.

Maybe this still sounds a little vague, so I created the implementation already, based on the K.DestructuringDeclaration: https://github.com/openrewrite/rewrite/pull/5293/commits/202167a0ffaa815b7a22efe3e86b3b9c6fe52ee1.

timtebeek:
Copying the failure we're seeing here for easier reference:
```
DestructuringTest > destructuringWithType() FAILED
    def (~~(non-whitespace)~~>String <~~key,~~(non-whitespace)~~> String <~~value) = "a1:b2".split(":")
    println("${key} ${value}")
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:324)
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:125)
        at org.openrewrite.groovy.tree.DestructuringTest.destructuringWithType(DestructuringTest.java:67)
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyVisitor.java
```diff
@@ -182,8 +182,23 @@ public Space visitSpace(Space space, GSpace.Location loc, P p) {
...
+        if (!(temp instanceof G.DestructuringDeclaration)) {
+            return temp;
+        } else {
+            d = (G.DestructuringDeclaration) temp;
+        }
+        d = d.withInitializer(visitAndCast(d.getInitializer(), p));
+        d = d.getPadding().withVariables(d.getPadding().getVariables());
+        d = d.getPadding().withVariables(visitContainer(d.getPadding().getVariables(), GContainer.Location.DESTRUCT_ASSIGNMENTS, p));
+        return d;
+    }
```
timtebeek, line 194: Linking the reference implementation here, for easy comparison.
https://github.com/openrewrite/rewrite/blob/01917b0213ee48dee46764f1e3acefec2dcdb9a7/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinVisitor.java#L142-L155

timtebeek:
Asking @jkschneider for a final review, as he suggested to introduce the `G.DestructuringDeclaration` element, and we want to ensure the design aligns with what he had intended, especially as LST changes are hard to change later on due to recipes having been writting using the elements and visitor, and LSTs potentially serialized.

jevanlingen:
```
DestructuringTest > destructuringWithType() FAILED
    def (~~(non-whitespace)~~>String <~~key,~~(non-whitespace)~~> String <~~value) = "a1:b2".split(":")
    println("${key} ${value}")
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:324)
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:125)
        at org.openrewrite.groovy.tree.DestructuringTest.destructuringWithType(DestructuringTest.java:67
```

This happens because of my change:
```java
public J visitDestructuringDeclaration(G.DestructuringDeclaration destructuringDeclaration, P p) {
  // ..
  d = d.getPadding().withVariables(visitContainer(d.getPadding().getVariables(), GContainer.Location.DESTRUCT_ASSIGNMENTS, p));
  return d;
}
```

Will call the subparts under the hood. Now when the

```
public J visitVariable(J.VariableDeclarations.NamedVariable variable, P p) {
  J.VariableDeclarations.NamedVariable v = variable;
  // visit members + build `v`
  return v;
}
```

is visited, the prefix is filled with the type (thus `"~~(non-whitespace)~~>String <~~key"` happens). Will spend more time tomorrow to look more into this.

sudouser777:
I tried to debug the issue. Found something.
image
Somehow type is coming in the prefix
@timtebeek @jevanlingen This is what causing above failure. I'm not sure on how to fix it though. I tried few things but they didn't work.

jevanlingen:
@timtebeek & @sudouser777 Types within the destructuring were not handled, which caused them to be treated as part of the prefix. This only became apparent once we started implementing `visitDestructuringDeclaration`, since that’s when we traverse the tree again and the check "Source file was parsed into an LST that contains non-whitespace characters in its whitespace" kicks in.

I [resolved](https://github.com/openrewrite/rewrite/pull/5293/commits/3e113185013ccb03b55a2cffa18ada54765ce54e) it by introducing an additional type marker. However, it makes me wonder whether we should introduce a `G.VariableDeclarations.NamedVariable`, which would explicitly allow attaching the type directly.

sudouser777:
> I resolved it by introducing an additional type marker. However, it makes me wonder whether we should introduce a `G.VariableDeclarations.NamedVariable`, which would explicitly allow attaching the type directly.

We can also change `variables` to use `JContainer<VariableDeclarations> variables` instead of ` JContainer<VariableDeclarations.NamedVariable> variables`
But it may complicate space handling.

jevanlingen:
Ah that's smart, didn't think about that! We modeled it after Kotlin deconstruction model, which does not have type declaration within the destruction¹, thus the model was not really tailored for having types within the destruction parts. I have to do work on other stuff now, but if you would have the time to look a bit into that `JContainer<VariableDeclarations> variables` direction instead of the custom marker, that would be awesome!

¹ _Or at least, not right now. See also [Kotlin 2.4 Introduces Name-based Destructuring: Stop Guessing Which Variable Goes Where](https://xuanlocle.medium.com/kotlin-2-4-introduces-name-based-destructuring-stop-guessing-which-variable-goes-where-2d22fa733837.)._

sudouser777:
@jevanlingen  I made the changes to use `VariableDeclarations` please check and let me know if you see any issues with the new changes.

jevanlingen:
Very nice @sudouser777, much better than my marker solution! I did make some small final changes, I think it's good now.

jevanlingen (approved):

greg-at-moderne:
To be honest I stopped following this PR and now it's hard to catch up with so many comments and history.

@timtebeek, @jevanlingen, @sudouser777, are there any reasons NOT to merge it as is?

sudouser777:
> To be honest I stopped following this PR and now it's hard to catch up with so many comments and history.
> 
> @timtebeek, @jevanlingen, @sudouser777, are there any reasons NOT to merge it as is?

@greg-at-moderne I’ve completed all the requested changes — from my side, we’re good to merge. I thought you all might still be waiting on something.

timtebeek (commented):
hi @sudouser777 ; thanks a lot so far! The reason it's been taking longer is that we're internally looking at the best way to map this across languages; Your work has helped kick off those discussions, so there's value even if we for now plan a slightly different direction. We won't ask you to change things yet again, but we're eyeing the use of
https://github.com/openrewrite/rewrite/blob/f5f5a9ac880fcac473535326ef8b4366462566e3/rewrite-java/src/main/java/org/openrewrite/java/tree/VariableDeclarator.java#L23-L31

That would similarly allow for multiple assignments, without the need for the new `DestructuringDeclaration` introduced here. This has been part of ongoing discovery, so don't feel like you have to keep up or make these changes. Just sharing our thoughts here such that you're looped in. We can take over your PR if that's ok with you, or you're welcome to stay involved; whichever you prefer.

sudouser777:
> hi @sudouser777 ; thanks a lot so far! The reason it's been taking longer is that we're internally looking at the best way to map this across languages; Your work has helped kick off those discussions, so there's value even if we for now plan a slightly different direction. We won't ask you to change things yet again, but we're eyeing the use of
> https://github.com/openrewrite/rewrite/blob/f5f5a9ac880fcac473535326ef8b4366462566e3/rewrite-java/src/main/java/org/openrewrite/java/tree/VariableDeclarator.java#L23-L31
> 
> That would similarly allow for multiple assignments, without the need for the new `DestructuringDeclaration` introduced here. This has been part of ongoing discovery, so don't feel like you have to keep up or make these changes. Just sharing our thoughts here such that you're looped in. We can take over your PR if that's ok with you, or you're welcome to stay involved; whichever you prefer.

Hi @timtebeek, apologies for the very delayed reply — I got caught up with some personal things and wasn’t able to get back to this sooner. Thanks a lot for the update and for keeping me in the loop. I’m totally fine with you taking over the PR. If you need anything from my side, feel free to reach out!

timtebeek:
No worries; I hope all is well there. We'll try to work this in, unless you beat us to it. :)

timtebeek:
Thanks again for the help here @sudouser777 ! Your efforts help shaped the solution we've landed on now. 🙏🏻

PR #5301, opened 2025-04-15
Add AddCommentToClass

SiBorea:
## What's changed?
Add a new recipe `AddCommentToClass`, similar to `AddCommentToMethod`

### Checklist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,5 +2,3 @@ org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.workers.max=6
 org.gradle.jvmargs=-Xmx3g
-org.gradle.java.installations.paths=/Users/xchen21/Documents/JDK/8/zulu-8.jdk,/Users/xchen21/Documents/JDK/11/zulu-11.jdk,/Users/xchen21/Documents/JDK/17/zulu-17.jdk,/Users/xchen21/Documents/JDK/21/zulu-21.jdk
-version=8.47.1-SNAPSHOT
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,3 +2,5 @@ org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.workers.max=6
 org.gradle.jvmargs=-Xmx3g
+org.gradle.java.installations.paths=/Users/xchen21/Documents/JDK/8/zulu-8.jdk,/Users/xchen21/Documents/JDK/11/zulu-11.jdk,/Users/xchen21/Documents/JDK/17/zulu-17.jdk,/Users/xchen21/Documents/JDK/21/zulu-21.jdk
+version=8.47.1-SNAPSHOT
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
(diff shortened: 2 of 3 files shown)


gradle.properties
```diff
@@ -2,3 +2,5 @@ org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.workers.max=6
 org.gradle.jvmargs=-Xmx3g
+org.gradle.java.installations.paths=/Users/xchen21/Documents/JDK/8/zulu-8.jdk,/Users/xchen21/Documents/JDK/11/zulu-11.jdk,/Users/xchen21/Documents/JDK/17/zulu-17.jdk,/Users/xchen21/Documents/JDK/21/zulu-21.jdk
```
timtebeek, line 5: Let's revert this change :)

rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
```diff
@@ -0,0 +1,109 @@
...
+    @Option(displayName = "Superclasses or interfaces",
+        description = "The FQNs of the superclasses or interfaces to match class to add the comment to. ",
+        example = "com.netflix.hystrix.HystrixCommand")
+    List<String> implementations;
+
+    @Option(displayName = "Multiline",
+        description = "Comments use by default single line // but they can use multiline /* */.",
+        required = false)
+    @Nullable
+    Boolean isMultiline;
```
sambsnyd, line 70: What about JavaDoc comments?

rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Option(displayName = "Comment",
+        description = "The comment to add.",
+        example = "This is a comment.")
+    String comment;
+
+    @Option(displayName = "Superclasses or interfaces",
+        description = "The FQNs of the superclasses or interfaces to match class to add the comment to. ",
+        example = "com.netflix.hystrix.HystrixCommand")
+    List<String> implementations;
```
sambsnyd, line 64: There are many different criteria one might wish to use to target comments. Path to the file, type being in use, something on the classpath, only for test sources, etc. 

The recipe could try to support parameters for all of these or it could leave it up to `Preconditions`. For this specific case `org.openrewrite.java.search.FindImplementations` can be used.

gradle.properties
sambsnyd: @SiBorea you can add this line to your personal ~/.gradle/gradle.properties file to have it in effect for all your gradle builds without having to edit any of them directly

rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
MBoegers: To be fair, `AddCommentToMethod` does not support JavaDoc ;) But I 100% get your motivation.

rewrite-java/src/main/java/org/openrewrite/java/AddCommentToClass.java
MBoegers: The same holds for `MethodDeclarations`. I find this approach very appealing, as it enables us to use the recipes' visitor within other imperative recipes if needed. Adding a comment is an atomic building block.
@sambsnyd lets follow this approach and mark the `AddCommentToMethod#methodPattern` parameter as optional to enable the same for Methods.

MBoegers: Cool idea! But it's not that simple when inner types are involved. Concidder:
```java
import java.util.HashMap;
                  
public class Example extends HashMap implements Runnable {
    @Override
    public void run() { }
    public class Inner { }
}
```

With a recipe like:
```markdown
name: org.openrewrite.java.AddCommentToClassTest
displayName: Add comment
description: Add a comment with a precondition.
preconditions:
    - org.openrewrite.java.search.FindImplementations:
       typeName: java.util.HashMap
recipeList:
    - org.openrewrite.java.AddCommentToClass:
      comment: " Short comment to add"
      isMultiline: false
```

The result is unexpeted as the precondition is evaluated on CompilationUnit level.
```diff
    public void run() { }
-    public class Inner { }
+    // Short comment to add
+   public class Inner { }
 }
```

MBoegers: Adding a `superclass` and `className` to the existing solution and checking the implements, extends and type of the class declaration would be a good fit here. Plus, migrating from a String to a `TypeMatcher` is beneficial too.

MBoegers (changes requested):
Hi @SiBorea I have to request a few changes.

1. Have what if I want to match the Class by its FQN? For instance, add a Comment to all classes within a package or with a Suffix?
2. Have you thought about the handling of inner classes? They should not get the same comment if their containing class is matched.

To overcome these I suggest

- Use [`TypeMatcher`](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/TypeMatcher.java) instead of a String comparesion to check if it is the desired type
- Add `superClass` and `className` as options to enable advanced matching capabilities
- seems a little duplicated but opt out of a `visitClassDeclaration` if not all type options are meet (like preconditions).

PR #5303, opened 2025-04-15
Validate newMethodName in ChangeMethodName

JohannisK:
## What's changed?
Added validation to prevent changing a method name to
 - blank names
 - invalid patterns
 - reserved keywords

## What's your motivation?
Fixes:
 - https://github.com/openrewrite/rewrite-static-analysis/issues/426

Are the tests sufficient? I don't think they explicitly cover every Visitor, but they do show the behavior

We could leave the JavaKeyword class, but it's a bit weird that we check methodNames against a `VariableNameUtils`

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodNameTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodNameTest.java
@@ -422,4 +422,70 @@ class ChangeMethodNameTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void ignoreInvalidMethodNames() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeMethodName("com.abc.B static1(String)", "123", null, null)),
+          java(b, SourceSpec::skip),
+          java(
+            """
+              package com.abc;
+              
+              import java.util.ArrayList;
+              
+              class A {
+                 public void test() {
+                     com.abc.B.static1("boo");
+                     new java.util.ArrayList<String>().forEach(B::static1);
+                 }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void ignoreBlankMethodNames() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeMethodName("com.abc.B static1(String)", "\t", null, null)),
+          java(b, SourceSpec::skip),
+          java(
+            """
+              package com.abc;
+              
+              import java.util.ArrayList;
+              
+              class A {
+                 public void test() {
+                     com.abc.B.static1("boo");
+                     new java.util.ArrayList<String>().forEach(B::static1);
+                 }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void ignoreReservedKeyWordMethodNames() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeMethodName("com.abc.B static1(String)", "this", null, null)),
+          java(b, SourceSpec::skip),
+          java(
+            """
+              package com.abc;
+              
+              import java.util.ArrayList;
+              
+              class A {
+                 public void test() {
+                     com.abc.B.static1("boo");
+                     new java.util.ArrayList<String>().forEach(B::static1);
+                 }
+              }
+              """
+          )
+        );
+    }
 }
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
@@ -19,6 +19,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.java.search.DeclaresMethod;
 import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.*;
@@ -26,6 +27,9 @@ import org.openrewrite.java.tree.*;
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class ChangeMethodName extends Recipe {
+
+    //A valid java method starts with a letter, an underscore or a $
+    public static final String VALID_JAVA_METHOD_PATTERN = "^[a-zA-Z_$][a-zA-Z0-9_$]*$";
 
     @Option(displayName = "Method pattern",
             description = MethodMatcher.METHOD_PATTERN_DESCRIPTION,
@@ -100,7 +104,9 @@ public class ChangeMethodName extends Recipe {
             @Override
             public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
                 J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
-
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return m;
+                }
                 J.NewClass newClass = getCursor().firstEnclosing(J.NewClass.class);
                 J.ClassDeclaration classDecl = getCursor().firstEnclosing(J.ClassDeclaration.class);
                 boolean methodMatches = newClass != null && methodMatcher.matches(method, newClass) ||
@@ -119,6 +125,9 @@ public class ChangeMethodName extends Recipe {
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return m;
+                }
                 if (methodMatcher.matches(method) && !method.getSimpleName().equals(newMethodName)) {
                     JavaType.Method type = m.getMethodType();
                     if (type != null) {
@@ -133,6 +142,9 @@ public class ChangeMethodName extends Recipe {
             @Override
             public J.MemberReference visitMemberReference(J.MemberReference memberRef, ExecutionContext ctx) {
                 J.MemberReference m = super.visitMemberReference(memberRef, ctx);
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return m;
+                }
                 if (methodMatcher.matches(m.getMethodType()) && !m.getReference().getSimpleName().equals(newMethodName)) {
                     JavaType.Method type = m.getMethodType();
                     if (type != null) {
@@ -153,6 +165,9 @@ public class ChangeMethodName extends Recipe {
             @Override
             public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, ExecutionContext ctx) {
                 J.FieldAccess f = super.visitFieldAccess(fieldAccess, ctx);
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return f;
+                }
                 if (getCursor().getParentTreeCursor().getValue() instanceof J.Import && methodMatcher.isFullyQualifiedClassReference(f)) {
                     Expression target = f.getTarget();
                     if (target instanceof J.FieldAccess) {
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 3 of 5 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
```diff
@@ -27,6 +28,9 @@
 @EqualsAndHashCode(callSuper = false)
 public class ChangeMethodName extends Recipe {
 
+    //A valid java method starts with a letter, an underscore or a $
+    public static final String VALID_JAVA_METHOD_PATTERN = "^[a-zA-Z_$][a-zA-Z0-9_$]*$";
```
timtebeek, line 32: Did you know these special character classes?
https://github.com/openrewrite/rewrite/blob/72d79f96e281b4554f65f22ff6eb85cf8464d732/rewrite-properties/src/main/java/org/openrewrite/properties/trait/PropertiesReference.java#L63-L66

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
```diff
@@ -100,7 +104,9 @@ public J visit(@Nullable Tree tree, ExecutionContext ctx) {
             @Override
             public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
                 J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
-
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return m;
+                }
```
greg-at-moderne, line 109: Shouldn't the validation be done in `validate()`?

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
```diff
@@ -27,6 +28,9 @@
 @EqualsAndHashCode(callSuper = false)
 public class ChangeMethodName extends Recipe {
 
+    //A valid java method starts with a letter, an underscore or a $
```
jevanlingen, line 31:
```suggestion
    // A valid java method starts with a letter, an underscore or a $
```

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
jevanlingen: Don't know if that's clearer here to be honest. Also notice regex can be written shorter:

```suggestion
    public static final String VALID_JAVA_METHOD_PATTERN = "^[a-zA-Z_$][\\w$]*$";
```

JohannisK: I asked Tim, because I didn't know the special Character classes. He mentioned we use this in several places, so I opted for that solution.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
JohannisK: I removed the comment in favor of the special Character representation.

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+package org.openrewrite.java;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+final class JavaKeywordUtils {
+    JavaKeywordUtils() {
+    }
```
jevanlingen, line 24:
```suggestion
@UtilityClass
final class JavaKeywordUtils {
```

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static final String[] RESERVED_WORDS = new String[]{
```
jevanlingen, line 26: What about the [reserved identifiers](https://en.wikipedia.org/wiki/List_of_Java_keywords#Reserved_Identifiers)? I don't know if you should expand this class or create a new one, but for this use case of the PR they can't be missing!

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+final class JavaKeywordUtils {
```
timtebeek, line 22:
```suggestion
// Intentionally package private, at least for now, to limit the publicly exposed API surface
final class JavaKeywordUtils {
```

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
Pankraz76, line 26:
```suggestion
    private static final Set<String> RESERVED_KEYWORDS = Set.of(
```

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+          "try",
+          "void",
+          "volatile",
+          "while",
+    };
+
+    private static final Set<String> RESERVED_WORDS_SET = new HashSet<>(Arrays.asList(RESERVED_WORDS));
+
+    public static boolean isReserved(String word) {
+        return RESERVED_WORDS_SET.contains(word);
```
Pankraz76, line 82:
```suggestion
        return RESERVED_KEYWORDS.contains(word);
```

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+          "throws",
+          "transient",
+          "try",
+          "void",
+          "volatile",
+          "while",
+    };
+
+    private static final Set<String> RESERVED_WORDS_SET = new HashSet<>(Arrays.asList(RESERVED_WORDS));
+
```
Pankraz76, line 80:
```suggestion
```

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
```diff
@@ -0,0 +1,84 @@
...
+          "synchronized",
+          "this",
+          "throw",
+          "throws",
+          "transient",
+          "try",
+          "void",
+          "volatile",
+          "while",
+    };
```
Pankraz76, line 77:
```suggestion
    );
```

rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java
```diff
@@ -293,68 +293,4 @@ private boolean isValidImportName(@Nullable JavaType targetType, String name) {
...
-                "throw",
-                "throws",
-                "transient",
-                "try",
-                "void",
-                "volatile",
-                "while",
-        };
-
-        private static final Set<String> RESERVED_WORDS_SET = new HashSet<>(Arrays.asList(RESERVED_WORDS));
```
Pankraz76, line 354: No need for this extra coupling.
Why not inline this and avoid Hungarian notation pattern (covering type in name as well) ?

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
Pankraz76, line 31:
```suggestion
    /** A valid java method starts with a letter, an underscore or a $ **/
```
This is valid information valuable enough to include in the Java-Doc. While having it in the source code is helpful, that alone may not cover all needs.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
```diff
@@ -153,6 +165,9 @@ public J.MemberReference visitMemberReference(J.MemberReference memberRef, Execu
             @Override
             public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, ExecutionContext ctx) {
                 J.FieldAccess f = super.visitFieldAccess(fieldAccess, ctx);
+                if (JavaKeywordUtils.isReserved(newMethodName) || StringUtils.isBlank(newMethodName) || !newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return f;
+                }
```
Pankraz76, line 170: Remove the unnecessary DRY abstraction and use the early `return` already present at line 180.

Rather than introducing a separate `if` block, simply invert and inline the following condition into the existing if  check following in line 171 — it should behave the same while improving cohesion:

`JavaKeywordUtils.isReserved(newMethodName) && StringUtils.isBlank(newMethodName)`

This keeps the logic focused, reduces coupling, and avoids cluttering the control flow. It’s just about refining the condition, not restructuring the logic.

```suggestion

```

Pankraz76 (commented):
nicely done thanks. kindly suggest some polish.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
JohannisK: I'm trying to reproduce the recipe;s behavior when I add this to the validate() method.
I do think this changes the behavior of this very abstract core recipe.
It used to change methods into not compiling java code.
My approach is to silently ignore changes that are breaking java (e.g. `void this_()` should not be changed to `void this()`.
I'm a bit hesitant because if I understand correctly, changing validate would cause the recipe to start throwing errors.

rewrite-java/src/main/java/org/openrewrite/java/JavaKeywordUtils.java
JohannisK: I should include the reserved literal values , true, false and null!

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
JohannisK: For now I expanded the condition.

JohannisK: Moved to isAcceptable

JohannisK:
I've moved the code to check for the valid method name to isAcceptable().
This works for the recipe itself, but it's used in other places with a direct call to the visit method bypassing the isAcceptable method (at least in the recipe where the bug was reported).
(e.g. `nameChange.getRecipe().getVisitor().visitNonNull(cu, ctx)`)

I'll discuss how to fix this tomorrow, just posting this here to prevent anybody from merging.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java
```diff
@@ -70,9 +73,22 @@ public Validated<Object> validate() {
...
         JavaIsoVisitor<ExecutionContext> condition = new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext executionContext) {
+                if (!JavaKeywordUtils.isReservedKeyword(newMethodName) &&
+                      !JavaKeywordUtils.isReservedLiteral(newMethodName) &&
+                      !StringUtils.isBlank(newMethodName) &&
+                      newMethodName.matches(VALID_JAVA_METHOD_PATTERN)) {
+                    return super.isAcceptable(sourceFile, executionContext);
+                }
```
timtebeek, line 88: Should this not go into the `validate` method just above?

JohannisK: Added it to validate and isAcceptable() to support direct invocations and downstream invocations.

jevanlingen (approved):

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodNameTest.java
```diff
@@ -422,4 +428,70 @@ public void test() {
...
+    void ignoreInvalidMethodNamesWhenCalledDownstream(String newMethodName) {
+        AdHocRecipe downstreamRecipe = new AdHocRecipe("DownStream", "downstream", false, () -> new JavaIsoVisitor<>() {
+            @Override
+            public J visit(Tree tree, ExecutionContext executionContext) {
+                return (J) new ChangeMethodName("com.abc.B singleArg(String)", newMethodName, null, null).getVisitor().visitNonNull(tree, executionContext);
+            }
+        }, List.of(), 1);
+
+        rewriteRun(
+          spec -> spec.recipe(downstreamRecipe),
```
timtebeek, line 443: This could be shorter by using `toRecipe` as a static import from `RewriteTest`.
```suggestion
        rewriteRun(
          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<ExecutionContext>() {
              @Override
              public J visit(Tree tree, ExecutionContext executionContext) {
                  return (J) new ChangeMethodName("com.abc.B static1(String)", newMethodName, null, null).getVisitor().visitNonNull(tree, executionContext);
              }
          })),
```

Also; your using `singleArg` in the matcher, but the text block below uses `static1`, so even with a valid name your variant made no match or change. That's also fixed with the above.

timtebeek (approved):
Approved provided we apply that last suggestion. Merge after the tests pass.

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodNameTest.java
JohannisK: The toRecipe did not behave the same as the explicit adhoc recipe

JohannisK: Thnx! the first version of toRecipe did not show the same behavior as the adhoc recipe, but this one does :)

Issue #5300, opened 2025-04-13
[Gradle] Support updating dependencies in the `dependencyManagement` block


ThomasVitale:
## What problem are you trying to solve?

When running the `org.openrewrite.java.dependencies.UpgradeDependencyVersion` recipe, BOM dependencies defined in the `dependencyManagement` block are updated in a Maven project, but not in a Gradle project.

## Describe the solution you'd like

I would like to get support in Gradle for updating BOM dependencies, similar to Maven.

## Have you considered any alternatives or workarounds?

Gradle projects defining BOM dependencies in the dependencies block as "platform dependencies" are correctly updated. Dropping the `dependencyManagement` block altogether is definitely an alternative. However, it's not always a good one since "platform dependencies" are configuration specific. In a typical Spring Boot application, it means duplicating the BOM dependencies for each configuration you need (e.g. `implementation`, `runtimeOnly`, `developmentOnly`...).

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes, I'm available to contribute this feature to OpenRewrite.

ThomasVitale:
I added a couple of tests to show the issue: https://github.com/openrewrite/rewrite/pull/5304/files

ThomasVitale:
A PR has been submitted to address this issue by @Jenson3210 in https://github.com/openrewrite/rewrite/pull/5623. It's pending review. More context in https://github.com/openrewrite/rewrite/pull/5304#issuecomment-2971954971

timtebeek:
Apologies! The bot didn't quite behave as we had wanted to; it wasn't support to add tags, only evaluate already tagged questions.

Hope all is well there, and thanks for the help!

sambsnyd:
I just merged that linked PR, hopefully that helps

timtebeek:
Picking up the downstream option in
- https://github.com/openrewrite/rewrite-java-dependencies/pull/145

timtebeek:
Thanks all! Hope that has you going. :)

ThomasVitale:
Thanks so much!

PR #5304, opened 2025-04-15
Gradle - Add tests for update in dependencyManagement

ThomasVitale:
## What's changed?

- Introduced tests for demonstrating the issue.

## What's your motivation?
https://github.com/openrewrite/rewrite/issues/5300

- How to implement the new use case?

@timtebeek 

Yes, workaround considered and explained in https://github.com/openrewrite/rewrite/issues/5300, but not feasible in certain scenarios.

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
@@ -397,8 +397,42 @@ class UpgradeDependencyVersionTest implements RewriteTest {
 
     @Issue("https://github.com/openrewrite/rewrite/issues/5300")
     @Test
+    void doesNotMakeChangesInDependencyManagement() {
+        rewriteRun(
+          spec -> spec.recipe(new UpgradeDependencyVersion("org.openrewrite", "*", "8.44.1", null)),
+          buildGradle(
+            """
+              plugins {
+                  id 'java'
+                  id 'io.spring.dependency-management' version '1.1.7'
+              }
+
+              repositories {
+                  mavenCentral()
+              }
+
+              ext {
+                  openRewriteVersion = "8.44.1"
+              }
+              
+              dependencies {
+                implementation platform("com.google.guava:guava:29.0-jre")
+              }
+
+              dependencyManagement {
+                  imports {
+                      mavenBom "org.openrewrite:rewrite-core:${openRewriteVersion}"
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/5300")
+    @Test
     @Ignore
-    void worksWithDependencyManagement() {
+    void makeChangesInDependencyManagement() {
         rewriteRun(
           spec -> spec.recipe(new UpgradeDependencyVersion("org.openrewrite", "*", "8.44.1", null)),
           buildGradle(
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -394,6 +395,99 @@ implementation platform("com.google.guava:guava:30.1.1-jre")
...
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/5300")
+    @Test
+    @Ignore
```
ThomasVitale, line 434: Ignored since this is currently failing:

```log
Recipe was expected to make a change but made no changes.
	at org.openrewrite.test.LargeSourceSetCheckingExpectedCycles.afterCycle(LargeSourceSetCheckingExpectedCycles.java:119)
	at org.openrewrite.Recipe.run(Recipe.java:438)
	at org.openrewrite.gradle.UpgradeDependencyVersionTest.makeChangesInDependencyManagement(UpgradeDependencyVersionTest.java:436)
```

timtebeek:
Thanks for the start here @ThomasVitale ! I'm going to ask if @shanman190 has some guiding hints here for whether we want to make these changes as part of this recipe, and if that should be in a separate visitor perhaps. He's our friendly community authority on all things Gradle. 🧙🏻‍♂️

ThomasVitale:
Perfect, thank you so much!

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -394,6 +395,99 @@ implementation platform("com.google.guava:guava:30.1.1-jre")
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/5300")
+    @Test
+    void doesNotMakeChangesInDependencyManagement() {
```
shanman190, line 400: This test feels a bit extraneous to me since it's effectively never going to do anything from a recipe standpoint.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
shanman190: We should also probably duplicate this test with a Gradle Kotlin DSL variation as well.

shanman190 (commented):
I think it is probably best as Tim suggested to do this as a separate dedicated visitor given that this section is 1) not type attributed and 2) extremely specific to the Spring Dependency Management plugin.

As you'll see with the other visitors, I would suggest utilizing the `JavaIsoVisitor`, then also performing similar updates to handle both Gradle Groovy DSL and Gradle Kotlin DSL both.

ThomasVitale:
@shanman190 thanks so much for the inputs. I'll look a bit deeper into the `JavaIsoVisitor` and consider support for both the Groovy DSL and Kotlin DSL. I hope I'll get a draft PR ready soon.

Jenson3210:
I have pushed a draft PR that addresses this one. The implementation is still quite dirty, but I feel this Trait could easily be added to other recipes like UpgradeDependencyVersion etc to make them also bump the depdencyManagement section.

PR #5307, opened 2025-04-16
Unwrap parentheses around method invocation arguments

timtebeek:
### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesVisitor.java
@@ -16,7 +16,6 @@
 package org.openrewrite.java.cleanup;
 
 import lombok.EqualsAndHashCode;
-import org.jetbrains.annotations.NotNull;
 import org.openrewrite.Cursor;
 import org.openrewrite.SourceFile;
 import org.openrewrite.internal.ListUtils;
@@ -236,7 +235,7 @@ public class UnnecessaryParenthesesVisitor<P> extends JavaVisitor<P> {
         }));
     }
 
-    private static @NotNull Space mergeSpace(Space outside, Space inside) {
+    private static Space mergeSpace(Space outside, Space inside) {
         return Space.build(
                 outside.getWhitespace() + inside.getWhitespace(),
                 ListUtils.concatAll(outside.getComments(), inside.getComments()));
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesVisitor.java
```diff
@@ -59,8 +60,16 @@ public <T extends J> J visitParentheses(J.Parentheses<T> parens, P ctx) {
...
+            J parentCursorValue = getCursor().getParentTreeCursor().getValue();
+            if (parentCursorValue instanceof J.Parentheses) {
                 return ((J.Parentheses<?>) par).getTree().withPrefix(Space.EMPTY);
+            } else if (parentCursorValue instanceof J.MethodInvocation) {
+                J tree = ((J.Parentheses<?>) par).getTree();
+                Space outside = par.getPrefix();
+                Space inside = tree.getPrefix();
+                return tree.withPrefix(Space.build(
+                        outside.getWhitespace() + inside.getWhitespace(),
+                        ListUtils.concatAll(outside.getComments(), inside.getComments())));
```
timtebeek, line 72: This seemed like something we ought to have a `merge` utility for in `Space`, as there could be comments or whitespace either side of the parentheses that we want to retain.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesVisitor.java
```diff
@@ -223,4 +216,27 @@ public J visitUnary(J.Unary unary, P ctx) {
         }
         return u;
     }
+
+    @Override
+    public J visitMethodInvocation(J.MethodInvocation method, P p) {
+        J.MethodInvocation mi = (J.MethodInvocation) super.visitMethodInvocation(method, p);
+        if (mi.getSelect() instanceof J.Parentheses &&
+                !(((J.Parentheses<?>) mi.getSelect()).getTree() instanceof J.Binary)) {
```
knutwannheden, line 224: I wonder if we also need to check for some other cases like J.Ternary and J.TypeCast.

timtebeek: Thanks! Addressed in https://github.com/openrewrite/rewrite/commit/cc4e2c063c56a25bab3118cede3373ca7b3a03d5

Issue #5211, opened 2025-03-24
Support JDK24


zepfred:
## What problem are you trying to solve?

I'm attempting to use JDK24 with the most recent rewrite version, but I'm encountering this error:

```
java.lang.AssertionError: Failed to parse sources or run recipe
	at org.openrewrite.test.RewriteTest.lambda$defaultExecutionContext$14(RewriteTest.java:638)
	at org.openrewrite.java.isolated.ReloadableJava21Parser.parseInputsToCompilerAst(ReloadableJava21Parser.java:244)
	at org.openrewrite.java.isolated.ReloadableJava21Parser.parseInputs(ReloadableJava21Parser.java:173)
	at org.openrewrite.java.Java21Parser.parseInputs(Java21Parser.java:39)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:280)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:129)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:124)
	at ai.timefold.solver.migration.v8.SingleConstraintAssertionRecipeTest.runTest(SingleConstraintAssertionRecipeTest.java:46)
	at ai.timefold.solver.migration.v8.SingleConstraintAssertionRecipeTest.constraintMethods(SingleConstraintAssertionRecipeTest.java:24)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.tryRemoveAndExec(ForkJoinPool.java:1434)
	at java.base/java.util.concurrent.ForkJoinPool.helpJoin(ForkJoinPool.java:2197)
	at java.base/java.util.concurrent.ForkJoinTask.awaitDone(ForkJoinTask.java:495)
	at java.base/java.util.concurrent.ForkJoinTask.join(ForkJoinTask.java:662)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1394)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1970)
[...]

PR #5309, opened 2025-04-16
feat: support JDK24

zepfred:
## What's changed?
The PR updates the module `rewrite-java-21` to be compatible with JDK 24. The main change is eliminating the use of  `UnknownType`, as it has been removed from JDK 24.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5211

It seems more straightforward to modify the `rewrite-java-21` module than to create a separate module with the new classes. 

### Checklist

### DIFF
--- a/rewrite-java-21/build.gradle.kts
+++ b/rewrite-java-21/build.gradle.kts
@@ -24,6 +24,7 @@ dependencies {
     implementation("io.micrometer:micrometer-core:1.9.+")
     implementation("io.github.classgraph:classgraph:latest.release")
     implementation("org.ow2.asm:asm:latest.release")
+    implementation("org.openrewrite:rewrite-templating:latest.release")
 
     testImplementation(project(":rewrite-test"))
     "javaTck"(project(":rewrite-java-tck"))
--- a/rewrite-java-21/build.gradle.kts
+++ b/rewrite-java-21/build.gradle.kts
@@ -54,7 +54,10 @@ tasks.withType<Javadoc> {
             "**/ReloadableJava21Parser**",
             "**/ReloadableJava21ParserVisitor**",
             "**/ReloadableJava21TypeMapping**",
-            "**/ReloadableJava21TypeSignatureBuilder**"
+            "**/ReloadableJava21TypeSignatureBuilder**",
+            "**/Javac**",
+            "**/JavacTreeMaker**",
+            "**/Permit**"
     )
 }
 
--- a/rewrite-java-21/build.gradle.kts
+++ b/rewrite-java-21/build.gradle.kts
@@ -24,7 +24,6 @@ dependencies {
     implementation("io.micrometer:micrometer-core:1.9.+")
     implementation("io.github.classgraph:classgraph:latest.release")
     implementation("org.ow2.asm:asm:latest.release")
-    implementation("org.openrewrite:rewrite-templating:latest.release")
 
     testImplementation(project(":rewrite-test"))
     "javaTck"(project(":rewrite-java-tck"))
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -48,8 +48,8 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
     private final JavaTypeCache typeCache;
 
     public JavaType type(com.sun.tools.javac.code.@Nullable Type type) {
-        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || matchesUnknownType(type)
-                || type instanceof NullType){
+        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || matchesUnknownType(type) ||
+                type instanceof NullType) {
             return JavaType.Class.Unknown.getInstance();
         }
 
@@ -734,8 +734,8 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
         for (Pair<Symbol.MethodSymbol, Attribute> attr : compound.values) {
             Object value = annotationElementValue(attr.snd.getValue());
             JavaType.Method element = requireNonNull(methodDeclarationType(attr.fst, annotType));
-            JavaType.Annotation.ElementValue elementValue = (value instanceof Object[] arrayValue) ?
-                    JavaType.Annotation.ArrayElementValue.from(element, (arrayValue)) :
+            JavaType.Annotation.ElementValue elementValue = value instanceof Object[] ?
+                    JavaType.Annotation.ArrayElementValue.from(element, ((Object[]) value)) :
                     JavaType.Annotation.SingleElementValue.from(element, value);
             elementValues.add(elementValue);
         }
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -37,8 +37,6 @@ import java.util.stream.Collectors;
 
 import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
-import static org.openrewrite.java.template.internal.Javac.CTC_UNKNOWN;
-import static org.openrewrite.java.template.internal.JavacTreeMaker.TypeTag.typeTag;
 import static org.openrewrite.java.tree.JavaType.GenericTypeVariable.Variance.*;
 
 @RequiredArgsConstructor
@@ -49,8 +47,8 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
     private final JavaTypeCache typeCache;
 
     public JavaType type(com.sun.tools.javac.code.@Nullable Type type) {
-        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || CTC_UNKNOWN.equals(typeTag(type))
-                || type instanceof NullType) {
+        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || type instanceof Type.UnknownType ||
+                type instanceof NullType) {
             return JavaType.Class.Unknown.getInstance();
         }
 
@@ -491,8 +489,7 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
             }
         }
 
-        if (selectType == null || selectType instanceof Type.ErrorType || symbol == null || symbol.kind == Kinds.Kind.ERR
-                || CTC_UNKNOWN.equals(typeTag(symbol.type))) {
+        if (selectType == null || selectType instanceof Type.ErrorType || symbol == null || symbol.kind == Kinds.Kind.ERR || symbol.type instanceof Type.UnknownType) {
             return null;
         }
 
@@ -557,7 +554,7 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
                     exceptionTypes.add(javaType);
                 }
             }
-        } else if (CTC_UNKNOWN.equals(typeTag(selectType))) {
+        } else if (selectType instanceof Type.UnknownType) {
             returnType = JavaType.Unknown.getInstance();
         }
 
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -481,8 +481,7 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
                     if (targetClass.getSimpleName().equals("RecoveryErrorType")) {
                         Field field = targetClass.getDeclaredField("candidateSymbol");
                         field.setAccessible(true);
-                        Symbol originalSymbol = (Symbol) field.get(selectType);
-                        return methodInvocationType(selectType.getOriginalType(), originalSymbol);
+                        return methodInvocationType(selectType.getOriginalType(), (Symbol) field.get(selectType));
                     }
                 }
             } catch (Exception e) {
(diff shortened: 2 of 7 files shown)


rewrite-java-21/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation("io.micrometer:micrometer-core:1.9.+")
     implementation("io.github.classgraph:classgraph:latest.release")
     implementation("org.ow2.asm:asm:latest.release")
+    implementation("org.openrewrite:rewrite-templating:latest.release")
```
timtebeek, line 27: Rather than add a dependency here I think it's likely best to add copies of the field and method used here in the rewrite-java-21 parser itself. This dependency contains an annotation processor that would be somewhat odd to pull in here.

timtebeek (commented):
Great to see this kicked off @zepfred ; I know a lot of folks will appreciate this effort.

A quick remark already; I think it should be easy to adjust and work from there.

timtebeek:
I had briefly looked at adding the following here; could be good to verify that it all still works from Java 24:
```diff
diff --git i/.github/workflows/ci.yml w/.github/workflows/ci.yml
index de127871a..8d5f33fb0 100644
--- i/.github/workflows/ci.yml
+++ w/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ concurrency:
 jobs:
   build:
     uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
+    with:
+      java_version: '24'
     secrets:
       gradle_enterprise_access_key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       ossrh_username: ${{ secrets.OSSRH_USERNAME }}
```

rewrite-java-21/build.gradle.kts
zepfred: Good point. I'll change the approach.

zepfred:
> I had briefly looked at adding the following here; could be good to verify that it all still works from Java 24:
> 
> ```diff
> diff --git i/.github/workflows/ci.yml w/.github/workflows/ci.yml
> index de127871a..8d5f33fb0 100644
> --- i/.github/workflows/ci.yml
> +++ w/.github/workflows/ci.yml
> @@ -19,6 +19,8 @@ concurrency:
>  jobs:
>    build:
>      uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
> +    with:
> +      java_version: '24'
>      secrets:
>        gradle_enterprise_access_key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
>        ossrh_username: ${{ secrets.OSSRH_USERNAME }}
> ```

Does it make sense to run the job using different JDK versions (17, 21, 24) instead of just the latest one?

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/Javac.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (C) 2009-2019 The Project Lombok Authors.
```
timtebeek, line 2: This looked slightly out of place at first, but I think this is a heavily reduced copy of https://github.com/projectlombok/lombok/blob/v1.18.38/src/utils/lombok/javac/Javac.java

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/JavacTreeMaker.java
```diff
@@ -0,0 +1,211 @@
...
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
```
timtebeek, line 17: I think we'll likely only want to apply a single header here.
```suggestion
/*
```

timtebeek:
> Does it make sense to run the job using different JDK versions (17, 21, 24) instead of just the latest one?

Thanks for pointing this out. Typically we've been able to get away with just the latest version, and then setting the target version to Java 8. I vaguely recall that beyond Java 21 that target on 8 might not be possible anymore, but we'll see soon enough from the CI output if that's the case. It does help reduce the CI usage if we can get away with testing on a single version.

timtebeek:
Copying the failure here:
```
* Where:
Build file '/home/runner/work/rewrite/rewrite/rewrite-java-test/build.gradle.kts' line: 45

* What went wrong:
Could not create task ':rewrite-java-test:test'.
> Could not create task of type 'Test'.
   > Could not create an instance of type org.gradle.api.internal.tasks.testing.DefaultTestTaskReports.
      > Could not create an instance of type org.gradle.api.reporting.internal.DefaultReportContainer.
         > Type T not present
```

Which is odd, since there is no line 45 in `main`. 🤔 
https://github.com/openrewrite/rewrite/blob/a4a70fdf3b39c96088b28589385d2718582f658e/rewrite-java-test/build.gradle.kts#L37-L42

timtebeek:
Reading the Gradle compatibility notes as of today: https://docs.gradle.org/8.13/userguide/compatibility.html
> Gradle runs on the Java Virtual Machine (JVM), which is often provided by either a JDK or JRE. A JVM version between 8 and 23 is required to execute Gradle. JVM 24 and later versions are not yet supported.

timtebeek:
We'll retry building on Java 24 once Gradle supports that; we'll update that separately, but no need to hold up this work over it: we can build on 21, and separately manually verify we can then support Java 24 before we merge this PR.

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/Javac.java
zepfred: I copied it from the project `rewrite-templating`, but you're probably right about it being reduced copy from the Lombok class.

zepfred:
> We'll retry building on Java 24 once Gradle supports that; we'll update that separately, but no need to hold up this work over it: we can build on 21, and separately manually verify we can then support Java 24 before we merge this PR.

Sounds good.

timtebeek:
Ok the steps I've taken to verify this so far:
1. Check out this branch & run `./gradlew pTML`
2. Check out rewrite-gradle-pluign and run `./gradlew pTML`
3. Check out https://github.com/tginsberg/gatherers4j and add `init.gradle` there
```gradle
initscript {
    repositories {
        mavenLocal()
        maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots/") }
        maven { url "https://plugins.gradle.org/m2" }
    }
    dependencies { classpath("org.openrewrite:plugin:latest.integration") }
}
rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite.recipe:rewrite-rewrite:0.4.2")
    }
    rewrite {
        activeRecipe("org.openrewrite.java.OrderImports")
        setExportDatatables(true)
    }
    repositories {
        mavenLocal()
        maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots/") }
        mavenCentral()
    }
}
```
4. Butcher the imports on one of the files, as we're running `OrderImports` above
5. Run `./gradlew --init-script init.gradle rewriteRun --info` & see output similar to
```
There were problems parsing some source files, run with --info to see full stack traces
There were problems parsing src/main/java/com/ginsberg/gatherers4j/ZipWithGatherer.java
There were problems parsing src/test/java/com/ginsberg/gatherers4j/ThrottlingGathererTest.java
All sources parsed, running active recipes: org.openrewrite.java.OrderImports
Printing available datatables to: /home/tim/Documents/workspace/gatherers4j/build/reports/rewrite/datatables/2025-04-18_14-12-29-761
Changes have been made to src/main/java/com/ginsberg/gatherers4j/InterleavingGatherer.java by:
    org.openrewrite.java.OrderImports
Please review and commit the results.
Estimate time saved: 5m
```

Repeating the above steps with the `main` branch of openrewrite/rewrite shows
```
* What went wrong:
Execution failed for task ':rewriteRun'.
> java.lang.IllegalStateException: Unable to construct Java21Parser.
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
```diff
@@ -47,8 +48,8 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
     private final JavaTypeCache typeCache;
 
     public JavaType type(com.sun.tools.javac.code.@Nullable Type type) {
-        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || type instanceof Type.UnknownType ||
-                type instanceof NullType) {
+        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || matchesUnknownType(type)
+                || type instanceof NullType){
```
jevanlingen, line 52:
```suggestion
        if (type == null || type instanceof Type.ErrorType || type instanceof Type.PackageType || matchesUnknownType(type) ||
                type instanceof NullType) {
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
```diff
@@ -732,8 +734,8 @@ private void completeClassSymbol(Symbol.ClassSymbol classSymbol) {
         for (Pair<Symbol.MethodSymbol, Attribute> attr : compound.values) {
             Object value = annotationElementValue(attr.snd.getValue());
             JavaType.Method element = requireNonNull(methodDeclarationType(attr.fst, annotType));
-            JavaType.Annotation.ElementValue elementValue = value instanceof Object[] ?
-                    JavaType.Annotation.ArrayElementValue.from(element, ((Object[]) value)) :
+            JavaType.Annotation.ElementValue elementValue = (value instanceof Object[] arrayValue) ?
+                    JavaType.Annotation.ArrayElementValue.from(element, (arrayValue)) :
```
jevanlingen, line 738: We try to keep our java 8, java 11, java 17 and 21 parsers the same as possible. So revert this change please.

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/Javac.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.openrewrite.java.isolated.internal.JavacTreeMaker.TypeTag;
+
+import static org.openrewrite.java.isolated.internal.JavacTreeMaker.TypeTag.typeTag;
+import static org.openrewrite.java.isolated.internal.JavacTreeMaker.TypeTag.typeTagPermissive;
+
+public final class Javac {
+    private Javac() {
+        // prevent instantiation
+    }
```
jevanlingen, line 33:
```suggestion
import lombok.experimental.UtilityClass;

@UtilityClass
public final class Javac {
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/JavacTreeMaker.java
```diff
@@ -0,0 +1,195 @@
...
+
+import static org.openrewrite.java.isolated.internal.Javac.CTC_UNKNOWN;
+import static org.openrewrite.java.isolated.internal.Javac.CTC_VOID;
+import static org.openrewrite.java.isolated.internal.Javac.sneakyThrow;
+
+@SuppressWarnings("SameParameterValue")
+public class JavacTreeMaker {
+
+    private JavacTreeMaker() {
+    }
```
jevanlingen, line 44:
```suggestion
import lombok.experimental.UtilityClass;

@UtilityClass
public class JavacTreeMaker {
```

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/JavacTreeMaker.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private static class SchroedingerType {
```
jevanlingen, line 46: Could also be lombokified.

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/internal/Javac.java
timtebeek: So these files are copied from Lombok; I wouldn't spend too much time maintaining these, as they are likely to be dropped when we do get around to rewrite-java-25. Until then they can stay as is. See also:
- https://github.com/openrewrite/rewrite/pull/5309/files/d39c4f226cbba2243666831712b76a60899929a3#r2047439216

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
timtebeek: To add a bit of context: we're unfortunately are having to maintain currently 4, but soon 5 implementations in parallel, each using the particular version compiler internals. A task we do often is fix something in one copy, and then have to replicate that to all other copies, and then it helps us a lot if there's only intentional changes related to compiler internals, for an easier (smaller) diff. As much as we _could_ leverage newer language features in some of these parsers, we've held off for these reasons. Thanks for taking that into account!

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
timtebeek: This is indeed a convention we maintain on the project; normally there's a bot to point these out on the other projects.

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
timtebeek:
```suggestion
            JavaType.Annotation.ElementValue elementValue = value instanceof Object[] ?
                    JavaType.Annotation.ArrayElementValue.from(element, ((Object[]) value)) :
```

sambsnyd:
Nice work @zepfred !

zepfred:
Thank you!

PR #5310, opened 2025-04-16
Semver not to choke on unusual version strings

greg-at-moderne:
## What's changed?

A safety guard in `Semver`'s logic to compare two versions. Basically if one of them doesn't follow the usual pattern, then the pick the other one. In case of both of them being non-standard, just pick one of them deterministically (then it's hard to decide what is max).

## What's your motivation?

Address issue reported by users seeing exceptions being thrown, e.g.
`Error comparing version number SOMETHING-2003.1.0_1 to 1.0.2 in DependencyVulnerabilityCheck`

## Anything specific to review?

Yeah. I tried to be defensive and pragmatic about the scenario of two versions being non-standard. I have a hunch this is closer to user expectations. But at the same time the code looks a bit off with not checking `isVersion` consistently in all branches. Happy to take better suggestions.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/Semver.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/Semver.java
@@ -28,7 +28,7 @@ public class Semver {
 
     @SuppressWarnings("BooleanMethodIsAlwaysInverted")
     public static boolean isVersion(@Nullable String version) {
-        if (version == null) {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Made a small adjustment to avoid duplication; didn't see harm there as the RELEASE_PATTERN we use already expects it to start with a number.

PR #5311, opened 2025-04-17
Fix: Ensure onComplete is called for DelegatingRecipe

amishra-u:
## What's changed?
Ensure onComplete method is invoked for DelegatingRecipe.

## What's your motivation?
Previously, the onComplete method was not called when a DeclarativeRecipe had a precondition, as the child recipes were wrapped in a DelegatingRecipe.

@knutwannheden @timtebeek 

### Checklist


timtebeek (approved):
Looks alright to me. Thanks for noticing and the immediate fix!

PR #5312, opened 2025-04-17
Improve UnfoldProperties recipe

jevanlingen:
## What's changed?
The unfold recipe:
- https://github.com/openrewrite/rewrite/pull/5245

 got a little improvement. It now support regexes, parent level pointers and merges same hierarchal keys together.

### Checklist


rewrite-yaml/src/test/java/org/openrewrite/yaml/UnfoldPropertiesTest.java
```diff
@@ -138,4 +138,97 @@ void exclusions() {
...
+    void mergeDuplicatedSections() {
+        rewriteRun(
+          yaml(
+            """
+              logging.level:
+                com.first.package: INFO
+              some.thing: else
+              logging.level.com.second.package: DEBUG
+              other: thing
+              logging.level.com.third.package: DEBUG
```
Jenson3210, line 215: I think the `MergeDuplicatedSectionsVisitor` will not work as expected when there are comments in the yaml, will it? 

Can't say that it does or does not, but I would expect a test that shows it.

```yaml
logging.level:
  com.first.package: INFO
  some.thing: else # comment 1 here
logging.level.com.second.package: DEBUG # comment 2 here
other: thing
# other:
#   commented: true
logging.level.com.third.package: DEBUG
```

jevanlingen: You are right, the comment model is hard to get right 😢. Let me see what I can do.

Jenson3210: If you figure it out, let me know! 
I need similar thing for a Draft I could never fix before: https://github.com/openrewrite/rewrite/pull/5002

rewrite-yaml/src/main/java/org/openrewrite/yaml/UnfoldProperties.java
```diff
@@ -42,7 +39,9 @@ public class UnfoldProperties extends Recipe {
     private static final Pattern LINE_BREAK = Pattern.compile("\\R");
 
     @Option(displayName = "Exclusions",
-            description = "The keys which you do not want to unfold",
+            description = "Specifies keys that should not be unfolded. Supports regex expressions. " +
+                    "A special format `[<parent-regex>]->[<child-regex>]` can be used to exclude specific child keys under a parent. " +
+                    "For example, `[logging.level]->.*` excludes all child keys under 'logging.level'.",
```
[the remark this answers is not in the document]
greg-at-moderne: Also is there any use case for the child regex? Or will it be always `.*` in practice?
If only `.*`, then we could simplify the option syntax keeping the `<parent-regex>` part only.

rewrite-yaml/src/test/java/org/openrewrite/yaml/UnfoldPropertiesTest.java
jevanlingen: Ignored comments by: https://github.com/openrewrite/rewrite/pull/5312/commits/90dfbe8277c1f199b870073ab55dff1888e2e1c3.

rewrite-yaml/src/main/java/org/openrewrite/yaml/UnfoldProperties.java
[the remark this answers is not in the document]
jevanlingen: > It's not clear to me whether [] is part of the special format or no.
> The top line suggests it's mandatory on both sides of the ->.
>The bottom line suggests it's not.

Ah, made a mistake here. It should have been: `[<parent-regex>]-><child-regex>`. The `[..]->` just indicate that we have a parent-child selector.

> Also is there any use case for the child regex? Or will it be always .* in practice?
We don't know and we don't care. Let the user use it is he/she wants.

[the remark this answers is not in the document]
jevanlingen: https://github.com/openrewrite/rewrite/pull/5312/commits/08c17fe9e0b8199e25d623b43919fe91b49c20d9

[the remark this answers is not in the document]
greg-at-moderne: > We don't know and we don't care. Let the user use it is he/she wants.

The benefit of not supporting this use-case is avoiding this strange syntax.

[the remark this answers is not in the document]
jevanlingen: > The benefit of not supporting this use-case is avoiding this strange syntax.

No. This syntax gives the user the option to target a mapping by its parent (which is exactly what a customer wants). Normally the exclusion regexes are compared to the lst elements you are visiting.  

Take this example:
```yaml
aaa:
  first: value:
zzz:
  first: value:
```

if you want to exclude the folding of only the first, you can't use the normal `first` (or `f.*t`, or whatever regex you can conjure up to match it), because now you are suddenly excluding all `first` elements. But by using `[aaa]->first` or `[a{3}]->.*`, or whatever regex you can think of, you can match it exactly like you want to.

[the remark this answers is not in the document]
greg-at-moderne: OK. So that's an example of a use-case for something else than `.*` in childRegex.

I continue to believe this syntax is error prone. Even looking at your comment above:
> [keep]-> this.example
There's a space after `>`, which I believe won't work. I am not trying to nitpick, but just flagging the syntax as something weird.
At the same time - I don't have a better suggestion (yet).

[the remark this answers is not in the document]
jevanlingen: Replaced custom syntax with JsonPath selectors 🎉.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeDuplicateSectionsVisitor.java
```diff
@@ -0,0 +1,90 @@
...
+            } else {
+                newEntries.put(key, e);
+            }
+        }
+
+        return changed ? m.withEntries(new ArrayList<>(newEntries.values())) : m;
+    }
+
+    // TODO We don't support comments yet, as supporting comments is a hard thing to get right
+    private static boolean hasComments(String prefix) {
```
Jenson3210, line 87: Can it not happen that i have > 1 newlines in the yaml which will then be seen as a comment

rewrite-yaml/src/main/java/org/openrewrite/yaml/UnfoldProperties.java
```diff
@@ -62,44 +63,150 @@ public String getDescription() {
...
+                    if (c.getValue() instanceof Yaml.Mapping.Entry) {
+                        parentKey.insert(0, ((Yaml.Mapping.Entry) c.getValue()).getKey().getValue() + ".");
+                    }
+                    c = c.getParent();
+                }
+                return parentKey.length() == 0 ? "" : parentKey.substring(0, parentKey.length() - 1);
+            }
+
+            // Custom JsonPathParser to parse keys with dots, like `logging.level`
+            private Optional<String> matchesWith(String ex, String key, String parentKey) {
```
Jenson3210, line 151: should we not use a name like `pattern`?

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeDuplicateSectionsVisitor.java
jevanlingen: https://github.com/openrewrite/rewrite/pull/5312/commits/5a1bb05c9b4297ce2828bf74d12533483a9e9363

rewrite-yaml/src/main/java/org/openrewrite/yaml/UnfoldProperties.java
jevanlingen: https://github.com/openrewrite/rewrite/pull/5312/commits/5a1bb05c9b4297ce2828bf74d12533483a9e9363

Jenson3210 (approved):
Great to see the test values. Makes it very clear. 
Personally, I would change the DocumentExample to another more representative example (like a Spring boot properties yaml file that gets unfolded, but the logging level gets kept etc...)

PR #5315, opened 2025-04-18
Improve performance of FindAnnotations precondition

jkschneider:
## What's changed?
Evaluate the precondition without navigating down the tree.

## What's your motivation?
`JUnitUpgrade` is on the top 3 of worst performers in DevCenter and it uses `FindAnnotations` under the covers:

image

I don't think so, it looks like the other part of that precondition was just a holdover from an earlier time.


timtebeek (approved):
Looks good. We'll keep an eye out of downstream builds, but I don't expect any issues.

Laurens-W (approved):

timtebeek:
Looks indeed like this was left over from single source applicability test we had previously; little sense in checking for the annotation twice once in precondition, and once more in the actual visitor. Merging with that in mind such that we can monitor today's downstream builds.

PR #5318, opened 2025-04-19
Feat/java add explicit imports

BelmoMusta:
### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddExplicitImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddExplicitImport.java
@@ -22,6 +22,11 @@ public class AddExplicitImport extends Recipe {
     public String getDisplayName() {
         return "Add explicit imports";
     }
+
+    @Override
+    public int maxCycles() {
+        return 1;
+    }
     @Override
     public String getDescription() {
         return "This recipe adds an explicit import to a single Java file.";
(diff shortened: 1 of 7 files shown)


timtebeek:
Hmm, we had intentionally not added a recipe for this, as adding an import by itself could lead to then having an unused import. Instead we handle imports from other recipes through a method called `maybeAddImport` that then ensures it's added only when referenced, and correctly grouped to match the project style. 

Could you describe the use case you see for having an explicit recipe? When would you need this in isolation, or how would you combine this? 

My worry is having a recipe like this might lead to confusion, and folks adopting this over self contained recipes that handle the imports for the elements they are adding.

BelmoMusta:
@timtebeek, I agree with you, but when it comes to the `imports style and organization` I don't think there is a standard style. I also believe that this kind of recipes will provide users with a certain autonomy when it comes to handling imports.

For example, I discourage the use of the `*` grouping way, as not every IDE is not having the same strategy of a such a grouping.

timtebeek:
For that wildcard option we have configurable styles: https://docs.openrewrite.org/concepts-and-explanations/styles
That ought to then apply those consistently when you run let's say `OrderImports`. I still feel that's preferred over recipes that explicitly add individual imports, however well-intentioned.

I have yet to see a convincing case before we merge and maintain this going forward, as I'm afraid folks (and AI assistants) will learn then wrong patterns when they see a recipe like this, and start adding explicit imports through this recipe.

timtebeek:
As much as the help is appreciated, given the concerns outlined in the comments above I propose we close this PR for now:
There are better ways to handle imports as opposed to directly force adding an import.

PR #5317, opened 2025-04-18
Add support for Groovy spread operator.

BrendanHart:
## What's changed?

Adding support for Groovy spread operator

## What's your motivation?.

- Closes #5297 

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #5321, opened 2025-04-20
Add support for "new array" in Groovy parser.

### LINKED ISSUE #5298: Groovy idempotency issue with parsing new array with fixed size

Current main = b504c17eb4055e959ac7d5edd5616c84664667f2

The following test:
```
    @Test
    void newArrayWithSize() {
        rewriteRun(groovy(
                """
        class A {
            static void main(String[] argv) {
                int[] addr = new int[4];
            }
        }
        """
        ));
    }
```

fails the idempotency check with:
```diff
- int[] addr = new int[4];
+ int[] addr = 4}ew int[4];
```
which is clearly invalid.

## Context

The same pattern has been found to cause problems in: https://github.com/spinnaker/spinnaker/blob/97cee4f3da768323984a4e1f404c96a9addf5306/clouddriver/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/common/AzureUtilities.groovy#L226

BrendanHart:
## What's changed?

- Adding support for mapping to `J.NewArray` in the Groovy parser.
- Handle array types in variable declarations.

## What's your motivation?

- Closes #5298 

That any padding/formatting is being handled correctly!

### Checklist


timtebeek:
Thanks for the help here @BrendanHart ! I've resolved the conflicts following your other PR
- https://github.com/openrewrite/rewrite/pull/5317

Was there any other reason you had kept this marked as draft?

BrendanHart:
Thanks @timtebeek!

Yes, I think the type declaration for assignment needs improving, I initially added `typeName = expression.getOriginType().toString();` as something to work through, but this doesn't work for types such as `List<String>[]`, and I think there's some issue with the type mapping to `JavaType` also, as at present in main, `List<String>[]` is mapped to `String[]`.

You could argue this is a separate issue, but since I need to make a change to handle the simple array cases like `int[]`, I was hoping to improve the naive `.toString()` I started with and solve it for both.

BrendanHart:
E.g.
```java
    @Test
    void arrayOfListOfStrings() {
        rewriteRun(groovy(
          """
            class TestMe{
              String[] getArgs() {
                List<String>[] r;
              }
            }
            """
        ));
    }
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -681,6 +681,58 @@ private Statement labeled(org.codehaus.groovy.ast.stmt.Statement statement, Supp
             return s;
         }
 
+        @Override
+        public void visitArrayExpression(ArrayExpression expression) {
+            Space whitespaceBeforeNew = whitespace();
```
jevanlingen, line 686:
```suggestion
            Space prefix = whitespace();
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -681,6 +681,58 @@ private Statement labeled(org.codehaus.groovy.ast.stmt.Statement statement, Supp
...
+            JContainer<Expression> initializer = buildNewArrayInitializer(expression);
+            queue.add(new J.NewArray(randomId(), whitespaceBeforeNew, Markers.EMPTY, typeTree, dimensions, initializer, typeMapping.type(expression.getElementType())));
+        }
+
+        private JContainer<Expression> buildNewArrayInitializer(ArrayExpression expression) {
+            if (expression.getSizeExpression() != null) {
+                return null;
+            }
+
+            Space whitespaceBefore = sourceBefore("{");
```
jevanlingen, line 699:
```suggestion
            Space fmt = sourceBefore("{");
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
BrendanHart: No problem, consistency is important! Thanks for the correction.

BrendanHart:
I think this works for `List<String>[]` now but doesn't yet handle multiple declarations as below would fail:
```
    @Test
    void multipleDeclarations() {
        rewriteRun(groovy(
          """
          int[] l, r;
          """
        ));
    }
```
however I think generally this needs more work due to 
- #5335 
will continue working on it as part of this PR but if it should be separate happy to mark this as ready and it can be a separate PR.

greg-at-moderne:
Thanks for the contributions again!

> it can be a separate PR.

IMHO it should be a separate one, but no strong feelings here. Just based on my experience, it will likely be easier to reason about these two PRs in isolation.

greg-at-moderne (approved):
Looks good!

I am gonna merge this one. And as mentioned above - looking forward to subsequent contributions.

Issue #5289, opened 2025-04-12
`StackOverFlowException` when parsing a Java file with recursive type definitions


uhafner:
## What version of OpenRewrite are you using?

I am using

<rewrite-maven-plugin.version>6.4.0</rewrite-maven-plugin.version>
<rewrite-testing-frameworks.version>3.6.1</rewrite-testing-frameworks.version>
<rewrite-static-analysis.version>2.6.0</rewrite-static-analysis.version>
<rewrite-migrate-java.version>3.7.0</rewrite-migrate-java.version>
<rewrite-recommendations.version>1.15.0</rewrite-recommendations.version>

## How are you running OpenRewrite?

Maven, see   https://github.com/jenkinsci/analysis-pom-plugin/blob/307b0a997b11b90498cb744bffb191aa8b9e8ec2/pom.xml#L269-L315

## What is the smallest, simplest way to reproduce the problem?

The original problem occurred for [DeltaComputationITest.java](https://github.com/jenkinsci/coverage-plugin/blob/main/plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/DeltaComputationITest.java), I stripped it down to:

```java
import java.util.Optional;

import static org.assertj.core.api.Assertions.*;

class BugTest {
    void run() {
        One<?, ?> firstBuild = null;
        assertThat(Optional.of(firstBuild))
                .isPresent()
                .satisfies(reference -> assertThat(reference.get()).isEqualTo(firstBuild));
    }

    abstract static class One<TwoT extends Two<TwoT, OneT>, OneT extends One<TwoT, OneT>> {
    }

    abstract static class Two<TwoT extends Two<TwoT, OneT>, OneT extends One<TwoT, OneT>> {
    }
}
```

## What did you expect to see?

The file should be read. 

## What did you see instead?

Exception, see below.

## What is the full stack trace of any errors you encountered?

```
Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:6.4.0:run failed: Error while visiting plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/BugTest.java: java.lang.StackOverflowError: null
  java.base/java.lang.StringBuilder.<init>(StringBuilder.java:106)
  org.openrewrite.java.tree.TypeUtils.toString(TypeUtils.java:884)
  org.openrewrite.java.tree.TypeUtils.toString(TypeUtils.java:911)
[...]

PR #5323, opened 2025-04-21
Fix recursive types

fmodesto:
## What's changed?
Fix handling recursive types.
- fixes https://github.com/openrewrite/rewrite/issues/5289

Not related, but should we handle the `$` symbol `BugTest$Rec<?>`? 

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
@@ -903,31 +903,44 @@ public class TypeUtils {
     }
 
     public static String toString(JavaType type) {
+        return toString(type, new IdentityHashMap<>());
+    }
+
+    private static String toString(JavaType type, IdentityHashMap<JavaType, Boolean> recursiveTypes) {
         if (type instanceof JavaType.Primitive) {
             return ((JavaType.Primitive) type).getKeyword();
         } else if (type instanceof JavaType.Class) {
             return ((JavaType.Class) type).getFullyQualifiedName();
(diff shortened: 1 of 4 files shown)


rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -903,31 +903,44 @@ static boolean deepEquals(@Nullable JavaType t, @Nullable JavaType t2) {
...
+                    recursiveTypes.remove(genericType);
+                    return variance + bound;
+                }
+                return "?";
             }
         } else if (type instanceof JavaType.Array) {
-            return toString(((JavaType.Array) type).getElemType()) + "[]";
+            return toString(((JavaType.Array) type).getElemType(), recursiveTypes) + "[]";
         }
         return type.toString();
```
knutwannheden, line 945: Looks good to me. But it would be best if we could add cases for JavaType.IbtersectionType and probably also JavaType.MultiCatch, which would otherwise end up here and could also indirectly be referencing recursive generics.

fmodesto: I had no idea `(Comparable<?> & Serializable)` was even a Java thing, until I saw one of the tests, let me add coverage for all of them

knutwannheden: Yeah, this is one of those Java features very few know about. But it is also very restricted in what it can be used for. I don't think I ever had to use it for anything except for OpenRewrite tests.

PR #5324, opened 2025-04-21
Drop long deprecated constructors

timtebeek:
## What's changed?
Drop deprecated constructors; in most cases these were added a long time ago, and we don't expect those to be required anymore due to changes downstream.

## What's your motivation?
Clear these out as we expand language support, such that have a cleaner slate to work with going forward.

This commit https://github.com/openrewrite/rewrite/commit/ab156d3bbf51e371c4fbd0a624b1fb62c8aae28f removes constructors from `IsNullSafe` and `KObject` where we now take _less_ fields than before with the remaining constrctor. That could have implications for deserialization so I'm calling it out here, but as these relate to Kotlin I don't expect a lot of folks to be affected. The solution would be to build those projects again, which we already recommend.

We could cherry pick 1147dda23ee3a8796ccd547a0b8451828b5a78a7 onto main first and only, and/or leave out ab156d3bbf51e371c4fbd0a624b1fb62c8aae28f for a smoother transition. Rejected initially as we're going to have to evaluate downstream anyway, and might as well then clear those out in one go.


timtebeek:
Looks like we'd have to adapt usage in https://github.com/openrewrite/rewrite-gradle-tooling-model as well.
```
ChangeDependencyGroupIdTest > findDependency(String, String) > [2] group=*, artifact=* FAILED
    java.lang.NoSuchMethodError: 'void org.openrewrite.maven.tree.MavenRepository.<init>(java.lang.String, java.lang.String, java.lang.String, java.lang.String, boolean, java.lang.String, java.lang.String, java.lang.Boolean)'
        at org.openrewrite.gradle.toolingapi.MavenRepository.toMarker(MavenRepository.java:61)
        at org.openrewrite.gradle.toolingapi.GradleProject$$Lambda$1094/0x000073cd2c8a8d90.apply(Unknown Source)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
```

PR #5330, opened 2025-04-22
Added option.release as a field that can set the java version.

Jenson3210:
This fixes one of the issues that our customers has been facing with Java version migration. 

according to [Gradle documentation](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:compiling_with_release), this property can also be used to set java compile version. 

What I am not sure about is if we do not want to go more in depth than just adapting all `options.release = ... `
What happens if i have another extension that also uses `options.release`? That one will also be changed with current code. 
I do not see a lot of risk there(and this risk is also there for the other properties `sourceCompatibility` and `targetCompatibility`). For this reason I've only added some assurance check by verifying the `target` is `options` instead of just putting all `release` assignments to the mentioned version. 

I've added a test for multiple ways of configuring the property (as we only use the internal assignment to match): 
```groovy
tasks.withType(JavaCompile) {
    options.release = ...
}

tasks.withType(JavaCompile).configureEach {
    options.release = ...
}

compileJava {
    options.release = ...
}

compileJava.options.release = ...
```

NOTE: After releasing this one, we also must release the [rewrite-migrate-java](https://github.com/openrewrite/rewrite-migrate-java) module for this change to take effect

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
@@ -132,7 +132,10 @@ public class UpdateJavaCompatibility extends Recipe {
                 } else if (a.getVariable() instanceof J.FieldAccess) {
                     J.FieldAccess fieldAccess = (J.FieldAccess) a.getVariable();
                     if (compatibilityType == null) {
-                        if (!("sourceCompatibility".equals(fieldAccess.getSimpleName()) || "targetCompatibility".equals(fieldAccess.getSimpleName()) || ("release".equals(fieldAccess.getSimpleName()) && fieldAccess.getTarget() instanceof J.Identifier && "options".equals(((J.Identifier) fieldAccess.getTarget()).getSimpleName())))) {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -132,7 +132,7 @@ public J visitAssignment(J.Assignment assignment, ExecutionContext ctx) {
                 } else if (a.getVariable() instanceof J.FieldAccess) {
                     J.FieldAccess fieldAccess = (J.FieldAccess) a.getVariable();
                     if (compatibilityType == null) {
-                        if (!("sourceCompatibility".equals(fieldAccess.getSimpleName()) || "targetCompatibility".equals(fieldAccess.getSimpleName()))) {
+                        if (!("sourceCompatibility".equals(fieldAccess.getSimpleName()) || "targetCompatibility".equals(fieldAccess.getSimpleName()) || ("release".equals(fieldAccess.getSimpleName()) && fieldAccess.getTarget() instanceof J.Identifier && "options".equals(((J.Identifier) fieldAccess.getTarget()).getSimpleName())))) {
```
Laurens-W, line 135: I wonder if we also have to take this into account on line 126

jevanlingen (approved):

PR #5336, opened 2025-04-23
Rename deprecated JaCoCo properties as part of Gradle 8 migration

Jenson3210:
- For more context: https://github.com/openrewrite/rewrite-migrate-java/pull/708

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
@@ -31,7 +31,7 @@ import org.openrewrite.java.tree.J;
 @EqualsAndHashCode(callSuper = false)
 public class JacocoReportDeprecations extends Recipe {
 
-    private static final String JACOCO_SETTINGS_PATH = "JACOCO_SETTINGS_PATH";
+    private static final String JACOCO_SETTINGS_INDEX = "JACOCO_SETTINGS_INDEX";
 
     @Override
     public String getDisplayName() {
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Set the `enabled` to `required` and the `destination` to `outputLocation` for Reports deprecations that were removed in gradle 8. " +
+                "See [the gradle docs on this topic](https://docs.gradle.org/current/userguide/upgrading_version_7.html#report_and_testreport_api_cleanup).";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsBuildGradle<>(), new GroovyIsoVisitor<ExecutionContext>() {
```
shanman190, line 49: Given these are working over common `J` types, if would be good to use a `JavaIsoVisitor` to handle support for the Kotlin migrations as well.

Jenson3210: Would you happen to have a syntax example where kotlin is used to use these deprecated properties?

I went back till gradle 6 in a demo project and for some reason, when using the enabled property the project configuration is invalid. It looks like that already needs the new properties. 

For that reason I did not make it a JavaIsoVisitor. But I Will try it again over here also.

Jenson3210: Added test for kotlin dsl + altered the recipe a bit to support the [documented gradle properties](https://docs.gradle.org/6.9.4/userguide/jacoco_plugin.html#sec:jacoco_report_configuration) also for kotlin dsl

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
```diff
@@ -0,0 +1,139 @@
...
+                if (parts.length >= 2 && !"reports".equalsIgnoreCase(parts[1])) {
+                    return false;
+                }
+                if (parts.length >= 3 && !"xml".equalsIgnoreCase(parts[2]) && !"csv".equalsIgnoreCase(parts[2]) && !"html".equalsIgnoreCase(parts[2])) {
+                    return false;
+                }
+                if (parts.length >= 4 && !"enabled".equalsIgnoreCase(parts[3]) && !"isEnabled".equalsIgnoreCase(parts[3]) && !"destination".equalsIgnoreCase(parts[3])) {
+                    return false;
+                }
+                return parts.length < 5;
```
sambsnyd, line 121: It feels inefficient that we end up allocating new arrays and re-checking the first fqn segments multiple times. But it's not a big deal in practice to have `O(n^2)` comparisons when `n == 5`.

Jenson3210: Triggering me to go the extra mile here. Love it! 
Used an alternative approach to only check the index we're currently at. This results in only 1 array being created when needed to do the last check (`visitAssignment`)

PR #5337, opened 2025-04-24
Add annotations to the type attribution for method parameters

knutwannheden:
The type attribution for method return types and method parameters currently lack annotations. Recipes might require these annotations to perform their task.


knutwannheden:
The approach in this PR is not complete:
* We also need to support `JavaType.Parameterized` and `JavaType.GenericTypeVariable` with annotations.
* Further, the approach here only works correctly for parsed source code. For Java types loaded from bytecode the code needs to look a bit different.
* Finally, possibly the annotated types should be cached by including the type use annotations in the signature (cache key).

sambsnyd (approved):
Nice

PR #5340, opened 2025-04-24
Support Scanning Recipes as preconditions in DeclarativeRecipe.

sambsnyd:
Back in rewrite 7 we had two kinds of preconditions: Repository level and file level. Repository level were uncommon but could be used to express things like "Run this recipe only if a particular type is in use or dependency is present". We lost that in the jump to rewrite 8, supporting only file-level preconditions. Until now.

With these changes a DeclarativeRecipe can use a scanning recipe as a precondition. This allows for not only repository level preconditions, but for any arbitrary granularity between file and repository. So this mechanism allows you to express nuanced preconditions like "only run this recipe on java sources gradle projects which apply a particular plugin".

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
@@ -337,18 +337,6 @@ public class DeclarativeRecipe extends ScanningRecipe<DeclarativeRecipe.Accumula
         return Preconditions.or(conditions.toArray(new TreeVisitor[0]));
     }
 
-    private static boolean isScanningRecipe(Recipe recipe) {
-        if (recipe instanceof ScanningRecipe) {
-            return true;
-        }
-        for (Recipe r : recipe.getRecipeList()) {
-            if (isScanningRecipe(r)) {
-                return true;
-            }
-        }
-        return false;
-    }
-
     private static List<Recipe> decorateWithPreconditionBellwether(PreconditionBellwether bellwether, List<Recipe> recipeList) {
         List<Recipe> mappedRecipeList = new ArrayList<>(recipeList.size());
         for (Recipe recipe : recipeList) {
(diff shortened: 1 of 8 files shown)


rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
```diff
@@ -32,7 +32,7 @@
 import static org.openrewrite.Validated.invalid;
 
 @RequiredArgsConstructor
-public class DeclarativeRecipe extends Recipe {
+public class DeclarativeRecipe extends ScanningRecipe<DeclarativeRecipe.Accumulator> {
```
knutwannheden, line 35: One worry I have here (yes, it is performance related) is that when every declarative recipe is a scanning recipe, these will always require two phases. Maybe the scheduler can check if the returned scanner is a noop or maybe the environment can be clever about either creating a DeclarativeRecipe or a DeclarativeScanningRecipe.

sambsnyd: That's a good point. I will consider what can be done there.

sambsnyd: I added this to `RecipeScheduler`'s scanning recipe check:
```
    private boolean hasScanningRecipe(Recipe recipe) {
        if (recipe instanceof ScanningRecipe) {
            // DeclarativeRecipe is technically a ScanningRecipe, but it only needs the
            // scanning phase if it or one of its sub-recipes or preconditions is a ScanningRecipe
            if(recipe instanceof DeclarativeRecipe) {
                for (Recipe precondition : ((DeclarativeRecipe) recipe).getPreconditions()) {
                    if (hasScanningRecipe(precondition)) {
                        return true;
                    }
                }
            } else {
                return true;
            }
        }
        for (Recipe r : recipe.getRecipeList()) {
            if (hasScanningRecipe(r)) {
                return true;
            }
        }
        return false;
    }
```

Maybe we shouldn't treat `DeclarativeRecipe` as a special snowflake, but I'd argue it merits it.

PR #5344, opened 2025-04-26
Drop unused JavaAIEditor and associated classes

timtebeek:
The only usage was in https://github.com/openrewrite/rewrite-generative-ai, but the code-davinci-edit-001 model was shut down on 2024-01-04. Rather than maintain these classes with a split between here and there, there's a branch to update usage in rewrite-generative-ai: https://github.com/openrewrite/rewrite-generative-ai/compare/main...switch-to-gpt-4o


timtebeek:
- Downstream changes made in https://github.com/openrewrite/rewrite-generative-ai/pull/23

PR #5348, opened 2025-04-28
Support two modes for handling `GenericTypeVariables` in `TypeUtils.isAssignableTo(..)` and `TypeUtils.isOfType(..)`

fmodesto:
## What's Changed

- `TypeUtils.isAssignableTo(..)` and `TypeUtils.isOfType(..)` now support two modes for handling `GenericTypeVariable`s:
  - **Bound mode**: Follows standard Java generic rules (e.g., `<T, S extends T>` `T` can only be assigned from `T` and `S`).
  - **Infer mode**: Attempts to infer if a type variable *could* be bound to the target type (e.g. `<T extends Number>` `T` can be assigned from `Integer`).

> ⚠️ Note: While Infer mode includes some validation, it does not fully track substitutions across type variables.
> For example, given `Map<T, T>` where `<T extends Number>`, it will:
> - ❌ Reject Map<String, String> (as String is not a Number)
> - ✅ Accept Map<Integer, Long> (as inference treats both as matching T, without enforcing consistency)

- **`SemanticallyEquals`**
  - All calls to `isAssignableTo` and `isOfType` are now protected methods, allowing them to be overridden.

- **`JavaTemplateSemanticallyEquals`**
  - Replaces all `isOfType` calls with `isAssignableTo` to enable more flexible generic matching.

- **`TypeUtils`**
  - Most logic has been moved to `Types`, which is now instance-based to support shared state (e.g., binding stacks) and avoid infinite recursion during generic inference.

- **`Types`**
  - Now contains the core logic for `isAssignableTo` and `isOfType`.
  - Improved support for matching parameterized types by propagating type parameters down the type hierarchy via `maybeResolveParameters`.

## What's your motivation?
Enable full support for generics in `JavaTemplate`, including type inference.

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/tree/Types.java
```diff
@@ -0,0 +1,542 @@
...
+        JavaType to;
+        JavaType from;
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (!(o instanceof TypePair)) return false;
+
+            TypePair that = (TypePair) o;
+            return operation == that.operation && mode == that.mode && to == that.to && from == that.from;
```
timtebeek, line 530: This `==` comparison will match then the types are identical as parsed from the same source, but I have a slight worry about how `to == that.to && from == that.from` will hold up when recipes are chained together; it's not uncommon for us to use `ShallowClass.build(String)` in recipes, which will then of course result in a different instance, even if the types they represent might still be the same as what we compare against.

Any thoughts here?

timtebeek (commented):
Did a quick round of polish and posed a question that would be good if others could weigh in on as well. I've not gone in depth fully on `Types` and `TypeUtils` just yet; I suggest that's where the others focus their review.

rewrite-java/src/main/java/org/openrewrite/java/tree/Types.java
fmodesto: The `==` identity check is used to prevent infinite recursion when dealing with recursive types (e.g., `E extends Enum<E>`). I was cautious about using `.equals()` or `.isOfType()` because those might themselves trigger recursion.

#### Expansion of `isAssignableTo(Enum<E>, MyEnum)`

| TO        | FROM         | Explanation |
|-----------|--------------|-------------|
| `Enum<E>` | `MyEnum`     | FQNs don't match, so check type hierarchy |
| `Enum<E>` | `Enum<MyEnum>` | FQNs match |
|   \ `<E>`     |   \ `<MyEnum>`   | Check if both lists are same size and positionally assignable |
| `E`       | `MyEnum`     | `E` is generic with bounds `E extends Enum<E>`, expand them |
| `Enum<E>` | `MyEnum`     | We have this in the expansion, is a recursive type, assume true |

This logic depends on the fact that when traversing the types  **their getters consistently return the same object references**. There are two cases where types are reconstructed:

- When resolving the type hierarchy (e.g., `MyEnum` to `Enum<MyEnum>`), a new type is created with potentially different type arguments—but the type arguments themselves are not modified.
- When expanding a type variable to its bounds (e.g., `E` to `Enum<E>`), if the bounds are a list, an `Intersection` type is created. These intersections are cached to preserve identity semantics via `==`. But if there is a cycle, it will be encountered before the expansion, so not sure if the cache is necessary.

sambsnyd:
Nice work @fmodesto !

Jenson3210:
I have created a unit test (borrowed from the static analysis module) after analyzing how that one would fail. 
It turns out it uses the `RenameVariable` internally -> coded the test for this one to showcase the issue. 

Other than that, I have mentioned in the classes where the "bug" got introduced that caused this regression. 
@fmodesto or someone else, if you have any idea how to tackle this one, I would be happy to take it's implementation further . At the moment, I am still not getting 100% of why `RenameVariable` needs this check. 

My next point of attcking this issue is by further analyzing this visitor to see if the `isOfType` information is really needed here or if we can do it differently. 

Bottom root cause: these changes introduced "unwrapping" the method into it's returned type pre-comparing with the type.

PR #5350, opened 2025-04-28
Allow parallel `RewriteTest` runs using `javaVersion`

timtebeek:
## What's changed?
Change the `static` fields in `Assertions` to use `ConcurrentHashMap` instead of `HashMap`.

## What's your motivation?
We saw `ConcurrentModificationException` when using the following trio of methods together with JUnit platform parallel tests:
https://github.com/openrewrite/rewrite/blob/20dc764c508a7be8d1c5e197f39fe7508a510c2e/rewrite-java/src/main/java/org/openrewrite/java/Assertions.java#L262-L276
This happened specifically in rewrite-static-analysis, where there's various test classes using these static methods, that could trigger at the same time to populate these maps.

I'd expect the performance impact to be small here: they are not used as often as other constructs, and could provide a speed up by allowing recipes to run in parallel.

We could make those methods above `synchronized`; `ConcurrentHashMap` seemed the better option.

## Any additional context
We saw test runs start to take longer and longer downstream, up to 30 minutes. Running test classes in parallel could speed that up.

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
@@ -32,7 +32,7 @@ class JacocoReportDeprecationsTest implements RewriteTest {
 
     @Test
     @DocumentExample
-    void enabledDeprecatedInNormalSyntax() {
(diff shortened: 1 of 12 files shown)


greg-at-moderne (approved):
It looks fine to me. I fail to understand why it would be needed TBH, but I see no harm. Thus approving.

Issue #5342, opened 2025-04-25
DependencyUseStringNotation whitespace behaviour inconsistent


Jenson3210:
## What version of OpenRewrite are you using?
I am using

- DependencyUseStringNotation of org.openrewrite:rewrite-gradle:8.51.0
- Moderne CLI latest version

## How are you running OpenRewrite?
`mod  run . --recipe DependencyUseStringNotation`

## What is the smallest, simplest way to reproduce the problem?
```groovy
plugins {
    id 'java'
}

dependencies {
    implementation( group: 'org.apache.commons', name: 'commons-lang3', version: '3.0' )
}

repositories {
    mavenCentral()
}
```

## What did you expect to see?
```groovy
implementation("org.apache.commons:commons-lang3:3.0")
// OR
implementation( "org.apache.commons:commons-lang3:3.0" )
```

## What did you see instead?

```groovy
implementation( "org.apache.commons:commons-lang3:3.0")
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
sure

Jenson3210:
[This line](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/JRightPadded.java#L156) gets reached which makes the string notation not have an after space. 

reaches through:
--> DependencyUseStringNotation:visitMethodInvocation:116 
--> J$MethodInvocation:withArguments:4034
--> JContainer:withElements:171
--> JRightPadded:withElements:156

I do not think we can without risk use the last before's space as an after in case it's a new element (as elements could also be inserted in between 2 others. 

Could we perhaps expose the concepts of `getArgumentsContainer()` and `withArgumentsContainer(JContainer<Expression> arguments)` and `withLastSpace(Space space)`? This way we would get back the JContainer which [offers the lastSpace](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/JContainer.java#L97-L99), call the wither to update it's space, call the withArgumentsContainer to add back that last space. 
However, then we open ourselves to a lot of new getters and withers as there would be other Containers like `typeParameters` etc also impacted atm...

shanman190:
This also gets especially difficult for the lambda argument case when a dependency declaration includes it in which case you want the 2nd to last before.

knutwannheden:
[...]

PR #5352, opened 2025-04-28
Allow JContainer content modifications through the JRIghtPadded elements rather than only on the internal Expressions

Jenson3210:
## What's changed?
Added a new method in MethodCall J objects so they can workn with the previous JRightPadded elements to keep certain stuff properly formatted.

## What's your motivation?
Trailing whitespaces are not consistenly kept like leading ones in all cases.
Fixes: #5342 

@shanman190 @timtebeek @knutwannheden 

## Any additional context
This is a draft to have easier discussions on the code/output

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
@@ -23,10 +23,7 @@ import org.openrewrite.TreeVisitor;
 import org.openrewrite.gradle.trait.GradleDependency;
 import org.openrewrite.groovy.GroovyVisitor;
 import org.openrewrite.groovy.tree.G;
-import org.openrewrite.java.tree.Expression;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JavaType;
-import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 
 import java.util.*;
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
```diff
@@ -109,12 +106,22 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                         return m;
                     }
 
-                    Expression lastArg = m.getArguments().get(m.getArguments().size() - 1);
-                    if (lastArg instanceof J.Lambda) {
-                        m = m.withArguments(Arrays.asList(stringNotation, lastArg));
-                    } else {
-                        m = m.withArguments(Collections.singletonList(stringNotation));
-                    }
+                    m = m.withArguments(originalArgs -> {
```
shanman190, line 109: callout: I think this would be possible without new API surface area by getting the elements off of the padding here instead. Then the internals of the lambda invocation would continue to be used as they are here.

ie. Something like:
```
List<JRightPadded<Expression>> paddedArguments = m.getPadding().getArguments().getPadding().getElements();
...
```

However, I do like how the declared helper method makes working with this segment a bit clearer and easier.

Jenson3210: hmm... 🤔  Wouldn't we still need a `withArgumentsContainer(JContainer)` method for the arguments to be added to the `MethodInvocation` in that case? 
If we must add something, I would add the the helper which indeed will make cleaner code.

shanman190: So the padding itself has a method `withArguments(JContainer)`. So it'd look something like this:

```
 List<Expression> originalArguments = m.getPadding().getArguments().getPadding().getElements();
 ...
 return m.getPadding().withArguments(JContainer.build(updatedArguments));
```

Jenson3210: @timtebeek @knutwannheden  what do you guys think here? 
I can adapt this recipe to use the Padding elements to fix this particular issue. 
However this introduces a lot of "styling code" for a single whitespace. 
OR 
we create a new API surface area for this helper method on MethodCall and make this (and future) recipes cleaner to manage these paddings.

timtebeek: I would lean towards not changing the `J` file here just yet. It seems we'd be copying some patterns from `ListUtils` to the LST element, whereas we've been trying to move logic out of those and into Traits where possible, since the LST elements should mostly just be dumb representations of what's available in the source code. In this case I think it's fine to fix this once in this recipe, and if we find a recurring need we can reevaluate where that might best fit.

knutwannheden: I agree with @timtebeek . If you can find some static utility on JContainer that would help, that could be an option.

PR #5356, opened 2025-04-28
Do not bump transitive dependencies if there is a direct dependency also

### LINKED ISSUE #5355: UpgradeTransitiveDependencyVersion adds constraint despite dependency being defined at depth 0

I am using latest in main

Added below to `UpgradeTransitiveDependencyVersionTest`:
```java
    @Test
    void test() {
        rewriteRun(
          buildGradle(
            """
              plugins {
                id 'java'
              }
              
              dependencies {
                  implementation('com.fasterxml.jackson.core:jackson-core:2.12.0')
                  implementation('com.fasterxml.jackson.core:jackson-databind:2.12.0')
              }
              """
          )
        );
    }
```

According to [the docs](https://docs.openrewrite.org/recipes/java/dependencies/upgradetransitivedependencyversion), it would seem a constraint shouldn't be added since `jackson-core` is a direct dependency (as well as being a nested dependency of `jackson-databind`).

## What did you see instead?

Yes

Jenson3210:
## What's changed?
We ignore the transitive dependency / configuration if there is a direct dependency for the same configuration. This way we can end up with cleaner build.gradle files AND we are more true to [the original documentation](https://docs.openrewrite.org/recipes/java/dependencies/upgradetransitivedependencyversion) of this recipe. 

> Upgrades the version of a transitive dependency in a Maven pom.xml or Gradle build.gradle. **Leaves direct dependencies unmodified.** Can be paired with the regular Upgrade Dependency Version recipe to upgrade a dependency everywhere, regardless of whether it is direct or transitive.

## What's your motivation?
- Fixes #5355

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -42,6 +42,7 @@ import org.openrewrite.marker.Markup;
 import org.openrewrite.marker.SearchResult;
 import org.openrewrite.maven.MavenDownloadingException;
 import org.openrewrite.maven.table.MavenMetadataFailures;
+import org.openrewrite.maven.tree.Dependency;
 import org.openrewrite.maven.tree.GroupArtifact;
 import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedDependency;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see! I've shortened it a bit with a label and applied the formatter. Hope you agree!

Jenson3210:
Here I am, learning something new about Java 8 in 2025 🥹

timtebeek:
> Here I am, learning something new about Java 8 in 2025 🥹 Never saw labels before. I can agree though I feel we should go lightly with these?

Hah, yes it's not something to overuse, but it this case it seemed a perfect match to clearly and early move onto the next iteration of the outer loop.

PR #5359, opened 2025-04-30
Java - not unwrapping parens around switch expressions

greg-at-moderne:
## What's changed?

Changing `UnwrapParentheses` not to remove parenthesis around switch expressions.

## What's your motivation?

- fixes https://github.com/openrewrite/rewrite-static-analysis/issues/534


knutwannheden:
@greg-at-moderne I suspect the same kind of change might also be required for the Parenthesize visitor. At some point it would be nice if there were no duplication of logic between the two.

PR #5363, opened 2025-05-01
Use Java primitives in Kotlin model to match types

timtebeek:
## What's changed?
Add type mapping from Kotlin primitive types to their Java counterparts, and use those in the type system.

## What's your motivation?
As described in https://github.com/openrewrite/rewrite-kotlin/pull/626 we had up to now mapped any primitive as a Kotlin primitive, even when those classes come from Java. That's problematic when say a method matcher for a Java method has a primitive type declared for the matcher. Those would then fail to match when used from Kotlin. While that's perhaps not very common, it's nonetheless important as we look to cover both with a single set of recipes, and for new recipes developed.

Should we continue to use Kotlin primitives types when methods are defined in Kotlin? How do we tell those apart? Can we make that work on 1.9? Would it help to move to 2.0 or 2.1?

## Any additional context
- Continues https://github.com/openrewrite/rewrite-kotlin/pull/626


timtebeek:
Linking some earlier comments on the older PR:
- https://github.com/openrewrite/rewrite-kotlin/pull/626#issuecomment-2585471464
- https://github.com/openrewrite/rewrite-kotlin/pull/626#issuecomment-2585830207

PR #5365, opened 2025-05-02
Default to no inference on Types

timtebeek:
## What's changed?
Add a no args constructor to Types that defaults to not using inference.

## What's your motivation?
Make it easier to reuse without having to think of an explicit argument to pass in.

Considered adding a `Types.withInference()` static method instead of `new Types(true)`, but figured for the single use we see today likely not worth adding just yet.

## Any additional context
- Follow up from https://github.com/openrewrite/rewrite/pull/5348


timtebeek:
Merging as this was discussed on Slack.

PR #5366, opened 2025-05-02
Replace `J.MethodInvocation`s with constant

MBoegers:
## What's changed?

Added a recipe to replace MethodInvocations with a constant values.

## Any additional context

See `System.getSecurtiyManager()` it always returns `null`from java 24. This recipe replaces the values and enable cleanup recipes.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ReplaceMethodInvocationWithConstant.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ReplaceMethodInvocationWithConstant.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ReplaceMethodInvocationWithConstant.java
```diff
@@ -0,0 +1,74 @@
...
+                return replacementTemplate.apply(ma.getCursor(), coordinateOf(ma).replace());
+            }
+
+            CoordinateBuilder.Expression coordinateOf(MethodAccess ma) {
+                return ((Expression) ma.getCursor().getValue()).getCoordinates();
+            }
+        };
+    }
+
+
```
timtebeek, line 73:
```suggestion
        return Preconditions.check(
                new UsesMethod<>(methodPattern),
                Traits.methodAccess(methodPattern)
                        .asVisitor(ma ->
                                JavaTemplate.apply(replacement, ma.getCursor(),
                                        ((Expression) ma.getCursor().getValue()).getCoordinates().replace())));
    }
```

timtebeek (approved):
Approved already, provided there's no new code suggestions to update examples.yml following the addition of a new `@DocumentExample` here.

knutwannheden:
I think it would be good with a test case for when a method is only called for side effects. To be decided what the recipe should do then. Replacing by a null literal would be a bad choice, I think.

PR #5369, opened 2025-05-02
Fix regression in RenameVariable

Jenson3210:
## What's changed?
The failing downstream test has been added also over here so we can catch such error later sooner.

## What's your motivation?
Fix regression introduced in 
- #5348

Is checking equals (where other code ended up) enough? 
Do we also have to check equality in the if-branches? 
I feel as all tests (including the added one) are now green, it should be good to go. 
We can always revise later but this breaking change could introduce a lot of wringly renamed variables / broken/non-building code.

@fmodesto, think you would be best positioned here to review also when I have a solution
@timtebeek or @sambsnyd if you have anything here also (as you are aware of the original PR)?
Please look at[ this commit](https://github.com/openrewrite/rewrite/pull/5369/commits/cf2138cd970a079ae04a441bb6bf09089455c91e) if you need to know the reason of failure/change

Further digging deeper into why the other if-blocks are present (however, we have no other failing tests anymore -> tests should be added when we know the code that is broken / why these branches of code exist)

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
@@ -995,4 +995,41 @@ class RenameVariableTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void troubleshooting() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new JavaVisitor<>() {
+                @Override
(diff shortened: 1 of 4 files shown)


timtebeek:
Thanks for the late work and fixes here! Not behind my laptop now, but will look when I can. Unless @sambsnyd beats me to it. Agree this seems safe to merge and iterate if needed.

timtebeek (approved):
Thanks a lot for the fix here! Appreciate you sharing the analysis you did, and holding off on more extensive changes that are harder to oversee. Great to have the downstream project passing again, without incorrect changes.

PR #5371, opened 2025-05-03
Add AddPlatformDependency recipe.

### LINKED ISSUE #5357: Support adding platform dependencies to Gradle project
It would be useful to be able to add platform dependencies to a project.

Following from this, with [RemoveRedundantDependencyVersions](https://docs.openrewrite.org/recipes/gradle/removeredundantdependencyversions)

- #3290

may become possible via a declarative recipe?

Improve the `AddDependency` recipe to support adding platform dependencies

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes

BrendanHart:
## What's changed?

Add recipe to support adding platform/enforcedPlatform dependencies to Gradle build files.

## What's your motivation?

- closes #5357 

### Checklist


BrendanHart:
Currently decided to not include "onlyIfUsing" in this recipe, mostly as unsure if it should be explicit if it would be best for users to be explicit in where they want to add a platform dependency?

shanman190:
So having only read your comment here about `onlyIfUsing`, in a multi module project, what would be the pattern for only adding the platform dependency to only some of the modules?

Maybe if we go ahead and include something like a `ProjectUsesType` recipe that was project aware and could be utilized as a project-level precondition? That would mean that the `onlyIfUsing` pattern could safely be soft deprecated like we did with `filePatterns` on other recipes.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddPlatformDependency.java
```diff
@@ -0,0 +1,281 @@
...
+    @Option(displayName = "Enforced",
+            description = "Used to determine whether the platform dependency should be enforcedPlatform.",
+            example = "true",
+            required = false)
+    @Nullable
+    Boolean enforced;
+
+    @Override
+    public String getDisplayName() {
+        return "Add Gradle dependency";
```
shanman190, line 93:
```suggestion
        return "Add Gradle platform dependency";
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddPlatformDependency.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    @Override
+    public String getInstanceNameSuffix() {
+        return String.format("`%s:%s:%s`", groupId, artifactId, version);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Add a gradle dependency to a `build.gradle` file in the correct configuration based on where it is used.";
```
shanman190, line 103:
```suggestion
        return "Add a gradle platform dependency to a `build.gradle` file in the correct configuration based on where it is used.";
```

BrendanHart:
Maybe there's the option of using the dependencies also to work out whether to add a platform dependency? E.g. if a module is using any dependencies matching a group `org.springframework*` then add the `spring-boot-dependencies` platform dependency.

Sorry, could you help me understand more on your second point, not sure I followed. Would that be for refactoring out the `onlyIfUsing` logic into something common that can be used across multiple recipes?

shanman190:
The short answer is that preconditions are a much more powerful and expressive way to capture these types of extra conditioning (types in use, acceptable file patterns, etc). 

With preconditions we can build all kinds of acceptance rules and combine them together arbitrarily in declarative recipes, including using DependencyInsight in the manner that you described for other Spring Boot dependencies being present. There are a ton of possible reasons that you may want to use a condition around when to add a dependency and by having the flexibility of preconditions it means that each of those different "reasons" can be utilized arbitrarily as a recipe author is building out their recipe(s).

BrendanHart:
Thanks for elaborating! Is the suggestion then that these recipes don't require such restrictions built in and we can assume them to be handled in the precondition recipes, and embedding logic into the recipe to restrict to specific modules isn't required?

shanman190:
That's kinda what I was thinking.

I'd probably go one step further to go ahead and create the precondition recipe (not 100% on the name of it though) while we're here thinking about it, but that's not strictly required. WDYT, @timtebeek ?

timtebeek:
Do you mean to create a recipe to check if a GradleProject is using any dependency managed by a BOM? And then to add that as a precondition check on the AddPlatformDependency recipe? would be ideal, I think, to make something like this just work out of the box:

```java
@Test
void twoProjects() {
    rewriteRun(
      spec -> spec.recipe(new AddPlatformDependency(
        "org.springframework.boot", "spring-boot-dependencies", "3.2.4", null, null, null)),
      mavenProject("projectA",
        buildGradle(
          """
            plugins {
                id "java-library"
            }
            repositories {
                mavenCentral()
            }
            dependencies {
                testImplementation "org.springframework:spring-core:6.1.5"
            }
            """,
          """
            plugins {
                id "java-library"
            }
            repositories {
                mavenCentral()
            }
            dependencies {
                testImplementation platform("org.springframework.boot:spring-boot-dependencies:3.2.4")
                testImplementation "org.springframework:spring-core"
            }
            """
        )
      ),
      mavenProject("projectB",
        buildGradle(
          """
            plugins {
                id "java-library"
            }
            repositories {
                mavenCentral()
            }
            """
          // No change, as we're not using any dependencies managed in the platform bom added
        )
      )
    );
}
```

shanman190:
So I was more thinking of:

* `AddPlatformDependency` adds unconditionally
* Create secondary recipes for the various conditioning requirements (ie: only if using, has managed dependency, etc) to be used as preconditions on a declarative recipe (or composite programmatic recipe).

I do like the idea of scheduling a `RemoveRedundantDependencyVersions` recipe execute as an after visit though.

rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleConfigurationFilter.java
```diff
@@ -0,0 +1,70 @@
...
+import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
+import org.openrewrite.gradle.marker.GradleProject;
+import org.openrewrite.maven.tree.GroupArtifact;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+public class GradleConfigurationFilter {
```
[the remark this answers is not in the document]
BrendanHart: Makes sense, thank you!

shanman190:
Just a FYI, I have an idea that I'd like to play with locally for a day or two that would change the need for the `GradleConfigurationNames` over to a `Trait` implementation for the specific feature that it's trying to cover. I've got the `JvmTestSuite` trait working locally, but am presently trying to adapt it to be able to have an `addDependency` helper function to completely centralize the configuration name editing in addition to the traits overall simplification and accuracy around JvmTestSuite usage and identification.

BrendanHart:
Added more test cases and made the changes mentioned. Marked PR as ready but of course pending any changes you wish to make from regarding the configuration names as already mentioned 😄

BrendanHart:
Thanks for the review @jevanlingen ! Will await @shanman190 's changes and any further comments I can address.

shanman190:
@BrendanHart, @jevanlingen, and @timtebeek, apologies for the delay on this, but I've pushed the new `JvmTestSuite` trait. I was able to run tests successfully pre merge from main, but then began to have issues with recent Gradle marker/model changes. I'm hopeful that CI doesn't produce any new errors, but let me know what you all think.

As part of this, you'll also notice that we can also now add a `dependencies` block where one was missing within a `JvmTestSuite` as well. So that'll be handy also.

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/FindJVMTestSuites.java
```diff
@@ -60,58 +55,31 @@ public String getDescription() {
...
-                return super.visitMethodInvocation(method, ctx);
-            }
-        });
+            return SearchResult.found(suite.getTree());
+        }));
     }
 
-    public static Set<String> jvmTestSuiteNames(Tree tree, boolean definesDependencies) {
-        return TreeVisitor.collect(new FindJVMTestSuites(definesDependencies, false).getVisitor(), tree, new HashSet<>())
+    public static Set<String> jvmTestSuiteNames(Tree tree) {
```
shanman190: This is technically no longer used, so we could remove it entirely, but it is a part of the public API surface.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/JvmTestSuite.java
```diff
@@ -0,0 +1,312 @@
...
+                                classifier,
+                                ctx
+                        );
+                    }
+
+                    return j;
+                }
+            };
+        } catch (MavenDownloadingException e) {
+            return TreeVisitor.noop();
```
sambsnyd: Perhaps instead leave a `Marker.warning()` so the failure isn't lost, otherwise it could be a debugging challenge to answer the dreaded "Why isn't this recipe making the changes I expect?"

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/JvmTestSuite.java
```diff
@@ -0,0 +1,312 @@
...
+                    if (j instanceof JavaSourceFile && j != tree) {
+                        return AddDependencyVisitor.addDependency(
+                                (JavaSourceFile) j,
+                                gradleProject.getConfiguration(targetConfiguration),
+                                new GroupArtifactVersion(groupId, artifactId, resolvedVersion),
+                                classifier,
+                                ctx
+                        );
+                    }
+
```
sambsnyd, line 107: ~I don't understand why `AddDependencyVisitor` is being invoked twice. Why can't this be just once?~
I understand now

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/JvmTestSuite.java
shanman190: I get that. Would we want the marker on the particular element or would the SourceFile be satisfactory?

PR #5372, opened 2025-05-03
Restore TypeUtils.isOfType behaviour

fmodesto:
## What's changed?
- Fix previous https://github.com/openrewrite/rewrite/pull/5348 unintended change, which cause regression https://github.com/openrewrite/rewrite/pull/5369
- Added more tests for `RenameVariable` and fix small issue.

## What's changed?
The original `TypeUtils.isOfType()` performed full checks (name, owner, parameters, etc.) for `JavaType.Method`, `JavaType.Variable`, and `JavaType.Annotation`. The new `Types.isOfType()` implementation normalizes these types to their expression type, aligning with `isAssignableTo` logic, to be used in `JavaTemplateSemanticallyEquals`. This causes some cases to now return `true` where the old logic returned `false`.

This change preserves backward compatibility by:
- Keeping the old behavior in `TypeUtils.isOfType()`
- Letting `Types.isOfType()` adopt the normalized type approach

The original `TypeUtils.isOfType()` had surprising edge cases:
- **Null handling**: Returns `false` if one type is `Unknown` but `true` if both are `null`
- **Method checks**: Validates _everything_ (name, return type, exceptions, type arguments)
- **Field checks**: Only verifies owner and type (_ignores field names_)

Is type normalization in `Types.isOfType()` the right approach, or should we keep the original behaviour there, or fix it instead.

---

I'm not sure what was the intended logic in `fieldAccessTargetsVariable`. In this test:
```
class Base {
    protected Base visible;
    protected Base hidden;

    Base base() {
        return hidden;
    }
}

class Middle extends Base {
    Middle middle() {
        return (Middle) hidden; // Case 2
    }
}

class Extended extends Middle {
    private Base hidden;

    public Extended test(Base hidden) {
        hidden.hidden = hidden; // Case 1
        return this;
    }
}
```
When renaming `Base.hidden` to `Base.changed`. 
- In `Extended.test()` `hidden.hidden` got both replaced to `changed.changed`. (Case 1)
- And it did not change the `hidden` in `Middle.middle()` method. (Case 2)

For case 1. I had to change the logic to ensure only identifiers whose parent was `FieldAccess` were changed, if they were the actual field and not the `target`. This lead to some simplification in the rules, only check for the `target` type to be the `owner` of the field. This pass all existing tests, except for the `Middle.middle()` case. But could there be any edge case?

To fix the `Middle.middle()` case 2. I have to introduce a check, if the Identifier fieldType is the actual variable to be rename, apply the rename. But that make the previous code never to be executed.

@timtebeek and @Jenson3210 

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
@@ -1033,4 +1033,288 @@ class RenameVariableTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void hiddenVariablesHierarchyRenameBase() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new JavaVisitor<>() {
+              @Override
+              public J visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
+                  J target = getCursor().getParentTreeCursor().getParentTreeCursor().getValue();
+                  if ("hidden".equals(multiVariable.getVariables().get(0).getSimpleName()) &&
+                    target instanceof J.ClassDeclaration && ((J.ClassDeclaration) target).getSimpleName().equals("Base")) {
+                      doAfterVisit(new RenameVariable<>(multiVariable.getVariables().get(0), "changed"));
+                  }
+                  return super.visitVariableDeclarations(multiVariable, ctx);
+              }
+          })),
+          java(
+            """
+              class Base {
+                  protected Base visible;
+                  protected Base hidden;
+              
+                  Base base() {
+                      return hidden;
+                  }
+              }
+              
+              class Middle extends Base {
+                  Middle middle() {
+                      return (Middle) hidden;
+                  }
+              }
+              
+              class Extended extends Middle {
+                  private Base hidden;
+              
+                  Extended extended() {
+                      return (Extended) hidden;
+                  }
+              
+                  public Extended test(Base hidden) {
+                      this.hidden = super.hidden;
+                      this.hidden = hidden.hidden;
+                      this.hidden = ((Middle) hidden).hidden;
+                      this.hidden = ((Extended) hidden).hidden;
+                      base().hidden = this;
+                      middle().hidden = this;
+                      extended().hidden = this;
+                      super.hidden = visible.hidden;
+                      visible.hidden = hidden;
+                      hidden.hidden.hidden = hidden.hidden = this.hidden = hidden;
+                      return this;
+                  }
+              }
+              """,
+            """
+              class Base {
+                  protected Base visible;
+                  protected Base changed;
+              
+                  Base base() {
+                      return changed;
+                  }
+              }
+              
+              class Middle extends Base {
+                  Middle middle() {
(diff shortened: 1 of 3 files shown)


Jenson3210 (approved):
This is some 'Extended' programming. I only had 'Base' understanding of the issue when I bypassed the introduced regression yesterday since I could not have hoped one would find a fix this fast. 
As all tests are still passing, and you only introduced more tests/coverage that cover these visitor changes, I have way above 'Middle' faith that the changes made here are correct. (Boy, I would hate to see these constructors from the new test used in Production code 😅)
Still can't say I understand 100% of these 'Typeutils' icw 'Types' etc yet though. 

Thanks a lot for the fast resolution and taking us (me at least) along in the what and why of certain changes with a very clear PR description!

knutwannheden:
Looking at that `isOfType(Annotation, Annotation) I introduced in https://github.com/openrewrite/rewrite/commit/d6d91ee1b55c971ce860a6af034e58f62704eb7b, I see that this was too hasty and misses a few things by not considering that the annotation elements could be in a different order or that an annotation might just repeat the default value and thus might be equivalent to another annotation which has omitted repeating the default value. But that isn't directly related to this PR.

timtebeek (approved):
Impressive test coverage here indeed; and great to see how that not only fixes the issue but also simplifies `RenameVariable`. Thanks a lot!

timtebeek:
Bit of a puzzling failure seen downstream in CI, but not locally for me in [org.openrewrite.staticanalysis.HiddenFieldTest](https://ge.openrewrite.org/s/qzsn6oqaav6em/tests/task/:test/details/org.openrewrite.staticanalysis.HiddenFieldTest).blocks()
```diff
diff --git a/A.java b/A.java	
index 7d6bcbe..358ebdf 100644	
--- a/A.java	
+++ b/A.java	
@@ -3,12 +3,12 @@ 	
 	
     public void blocks() {	
         {	
-            int n1 = 0;	
-            int x = n1;	
+            int n2 = 0;	
+            int x = n2;	
         }	
         {	
-            int n1 = 0;	
-            int x = n1;	
+            int n2 = 0;	
+            int x = n2;	
         }	
     }	
 }	
\ No newline at end of file
```
As you can see there's two blocks here which are each unique scopes, and the recipe should only rename a variable if it's hiding another one. Not sure I even agree with the original asserted output to begin with, but it's odd that we're picking `n2` now:
https://github.com/openrewrite/rewrite-static-analysis/blob/a682e11ee59d2d6bddcd34040853823e4779d405/src/test/java/org/openrewrite/staticanalysis/HiddenFieldTest.java#L266-L305

Jenson3210:
@fmodesto @timtebeek @MBoegers 

I created 2 more tests (1 is the failing one) and was able to replicate the failure in this project by calling the postVisit with the correct statements to show issue. 
Notice how I made sure to only run the visitor that renames for 1 of the 2 blocks, yet the output in the failing test has updated both. 
I identified that this is caused by the if statement (I commented above it)
putting `currentNameScope.size() == 1 && ` in the if statement that checks the types fixes this one test but breaks the Base/Middle/Extended test case. 

If you have any idea, very welcomed 😄

PR #5373, opened 2025-05-03
Move file to another directory

Jenson3210:
## What's changed?
Added a recipe that moves files to another directory (resulting in renamed directories if all files are moved)

## What's your motivation?
As there could be many use cases to in bulk rename directories, move files... 

@timtebeek

## Any additional context
[Requested in slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1746290845227519)

### Checklist


shanman190:
I spy *nix specific paths which would also make these tests fail when executed on a Windows machine or when using Windows based paths.

Jenson3210:
I indeed needed to check still with @timtebeek how Windows would behave on the parser. Was kinda hoping parsing mapped to single notation. But apparently not then 😖

Jenson3210:
> I spy *nix specific paths which would also make these tests fail when executed on a Windows machine or when using Windows based paths.
@shanman190 
🙏 thanks for the observation!
I indeed needed to check still with @timtebeek how Windows would behave on the parser. Was kinda hoping parsing mapped to single notation. But apparently not then 😖

I did write some behaviour to how I hope it is working plus the tests, but need to check :)

shanman190:
@Jenson3210 , is the plan to only support *nix style paths for the `moveTo` option? Otherwise, that probably needs some adaptation to support Windows-style paths.

Jenson3210:
@shanman190 yes indeed. 

As the SourcePath is always relative (still have to check if this is the case with Tim) to the repo root, I was thinking to have 1 uniform syntax for moveTo math (absolute inside repo root or relative to the current file/folder) 

Otherwise our recipes using this one always have to call it with a syntax for Windows and one for Unix, resulting in 2 recipes in the recipe list or complexer custom recipes.

Jenson3210:
@shanman190 yes indeed. 

As the SourcePath is always relative (still have to check if this is the case with Tim) to the repo root, I was thinking to have 1 uniform syntax for moveTo math (absolute inside repo root or relative to the current file/folder) 

Otherwise our recipes using this one always have to call it with a syntax for Windows and one for Unix, resulting in 2 recipes in the recipe list or complexer custom recipes.

shanman190:
Or you choose to accept either and normalize the input within the recipe to allow the user to specify either, but handle it internally in a uniform way. This is what is done internally within `PathUtils` for matching as an example.

Jenson3210:
Adapted the recipe to allow for window paths also in the moveTo option (and folder option)

PR #5378, opened 2025-05-05
RPC-based JavaScript parser

knutwannheden:
TODOs:
- [ ] Decide how to install JS Node dependencies. As the OpenRewrite JS package has its own dependencies (e.g. TS compiler), I think it would make most sense to create a `package.json` file (extract from classpath or create on the fly) and let Node download any dependencies into a `node_modules` dir and then run from there.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/parser.ts
@@ -29,7 +29,6 @@ import {
 import {Expression, isJavaScript, JS, TypedTree, TypeTree} from '.';
 import {
     emptyMarkers,
-    ExecutionContext,
     markers,
     Markers,
     MarkersKind,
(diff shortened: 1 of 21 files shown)


rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -55,15 +57,17 @@ import {
...
 export class JavaScriptParser extends Parser {
 
     private readonly compilerOptions: ts.CompilerOptions;
     private readonly sourceFileCache: Map<string, ts.SourceFile> = new Map();
     private oldProgram?: ts.Program;
 
-    constructor() {
-        super();
+    constructor(ctx: ExecutionContext = new ExecutionContext(),
+                readonly relativeTo?: string) {
```
knutwannheden, line 69: I don't really like this part. I think it would probably be better if the `Parser` class defined the constructor as something like:

image

I think that should allow the `JavaScriptParser` to extend that `ParserOptions` type for its own options.

PR #5380, opened 2025-05-05
Support creation of a new guard when no guard was present

### LINKED ISSUE #5379: JCase where guard is not present has invalid syntax after calling WithGuard
I am using `latest.release` of gradle plugin

gradle plugin to run the unit test/write recipe

Call `withGuard` on the second `JCase`
```java
class Test {
    void guardedCases(Object o) {
        switch (o) {
            case Integer i when i > 0 -> System.out.println("Perfect");
            case String s -> System.out.println("Great");
            default -> System.out.println("Ok");
        }
    }
}
```

```java
case String s when "YES".equalsIgnoreCase(s) -> System.out.println("Great");
```

## What did you see instead?
No space between the labels and the newly added guard
```java
case String swhen "YES".equalsIgnoreCase(s) -> System.out.println("Great");
```

`getLabels` returns a `JContainer` of `JRightPadded` labels. When a guard is present, the last rightpadded contains a space/newline/... to print before the when. When the original case has no JRightPadded space (as the lambda has the space?) and we add a guard, the when is printed immediatly after. 
We could also change this in the Printer itself -> print " when" if the labels last expression had no/empty space.

branch on the way

Jenson3210:
## What's changed?
When adding a new guard using `withGuard` the printer would print labels, followed by "when" followed by the guard.
When no guard was present yet, the rightpadded labels would not have a space. 
the "when" keyword was immediately after the label printed. 

## What's your motivation?
This is incorrect syntax that is printed.

This will: 
- Fix #5379 

Adapt it in the JavaPrinter$492.

## Any additional context
I went for a default space if nothing is present. We can also go for newline/tab but think as a default space makes sense.

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -1113,6 +1113,23 @@ public Case(UUID id, Space prefix, Markers markers, Type type, @Deprecated @Null
...
+            }
+            if (this.guard == null) {
+                List<JRightPadded<J>> caseLabels = ListUtils.mapLast(getPadding().getCaseLabels().getPadding().getElements(), last -> {
+                    if (last != null && Space.EMPTY.equals(last.getAfter())) {
+                        return last.withAfter(Space.SINGLE_SPACE);
+                    }
+                    return last;
+                });
+
+                return new Case(this.padding, this.id, this.prefix, this.markers, this.type, JContainer.build(this.caseLabels.getBefore(), caseLabels, this.caseLabels.getMarkers()), this.statements, this.body, guard);
```
knutwannheden, line 1128: Here you would typically just use `JContainer#withElements()`:

```suggestion
                return new Case(this.padding, this.id, this.prefix, this.markers, this.type, this.caseLabels.withElements(caseLabels), this.statements, this.body, guard);
```

But actually, this type of problem we tend to solve by adding logic to the formatter. In this case specifically to `MinimumViableSpacingVisitor`, as this space is required to make the code compile.

knutwannheden: Also note that `JavaTemplate` should automatically apply the formatter, so the recipe shouldn't have to do that itself.

Jenson3210: Thanks for the feedback, @knutwannheden !
I adapted the logic to be in the `MinimumViableSpacingVisitor` but it seems that it is not automatically being called. 
I am not using a JavaTemplate here, so that might be it (not for the guard at least, only for the condition). The test added in this PR fails when using the `MinimumViableSpacingVisitor` unless I specifically call it in my test (see commit just made.

Jenson3210:
had a discussion with @knutwannheden and adapted the test accordingly. 
We call autoFormat to call the `MinimumViableSpacingVisitor` in which I added the spacing check. 
We do not put this in the `withGuard(...)` method as that is not how we tackled these issues in the past.

timtebeek (approved):
Thanks for noticing and fixing! I've applied some light polish to help me understand and not ignore type validation issues. Hope this helps you downstream.

PR #5382, opened 2025-05-05
Not visiting Javadoc contents in JavaTemplate.apply

greg-at-moderne:
## What's changed?

Changing `JavaTemplate.apply` not to go into Method Invocations within Javadoc comments contents.

## What's your motivation?

Prevent recipe run failures like:
```
java.lang.ClassCastException: class org.openrewrite.java.tree.Javadoc$Reference cannot be cast to class org.openrewrite.java.tree.J (org.openrewrite.java.tree.Javadoc$Reference and org.openrewrite.java.tree.J are in unnamed module of loader 'app')
  org.openrewrite.java.internal.template.JavaTemplateParser.isStatement(JavaTemplateParser.java:219)
  org.openrewrite.java.internal.template.JavaTemplateParser.parseMethodArguments(JavaTemplateParser.java:204)
  org.openrewrite.java.internal.template.JavaTemplateJavaExtension$1.visitMethodInvocation(JavaTemplateJavaExtension.java:392)
  org.openrewrite.java.internal.template.JavaTemplateJavaExtension$1.visitMethodInvocation(JavaTemplateJavaExtension.java:55)
  org.openrewrite.java.tree.J$MethodInvocation.acceptJava(J.java:4079)
```
without manually handling this in recipe code like we did in:
- https://github.com/openrewrite/rewrite-spring/pull/703
- or https://github.com/openrewrite/rewrite-static-analysis/pull/527


rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -106,6 +107,9 @@ protected JavaTemplate(String code, Set<String> genericTypes, Consumer<String> o
     @Override
     @SuppressWarnings("unchecked")
     public <J2 extends J> J2 apply(Cursor scope, JavaCoordinates coordinates, Object... parameters) {
+        if (scope.firstEnclosing(Javadoc.DocComment.class) != null) {
```
knutwannheden, line 110: Let's try to make it more specific to method references, as certain substitutions in Javadoc, like types should work and this change might cause test failures.

knutwannheden: I wonder why we didn't get test failures. I am quite sure we support simpler updates in Javadoc. Maybe no tests?

So I suggest that we place the check in that other visitor which gets called a few lines further down.

greg-at-moderne: 60178127f9134c20efd23140fead9c379552d174

PR #5384, opened 2025-05-05
Add expression type in JavaTemplate

fmodesto:
## What's changed?
Added support for explicit setting `expressionType` in `JavaTemplate`
- Enables better matching in templates
- Incidentally adds support for MemberReference and Lambda expressions in context-free templates

### Have you considered any alternatives or workarounds?
It is possible to write code that doesn't require setting the expression type.
- Omitting diamond operators (<>)
- Specifying generic types in method invocations

However, this approach presents challenges:
- Difficult to implement this type inference in rewrite-templating
- Or requires users to understand non-standard workarounds

### Checklist


timtebeek:
Thanks again! Looking at the overlap in changes I think this PR should go first, right?
- https://github.com/openrewrite/rewrite/pull/5420

fmodesto:
> Thanks again! Looking at the overlap in changes I think this PR should go first, right?
> 
> * [Support infer mode #5420](https://github.com/openrewrite/rewrite/pull/5420)

Yes, that should go first. It adds some of the functionality removed in https://github.com/openrewrite/rewrite/pull/5394.
This one should add expression type only.

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
```diff
@@ -908,4 +910,639 @@ public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, Object o) {
...
+        );
+    }
+    abstract class CompT<T extends CompT<T>> implements Comparable<T> {}
+    abstract class ExtT<T> extends CompT<ExtT<T>> {}
+    @Test
+    void recursiveTypes() {
+        CompT<?> compT;
+        CompT<ExtT<Integer>> compExtT;
+        ExtT<Integer> extT = new ExtT<Integer>() {
+            @Override
```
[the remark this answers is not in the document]
timtebeek: Hmm; the bot is off; the actual patch is this one:
```diff
diff --git a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
index b78bdc1..c55eb05 100644
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite.java.tree;
 
-import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.InMemoryExecutionContext;
@@ -1432,9 +1431,9 @@ abstract class ExtT<T> extends CompT<ExtT<T>> {}
     void recursiveTypes() {
         CompT<?> compT;
         CompT<ExtT<Integer>> compExtT;
-        ExtT<Integer> extT = new ExtT<Integer>() {
+        ExtT<Integer> extT = new ExtT<>() {
             @Override
-            public int compareTo(@NotNull TypeUtilsTest.ExtT<Integer> o) {
+            public int compareTo(TypeUtilsTest.ExtT<Integer> o) {
                 return 0;
             }
         };
```

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
```diff
@@ -908,4 +910,639 @@ public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, Object o) {
...
+            public int compareTo(@NotNull TypeUtilsTest.ExtT<Integer> o) {
```
timtebeek, line 1437:
```suggestion
            public int compareTo(TypeUtilsTest.ExtT<Integer> o) {
```

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.tree;
 
+import org.jetbrains.annotations.NotNull;
```
timtebeek, line 18:
```suggestion
```

rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -974,4 +1186,54 @@ public static String toGenericTypeString(JavaType.GenericTypeVariable type) {
             return type.getName() + " extends " + bounds;
         }
     }
+
+    public static class TypeMode {
```
knutwannheden, line 1190: I am wondering how we can improve the naming here of `TypeMode`, `TypeMode.Mode`, `TypeMode#get()` and possibly also the `push()` and `contains()` methods. I think that might help when trying to understand the code. Without having thought much about it, should we rename `TypeMode.Mode` to `TypeMode.Direction` and then possibly also `get()` to `direction()`? Or what ideas do you have?

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateMatchTest.java
```diff
@@ -1046,4 +1046,62 @@ Predicate<Object> test() {
           )
         );
     }
+
+    @Test
+    void matchMemberReferenceAndLambda() {
```
knutwannheden, line 1051: I would still like to see a test which verifies that we can also use `expressionType()` for `JavaTemplate#apply()` to e.g. replace a lambda (or anonymous class) with a method reference as an example.

rewrite-java/src/main/java/org/openrewrite/java/JavaTemplate.java
```diff
@@ -212,6 +213,25 @@ public Builder contextSensitive() {
...
+         * <p>
+         * Common examples include:
+         * <ul>
+         *   <li>{@code new ArrayList<>()}</li>
+         *   <li>{@code Collections.emptyList()}</li>
+         *   <li>{@code String::valueOf}</li>
+         * </ul>
+         * In such cases, the type must be specified manually.
+         */
+        public Builder expressionType(String expressionType) {
```
knutwannheden, line 227: I also wonder if any of these names would be better: `expectedType()`, `bindType()`, or `typeConstraint()`?

rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
fmodesto: Yes, naming things is hard 😅. I've pushed the changes to the other branch: https://github.com/openrewrite/rewrite/pull/5420

I hope the new names make things a bit clearer.

The ComparisonContext has two main responsibilities:

- It tracks the current comparisons using a stack. Since parameterized types can be recursive, it checks whether a type pair is already being compared to avoid infinite loops.

- It tracks which side (if any) we're inferring types from during the comparison. Normally, the inference side is TO, but in some cases it might flip to FROM—for example, in `Comparator<? super Comparable<T>>`, the `super` bound causes the sides to be swapped.

knutwannheden: Yes, naming is one of the 2 hard problems 😄 Let's then also add a test to make sure it works correctly for more complicated cases like `Comparator<? super Comparable<List<? extends T>>>`.

knutwannheden: Note: I merged the other PR to make it a bit easier to continue with this.

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateGenericsTest.java
```diff
@@ -206,4 +289,118 @@ Stream<Integer> test() {
           )
         );
     }
+
+    @Test
+    @ExpectedToFail
```
fmodesto, line 294: This is failing because `T::toString`, matches correctly the memberReference but then when visiting the containing `                visit(memberRef.getContaining(), compareTo.getContaining());` is visiting two identifiers T and Object and it fails to match equal names.

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateGenericsTest.java
```diff
@@ -206,4 +289,118 @@ Stream<Integer> test() {
...
+                  void test(Function<Object, String> fn) {
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    @ExpectedToFail
```
fmodesto, line 351: This one it's more to showcase a scenario we are not handling, even if the apply had arguments, T is not really a parameter, and `#{type:any(T)}` would not match a JavaType

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateMatchTest.java
fmodesto: Instead of match I have use apply with the other template

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateGenericsTest.java
knutwannheden: If we can add an example without generics that will already be useful as a test.

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateGenericsTest.java
fmodesto: We could have some other kind of argument. `#{type(T)}` where instead of using `__P__.<T>any()` we could add the `TypeUtils.toString()` that would also make the test above pass, being an argument the containing would match.

fmodesto: I have try this templates in error-prone refaster and they work. But I haven't found any use on the Picnic tests.
```java
  static final class RefToLambda<T> {
    @BeforeTemplate
    Function<T, String> before() {
      return T::toString;
    }

    @AfterTemplate
    Function<T, String> after() {
      return e -> e.toString();
    }
  }

  static final class LambdaToRef<T> {
    @BeforeTemplate
    Function<T, String> before() {
      return e -> e.toString();
    }

    @AfterTemplate
    Function<T, String> after() {
      return T::toString;
    }
  }

---

  ImmutableSet<Function<Object, String>> testRefToLambda() {
    return ImmutableSet.of(
        Object::toString);
  }

  ImmutableSet<Function<Object, String>> testLambdaToRef() {
    return ImmutableSet.of(
        e -> e.toString());
  }

```

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateGenericsTest.java
knutwannheden: I fixed the `SemanticallyEqualVisitor` and now the test passes.

PR #5394, opened 2025-05-07
Temporarily rollback TypeUtils enhancements

MBoegers:
## What's changed?
With the PR https://github.com/openrewrite/rewrite/pull/5348 a valuable contribution was made that caused some CI failures in a downstream project. The issues were partially resolved by the PRs https://github.com/openrewrite/rewrite/pull/5369 and https://github.com/openrewrite/rewrite/pull/5372.

In this PR the changes reverted and the new test cases are preserved.

See https://github.com/moderneinc/customer-requests/issues/1150 for plan and details.

## What's your motivation?

With the upcoming Code Remix Summit we decided to restore the original behavior to prevent downstream problems.
After the CRS we will readdress these enhancements as we think they are correct and of high value.

We have to double-check every downstream project to avoid CI failures.

@fmodesto supplied multiple fixes and also one for the current problems. We will use this PRs with him to merge the changes after CRS.

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java
```diff
@@ -929,7 +932,6 @@ public J.MemberReference visitMemberReference(J.MemberReference memberRef, J j)
                     return memberRef;
                 }
 
-                visit(memberRef.getContaining(), compareTo.getContaining());
                 visitList(memberRef.getTypeParameters(), compareTo.getTypeParameters());
```
timtebeek, line 935: Do we need to revert this as well? Or can we keep this part in?
```suggestion
                visit(memberRef.getContaining(), compareTo.getContaining());
                visitList(memberRef.getTypeParameters(), compareTo.getTypeParameters());
```

MBoegers: Thought about this line, it made no difference in the tests, and I decided to stick with the _brute force_ revert approach rather than backporting in this state.

timtebeek (approved):
Quick step back the release; hopefully reintroduced soon after with the latest fix suggestions.

PR #5396, opened 2025-05-07
Added Java parser tests for constructors, with some statements before super/this

JohannisK:
## What's changed?
Add a parser test for contructors, testing common constructors, but specifically adding tests for constructors with statement before calling `this(..)` or `super(..)`.

## What's your motivation?
Research for:
- https://github.com/moderneinc/customer-requests/issues/1144

### DIFF
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
@@ -17,6 +17,7 @@ package org.openrewrite.java.tree;
 
 import org.junit.jupiter.api.Test;
 import org.openrewrite.Issue;
+import org.openrewrite.java.MinimumJava21;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
@@ -94,6 +95,7 @@ class ConstructorTest implements RewriteTest {
     }
 
     @Issue("https://openjdk.org/jeps/513")
+    @MinimumJava21
     @Test
     void validationBeforeThisConstructor() {
         rewriteRun(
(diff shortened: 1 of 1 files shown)


rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
```diff
@@ -0,0 +1,171 @@
...
+                      stringA = a;
+                      this();
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/moderneinc/customer-requests/issues/1144")
```
timtebeek, line 154: Let's link to a public source link `https://openjdk.org/jeps/513#Constructors-are-too-restrictive` instead of a private issue.

JohannisK: Done

timtebeek:
Looks like we see failures on Java 8 and 11: https://ge.openrewrite.org/s/i2grybiw3pm7q/tests/overview?outcome=FAILED
Did you see these pass on say Java 21 ?

Wondering if we should add `@MinumumJava21` to the tests that use statements before super/this, as at most we'd want to check compatibility there pending a proper Java 25 parser.

JohannisK:
> Looks like we see failures on Java 8 and 11: https://ge.openrewrite.org/s/i2grybiw3pm7q/tests/overview?outcome=FAILED Did you see these pass on say Java 21 ?
> 
> Wondering if we should add `@MinumumJava21` to the tests that use statements before super/this, as at most we'd want to check compatibility there pending a proper Java 25 parser.

I think I've seen them pass on 17 also, but since this is only relevant for 25+ I added `@MinimumJava21`

timtebeek (approved):
Thanks! Great to know we don't have additional work to do here when it comes to supporting Java 25+

Issue #5334, opened 2025-04-23
PropertyPlaceholderHelper doesn't handle properties without separation


CasualSuperman:
## What version of OpenRewrite are you using?
I am using

- rewrite-core v8.50.2

## How are you running OpenRewrite?
I am writing a Java recipe.

## What is the smallest, simplest way to reproduce the problem?
Remove the space in the existing single PropertyPlaceholderHelperTest test case
```java
class PropertyPlaceholderHelperTest {
	@Test
	void noSpace() {
		var helper = new PropertyPlaceholderHelper("%%{", "}", null);
		var s = helper.replacePlaceholders("%%{k1}%%{k2}", k -> switch (k) {
			case "k1" -> "hi";
			case "k2" -> "jon";
			default -> throw new UnsupportedOperationException();
		});
		assertThat(s).isEqualTo("hijon");
	}
}
```

## What did you expect to see?
`"hijon"`

## What did you see instead?
`"hi%%{k2}"`

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
I am unfortunately unable to do so.

timtebeek:
Thanks for the helpful reproducer! Curious case indeed why that fails. Something to investigate further.

PR #5398, opened 2025-05-07
PropertyPlaceholderHelper Not Handling Lack of Separation Between Placeholders

steve-aom-elliott:
- #5334 

Previously not having any separation between placeholders when invoking `replacePlaceholders` on `PropertyPlaceholderHelper` would result in only the first placeholder being replaced.

## What's changed?
- Changed the calculation of the new start index after a replacement occurs to start searching at just the new end index  (was previously adding the size of the placeholder suffix to the new end index for the search start)

### Checklist
- [X] Figuring out why it fails to replace the second placeholder when there's no separation between placeholders (wrapped in their prefix and suffix)
- [X] I've added the unit test provided for reproduction of the bug
- [X] Added additional tests to cover other functionality present in the `PropertyPlaceholderHelper`, including nested placeholders and use of a non-null `valueSeparator` parameter.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/PropertyPlaceholderHelper.java
@@ -121,7 +121,7 @@ public class PropertyPlaceholderHelper {
                 }
 
                 // Proceed with unprocessed value.
-                startIndex = result.indexOf(placeholderPrefix, endIndex + placeholderSuffix.length());
+                startIndex = result.indexOf(placeholderPrefix, endIndex);
                 visitedPlaceholders.remove(originalPlaceholder);
             } else {
                 startIndex = -1;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see the tests you've added; gives us more confident with this particular change, thanks!

timtebeek:
Had a brief look at the parallel [PlaceholderParser.java#L62](https://github.com/spring-projects/spring-framework/blob/733e695e18919371a3e9d8c8bb82d6bddda6c26f/spring-core/src/main/java/org/springframework/util/PlaceholderParser.java#L62), but they seem to have diverged so much that we do not see the same handling there anymore.

steve-aom-elliott:
> Had a brief look at the parallel [PlaceholderParser.java#L62](https://github.com/spring-projects/spring-framework/blob/733e695e18919371a3e9d8c8bb82d6bddda6c26f/spring-core/src/main/java/org/springframework/util/PlaceholderParser.java#L62), but they seem to have diverged so much that we do not see the same handling there anymore.

So would the move be to adopt the implementation from `spring-framework` over what's in the `rewrite` repo (assuming it will need to be reviewed to ensure it doesn't affect the desired and already in use functionality of the one in the `rewrite` repo), or would they just remain two different implementations?

timtebeek:
Oh no it just means we're on our own supporting this implementation going forward; No need to borrow again, you've done well adjusting our implementation. Thanks!

PR #5399, opened 2025-05-07
Paving the core concepts for LockFile creation recipe

Jenson3210:
In order to support the changes made here, I am also preparing `withToolingApi` support. 
However, to add a clean unit test for this in the gradle-tooling module, I need to have the Assertions available. 

Furthermore, there is a bug I think in the PlainText. 
When the adapted for loop in RewriteTest executes for a plainText file, instead of updating the map, it adds the new one. 
This is because the equals contract is not the same as CompilationUnit's. 
adapted that also in this PR

Lastly, I made the flatMap method Function parameter adhere to the contract of the method which it is delegating to. 
If the flatmap function returns null, the entry will be deleted. 
However, the function does not support null, the bifunction which gets called does.

@sambsnyd

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
@@ -350,7 +350,7 @@ public final class ListUtils {
      * @return A new list with expanded or modified elements, or the original list if unchanged.
      */
     @Contract("null, _ -> null; !null, _ -> !null")
-    public static <T> @Nullable List<T> flatMap(@Nullable List<T> ls, Function<T, Object> flatMap) {
+    public static <T> @Nullable List<T> flatMap(@Nullable List<T> ls, Function<T, @Nullable Object> flatMap) {
         return flatMap(ls, (i, t) -> flatMap.apply(t));
     }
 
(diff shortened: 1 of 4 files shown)


timtebeek:
Copying failure here:
```
java.lang.AssertionError: Expected recipe to complete in 3 cycles, but took 2 cycles. This usually indicates the recipe is making changes after it should have stabilized.	
at org.openrewrite.test.LargeSourceSetCheckingExpectedCycles.afterCycle(LargeSourceSetCheckingExpectedCycles.java:121)	
at org.openrewrite.RecipeScheduler.runRecipeCycles(RecipeScheduler.java:99)	
at org.openrewrite.Recipe.run(Recipe.java:441)	
at org.openrewrite.config.DeclarativeRecipeTest.maxCyclesNested(DeclarativeRecipeTest.java:374)	
at java.lang.reflect.Method.invoke(Method.java:569)	
at java.util.ArrayList.forEach(ArrayList.java:1511)	
at java.util.ArrayList.forEach(ArrayList.java:1511)
```
https://github.com/openrewrite/rewrite/blob/f358c66b101c3673223f8b84948d264e9dc2c8e6/rewrite-core/src/test/java/org/openrewrite/config/DeclarativeRecipeTest.java#L356-L379

sambsnyd (approved):
There's a test failure but that looks wholly unrelated.
We should also update rewrite-gradle-tooling-model to write these lockfiles out to disk when `withToolingApi` is invoked during a test:
https://github.com/openrewrite/rewrite-gradle-tooling-model/blob/910908dc417c3d50900725b95055791ddf79c16d/model/src/main/java/org/openrewrite/gradle/toolingapi/Assertions.java#L66

Jenson3210:
Already working on that. That's how I found these bugs here. :)  https://github.com/openrewrite/rewrite-gradle-tooling-model/compare/main...gradle-lock-marker

Jenson3210:
The failure is caused by the equals annotations. Will check and fix!

timtebeek (approved):

PR #5401, opened 2025-05-08
Cleanup RemoveRedundantDependencyVersions recipe

jevanlingen:
## What's changed?
Nothing, just small refactoring.

## What's your motivation?
I wil make adjustments to this recipe in the future, it's better not to include this small refactoring in that recipe.

### Checklist


timtebeek (approved):
Thanks for doing this as a separate step; makes reviewing the changes to come a lot easier.

PR #5403, opened 2025-05-08
Update gradle lockfiles (step 1 of many)

Jenson3210:
## What's changed?
A new recipe was added to maintain lockfiles of gradle

## What's your motivation?
In order to build sucsefully for projects where a lockfile is present after all kind of dependency adding recipes, we need to be able to modify the lockfile post-recipe runs. This attempt is the **first** step of many in introducing this capability. 
With this recipe we can already override single versions by GA. 
Next steps also include:
- bom's
- transitive dependency version overrides
- introduce it in UpgradeTransitiveDependency

As this is the bases of many other recipe entrypoints, I tried to kept it rather "simple". Building the lockfile from scratch. 
For other recipes we will be able to introduce different concepts in the acc so that we can add dependencies in configurations etc... 

Now, I know it is probably not the best implementation atm, but I was really limited by the fact that I do not have access to the updated gradleproject post recipe run while visiting the lockfile. I could only come up with scanning recipe (hence I introduced it in UpgradeDependencyVersion first), but any feedback is welcomed! 

@sambsnyd @shanman190 @timtebeek 

Very open to suggestions! Having the framework write the lockfile during deserialization of sourcefiles would probably be an easier code, but opens us up to all kinds of these ocncepts leaking through the recipe stages. 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLockFile.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLockFile.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.gradle;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
+import org.openrewrite.gradle.marker.GradleProject;
+import org.openrewrite.groovy.tree.G;
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.kotlin.tree.K;
+import org.openrewrite.maven.tree.GroupArtifact;
+import org.openrewrite.maven.tree.GroupArtifactVersion;
+import org.openrewrite.text.PlainText;
+import org.openrewrite.text.PlainTextVisitor;
+
+import java.util.*;
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class UpdateDependencyLockFile extends ScanningRecipe<UpdateDependencyLockFile.Accumulator> {
+    private static final String[] EMPTY = new String[0];
+
+    @Override
+    public String getDisplayName() {
+        return "Update gradle dependency lock file";
+    }
+
+    @Override
+    public String getDescription() {
+        //language=markdown
+        return "Update the version of a dependency in a gradle lockfile.";
+    }
+
+    public static class Accumulator {
+        Map<String, GradleProject> modules = new HashMap<>();
+    }
+
+    @Override
+    public Accumulator getInitialValue(ExecutionContext ctx) {
+        return new Accumulator();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+        //noinspection BooleanMethodIsAlwaysInverted
+        return new JavaVisitor<ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                return (sourceFile instanceof G.CompilationUnit && sourceFile.getSourcePath().toString().endsWith(".gradle")) ||
+                        (sourceFile instanceof K.CompilationUnit && sourceFile.getSourcePath().toString().endsWith(".gradle.kts"));
+            }
+
+            @Override
+            public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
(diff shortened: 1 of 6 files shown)


shanman190 (commented):
Overall looks good.

With other recent changes, you should be able to just directly update the lockfile without needing to keep track of the related GradleProject markers -- since the lockfiles will have them directly attached -- which would simplify things down quite a bit.

Also not sure if we considered starting out with just enabling dependency locking as a first step over updating.

Just some thoughts. Nice work, @Jenson3210 !

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLockFile.java
```diff
@@ -0,0 +1,154 @@
...
+            if (versionOverrides.containsKey(gav.asGroupArtifact())) {
+                return gav.withVersion(versionOverrides.get(gav.asGroupArtifact()));
+            }
+            return gav;
+        }
+    }
+
+    @Override
+    public boolean isAcceptable(SourceFile sourceFile, ExecutionContext executionContext) {
+        return sourceFile instanceof PlainText;
```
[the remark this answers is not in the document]
Jenson3210: Thanks! I used to have more checks indeed. Well spotted!

Jenson3210:
@timtebeek tagging/requesting you here. 
This is the first step I mentioned for building/changing the lockfile. 
I have the changes locally ready for the transitive ones but do not want to push too many of these changes in 1 go to ease the "reviewer" allowing them to focus on small increments. 

The version locally already allows for transitive dependency upgrades (those that come transitively with the direct dependency, not the UpgradeTransitiveDependency recipe) + bom upgrades if necessary. 

But deliberately not pushing it here as we discussed

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLockFile.java
```diff
@@ -0,0 +1,198 @@
...
+import org.openrewrite.text.PlainText;
+import org.openrewrite.text.PlainTextVisitor;
+
+import java.util.*;
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class UpdateDependencyLockFile extends ScanningRecipe<UpdateDependencyLockFile.GradleProjectDependencyState> {
```
sambsnyd, line 40: Do we want a separate recipe for this?
When I imagine the venn diagram of situations I want a recipe updating the lockfile and situations where I'm running upgrade dependency version I'm picturing a circle.

Jenson3210: I do not consider this as cyclic. The recipe itself locks the gradle to the state of the marker in the lst. 
The visitor however is also used from the locations where you update the marker. The recipe is equivalent of locking for the first time (there's a `generate` still missing for this). 
I could remove the surrounding wrapper recipe leaving only the visitor? But then we can not "introduce" locking in a project that does not have it yet.

Jenson3210: @sambsnyd , I pushed some change that hopefully addresses your concern introducing a clear separation of concern between the 2 recipe and the visitor (different names, different accumulator, lock file is created with generate method...)
Let me know what you think about it!

shanman190:
> I tried... But for the recipes that adapt the GradleMarker, they only update it on the gradle file marker. So not sure if putting the marker on the lockfile also was a good idea. :/

So based on the tooling API and Gradle plugin edits, there would be a GradleProject marker on the lock files. With that said, if you observe the recipe logic carefully, you'll see that the SCAN phase determines a mapping of GA -> newVersion. Each of the EDIT recipes then take that information to perform their localized updates. So in the case of lockfiles, you'd take the GradleProject marker off of the lockfile itself, then take each GA and check if the versions had been modified and apply those changes to all of the first orders.


> For the usecase of creating a lockfile we would also not have a lockfile with marker to build one from.

This case is most definitely true. For that instance, you'd SCAN to find all of the GradleProject markers, then for each one found GENERATE lockfile entries within the projectDirs. This case would also involve an EDIT visitor modifying the build configuration to likely enable dependency locking and should be a separate recipe entirely.

Jenson3210:
>So based on the tooling API ... determines a mapping of GA -> newVersion. .... So in the case of lockfiles, you'd take the GradleProject marker off of the lockfile itself, then take each GA and check if the versions had been modified and apply those changes to all of the first orders.

Implemented that GA->newVersion cleaner. 
Goal is to in the next support for delete allow this map to contain null values and also new entries would result in added entries. 

>This case is most definitely true. For that instance, you'd SCAN to find all of the GradleProject markers, then for each one found GENERATE lockfile entries within the projectDirs. This case would also involve an EDIT visitor modifying the build configuration to likely enable dependency locking and should be a separate recipe entirely.

For nowmove the recipe to be a test-recipe for the visitor as also Sam had doubts there and we can reintroduce it when we have the gradle changes etc also. 

Thanks @sambsnyd and @shanman190 for your patience here!

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -224,6 +224,11 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
 
                         String versionVariableName = declaredVersion;
                         GroupArtifact ga = new GroupArtifact(declaredGroupId, declaredArtifactId);
+                        if (gradleProject != null) {
+                            acc.getConfigurationPerGAPerModule().computeIfAbsent(gradleProject.getPath(), k -> new HashMap<>())
```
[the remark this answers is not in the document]
Jenson3210: In order to succesfully/correctly update the marker we need the dependency and its configuration. 
If I use the gradleMarker present on the lockfile, I cannot access the Map's gradle configurations detected during scanning from the build.gradle files. 
I will for now create a specific key which takes both modulename AND group as key.
Perhaps we can then later transform something like this in the GradleProject itself. 
I do not think overriding  equals is justified in this case here to make the 2 markers equal to each other just when the key matches.

PR #5404, opened 2025-05-08
Add fallback when attempting to load `lombok.launch.ShadowClassLoader`

Laurens-W:
## What's changed?

Added a fallback when loading `lombok.launch.ShadowClassLoader` with the parserClassLoader fails

## What's your motivation?

Lombok integration seemingly not working with Maven

### Checklist

### DIFF
--- a/rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
+++ b/rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
@@ -59,8 +59,13 @@ public class LombokSupport {
         }
 
         String oldValue = System.setProperty("shadow.override.lombok", String.join(File.pathSeparator, overrideClasspath));
+        Class<?> shadowLoaderClass;
         try {
-            Class<?> shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader);
+            shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader);
+        } catch (Exception e) {
+            shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader.getParent());
+        }
+        try {
             Constructor<?> shadowLoaderConstructor = shadowLoaderClass.getDeclaredConstructor(
                     Class.forName("java.lang.ClassLoader"),
                     Class.forName("java.lang.String"),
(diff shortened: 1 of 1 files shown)


rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
```diff
@@ -59,8 +59,13 @@ public class LombokSupport {
         }
 
         String oldValue = System.setProperty("shadow.override.lombok", String.join(File.pathSeparator, overrideClasspath));
+        Class<?> shadowLoaderClass;
+        try {
+            shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader);
+        } catch (Exception e) {
+            shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader.getParent());
+        }
         try {
```
timtebeek, line 68: Why pull this out of the original `try`? Now it looks like it could happen that the finally below is not called to restore/clear the property.

rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
```diff
@@ -59,8 +59,13 @@ public class LombokSupport {
         }
 
         String oldValue = System.setProperty("shadow.override.lombok", String.join(File.pathSeparator, overrideClasspath));
+        Class<?> shadowLoaderClass;
+        try {
+            shadowLoaderClass = Class.forName("lombok.launch.ShadowClassLoader", true, parserClassLoader);
+        } catch (Exception e) {
```
jevanlingen, line 65: I think it's better to only catch:

```suggestion
        } catch (ClassNotFoundException e) {
```

rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
jevanlingen: On top of that, please add a comment why the fallback is there in the first place. Something like:
```java
// `rewrite-java-lombok` is shaded within the cli's `modmaven` module, thus is only visible in the parent class loader
```
will do.

Laurens-W: I realized yesterday the way I changed things isn't valid, I'll move trying to create the ShadowClassLoader back into the original try

I don't think I want to expose the CLI context in OpenRewrite, but I'll try to add a comment without that

rewrite-java-lombok/src/main/java/org/openrewrite/java/lombok/LombokSupport.java
Laurens-W: Invoking `Class.forName` can throw different kinds of exceptions, In this case I think catching all is okay as in any of these cases I think we want to retry

timtebeek (approved):

Issue #4876, opened 2025-01-09
RecipeRunStats source file count exceeds number of files in project


Bananeweizen:
## What is the smallest, simplest way to reproduce the problem?
Enable datatables on a small project, open RecipeRunStats.csv and sort by first column (edited source files).

## What did you see instead?
![image](https://github.com/user-attachments/assets/14ab5d82-0f3c-44c5-9fb7-8691bad38778)

That's from a project with 2600 files in the repo (Java, Maven, adoc, ....), therefore I would expect none of the numbers to be larger than 2600.

If you agree to that assumption, then https://github.com/openrewrite/rewrite/blob/d11eae939d91309a414e3028bd800457c84ef30c/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java#L122 seems to be wrong. This is called unconditionally inside 2 nested loops for recipes and files, thereby incrementing the counter by files*recipes or something like that.
It might need to be moved inside the `if visitor.isAcceptable()` condition.

This bug does not create any wrong code with rewrite recipes. It just makes it harder to use the numbers in the data tables to look for "extreme" recipes or similar oddities.

timtebeek:
Thanks for the analysis with examples here @Bananeweizen ! Agree that it looks off, and would be better if we ensure these numbers line up with the number of files, as opposed to number of files * number of recipes (?). I'll raise this to the team!

arodionov:
I see that only particular recipes produce the wrong `sourceFiles` stats 

In the test repo are 12056 files.
For `org.openrewrite.staticanalysis.CommonStaticAnalysis`, the stats were correct.
For `org.openrewrite.java.migrate.UpgradeToJava17` it doesn't:

<img width="560" alt="Image" src="https://github.com/user-attachments/assets/e9d315a2-6bc8-46d4-ac05-93f5e40a2257" />

timtebeek:
At a quick glance those all seem to be multiples of the number of files, where `AddJaxwsRuntime` likely only runs once, whereas the others are configurable and may thus be included more than once.

What are you proposing in terms of a fix?

arodionov:
Yes, it seems that if a recipe is run with different parameters, the number of files visited is summed up.

In terms of fix: 
- trying to create a unit test to fail (because initial unit tests passed successfully)
- trying to find the place where the multiplication happens

arodionov:
The counter issue is related to `recordEdit` method:
[...]

PR #5408, opened 2025-05-09
RecipeRunStats several times counts the same recipe runs with different params

arodionov:
When the same recipe is executed several times with different parameters, the `tag("name", recipe.getName())` will be the same, and it will produce wrong (n-times bigger) stats for `source-files` ("The number of source files the recipe ran over") in  `RecipeRunStats`.

As a fix, a HashSet was introduced to track all visited sources.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -174,66 +174,67 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
         // skip edits made to generated source files so that they don't show up in a diff
         // that later fails to apply on a freshly cloned repository
         // consider any recipes adding new messages as a changing recipe (which can request another cycle)
-        return sourceSetEditor.apply(sourceSet, sourceFile ->
-                allRecipeStack.reduce(sourceSet, recipe, ctx, (source, recipeStack) -> {
-                    Recipe recipe = recipeStack.peek();
-                    if (source == null) {
-                        return null;
-                    }
-
-                    recipeRunStats.recordSourceVisited(source);
-                    SourceFile after = source;
-
-                    try {
-                        Duration duration = Duration.ofNanos(System.nanoTime() - cycleStartTime);
-                        if (duration.compareTo(ctx.getMessage(ExecutionContext.RUN_TIMEOUT, Duration.ofMinutes(4))) > 0) {
-                            if (thrownErrorOnTimeout.compareAndSet(false, true)) {
-                                RecipeTimeoutException t = new RecipeTimeoutException(recipe);
-                                ctx.getOnError().accept(t);
-                                ctx.getOnTimeout().accept(t, ctx);
-                            }
-                            return source;
+        return sourceSetEditor.apply(sourceSet, sourceFile -> {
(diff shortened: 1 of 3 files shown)


arodionov:
Checked a new version on the same repo https://github.com/openrewrite/rewrite/issues/4876#issuecomment-2858443748
The `org.openrewrite.staticanalysis.CommonStaticAnalysis` and `org.openrewrite.java.migrate.UpgradeToJava17` have the same numbers of `sourceFiles` in a RecipeRunStats table.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -181,6 +181,7 @@ public LSS editSources(LSS sourceSet) {
                         return null;
                     }
 
+                    recipeRunStats.recordSourceVisited(source);
```
timtebeek, line 184: Perhaps good to get input on the approach first before we continue, to make changes, but pointing out a detail:

By calling this method here we're executing this not once for every file, but for every recipe that handles a file (so again a multiplication). Perhaps it's better called not for every items on the recipeStack, but between lines 177 and 178 for every sourceFile processed there.

jkschneider:
I'm not clear on whether this table is shared across repositories and whether then it will undercount source file paths that reoccur across repos.

arodionov:
@jkschneider  I've tested the `mod-cli` with the fix on the Apache organization. The new version gives correct and consistent results for `org.openrewrite.staticanalysis.CommonStaticAnalysis` and `org.openrewrite.java.migrate.UpgradeToJava17`

okundzich:
@arodionov would you check on what Jonathan asked about - 

> whether this table is shared across repositories and whether then it will undercount source file paths that reoccur across repos

arodionov:
@okundzich I thought that running recipes across an organization with multiple repositories should check this. 
But as I understand, checking this via `mod-cli` is not enough, and it should be checked in a Saas platform.

okundzich:
please make sure you have same file names across repos while testing

arodionov:
After the source code analysis for `mod-cli` and `moderne-worker` can conclude that `RecipeRunStats` (which contains the initial stats) is not shared across repositories.

For `mod-cli`, the `RecipeRunStats` initialised per `sourceSet`, per `recipe` as 
`new RecipeScheduler().scheduleRun(recipe, lss, ctx, 1, 1)` 

For `moderne-worker`, the `RecipeRunStats` initialised as a part of `new CheckpointingRecipeRun.StandardDataTables()`, during the `new StoredRecipeRun(...)` constructor call:
https://github.com/moderneinc/moderne-worker/blob/550b08b827213533b710d546628405e8bba5f523/src/main/java/io/moderne/worker/execution/StoredRecipeRun.java#L590-L604 
The `new StoredRecipeRun(...)` is also calling per repository

timtebeek (approved):
Thanks for confirming, and the helpful link into moderne-worker. With that I'm leaning towards a merge: we're copying a pattern we already used for `sourceFileChanged` to add a `sourceFileVisited`, with both the CLI and worker only using those per repository from what we've seen here. Nice to see this fixed.

We can of course wait a day or two before we merge in case Jonathan or Olga has any concerns (there's a US public holiday), as there's no rush getting this out.

PR #5409, opened 2025-05-09
RemoveRedundantDependencyVersions recipe removes direct dependencies if dependency is pulled in transitively

jevanlingen:
## What's changed?
Redundant direct dependencies are removed as well.

Interestingly, the recipe did remove direct dependencies with a lower version number already. This was actually a side effect by removing redundant constraints. I just added an extra check to see if the version 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
@@ -38,6 +38,7 @@ import org.openrewrite.maven.MavenDownloadingException;
 import org.openrewrite.maven.internal.MavenPomDownloader;
 import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedDependency;
+import org.openrewrite.maven.tree.ResolvedManagedDependency;
 import org.openrewrite.maven.tree.ResolvedPom;
 import org.openrewrite.semver.ExactVersion;
 import org.openrewrite.semver.LatestIntegration;
@@ -96,7 +97,7 @@ public class RemoveRedundantDependencyVersions extends Recipe {
 
     @Override
     public String getDescription() {
-        return "Remove explicitly-specified dependency versions that are managed by a Gradle `platform`/`enforcedPlatform`.";
+        return "Remove explicitly-specified dependencies and dependency versions that are managed by a Gradle `platform`/`enforcedPlatform`.";
     }
 
     private static final MethodMatcher CONSTRAINTS_MATCHER = new MethodMatcher("DependencyHandlerSpec constraints(..)");
@@ -108,6 +109,7 @@ public class RemoveRedundantDependencyVersions extends Recipe {
         return Preconditions.check(new IsBuildGradle<>(), new GroovyIsoVisitor<ExecutionContext>() {
                     GradleProject gp;
                     final Map<String, List<ResolvedPom>> platforms = new HashMap<>();
+                    final List<ResolvedDependency> directDependencies = new ArrayList<>();
 
                     @Override
                     public G.CompilationUnit visitCompilationUnit(G.CompilationUnit cu, ExecutionContext ctx) {
@@ -124,6 +126,13 @@ public class RemoveRedundantDependencyVersions extends Recipe {
                             @Override
                             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                                 J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+
+                                new GradleDependency.Matcher()
+                                        .groupId(groupPattern)
+                                        .artifactId(artifactPattern)
+                                        .get(getCursor())
+                                        .ifPresent(it -> directDependencies.add(it.getResolvedDependency()));
+                                
                                 if (!platformMatcher.matches(m) && !enforcedPlatformMatcher.matches(m)) {
                                     return m;
                                 }
@@ -195,7 +204,7 @@ public class RemoveRedundantDependencyVersions extends Recipe {
                                         return m;
                                     }
                                     //noinspection DataFlowIssue
-                                    if (shouldRemoveRedundantConstraint((String) ((J.Literal) m.getArguments().get(0)).getValue(), m.getSimpleName())) {
+                                    if (shouldRemoveRedundantConstraintOrImplementation((String) ((J.Literal) m.getArguments().get(0)).getValue(), m.getSimpleName())) {
                                         return null;
                                     }
                                 }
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -211,33 +220,69 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                            private boolean isPulledInExactlyTransitively(Dependency dep) {
+                                if (dep.getVersion() == null || !platforms.containsKey("implementation") || platforms.get("implementation").isEmpty()) {
+                                    return false;
+                                }
+
+                                // TODO support multiple platform tags
+                                ResolvedManagedDependency x = platforms.get("implementation").get(0).getDependencyManagement().stream()
+                                        .filter(it -> it.getGroupId().equals(dep.getGroupId()) && it.getArtifactId().equals(dep.getArtifactId()))
+                                        .findFirst()
+                                        .orElse(null);
```
shanman190, line 267: callout: remember that Gradle constraints from platforms are highest resolved version from any platforms that import them.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -96,7 +97,7 @@ public String getDisplayName() {
 
     @Override
     public String getDescription() {
-        return "Remove explicitly-specified dependency versions that are managed by a Gradle `platform`/`enforcedPlatform`.";
+        return "Remove explicitly-specified dependencies and dependency versions that are managed by a Gradle `platform`/`enforcedPlatform`.";
```
Jenson3210, line 100: The name of this recipe says it will remove versions, the description says it will remove dependencies also. This PR is for the second. But should we not have a better recipe name then?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -211,10 +233,40 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
-                            boolean shouldRemoveRedundantConstraint(String dependencyNotation, String configurationName) {
-                                return shouldRemoveRedundantConstraint(
-                                        DependencyStringNotationConverter.parse(dependencyNotation),
-                                        gp.getConfiguration(configurationName));
+                            private boolean shouldRemoveRedundantImplementation(Dependency dependency) {
+                                if(platforms == null || platforms.get("implementation") == null || platforms.get("implementation").isEmpty()) {
+                                    return false;
+                                }
+
+                                if ((groupPattern != null && !StringUtils.matchesGlob(dependency.getGroupId(), groupPattern)) ||
```
Jenson3210, line 241: both group and artifact are nullable but then this one will always return false so no adaptations will be made?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -211,10 +233,40 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                                    //ignore self
+                                    if (d.getGroupId().equals(dependency.getGroupId()) && d.getArtifactId().equals(dependency.getArtifactId())) {
+                                        continue;
+                                    }
+
+                                    if (d.getDependencies() == null) {
+                                        continue;
+                                    }
+
+                                    String dependencyManagedVersion = platforms.get("implementation")
```
Jenson3210, line 255: I do not see the loop variable used here so if this is intentional you could pull it up outside loop.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -211,10 +233,40 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                                            .stream()
+                                            .flatMap(e -> e.getDependencyManagement().stream())
+                                            .filter(e -> e.getGroupId().equals(dependency.getGroupId()))
+                                            .filter(e -> e.getArtifactId().equals(dependency.getArtifactId()))
+                                            .map(ResolvedManagedDependency::getVersion)
+                                            .max(VERSION_COMPARATOR)
+                                            .orElse(null);
+
+                                    ResolvedDependency resolvedDependency = d.findDependency(dependency.getGroupId(), dependency.getArtifactId());
```
Jenson3210, line 264: not sure if this findDependency is going to go transitive, to be checked
I thought, at this moment it does not yet.

Jenson3210: `shouldRemoveRedundantConstraint` removes if the constraint the transitive/direct version is higher than the constraint. 
I do not think we can do that. They might have a reason for keeping it pinned to a lower version. By removing the constraint it will become a higher version suddenly.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
JohannisK: No, the groupPattern and artifactPattern are options. If they are not set the recipe will operate on all dependencies. If they are set, and the current dependency does not match, no change will be made.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
JohannisK: Done

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
JohannisK: > not sure if this findDependency is going to go transitive, to be checked I thought, at this moment it does not (yet) as I was facing this for the gradle lockfiles. I had a similar need for resolvedDependencies resolution in GradleConfiguration's `findResolvedDependency`

I think it does. I did see instances where it found resolved dependencies

JohannisK: > `shouldRemoveRedundantConstraint` removes if the constraint the transitive/direct version is higher than the constraint. I do not think we can do that. They might have a reason for keeping it pinned to a lower version. By removing the constraint it will become a higher version suddenly.

I haven't changed this behavior, but you might be right. I will also try to find out it the behavior for implementations is correct.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
Jenson3210: ah, misread indeed

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
JohannisK: I think I addressed this by ordering the versions.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -211,10 +225,42 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                                 return r;
                             }
 
-                            boolean shouldRemoveRedundantConstraint(String dependencyNotation, String configurationName) {
-                                return shouldRemoveRedundantConstraint(
-                                        DependencyStringNotationConverter.parse(dependencyNotation),
-                                        gp.getConfiguration(configurationName));
+                            private boolean shouldRemoveRedundantImplementation(Dependency dependency) {
+                                if(platforms == null || platforms.get("implementation") == null || platforms.get("implementation").isEmpty()) {
```
shanman190, line 229: If we pass in the configuration for the current dependency, then we can be able to perform redundant dependency removal across all configurations.

As an example, you might put the junit-bom in testImplementation which then wouldn't be in scope with the current logic.

shanman190 (commented):
optional: should we also consider moving from Groovy visitors to the more general Java visitors in order to bring Kotlin DSL support to this recipe? There doesn't appear to be any specific functionality used, so it should be a pretty straight forward class swap.

sambsnyd:
The [Maven version](https://github.com/openrewrite/rewrite/blob/9d7fe7709a41e66816033b29786deab0173b3091/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java#L56-L74) of this recipe has controls to allow the user to specify when to remove. If we have the same options on the gradle recipe we can group them together into a single remove-redundants in rewrite-java-dependencies

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
JohannisK: Added support for all dependency methods.

PR #5412, opened 2025-05-09
Mark methods overridden from classes as `isOverride()`

timtebeek:
Quite surprised we didn't mark these before; we only looked at interfaces. it seems.


Jenson3210 (approved):
Great to see that it can be fixed with few lines of code!

PR #5418, opened 2025-05-13
Add ModuleHasDependency recipes to add as preconditions to recipes

kmccarp:


rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasDependency.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                assert tree != null;
+                tree.getMarkers()
+                        .findFirst(JavaProject.class)
+                        .ifPresent(jp -> {
+                            Tree t = new FindDependency(groupId, artifactId, version, versionPattern).getVisitor().visit(tree, ctx);
```
shanman190, line 92: Did you possibly want to use `maven.DependencyInsight` here like was done with the Gradle variation?

sambsnyd:
I checked this out locally to investigate the failing tests, but everything passed. Not sure why there are some CI failures

rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasDependency.java
kmccarp: Good call, will update.

kmccarp: Updated!

kmccarp:
> I checked this out locally to investigate the failing tests, but everything passed. Not sure why there are some CI failures

I got a ConcurrentModificationException in `FindPluginTest` but a rerun cleared it up; we may want to investigate that at some point.

PR #5422, opened 2025-05-14
mapAsync to enforce order of execution

greg-at-moderne:
## What's changed?

Changing `mapAsync` implementation to one which guarantees order of execution.

## What's your motivation?

Otherwise the Javascript visitors get invalid `cursor` data when traversing an LST. As `mapAsync` is heavily used in visitors.

The current code of `mapAsync` is not as nicely functional as it used to be. I don't know how to write it in a cleaner way.

### DIFF
--- a/rewrite-javascript/rewrite/src/util.ts
+++ b/rewrite-javascript/rewrite/src/util.ts
@@ -15,8 +15,8 @@
  */
 export async function mapAsync<T, U>(arr: T[], fn: (t: T, i: number) => Promise<U | undefined>): Promise<U[]> {
     const results: U[] = [];
-    for await (const [i, item] of arr.entries()) {
-        const result = await fn(item, i);
+    for (let i = 0; i < arr.length; i++) {
+        const result = await fn(arr[i], i);
         if (result !== undefined) {
             results.push(result);
         }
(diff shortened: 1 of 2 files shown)


knutwannheden:
I thought that Promise.all() does guarantee the order. Can we verify this?

knutwannheden:
Right. Now I understand what you mean.

rewrite-javascript/rewrite/src/util.ts
```diff
@@ -14,5 +14,12 @@
...
  */
 export async function mapAsync<T, U>(arr: T[], fn: (t: T, i: number) => Promise<U | undefined>): Promise<U[]> {
-    return (await Promise.all(arr.map(fn))).filter(v => v !== undefined);
+    const results: U[] = [];
+    for await (const [i, item] of arr.entries()) {
+        const result = await fn(item, i);
+        if (result !== undefined) {
+            results.push(result);
+        }
+    }
```
knutwannheden, line 23: We don't have an async iterable here, so there is no need for `for async`.
```suggestion
    for (let i = 0; i < arr.length; i++) {
        const result = await fn(arr[i], i);
        if (result !== undefined) {
            results.push(result);
        }
    }
```

PR #5425, opened 2025-05-14
Update transitive dependencies and platforms in a gradle lockfile

Jenson3210:
## What's changed?
Adopted the lockfiles recipe to also update the transitive dependencies and the platforms

## What's your motivation?
In order to build successefully for projects where a lockfile is present after all kind of dependency adding recipes, we need to be able to modify the lockfile post-recipe runs. This attempt is the **second** step of many in introducing this capability. 
With this recipe we can already override versions by UpgradeDependencyVersion:
- bom's
- transitive dependency version overrides

@sambsnyd, @shanman190 or @timtebeek 

Very open to suggestions! 

### Checklist


Jenson3210:
I just pushed the last change that I think we can possible make (with current knowledge) after discussions with @shanman190 but I will leave it up to @shanman190 to decide if this is sufficient to resolve the draft status.

sambsnyd (approved):

PR #5429, opened 2025-05-15
Add support for partially unfolding single line yaml properties with exclusions in UnfoldProperties

JohannisK:
## What's changed?
When breaking single line declarations in a yaml file into parts we reiterate over sub-declaration when we find matches.

## What's your motivation?
The recipe only matched entire keys.

When matching 
`a.b.c.d` agains `$..[a.b][c.d]` the recipe now takes in account that when $..[a.b] is matched we then need to delve deeper matching `c.d` against [c.d] to actually find the exclusion for a single line key.

The solution does introduce some more recursion, but since this recipe only target yaml properties line by line I don't believe it will be a problem.


Jenson3210 (approved):
As discussed offline we should find a better way of matching but for this "bug" fix we can go ahead imo

PR #5433, opened 2025-05-15
Do not `AddDependency` for transitive dependency managed by a bom, even if `onlyIfUsing == null`

mrakshith21:
## What's changed?
Added a unit test for `AddDependency` recipe to not add a transitive dependency when managed by a bom, with `acceptTransitive` set to true. This test is failing.
In the below test spring-boot-starter-web is managed by spring-boot-dependencies and transitively uses spring-webmvc, so the recipe should not add it.

@timtebeek. I am not sure if any other argument to the recipe must be used for this case, so let me know. 

No, but do let me know if this can be done in any other way.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
@@ -279,6 +279,6 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
     }
 
     private boolean hasAcceptableTransitivity(ResolvedDependency d, Scanned acc) {
-        return d.isDirect() || Boolean.TRUE.equals(acceptTransitive) && !acc.scopeByProject.isEmpty();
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the clear & runnable example @mrakshith21 ! Indeed would not expect this dependency to be added, as the dependency already comes in transitively:
https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-web/2.7.18/spring-boot-starter-web-2.7.18.pom
image

mrakshith21:
@timtebeek I have added a fix in `hasAcceptableTransitivity()` to ignore the check when `onlyIfUsing` is null. It looks to me as if acceptTransitive only worked when java source files used the `onlyIfUsing` parameter, hence the change.

Let me know if this is okay.

timtebeek (approved):
That looks perfect, yes thanks @mrakshith21 ! Good to see you worked that out into this fix. Should help prevent unwanted dependencies downstream.

PR #5438, opened 2025-05-16
Added support for index visitation and all matching elements when filtering

Jenson3210:
## What's changed?

We had this issue pop up during a connect. 

## What's your motivation?
We still won;t support full usecases of jsonpath -> opened another ticket to track a more uniform solving of these issues.

### Checklist


sambsnyd:
Nice work @Jenson3210 !

Issue #5435, opened 2025-05-15
Managing imports post code trasformation.


amishra-u:
## What problem are you trying to solve?
Currently, recipe authors must manually invoke AddImport or RemoveImport visitors based on the specific code transformations they apply. This process is verbose, especially since import changes can be inferred from the resulting type or member references in the transformed code. 

Moreover, AddImport may add ambiguous imports, and there's no easy way fix to prevent in all cases. 

## Describe the solution you'd like
Introduce a FixImport visitor that can be invoked after the main code transformation. This visitor would automatically:

1.  Add any required imports,
2. Remove unused imports,
3. Convert references to fully qualified names when an import would result in ambiguity.

Optionally, the FixImport visitor could accept an import layout configuration to maintain consistent formatting.

#### Before
{
   someTransformation()
   maybeAddImport("foo")
   maybeAddImport("bar")
   maybeRemoveImport("baz")
}

#### After 
{
   someTransformation()
   doAfterVisit(new FixImport(/*optional import layout*/))
}

We can also create recipe on top of FixImport Visitor which anyone can add to declarative recipe to fix the erroneous imports caused by other recipes.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
Yes

timtebeek:
Thanks for the suggestion and offer to help @amishra-u ! It's indeed a common sight to have to wrangle explicit imports even when using JavaTemplate to make changes. I suppose we could indeed deduce which types where dropped by diffing the replaced element tree, but I wonder if we're able to do so efficiently enough to forgo the explicit import handling.

We'd also have to account for any missing types, which unfortunately haven't been completely phased out despite your help to close the gap when related to Lombok. These are a particular concern when it comes to removing imports, as we still see missing imports when frameworks like Mockito are involved.

Not quite sure if we'd be able to cover all those cases before fully switching over, and it's a bit of a high bar for us to have two ways to do things. Any thoughts here @knutwannheden ?

PR #5439, opened 2025-05-16
AddImport: Handle import ambiguity

amishra-u:
## What's changed?
Previously, the AddImport logic did not account for import ambiguity, leading to compilation errors when ambiguous imports were added. This update introduces a check for ambiguity before adding an import. If ambiguity is detected, the code now fully qualifies the relevant identifiers instead of adding the import

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java
@@ -223,19 +223,19 @@ class AddImportTest implements RewriteTest {
           java(
             """
               package com.acme.bank;
-              
+
               import com.acme.bank.*;
-              
+
               class Foo {
               }
               """,
             """
               package com.acme.bank;
-              
+
               import com.acme.bank.*;
-              
+
               import com.acme.bank.Record;
-              
+
               class Foo {
               }
               """,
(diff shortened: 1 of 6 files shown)


sambsnyd:
Nice work @amishra-u !

knutwannheden:
Personally, I prefer doing this the other way around by using fully qualified names in the template and then apply the ShortenFullyQualifiedTypeReferences visitor to add imports where possible.

amishra-u:
@knutwannheden Just to clarify—are you suggesting that we update all existing recipes to use fully qualified names in the template, and then rely on the ShortenFullyQualifiedTypeReferences visitor to handle the imports?

amishra-u:
@sambsnyd @knutwannheden When you have a moment, could you please share your thoughts on this?
https://github.com/openrewrite/rewrite/issues/5435

knutwannheden:
@amishra-u No, I think it makes sense for AddImport to have this extra safety, as it otherwise ends up braking code. This way we can keep both alternatives in place and let people use either one.

timtebeek:
We seem to be having some downstream tests fail which now see FQN not shortened. Could that be cause by the changes here?
https://ge.openrewrite.org/s/p2qtwe6x6pszu rewrite-apache
https://ge.openrewrite.org/s/ovb4rqwsvm5f6 rewrite-testing-frameworks
https://ge.openrewrite.org/s/5k7lxq4ixeqre rewrite-migrate-java

amishra-u:
@timtebeek fixed it here
https://github.com/openrewrite/rewrite/pull/5452

also fixed all the test in rewrite-testing-frameworks. https://github.com/openrewrite/rewrite-testing-frameworks/pull/731

we may use the same approach for rewrite-apache and  rewrite-migrate-java

PR #5442, opened 2025-05-17
Only allow gradle dependency locking for already locked configurations/artifacts

Jenson3210:
## What's changed?
I used to completely write the lockfile based on the marker, but did miss that configurations can be ignored for locking. Even artifacts themselves can be ignored. 
This fix addresses to make sure we only put the lock if:
- The configuration was in the lockfile before
- The GroupArtifact was locked before

This way we will address issues like [this repo's lockfile](https://github.com/atlassian/aws-infrastructure/blob/master/gradle.lockfile) not being wrongfully updated to contain locks [for excluded configurations](https://github.com/atlassian/aws-infrastructure/blob/master/build.gradle.kts#L21-L23).

@shanman190 I think this is one quick-win yet big improvement to the "wrongfully updating the lockfile"
@sambsnyd Shannon reached out to mention we might have been to fast. Rather than rolling back, I hope we can fix-forward instead

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLock.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLock.java
@@ -24,14 +24,17 @@ import org.openrewrite.Tree;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
 import org.openrewrite.gradle.marker.GradleProject;
 import org.openrewrite.internal.StringUtils;
+import org.openrewrite.maven.tree.GroupArtifact;
 import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedDependency;
 import org.openrewrite.text.PlainText;
 import org.openrewrite.text.PlainTextVisitor;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.emptySortedSet;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLock.java
```diff
@@ -62,26 +65,32 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext executionCon
...
+            }
+        });
+
         GradleProject project = gradleProject.get();
         project.getConfigurations().stream()
                 .filter(GradleDependencyConfiguration::isCanBeResolved)
+                .filter(configuration -> lockedConfigurations.contains(configuration.getName()))
                 .forEach(conf -> {
                     List<ResolvedDependency> transitivelyResolvedDependencies = conf.getResolved();
+                    String configuration = conf.getName();
```
timtebeek, line 82: It looks like we only ever use the configuration name, but in one lambda used `GradleDependencyConfiguration configuration`, while in the next assigned `String configuration`. Figured make that consistent through these changes:
```suggestion
                .map(GradleDependencyConfiguration::getName)
                .filter(lockedConfigurations::contains)
                .forEach(configuration -> {
                    List<ResolvedDependency> transitivelyResolvedDependencies = conf.getResolved();
```

timtebeek (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateDependencyLock.java
Jenson3210: conf.getResolved() would not exist in this suggestion, would it?

Jenson3210: We need the dependencies in the configuration and its name.

timtebeek: Ah I missed that one line, thanks!

shanman190:
I think capturing the current locked configurations should better align us with Gradle.

Personally, I'd forgo using the currently locked artifacts as a filter for what new artifacts to lock. For simple dependency changes, it'd probably work. But for migrations such as Spring Boot 2.x to 3.x, the lockfile would be incorrect and would result in a broken build. 

I also suspect that ignoring particular artifacts is probably much less used compared to only including specific configurations.

Jenson3210:
Adapted by removing the artifacts filtering.

shanman190 (approved):
Looks good! Thanks @Jenson3210

PR #5450, opened 2025-05-18
Add test for all dependency management entry variants

torsten-liermann:
For more context and background information, please see the discussion in [issue #983](https://github.com/openrewrite/rewrite-maven-plugin/issues/983).


torsten-liermann:
I’ve pushed the test as a draft PR.

I’m currently not working on the actual fix, as I would prefer to refactor the Maven parser and `ResolvedPom` to be more closely aligned with Maven’s behavior.
In particular, I think it would be helpful to replace `null` values with Maven defaults early on — such as `type = "jar"`, `classifier = ""`, and `scope = Scope.Compile`.
As a quick workaround, something like this could work:

```java
      @Value
        private class GroupArtifactScope {
            GroupArtifact ga;
            @Nullable
            Scope scope;
            @Nullable
            String classifier;
            @Nullable
            String type;
        }
```

But ideally, we’d handle this consistently within the model itself.

timtebeek:
Great to see the tests pass here; any reason this is still marked as draft? Or can we already review & merge?

sambsnyd:
Thanks @torsten-liermann !

PR #5452, opened 2025-05-19
Enhance AddImport to shorten fully qualified names

amishra-u:
## What's changed?
PR #5439 introduced logic to handle import ambiguity by using fully qualified names when conflicting usages are present. However, this led to regressions in cases where such conflicts only existed temporarily during multi-step recipe transformations.
### Issue
When multiple recipes are applied, intermediate steps may introduce conflicting usages. But at the end might all such usage can be safely replaced to shortend names.

To handle this, updated AddImport to shorten existing fullyqualified names for the type or member for which import is being added.

### Testing
Validated the fix by running all tests in the rewrite-testing-frameworks.

### Checklist


timtebeek (approved):
Thanks for the quick fixes here and downstream in https://github.com/openrewrite/rewrite-testing-frameworks/pull/731; great to roll forwards.

PR #5453, opened 2025-05-19
Fixed the scanner being executed in a loop that then causes all scans to be updated in a single run.

Jenson3210:
## What's changed?
Fixed some regression introduced by moving the `updatesPerProject` to an accumulatro -> 1 run of visitor would upgrade all of the scanned newVersions resulting in the because being the same. 
We only update if the recipe is for the provided group/artifact BUT we keep adding the new version in scanner so that other recipes can use 1 grouped accumulator.

### Checklist


timtebeek (approved):
Thanks for the quick fix!

PR #5454, opened 2025-05-19
Rename variables if the name matches class name

radoslaw-panuszewski:
## What's your motivation?
When I rename a type in IntelliJ IDEA (for example `SomeType` to `AnotherType`) it will automatically rename variables with the same name as the class, but starting with lowercase character (so `someType` to `anotherType`). 

This PR introduces the above feature to `org.openrewrite.java.ChangeType` recipe.

### Checklist


rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
```diff
@@ -732,4 +732,14 @@ public static boolean containsWhitespace(String s) {
...
+
+    /**
+     * @return string with first character changed to lowercase (or empty string)
+     */
+    public static String decapitalize(@Nullable String string) {
+        if (string != null && !string.isEmpty()) {
+            return Character.toLowerCase(string.charAt(0)) + string.substring(1);
+        }
+        return "";
+    }
```
timtebeek, line 744: Could we start this out as just a private method inside ChangeType itself? We try to minimize exposed API, as folks will use anything they can see, which limits our ability to evolve for something essentially only used once so far.

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -1592,8 +1592,8 @@ public A1 method(A1 a1) {
               import a.A2;
               
               public class Example {
-                  public A2 method(A2 a1) {
-                      return a1;
+                  public A2 method(A2 a2) {
+                      return a2;
```
timtebeek, line 1596: Neat idea indeed to rename these as well!

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -368,6 +369,19 @@ public J visitIdentifier(J.Identifier ident, ExecutionContext ctx) {
                     }
                 }
 
+                // Rename variable if it matches class name (starting with a lowercase character)
+                if (ident.getSimpleName().equals(decapitalize(className))) {
+                    if (targetType instanceof JavaType.FullyQualified) {
+                        String newName = decapitalize(((JavaType.FullyQualified) targetType).getClassName());
```
timtebeek, line 375: Would you mind wrapping this in a use of `VariableNameUtils.generateVariableName`? in 99.9...% of the cases those will be the same, but it's best to avoid conflicts where we can.
https://github.com/openrewrite/rewrite/blob/3be17c88ccefef6820736a7dce12fbe514605d03/rewrite-java/src/main/java/org/openrewrite/java/VariableNameUtils.java#L49

rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
radoslaw-panuszewski: Moved to private method in `ChangeType`

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
radoslaw-panuszewski: Done as suggested 😉

rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
```diff
@@ -732,4 +732,14 @@ public static boolean containsWhitespace(String s) {
 
         return false;
     }
+
+    /**
+     * @return string with first character changed to lowercase (or empty string)
+     */
+    public static String decapitalize(@Nullable String string) {
+        if (string != null && !string.isEmpty()) {
+            return Character.toLowerCase(string.charAt(0)) + string.substring(1);
```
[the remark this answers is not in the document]
radoslaw-panuszewski: Added support for unicode characters + test case for that (`Ą1` -> `Ą2`)

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -387,6 +405,15 @@ public J visitIdentifier(J.Identifier ident, ExecutionContext ctx) {
             return visitAndCast(ident, ctx, super::visitIdentifier);
         }
 
+        @Override
+        public J.VariableDeclarations.NamedVariable visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionContext executionContext) {
+            J.VariableDeclarations.NamedVariable v = (J.VariableDeclarations.NamedVariable) super.visitVariable(variable, executionContext);
+            if (v.getVariableType() != null && !v.getSimpleName().equals(v.getVariableType().getName())) {
+                return v.withVariableType(v.getVariableType().withName(v.getSimpleName()));
+            }
```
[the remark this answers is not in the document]
radoslaw-panuszewski: This piece of code only updates variable type name if the underlying `Identifier` has been renamed earlier. 

I've added a test case `doNotRenameRandomVariablesMatchingClassName` to verify that random variables with matching name are not renamed.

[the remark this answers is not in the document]
timtebeek: Great thanks! I'm rushing out the door here, but will ask the wider team to review while I'm out over the next few days.

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -1630,6 +1630,92 @@ public class Test {
         );
     }
 
+    @Test
+    void updateMethodTypeWithUnicodeCharacter() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Ą1", "a.Ą2", false)),
```
mkosmul, line 1636: Since `Ą` is in the BMP, this test should pass with the old code as well. In order to really see the change going from characters to codepoints, you probably need something like `𐐓` (\uD801\uDC13).

radoslaw-panuszewski: I tried changing `Ą/ą`  to `𐐅/𐐭` but I encountered strange issues from unrelated parts of code (like `import` statement started disappearing). I've spent some time debugging it but eventually gave up as it seems to be a broader issue and fixing it is out of scope of this PR.

I reverted the code of `decapitalize` method is reverted to the previous version and removed the test case with special characters. All of my experiments are in the git history in case someone would like to continue the fight from there 😄

radoslaw-panuszewski:
Hey @timtebeek, could you ping rest of the team to take a look? 😉

radoslaw-panuszewski:
Hi @timtebeek, thanks for the changes! The only thing I don't understand is the commit [Do not expect Kotlin changes just yet](https://github.com/openrewrite/rewrite/pull/5454/commits/cc6431165f4945a5749bb39347cac4570a2b4760). Does it mean this recipe won't rename variables in Kotlin? That would be bad news for my company, which have most of the code written in Kotlin 😞 

Or maybe the "just yet" means that more commits will follow? 😛

timtebeek:
I've not yet worked out why we're not seeing change for Kotlin, and my IDE was throwing up some unrelated issues in trying to figure it out. It might be that `RenameVariable` visitor, or something else still. I'm definitely open to fixing that, but perhaps best done on a separate PR such that we have something to work from already. Appreciate any help debugging that addition!

timtebeek:
Looks to fail on 
image

radoslaw-panuszewski:
Thanks for your help @timtebeek, looking forward to the release! 😄

timtebeek:
Unfortunately we saw some undesired incomplete changes downstream, so we've decided to revert this for now:
- https://github.com/openrewrite/rewrite/commit/982ad13ca1282bc9831864670f8bb800454c8e04

It's at times hard to make guaranteed correct changes, so as much as we like the suggestion, in this case it's too risky.

radoslaw-panuszewski:
> Unfortunately we saw some undesired incomplete changes downstream, so we've decided to revert this for now:
> 
> * [982ad13](https://github.com/openrewrite/rewrite/commit/982ad13ca1282bc9831864670f8bb800454c8e04)
> 
> It's at times hard to make guaranteed correct changes, so as much as we like the suggestion, in this case it's too risky.

@timtebeek could you give some examples of the "undesired incomplete changes downstream"? If this change is too risky, I would love to understand why 🙏

radoslaw-panuszewski:
@timtebeek maybe it would be possible to hide this feature under a boolean flag, something like `renameVariables = true`?

timtebeek:
hi! So cases of undesired renames is anything that's public, and might then be referenced elsewhere, or implied public through the use of say Lombok annotations or (de)serialization. In short it'd only be safe if we know the exact scope, and even then might result in conflicts that need to be handled correctly. While we could potentially iterate on this, ChangeType is very much a corner stone recipe for a lot of migrations, and changes here are likely to ripple out to cause friction downstream. For that same reason we do not like to add boolean flags to recipes: those tend to accumulate over time, and result in a need to adopt ever more optional arguments in constructors. While we understand it's perhaps a desirable change on a small scale, the potential effects on a large scale mean we can not safely add this at this time. Appreciate the suggestion though!

Issue #5227, opened 2025-03-27
Type attribution terminated after first Exception thrown


Knut-Ove:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.49.1

## How are you running OpenRewrite?
I am running the parser from in-house CLI, separate from the regular builds.
My entry-point is:
I set up a JavaParser with a classpath that sometimes is incomplete (typically dependencies from sub-modules in same project are missing - unless they have been published to maven repo)
Normally using the Java21 parser
Deep down in the java compiler there are asserts throwing exception. Have seen this related to switch statements involve these missing types.

Navigating into the ReloadableJava21Parser, line 240 shows the call that throws exception:
https://github.com/openrewrite/rewrite/blob/7ac1bc8f07d030e9861c97f7a12b21b81da0b4e7/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21Parser.java#L240

In such case, the compiler.todo still has a (potentially large) number of entries that are left unhandled.

Inspection of the attribute method shows that it itereates over the list and remove/process one element at a time.
So I have experimented by wrapping the call to attribute with these lines (continuing processing even after a failure)
```java
 while (!compiler.todo.isEmpty()) {
      try {
          compiler.attribute(compiler.todo);
      } catch (Throwable t) {
          ctx.getOnError().accept(new JavaParsingException("Failed symbol entering or attribution", t));
      }
  } 
```
That seems to do the trick.

Even if there are some issues with type attribution for some classes, most of them are handled correctly.
I do analysis of log statements, and need type info for both the method invocation to log4j/slf4, and also the type info for the parameters.
I do have a fallback processing on my side (needed that before lombok was supported) where method name is used to categorize the invocation (debug,info,warn..etc) - but that is a second best solution.

I also experimented by manually patching in the missing jar files to the classpath (not sustainable in large scale) - and the attribution was correct (without the above code patch)

Hope this is something you can have a look at even if procedure to recreate / stack trace is missing.

timtebeek:
Thanks for the detailed report& suggestion @Knut-Ove ! Indeed seems promising to have more type information despite failures. I'll leave it to my more parser minded colleagues on whether to adopt the suggestion, but it sounds like a safe and measured approach.

/cc @knutwannheden

arodionov:
[...]

PR #5456, opened 2025-05-19
Continue type attribution in a case of exception

arodionov:
Since type attribution is a best-effort, we continue it even in the presence of errors. But in a current implementation for `JavaCompiler#attribute(Queue)`, it will be stopped in case of error, and not all entries in a queue will be processed.

In the fix, the attribution will be continued for the rest entries in `compiler.todo`

- Fix https://github.com/openrewrite/rewrite/issues/5227

Kudos [@Knut-Ove](https://github.com/Knut-Ove) for the finding and a fix


timtebeek (approved):
Thanks for the detailed look here @arodionov and thanks @knut-ove for the suggestion. Hope this helps clear out yet another case of missing types going forward.

PR #5458, opened 2025-05-19
 Implement ImportManager for managing imports

amishra-u:
## What's changed?
Introduced the ImportManager to handle import management. It provides APIs to:

- Detect missing or ambiguous imports
- Check whether a type or member can be referenced via an import safely

This utility can be leveraged by existing visitors AddImport and RemoveImport, and we can also create a new visitor (e.g., FixImport, see #5435) to automatically add or remove imports based on usage. We can also build recipes like RemoveUnusedImport on top of this.

Additionally, ImportManager can support APIs to determine whether a list of imports can be safely folded or to unfold wildcard imports. Currently not implemented but could be easily built on top of the existing structure.

## What's your motivation?
While working on #5439, I noticed that AddImport misses some edge cases. I attempted to address import ambiguity, but the solution still doesn't cover all scenarios.  This utility class provides a centralized and robust solution to handle all such cases.

## Additional context
- https://github.com/openrewrite/rewrite/issues/5435
- https://github.com/openrewrite/rewrite/issues/5439

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/ImportManager.java
```diff
@@ -0,0 +1,370 @@
...
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.java.ImportManager.ImportStatus.*;
+
+public class ImportManager {
```
MBoegers: Can we find a better name than `Manager`? I think of a manager like something that not only knows about something but also can alter them.

We ususally use the suffix `Util` for helper, but I get that it is not a classic Util as it maintains state.

Maybe `ImportReport` as it can answer questions?

MBoegers: Would be final reasonable?

rewrite-java/src/main/java/org/openrewrite/java/ImportManager.java
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    public enum ImportStatus {
+        NOT_IMPORTED,
+        IMPLICITLY_IMPORTED,
+        EXPLICITLY_IMPORTED,
+        IMPORT_AMBIGUITY,
+    }
+
+    @Getter
```
MBoegers: Maybe a `@Value` ?

rewrite-java/src/main/java/org/openrewrite/java/ImportManager.java
```diff
@@ -0,0 +1,370 @@
...
+            resolvedMembers.addAll(matched);
+            matched.forEach(memberReferences::remove);
+        }
+    }
+
+    private static class FindUnqualifiedTypeReference extends JavaVisitor<Set<JavaType.FullyQualified>> {
+        @Override
+        public J visitFieldAccess(J.FieldAccess fieldAccess, Set<JavaType.FullyQualified> types) {
+            if (fieldAccess.getType() instanceof JavaType.FullyQualified) {
+                if (fieldAccess.isFullyQualifiedClassReference(((JavaType.FullyQualified) fieldAccess.getType()).getFullyQualifiedName())) {
```
MBoegers: We have`TypeUtils.asFullyQualified` defined which does the same.

rewrite-java/src/main/java/org/openrewrite/java/ImportManager.java
```diff
@@ -0,0 +1,370 @@
...
+            return super.visitMethodInvocation(method, memberReferences);
+        }
+
+        @Override
+        public J.Identifier visitIdentifier(J.Identifier identifier, Set<JavaType> memberReferences) {
+            if (!(identifier.getType() instanceof JavaType.Variable)) {
+                return identifier;
+            }
+            JavaType.Variable variable = identifier.getFieldType();
+            if (variable != null && variable.getOwner() instanceof JavaType.Class && variable.hasFlags(Flag.Static)) {
```
MBoegers: You have `ImportManager.isStaticClassVariable()` which does the same.

MBoegers (changes requested):
Sorry for the long run.. Overall looks good, just some ideas around the namings and annotations

rewrite-java/src/main/java/org/openrewrite/java/ImportManager.java
amishra-u: I agree with you. What about `ImportAnalyzer`?

amishra-u: IMO final could be optional considering it can be used for others JVM based languages with few modifications.

amishra-u:
@MBoegers incorporated feedback. PTAL

amishra-u:
@timtebeek @MBoegers Got a stale notification for this pr. I believe I have already incorporated all feedbacks. PTAL

PR #5460, opened 2025-05-20
Added Gradle caching for npm test

greg-at-moderne:
## What's changed?

Adding Gradle bulld caching for NPM tests in `rewrite-javascript`.

Note - this is coarse-grained caching - i.e. any change in `rewrite-javascript` makes the cache invalid. No fine-grained caching for specific files, etc. I don't think this matters for now.

Another note - this assumes the tests are deterministic. I.e. the same code in specified input files gives the same result.

## What's your motivation?
- To prevent Gradle from running these tests for unrelated PRs - i.e. these with no Javascript changes at all

## Testing performed

- First run: https://ge.openrewrite.org/s/idawlbmqy6qbw
- Then the second with no changes: https://ge.openrewrite.org/s/q2dow5pyppqrk (retrieved from cache 👍 )
- Then applying some to some of the source files: https://ge.openrewrite.org/s/omxettigqfpos (all tests were run again 👍 )
- Then re-running with no further change: https://ge.openrewrite.org/s/ihuosa6dbzcbc (retrievd from cache 👍 )
- etc.
- Then make one of the tests fail: https://ge.openrewrite.org/s/4e5cmfj2myiiw (all tests were run 👍 )
- Then re-run with no further change: https://ge.openrewrite.org/s/t3rbgbinchhie (all tests were run again and failed too 👍 )

### DIFF
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -62,6 +62,7 @@ npmTest.configure {
         include("jest.config.js")
     })
     inputs.files(fileTree("rewrite/src"))
+    inputs.files(fileTree("rewrite/test"))
     outputs.files("rewrite/README.md") // A fake output entry; without it Gradle build caching doesn't work
 }
 tasks.check {
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -56,6 +56,14 @@ extensions.configure<NodeExtension> {
 }
 
 val npmTest = tasks.named("npm_test")
+npmTest.configure {
+    inputs.files(fileTree("rewrite") {
+        include("*.json")
+        include("jest.config.js")
+    })
+    inputs.files(fileTree("rewrite/src"))
+    outputs.files("rewrite/README.md") // A fake output entry; without it Gradle build caching doesn't work
+}
 tasks.check {
     dependsOn(
         tasks.named("npmInstall"),
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -64,6 +64,8 @@ npmTest.configure {
     inputs.files(fileTree("rewrite/src"))
     inputs.files(fileTree("rewrite/test"))
     outputs.files("rewrite/README.md") // A fake output entry; without it Gradle build caching doesn't work
+
+    dependsOn(tasks.named("npmInstall"))
 }
 tasks.check {
     dependsOn(


rewrite-javascript/build.gradle.kts
```diff
@@ -56,6 +56,14 @@ extensions.configure<NodeExtension> {
 }
 
 val npmTest = tasks.named("npm_test")
+npmTest.configure {
+    inputs.files(fileTree("rewrite") {
+        include("*.json")
+        include("jest.config.js")
+    })
+    inputs.files(fileTree("rewrite/src"))
```
knutwannheden, line 64: We definitely also need the `rewrite/test` source files as input.

rewrite-javascript/build.gradle.kts
```diff
@@ -56,6 +56,14 @@ extensions.configure<NodeExtension> {
...
+    outputs.files("rewrite/README.md") // A fake output entry; without it Gradle build caching doesn't work
```
knutwannheden, line 65: I would have expected that we would have to have the task "touch" an output file somewhere. I think we can easily create that in the output folder somewhere.

rewrite-javascript/build.gradle.kts
greg-at-moderne: Right. Missed that. Confused with `src/test`.

1b5aaf042

rewrite-javascript/build.gradle.kts
greg-at-moderne: Why would we do that?

greg-at-moderne:
CI failure:
```
    Reason: Task ':rewrite-javascript:npm_test' uses this output of task ':rewrite-javascript:npmInstall' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```
I didn't see that locally. 👀

rewrite-javascript/build.gradle.kts
knutwannheden: It just seems weird. I don't know if it can have any other side effects, but at least if I were to change that README.md file manually, it would end up rerunning the tests.

greg-at-moderne: OK. I picked the file as:

- one that already exists
- is the least likely to be changed
- it's clear that it doesn't have impact to correctness of the tests

Adding a fake magic file fails the criteria above.

knutwannheden: I also had to add a line like this: `dependsOn(tasks.named("npmInstall"))`, otherwise I would get an error like this when running e.g. `:rewrite-javascript:check`:

```
A problem was found with the configuration of task ':rewrite-javascript:npm_test' (type 'NpmTask').
  - Gradle detected a problem with the following location: '/Users/knut/git/openrewrite/rewrite/rewrite-javascript/rewrite'.
    
    Reason: Task ':rewrite-javascript:npm_test' uses this output of task ':rewrite-javascript:npmInstall' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':rewrite-javascript:npmInstall' as an input of ':rewrite-javascript:npm_test'.
      2. Declare an explicit dependency on ':rewrite-javascript:npmInstall' from ':rewrite-javascript:npm_test' using Task#dependsOn.
      3. Declare an explicit dependency on ':rewrite-javascript:npmInstall' from ':rewrite-javascript:npm_test' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/8.14/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.

```

greg-at-moderne: That's already accounted for - see below.

knutwannheden: As an alternative proposed by AI we can maybe set `outputs.upToDateWhen { true }` rather than using a fake output. Didn't try that, however.

We could also look into running Jest with `--reporters=jest-junit` if that helps us with the reporting in Develocity.

rewrite-javascript/build.gradle.kts
```diff
@@ -56,6 +56,17 @@ extensions.configure<NodeExtension> {
...
+npmTest.configure {
+    inputs.files(fileTree("rewrite") {
+        include("*.json")
+        include("jest.config.js")
+    })
+    inputs.files(fileTree("rewrite/src"))
+    inputs.files(fileTree("rewrite/test"))
+    outputs.files("rewrite/README.md") // A fake output entry; without it Gradle build caching doesn't work
+
+    dependsOn(tasks.named("npmInstall"))
```
shanman190, line 68: This is better described as `inputs.files(tasks.named("npmInstall"))` instead. 

Gradle best practices state that you should avoid explicit `dependsOn` statements and should instead prefer linking the outputs of the former into the inputs of the later.

The only exception to this is for lifecycle tasks, such as `check` below.

rewrite-javascript/build.gradle.kts
shanman190: I'd recommend configuring jest to generate something as output and then utilize that as the defined output. JUnit XML files are certainly one option.

At present, I don't know of a way to force the Develocity Gradle Plugin to ingest test results outside of the native JVM `Test` task.

knutwannheden:
I configured the JUnit report as output, but I was unable to get the reporting in Develocity working: https://github.com/openrewrite/rewrite/commit/24f3d160dde4d40bcfb576c8c8c1f1150ec8c4a1

Issue #5457, opened 2025-05-19
ConcurrentModificationException in TypeTable


Jenson3210:
We're seeing some ConcurrentModificationException when running parallel tests in our security module as seen [here](https://github.com/moderneinc/rewrite-java-security/actions/runs/15118311858/job/42494475825#step:6:840)

<img width="1208" alt="Image" src="https://github.com/user-attachments/assets/2dfd079d-3f5e-47fc-88e9-ff0468ff8fa7" />

Stack trace:
```
java.util.ConcurrentModificationException
        at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1023)
        at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:1046)
        at org.openrewrite.java.internal.parser.TypeTable.artifactsNotYetWritten(TypeTable.java:140)
        at org.openrewrite.java.internal.parser.TypeTable.read(TypeTable.java:116)
        at org.openrewrite.java.internal.parser.TypeTable.<init>(TypeTable.java:111)
        at org.openrewrite.java.internal.parser.TypeTable.fromClasspath(TypeTable.java:97)
        at org.openrewrite.java.JavaParser.dependenciesFromResources(JavaParser.java:123)
        at org.openrewrite.gradle.GradleParser.defaultClasspath(GradleParser.java:188)
        at org.openrewrite.gradle.GradleParser.parseInputs(GradleParser.java:51)
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:281)
        at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:130)
        at org.openrewrite.java.dependencies.DependencyVulnerabilityCheckTest.gradleTransitive(DependencyVulnerabilityCheckTest.java:46)
```

timtebeek:
Figured link that specific part of the stack trace to the implementation here:
https://github.com/openrewrite/rewrite/blob/7d657228044bf0a45474bd918ab4bd399f429922/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java#L136-L149

PR #5462, opened 2025-05-21
Fix `ConcurrentModificationException` in `TypeTable`

knutwannheden:
The `classesDirByArtifact` map in `TypeTable.java` was susceptible to
concurrent modification exceptions when accessed by multiple threads
simultaneously, particularly during parallel test execution.

This commit introduces the following changes to ensure thread safety:
1.  The `classesDirByArtifact` map is now wrapped using
    `Collections.synchronizedMap(new LinkedHashMap<>())`. This retains
    the insertion order characteristics of `LinkedHashMap` while
    providing synchronized access to its methods.
2.  Explicit `synchronized(classesDirByArtifact)` blocks have been added
    around iterations of `classesDirByArtifact.keySet()` in the
    `artifactsNotYetWritten` method and `classesDirByArtifact.entrySet()`
    in the `load` method. This prevents `ConcurrentModificationException`
    during these iteration processes even if the map is modified by another
    thread between the start of the iteration and an access.

These changes ensure that all accesses and iterations involving
`classesDirByArtifact` are properly synchronized, preventing race
conditions and ensuring stable operation in multi-threaded environments.

Fixes:
 - #5457

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
@@ -88,7 +88,7 @@ public class TypeTable implements JavaParserClasspathLoader {
 
     public static final String DEFAULT_RESOURCE_PATH = "META-INF/rewrite/classpath.tsv.zip";
 
-    private static final Map<GroupArtifactVersion, Path> classesDirByArtifact = new LinkedHashMap<>();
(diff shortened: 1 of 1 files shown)


greg-at-moderne:
For context:
- https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedMap-java.util.Map-
- mentions: "It is imperative that the user manually synchronize on the returned map when iterating over any of its collection views"

knutwannheden:
That second commit should have been a separate PR. Apparently I don't know how to properly operate this thing yet.

PR #5463, opened 2025-05-21
Adding OrderImports recipe in Javascript

greg-at-moderne:
## What's changed?

Adding a new recipe to Javascript: OrderImports.

## What's your motivation?

- I wanted to write a real-life Javascript recipe.
- Picked one which doesn't have any input and needs to be JS-specific (as JS/TS imports are different than Java's).
- Still it might be useful in real-life for people.


rewrite-javascript/rewrite/test/recipe/order-imports.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe("order-imports", () => {
+    const spec = new RecipeSpec()
+    spec.recipe = new OrderImports();
+
+    test("basic", () => {
+        return spec.rewriteRun(
+            typescript(
+            `
+            import {delta, gamma} from 'delta.js';
```
knutwannheden, line 29: Can we also order within `{}`? IIRC, IDEA does this by default using case insensitive comparisons. I suppose that is a style option somewhere.

greg-at-moderne: f5e7d32709b34441be3d9c78af8ef08d6d93517f without handling case insensitive/sensitive setting for now

Issue #5446, opened 2025-05-17
Improve gradle dependency resolution/handling


Jenson3210:
## What problem are you trying to solve?
Up until the introduction of lockfiles we could "decently" resolve dependencies for gradle projects using the Maven classes. During the introduction of lockfiles we had some improvements of the transitive dependency gradle project marker but this only showed the gaps for gradle using the maven resolution even more. 

Together with @shanman190 we decided to create this ticket to at least log the impact of not doing a gradle specific dependency resolution. I quote _the legend himself_:
> We have made it a LONG way by only using the Maven-worldview. Given lockfiles are a Gradle specific feature, we have now kinda opened that door and have inadvertently raised this long known, but hidden behind the closed door of Gradle specific dependency resolution. Now that is the very question that I think we(larger collective) probably need to answer or write a ticket for.

At the moment, there are a few gaps that make gradle specific stuff hard like: 
- correct platform resolution
- user configurations
- multimodular projects with user specified paths, ...
- ...

(FEEL FREE TO EDIT TICKET DESCRIPTION TO ADD/EDIT!)

## Describe the solution you'd like
Nothing just yet. But we need to identify, discuss, divide in subtickets and take up...

PR #5465, opened 2025-05-21
Fixed 2 things in the lockfile

Jenson3210:
## What's changed?
- do not touch the locked lines that are there for unexisting configurations as they might have been put there by a plugin
- transitive upgrade constraints are only bumping the spring managed dependency and not bringing their transitive dependencies according to the lockfile resolution when the spring boot plugin is being used.

## What's your motivation?
We've seen that the lockfile is wrong without these changes. This will be already less wrong than it was before without properly working out: 
- https://github.com/openrewrite/rewrite/issues/5446

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -51,7 +51,6 @@ import org.openrewrite.semver.Semver;
 import org.openrewrite.semver.VersionComparator;
 
 import java.util.*;
-import java.util.function.Function;
 
 import static java.util.Collections.*;
 
@@ -851,7 +850,7 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
         }
     }
 
-    private static GradleProject replaceVersion(GradleProject gp, ExecutionContext ctx, GroupArtifactVersion gav, Set<String> configurations) {
+    public static GradleProject replaceVersion(GradleProject gp, ExecutionContext ctx, GroupArtifactVersion gav, Set<String> configurations) {
         try {
             //noinspection ConstantValue
             if (gav.getGroupId() == null || gav.getArtifactId() == null) {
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -911,15 +914,24 @@ private static org.openrewrite.maven.tree.Dependency maybeUpdateDependency(
...
+            Set<ResolvedDependency> traversalHistory) {
+        if (traversalHistory.contains(dep)) {
+            return dep;
+        }
+        if (Objects.equals(dep.getGroupId(), gav.getGroupId()) && Objects.equals(dep.getArtifactId(), gav.getArtifactId()) && Objects.equals(dep.getVersion(), gav.getVersion())
+                || wouldDowngrade(dep.getGav().asGroupArtifactVersion(), gav.asGroupArtifactVersion())) {
+            return dep;
+        }
+        if (Objects.equals(dep.getGroupId(), gav.getGroupId()) && Objects.equals(dep.getArtifactId(), gav.getArtifactId())) {
+            if (!isSpringPluginManaged(dep)) {
```
shanman190, line 929: callout: For this we generally look at the plugins available on the marker itself rather than the current dependency being a `org.springframework*` group. The rationale for this is because the `org.springframework.boot:spring-boot-dependencies` BOM which is typically utilized manages all Spring _and_ third party versions.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -929,20 +941,38 @@ private static ResolvedDependency maybeUpdateResolvedDependency(
         if (traversalHistory.contains(dep)) {
             return dep;
         }
-        if (Objects.equals(dep.getGroupId(), newDep.getGroupId()) && Objects.equals(dep.getArtifactId(), newDep.getArtifactId()) && Objects.equals(dep.getVersion(), newDep.getVersion())) {
+        if (Objects.equals(dep.getGroupId(), newDep.getGroupId()) && Objects.equals(dep.getArtifactId(), newDep.getArtifactId()) && Objects.equals(dep.getVersion(), newDep.getVersion())
+                || wouldDowngrade(dep.getGav().asGroupArtifactVersion(), newDep.getGav().asGroupArtifactVersion())) {
             return dep;
         }
         if (Objects.equals(dep.getGroupId(), newDep.getGroupId()) && Objects.equals(dep.getArtifactId(), newDep.getArtifactId())) {
+            if (isSpringPluginManaged(newDep)) {
```
shanman190, line 949: callout: see above

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
Jenson3210: Do we know by looking at the plug-in if this particular dependency is managed by the bom? My goal here was to have that list of managed libs somehow in that method (until we have better dependency resolution insights). But indeed, i should add gradleproject as a parameter so that that method only returns true when plugin is present AND it is a managed dependency. We are recursively going through the dependencies so we might have other constraints that have nothing to do with the plugin and I do want to have the other behaviour for the non-plugin-managed ones.

shanman190: So there's not a super easy way with what we presently have stored.

If I were to propose a workaround, it would be codifying the implicit (just spring-boot-dependencies) or discovering any explicit configurations defined for the Spring Dependency Management plugin (assuming those are visible in the current source file).

PR #5470, opened 2025-05-21
Fix XPathMatcher: correctly handle spaces around '=' in predicates with conjunctions

torsten-liermann:
ensuring EE10 migration recipes (e.g. JavaxXmlToJakartaXmlXJCBinding) reliably match intended elements, regardless of whitespace.

## What's changed?

 * Updated the regular expression logic in XPathMatcher to correctly handle whitespace before and after the = sign in XPath predicate conjunctions.
 * Added a regression test for ChangeTagAttribute to ensure correct behavior when spaces are present around = in XPath predicates.

## What's your motivation?

The previous implementation of XPathMatcher failed to match such predicates, which made the official recipe
org.openrewrite.java.migrate.jakarta.JavaxXmlToJakartaXmlXJCBinding unusable in certain scenarios (e.g., when targeting attributes like version="1.0").

Please verify that the regular expression change in XPathMatcher does not cause regressions for other XPath patterns.

No specific reviewers.

As a temporary workaround, the affected recipes could be adjusted to avoid whitespace in XPath expressions.
However, this is not practical for real-world XML, so a generic fix in the matcher was preferred.

## Any additional context

This change will improve the robustness of OpenRewrite's XML recipes for migration and transformation tasks in the presence of varied XML formatting.

### Checklist
- [x] I've added unit tests to cover both positive case
- [-] I've used the IntelliJ IDEA auto-formatter on affected files


sambsnyd:
Thanks @torsten-liermann !

PR #5471, opened 2025-05-22
SpacesVisitor - make sure there's a space after `throw`

greg-at-moderne:
## What's changed?

Minor amendment to `SpacesVisitor` in Javascript.
Adding a mandatory space after `throw`.

## What's your motivation?

Otherwise the following code change is observed:
```diff
-     throw new Error("test");
+     thrownew Error("test");
```

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/format.ts
+++ b/rewrite-javascript/rewrite/src/javascript/format.ts
@@ -704,6 +704,13 @@ export class MinimumViableSpacingVisitor extends JavaScriptVisitor<ExecutionCont
         });
     }
 
+    protected async visitThrow(thrown: J.Throw, p: ExecutionContext): Promise<J | undefined> {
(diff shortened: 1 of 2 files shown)


rewrite-javascript/rewrite/src/javascript/format.ts
```diff
@@ -704,6 +704,13 @@ export class MinimumViableSpacingVisitor extends JavaScriptVisitor<ExecutionCont
         });
     }
 
+    protected async visitThrow(thrown: J.Throw, p: ExecutionContext): Promise<J | undefined> {
+        const ret = await super.visitThrow(thrown, p) as J.Throw;
+        return ret && produce(ret, draft => {
+           draft.exception.prefix.whitespace = " ";
```
knutwannheden, line 710: We should take care to only update the `Space` when required. This is basically only when `whitespace` is empty and there are no comments. This should apply to all modifications performed by this visitor.

PR #5474, opened 2025-05-22
Fir for parse errors when build.gradle contains escaped shell commands

JohannisK:
## What's changed?
- Added tests showing the issue

## What's your motivation?
When using shell commands in a build.gradle, combined with escape characters the build.gradle is not printIdempotent anymore.

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleParserTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleParserTest.java
@@ -16,8 +16,16 @@
 package org.openrewrite.gradle;
 
 import org.junit.jupiter.api.Test;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.Parser;
+import org.openrewrite.SourceFile;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.test.RewriteTest;
+import org.openrewrite.tree.ParseError;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/GradleParserTest.java
```diff
@@ -262,4 +270,38 @@ void kotlinDsl() {
           )
         );
     }
+
+    @Test
+    void error() {
```
JohannisK, line 275: Is there a better way to do this? I have to bypass `RewriteTest` to test the wrong behavior, because RewriteTest explicitly instructs the parser to disregard idempotency issues:

RewriteTest.java line 276 - 289
```
if (requirePrintEqualsInput) {
    // this gets checked by the test framework itself a few statements further down
    ctx.putMessage(ExecutionContext.REQUIRE_PRINT_EQUALS_INPUT, false);
}

List<SourceFile> sourceFiles = parser.parseInputs(inputs.values(), relativeTo, ctx)
                    .collect(Collectors.toList());
assertThat(sourceFiles.size())
                    .as("Every input should be parsed into a SourceFile.")
                    .isEqualTo(inputs.size());

if (requirePrintEqualsInput) {
    ctx.putMessage(ExecutionContext.REQUIRE_PRINT_EQUALS_INPUT, true);
}
```

Jenson3210: I would say this is valid approach. IMHO, you are not really testing a rewriteTest / rewriteRun. Just testing the gradleParser only. More as if unit-testing the single parser unit, not the parser in a rewrite/recipe context

PR #5476, opened 2025-05-22
Fix for idempotency exception when build gradle contains shell commands with escape characters

JohannisK:
## What's changed?
- Added tests showing the issue

## What's your motivation?
When using shell commands in a build.gradle, combined with escape characters the build.gradle is not printIdempotent anymore.

Example:
```
plugins {
    id 'java-library'
}
          
task executeShellCommands {
    doLast {
        exec {
            commandLine 'bash', '-c', "RESPONSE=\$(curl --location -s --request POST \"https://localhost/$path\")"
        }
    }
}
```

Notice the first $sign is escaped, because it's part of the shell command, while the second $sign is not escaped because it's a Groovy variable


Jenson3210 (approved):
I do not know a lot about the parsers yet, but this looks like a correct fix to me

knutwannheden (approved):

JohannisK:
> fyi: This could be simplified even a layer further down into the Groovy parser with an escaped $ and variable replacement in the same string literal. The syntax itself is valid Groovy, but wasn't being handled.

Could you point me to the java file I could do this in? While debugging I found my way to this change, where it seemed to base everything on the node delivered by the Groovy Compiler, but I'm happy to revisit this PR if there is a better solution.

shanman190:
@JohannisK , probably in [LiteralTest.java](https://github.com/openrewrite/rewrite/blob/main/rewrite-groovy%2Fsrc%2Ftest%2Fjava%2Forg%2Fopenrewrite%2Fgroovy%2Ftree%2FLiteralTest.java) would have been where I'd put it for the Groovy specific part of it.

PR #5484, opened 2025-05-22
Fixing parsing of method invocations in Javascript

greg-at-moderne:
## What's changed?

Fixing parsing of method and other function invocations in Javascript. Currently `console.log("something")` gets parsed as `MethodInvocation(select=FieldAccess(console.log), name="")` which is understandably not desired.

The design decision has been to:
- parse "normal" method invocations (with proper select and proper name) as `J.MethodInvocation`
- and any other function calls as a new type - called `JS.FunctionCall`

## What's your motivation?

Avoiding wrong parsing. And providing reasonable level of compatibility with Java parsing.


knutwannheden:
@greg-at-moderne Did we decide what to do about this?

greg-at-moderne:
We have not, forgot about this one. It was several weeks ago.

rewrite-javascript/rewrite/src/javascript/tree.ts
```diff
@@ -643,6 +644,21 @@ export namespace JS {
...
+    /**
+     * Represents function calls which are not method invocations.
+     * @example f(5, 0, 4)
+     * @example f?.(5, 0, 4)
+     * @example data["key"](5, 0, 4)
+     * @example (() => { return 3 + 5 })()
+     */
+    export interface FunctionCall extends JS, Expression {
+        readonly kind: typeof Kind.FunctionCall;
+        readonly select?: J.RightPadded<Expression>;
```
knutwannheden: I think we should instead use something like `function`.
```suggestion
        readonly function?: J.RightPadded<Expression>;
```

Also, to clarify, is the `J.RightPadded` required for use cases like `f?.(1)` where the padding would be what is between `f` and `?.`?

rewrite-javascript/rewrite/src/javascript/tree.ts
```diff
@@ -56,6 +56,7 @@ export namespace JS {
         IndexedAccessTypeIndexType: "org.openrewrite.javascript.tree.JS$IndexedAccessType$IndexType",
         InferType: "org.openrewrite.javascript.tree.JS$InferType",
         Intersection: "org.openrewrite.javascript.tree.JS$Intersection",
+        FunctionCall: "org.openrewrite.javascript.tree.JS$FunctionCall",
```
knutwannheden, line 59: I think this looks pretty good. We need to extend the `JS.java` as well as both `JavaScriptSender` and `JavaScriptReceiver` classes, as well as the `JavaScriptValidator`.

rewrite-javascript/rewrite/src/javascript/tree.ts
greg-at-moderne: > I think we should instead use something like function.

3b998be37

rewrite-javascript/rewrite/src/javascript/tree.ts
greg-at-moderne: 3887d8d and the next two commits.

rewrite-javascript/rewrite/src/javascript/tree.ts
greg-at-moderne: Re padding:
Currently (9c8d727789ce4d74e379e2de9777e26a75a6b143) the code:

| code | parsed as |
| ------| ----| 
| `func/*a*/?./*b*/("TS")` | `FunctionCall(function=func, markers = Optional(prefix=/*a*/), arguments=Container(prefix=/*b*/))` |
| `func/*a*/?./*b*/call("TS")` | `MethodInvocation(select=func, markers = Optional(prefix=/*a*/), name=(Identifier(call), prefix=/*b*/))` |
| `func/*a*/.call/*b*/("TS")` |  `MethodInvocation(select=(func, after=/*a*/), markers = empty, name = (Identifier(call), prefix=/*b*/))` |

=> So I think the `/*b*/` is fine.
But the `/*a*/` needs more work, i.e. needs to be unified not to use `Markers.prefix`

greg-at-moderne: After 5fb8c1d

It's:

| code | parsed as |
| ------| ----| 
| `func/*a*/?./*b*/("TS")` | `FunctionCall(function=(func, after=/*a*/), markers = Optional(prefix=empty), arguments=Container(prefix=/*b*/))` |
| `func/*a*/?./*b*/call("TS")` | `MethodInvocation(select=(func, after=/*a*/), markers = Optional(prefix=empty), name=(Identifier(call), prefix=/*b*/))` |
| `func/*a*/.call/*b*/("TS")` |  `MethodInvocation(select=(func, after=/*a*/), markers = empty, name = (Identifier(call), prefix=/*b*/))` |

which I think is satisfyingly consistent.

rewrite-javascript/src/main/java/org/openrewrite/javascript/tree/JS.java
```diff
@@ -820,6 +821,165 @@ public ExpressionWithTypeArguments withTypeArguments(@Nullable JContainer<Expres
...
+        public @Nullable Expression getFunction() {
+            return function == null ? null : function.getElement();
+        }
+
+        public FunctionCall withFunction(@Nullable Expression function) {
+            return getPadding().withFunction(JRightPadded.withElement(this.function, function));
+        }
+
+        @Nullable
+        @With
```
knutwannheden: AFAICT this was copied from `J.MethodInvocation`. Looking at this I think this is actually wrong. There should not be any `@With` annotation and instead there should be a custom with method with a `List<Expression>` parameter.

rewrite-javascript/src/main/java/org/openrewrite/javascript/tree/JS.java
```diff
@@ -820,6 +821,165 @@ public ExpressionWithTypeArguments withTypeArguments(@Nullable JContainer<Expres
...
+            throw new UnsupportedOperationException("To change the return type of this function call, use withFunctionType(..)");
+        }
+
+        public FunctionCall withDeclaringType(JavaType.FullyQualified type) {
+            if (this.functionType == null) {
+                return this;
+            } else {
+                return withMethodType(this.functionType.withDeclaringType(type));
+            }
+        }
```
knutwannheden: I suggest we remove this method. I don't think it is necessary. I know we have it on `J.MethodInvocation`, but the only place I see it being used is [here](https://github.com/openrewrite/rewrite-logging-frameworks/blob/966498f0656a6e0e60e1ad7eb498360c185e68b7/src/main/java/org/openrewrite/java/logging/log and I don't think it would be necessary (I will try to remove that call).

rewrite-javascript/src/main/java/org/openrewrite/javascript/JavaScriptVisitor.java
```diff
@@ -194,6 +194,35 @@ public J visitExpressionWithTypeArguments(JS.ExpressionWithTypeArguments express
...
+        }
+        if (f.getPadding().getFunction() != null && f.getPadding().getFunction().getElement() instanceof NameTree &&
+                functionCall.getMethodType() != null && functionCall.getMethodType().hasFlags(Flag.Static)) {
+            //noinspection unchecked
+            f = f.getPadding().withFunction(
+                    (JRightPadded<Expression>) (JRightPadded<?>)
+                            visitTypeName((JRightPadded<NameTree>) (JRightPadded<?>) f.getPadding().getFunction(), p));
+        }
+        if (f.getPadding().getFunction() != null) {
+            f = f.getPadding().withFunction(visitRightPadded(f.getPadding().getFunction(), JRightPadded.Location.METHOD_SELECT, p));
```
knutwannheden: Normally we add a new enum constant to `JSRightPadded.Location`. To be consistent I think we should do that here as well. The other two seem OK to me, as we also reuse them in other places.

rewrite-javascript/src/main/java/org/openrewrite/javascript/JavaScriptVisitor.java
```diff
@@ -194,6 +194,35 @@ public J visitExpressionWithTypeArguments(JS.ExpressionWithTypeArguments express
...
+        f = f.withPrefix(visitSpace(f.getPrefix(), JsSpace.Location.FUNCTION_CALL_PREFIX, p));
+        f = f.withMarkers(visitMarkers(f.getMarkers(), p));
+        Expression temp2 = (Expression) visitExpression(f, p);
+        if (!(temp2 instanceof JS.FunctionCall)) {
+            return temp2;
+        } else {
+            f = (JS.FunctionCall) temp2;
+        }
+        if (f.getPadding().getFunction() != null && f.getPadding().getFunction().getElement() instanceof NameTree &&
+                functionCall.getMethodType() != null && functionCall.getMethodType().hasFlags(Flag.Static)) {
```
knutwannheden: For consistency I think we should here always refer to `f` instead of `functionCall`.

rewrite-javascript/rewrite/test/javascript/parser/call.test.ts
```diff
@@ -50,14 +51,27 @@ describe('call mapping', () => {
...
+            afterRecipe: (cu: JS.CompilationUnit) => {
+                const inits = [1, 2, 3].map(i => (cu.statements[i].element as J.VariableDeclarations).variables[0].element.initializer!.element);
+                expect(inits[0].kind).toEqual(JS.Kind.FunctionCall);
+                expect(inits[1].kind).toEqual(J.Kind.MethodInvocation);
+                expect(inits[2].kind).toEqual(J.Kind.MethodInvocation);
+
+                // TODO assert that the comments are placed properly:
+                //      /*a*/ to go into after of select/function,
+                //      /*b*/ to go into prefix of name
+                // Currently it's broken, comments go into whitespace!
```
knutwannheden: This is resolved now, right?

rewrite-javascript/rewrite/src/javascript/print.ts
```diff
@@ -612,6 +612,23 @@ export class JavaScriptPrinter extends JavaScriptVisitor<PrintOutputCapture> {
         return mod;
     }
 
+    override async visitFunctionCall(functionCall: JS.FunctionCall, p: PrintOutputCapture): Promise<J | undefined> {
+        await this.beforeSyntax(functionCall, p);
+
+        if (functionCall.function) {
+            await this.visitRightPadded(functionCall.function, p);
+            if (functionCall.function.element.markers.markers.find(m => m.kind === JS.Markers.Optional)) {
+                p.append("?.");
```
knutwannheden, line 627: It would be nice if we could detangle this a bit and have the `?` be printed by the presence of the marker and then only here print the `.`. Would that make sense?

knutwannheden (commented):
Looks good to me. Left a few small comments.

rewrite-javascript/src/main/java/org/openrewrite/javascript/JavaScriptVisitor.java
greg-at-moderne: fadd2bb6a

rewrite-javascript/rewrite/test/javascript/parser/call.test.ts
greg-at-moderne: 2d4e017cb

rewrite-javascript/src/main/java/org/openrewrite/javascript/tree/JS.java
greg-at-moderne: 04cdf8dc3

rewrite-javascript/src/main/java/org/openrewrite/javascript/JavaScriptVisitor.java
greg-at-moderne: 752d3008e

rewrite-javascript/src/main/java/org/openrewrite/javascript/tree/JS.java
greg-at-moderne: 5d4fd996f

rewrite-javascript/rewrite/src/javascript/print.ts
greg-at-moderne: I had it like you suggest before in my code, but then the Space handling is wrong:
```diff
- const result0 = func/*a*/?./*b*/("TS");
+ const result0 = func?/*a*/./*b*/("TS");
```
For that to work, the `/*a*/` comment would need to land as `Optional.prefix`.
That in turn would make it inconsistent with how the `/*a*/` comment is placed in other similar syntaxes.
See `call.test.ts / with optional chaining operator`.

I suggest to keep it like it is. After all the `?.` is a single token with its own meaning, thus it feels OK to print it in one go.
Wdyt?

knutwannheden: I understand. In that case I think it probably makes sense to keep it this way, unless we want to add some kind a dedicated `optionalChaining` property to the `JS.FunctionCall` type. That would then however require parsing any calls using `?.` into a `JS.FunctionCall`.

PR #5158, opened 2025-03-12
Gradle Kotlin DSL recipes

shanman190:
## What's changed?
Update Gradle recipes to handle Gradle Kotlin DSL.

- [x] AddDependency
- [x] ChangeDependency
- [x] UpgradeDependencyVersion
- [x] UpgradeTransitiveDependencyVersion
- [x] RemoveDependency

## What's your motivation?
- Closes #3291.

## Anything in particular you'd like reviewers to focus on?
Nothing in particular

## Anyone you would like to review specifically?
Anybody

## Have you considered any alternatives or workarounds?
N/A

## Any additional context
There are many recipes to handle, but for now let's just focus on the ones that are primarily in just the critical path of other high profile recipes (Spring Boot migrations, Find and fix vulnerable dependencies, etc).

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #5502, opened 2025-05-27
Added kotlin support to UseHttpsForRepositories

JohannisK:
## What's changed?
Added kotlin support to UseHttpsForRepositories recipe

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
@@ -33,8 +33,6 @@ import static java.util.Collections.singleton;
 
 public class UseHttpsForRepositories extends Recipe {
 
-    public static final List<String> PARENT_METHOD_INVOCATIONS = Arrays.asList("repositories", "maven");
-
     @Override
     public String getDisplayName() {
         return "Use HTTPS for repositories";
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
```diff
@@ -69,16 +68,20 @@ private J.Literal fixupLiteralIfNeeded(J.Literal arg) {
             }
 
             @Override
-            public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
-                J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (REPO_URL.matches(method)) {
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+                if (method.getSimpleName().equals("url") || method.getSimpleName().equals("uri")) {
```
Jenson3210, line 73: I think by removing the `MavenArtifactRepositorySpec` there are suddenly going to be a lot more url() method that match. 
Not a real world example as artifactory does not have this property but this would suddenly match: 
```groovy
artifactory{
    url "http://artifactory.jfrog.com
}
```
where it is not this recipe to make this change as that would not be a repository. 
If you want me to figure out a real working example for this url/uri, let me know!

rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
```diff
@@ -57,7 +56,7 @@ public Duration getEstimatedEffortPerOccurrence() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(new UsesMethod<>(REPO_URL), new GroovyVisitor<ExecutionContext>() {
```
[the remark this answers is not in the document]
timtebeek: `IsBuildGradle` should be fine; without type information we can't do much more, and we do not expect thousands of build gradle (.kts) files per repository for that to matter much.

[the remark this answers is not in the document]
jevanlingen: Yep, you need to broaden it. Instead of just Groovy you go up one more level to support all Java like languages. @shanman190 did the same in 
- https://github.com/openrewrite/rewrite/pull/5158

so this is good!

rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
```diff
@@ -69,16 +69,21 @@ private J.Literal fixupLiteralIfNeeded(J.Literal arg) {
             }
 
             @Override
-            public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
-                J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (REPO_URL.matches(method)) {
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation m, ExecutionContext ctx) {
+                if (PARENT_METHOD_INVOCATIONS.contains(m.getSimpleName())) {
```
shanman190, line 73: While it is discouraged, this also would miss `allprojects`, `subprojects`, and `configure(..)` subsections.

Jenson3210 (approved):
Great to see the kotlin support being added now!

rewrite-gradle/src/main/java/org/openrewrite/gradle/security/UseHttpsForRepositories.java
Jenson3210: Thanks for pointing that out! 
I personally think we can add these still to the PARENT_METHOD_INVOCATIONS, what do you think?

shanman190: Yep, I think that's the easiest path forward to ensure that bases are covered.

JohannisK: I can still do that, but because drilling down would have us adding edge cases I tried another approach of drilling up from uri/url and validating the parent calls are first "maven" and then "repositories" 

I think that's a more robust approach

Issue #4527, opened 2024-09-26
Generalize property handling in Maven recipes?


gsmet:
Hi,

We have been very happy with the property handling in [UpgradeDependencyVersion](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java): when a version is a property, the property is updated instead of the original version tag.
Which is awesome as a lot of versions are properties.

All good but... in Quarkus, we are using properties in generated projects also for the groupId/artifactId of the Quarkus BOM (and it's something that is important to us as we have several "Platforms" with different coordinates and we need to enforce the Platform consistently to the BOM and the Quarkus Maven plugin, thus why we use properties).

While working on a PR to handle changing the Platform entirely (i.e. including the groupId/artifactId, until now we were just updating the version), I stumbled upon the fact that this nice property handling is actually not generalized to the Maven recipes.

For instance, there is nothing to handle it in [ChangeDependencyGroupIdAndArtifactId](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java) or [ChangeManagedDependencyGroupIdAndArtifactId](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java) (unless I missed something  but the code seems in line with the behavior I'm experiencing).

I'm not sure exactly why this is handled only in a very specific case. I'm also not completely sure it should be generalized everywhere but not handling it for the groupId/artifactId is actually quite blocking for us. And to be fair, the version is not handled either in `ChangeDependencyGroupIdAndArtifactId` and `ChangeManagedDependencyGroupIdAndArtifactId` so it's not very consistent.

I was wondering if it's something you would consider? If you had some advice on how to fix it? I could probably dedicate some time to it as it's in our way but I definitely need some guidance. Also, I might not be able to generalize it everywhere in the time I would be able to dedicate to it but I could at least put the infrastructure in place and fix the cases that are in our way, opening the gate for more in the future if someone needs it elsewhere.

Thoughts?

timtebeek:
Hi @gsmet ! Thanks for the additional context! I think the Jenkins folks have similar requirements, as seen on this existing PR:
- https://github.com/openrewrite/rewrite/pull/4404

[...]

PR #5503, opened 2025-05-27
Generalize changing of properties in Maven recipes

jevanlingen:
## What's changed?
The "ChangeDependencyGroupIdAndArtifactId" and "ChangeManagedDependencyGroupIdAndArtifactId" recipes now handle a value change for the version tag including cases where a property is used.

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4527

I moved the logic of the property resolvent out of the "UpgradeDependencyVersion" and into the "MavenVisitor" (and used it in the other recipes). Is this a good solution?

### Checklist


Jenson3210 (approved):
I am not a maven person... But i like the move of reused code to more central location which might make future changes easier and also more wide-spread. 👍 
Also, it does seem to do what is needed

Issue #5361, opened 2025-04-30
Plugin management support for Gradle Kotlin scripts (.kts)


bmuschko:
## What problem are you trying to solve?

Dependency management has been enhanced to support the Gradle Kotlin DSL via https://github.com/openrewrite/rewrite/pull/5158. Recipes for managing plugins only works for the Gradle Groovy DSL, but should be supported for the Kotlin DSL as well.

## Describe the solution you'd like

Support the Kotlin DSL for the recipes in the package `org.openrewrite.gradle.plugins`.

## Have you considered any alternatives or workarounds?

No workarounds are available right now except for writing a custom recipe.

## Additional context

Some upgrade recipes e.g. for Spring Boot may need to touch the applied plugin(s) in a Gradle build.

shanman190:
This should be pretty a pretty straightforward migration from `Groovy*Visitor` to `Java*Visitor` because the plugin syntax is much more rigid in terms of it's definition.

timtebeek:
Thanks for the help & suggestions both! Also let us know if my team can help out bring this forward. I imagine the changes are straightforward, and a review is easier to fit in.

timtebeek:
A first change to `UpgradePluginVersion` was already pushed in https://github.com/openrewrite/rewrite/commit/b5c3884eb74200f3711f0a577a9b4aabd62fd6d6

Jenson3210:
Did some check where we are at today regarding gradle recipes and groovy/kotlin support.

#### Change a plugin version
AFAICS there are 2 recipes to change the version of a plugin. 
We have `ChangePluginVersion` which is only a GroovyVisitor and `UpgradePluginVersion` which uses a JavaVisitor. 
- [ ] Can we deprecate `ChangePluginVersion` so that we support kotlin and groovy with 1 recipe? 

#### DependencyUseMap/StringNotation
These ones only support groovy for now. 
- [ ] Do we add support for kotlin map notation? `implementation(group = "org.openrewrite", name = "rewrite-xml")`

#### RemoveRepository
- [ ] we should add support for kotlin

#### UpdateJavaCompatibility
- [ ] we should add support for kotlin

#### FindDependency
- [ ] we should add support for kotlin

#### UseHttpsForRepositories
- [ ] we should add support for kotlin

I can create tickets for the last 5 as they are not related to plugins. The first one should be handled / deprecated and then I think we can close this ticket.

shanman190:
[...]

PR #5504, opened 2025-05-27
Deprecated ChangePluginVersion in favor of UpgradePluginVersion as that one is Kotlin aware.

Jenson3210:
As described here: 
- https://github.com/openrewrite/rewrite/issues/5361

>AFAICS there are 2 recipes to change the version of a plugin. We have ChangePluginVersion which is only a GroovyVisitor and UpgradePluginVersion which uses a JavaVisitor. Can we deprecate ChangePluginVersion so that we support kotlin and groovy with 1 UpgradePluginVersion recipe?

> ChangePluginVersion and UpgradePluginVersion both are only impacting the version, have the same amount/type of parameters.UpgradePluginVersion goes further in the way that if the version is a variable, it will update the variable value and not overwrite the variable usage with the updated version.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/ChangePluginVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/ChangePluginVersion.java
@@ -40,8 +40,12 @@ import org.openrewrite.semver.Semver;
 
 import java.util.List;
 
+/**
+ * @deprecated in favor of {@link org.openrewrite.gradle.plugins.UpgradePluginVersion}.
+ */
 @Value
 @EqualsAndHashCode(callSuper = false)
+@Deprecated
 public class ChangePluginVersion extends Recipe {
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/ChangePluginVersion.java
```diff
@@ -17,6 +17,7 @@
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.gradle.internal.impldep.com.google.errorprone.annotations.InlineMe;
```
jkschneider, line 20: Not a fan of using InlineMe annotations in our codebase, especially this shaded one.

Jenson3210: @timtebeek told me to add this one if I understood the comment. 
Can remove it again!

PR #5511, opened 2025-05-27
Removing parentheses following annotation if they are empty and you try to change the annotation attribute name.

steve-aom-elliott:
There was a bit of an incorrect modification of annotation attributes when using `ChangeAnnotationAttributeName`, specifically when you have an annotation that was written like this: `@Schema()`. To the user, obviously the presence or absence of the parentheses doesn't make a difference in how it behaves, but to openrewrite, it could end up performing operations resulting in an inability to compile.

When written as `@Schema()`, calling `ChangeAnnotationAttributeName` to change `"value"` to `"somethingelse"` would result in openrewrite changing the annotation to `@Schema(somethingelse = )`, breaking the code.

This PR will filter the arguments on the annotation to remove those that are instances of `J.Empty` (there should only ever be at most one of these, and in that situation it would be the only argument anyway) before performing the iterating through the arguments to try to rename the correct one, meaning it should result in `@Schema` instead.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeAnnotationAttributeNameTest.java
@@ -102,7 +102,7 @@ class ChangeAnnotationAttributeNameTest implements RewriteTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
I've changed it slightly to make no change at all, which I think would be best here to avoid complaints about such cases. Hope you agree!

PR #5512, opened 2025-05-28
Added Kotlin support to "UpdateJavaCompatibility"

JohannisK:
## What's changed?
Added Kotlin support to "UpdateJavaCompatibility"

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
@@ -271,9 +271,8 @@ public class UpdateJavaCompatibility extends Recipe {
                         return m.withArguments(
                                 Collections.singletonList(versionArg.withValue(version)
                                         .withValueSource(version.toString())));
-                    } else {
-                        return m;
                     }
+                    return m;
                 }
             }
 
@@ -295,9 +294,8 @@ public class UpdateJavaCompatibility extends Recipe {
                     DeclarationStyle actualStyle = declarationStyle == null ? currentStyle : declarationStyle;
                     //noinspection DataFlowIssue
                     return m.withArguments(ListUtils.mapFirst(m.getArguments(), arg -> changeJavaVersion(arg, actualStyle)));
-                } else {
-                    return m;
                 }
+                return m;
             }
 
             return SearchResult.found(m, "Attempted to update to Java version to " + version +
@@ -374,12 +372,11 @@ public class UpdateJavaCompatibility extends Recipe {
             String removePrefix = version.substring(version.indexOf("_") + 1);
             if (removePrefix.startsWith("1_")) {
                 return removePrefix.substring(removePrefix.indexOf("_") + 1);
-            } else {
-                return removePrefix;
             }
-        } else {
-            return version.substring(version.indexOf(".") + 1);
+            return removePrefix;
         }
+
+        return version.substring(version.indexOf(".") + 1);
     }
 
     private Expression changeJavaVersion(Expression expression, @Nullable DeclarationStyle style) {


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -392,13 +406,28 @@ private G.CompilationUnit addCompatibilityTypeToSourceFile(G.CompilationUnit c,
...
-            sourceFile.getStatements();
+            c = c.withStatements(ListUtils.concatAll(c.getStatements(), sourceFile.getStatements()));
+        }
+        return c;
+    }
+
+    private K.CompilationUnit addCompatibilityTypeToSourceFile(K.CompilationUnit c, String compatibilityType, ExecutionContext ctx) {
+        if ((this.compatibilityType == null || compatibilityType.equals(this.compatibilityType.toString())) && Boolean.TRUE.equals(addIfMissing)) {
+            K.CompilationUnit sourceFile = (K.CompilationUnit) KotlinParser.builder()
+                    .isKotlinScript(true)
```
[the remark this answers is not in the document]
JohannisK: Let me double check that.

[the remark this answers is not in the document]
JohannisK: @timtebeek At this point in the code the cursor is at the "root" which is not a J interface. This means the cursor isn't a valid "scope" cursor for a KotlinTemplate.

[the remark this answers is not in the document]
jevanlingen: @knutwannheden / @shanman190 does it make sense to use the `isKotlinScript` method here?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -91,35 +94,43 @@ public Validated<Object> validate() {
...
                 }
-                if (getCursor().pollMessage(TARGET_COMPATIBILITY_FOUND) == null) {
-                    c = addCompatibilityTypeToSourceFile(c, "target", ctx);
+                if (t instanceof K.CompilationUnit) {
+                    if (getCursor().pollMessage(SOURCE_COMPATIBILITY_FOUND) == null) {
+                        t = addCompatibilityTypeToSourceFile((K.CompilationUnit) t, "source", ctx);
+                    }
+                    if (getCursor().pollMessage(TARGET_COMPATIBILITY_FOUND) == null) {
+                        t = addCompatibilityTypeToSourceFile((K.CompilationUnit) t, "target", ctx);
+                    }
```
jevanlingen, line 118: I think I would write this differently :

```java
if (getCursor().pollMessage(SOURCE_COMPATIBILITY_FOUND) == null) {
  t = addCompatibilityTypeToSourceFile(t, "source", ctx);
}
if (getCursor().pollMessage(TARGET_COMPATIBILITY_FOUND) == null) {
  t = addCompatibilityTypeToSourceFile(t, "target", ctx);
}
```

And do those `instanceof` checks in the `addCompatibilityTypeToSourceFile` method. Because the current two methods do also have code duplication. Something like:

```java
private Tree addCompatibilityTypeToSourceFile(Tree tee, String compatibilityType, ExecutionContext ctx) {
  if ((this.compatibilityType == null || compatibilityType.equals(this.compatibilityType.toString())) && Boolean.TRUE.equals(addIfMissing)) {
    if (t instanceof G.CompilationUnit) {
      // impl
    } else if (t instanceof K.CompilationUnit) {
      // impl
    }
  }
  return tree;
}   
```

JohannisK: I've changed it to be more like you example. I did put one method in between because the handling for Groovy is different from Kotlin.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -384,21 +386,84 @@ private Expression changeExpression(Expression expression, @Nullable Declaration
...
+                                    .isKotlinScript(true)
+                                    .build().parse(ctx, "\n    " + compatibilityType + "Compatibility = " + styleMissingCompatibilityVersion(DeclarationStyle.Enum))
+                                    .findFirst()
+                                    .orElseThrow(() -> new IllegalStateException("Unable to parse compatibility type as a Gradle file"));
+                            return lambda.withBody(body.withStatements(ListUtils.concatAll(statements, sourceFile.getStatements())));
+                        }
+                    }.visitMethodInvocation(method, ctx);
+                }
+                return super.visitMethodInvocation(method, ctx);
+            }
```
jevanlingen, line 458:
```suggestion
            @Override
            public J.Lambda visitLambda(J.Lambda lambda, ExecutionContext ctx) {
                Cursor cursor = getCursor().dropParentUntil(e -> e instanceof J.MethodInvocation);
                if ("java".equals(((J.MethodInvocation) cursor.getValue()).getSimpleName())) {
                    J.Block body = (J.Block) lambda.getBody();
                    List<Statement> statements = body.getStatements();
                    K.CompilationUnit sourceFile = (K.CompilationUnit) KotlinParser.builder()
                            .isKotlinScript(true)
                            .build().parse(ctx, "\n    " + compatibilityType + "Compatibility = " + styleMissingCompatibilityVersion(DeclarationStyle.Enum))
                            .findFirst()
                            .orElseThrow(() -> new IllegalStateException("Unable to parse compatibility type as a Gradle file"));
                    return lambda.withBody(body.withStatements(ListUtils.concatAll(statements, sourceFile.getStatements())));
                }
                return lambda;
            }
```

I think a visitor in a visitor is harder to read than 1 visitor with a `getCursor().dropParentUntil` call.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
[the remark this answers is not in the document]
shanman190: So what I've seen in other `visit` methods is to explicitly "re-set" the parent to be the current element again or to build up the child cursor. I've done both.

One thing that KotlinTemplate has a bit of an issue with though is formatting at the root due to it being nestled within a `Block` and the template returning a prefix of `\n`. This then triggered an indent of the entire statement which is undesirable when at the root.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -91,35 +95,33 @@ public Validated<Object> validate() {
...
-            public J visitAssignment(J.Assignment assignment, ExecutionContext ctx) {
-                J.Assignment a = (J.Assignment) super.visitAssignment(assignment, ctx);
+            public J.Assignment visitAssignment(J.Assignment assignment, ExecutionContext ctx) {
+                J.Assignment a = super.visitAssignment(assignment, ctx);
 
                 if (a.getVariable() instanceof J.Identifier) {
                     J.Identifier variable = (J.Identifier) a.getVariable();
                     if ("sourceCompatibility".equals(variable.getSimpleName())) {
-                        getCursor().putMessageOnFirstEnclosing(G.CompilationUnit.class, SOURCE_COMPATIBILITY_FOUND, a.getAssignment());
+                        getCursor().getRoot().putMessage(SOURCE_COMPATIBILITY_FOUND, a.getAssignment());
```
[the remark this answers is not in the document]
JohannisK: I've tried using first enclosing SourceFile, but that does not work.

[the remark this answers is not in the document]
jevanlingen: Took this PR over to fix this. I used inline GroovyIsoVisitor and KotlinIsoVisitor to make it work.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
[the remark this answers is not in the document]
jevanlingen: > So what I've seen in other visit methods is to explicitly "re-set" the parent to be the current element again or to build up the child cursor. I've done both.

That's sounds as a hassle as well. Let's leave it in, I see other recipes like [DependencyConstraintToRule](https://github.com/openrewrite/rewrite/blob/b777a6dfcccdd5fcfeaee214649f2a8dd775ba11/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyConstraintToRule.java#L227C48-L227C62) do also use this `isKotlinScript` functionality.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -89,316 +97,389 @@ public Validated<Object> validate() {
...
+                        return m.withArguments(
+                                Collections.singletonList(versionArg.withValue(version)
+                                        .withValueSource(version.toString())));
+                    } else {
+                        return m;
+                    }
                 }
+            }
 
+            return SearchResult.found(m, "Attempted to update to Java version to " + version +
```
Jenson3210, line 279: I like this idea, but could we not instead add a new Marker type that adds a `//TODO: ...` comment to the marked element rather than a SearchResult?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -89,316 +97,389 @@ public Validated<Object> validate() {
...
+        }
+    }
 
-            private String normalize(String version) {
-                if (version.contains("\"") || version.contains("'")) {
-                    version = version.replace("\"", "").replace("'", "");
+    private Expression changeExpression(Expression expression, @Nullable DeclarationStyle style) {
+        if (expression instanceof J.Literal) {
+            J.Literal literal = (J.Literal) expression;
+            if (style == DeclarationStyle.String) {
```
Jenson3210, line 389: can we introduce a method for the styles to avoid the duplications?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
jevanlingen: Maybe, I won't do that in this PR. The current code (on the main branch) does already do this, so let's not go for extra improvements yet. Let's keep it a "Support Kotlin too" PR.

Jenson3210: I did not mean for this PR indeed, but more as a generic improvement issue

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
jevanlingen: https://github.com/openrewrite/rewrite/pull/5512/commits/6f9153fadf661e315f8b801c5697634809e08c00

Jenson3210 (approved):
Nice!

PR #5515, opened 2025-05-28
Made DependencyUseMapNotation kotlin-aware

Jenson3210:
As described here:

- https://github.com/openrewrite/rewrite/issues/5361

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseMapNotation.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseMapNotation.java
@@ -47,7 +47,8 @@ public class DependencyUseMapNotation extends Recipe {
     @Override
     public String getDescription() {
         return "In Gradle, dependencies can be expressed as a `String` like `\"groupId:artifactId:version\"`, " +
-                "or equivalently as a `Map` like `group: 'groupId', name: 'artifactId', version: 'version'`. " +
+                "or equivalently as a `Map` like `group: 'groupId', name: 'artifactId', version: 'version'` (groovy)" +
+                "or `group = \"groupId\", name = \"artifactId\", version = \"version\"` (kotlin). " +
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseMapNotation.java
```diff
@@ -52,137 +53,178 @@ public String getDescription() {
...
-    private static G.MapEntry mapEntry(String key, Expression e) {
-        return new G.MapEntry(
-                randomId(),
-                Space.format(" "),
-                Markers.EMPTY,
-                JRightPadded.build(new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), key, null, null)),
-                e,
-                null
-        );
+    private static <T extends Expression, P extends J, R extends Expression> J.MethodInvocation forStringTemplate(J.MethodInvocation m,
```
Jenson3210, line 183: Note: I am not really happy with the 9-args forStringTemplate method myself but i think it is still better compared to duplicating all this code.

rewrite-gradle/src/test/java/org/openrewrite/gradle/DependencyUseMapNotationTest.java
```diff
@@ -69,6 +70,47 @@ void basicString() {
         );
     }
 
+    @DocumentExample
```
jevanlingen:
```suggestion
```
I don't think multiple `@DocumentExample` do work.

rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseMapNotation.java
```diff
@@ -52,137 +53,178 @@ public String getDescription() {
 
     @Override
```
jevanlingen, line 55: I can't place a comment on line 48 (because it's collapsed), but improve the description a bit. It clearly targeted towards groovy only, so I should be changed slightly to include Kotlin gradle files as well.

PR #5519, opened 2025-05-28
Annotate nullable methods that return from ternary

timtebeek:
- Following https://github.com/openrewrite/rewrite-static-analysis/pull/569


greg-at-moderne:
Post-merge LGTM.

Issue #5498, opened 2025-05-26
`RemoveUnusedImports` adds an invalid import of `java.util.Entry`


greg-at-moderne:
## Original problem

Happened in real-life:
https://github.com/openrewrite/rewrite-hibernate/commit/a4c7d867153c95f8cd46d45a0b082e5b2c455bff
The change was added by Moderne, which was running `RemoveUnusedImports` among others.

## Steps to reproduce

In current main (8261b9038496cb446697b6b30fb82c470a247f02)
the following test:
```java
    @Test
    void javaUtilMapEntry() {
        // language=java
        rewriteRun(
          java(
            """
              import java.util.*;

              public class Usages {
                  HashMap<String, String> hashMap;
                  Map<String, String> map;
                  Optional<String> optional;
                  Map.Entry<String, String> favoriteEntry;
              }
            """,
            """
              import java.util.HashMap;
              import java.util.Map;
              import java.util.Optional;

              public class Usages {
                  HashMap<String, String> hashMap;
                  Map<String, String> map;
                  Optional<String> optional;
                  Map.Entry<String, String> favoriteEntry;
              }
              """
          )
        );
    }
```
- is not idempotent, takes two cycles.
- adds the spurious `import java.util.Entry` in one of the cycles.

PR #5525, opened 2025-05-30
Skip importing inner type if outer class is already imported

e5LA:
## What's changed?
Updated `RemoveUnusedImports` recipe to recognize that nested types (e.g., `Map.Entry`) are covered by the outer class import (`Map`) and do not require separate import statements. 

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5498

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/MoveFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/MoveFile.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


greg-at-moderne (approved):
Thanks for fixing this.

Based on what you have learned and your code. I am wondering if it would work as well with `Map` wouldn't be used by itself. Can you add a similar test case which has usages of `Map.Entry`, but no mention of `Map` "separately"?

e5LA:
Thanks for checking. 
I've added an additional test case to cover the use case you mentioned: https://github.com/openrewrite/rewrite/pull/5525/commits/8f1a6549be1b43d9501ba10f2d6bb37da6e59172


On top of that, there are already few extra tests covering `Map.Entry`: 
- https://github.com/openrewrite/rewrite/blob/f8171dc4ecacdf767a93a4f9242e6dde56e01a10/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java#L105
- https://github.com/openrewrite/rewrite/blob/f8171dc4ecacdf767a93a4f9242e6dde56e01a10/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java#L120
- https://github.com/openrewrite/rewrite/blob/f8171dc4ecacdf767a93a4f9242e6dde56e01a10/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java#L1509

PR #5529, opened 2025-05-30
Fix bug in `ResolvedPom#mergeDependencyManagement()`

knutwannheden:
The merging of managed dependency properties was not quite correct in `ResolvedPom`. I didn't find an easy way to fix it, so the changes are a bit more exhaustive than I was hoping for.

Additionally, the `dependencyManagement` list is now sorted (using TimSort 😄), so that the `getManagedXyz()` operations are faster.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -84,24 +84,18 @@ public class ResolvedPom {
         this.requested = requested;
         this.activeProfiles = activeProfiles;
         this.properties = properties;
-        this.dependencyManagement = sortDependencyManagement(dependencyManagement);
-        this.initialRepositories = initialRepositories;
+        this.dependencyManagement = dependencyManagement;
+        // TO-BE-REMOVED(2025-09-01): sorting added for backwards compatibility with older LSTs
+        this.dependencyManagement.sort(MANAGED_DEPENDENCY_COMPARATOR);
+        if (initialRepositories != null) {
+            this.initialRepositories = initialRepositories;
+        }
         this.repositories = repositories;
         this.pluginRepositories = pluginRepositories;
         this.requestedDependencies = requestedDependencies;
         this.plugins = plugins;
         this.pluginManagement = pluginManagement;
         this.subprojects = subprojects;
-    }
-
-    private static List<ResolvedManagedDependency> sortDependencyManagement(List<ResolvedManagedDependency> sorted) {
-        if (sorted.isEmpty()) {
-            return sorted;
-        }
-
-        // TimSort is O(n) for already sorted data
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -440,35 +456,36 @@ private void resolveParentPropertiesAndRepositoriesRecursively(List<Pom> pomAnce
         }
 
         private void resolveParentDependenciesRecursively(List<Pom> pomAncestry) throws MavenDownloadingException {
-            if(pomAncestry.isEmpty()) {
+            if (pomAncestry.isEmpty()) {
                 return;
             }
-            Map<GroupArtifactScope, ResolvedManagedDependencyDepth> gaToNearest = dependencyManagement.stream()
+            Map<GroupArtifactClassifierType, ResolvedManagedDependencyDepth> managedDependencyMap = dependencyManagement.stream()
```
knutwannheden, line 462: As can be seen here the managed dependencies are now stored in a map, where `(groupId, artifactId, classifier, type)` is used as key rather than `GroupArtifactScope`, which also included the scope (`(groupId, artifactId, classifier, type, scope)`), which was wrong.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -89,6 +94,16 @@ public ResolvedPom(Pom requested, Iterable<String> activeProfiles) {
...
 
+    private static List<ResolvedManagedDependency> sortDependencyManagement(List<ResolvedManagedDependency> sorted) {
+        if (sorted.isEmpty()) {
+            return sorted;
+        }
+
+        // TimSort is O(n) for already sorted data
+        sorted.sort(MANAGED_DEPENDENCY_COMPARATOR);
+        return sorted;
+    }
```
sambsnyd, line 105: Hmm I'm not sure if it is correct / matches Maven's behavior to sort dependency management.
Maven's version conflict resolution strategy is "nearest wins" but if there are two equally-near options it _probably_ comes down to order, although I don't know off the top of my head if it would be first wins or last wins.

If something other than order determines which version of jackson-core wins in a situation like this one then it's probably fine
```
org:transitive-jackson-dep
+- org:depends-on-jackson
|  +- com.fasterxml.jackson.core:jackson-core:2.9.9:compile
+- org:also-depends-on-jackson
   +- com.fasterxml.jackson.core:jackson-core:2.12.0:compile
```

knutwannheden: I think here whichever dependency comes first (here 2.9.9) should win. I don't think the sorting should affect this, as the sorting isn't performed until after the resolution. But it would make sense to add a test for this.

PR #5530, opened 2025-05-30
Regression in `AddImport`

BoykoAlex:
Regression in calculation of import type for nested classes imports: exact, ambiguous, none etc. It related to package name and simple name calculations as these would miss enclosing class names.


timtebeek (approved):
Thanks for the fix here! The test is also helpful to drop future regressions. Surprised we hadn't picked this up with any downstream tests

Issue #5196, opened 2025-03-19
Failure when parsing Javadoc containing `{@link` and no closing brace


greg-at-moderne:
## What version of OpenRewrite are you using?

Current main = 7de8e67cbdef0c67a2c48d505e549d953f1f684d

## What is the smallest, simplest way to reproduce the problem?
```
    @Test
    void unclosedBraceOnLink() {
        rewriteRun(
          java(
            """
              class Test {
                 /**   
                  * {@link int
                  * Some other text.
                  * See {@link java.lang.String}
                  * @param arg description
                  */                  
                  void method(String arg) {
                  }
              }
              """
          )
        );
    }

```
in `JavadocTest`

## What did you see instead / What is the full stack trace of any errors you encountered?
```
Caused by: org.openrewrite.java.JavaParsingException: Failed to convert for the following cursor stack:--- BEGIN PATH ---
JCCompilationUnit(sourceFile = Test.java)
JCClassDecl(name = Test, line = 1)
--- END PATH ---

	... 32 more
Caused by: java.lang.AssertionError
	at org.openrewrite.java.isolated.ReloadableJava17JavadocVisitor.whitespaceBefore(ReloadableJava17JavadocVisitor.java:1074)
	at org.openrewrite.java.isolated.ReloadableJava17JavadocVisitor.visitDocComment(ReloadableJava17JavadocVisitor.java:340)
	at org.openrewrite.java.isolated.ReloadableJava17JavadocVisitor.visitDocComment(ReloadableJava17JavadocVisitor.java:52)
	at com.sun.tools.javac.tree.DCTree$DCDocComment.accept(DCTree.java:138)
	at com.sun.source.util.DocTreeScanner.scan(DocTreeScanner.java:84)
	at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.formatWithCommentTree(ReloadableJava17ParserVisitor.java:2385)
	at org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert(ReloadableJava17ParserVisitor.java:1758)
	... 30 more
```

## Context

- the problem goes away if I close the `{@link` with `}` on the same line. But this mistake is kind of popular when looked through the open-source Java code, so I think we should do a better job handling this admittedly invalid input.

e5LA:
@greg-at-moderne I can’t reproduce this on `main` branch — seems like it’s already fixed. Can you please confirm on your end?

e5LA:
[...]

Issue #3650, opened 2023-10-30
edge cases for ReloadableJava17JavadocVisitor


bnaudts:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite 8.9.0-SNAPSHOT, main branch, commit [0d884d4](https://github.com/openrewrite/rewrite/commit/0d884d4d6a0c15591bae6111b166029053e34f05)

## How are you running OpenRewrite?
In IntelliJ, but also using the latest Gradle plugin, latest SNAPSHOT version.

## What is the smallest, simplest way to reproduce the problem?

The following are a number of tests that can be inserted into `org.openrewrite.java.tree.JavadocTest`.
Each detail a minor JavaDoc parsing issue.

```java

    @Test
    void unicodeAmpersand() {
        rewriteRun(
          java("""
            package com.foo;

             /**
              * {@code
              * \\u0040Override
              * public void method() {
              * }
              */
             @SuppressWarnings
             class Bar { }
             """)
        );
    }

    @Test
    void unicodeAccent() {
        rewriteRun(
          java("""
            package com.foo;

             /**
              * @author Xx\\u00e9
             */
             class Bar { }
             """)
        );
    }

    @Test
    void badlyClosedXmlTags() {
        rewriteRun(
          java("""
            package com.foo;
            /**
             <!--Optional:->
             <urn:portalId>?</urn:portalId-->
            */

            class Bar { }
            """)
        );
    }

    @Test
    void unicodeSpace() {
        rewriteRun(java("""
          package com.foo;

          /**
           * Strip characters that are not allowed in an XML document : those below ASCII 32 (HEX 20 -> in java < '\\u0020')
           */
          public class Bar { }
          """));
    }

    @Test
    void splitAtDot() {
        rewriteRun(java("""
          package com.foo;
          	/**
          	 * @see a.b.Type#method(a.b.c.
          	 * d.Type)
          	 */
          class Bar { }
          """));
    }

    @Test
    void combinedCommentsWithoutSpace() {
        // solved by adding a space between */ and /*
        rewriteRun(java("""
          package com.foo;
          class Bar {
          	/**
          	 * xx
          	 *//*
               void method() {
               }*/
          }
          """));
    }

    @Test
    void spaceStarAfterStartOfJavaDoc() {
        // remove the * after /**
        rewriteRun(java("""
          package com.foo;
           /**	 *
            * @author x
            */
          class Bar { }
          """));
    }

    @Test
[...]

PR #5531, opened 2025-05-31
Handling unterminated inline tags

e5LA:
## What's changed?
First, I added tests related to issues (https://github.com/openrewrite/rewrite/pull/5531/commits/20175de4b97861132d9773f9f668e6094c41434f): 
- #5196 and
- #3650

Then it turned out that tests failed for Java 8, 11, 17

Therefore I applied the fix (https://github.com/openrewrite/rewrite/pull/5531/commits/a6dbe422a0794664f51ef51840ac344625e303a6) for #5196 

## Any additional context
In Java 21, `DCLink` contains:
`{DCTree$DCErroneous@5277} "Some other text.\n See {@link java.lang.String}\n @param arg description"`

While in lower versions, it contains:
`{DCTree$DCErroneous@5338} "Some other text.\n See {@link java.lang.String}"`
 
that's the main reason why the test passed on Java 21 but failed below.

## What's your motivation?
Initially, as suggested by @greg-at-moderne in https://github.com/openrewrite/rewrite/issues/5196#issuecomment-2922440000, adding these test cases will help prevent regressions in the future.

Then realized that in fact #5196 still requires a fix for lower versions of java.

### Checklist


rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
```diff
@@ -1087,9 +1087,14 @@ private List<Javadoc> endBrace() {
...
             List<Javadoc> end = whitespaceBefore();
-            if (cursor < source.length() && source.charAt(cursor) == '}') {
-                end = ListUtils.concat(end, new Javadoc.Text(randomId(), Markers.EMPTY, "}"));
-                cursor++;
+            if (cursor < source.length()) {
+                boolean containsEndLine = end.stream().anyMatch(p -> p instanceof Javadoc.LineBreak);
+                if (source.charAt(cursor) == '}') {
+                    end = ListUtils.concat(end, new Javadoc.Text(randomId(), Markers.EMPTY, "}"));
+                    cursor++;
+                } else if (containsEndLine) {
```
e5LA, line 1095: Not sure if this is the best way to handle it.. We expect a closing `}`, but get a newline instead (as the javadoc is invalid)

e5LA:
Thanks for checking it out @timtebeek - I’ve updated the PR description with more context. I initially thought it would just be about adding tests, but it turned out the issue required an actual fix for lower versions of java.

Please let me know what you think about the approach.

rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
greg-at-moderne: I think this is what exactly needs to be checked in this specific case.

greg-at-moderne (approved):
Thanks so much for your engagement here!

PR #5535, opened 2025-06-01
Handle invalid JavaDoc references in parser and printer

### LINKED ISSUE #5534: Rewrite calls from deprecated API to target API
Certain API got deprecated **in favor of something new**. There is a replacement, and we need to define the rules on how old, deprecated API calls are transitioned into a new call, that is open to discussion

We have calls to an old API. The old API produces the type A. The old API produces also either the type A, or any of its children. So it is backward compatible **on the source code level**. But the input parameters are either changed, or more qualified, or re-arranged or etc.

```java
class A {

    void foo(String a, String b) {
        bar(a, b);
    }
    
    /**
     * @deprecated in favor of #bar(Integer, Boolean)
     */
    @Deprecated
    void bar(Integer int, String value) {
    }

    void bar(Integer int, Boolean value) {
    }
}
```

```java
class A {

    void foo(String a, String b) {
        bar(a, Boolean.valueOf(b));
    }
    
    /**
     * @deprecated in favor of #bar(Integer, Boolean)
     */
    @Deprecated
    void bar(Integer int, String value) {
    }

    void bar(Integer int, Boolean value) {
    }
}
```

[Based on documentation](https://docs.openrewrite.org/recipes), there are no workarounds to this in openrewrite, yet.

Yes, I would like to

timtebeek:
- Fixes #5534

### DIFF
--- a/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17JavadocVisitor.java
+++ b/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17JavadocVisitor.java
@@ -668,7 +668,7 @@ public class ReloadableJava17JavadocVisitor extends DocTreeScanner<Tree, List<Ja
                         randomId(),
                         qualifier == null ? Space.EMPTY : qualifier.getPrefix(),
                         Markers.EMPTY,
-                        qualifier == null ? null : JRightPadded.build(qualifier.withPrefix(Space.EMPTY)),
+                        JRightPadded.build(qualifier == null ? null : qualifier.withPrefix(Space.EMPTY)),
                         JContainer.empty(),
                         JLeftPadded.build(name),
                         null,
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavadocPrinter.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavadocPrinter.java
@@ -397,8 +397,13 @@ public class JavadocPrinter<P> extends JavadocVisitor<PrintOutputCapture<P>> {
         @Override
         public J visitMemberReference(J.MemberReference memberRef, PrintOutputCapture<P> p) {
             beforeSyntax(memberRef, Space.Location.MEMBER_REFERENCE_PREFIX, p);
-            visit(memberRef.getContaining(), p);
-            visitLeftPadded("#", memberRef.getPadding().getReference(), JLeftPadded.Location.MEMBER_REFERENCE_NAME, p);
+            Expression containing = memberRef.getContaining();
+            if (containing != null) { // Invalid references will have a null containing
+                visit(containing, p);
+                visitLeftPadded("#", memberRef.getPadding().getReference(), JLeftPadded.Location.MEMBER_REFERENCE_NAME, p);
+            } else {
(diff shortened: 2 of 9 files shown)


rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17JavadocVisitor.java
```diff
@@ -668,7 +668,7 @@ public Tree visitProvides(ProvidesTree node, List<Javadoc> body) {
                         randomId(),
                         qualifier == null ? Space.EMPTY : qualifier.getPrefix(),
                         Markers.EMPTY,
-                        qualifier == null ? null : JRightPadded.build(qualifier.withPrefix(Space.EMPTY)),
+                        JRightPadded.build(qualifier == null ? null : qualifier.withPrefix(Space.EMPTY)),
```
timtebeek, line 671: This mimics what we do for the Java parser: there we also unconditionally create a JRightPadded with a nullable containing element.

rewrite-java/src/main/java/org/openrewrite/java/JavadocPrinter.java
```diff
@@ -397,8 +397,13 @@ public J visitFieldAccess(J.FieldAccess fieldAccess, PrintOutputCapture<P> p) {
...
         public J visitMemberReference(J.MemberReference memberRef, PrintOutputCapture<P> p) {
             beforeSyntax(memberRef, Space.Location.MEMBER_REFERENCE_PREFIX, p);
-            visit(memberRef.getContaining(), p);
-            visitLeftPadded("#", memberRef.getPadding().getReference(), JLeftPadded.Location.MEMBER_REFERENCE_NAME, p);
+            Expression containing = memberRef.getContaining();
+            if (containing != null) { // Invalid references will have a null containing
+                visit(containing, p);
+                visitLeftPadded("#", memberRef.getPadding().getReference(), JLeftPadded.Location.MEMBER_REFERENCE_NAME, p);
+            } else {
+                visitLeftPadded(null, memberRef.getPadding().getReference(), JLeftPadded.Location.MEMBER_REFERENCE_NAME, p);
```
timtebeek, line 405: These `visitLeftPadded` might be a little too hyper focused on the failing test that was provided, but still I think it's common enough to keep this as is here. If we were to unconditionally use a `#` prefix, then that would be another print idempotency issue.

greg-at-moderne:
```
JavadocTest > parsingIncorrectJavadocValueReference() FAILED
    Identifier->InlinedValue->DocComment->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
    /*~~(Identifier type is missing or malformed)~~>*/BAR
    https://docs.openrewrite.org/reference/faq#im-seeing-lst-contains-missing-or-invalid-type-information-in-my-recipe-unit-tests-how-to-resolve
 ```
    is the CI failure.

timtebeek:
Thanks! Weird how those only show on Java 11 and 17; they are true, and expected, so I've updated the recipe spec to allow for those, as indeed we're using an invalid reference; we're aiming for print idempotency, not type fixes.

greg-at-moderne (approved):
LGTM

rewrite-java/src/main/java/org/openrewrite/java/JavadocPrinter.java
greg-at-moderne: I think I see what you mean here. But I see no better option.

PR #5540, opened 2025-06-02
Added test showing incompatibility between ChangeType and RenameVariable

JohannisK:
## What's changed?
Added test showing incompatibility between ChangeType and RenameVariable

## What's your motivation?
When using `ChangeType`, change type tries to rename the variable through `RenameVariable`. Normally RenameVariable renames all relevant variables, but when invoked through `ChangeType` it seems to only rename the variable where it is declared.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
@@ -15,11 +15,9 @@
  */
 package org.openrewrite.java;
 
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
-import org.openrewrite.DocumentExample;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Issue;
-import org.openrewrite.Recipe;
+import org.openrewrite.*;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.Statement;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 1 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
```diff
@@ -1427,4 +1425,87 @@ public void blocks() {
...
+        @Test
+        void replacingTheNameAsAnEffectOfChangeTypeFails() {
+            rewriteRun(
+              spec -> spec.recipe(toRecipe(() -> new JavaVisitor<>() {
+                  @Override
+                  public J visit(Tree tree, ExecutionContext ctx) {
+                      J t = super.visit(tree, ctx);
+                      Tree visited = new ChangeType("java.util.Stack", "java.util.ArrayDeque", false)
+                        .getVisitor().visit(t, ctx, getCursor());
+                      return (J) visited;
```
greg-at-moderne: Since it's `visit(Tree)` method, this test case causes the `ChangeType` recipe to be called several times with various scopes, right?

JohannisK: Oeps... yeah.

I've changed it to test a few specific scopes, and you clearly see the issue. 
On one hand it's good to be able to rename variables in an isolated scope, but I think the rename variables recipe should broaden it's scope to make all the necessary "transitive" changes outside the provided scope.

Right now it just operates in the provided scope.

timtebeek:
The underlying issue has seen been fixed in a different way, so I think we can close this effort for now
- https://github.com/openrewrite/rewrite-static-analysis/issues/582
- https://github.com/openrewrite/rewrite-static-analysis/commit/0b4b4c308c2cc09b8650ce95a611e6c655fde050
- https://github.com/openrewrite/rewrite/pull/5454

PR #5547, opened 2025-06-03
Added support for dependency versions in buildscript.ext {}

JohannisK:
## What's changed?
- Added test for UpgradeDependencyVersion with variables in `buildscript.ext {}` when using dependencies.gradle
- Added implementation
- Added support for string concatenations when left is a dependency and right is a version variable

## What's your motivation?
- Although discouraged there are still usecases when buildscript.ext {} is used, and the use is not officially deprecated


shanman190:
I'm not sure if this test case captures a valid Gradle build. For some reason I think the buildscript.ext is treated separately from project.ext without overlap.

Secondarily, variable substitution was intentionally limited to the current build descriptor that contains a dependency that utilizes the variable (effectively enforcing scoping rules) and gradle.properties because otherwise we have no linkage to knowing that we're updating the correct variable or just some other random one that just so happens to have the same name.

rewrite-gradle/src/main/java/org/openrewrite/gradle/Assertions.java
```diff
@@ -181,6 +181,52 @@ public static SourceSpecs settingsGradleKts(@Language("kotlin") @Nullable String
         return gradle;
     }
 
+    public static SourceSpecs dependenciesGradle(@Language("groovy") @Nullable String before) {
```
shanman190: In my opinion, I'm hesitant to add these additional helper methods because these are for freestanding scripts which have no naming requirements imposed upon them and can be named literally anything. I personally feel that the `spec.path(..)` shows that distinction better than adding API surface area here.

JohannisK: Done :)

JohannisK:
> I'm not sure if this test case captures a valid Gradle build. For some reason I think the buildscript.ext is treated separately from project.ext without overlap.
> 
> Secondarily, variable substitution was intentionally limited to the current build descriptor that contains a dependency that utilizes the variable (effectively enforcing scoping rules) and gradle.properties because otherwise we have no linkage to knowing that we're updating the correct variable or just some other random one that just so happens to have the same name.

I'm not that big of a gradle expert, but what I found so far is that variables declared in `buildscript.ext {}` should apply to dependencies in `dependencies.gradle`. The reported issue also contained examples showing the scenario is used.

I could expand on the solution, by actually keeping track of files where dependencies are declared in the scanner (so I know `implementation "com.google.guava:guava:${guavaVersion}"` was declared in `dependencies.gradle`) and then retrace those steps when I encounter `apply from: 'dependencies.gradle'` in the `build.gradle`

shanman190:
Ok, so researched this a little bit more.

```
buildscript { // receiver = org.gradle.api.Project
  ext { // receiver = org.gradle.api.Project (extension = ext, implementation class = org.gradle.api.plugins.ExtraPropertiesExtension)
    guavaVersion = "29.0-jre" // receiver = org.gradle.api.plugins.ExtraPropertiesExtension
  }
}
```

So any of these properties are going to be project specific and the imported standalone script would have it's version be directly dependent upon the current project's state of their extension. 

So as an illustration:

_settings.gradle_
```
include("projectA")
include("projectB")
```


_projectA/build.gradle_
```
ext {
  guavaVersion = "29.0-jre"
}

apply from: "dependencies.gradle"
```


_projectB/build.gradle_
```
ext {
  guavaVersion = "30.0-jre"
}

apply from: "dependencies.gradle"
```

The above project is valid, but each project would include a different version of Guava because the extra properties are evaluated individually within the context of the script application. This is one of the challenges with respect to global dataflow analysis in terms of Gradle scripts.

JohannisK:
@shanman190 , but doesn't that mean in this context we can safely update the version in the ext block?

shanman190:
It's probably ok. We just have to realize that we're tiptoeing into global analysis land here and that may come with some potential risks.

Some examples:
* What repositories do you search when looking for a new version? Each project may declare a different set, in one repository a newer version may be available that isn't available in another project (honestly unlikely), etc. 
* Extra properties don't inherit from parent projects, so in the above example there are two distinct and independent dependency resolution contexts.
* The cross section of preconditions only selecting part of the repository and the `dependencies.gradle` (or another included script) existing outside of that context.

sambsnyd (approved):

JohannisK:
Discussed with @sambsnyd and we think the recipe will bring value and the risk is low.
Requested an extra opinion from @jkschneider , but we feel like we should proceed with the PR.

PR #5549, opened 2025-06-03
Fixed that the YAML DeleteKey does not remove unused parent keys in all situations

mzeijen:
## What's changed?
This solves that the YAML `DeleteKey` recipe doesn't always remove unused parent keys, specifically in combination with a regex match and sequence entries. 

I encountered the issue in the following scenario:

**When** the `DeleteKey` recipe is used with  a JSON path that matches using a regex on a child value of a YAML sequence entry
**And** this results in the sequence becoming empty
**Then** the sequence key and any empty parent keys are not removed
**However** I expected them to be removed, because this does happen when you use an exact match on the child value

The added `deleteSequenceEntriesUsingChildRegexMatchRemovingUnusedKeysRecursively` unit test method shows the expected behavior. Without this fix the `after` looks like this:

```yaml
a: a-value
b:
  c:
```

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/DeleteKeyTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/DeleteKeyTest.java
@@ -224,4 +224,46 @@ class DeleteKeyTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void deleteSequenceEntryUsingChildRegexMatch() {
+        rewriteRun(
+          spec -> spec.recipe(new DeleteKey("$.b.c[?(@.name =~ 'b1-.*')]", null)),
+          yaml(
+            """
+                  a: a-value
+                  b:
+                    c:
+                      - name: b1-value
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/DeleteKeyTest.java
```diff
@@ -224,4 +224,46 @@ void deleteKeyKeepingUnrelatedUnusedKeys() {
...
+            """
+                  a: a-value
+                  b:
+                    c:
+                      - name: b1-value
+                      - name: b2-value
+              """,
+            """
+              a: a-value
+              """
```
greg-at-moderne:
```suggestion
            """
            a: a-value
            b:
              c:
                - name: b1-value
                - name: b2-value
            """,
            """
            a: a-value
            """
```

timtebeek (approved):
Thanks a lot for the thoughtful fix here. Indeed best to clear these out.

PR #5550, opened 2025-06-03
Support `MarkerPrinter` in RPC `Print` request

knutwannheden:


knutwannheden:
@jetbrains-junie Please review.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -204,32 +255,50 @@ public Recipe prepareRecipe(String id, Map<String, Object> options) {
         return new RpcRecipe(this, r.getId(), r.getDescriptor(), r.getEditVisitor(), r.getScanVisitor());
     }
 
-    public List<SourceFile> parse(String parser, Iterable<Parser.Input> inputs, @Nullable Path relativeTo) {
+    public List<SourceFile> parse(Parser parser, String language, Iterable<Parser.Input> inputs, @Nullable Path relativeTo, ExecutionContext ctx) {
```
jkschneider: try to remove `language` until such a time as a single RPC server supports parsing multiple languages.

jkschneider: consider changing `ParseError#build` to not require a full `Parser` instance.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -204,32 +255,50 @@ public Recipe prepareRecipe(String id, Map<String, Object> options) {
...
         }
         return emptyList();
     }
 
     public String print(SourceFile tree) {
-        return print(tree, new Cursor(null, Cursor.ROOT_VALUE));
+        return print(tree, new Cursor(null, Cursor.ROOT_VALUE), null);
+    }
+
+    public String print(SourceFile tree, Print.MarkerPrinter markerPrinter) {
```
jkschneider: Should be `@Nullable` if the 3 arg for is as well.

rewrite-javascript/rewrite/src/rpc/request/print.ts
```diff
@@ -15,11 +15,17 @@
...
 
+export const enum MarkerPrinter {
+    DEFAULT = "DEFAULT",
+    FENCED = "FENCED",
+    SANITIZED = "SANITIZED"
+}
+
 export class Print {
-    constructor(private readonly treeId: UUID, private readonly cursor?: string[]) {
+    constructor(private readonly treeId: UUID, private readonly cursor?: string[], readonly markerPrinter?: MarkerPrinter) {
```
jkschneider: Can we say `= PrintMarkerPrinter.DEFAULT` and remove the `?`?

rewrite-javascript/src/integTest/java/org/openrewrite/javascript/rpc/JavaScriptRewriteRpcTest.java
```diff
@@ -56,14 +55,11 @@ class JavaScriptRewriteRpcTest implements RewriteTest {
     @BeforeEach
     void before() throws FileNotFoundException {
         this.log = new PrintStream(new FileOutputStream("rpc.java.log"));
-        this.client = JavaScriptRewriteRpc.start(
-          Environment.builder().build(),
-          "node",
-          "--enable-source-maps",
-          // Uncomment this to debug the server
-//          "--inspect-brk",
```
jkschneider, line 64: Add builder option for `--inspect-brk` to be able to debug in IDE.

rewrite-javascript/src/main/java/org/openrewrite/javascript/JavaScriptParser.java
```diff
@@ -19,28 +19,41 @@
...
-        this.client = client;
+    private JavaScriptParser(JavaScriptRewriteRpc rewriteRpc) {
+        this.rewriteRpc = rewriteRpc;
     }
 
     @Override
     public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path relativeTo, ExecutionContext ctx) {
-        return client.parse("javascript", sources, relativeTo).stream();
+        JavaScriptValidator<Integer> validator = new JavaScriptValidator<>();
+        return rewriteRpc.parse(this, "javascript", sources, relativeTo, ctx).stream()
```
jkschneider: Should `parse` on `RewriteRpc` return a stream to support large source sets?

rewrite-javascript/src/main/java/org/openrewrite/javascript/rpc/JavaScriptRewriteRpc.java
```diff
@@ -112,16 +104,189 @@ public int installRecipes(String packageName, @Nullable String version) {
...
+
+        private Environment marketplace = Environment.builder().build();
+        private Path nodePath = Paths.get("node");
+        private @Nullable Path installationDirectory;
+        private int port;
+        private boolean trace;
+        private @Nullable Path logFile;
+        private @Nullable Duration timeout;
+
+        private static Path getDefaultInstallationDirectory() {
```
jkschneider: Discussed, and we agree we should move toward bundling instead.

rewrite-javascript/src/main/java/org/openrewrite/javascript/rpc/JavaScriptRewriteRpc.java
```diff
@@ -112,16 +104,189 @@ public int installRecipes(String packageName, @Nullable String version) {
...
+            this.installationDirectory = installationDirectory;
+            return this;
+        }
+
+        public Builder socket(int port) {
+            this.port = port;
+            return this;
+        }
+
+        public Builder trace(boolean trace) {
```
jkschneider, line 204: Change this to `MessageHandler` instead, which would flow down to an argument on `JsonRpc` as a substitute for the `trace` boolean flag. In addition to `TraceMessageHandler`, we may want to choose an alternative `MessageHandler` to change the over-the-wire serialization format to msgpack, etc.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
knutwannheden: For now I will leave the `Parser` parameter, as I am also using it to do the print-idempotence check.

rewrite-javascript/src/main/java/org/openrewrite/javascript/internal/rpc/JavaScriptValidator.java
```diff
@@ -0,0 +1,945 @@
...
+
+    private <T extends Tree> @Nullable T visitAndValidate(@Nullable T tree, Class<? extends Tree> expected, P p) {
+        if (tree != null && !expected.isInstance(tree)) {
+            throw new ClassCastException("Type " + tree.getClass() + " is not assignable to " + expected);
+        }
+        // noinspection unchecked
+        return (T) visit(tree, p);
+    }
+
+    private <T extends Tree> T visitAndValidateNonNull(@Nullable T tree, Class<? extends Tree> expected, P p) {
```
greg-at-moderne, line 38: Don't we need the same validation on the TS side?

greg-at-moderne (approved):
The code looks fine to me.
But I need to disclose I have no experience with OR RPC technology, so I am likely missing a lot of context of what to look at.

rewrite-javascript/src/main/java/org/openrewrite/javascript/internal/rpc/JavaScriptValidator.java
knutwannheden: We could implement the validation on the JS side instead (or as well). For me it was simpler to generate this code for Java. The idea behind this class is to use it to validate the parser output, while the parser is still under development and considered immature. We want to avoid situations where an LST model can't be serialized by Jackson because it is strict about the runtime types of objects, whereas the JVM doesn't care when there is type erasure.

At some point we might disable this thing or maybe even enable it as part of the recipe scheduler to prevent cases when a recipe might break an LST model.

greg-at-moderne:
CI complains about some implicit dependencies in Gradle task definitions.

PR #5551, opened 2025-06-03
Retrieving all licenses currently used in recipe jars

Jenson3210:
In order to be able to build in compliance checks in plugins, we need to be able to extract what licenses are required by the recipe jars currently present. 

Goal is then to in these plugins do the necessary checks if we adhere to the licenses present (Skipping validations for eg. Apache as there are none there).

### Checklist


rewrite-core/build.gradle.kts
```diff
@@ -25,5 +25,6 @@ dependencies {
     implementation("org.objenesis:objenesis:latest.release")
 
     testImplementation("org.assertj:assertj-core:latest.release")
+    testImplementation("org.openrewrite.recipe:rewrite-hibernate:2.7.0") // For testing license detection TODO: can this stay here? Do we have the license to use this licensed one?
```
timtebeek: Had brief look whether test dependencies would show up as such in the pom, but thankfully not:
https://repo1.maven.org/maven2/org/openrewrite/rewrite-core/8.54.0/rewrite-core-8.54.0.pom

Still, some folks/scanners might get nervous about seeing a non-Apache licensed dependency here; can we work around this somehow?

Jenson3210: I could adapt the test now that I know it works and it then only returns Apache license?

rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -248,6 +289,12 @@ public Map<String, List<RecipeExample>> listRecipeExamples() {
         return recipeExamples;
     }
 
+    @Override
+    public Set<License> getRequiredLicenses() {
```
timtebeek: Slightl wonder about the naming here: would it make sense to conform to what we have for `listContributors` and `listRecipeExamples` ? 
We tend to use `get` only when it's a "pure" get, not when there might be calculation involved. Then the `Required` part here only applies in the context of checking compliance on other projects downstream.

jkschneider:
Licenses should be facts about recipes, not about a `ResourceLoader`. Reverted this design with 2b9bc08784f7f7c0e265f268b783f55daba620fb.

PR #5554, opened 2025-06-03
Support `file://` URL in `RemoteFile` for use in `CreateYamlFile` url

BoykoAlex:


timtebeek (commented):
Thanks! I had a brief look at how we handle this for `RemoteArchive` as compared to `RemoteFile`; there we seem to handle it in the class itself:
https://github.com/openrewrite/rewrite/blob/d33ff3f5a96e754aeefc7f7135544c3b7b46576f/rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java#L91-L98

I wonder if that would make the most sense here as well, as it's hardly Yaml specific.

Could you explain a bit of how you plan to use this?

BoykoAlex:
@timtebeek in other words you'd make a similar change to `RemoteFile`?
The plan is to support `file` URLs with `fileContentsUrl` parameter in the `CreateYamlFile` recipe

rewrite-core/src/main/java/org/openrewrite/remote/RemoteFile.java
```diff
@@ -68,6 +69,9 @@ public InputStream getInputStream(ExecutionContext ctx) {
         RemoteArtifactCache cache = RemoteExecutionContextView.view(ctx).getArtifactCache();
         try {
             Path localFile = cache.compute(uri, () -> {
+                if ("file".equals(uri.getScheme())) {
+                    return Files.newInputStream(Paths.get(uri));
+                }
```
timtebeek, line 74: Thanks! This indeed matches what we do for RemoteArchive already; highlighting that here for colleagues:
https://github.com/openrewrite/rewrite/blob/982ad13ca1282bc9831864670f8bb800454c8e04/rewrite-core/src/main/java/org/openrewrite/remote/RemoteArchive.java#L94-L97

BoykoAlex:
@timtebeek any news on this PR? Still evaluating?

timtebeek:
Hmm; given the feedback received internally we think it's best not to support local file URLs just yet here. Thanks for the suggestion though! I hope you're able to make this work with local only changes there.

PR #5556, opened 2025-06-04
refactor: OpenRewrite Recipe best practices

timtebeek:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.recipes.rewrite.OpenRewriteRecipeBestPractices?organizationId=T3BlblJld3JpdGU%3D


rewrite-core/src/main/java/org/openrewrite/rpc/RpcReceiveQueue.java
```diff
@@ -100,7 +100,7 @@ public <T> T receive(@Nullable T before) {
      * @return The received value.
      */
     @SuppressWarnings("DataFlowIssue")
-    public <T> T receive(@Nullable T before, @Nullable UnaryOperator<T> onChange) {
+    public <T> @Nullable T receive(@Nullable T before, @Nullable UnaryOperator<T> onChange) {
```
timtebeek, line 103: Figured call out this change: it's correct based on the code below, but I haven't seen this API used much before.

rewrite-javascript/src/integTest/java/org/openrewrite/javascript/JavaScriptParserTest.java
```diff
@@ -73,10 +72,9 @@ void helloTypeScript() {
         Parser.Input input = Parser.Input.fromString(Paths.get("helloworld.ts"), helloWorld);
         Optional<SourceFile> typescript = parser.parseInputs(List.of(input), Paths.get("helloworld.ts").toAbsolutePath().getParent(), new InMemoryExecutionContext()).findFirst();
         assertThat(typescript).containsInstanceOf(JS.CompilationUnit.class);
-        assertThat(typescript.get()).satisfies(cu -> {
+        assertThat(typescript.get()).satisfies(cu ->
 //            assertThat(cu.printAll()).isEqualTo(helloWorld);
-            assertThat(cu.getSourcePath()).isEqualTo(input.getPath());
-        });
+            assertThat(cu.getSourcePath()).isEqualTo(input.getPath()));
     }
```
greg-at-moderne, line 78: I don't mind this change, but an interesting case.
Replacing a block with a single expression, when in the middle there's a commented out line. If it wasn't commented, a block would be needed.

But I hate commented out code. So consider this just an interesting remark.

timtebeek: Yes noticed that as well; had a brief look at the recipe used there, but that didn't immediately replicate there
```java
@Test
void retainComment() {
    rewriteRun(
      //language=java
      java(
        """
          import java.util.function.Function;
          class Test {
              Function<Integer, Integer> f = n -> {
                  // Comment
                  return n+1;
              };
          }
          """,
        """
          import java.util.function.Function;
          class Test {
              Function<Integer, Integer> f = n ->
                  // Comment
                  n+1;
          }
          """
      )
    );
}
```
Would need a closer look.

PR #5559, opened 2025-06-04
Enhance `ListUtils`

MBoegers:
## What's changed?

`ListUtils` provides an iterative alternative to the Stream API. It also provides a `map` method, but this method only maps values to the same type. 
To transform a List of `J.VariableDeclarations` to its `JavaType.FullyQualified` we had no method available.
I propose `ListUtils.transform` to allow mappings between Java types simlar to the `Stream#map` method. I chose the name `transform` to avoid name clashes and denote the semantically difference. Transform has to return a new instance, even if provided with `Function.identity()`

## What's your motivation?

Allow mappings of data and allow type changes at the same time.

Stream API or redefining helper methods in recipes.

## Any additional context

Would help to get rid of the Stream API usage in https://github.com/openrewrite/rewrite/pull/5458

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
@@ -570,7 +570,7 @@ public final class ListUtils {
             return emptyList();
         }
 
-        HashSet<T> newS = new HashSet<>(ls);
+        Set<T> newS = new LinkedHashSet<>(ls);
         return new ArrayList<>(newS);
     }
 }


rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
```diff
@@ -517,4 +515,62 @@ public static <T> List<T> insertAll(@Nullable List<T> ls, int index, @Nullable L
...
+     * @param <T> The type of elements in the list.
+     * @return A new list with distinct elements.
+     */
+    @Contract(pure = true, value = "_ -> !null")
+    public static <T> List<T> distinct(@Nullable List<T> ls) {
+        if (ls == null || ls.isEmpty()) {
+            return emptyList();
+        }
+
+        HashSet<T> newS = new HashSet<>(ls);
```
knutwannheden: I think it makes sense to retain the element order from the input list.
```suggestion
        Set<T> newS = new LinkedHashSet<>(ls);
```

PR #5563, opened 2025-06-04
Tests that demonstrate the correct runtime behavior of the value returned as estimated effort

bmuschko:
## What's changed?

This PR only adds tests to the code base. The tests verify the correct behavior of the method `Recipe#getEstimatedEffortPerOccurrence`. This is a follow up to PR https://github.com/openrewrite/rewrite/pull/5064.

## What's your motivation?

I want to make sure that the value returned by the method `Recipe#getEstimatedEffortPerOccurrence` is correctly reflected in the `SourcesFileResults` report. There are  a couple of use cases to be covered:

1. A recipe doesn't make any changes. No `SourcesFileResults` report will be created.
2. A recipe doesn't make any changes but generates custom data tables. No `SourcesFileResults` report will be created.
3. A recipe makes changes to an existing file. Return the default estimated effort and reflect it in the `SourcesFileResults` report. If the recipe author overrode `Recipe#getEstimatedEffortPerOccurrence` then the returned value will be reflected in the `SourcesFileResults` report.
4. A recipe only generates new files and doesn't modify existing ones. Return the default estimated effort and reflect it in the `SourcesFileResults` report. If the recipe author overrode `Recipe#getEstimatedEffortPerOccurrence` then the returned value will be reflected in the `SourcesFileResults` report.

Based on the logic demonstrated by the test cases, can you clarify if the SaaS will simply read from the values in the `SourcesFileResults` report and create the sum of it. That sum will then be rendered in the UI. Is that a correct assumption?

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/RecipeEstimatedEffortTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/RecipeEstimatedEffortTest.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
+
+import com.opencsv.CSVReaderHeaderAware;
+import com.opencsv.exceptions.CsvValidationException;
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.marker.AlreadyReplaced;
+import org.openrewrite.marker.GitProvenance;
+import org.openrewrite.marker.Marker;
+import org.openrewrite.table.DistinctGitProvenance;
+import org.openrewrite.table.RecipeRunStats;
+import org.openrewrite.table.SourcesFileResults;
+import org.openrewrite.test.RewriteTest;
+import org.openrewrite.text.*;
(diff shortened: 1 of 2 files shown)


bmuschko:
To verify the correct behavior, I ran the `org.openrewrite.text.CreateTextFile` recipe used in the test cases with the Moderne CLI and the SaaS.

From the Moderne CLI, the `SourceFileResults` data table contains the estimated time savings of 300 seconds.

image

Running the same recipe from the SaaS produces 0 estimated time savings.

image

Correct me if I am wrong, but it seems that the SaaS doesn't properly parse, calculate, or treat the estimated time savings. Or my assumption that the estimated time saving is used from the `SourcesFileResults` data table is wrong.

rewrite-core/src/test/java/org/openrewrite/RecipeEstimatedEffortTest.java
```diff
@@ -0,0 +1,324 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite;
+
+import com.opencsv.CSVReaderHeaderAware;
+import com.opencsv.exceptions.CsvValidationException;
```
timtebeek: Instead of this library, were you aware of the RecipeSpec methods to validate data table rows produced?
https://github.com/openrewrite/rewrite/blob/5f2d7e4c39876a987190278690ce3342ce0e0065/rewrite-test/src/main/java/org/openrewrite/test/RecipeSpec.java#L209-L252

bmuschko: I wasn't aware of that. Thanks for the pointer. Are you planning to merge those tests? If yes, I'll change the code to use the RecipeSpec methods.

timtebeek: Up to you whether you'd want to see this merged or not; I think you've effectively shown the problem lies elsewhere, but we could merge these still to codify the expectations.

bmuschko: I cleaned up the code. Let's merge the tests in case this topic comes up again.

PR #5564, opened 2025-06-05
Move accepting licenses to Environment$Builder / Environment

Jenson3210:
A first swing at moving license acceptations as close enough to core so that all the tooling modules can make use of it.

The empty stacktrace is debatable if this is a practice we want to do, but in this case it does result in cleaner error message AND the class name is printed -> should not be very hard to figure out where it is thrown. 

Screenshot 2025-06-05 at 09 52 41

### Checklist


sambsnyd:
I think this is a good start. Agree that the empty stack trace makes for clearer messaging.

PR #5565, opened 2025-06-05
Javascript - Literal extends VariableDeclarator

greg-at-moderne:
## What's changed?

Making Javascript's Literal extend VariableDeclarator too.

## What's your motivation?

In short - because this is actually happening in TS/JS.

Longer answer:

Fixing this error:
```
class org.openrewrite.java.tree.J$Literal cannot be cast to class org.openrewrite.java.tree.VariableDeclarator (org.openrewrite.java.tree.J$Literal and org.openrewrite.java.tree.VariableDeclarator are in unnamed module of loader 'app')
java.lang.ClassCastException: class org.openrewrite.java.tree.J$Literal cannot be cast to class org.openrewrite.java.tree.VariableDeclarator (org.openrewrite.java.tree.J$Literal and org.openrewrite.java.tree.VariableDeclarator are in unnamed module of loader 'app')
	at org.openrewrite.java.internal.rpc.JavaReceiver$$Lambda/0x000000d001234d78.apply(Unknown Source)
	at org.openrewrite.rpc.RpcReceiveQueue.receive(RpcReceiveQueue.java:133)
	at org.openrewrite.java.internal.rpc.JavaReceiver.visitVariable(JavaReceiver.java:567)
	at org.openrewrite.java.internal.rpc.JavaReceiver.visitVariable(JavaReceiver.java:32)
	at org.openrewrite.java.tree.J$VariableDeclarations$NamedVariable.acceptJava(J.java:6271)
	at org.openrewrite.java.tree.J.accept(J.java:63)
	at org.openrewrite.TreeVisitor.visit(TreeVisitor.java:245)
	at org.openrewrite.j
```
happening when trying to RPC code containing:
```
              export interface Foo {
                ' bar': any;
              }
```


knutwannheden:
We will also want to adjust this: https://github.com/openrewrite/rewrite/blob/4a54c71b01400de2f046282e58f7d00e90c3a8e9/rewrite-javascript/rewrite/src/java/tree.ts#L450

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3726,6 +3726,11 @@ public static boolean isLiteralValue(@Nullable Expression maybeLiteral, @Nullabl
             return false;
         }
 
+        @Override
+        public List<Identifier> getNames() {
+            return Collections.emptyList();
```
knutwannheden: I think it will be better to here return a new `J.Identifier` instance. Let's also add a TODO comment, as we will want to revisit that design at some point.

greg-at-moderne: 4d140a8b3114384f6e5168369f3502276f89f2ba

knutwannheden:
> We will also want to adjust this:
> 
> https://github.com/openrewrite/rewrite/blob/4a54c71b01400de2f046282e58f7d00e90c3a8e9/rewrite-javascript/rewrite/src/java/tree.ts#L450

@greg-at-moderne I still think we want this part.

greg-at-moderne:
> @greg-at-moderne I still think we want this part.

That's already done in 067f895a7f362d59c25c7eb71e16bc71b89844e2.

PR #5570, opened 2025-06-06
Bump plugin version for gradle in buildscript classpath dependency scope

Jenson3210:
## What's changed?
Dependency Trait now allows for `withVersion` to easily bump the version of a Dependency J.MethodInvocation and return the resulting J.MethodInvocation
Plugins can now be updated by the buildscript dependencies also. 

## What's your motivation?
The plugin dependencies in buildscript are methodinvocations. Rather than copying/gathering stuff from elsewhere I think this approach would allow other recipes to bump a version rather easily. (later perhaps also groups/artifactids?

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
@@ -25,6 +25,8 @@ import org.openrewrite.gradle.IsSettingsGradle;
 import org.openrewrite.gradle.internal.ChangeStringLiteral;
 import org.openrewrite.gradle.marker.GradleProject;
 import org.openrewrite.gradle.marker.GradleSettings;
+import org.openrewrite.gradle.trait.GradleDependency;
+import org.openrewrite.gradle.trait.Traits;
 import org.openrewrite.groovy.tree.G;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.StringUtils;
@@ -53,6 +55,7 @@ import static java.util.Collections.singletonList;
 @EqualsAndHashCode(callSuper = false)
 public class UpgradePluginVersion extends ScanningRecipe<UpgradePluginVersion.DependencyVersionState> {
     private static final String GRADLE_PROPERTIES_FILE_NAME = "gradle.properties";
+    public static final String PLUGIN_SUFFIX = ".gradle.plugin";
 
     @EqualsAndHashCode.Exclude
     transient MavenMetadataFailures metadataFailures = new MavenMetadataFailures(this);
@@ -101,6 +104,7 @@ public class UpgradePluginVersion extends ScanningRecipe<UpgradePluginVersion.De
     }
 
     public static class DependencyVersionState {
+        Map<String, String> versionAssignmentToPluginId = new HashMap<>();
         Map<String, String> versionPropNameToPluginId = new HashMap<>();
         Map<String, @Nullable String> pluginIdToNewVersion = new HashMap<>();
     }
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -149,50 +156,52 @@ public TreeVisitor<?, ExecutionContext> getScanner(DependencyVersionState acc) {
             @Override
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (!isPluginVersion(getCursor())) {
-                    return m;
-                }
-                assert m.getSelect() != null;
-                List<Expression> pluginArgs = ((J.MethodInvocation) m.getSelect()).getArguments();
-                if (!(pluginArgs.get(0) instanceof J.Literal)) {
+                Optional<GradleDependency> gradleDependency = Traits.gradleDependency().get(getCursor());
```
shanman190, line 159:
```suggestion
                Optional<GradleDependency> gradleDependency = Traits.gradleDependency().configuration("classpath").get(getCursor());
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -278,6 +315,22 @@ public J visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionC
...
+                pluginId = literalValue(pluginArgs.get(0));
+            }
+        } else if (gradleDependency.isPresent()) {
+            pluginId = gradleDependency.get().getResolvedDependency().getGroupId();
+        }
+        if (pluginId == null || !StringUtils.matchesGlob(pluginId, pluginIdPattern)) {
+            return null;
+        }
+        return pluginId;
+    }
```
shanman190, line 332: So this _only_ tracks when the imported Gradle dependency happens to be the actual plugin marker dependency and thus pulls in the actual dependency as a transitive. 

In a majority of cases where the buildscript block is being utilized, the artifact being imported doesn't share in this quality...

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -46,6 +48,23 @@ public class GradleDependency implements Trait<J.MethodInvocation> {
     @Getter
     ResolvedDependency resolvedDependency;
 
+    @Getter
+    org.openrewrite.gradle.internal.Dependency requestedDependency;
+
+    public J.MethodInvocation withVersion(String version) {
```
shanman190, line 54: We should probably return the trait itself in order to continue updating this making the method invocation update to a new cursor which replaces the current one.

See `YamlValue` for details.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -149,50 +156,52 @@ public TreeVisitor<?, ExecutionContext> getScanner(DependencyVersionState acc) {
...
-                        String resolvedPluginVersion = new DependencyVersionSelector(metadataFailures, gradleProject, gradleSettings)
-                                .select(new GroupArtifact(pluginId, pluginId + ".gradle.plugin"), "classpath", newVersion, versionPattern, ctx);
+                String currentVersion;
+                //buildscript
+                if (gradleDependency.isPresent()) {
+                    GradleDependency dependency = gradleDependency.get();
+                    currentVersion = dependency.getResolvedDependency().getVersion();
+                    if (dependency.getRequestedDependency().getVersion() != null) {
+                        acc.versionAssignmentToPluginId.put(dependency.getRequestedDependency().getVersion(), pluginId);
+                    }
```
shanman190, line 175: The resolved dependency will be out of the `GradleProject` marker, if present, or directly from source if it's not available for some reason. I'm not sure that the requested dependency as a secondary level is needed (read: simplification).

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
shanman190: Just as an extension, it is for this reason why `UpgradeDependencyVersion` handles the dependencies section of both the current project and it's buildscript block rather than trying to make `UpgradePluginVersion` work when it only has the pluginId which may have nearly zero relationship to the classpath dependency declaration.

Jenson3210:
Going to close this one as @shanman190 correctly referred me to the `UpgradeDependencyVersion` for this purpose. Thanks 🙏

PR #5574, opened 2025-06-06
Add GradlePlugin trait

shanman190:
## What's changed?

Added a trait implementing to find direct and transitive Gradle plugins and relevant information about them.

## What's your motivation?

More traits? 😆 

Nothing in particular

Anybody.

N/A

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradlePlugin.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradlePlugin.java
@@ -141,26 +141,20 @@ public class GradlePlugin implements Trait<J> {
 
                         return maybeGradlePlugin(cursor, null, null, null, true);
                     } else if (APPLY_DSL_MATCHER.matches(m, true)) {
-                        if (!(m.getArguments().get(0) instanceof J.Literal)) {
-                            return null;
-                        }
-
-                        if (!(m.getSelect() instanceof J.MethodInvocation && PLUGIN_VERSION_DSL_MATCHER.matches((J.MethodInvocation) m.getSelect(), true))) {
+                        if (!(m.getArguments().get(0) instanceof J.Literal) || !(m.getSelect() instanceof J.MethodInvocation)) {
                             return null;
                         }
 
                         J.MethodInvocation versionSelect = (J.MethodInvocation) m.getSelect();
-                        if (!(versionSelect.getArguments().get(0) instanceof J.Literal)) {
-                            return null;
-                        }
-
-                        if (!(versionSelect.getSelect() instanceof J.MethodInvocation &&
-                                (PLUGIN_ID_DSL_MATCHER.matches((J.MethodInvocation) versionSelect.getSelect(), true) || KOTLIN_PLUGIN_DSL_MATCHER.matches((J.MethodInvocation) versionSelect.getSelect(), true)))) {
+                        if (!PLUGIN_VERSION_DSL_MATCHER.matches(versionSelect, true) ||
+                                !(versionSelect.getArguments().get(0) instanceof J.Literal) ||
+                                !(versionSelect.getSelect() instanceof J.MethodInvocation)) {
                             return null;
                         }
 
                         J.MethodInvocation idSelect = (J.MethodInvocation) versionSelect.getSelect();
-                        if (!(idSelect.getArguments().get(0) instanceof J.Literal)) {
+                        if (!(PLUGIN_ID_DSL_MATCHER.matches(idSelect, true) || KOTLIN_PLUGIN_DSL_MATCHER.matches(idSelect, true)) ||
+                                !(idSelect.getArguments().get(0) instanceof J.Literal)) {
                             return null;
                         }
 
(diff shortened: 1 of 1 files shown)


shanman190:
It would be most ideal that the FieldAccess and Identifier cases were able to specify the full class name via type attribution. Unfortunately, the types weren't available even after adding the OpenRewrite plugin to the classpath.

I didn't feel comfortable either with forcing the class name in from either value due to the possibility of all other kinds of code combinations that could arise from real world usage.

TODOs: 
- [x] Add `alias` support from Gradle Version Catalogs. Like FieldAccess and Identifier, this probably won't carry much useful information at first due to the lack thereof in the model and markers.
- [x] Add test case for `settings.gradle(.kts)`

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradlePluginTest.java
```diff
@@ -0,0 +1,775 @@
...
+                  buildscript {
+                      repositories {
+                          mavenCentral()
+                      }
+                      dependencies {
+                          classpath("org.openrewrite.rewrite:org.openrewrite.rewrite.gradle.plugin:8.0.0")
+                      }
+                  }
+                  /*~~>*/apply(plugin="org.openrewrite.rewrite")
+                  """
```
Jenson3210: Is it expected that we have null version here after changing to printing applied/version?
```suggestion
              spec -> spec.recipe(RewriteTest.toRecipe(() -> plugin().asVisitor(plugin -> SearchResult.found(plugin.getTree(), plugin.getPluginId() + ":" + plugin.getVersion() + ":" + plugin.isApplied())))),
              buildGradleKts(
                """
                  buildscript {
                      repositories {
                          mavenCentral()
                      }
                      dependencies {
                          classpath("org.openrewrite.rewrite:org.openrewrite.rewrite.gradle.plugin:8.0.0")
                      }
                  }
                  apply(plugin="org.openrewrite.rewrite")
                  """,
                """
                  buildscript {
                      repositories {
                          mavenCentral()
                      }
                      dependencies {
                          classpath("org.openrewrite.rewrite:org.openrewrite.rewrite.gradle.plugin:8.0.0")
                      }
                  }
                  /*~~(org.openrewrite.rewrite:null:true)~~>*/apply(plugin="org.openrewrite.rewrite")
                  """
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradlePluginTest.java
```diff
@@ -0,0 +1,775 @@
...
+                """
+                  buildscript {
+                      repositories {
+                          mavenCentral()
+                      }
+                      dependencies {
+                          classpath "org.openrewrite.rewrite:org.openrewrite.rewrite.gradle.plugin:8.0.0"
+                      }
+                  }
+                  /*~~>*/apply plugin: "org.openrewrite.rewrite"
```
Jenson3210, line 153: Same question as the other one

Jenson3210:
I think this is awesome! 
If possible, we could enhance it with the versions if the version itself is declared higher up / somewhere else in the cursor. 
But if not possible in an accurate way, we can also leave it.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradlePluginTest.java
shanman190: So at the invocation site and even in the marker for this case there would be no version. This relates back to not having a relationship for the plugin back to its GAV (though in these tests, I am using the marker artifact). We could potentially "try" to discover the version as best as we can, but there are many situations where this would fail.

An idea that I have looking forward is during LST generation scanning the Gradle build's classpath for marker IDs which we can then use for two purposes:

1. Ensuring every plugin application has both its id and implementation class.
2. We should be able to associate a GAV to a plugin because we will know where it came from.

The second quality is especially useful for another best practice recipe that I'm wanting which is to convert from `buildscript` to `plugins`.

Jenson3210: As it is not possible to know it for sure, not much we can do.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/Traits.java
```diff
@@ -26,4 +26,8 @@ public static GradleDependency.Matcher gradleDependency() {
     public static JvmTestSuite.Matcher jvmTestSuite() {
         return new JvmTestSuite.Matcher();
     }
+
+    public static GradlePlugin.Matcher plugin() {
+        return new GradlePlugin.Matcher();
+    }
```
timtebeek: As a quick FYI we're moving away from `Trait` classes as they can lead to import conflicts:
- https://github.com/openrewrite/rewrite/pull/5613
- https://github.com/openrewrite/rewrite-rewrite/pull/16

The suggested replacement is just an inlined `new GradlePlugin.Matcher()`. Likely best to then not add this here & adjust the tests.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -73,6 +77,21 @@ public Matcher artifactId(@Nullable String artifactId) {
...
+                public J visitMethodInvocation(J.MethodInvocation method, P p) {
+                    GradleDependency dependency = test(getCursor());
+                    return dependency != null ?
+                            (J) visitor.visit(dependency, p) :
+                            super.visitMethodInvocation(method, p);
+                }
+            };
+        }
+
+
```
timtebeek:
```suggestion
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -73,6 +77,21 @@ public Matcher artifactId(@Nullable String artifactId) {
...
+
         @Override
```
timtebeek:
```suggestion
        }

        @Override
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/Traits.java
shanman190: No problem! I'll get this cleaned up tomorrow as a part of that discussions outcome (I was a small part of that discussion as well. 😉).

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradlePluginTest.java
```diff
@@ -0,0 +1,794 @@
...
+                  """
+              )
+            );
+        }
+
+        @Test
+        void alias() {
+            rewriteRun(
+              spec -> spec.recipe(RewriteTest.toRecipe(() -> new GradlePlugin.Matcher().asVisitor(plugin -> SearchResult.found(plugin.getTree())))),
+              toml(
```
knutwannheden, line 751: Wondering if we should have a gradleVersionCatalog() or similar.

shanman190: Definitely a good candidate for a potential helper method. The default file name that is picked up automatically by Gradle is `gradle/libs.versions.toml`, so we'd probably align in some capacity to that if we chose to adopt such a helper.

PR #5575, opened 2025-06-08
Add support for Kotlin 2.x

barbulescu:
### Checklist


timtebeek:
[Gradle 9](https://github.com/gradle/gradle/releases/tag/v9.0.0-RC1) should be interesting, as that now requires Kotlin 2.1. All the more appreciative of your efforts here!

timtebeek:
Linking the relevant [thread in the Kotlin lang Slack](https://slack-chats.kotlinlang.org/t/28493795/what-is-the-recommended-way-to-migrate-firsessionfactoryhelp#997dab2b-2e41-4549-9721-ae0192a3f337) here as well. Let us know if you could use more help!

e5l:
Hey @barbulescu, we're looking forward to having Kotlin 2.x support 🙂.

Could you please let me know if you have any issues or require assistance?

timtebeek:
Thanks for your interest @e5l ; there's also this parallel alternative approach, where we're exploring a separate Kotlin 2 parser which more closely matches the model of the underlying compiler:
- https://github.com/openrewrite/rewrite/pull/5963

It's expected that we can take learnings from the work done here, but might have that flow into the new Kotlin 2 parser, such that we can support these two in parallel rather than a hard switch over.

jkschneider:
Superseded by #6766. Thanks for all your work on this @barbulescu. I pulled forward a lot of your work and added you as a co-author there to respect your contribution to it!

PR #5576, opened 2025-06-08
`AddOrUpdateAnnotationAttribute` mistakenly matches named attribute with field reference

timtebeek:
As reported.


jevanlingen (approved):

Laurens-W:
~I wonder whether this actually addresses the issue reported~
After taking a closer look I think it does

Laurens-W (approved):

PR #5581, opened 2025-06-09
Correctly parse comments in doctype declarations

### LINKED ISSUE #4930: XML parser parses comments inside doctype subset into `Xml.Ident`
Given an XML document like the following:

```xml
  <?xml version="1.0" encoding="iso-8859-1" standalone="no"?>
  <!DOCTYPE xsl:stylesheet [
  <!-- EXSLT-Math -->
  <!ENTITY foons "https://www.foo.org/bar">
  ]>
  <xsl:stylesheet version="1.0"/>
```

the parser parses the comment into an `Xml.Ident` element rather than creating an `Xml.Comment`: https://github.com/openrewrite/rewrite/blob/b6f178b30a4298eb5f06015783638e17c77a75a9/rewrite-xml/src/main/java/org/openrewrite/xml/internal/XmlParserVisitor.java#L334-L338.

The name of the created `Xml.Ident` is `<!-- EXSLT-Math -->` which then can cause recipes to fail processing this.

jhl221123:
## What's changed?
This change updates the logic in `visitDoctypedecl` to correctly create `Xml.Comment` objects for comments found within a doctype's internal subset. All other markup, such as entity declarations, is parsed as `Xml.CharData` to preserve its raw text.

## What's your motivation?
Closes #4930

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
@@ -740,7 +740,7 @@ public interface Xml extends Tree {
             }
 
             Markers markers;
-            List<Element> elements;
+            List<Content> elements;
 
             @Override
             public <P> Xml acceptXml(XmlVisitor<P> v, P p) {
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/internal/XmlParserVisitor.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/internal/XmlParserVisitor.java
@@ -31,7 +31,6 @@ import org.openrewrite.xml.tree.Xml;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.function.BiFunction;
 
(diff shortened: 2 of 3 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
```diff
@@ -740,7 +740,7 @@ public String getPrefix() {
             }
 
             Markers markers;
-            List<Element> elements;
+            List<Content> elements;
```
timtebeek, line 743: I have a slight worry here about deserialization, but I'm not on my laptop to check.. Pointing it out here for a more thorough review.

timtebeek:
Thanks a lot for the help here @jhl221123 ! I've tagged Greg for an additional review of the concern I posted above.

rewrite-xml/src/main/java/org/openrewrite/xml/tree/Xml.java
greg-at-moderne: - Indeed, types `Element` and `Content` seem unrelated (other than descending from `Xml`).
- But I agree, the external subsets shouldn't be modelled as a collection of elements. These are not elements from logical perspective.
- Having said that, I think we should go for it. As far as I understand,tThe failure scenario is that for some XML files which use these declarations, their LSTs will no longer match until re-built.

All in all - let's try.

greg-at-moderne (approved):

greg-at-moderne:
Thank you so much for the contribution, @jhl221123.
I appreciate the thorough test case provided.

knutwannheden:
Something seems wrong here. Note that Xml.Element isn't a normal XML element. For that we have Xml.Tag. Perhaps it is meant to correspond to https://www.w3.org/TR/xml/#NT-elementdecl?

AFAICT, we now no longer create any Xml.Element instances anywhere. That would mean that any recipes which operate on these or on owned objects would be broken now. Can we check this again?

PR #5583, opened 2025-06-09
Fail to resolve property values if they can't be resolved

kmccarp:
## What's changed?

Prevent `StackOverflowError` when a property cannot be resolved.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -1032,7 +1032,7 @@ public class ResolvedPom {
                     if ((d.getGav().getGroupId() != null && d.getGav().getGroupId().startsWith("${") && d.getGav().getGroupId().endsWith("}")) ||
                         (d.getGav().getArtifactId().startsWith("${") && d.getGav().getArtifactId().endsWith("}")) ||
                         (d.getGav().getVersion() != null && d.getGav().getVersion().startsWith("${") && d.getGav().getVersion().endsWith("}"))) {
-                        throw new MavenDownloadingException("Could not resolve property", null, d.getGav());
+                        throw new MavenDownloadingException("Could not determine values for all properties in " + d.getGav(), null, d.getGav());
                     }
 
                     Pom dPom = downloader.download(d.getGav(), null, dd.definedIn, getRepositories());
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/tree/ResolvedPomTest.java
```diff
@@ -107,6 +109,56 @@ void resolveDependencyWithPlaceholderClassifier() {
...
+                        </dependency>
+                    </dependencies>
+                </project>
+                """,
+              spec -> spec
+                .afterRecipe(doc -> {
+                    List<ResolvedDependency> deps = doc.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow().getDependencies()
+                      .get(Scope.Compile);
+                    assertThat(deps).hasSize(1);
+                    assertThat(deps.getFirst().getVersion()).isEqualTo("3.17.0");
```
steve-aom-elliott: Can it actually theoretically resolve this version number given it's not defined at all in the files provided? In the test below this, `projectVersion`, it at least has a place to eventually pull that from.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -1029,10 +1029,14 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
...
-                        throw new MavenDownloadingException("Could not resolve property", null, d.getGav());
+                    if (d.getGav().getGroupId() != null && d.getGav().getGroupId().startsWith("${") && d.getGav().getGroupId().endsWith("}")) {
+                        throw new MavenDownloadingException("Could not resolve property " + d.getGav().getGroupId() + " for pom " + d.getGav(), null, d.getGav());
+                    }
+                    if (d.getGav().getArtifactId().startsWith("${") && d.getGav().getArtifactId().endsWith("}")) {
+                        throw new MavenDownloadingException("Could not resolve property " + d.getGav().getArtifactId() + " for pom " + d.getGav(), null, d.getGav());
+                    }
+                    if (d.getGav().getVersion() != null && d.getGav().getVersion().startsWith("${") && d.getGav().getVersion().endsWith("}")) {
+                        throw new MavenDownloadingException("Could not resolve property " + d.getGav().getVersion() + " for pom " + d.getGav(), null, d.getGav());
```
sambsnyd: I don't think this adds anything... the whole gav, unresolved property placeholder included, is in the message already

rewrite-maven/src/test/java/org/openrewrite/maven/tree/ResolvedPomTest.java
kmccarp: Nope, that was a copy/paste error, good catch. Removed.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
kmccarp: Solid point. Updated the logic and the message.

kmccarp:
This PR caused more problems than it fixed, so closing.

PR #5586, opened 2025-06-10
Maven self-referential property resolution

jkschneider:
## What's changed?

We're trying to address a StackOverflowError in Maven resolution for dependencies that use ${project.version} as their version.

The StackOverflowError occurred when a POM defines a property that shadows the built-in project.version with a
  self-referential value:

```
  <properties>
    <project.version>${project.version}</project.version>
  </properties>
```

  This is a valid Maven scenario (though unusual) where someone inadvertently creates a circular reference by
  defining a user property with the same name as a built-in property.

  When resolving `${project.version}`:
  1. ResolvedPom.getProperty() checks user-defined properties first
  2. It finds project.version=${project.version}
  3. Returns "${project.version}" which needs resolution
  4. This creates infinite recursion

  When a property references itself, we skip the user-defined value and fall through to the built-in property
  resolution, which correctly returns the actual project version.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -2,7 +2,8 @@
   "permissions": {
     "allow": [
       "Bash(./gradlew:*)",
-      "Bash(find:*)"
+      "Bash(find:*)",
+      "Bash(grep:*)"
     ],
     "deny": []
   }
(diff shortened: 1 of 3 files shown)


kmccarp:
[This unit test](https://github.com/openrewrite/rewrite/blob/b269851bdce370df8d1cc6a8ce131e5486777114/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java#L3368-L3414) and the logic that backs it should likely change too?

PR #5588, opened 2025-06-10
Improve annotation support for records on constructors

jevanlingen:
## What's changed?
Records with a `@Target(value=PARAMETER)` annotation are supported as well.

## What's your motivation?
- https://github.com/openrewrite/rewrite-spring/pull/713

## Anyone you would like to review specifically? Have you considered any alternatives or workarounds?
This code is a workaround at best 😢. The problem is that the AST does not contain all the proper information where you would expect it to be. So I grabbed the information from another element and merged it with the stateVector. It feels dirty and wrong to do this (as you can see in the PR, this has consequences; to make it work I also needed to "upgrade" the visitLiteral code, because the `endPosTable` does not contain the information about these annotations either). I don't see another way around this, so I'll say we move forward. But if there is better way, please let me know!

### Checklist


jkschneider:
Converting to draft because this needs serious review.

Laurens-W:
Under water a record still becomes a class, why don't we map it as such?
-> after mapping the constructor and moving the cursor we should go over the `stateVector` list and map the `VariableDecls` to `members` (without moving the cursor). That way we have all of the context we need in the LST

We do then have to do major refactoring to how we print back record classes

Laurens-W:
Or better yet, we should construct the constructor without moving the cursor (as it is derived from the variable declarations) and parse the record parameters into `VariableDeclarations` that are members, while moving the cursor

jevanlingen:
> Under water a record still becomes a class, why don't we map it as such?

We do map it as a class, but just that it has a `primaryConstructor`. See also the original [support-record-commit](https://github.com/openrewrite/rewrite/commit/5e320edc65f14b40b0c9c2edd33d76f40ac20b60). I don't know if we really want to change the idea of a primaryConstructor. It's interesting though that for example for Kotlin, the primaryConstructor is not used (but instead there is a PrimaryConstructor marker).

jevanlingen:
This is possibly not the direction we want to take, but can serve as inspiration. Closed it for now.

jevanlingen:
We could go for a delegate JCVariableDecl class instead of using reflection to alter the original AST element. That would be somewhat better I think.

PR #5591, opened 2025-06-10
Revert "Correctly parse comments in doctype declarations"

greg-at-moderne:
Reverts openrewrite/rewrite#5581
as it's causing downstream problems, e.g.
```
/home/runner/work/rewrite-java-security/rewrite-java-security/src/main/java/org/openrewrite/java/security/xml/ExternalDTDAccumulator.java:32: error: incompatible types: Content cannot be converted to Element
                    for (Xml.Element element : docTypeDecl.getExternalSubsets().getElements()) {
```
Also this comment:
https://github.com/openrewrite/rewrite/pull/5581#issuecomment-2957716271
needs to be addressed.

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/internal/XmlParserVisitor.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/internal/XmlParserVisitor.java
@@ -31,6 +31,7 @@ import org.openrewrite.xml.tree.Xml;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


MBoegers:
We should defnetly have a second look if we could find a way to do both ;)

greg-at-moderne:
FYI @jhl221123

jhl221123:
@greg-at-moderne, Thanks for the heads-up and for handling the revert so quickly. My apologies that my change caused this downstream issue; Please let me know if there's anything I can do to assist with the new approach.

Issue #5545, opened 2025-06-02
Kotlin Recipe Declarative String Arguments Instantiate as Empty String


ryan-hudson:
As of OpenRewrite 8.50.0, String options passed declaratively into Kotlin recipes are being instantiated as "". 

Debugging a bit, it looks like this issue was caused by [this](https://github.com/openrewrite/rewrite/commit/ffca900d5866aecff0c00e09c76413b7301bbf4d#diff-4f7f19d4eafba083e6cda1dc9b9b8689b8c664e3afe65374e59897f7ac2f152e) change to the Recipe Loader, which is now delegating recipe instantiation to `RecipeIntrospectionUtils.constructRecipe`, which sets Kotlin String arguments to "" [here](https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/internal/RecipeIntrospectionUtils.java#L123-L127). That logic appears to have been added purely for validation, and was not intended to instantiate recipe args. 
- See https://github.com/openrewrite/rewrite/issues/3508

## What version of OpenRewrite are you using?
8.54.0

## How are you running OpenRewrite?
Unit Tests in private repo. 

### Declarative Yaml
```yaml
type: specs.openrewrite.org/v1beta/recipe
name: net.sample.kotlin.KotlinReplaceAllRecipeWrapper
displayName: Kotlin Replace All Wrapper
description: Tests a simple Kotlin recipe that replaces all text with the provided option.
recipeList:
  - net.sample.kotlin.KotlinReplaceAllRecipe:
      replacementText: "REPLACEMENT_TEXT"
```

### Imperative Kotlin Recipe
```kotlin
package net.sample.kotlin

import com.fasterxml.jackson.annotation.JsonCreator
import com.fasterxml.jackson.annotation.JsonProperty
import org.openrewrite.*
import org.openrewrite.text.PlainText
import org.openrewrite.text.PlainTextVisitor


class KotlinReplaceAllRecipe @JsonCreator constructor(
  @Option(
    displayName = "Replacement Text",
    description = "Resulting text after the replacement occurs."
  ) @JsonProperty("replacementText") val replacementText: String
): Recipe() {
  override fun getDisplayName() = "Tests a simple Kotlin Recipe With Options"
  override fun getDescription() = "A test case for recipe authoring in Kotlin."

  override fun getVisitor(): TreeVisitor<*, ExecutionContext> =
    object: PlainTextVisitor<ExecutionContext>() {
[...]

PR #5595, opened 2025-06-11
Always instantiate Kotlin recipes via Jackson

knutwannheden:
The `RecipeIntrospectionUtils#constructRecipe()` is a lightweight alternative to Jackson, but shouldn't try to handle Kotlin recipes, as typically, the Kotlin compiler is not configured to emit parameter names into the Java bytecode, as that is already part of the Kotlin metadata (in the `kotlin.Metadata` annotation). Also, the metadata can contain default values, which would need handling, so it makes more sense to let Jackson handle Kotlin recipes using its dedicated Kotlin module.

Additionally, the `RecipeLoader` will add the `KotlinModule` to the `ObjectMapper` if it can be found on the classpath, so that Kotlin recipes can be instantiated properly even if they aren't annotated using Jackson annotations (like `@JsonCreator`).

Fixes:
 - #5545


rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -186,7 +191,7 @@ private void configureRecipes(ScanResult result, String className) {
             Timer.Builder builder = Timer.builder("rewrite.scan.configure.recipe");
             Timer.Sample sample = Timer.start();
             try {
-                Recipe recipe = constructRecipe(recipeClass);
+                Recipe recipe = recipeLoader.load(recipeClass, emptyMap());
```
knutwannheden, line 194: This should be safe as `RecipeLoader#load()` internally uses `constructRecipe()` but at the same time tries to delegate to Jackson when that fails.

rewrite-core/src/main/java/org/openrewrite/internal/RecipeIntrospectionUtils.java
```diff
@@ -113,18 +113,13 @@ public static Recipe constructRecipe(Class<?> recipeClass, Map<String, Object> a
...
             java.lang.reflect.Parameter param = constructor.getParameters()[i];
             if (args != null && args.containsKey(param.getName())) {
                 constructorArgs[i] = convert(args.get(param.getName()), param.getType());
             } else if (param.getType().isPrimitive()) {
                 constructorArgs[i] = getPrimitiveDefault(param.getType());
-            } else if (param.getType().equals(String.class) && isKotlin(clazz)) {
-                // Default Recipe::validate is more valuable if we pass null for unconfigured Strings.
-                // But, that's not safe for Kotlin non-null types, so use an empty String for those
-                // (though it will sneak through default recipe validation)
-                constructorArgs[i] = "";
```
knutwannheden, line 127: Glad we can remove this block, as it seemed kind of arbitrary.

nmck257: Agree, I am not proud of this code 😉

rewrite-kotlin/build.gradle.kts
```diff
@@ -39,3 +40,9 @@ kotlin {
         jvmTarget = JvmTarget.JVM_1_8
     }
 }
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    if (name == "compileTestKotlin") {
+        kotlinOptions.jvmTarget = "21"
```
jkschneider: This seems like the wrong target which still should be 8

knutwannheden: I think I had to do this because it claimed we use Java 21 for the Java tests. But I will check if I can change this.

knutwannheden: Possibly due to that rewrite-java-21 runtimeTestOnly dependency? 🤔

nmck257: if it's relevant, I have this snippet in my spring-wrapper recipe repo:
```kts
configurations["testWhatever"]
        .attributes {
            // allow the test cases to pull deps which use a newer java version than the recipes themselves
            attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 17)
        }
```

PR #5596, opened 2025-06-11
Handle `RecipeRunException` in `generate` phase

knutwannheden:
When a recipe throws a `RecipeRunException` (typically via visitor) in the `generate` phase, that should not cause the recipe scheduler to fail.


pstreef (approved):

timtebeek:
Consensus on Slack was that this was already low risk, but still tagging for additional review as was indicated on the associated thread on Wednesday.

PR #5603, opened 2025-06-11
add reproducer showing `gradle.FindDependency` finds gradle dependency constraints

MBoegers:
### Checklist


shanman190:
The issue here is with the custom visitor defined for `FindGradleDependency` which doesn't account for the constraints blocks.

I think the easiest fix is to refactor `FindGradleDependency` to use the much more robust `GradleDependency` trait.

PR #5605, opened 2025-06-11
Use `@Nullable` fields instead of ``@JsonCreator`

pstreef:
The decision was made to avoid `@JsonCreator` throughout the rewrite code. By making the fields nullable and adding checks we can avoid failure for old LSTs just like with the `@JconCreator`

This replaces: https://github.com/openrewrite/rewrite/pull/5597

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/WrappingAndBracesVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/WrappingAndBracesVisitor.java
@@ -207,8 +207,8 @@ public class WrappingAndBracesVisitor<P> extends JavaIsoVisitor<P> {
         });
     }
 
-    private Space wrapElement(Space prefix, String whitespace, WrappingAndBracesStyle.Annotations annotationsStyle) {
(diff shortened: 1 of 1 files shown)


pstreef:
> Can we add a short description? I take it the idea is to not format anything unless we have some explicit styles? That was the other option we considered.

Yes, I just wanted to make the draft PR to share the in progress state. Tests are now failing which I still need to figure out.

the reason for the change was:

> jonathan: i’d like to change the approach. we’ve tried to avoid propagating @JsonCreators throughout the core code as much as possible

jkschneider:
> I take it the idea is to not format anything unless we have some explicit styles? 

Yes, this is part of the benefit of not doing the `@JsonCreator` approach as well. An old LST will be missing the styles, and will have missed the opportunity to autodetect them at LST creation time, so we wouldn't want autoformat to potentially change all the formatting just because we backfilled with a default. It's better to just avoid that formatting until another ingestion cycle passes and we're able to benefit from the autodetected style.

jkschneider (approved):

PR #5606, opened 2025-06-11
Enhance MavenSettings to handle password decryption

Jammy-Louie:
## What's changed?
fixes https://github.com/moderneinc/customer-requests/issues/1155

We decrypt the passwords if we are loading the Maven files from disk but we don't decrypt the settings if loaded from parsing.

## What's your motivation?

- consistent behaviour when loading the Maven `settings.xml`
- fix customer issue where they are using a `security-setting.xml` to encrypt their passwords

approach?

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSecuritySettings.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSecuritySettings.java
@@ -201,7 +201,8 @@ public class MavenSecuritySettings {
             System.arraycopy(clearBytes, 0, decryptedBytes, 0, decryptedBytes.length);
             return new String(decryptedBytes, StandardCharsets.UTF_8);
         } catch (NoSuchPaddingException | NoSuchAlgorithmException | BadPaddingException | IllegalBlockSizeException |
-                 InvalidKeyException | InvalidAlgorithmParameterException | IllegalArgumentException e) {
+                 InvalidKeyException | InvalidAlgorithmParameterException | NegativeArraySizeException |
+                 ArrayIndexOutOfBoundsException | IllegalArgumentException e) {
             return null;
         }
     }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
@@ -85,8 +85,10 @@ public class MavenSettings {
 
     public static @Nullable MavenSettings parse(Parser.Input source, ExecutionContext ctx) {
         try {
-            return new Interpolator().interpolate(
+            MavenSettings settings = new Interpolator().interpolate(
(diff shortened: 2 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/MavenSecuritySettings.java
```diff
@@ -201,7 +201,8 @@ String decrypt(@Nullable String fieldValue, @Nullable String password) {
             System.arraycopy(clearBytes, 0, decryptedBytes, 0, decryptedBytes.length);
             return new String(decryptedBytes, StandardCharsets.UTF_8);
         } catch (NoSuchPaddingException | NoSuchAlgorithmException | BadPaddingException | IllegalBlockSizeException |
-                 InvalidKeyException | InvalidAlgorithmParameterException | IllegalArgumentException e) {
+                 InvalidKeyException | InvalidAlgorithmParameterException | NegativeArraySizeException |
+                 ArrayIndexOutOfBoundsException | IllegalArgumentException e) {
```
Jammy-Louie, line 205: added `NegativeArraySizeException` and `ArrayIndexOutOfBoundsException` to handle situations where we are attempting to decrypt plain text passwords.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -85,23 +85,31 @@ public MavenSettings(@Nullable String localRepository, @Nullable Profiles profil
 
     public static @Nullable MavenSettings parse(Parser.Input source, ExecutionContext ctx) {
         try {
-            return new Interpolator().interpolate(
+            MavenSettings settings = new Interpolator().interpolate(
                     MavenXmlMapper.readMapper().readValue(source.getSource(ctx), MavenSettings.class));
+            settings.maybeDecryptPasswords(ctx);
```
Laurens-W, line 90: Should we wrap this in a null check too?

Jammy-Louie: I don't think it is necessary, `Interpolator().interpolate` should return a non nullable `MavenSettings` object, 

code for reference, https://github.com/openrewrite/rewrite/blob/2a89e1de8c5a54524dcd12597496fcfb749a8a4b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java#L237-L244

Jammy-Louie:
there were concerns about storing plain text passwords but that is already being done outside of this change.

see code for reference:

https://github.com/openrewrite/rewrite-maven-plugin/blob/4eb743cf62315ada7c71ba903c350f48a88814ee/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L737-L754

sambsnyd:
> there were concerns about storing plain text passwords but that is already being done outside of this change.

The actual reason this isn't a problem is that we explicitly mark credentials as not for serialization so that they don't ever get persisted as part of the LST.

PR #5607, opened 2025-06-11
Datatable should not have the direct column

Jenson3210:
## What's changed?
Datatable should not have the direct column

**AFAICT, I will also need to followup release in rewrite-java-dependencies as that one also uses the constructor [here](https://github.com/openrewrite/rewrite-java-dependencies/blob/f6d1cff5be4c865fa3ef4b09ff612394655d482b/src/main/java/org/openrewrite/java/dependencies/search/FindMinimumDependencyVersion.java#L176)**

### Checklist


Jenson3210:
@jkschneider as discussed removing the column and updating its consumer.

PR #5608, opened 2025-06-11
Widening the type of ForEachLoop.Control.variable to Statement

greg-at-moderne:
## What's changed?

In Java:
- changing the type of `ForEachLoop.Control.variable` from `VariableDeclarations` to `Statement`.

In Javascript:
- using this to handle cases where `for-of` or `for-in` loops use a variable declared before the loop.

## What's your motivation?

As designed with @jkschneider and @knutwannheden this is to accommodate a broader set of JS/TS for loops.

Note - this slightly changes the API for `J.ForEachLoop`. As checked before, this will cause some minor compilation issues in a few (roughly around 5) places downstream in `openrewrite` org.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/parser.ts
@@ -2723,7 +2723,7 @@ export class JavaScriptParserVisitor {
                             } as JS.ArrayBindingPattern, this.suffix(node.initializer))
                         } else if (ts.isObjectLiteralExpression(node.initializer)) {
                             return this.rightPadded({
-                                kind: JS.Kind.ObjectBindingDeclarations,
+                                kind: JS.Kind.ObjectBindingPattern,
                                 id: randomId(),
                                 leadingAnnotations: [],
                                 modifiers: [],
@@ -2738,7 +2738,7 @@ export class JavaScriptParserVisitor {
                                 } as J.Container<Expression>,
                                 markers: emptyMarkers,
                                 prefix: emptySpace
-                            } as JS.ObjectBindingDeclarations, this.suffix(node.initializer))
+                            } as JS.ObjectBindingPattern, this.suffix(node.initializer))
                         }
                         else {
                             return this.rightPadded(this.visit(node.initializer), this.suffix(node.initializer))
--- a/rewrite-javascript/rewrite/test/javascript/parser/for.test.ts
+++ b/rewrite-javascript/rewrite/test/javascript/parser/for.test.ts
@@ -318,8 +318,8 @@ describe('for mapping', () => {
             `),
             afterRecipe: (cu: JS.CompilationUnit) => {
                 const forOfLoop = <JS.ForOfLoop>cu.statements[2].element;
-                expect(forOfLoop.loop.control.variable.element.kind).toBe(JS.Kind.ObjectBindingDeclarations);
-                const objectBinding = <JS.ObjectBindingDeclarations>forOfLoop.loop.control.variable.element;
+                expect(forOfLoop.loop.control.variable.element.kind).toBe(JS.Kind.ObjectBindingPattern);
+                const objectBinding = <JS.ObjectBindingPattern>forOfLoop.loop.control.variable.element;
                 expect(objectBinding.bindings.elements.length).toBe(3);
                 for (let i = 0; i < 3; i++) {
                     expect(objectBinding.bindings.elements[i].element.kind).toBe(JS.Kind.PropertyAssignment);


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1537,9 +1537,8 @@ public void visitForLoop(ForStatement forLoop) {
                         forEachMarkers = forEachMarkers.add(new InStyleForEachLoop(randomId()));
                     }
 
-                    JRightPadded<J.VariableDeclarations> variable = JRightPadded.build(new J.VariableDeclarations(randomId(), paramFmt,
-                            Markers.EMPTY, emptyList(), modifiers, paramType, null, emptyList(),
-                            singletonList(paramName))
+                    JRightPadded<Statement> variable = JRightPadded.build(
+                            (Statement) new J.VariableDeclarations(randomId(), paramFmt, Markers.EMPTY, emptyList(), modifiers, paramType, null, emptyList(), singletonList(paramName))
```
knutwannheden: I suggest we change it to this, so that if anything changes we rather get a compiler warning than a runtime exception.
```suggestion
                    JRightPadded<Statement> variable = JRightPadded.<Statement>build(
                            new J.VariableDeclarations(randomId(), paramFmt, Markers.EMPTY, emptyList(), modifiers, paramType, null, emptyList(), singletonList(paramName))
```

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -2294,13 +2294,16 @@ public static final class Control implements J {
             @Getter
             Markers markers;
 
-            JRightPadded<VariableDeclarations> variable;
+            // If used to be VariableDeclarations, but got widened to Statement for JS/TS sake as one can
+            // have other statements there. For instance, use a variable defined before the loop.
+            // Keeping the "variable" name as this is the most prominent usage anyway.
```
knutwannheden: FYI: Keeping the `variable` name is important for another reason: Jackson would not be able to deserialize the LST models if we would just change that without any other changes (like a `@JsonProperty` annotation or an `@JsonCreator` annnotated constructor to do the mapping).

rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -2685,7 +2698,37 @@ export class JavaScriptParserVisitor {
...
+                        if (node.initializer.kind == ts.SyntaxKind.Identifier) {
+                            const ident = this.visit(node.initializer);
+                            return this.rightPadded({
+                                kind: JS.Kind.ExpressionStatement,
+                                id: randomId(),
+                                expression: ident,
+                                markers: emptyMarkers,
+                                prefix: emptySpace
+                            } as JS.ExpressionStatement, this.suffix(node.initializer));
+                        } else if (node.initializer.kind == ts.SyntaxKind.ArrayLiteralExpression) {
```
knutwannheden: Can we also check that we properly support this here:

```ts
const users = [
    {name: 'Alice', age: 30, city: 'NYC'},
    {name: 'Bob', age: 25, city: 'LA'},
    {name: 'Carol', age: 35, city: 'Chicago'}
];

let n, a, c;
for ({name: n, age: a, city: c} of users) {
    console.log(`${n} is ${a} years old and lives in ${c}`);
}
```

and also this (I expect this will work already):
```ts
let first, rest;
for ([first, ...rest] of [[1,2],[3,4]]) {
    console.log(first, rest);
}
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
greg-at-moderne: 0e45d2188

rewrite-javascript/rewrite/src/javascript/parser.ts
greg-at-moderne: Regarding the first one. It's mapped as `J.NewClass` as per 72518a8b8.
Typescript maps that as `ObjectLiteralExpression`.
I don't have a better idea how to model it.

greg-at-moderne: The other one is mapped as array deconstruction: 8bd38908f

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
greg-at-moderne: Right.
21be47e9d

rewrite-javascript/rewrite/src/javascript/parser.ts
knutwannheden: I think before we also had the array deconstruction case mapped as `J.NewArray`. So if we changed that, it would seem to also make sense to change `J.NewClass` to `JS.ObjectBindingDeclarations`. What do you think?

greg-at-moderne: Ah. I missed this one. a8fd4c524

rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -2687,7 +2700,55 @@ export class JavaScriptParserVisitor {
...
+                                    id: randomId(),
+                                    before: emptySpace,
+                                    elements: arrayLiteral.elements.map(e => this.rightPadded(this.visit(e) as Expression, this.suffix(e))),
+                                    markers: emptyMarkers
+                                } as J.Container<Expression>,
+                                markers: emptyMarkers,
+                                prefix: emptySpace
+                            } as JS.ArrayBindingPattern, this.suffix(node.initializer))
+                        } else if (node.initializer.kind == ts.SyntaxKind.ObjectLiteralExpression) {
+                            const objectLiteral = node.initializer as ts.ObjectLiteralExpression;
```
knutwannheden: I think the TS Compiler API has type guards for most types. So you should be able to call `isObjectLiteralExpression(node.initializer)` in the `if` condition.

greg-at-moderne: 5f569575f

PR #5609, opened 2025-06-11
Include tgz-type dependencies in Maven resolution

kmccarp:
## What's changed?

Ensure that we can pull in dependencies of type `tgz` for users of Maven resolution.


jkschneider:
What consequence does this have if used in combination with `MavenArtifactDownloader`?

jkschneider:
image

PR #5610, opened 2025-06-12
Polish the Gradle build's configuration phase

shanman190:
## What's changed?
Gradle build cleanup.

Before: 1m 58s
After: 18s

## What's your motivation?
Get to actually having Gradle do work sooner

rewrite-javascript had a bit of restructuring, but is fundamentally the same. It should be reviewed carefully though.

Several of the previous rewrite-javascript tasks were using the node-gradle plugin's internal Gradle rule (ie. `npm_*` which resulted in eager task creation).

Anybody

N/A

## Any additional context
There are additional pieces that can be cleaned up. There are several plugins -- the license plugin being the worst offender -- that still eagerly create tasks and we could investigate next enabling configuration cache as well.

### Checklist

### DIFF
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -57,7 +57,7 @@ extensions.configure<NodeExtension> {
 }
 
 val datedSnapshotVersion by extra {
-    if (System.getenv("CI")?.toBoolean() ?: false) {
+    if (System.getenv("CI") != null) {
         project.version.toString().replace(
             "SNAPSHOT",
             LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss"))
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -199,7 +199,7 @@ val npmPublish = tasks.register<NpmTask>("npmPublish") {
         .withPathSensitivity(PathSensitivity.RELATIVE)
     dependsOn(setupNpmrc)
 
-    args = listOf("publish", "--dry-run")
+    args = listOf("publish", npmPack.get().archiveFile.get().asFile.absolutePath, "--dry-run")
     if (!project.hasProperty("releasing")) {
         args.addAll("--tag", "next")
     }
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -199,7 +199,7 @@ val npmPublish = tasks.register<NpmTask>("npmPublish") {
         .withPathSensitivity(PathSensitivity.RELATIVE)
     dependsOn(setupNpmrc)
 
-    args = listOf("publish", npmPack.get().archiveFile.get().asFile.absolutePath, "--dry-run")
+    args = provider { listOf("publish", npmPack.get().archiveFile.get().asFile.absolutePath, "--dry-run") }
     if (!project.hasProperty("releasing")) {
         args.addAll("--tag", "next")
     }
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -10,6 +10,7 @@ plugins {
     id("org.openrewrite.build.moderne-source-available-license")
     id("com.netflix.nebula.integtest-standalone")
     id("com.github.node-gradle.node") version "latest.release"
+    id("publishing")
 }
 
 dependencies {
@@ -34,7 +35,7 @@ dependencies {
     integTestRuntimeOnly("org.junit.platform:junit-platform-suite-engine:latest.release")
 }
 
-tasks.withType<Javadoc> {
+tasks.withType<Javadoc>().configureEach {
     // generated ANTLR sources violate doclint
     (options as StandardJavadocDocletOptions).addStringOption("Xdoclint:none", "-quiet")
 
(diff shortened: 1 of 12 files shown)


shanman190:
I'll investigate the build failure for rewrite-javascript. My bet is that it's missing a `npmInstall` task dependency.

rewrite-javascript/build.gradle.kts
```diff
@@ -55,68 +56,64 @@ extensions.configure<NodeExtension> {
...
         include("jest.config.js")
-    })
+    }).withPathSensitivity(PathSensitivity.RELATIVE)
     inputs.files(fileTree("rewrite/src"))
+        .withPathSensitivity(PathSensitivity.RELATIVE)
     inputs.files(fileTree("rewrite/test"))
+        .withPathSensitivity(PathSensitivity.RELATIVE)
     outputs.files("rewrite/build/test-results/jest/junit.xml")
 
-    dependsOn(tasks.named("npmInstall"))
```
knutwannheden, line 68: I can see that we now instead have `inputs.files(npmInstall)`. But is that the same as also depending on the task?

shanman190: It is! As we can see [upstream](https://github.com/node-gradle/gradle-node-plugin/blob/main/src/main/kotlin/com/github/gradle/node/npm/task/NpmInstallTask.kt) the outputs of the `npmInstall` task include the `node_modules` directory which is exactly what we depend upon here for this task to execute properly.

Internally, to Gradle when inputs and outputs are tied together there are dependencies between the tasks included directly from that modeling.

Additionally, Gradle best practices recommend the following:
* By default, model inputs and outputs between tasks and avoid explicit `dependsOn` statements.
* Ignore the above for lifecycle tasks only (ie. build, assemble, check, publish, etc).

rewrite-javascript/build.gradle.kts
```diff
@@ -55,68 +56,64 @@ extensions.configure<NodeExtension> {
...
+    npmCommand = listOf("test")
 }
 
-tasks.register<Test>("npmTestReporting") {
```
shanman190, line 71: Gradle has an npm agent in beta that will allow for reporting of the tests to Develocity.
https://docs.gradle.com/develocity/npm-agent/

This'll achieve the goal where the desire was to publish npm tests alongside the build scan.

timtebeek:
@sambsnyd / @jkschneider this seems a clear performance boost to the builds and the responsiveness in the IDE, but of course we've not been able to test drive the publication for rewrite-javascript while on a branch. Are you comfortable merging this in and watch the snapshot publish from there? Or would you rather hold off until a better moment as to not get in the way of being able to release at any moment?

shanman190:
I'll plan to fix the merge conflicts from the recent JS changes from Knut in order to make this PR mergable again. This'll include optimizing any newly introduced tasks from Knut's PR as well.

rewrite-javascript/build.gradle.kts
```diff
@@ -55,68 +56,64 @@ extensions.configure<NodeExtension> {
...
-        listOf(
-            "version", project.version.toString().replace(
-                "SNAPSHOT",
-                LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMdd.HHmmss"))
-            )
-        )
-    )
-}
-
 // This task creates a `.npmrc` file with the given token, so that the `npm publish` succeeds
```
timtebeek, line 185: On `main` we see this added instead.
```suggestion
val npmVersion = tasks.register("npmVersion", NpmTask::class) {
    val versionProperty = "npmVersionGenerated"

    args.set(provider {
        // Check if we already generated a version for this build
        val existingVersion = project.extensions.findByName(versionProperty) as String?
        if (existingVersion != null) {
            listOf("version", "--no-git-tag-version", existingVersion)
        } else {
            val generatedVersion = project.version.toString().replace(
                "SNAPSHOT",
                LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss"))
            )
            // Store the generated version for reuse
            project.extensions.add(versionProperty, generatedVersion)
            listOf("version", "--no-git-tag-version", generatedVersion)
        }
    })

    workingDir.set(file("rewrite"))
}

val npmResetVersion = tasks.register<NpmTask>("npmResetVersion") {
    args.set(listOf("version", "0.0.0", "--no-git-tag-version"))
    workingDir.set(file("rewrite"))
}

// This task creates a `.npmrc` file with the given token, so that the `npm publish` succeeds
```

rewrite-javascript/build.gradle.kts
```diff
@@ -126,11 +123,11 @@ tasks.register("setupNpmrc") {
...
 // Implicitly `--tag latest` if not specified
-val npmPublish = tasks.named<NpmTask>("npm_publish") {
+val npmPublish = tasks.register<NpmTask>("npmPublish") {
     if (!project.hasProperty("releasing")) {
-        args.set(listOf("--tag", "next"))
+        args = listOf("publish", "--tag", "next")
     }
-    dependsOn(tasks.named("setupNpmrc"))
+    dependsOn(setupNpmrc, npmRunBuild)
 }
```
timtebeek: Also some conflicts here with main:
```gradle
val npmPublish = tasks.named<NpmTask>("npm_publish") {
    dependsOn(tasks.named("setupNpmrc"), npmRunBuild, npmVersion)
    finalizedBy(npmResetVersion)

    args.set(provider {
        buildList {
            add("--dry-run")
            if (!project.hasProperty("releasing")) {
                add("--tag")
                add("next")
            }
        }
    })

    workingDir.set(file("rewrite"))
```

shanman190:
Alright, I ended up reapplying my first commit in isolation without rewrite-javascript on top of the latest main. Then redoing the rewrite-javascript updates because there were a few too many changes to keep the direct merge straight for.

I did opt to take rewrite-javascript a little bit further with this latest refactor, but there's still a lot more that can be done.

rewrite-javascript/build.gradle.kts
```diff
@@ -55,80 +56,91 @@ extensions.configure<NodeExtension> {
     nodeProjectDir.set(projectDir.resolve("rewrite"))
 }
 
-val npmTest = tasks.named("npm_test")
-npmTest.configure {
+val datedSnapshotVersion by extra {
+    if (System.getenv("CI")?.toBoolean() ?: false) {
```
shanman190:
```suggestion
    if (System.getenv("CI") != null) {
```

sambsnyd:
Thanks @shanman190 !

PR #5612, opened 2025-06-12
Fix parsing of import type =

greg-at-moderne:
## What's changed?

Fix parsing of `import type something = require('something')`.

## What's your motivation?

Without the fix, there's no space before `type`:
```
Expected: "import   type my = require('my-library');"
Received: "importtype my = require('my-library');"
```

## OSS repro
https://github.com/microsoft/vscode/blob/74430769757679c680f4679b617a3e428ac8fe1d/extensions/markdown-language-features/src/markdownEngine.ts#L6

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/parser.ts
@@ -3262,7 +3262,14 @@ export class JavaScriptParserVisitor {
             importClause: {
                 kind: JS.Kind.ImportClause,
                 id: randomId(),
-                prefix: this.prefix(node),
+                prefix: (() => {
+                    if (node.isTypeOnly) {
+                        const typeKeyword = node.getChildren().find(n => n.kind === ts.SyntaxKind.TypeKeyword);
+                        return this.prefix(typeKeyword!);
+                    } else {
+                        return emptySpace;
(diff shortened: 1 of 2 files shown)


rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -3262,7 +3262,14 @@ export class JavaScriptParserVisitor {
...
                 id: randomId(),
-                prefix: this.prefix(node),
+                prefix: (() => {
+                    if (node.isTypeOnly) {
+                        const typeKeyword = node.getChildren().find(n => n.kind === ts.SyntaxKind.TypeKeyword);
+                        return this.prefix(typeKeyword!);
+                    } else {
+                        return emptySpace;
+                    }
+                })(),
```
knutwannheden, line 3272: Out out curiosity: Why the indirection via a lambda?

greg-at-moderne: Not much of a reason. The alternative using Elvis operator is OK:
```
prefix: node.isTypeOnly ? this.prefix(node.getChildren().find(n => n.kind === ts.SyntaxKind.TypeKeyword)) : emptySpace,
```

PR #5614, opened 2025-06-12
Return an identifier instead of a memberreference when containing is null

Laurens-W:
## What's changed?

When containing is null we now return an identifier instead of a MemberReference where the containing element is null

## What's your motivation?

After https://github.com/openrewrite/rewrite/pull/5535 we were seeing some
```
org.openrewrite.internal.RecipeRunException: java.lang.NullPointerException: Cannot invoke "org.openrewrite.java.tree.JRightPadded.getElement()" because "this.containing" is null
```
because the LST would get changed here
https://github.com/openrewrite/rewrite/blob/4c012dde6792c62436434026410dbe2eee1e0c99/rewrite-java/src/main/java/org/openrewrite/java/JavaVisitor.java#L1377-L1387

@jkschneider 

### Checklist

### DIFF
--- a/rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8JavadocVisitor.java
+++ b/rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8JavadocVisitor.java
@@ -621,11 +621,14 @@ public class ReloadableJava8JavadocVisitor extends DocTreeScanner<Tree, List<Jav
                         methodRefType
                 );
             } else {
+                if (qualifier == null) {
+                    return new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), name.getSimpleName(), qualifierType, fieldRefType);
+                }
                 return new J.MemberReference(
                         randomId(),
-                        qualifier == null ? Space.EMPTY : qualifier.getPrefix(),
+                        qualifier.getPrefix(),
                         Markers.EMPTY,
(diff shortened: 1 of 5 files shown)


rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8JavadocVisitor.java
```diff
@@ -621,11 +621,14 @@ public Tree visitParam(ParamTree node, List<Javadoc> body) {
...
+                if (qualifier == null) {
+                    return new J.Identifier(randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), name.getSimpleName(), qualifierType, fieldRefType);
+                }
                 return new J.MemberReference(
                         randomId(),
-                        qualifier == null ? Space.EMPTY : qualifier.getPrefix(),
+                        qualifier.getPrefix(),
                         Markers.EMPTY,
-                        JRightPadded.build(qualifier == null ? null : qualifier.withPrefix(Space.EMPTY)),
+                        JRightPadded.build(qualifier.withPrefix(Space.EMPTY)),
```
Laurens-W, line 631: We'd considered returning an empty identifier here but this already happens for example with this piece of JavaDoc:
```
/**
 * {@link #n} refers to Inner
 */
```
so there would be no trivial way to determine whether to place the `#` or not, both are valid JavaDoc

knutwannheden (approved):

PR #5615, opened 2025-06-12
fix(maven): POM file path in deserialization exception message

### LINKED ISSUE #5558: Dependency POM parse failure: rewrite-maven-plugin not telling which file has the issue
Opening the issue here, as as far I could analyze the issue seems to be with `rewrite-maven`.

Version of rewrite-maven: 8.48.1
Version of rewrite-maven-plugin: 6.3.0
Version of maven: 3.9.9
Version of java: JDK 17.0.15

I am using the Maven plugin, and my project is a single module project.
```xml
<plugin>
        <artifactId>rewrite-maven-plugin</artifactId>
        <version>6.3.0</version>
        <dependencies>
          <dependency>
            <groupId>org.openrewrite</groupId>
            <artifactId>rewrite-maven</artifactId>
            <version>8.48.1</version>
          </dependency>
          <dependency>
            <groupId>org.openrewrite</groupId>
            <artifactId>rewrite-java</artifactId>
            <version>8.48.1</version>
          </dependency>
        </dependencies>
        <executions>
          <execution>
            <id>run</id>
            <phase>package</phase>
            <goals>
              <goal>run</goal>
            </goals>
            <configuration>
              <activeRecipes>
                <recipe>org.openrewrite.java.format.AutoFormat</recipe>
              </activeRecipes>
            </configuration>
          </execution>
        </executions>
      </plugin>
```

I am not running any specific recipe, the issue is happening before any recipe is applied.

Reproducer provided:

1. Unzip [reproducer-openrewrite-maven-pom-parse-error.zip](https://github.com/user-attachments/files/20590607/reproducer-openrewrite-maven-pom-parse-error.zip)
2. Run from root:

```
cd incorrect-pom-project
mvn clean install
cd dependent-project
mvn clean install  
```

The path of the POM file that cannot be parsed in the standard output.

Adding `--debug` does not fix the issue.

## What did you see instead?

The path is absent from the output.

[reproducer-openrewrite-maven-pom-parse-error.log](https://github.com/user-attachments/files/20590623/reproducer-openrewrite-maven-pom-parse-error.log)

Yes

edeweerd1A:
- Fixes #5558.

It is probably a very narrow corner case.
Reproduced with a POM file with an incorrect XML syntax:

```xml
<!-- EMPTY LINE -->
<project>
```

Notice the `<!-- EMPTY LINE -->` before `<?xml` opening tag. In the original reproducer furnished with the issue linked it is really an empty line not a XML comment.
Seems Maven parsers are very confortable with this but not Jackson Databind XML.

~Here we add an argument to RawPom.parse with the Path to the pom file being parsed. In case of exception the path is added to the exception message.~

Tests adapted and 1 additional test to ensure the POM path is in the exception message.

## What's changed?

rewrite-maven and RawPom class

## What's your motivation?

Enhance user experience by telling which POM could not be deserialized by Jackson.

The failing run of the rewrite-maven-plugin, as communicated in the linked issue, does not tell which is the POM file that failed in its deserialization, even with --debug flag.

### Checklist
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files --> Sorry I could not find where the configuration of the formatter to use.


timtebeek:
Thanks for the help here @edeweerd1A ! Just for my own understanding: did the recently added diagnostics not already help pinpoint the cause? Just to know if paths are missed there for other unexpected failures.

Also downstream from this change we'd need to update the [usage in rewrite-maven-plugin](https://github.com/openrewrite/rewrite-maven-plugin/blob/43199377036135d417cdbb8d5c7ab2568197d403/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L867) as well. Wondering if we should do that backwards compatible or not; checking potential impact here.

edeweerd1A:
> Thanks for the help here @edeweerd1A ! Just for my own understanding: did the recently added diagnostics not already help pinpoint the cause? Just to know if paths are missed there for other unexpected failures.
> 
> Also downstream from this change we'd need to update the [usage in rewrite-maven-plugin](https://github.com/openrewrite/rewrite-maven-plugin/blob/43199377036135d417cdbb8d5c7ab2568197d403/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L867) as well. Wondering if we should do that backwards compatible or not; checking potential impact here.

Are you referring at running the [following recipe DependencyResolutionDiagnostic](https://docs.openrewrite.org/recipes/java/dependencies/dependencyresolutiondiagnostic#data-tables) ? because the failure is happening at POM resolution time, so no single recipe will be applied.

About doing the reporting in the maven plugin I let you decide on this.

timtebeek:
Added some more details here now that I've found some time for a closer look:
- https://github.com/openrewrite/rewrite/issues/5558#issuecomment-2971150138

Would you mind chiming in there (for SEO) on whether that alternative approach is helpful?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -680,7 +680,7 @@ public Pom download(GroupArtifactVersion gav,
                             //If the exception is a common, client-side exception, cache an empty result.
                             mavenCache.putPom(resolvedGav, null);
                         }
-                    } catch (IOException e) {
+                    } catch (IOException | UncheckedIOException e) {
```
timtebeek, line 683: `UncheckedIOException` can be thrown here by `RawPom.parse`:
https://github.com/openrewrite/rewrite/blob/ca705bcad67d2ef4ce86adbfce9ec4bc676273c1/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java#L153-L163

Or by OkHttpSender:
https://github.com/openrewrite/rewrite/blob/cdb945ff320f58af236f844c66a60c3d3582219f/rewrite-maven/src/test/java/org/openrewrite/maven/http/OkHttpSender.java#L72-L79

Or by HttpUrlConnectionSender:
https://github.com/openrewrite/rewrite/blob/52c770380b60955facd93db47d91a0aa2ae09462/rewrite-core/src/main/java/org/openrewrite/ipc/http/HttpUrlConnectionSender.java#L71-L126

timtebeek: I did not find additional cases where we `catch(IOException ...)` and ourselves `throw new UncheckedIOException`. I'd think it's an improvement to now always catch those `UncheckedIOException`s as well, and treat them the same, such that we get the uniform `MavenDownloadingException` along with details needed to follow up, as opposed to the missing details reported in #5558.

timtebeek: @jkschneider The above two lines are all that need a quick review here; I've contextualized the potential impact above and think this looks good, but I don't want to push this through on a weekend without your ok.

jkschneider: LGTM

PR #5617, opened 2025-06-13
Javascript - fixing parsing of export import...

greg-at-moderne:
## What's changed?

In Javascript fixing parsing of a special case of `export import ...` statements.

## What's your motivation?

Otherwise the parser fails to recognize it properly and swallows the `export` keyword.


rewrite-javascript/rewrite/src/javascript/print.ts
```diff
@@ -1777,6 +1783,8 @@ export class JavaScriptPrinter extends JavaScriptVisitor<PrintOutputCapture> {
     protected async visitRightPadded<T extends J | boolean>(right: J.RightPadded<T>, p: PrintOutputCapture): Promise<J.RightPadded<T>> {
         if (isTree(right.element)) {
             await this.visit(right.element, p);
+        } else if (typeof right.element !== 'boolean' && typeof right.element === 'object' && 'kind' in right.element && (right.element as any).kind === J.Kind.Modifier) {
+            await this.visitModifier(right.element as J.Modifier, p);
```
knutwannheden: I don't understand this part. AFAICT a `J.Modifier` should be a tree. Can we check why `isTree()` thinks differently?

rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -3259,6 +3259,18 @@ export class JavaScriptParserVisitor {
             kind: JS.Kind.Import,
             id: randomId(),
             prefix: this.prefix(node),
+            exportModifier: ( () => {
+                const exportModifier = node.modifiers?.find(m => m.kind === ts.SyntaxKind.ExportKeyword);
```
knutwannheden, line 3260: Since the TS Compiler API is prepared for multiple modifiers, maybe we should do the same?

rewrite-javascript/rewrite/src/javascript/print.ts
greg-at-moderne: Hmm... It works without it:
68f043863
But I am pretty sure I added it for a reason. Can't tell right now what it was :)

rewrite-javascript/rewrite/src/javascript/parser.ts
greg-at-moderne: OK: e55ea8b29

BTW, @knutwannheden, is `List<JRightPadded<J.Modifier>>` OK?
I am not fully sure.

knutwannheden: `List<J.Modifer>` should be enough here. The right-padding would be in case we have something beyond a space in between the modifiers (like a comma or a semicolon).

greg-at-moderne: > List<J.Modifer> should be enough here.

b310d49df

Issue #5329, opened 2025-04-22
Java parser idempotency issue when using generics, annotations and arrays


greg-at-moderne:
## What version of OpenRewrite are you using?

Current main = a344e18ff346684b0ad9d18cedc57868c1ca0ab6

## What is the smallest, simplest way to reproduce the problem?
```java
    @Test
    void arraysWithAnnotations() {
        rewriteRun(
          java(
            """
            import java.lang.annotation.ElementType;
            import java.lang.annotation.Target;
            
            class A {
               @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
               private static @interface C {
               }

               Comparable<@C Object @C []> specialArray; 
            }
            """
          )
        );
    }
```
fails with:
```
[When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error for "A.java":
```

```diff
-   Comparable<@C Object @C []> specialArray;
+   Comparable<Objectect @C []> specialArray;
```

## Context

OSS repro: https://github.com/classgraph/classgraph/blob/6f9012f2a193ebfefe4a4384e7642820e7aab0f5/src/test/java/io/github/classgraph/issues/issue402/TypeAnnotationTest.java#L65

PR #5618, opened 2025-06-13
Fix: handle nested type annotations on array

e5LA:
## Any additional context
For the following case: `@C Object @B []` 

Currently the `annotationPosTable` in `visitAnnotatedType` method contains only 1 annotation: 
```
annotationPosTable = {HashMap@5269}  size = 1
 {Integer@5275} 365 -> {JCTree$JCAnnotation@5276} "@B"
```
which makes `leadingAnnotations` to be empty (`size = 0`). That's wrong for `JCAnnotatedType` like `@C Object`

## What's changed?
Adding annotation to the `annotationPosTable` when handling `JCAnnotatedType`.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5329 

Not. 

### Checklist


timtebeek (approved):
Indeed thanks for yet another quality fix in the parsers. Much appreciated!

PR #5623, opened 2025-06-13
Initial proposal that will add DependencyManagement handling

Jenson3210:
**EARLY REVIEW REQUEST FOR EARLY FEEDBACK** -> Not final review yet. 

## What's changed?
By using a Trait we can both match dependencyManagement sections from recipe perspectives. I also wrote a method that updates the Trait's cursor with an updated dependencyManagement. We can have several more overload methods that call this one with null. Passing null for `group`, `artifact` or `version` makes these fields untouched. 

## What's your motivation?
Fix #5304

For now this is only in draft as I see too much repetition in the code. I want to go over it a second time optimizing reusable parts to private methods etc... 

@shanman190 or @sambsnyd would be great for an early/conceptual review. Please do not focus on code duplications etc yet. I want to fail fast before spending time in optimizing.

I think the way to introduce this is shown in the ChangeDependency recipe. It's really clear over there. 
ATM, the Trait does not allow for templated strings yet as we would need scanning phases for that so want to come up with support for no-templates strings first and then add templated strings in a separate PR.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -43,6 +43,7 @@ import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
 import org.openrewrite.semver.DependencyMatcher;
 import org.openrewrite.semver.Semver;
+import org.openrewrite.trait.Trait;
 
 import java.util.*;
 
@@ -82,9 +83,9 @@ public class ChangeDependency extends Recipe {
 
     @Option(displayName = "New version",
             description = "An exact version number or node-style semver selector used to select the version number. " +
-                          "You can also use `latest.release` for the latest available version and `latest.patch` if " +
-                          "the current version is a valid semantic version. For more details, you can look at the documentation " +
-                          "page of [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors).",
+                    "You can also use `latest.release` for the latest available version and `latest.patch` if " +
+                    "the current version is a valid semantic version. For more details, you can look at the documentation " +
+                    "page of [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors).",
             example = "29.X",
             required = false)
     @Nullable
@@ -92,7 +93,7 @@ public class ChangeDependency extends Recipe {
 
     @Option(displayName = "Version pattern",
             description = "Allows version selection to be extended beyond the original Node Semver semantics. So for example," +
-                          "Setting 'version' to \"25-29\" can be paired with a metadata pattern of \"-jre\" to select Guava 29.0-jre",
+                    "Setting 'version' to \"25-29\" can be paired with a metadata pattern of \"-jre\" to select Guava 29.0-jre",
             example = "-jre",
             required = false)
     @Nullable
@@ -100,8 +101,8 @@ public class ChangeDependency extends Recipe {
 
     @Option(displayName = "Override managed version",
             description = "If the old dependency has a managed version, this flag can be used to explicitly set the version on the new dependency. " +
-                          "WARNING: No check is done on the NEW dependency to verify if it is managed, it relies on whether the OLD dependency had a managed version. " +
-                          "The default for this flag is `false`.",
+                    "WARNING: No check is done on the NEW dependency to verify if it is managed, it relies on whether the OLD dependency had a managed version. " +
+                    "The default for this flag is `false`.",
             required = false)
     @Nullable
     Boolean overrideManagedVersion;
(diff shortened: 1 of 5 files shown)


shanman190:
When I think about the logistics of it, it feels like `ChangeDependency` should be altering an external dependency that the project explicitly pulls in. The Spring Dependency Management plugin is more about declaring a managed dependency which at least to me is more akin to declaring a dependency constraint. Neither has any effect upon dependencies until it exists within the dependency graph.

What if we introduced a `ChangeManagedDependency` and had it update the constraints and dependency management sections? Like in Maven managed dependency has no effect until the dependency is actually present, so I think this aligns more closely in expectations.

Jenson3210:
@shanman190, 
- you would then have to call both these recipes from everywhere you want to bump a dependency? Or would you still call it from the `ChangeDependency`? 
Are we then not also going to have to create `UpgradeManagedDependencyVersion` etc? 
Or do they all delegate to that one? 
- You then mean not to put the actual `dependencyManagement` change in the `Trait` but in a recipe? Or is it fine to keep it in the `Trait` and then from that recipe for the Spring's dependency management part use the `Trait` over there and for constraints use another `visitor/Trait` and...?

shanman190:
Here's my analysis of rewrite-maven as a parallel:

- `maven.AddDependency` - only add direct dependency
- `maven.AddManagedDependency` - only add managed dependency
- `maven.ChangeDependencyGroupIdAndArtifactId` - has option defaulted to true to change managed dependencies as well via doAfterVisit
- `maven.ChangeManagedDependencyGroupIdAndArtifactId` - only change managed dependency
- `maven.RemoveDependency` - only remove direct dependency
- `maven.RemoveManagedDependency` - only remove managed dependency
- `maven.UpgradeDependencyVersion` - upgrades a direct or managed dependency

So it would make sense I suppose to follow the established pattern. Honestly, I didn't realize that the `maven.ChangeDependencyGroupIdAndArtifactId` had been updated to effect managed dependencies by default. I probably would have kept them separate following the single responsibility principle myself, but at least the recipes are separate with the one invoking the other when requested.

I think having the trait is fine, but I think the design that I think would work best is a `gradle.ChangeManagedDependency` which encompasses changing a constraint using the trait from https://github.com/openrewrite/rewrite/pull/5604 and then your dependency management trait here. Then we update `gradle.ChangeDependency` to have an option akin to the Maven variant and when that option is true execute to the `gradle.ChangeManagedDependency` recipe in a doAfterVisit.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -82,36 +81,58 @@ public class ChangeDependency extends Recipe {
...
+    @Nullable
+    Boolean changeManagedDependency;
+
     // Individual dependencies tend to appear in several places within a given dependency graph.
     // Minimize the number of allocations by caching the updated dependencies.
     transient Map<org.openrewrite.maven.tree.Dependency, org.openrewrite.maven.tree.Dependency> updatedRequested = new HashMap<>();
     transient Map<org.openrewrite.maven.tree.ResolvedDependency, org.openrewrite.maven.tree.ResolvedDependency> updatedResolved = new HashMap<>();
     transient MavenMetadataFailures mavenMetadataFailures = new MavenMetadataFailures(this);
 
+    @JsonCreator
```
Jenson3210, line 121: I still need to find a way to have this here without it failing but without `@JsonCreator` as @jkschneider mentioned using `@JsonCreator` is to be avoided. I looked at the other maven counterpart, but there it is done using the annotation.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -173,8 +194,12 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
                     sourceFile = (JavaSourceFile) super.visit(sourceFile, ctx);
                     if (sourceFile != tree) {
                         sourceFile = sourceFile.withMarkers(sourceFile.getMarkers().setByType(updateGradleModel(gradleProject)));
+                        if (changeManagedDependency == null || changeManagedDependency) {
```
shanman190, line 198:
```suggestion
                        if (Boolean.TRUE.equals(changeManagedDependency)) {
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeManagedDependency.java
```diff
@@ -0,0 +1,227 @@
...
+    String versionPattern;
+
+    // Individual dependencies tend to appear in several places within a given dependency graph.
+    // Minimize the number of allocations by caching the updated dependencies.
+    transient Map<org.openrewrite.maven.tree.Dependency, org.openrewrite.maven.tree.Dependency> updatedRequested = new HashMap<>();
+    transient Map<org.openrewrite.maven.tree.ResolvedDependency, org.openrewrite.maven.tree.ResolvedDependency> updatedResolved = new HashMap<>();
+
+    @Override
+    public String getDisplayName() {
+        return "Change Gradle dependency";
```
shanman190:
```suggestion
        return "Change Gradle managed dependency";
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeManagedDependency.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public String getInstanceNameSuffix() {
+        return String.format("`%s:%s`", oldGroupId, oldArtifactId);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Change a Gradle dependency coordinates. The `newGroupId` or `newArtifactId` **MUST** be different from before.";
```
shanman190:
```suggestion
        return "Change a Gradle managed dependency coordinates. The `newGroupId` or `newArtifactId` **MUST** be different from before.";
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencyManagementEntry.java
```diff
@@ -0,0 +1,999 @@
...
+                            group = (String) value.getValue();
+                        } else if ("version".equals(name)) {
+                            version = (String) value.getValue();
+                        }
+                    }
+
+                    if (group != null) {
+                        dependency = new GroupArtifactVersion(group, entry, version);
+                    }
+                } else if (argument instanceof J.MethodInvocation && "mapOf".equals(((J.MethodInvocation) argument).getSimpleName())) {
```
shanman190: callout: Not a part of this PR, but this may be an improvement that we can make for the `GradleDependency` trait as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencyManagementEntry.java
```diff
@@ -0,0 +1,999 @@
...
+import org.openrewrite.trait.Trait;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+import static org.openrewrite.internal.StringUtils.matchesGlob;
+
+@Value
+public class GradleDependencyManagementEntry implements Trait<J.MethodInvocation> {
```
shanman190: question: is there an idea to make this not Spring Dependency Management plugin specific and more of a general purpose finder? If, not maybe we make the class name more specific to Spring?

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
shanman190: On second though, null here is valid because of the wrapper type and deserialization.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencyManagementEntry.java
Jenson3210: As discussed offline -> renaming the Trait

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencyManagementEntry.java
Jenson3210: 👍  Leaving this open for now so that we can create a new issue from this comment later on

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/Traits.java
```diff
@@ -23,6 +23,10 @@ public static GradleDependency.Matcher gradleDependency() {
         return new GradleDependency.Matcher();
     }
 
+    public static SpringDependencyManagementPluginEntry.Matcher springDependencyManagementPluginEntry() {
+        return new SpringDependencyManagementPluginEntry.Matcher();
+    }
```
shanman190: As @timtebeek confirmed for me, we're no longer going to use the factory pattern going forward, so we can drop this method.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
sambsnyd: I'm not aware of any alternatives. If you have multiple constructors, and it can make sense to do so briefly for backwards compatibility, one of them must be annotated with `@JsonCreator` 🤷

Issue #5296, opened 2025-04-14
Error parsing Groovy with double-generics and anonymous class instantiation.


greg-at-moderne:
## What version of OpenRewrite are you using?

current main = b504c17eb4055e959ac7d5edd5616c84664667f2

## What is the smallest, simplest way to reproduce the problem?
```
    @Test
    void classEnum() {
        rewriteRun(groovy("new ArrayList<Map<String, String>>() {}"));
    }
```
fails with:
```diff
- new ArrayList<Map<String, String>>() {}
+ new ArrayList<Map,String, String>()>() {}
```

## Context

Found the same failure in https://github.com/spinnaker/spinnaker/blob/97cee4f3da768323984a4e1f404c96a9addf5306/clouddriver/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppengineHealth.groovy#L41C1-L43C4

PR #5625, opened 2025-06-14
Fix parsing of generics in anonymous class in Groovy

e5LA:
## What's changed?
1) For anonymous class instantiations like `new ArrayList<Map<String, String>>() {}`, we now call `visitTypeParameterizations` with `classNode.getUnresolvedSuperClass().getGenericsTypes()` instead of `classNode.getGenericsTypes()` which is `null` in this case. 

2) After adding a test case for `new ArrayList<>() {}` - it failed as well. It turned out that  `ctor.getNodeMetaData(StaticTypesMarker.INFERRED_TYPE)` was not sufficient to detect inferred generics. Looks like Groovy doesn't always set this metadata? 
To address this, I introduced private helper method `isInferred(ConstructorCallExpression)` that detects inferred generic usage (e.g., diamond operator).

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5296

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1374,7 +1374,7 @@ public class GroovyParserVisitor {
                 }
 
                 skip("new");
-                TypeTree clazz = visitTypeTree(ctor.getType(), ctor.getNodeMetaData().containsKey(StaticTypesMarker.INFERRED_TYPE));
+                TypeTree clazz = visitTypeTree(ctor.getType(), isInferred(ctor));
                 JContainer<Expression> args = visit(ctor.getArguments());
                 J.Block body = null;
                 if (ctor.isUsingAnonymousInnerClass() && ctor.getType() instanceof InnerClassNode) {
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -3015,6 +3025,21 @@ private static ClassNode staticType(Parameter parameter) {
...
+        }
+
+        ClassNode innerClass = ctor.getType();
+        if (!(innerClass instanceof InnerClassNode)) {
+            return false;
+        }
+
+        ClassNode superClass = innerClass.getUnresolvedSuperClass();
+        GenericsType[] generics = superClass != null ? superClass.getGenericsTypes() : null;
+        return generics != null && generics.length == 0;
```
e5LA, line 3040: In case of `new ArrayList<>() {}` : `superClass.getGenericsTypes()` is: `[]`

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2443,7 +2443,17 @@ private <T extends TypeTree & Expression> T typeTree(@Nullable ClassNode classNo
 
         assert expr != null;
         if (classNode != null) {
-            if (classNode.isUsingGenerics() && !classNode.isGenericsPlaceHolder()) {
+            boolean isAnonymousClassWithGenericSuper = classNode instanceof InnerClassNode
+                    && classNode.getUnresolvedSuperClass() != null
+                    && classNode.getUnresolvedSuperClass().isUsingGenerics()
+                    && !classNode.getUnresolvedSuperClass().isGenericsPlaceHolder()
+                    && classNode.getGenericsTypes() == null;
+            if (isAnonymousClassWithGenericSuper) {
```
e5LA, line 2451: Not sure if there is better way to distinguish this case..

timtebeek:
Thanks a lot for yet another fix @e5LA ! I've tagged @greg-at-moderne for review as he's the better judge here.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
greg-at-moderne: I fail to know either.

Having said that, I don't see any problem with the code change as-is.

greg-at-moderne (approved):

greg-at-moderne:
The solution provided LGTM. I don't see any problems. And if it solved additional cases, I am all for merging it.

Thank you one more time!

greg-at-moderne:
(accidentally clicked Close, sorry)

PR #5630, opened 2025-06-16
UpgradeTransitiveDependencyVersion maven recipe doesn't update all matching transitive dependencies with * wildcard for artifact ID

jevanlingen:
## What's changed?
A user may use the `*` wildcard identifier when declaring the artifact ID for replacement. It seems as if it only adds the version for the last matching transitive dependency in case there's more than one. This did not work for the Maven's variant of the recipe. 

## What's your motivation?
Make Gradle and Maven `UpgradeTransitiveDependencyVersion` recipes act similar.

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
@@ -958,7 +958,7 @@ class UpgradeTransitiveDependencyVersionTest implements RewriteTest {
     }
 
     @Test
-    void canUseAnyWildcardForArtifactId() {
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
```diff
@@ -957,6 +957,41 @@ void canHandleNullScannedAccumulator() {
         );
     }
 
+    @Test
+    void canUseAnyWildcardForArtifactId() {
```
bmuschko: I'd probably rename the test method to `.canUseAnyWildcardForMultipleMatchingArtifactIds`

bmuschko (approved):

JohannisK (approved):

PR #5633, opened 2025-06-17
UpgradeDependencyVersion recipe does not update properties defined in parent  

jevanlingen:
## What's changed?
When a parent build.gradle defines a property and it's used in a submodule, this property is updated as well.

### Checklist


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -801,6 +801,111 @@ void versionInParentSubprojectDefinitionWithPropertiesFiles() {
         );
     }
 
+    @Test
+    void versionInParentPropertiesFiles() {
```
jevanlingen, line 914: 💡 This test did already pass before this PR. It's just an extra test to prove this kind of setup does also work.

okundzich:
please give time to @jkschneider to review

PR #5634, opened 2025-06-17
Fix for situation that could bypass a check inadvertently on `UpdateMavenProjectPropertyJavaVersion`

steve-aom-elliott:
Adding test case and fix for situation where if the `maven-compiler-plugin` made use of `java.version` or similar for its `target` or `release` configuration settings, it could bypass the check for whether the version was already high enough, which caused all the `java.version`s to get downgraded potentially.

### Checklist


timtebeek (approved):
Nicely spotted!

PR #5636, opened 2025-06-18
js rpc

knutwannheden:
- **Remove `extractBundledInstallation` from API**
- **Remove `batchSize()` from Builder API**
- **JS server will now append to given log file**

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
@@ -15,6 +15,8 @@
  */
 package org.openrewrite.rpc;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import io.moderne.jsonrpc.JsonRpc;
 import io.moderne.jsonrpc.JsonRpcMethod;
 import io.moderne.jsonrpc.JsonRpcRequest;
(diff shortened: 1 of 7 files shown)


rewrite-core/src/main/java/org/openrewrite/rpc/request/GetRef.java
```diff
@@ -0,0 +1,99 @@
...
+    public static class Handler extends io.moderne.jsonrpc.JsonRpcMethod<GetRef> {
+        private static final ExecutorService forkJoin = ForkJoinPool.commonPool();
+
+        private final Map<Integer, Object> remoteRefs;
+        private final IdentityHashMap<Object, Integer> localRefs;
+        private final AtomicInteger batchSize;
+        private final AtomicBoolean trace;
+
+        private final Map<Integer, BlockingQueue<List<RpcObjectData>>> inProgress = new ConcurrentHashMap<>();
+
```
sambsnyd, line 48: So is the idea here that one side or the other of an RPC setup owns the reference to a particular object and that can be used to figure out if it is safe to evict it from a cache?

knutwannheden: The idea is that either party should be allowed to cache both "objects" and "refs" (which ultimately might converge) without needing to coordinate the cache eviction with the other party. When a "ref" is unknown, it will then just send a "GetRef" request.

PR #5638, opened 2025-06-18
Javascript - programmatic comparator

greg-at-moderne:
## What's changed?

Changing the implementation of `JavaScriptComparatorVisitor` to use reflection and programmatic comparison of objects instead of hundreds of lines of generated code.

## What's your motivation?

I find it hard to maintain and ensure consistency of the comparator when applying changes to the LST elements.


rewrite-javascript/rewrite/src/javascript/comparator.ts
```diff
@@ -79,44 +78,36 @@ export class JavaScriptComparatorVisitor extends JavaScriptVisitor<J> {
...
-        if (!this.match) return binary;
+        for (const key of Object.keys(j)) {
+            const value = (j as any)[key];
+            const otherValue = (p as any)[key] as J;
 
-        await this.visit(binary.right, otherBinary.right);
-        return binary;
+            if (isTree(value) && isTree(otherValue)) {
+                await this.visit(value, otherValue, parent);
+            } else if (["value", "valueSource", "simpleName", "terminatedWithSemicolon", "parenthesized", "codePoint", "keyword", "implicit", "text", "typeOnly", "exportEquals", "multiline", "element"].includes(key)) {
```
knutwannheden, line 91: I'd rather we move this if into some method and if we there either also check the type of the property owner or instead check the type of the property value (e.g. check for number, string, boolean, etc).

knutwannheden: Thinking a bit more, I'd rather keep the current structure with all overrides and then have some kind of defaultVisit() method which they all delegate to and which essentially contains this code. The reason being that we still want to customize individual methods to implement a semantic equality, rather than a basically pure syntactic check (minus whitespace). Also, we need to properly check type attribution as well as markers.

greg-at-moderne: OK. Then we need some mechanism to enforce the completeness of the checks. A unit test to read the `tree.ts` and `comparator.ts` code?

greg-at-moderne: > Then we need some mechanism to enforce the completeness of the checks.

#5655

PR #5639, opened 2025-06-19
Recipe to find maven modules that have no parent

JohannisK:
## What's changed?
Added a recipe to mark Maven Modules that have no parent

## What's your motivation?
When adding recipe (as part of it's changes) adds a maven parent, the current implementation of the AddParentPom recipe does notting when a parent is already present.

Especially when changes in the recipe depend on the correct parent being present this will cause unexpected and breaking changes.

With this recipe the module can be marked when there is no parent, so it can be used as a precondition for a recipe that requires the correct parent to be present.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasNoParent.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasNoParent.java
@@ -28,37 +28,25 @@ import java.util.Set;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
-public class ModuleHasPlugin extends ScanningRecipe<ModuleHasPlugin.Accumulator> {
+public class ModuleHasNoParent extends ScanningRecipe<ModuleHasNoParent.Accumulator> {
 
     @Override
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasNoParent.java
```diff
@@ -0,0 +1,97 @@
...
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                assert tree != null;
+                Optional<JavaProject> maybeJp = tree.getMarkers().findFirst(JavaProject.class);
+                if (!maybeJp.isPresent()) {
+                    return tree;
+                }
+                JavaProject jp = maybeJp.get();
+                if (!acc.getProjectsWithParent().contains(jp)) {
```
Laurens-W, line 90: What is the reasoning for this recipe being a scanning recipe?
I don't see why you couldn't just call `parentPomInsight` here and save the scanning cycle of the recipe :)

PR #5640, opened 2025-06-19
UpgradeDependencyVersionTest recipe support for dependency-management plugin

jevanlingen:
## What's changed?
Dependencies defined in a dependencyManagement block are updated as well.

## What's your motivation?
Currently, using the [Dependency Management](https://plugins.gradle.org/plugin/io.spring.dependency-management) plugin does not yield any changes.

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
@@ -1298,6 +1298,81 @@ class UpgradeDependencyVersionTest implements RewriteTest {
     }
 
     @Test
+    void updateDependencyManagementVersion() {
+        rewriteRun(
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -1437,6 +1512,38 @@ void kotlinDslStringInterpolation() {
         );
     }
 
+    @Test
+    void kotlinVersionInPropertiesFile() {
```
jevanlingen, line 1516: 💡 This test does not really have anything to do with this PR; just an (already succeeding) missing test we need to have covered as well.

bmuschko (commented):
Please see my comment on the originating issue that drives the test cases. I personally think that we should not support the Spring Boot dependency management plugin directly. There should be a recipe that migrate away from it instead.

I am proposing not to add the tests in this PR until we make a decision.

shanman190:
Related PR:  https://github.com/openrewrite/rewrite/pull/5623

PR #5644, opened 2025-06-19
Skip update if existing compiler plugin version config is newer

bmuschko:
## What's changed?

Refined the logic that would determine if Maven compiler plugin needs to be updated based on the existing configuration.

## What's your motivation?

Given the following Maven project that configures Java 17 for the source/target compatibility via the Maven compiler plugin.

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>com.example</groupId>
    <artifactId>demo</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <build>
        <plugins>
          <plugin>
            <artifactId>maven-compiler-plugin</artifactId>
            <version>3.14.0</version>
            <configuration>
              <source>17</source>
              <target>17</target>
            </configuration>
          </plugin>
        </plugins>
      </build>
</project>
```

Running the [Java8toJava11](https://docs.openrewrite.org/recipes/java/migrate/java8tojava11) recipe produces a fix.patch even though the currently used Java version is already higher than 11.

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/UseMavenCompilerPluginReleaseConfigurationTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/UseMavenCompilerPluginReleaseConfigurationTest.java
@@ -16,10 +16,15 @@
 package org.openrewrite.maven;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.Issue;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
+
+import java.util.stream.Stream;
 
 import static org.openrewrite.java.Assertions.mavenProject;
 import static org.openrewrite.maven.Assertions.pomXml;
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/UseMavenCompilerPluginReleaseConfigurationTest.java
```diff
@@ -621,4 +626,86 @@ void replacesSourceAndTargetConfig_InProfileSection() {
           )
         );
     }
+
+    @ParameterizedTest
+    @MethodSource("mavenCompilerPluginConfig")
```
timtebeek, line 631: Would `@ValueSource` not make more sense here? Saves some indirection

timtebeek (approved):
Approved already, with comment for the test

rewrite-maven/src/test/java/org/openrewrite/maven/UseMavenCompilerPluginReleaseConfigurationTest.java
bmuschko: The method header would probably become a little bit unreadable with having those Strings spelled out with `@ValueSource` directly. That's why I feel the `@MethodSource` is a better fit.

PR #5645, opened 2025-06-20
RemoveRedundantDependencyVersions recipe removes unused properties

jevanlingen:
## What's changed?
RemoveRedundantDependencyVersions recipe removes unused properties if the recipe removes a version with said properties.

## What's your motivation?
After the recipe in its current state has run, it's possible there are now properties in the properties block that are no longer used.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
@@ -28,6 +28,7 @@ import org.openrewrite.semver.ExactVersion;
 import org.openrewrite.semver.LatestIntegration;
 import org.openrewrite.semver.Semver;
 import org.openrewrite.semver.VersionComparator;
+import org.openrewrite.xml.tree.Content;
 import org.openrewrite.xml.tree.Xml;
 
 import java.util.*;
@@ -173,7 +174,7 @@ public class RemoveRedundantDependencyVersions extends Recipe {
                     ResolvedDependency d = findDependency(tag);
                     if (d != null && matchesGroup(d) && matchesArtifact(d) && matchesVersion(d) && isNotExcepted(d)) {
                         Xml.Tag version = tag.getChild("version").orElse(null);
-                        return tag.withContent(ListUtils.map(tag.getContent(), c -> c == version ? null : c));
+                        return tag.withContent(withoutVersion(tag, version));
                     }
                 } else if (isManagedDependencyTag()) {
                     ResolvedManagedDependency managed = findManagedDependency(tag);
@@ -199,17 +200,30 @@ public class RemoveRedundantDependencyVersions extends Recipe {
                                 return null;
                             }
                             // some other element is also declared (executions, configuration, dependencies…), so just remove the version
-                            return tag.withContent(ListUtils.map(tag.getContent(), c -> c == version ? null : c));
+                            return tag.withContent(withoutVersion(tag, version));
                         }
                     } else {
                         Plugin p = findPlugin(tag);
                         if (p != null && matchesGroup(p) && matchesArtifact(p) && matchesVersion(p)) {
                             Xml.Tag version = tag.getChild("version").orElse(null);
-                            return tag.withContent(ListUtils.map(tag.getContent(), c -> c == version ? null : c));
+                            return tag.withContent(withoutVersion(tag, version));
                         }
                     }
                 }
                 return super.visitTag(tag, ctx);
+            }
+
+            private @Nullable List<? extends Content> withoutVersion(Xml.Tag tag, Xml.@Nullable Tag version) {
+                return ListUtils.map(tag.getContent(), c -> {
+                    if (c == version) {
+                        if (((Xml.Tag) c).getValue().isPresent() && ((Xml.Tag) c).getValue().get().contains("${")) {
+                            final String propertyName = ((Xml.Tag) c).getValue().get().replace("${", "").replace("}", "");
+                            doAfterVisit(new RemoveUnusedProperties(propertyName).getVisitor());
+                        }
+                        return null;
+                    }
+                    return c;
+                });
             }
 
             private boolean matchesGroup(ResolvedManagedDependency d) {
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
@@ -1763,6 +1763,67 @@ class RemoveRedundantDependencyVersionsTest implements RewriteTest {
     }
 
     @Test
+    void unusedPluginPropertiesAreRemoved() {
+        rewriteRun(
+          spec -> spec.recipe(new RemoveRedundantDependencyVersions(null, null, RemoveRedundantDependencyVersions.Comparator.GTE, null)),
+          pomXml(
+            """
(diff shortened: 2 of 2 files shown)


jevanlingen:
Don't merge until @jkschneider has time to review. No high priority either to get this merged!

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -1762,6 +1762,67 @@ void pluginPropertySubstitution() {
         );
     }
 
+    @Test
```
Jenson3210, line 1827: Should we add a test that verifies used parameters are not removed?

jevanlingen: I thought about this, but discarded the idea. That test does already exist in the test file of RemoveUsedProperties recipe:

https://github.com/openrewrite/rewrite/blob/a3210365767fd12e58286d76c93b10c2016c3551/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveUnusedPropertiesTest.java#L207-L258

Doing it over here again does not make sense to me (in the sense, you should be able to trust the building blocks you are using; in this case the building block is "just" another recipe). What do you think?

Jenson3210: I think there is no test that we use this visitor for the cleanup meaning that we should not "rely" on the fact that this is covered down the line. 
Similarly, one could argue then that the other 2 tests are also not needed then. 

Not sure if possible -> can we test if that recipe was postVisited? Then there would be no need for the tests, just for the one that verifies we called the other recipe and we can rely on the other tests being complete

jevanlingen: > I think there is no test that we use this visitor for the cleanup meaning that we should not "rely" on the fact that this is covered down the line.
Similarly, one could argue then that the other 2 tests are also not needed then.

I don't really agree with this argumentation. The current tests prove the building block "RemoveUsedProperties recipe" has been used properly. Test coverage for everything a building block does is something totally different than having a test to prove the building block has been used. It like having a test if a car contains wheels vs having a lot of tests to prove the wheels also contains rims, tires, etc.

> Not sure if possible -> can we test if that recipe was postVisited? Then there would be no need for the tests, just for the one that verifies we called the other recipe and we can rely on the other tests being complete.

I don't know if it's possible, but it makes the tests not easier to read. This PR states "if a version with a property is used, remove it from the properties list if possible". The RemoveUsedProperties does already take care of most of the problem. But let me just alter the current tests, I think I can catch that "if possible" part quite easily without introducing to much extra lines of code.

Jenson3210: > The current tests prove the building block "RemoveUsedProperties recipe" has been used properly.

That's my point exactly. The current tests do not prove that imo. The current test shows that there is some logic that removes properties. Not that it delegated that to the "RemoveUsedProperties recipe" building block.

> I don't know if it's possible, but it makes the tests not easier to read

I do not know how it would look. But if we can "only" test that it was called, that would be sufficient (eg. not even having to test that the versions are gone (though we might want to add 1 to be sure)

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -1818,6 +1879,65 @@ void dependencyPropertySubstitution() {
...
+                    <version>2.5.4</version>
+                    <relativePath/>
+                </parent>
+
+                <groupId>com.example</groupId>
+                <artifactId>test</artifactId>
+                <version>1.0.0-SNAPSHOT</version>
+                <properties>
+                    <spring-web.version>5.3.9</spring-web.version>
+                </properties>
```
timtebeek, line 1985: I have some concerns about removing properties like this; in a multi module project we could remove a property from a parent that's still used in a child project, since `doAfterVisit` can only look at a single source file.

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
jevanlingen: Ok I stand corrected, because my current implementation is not good enough! I updated the `unusedDependencyPropertiesAreRemoved` and the `<shared.version>5.3.1</shared.version>` tag is removed. 

I forgot, but the problem lies in the fact that doAfterVisit calls the vistor instead of the scanner. Thus the "does properties still exist in pom" phase is skipped!

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
jevanlingen: Yep, you are right about this. Added a test to reflect this problem: https://github.com/openrewrite/rewrite/pull/5645/commits/6aa0e648b380f9dc824eb721ef95f3883cd90075.

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
jevanlingen: Fix: https://github.com/openrewrite/rewrite/pull/5645/commits/2c9bf226ece6b24dfa64a3fe158e21fb8ffee777

rewrite-maven/src/test/java/org/openrewrite/maven/RemoveRedundantDependencyVersionsTest.java
jevanlingen: Would a

```java
getResolutionResult().getPom().getPackaging().equals("jar")
```

[check](https://github.com/openrewrite/rewrite/pull/5645/commits/ad0551aa5bb6ecea9c8d6e54b7a874e62d4af18d) be enough?

rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java
```diff
@@ -199,19 +207,32 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
...
                     }
                 }
                 return super.visitTag(tag, ctx);
             }
 
+            private @Nullable List<? extends Content> withoutVersion(Xml.Tag tag, Xml.@Nullable Tag version) {
+                return ListUtils.map(tag.getContent(), c -> {
+                    if (c == version) {
+                        if (((Xml.Tag) c).getValue().isPresent() && ((Xml.Tag) c).getValue().get().contains("${")) {
+                            final String propertyName = ((Xml.Tag) c).getValue().get().replace("${", "").replace("}", "");
```
Laurens-W: As indicated by Sam here https://github.com/openrewrite/rewrite/pull/5194#discussion_r2149906102 there could be multiple properties used

jevanlingen: Fix: https://github.com/openrewrite/rewrite/pull/5645/commits/fb7a0da5c650624585edafd602fc1ad482360d0a

PR #5646, opened 2025-06-20
Gradle: Add java toolchain if missing

sullis:
## What's changed?

This PR modifies 
```
org.openrewrite.gradle.UpdateJavaCompatibility
```

## What's your motivation?

I want to be able to add a Java toolchain to a Gradle build file


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -147,6 +211,11 @@ public K.CompilationUnit visitCompilationUnit(K.CompilationUnit cu, ExecutionCon
             if (getCursor().pollMessage(TARGET_COMPATIBILITY_FOUND) == null) {
                 c = addCompatibilityTypeToSourceFile(c, "target", ctx);
             }
+            if (Boolean.TRUE.equals(addToolchainIfMissing)) {
+                if (getCursor().pollMessage(JAVA_TOOLCHAIN_FOUND) == null) {
+                    // todo : c = addJavaToolchainToSourceFile(c, ctx);
```
sullis: This TODO comment needs to be addressed.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -135,6 +149,56 @@ private G.CompilationUnit addCompatibilityTypeToSourceFile(G.CompilationUnit c,
             }
             return c;
         }
+
+        private G.CompilationUnit addJavaToolchainToSourceFile(G.CompilationUnit c, ExecutionContext ctx) {
```
sullis: q:  is there a better way to implement "addJavaToolchainToSourceFile"

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -79,8 +81,15 @@ public class UpdateJavaCompatibility extends Recipe {
     @Nullable
     Boolean addIfMissing;
 
+    @Option(displayName = "Add Java toolchain if missing",
+            description = "Adds the specified Java toolchain if one is not found.",
+            required = false)
+    @Nullable
+    Boolean addToolchainIfMissing;
```
shanman190: Just an idea, but we could possibly use DeclarationStyle as the mechanism to describe how you as a user want the version to be defined. There's already the `addIfMissing` option, so that could work as a possibility.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
shanman190: Yeah, this is basically what we've done elsewhere. Another option could be a GroovyTemplate (or KotlinTemplate respectively), but overall it's essentially the same.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateJavaCompatibility.java
```diff
@@ -135,6 +149,56 @@ private G.CompilationUnit addCompatibilityTypeToSourceFile(G.CompilationUnit c,
...
+                    if (s instanceof J.MethodInvocation) {
+                        J.MethodInvocation invocation = (J.MethodInvocation) s;
+                        if (invocation.getName().getSimpleName().equals("java")) {
+                            newStatements.add(newJavaStatement);
+                            continue;
+                        }
+                    }
+                    newStatements.add(s);
+                }
+                c = c.withStatements(newStatements);
```
shanman190: Maybe we use `ListUtils.concat(before, toolchain)` like we do on the other side. It's a little more idiomatic that way.

PR #5647, opened 2025-06-20
Add recipe to update SCM tag based on Git origin

### LINKED ISSUE #5560: Maven - Update the `scm` configuration to match the `origin` of the repository

Following a migration to a new git server, many of my projects now have outdated SCM configuration properties.

Although it is possible to update the SCM configuration manually or with a generic xml file, doing so at scale is not practical as each repository may have different inputs 

We should have a recipe that detects the `origin` of the repository and fixes the `scm` as needed

The maven project has a `scm` configuration that does not match the server where the repository is

```shell
git remote get-url origin
https://new-server.example.com/org/repo.git
```

A pom file with the `scm` configuration set with a configuration that does not match the origin server

```xml
<scm>
        <url>https://old-server.example.com/org/repo</url>
        <connection>scm:git:https://old-server.example.com/org/repo.git</connection>
        <developerConnection>scm:git:git@old-server.example.com:org/repo.git</developerConnection>
    </scm>
```

The `scm` configuration is updated to match the origin

```xml
<scm>
        <url>https://new-server.example.com/org/repo</url>
        <connection>scm:git:https://new-server.example.com/org/repo.git</connection>
        <developerConnection>scm:git:git@new-server.example.com:org/repo.git</developerConnection>
    </scm>
```

### LINKED ISSUE #5561: Maven - Add `scm` configuration

Many tools and plugins rely on the Maven project's SCM configuration to register and publish project metadata. Common examples include:

- Maven Release Plugin (for tagging and releasing)
- Maven Site Plugin (for generating project sites with SCM links)
- Dependency management tools (for traceability and auditing)
- Sonar (project information and metadata)

If this property is absent, it can limit the visibility and discoverability of project details in important locations.

We should have a recipe that detects if a project does not have the `scm` configuration and add it

- The project does not have a `scm` configuration defined
- If there is a `scm` entry, the recipe should not make any change

```shell
git remote get-url origin
https://server.example.com/org/repo.git
```

A `pom` file with no `scm` section defined

```xml
<scm>
        <url>https://server.example.com/org/repo</url>
        <connection>scm:git:https://server.example.com/org/repo.git</connection>
        <developerConnection>scm:git:git@server.example.com:org/repo.git</developerConnection>
    </scm>
```

## Any additional context

We may be able to re-use most of the logic of https://github.com/openrewrite/rewrite/issues/5560

The difference between this recipe and #5560 is that this recipe will focus on adding it if and only if it is missing.

e5LA:
## What's changed?
- Added a new recipe `UpdateScmFromGitOrigin` that updates the `<scm>` section in `pom.xml` based on the Git origin URL.
- Added helper class `ScmValues` for URL generation logic.
- Added search recipe `FindScm` used as a precondition to check if  `<scm>` tag is present.

## What's your motivation?
- Closes #5560

As it's my first recipe, I would say:
- Use of `Visitor` +  use of `Preconditions.check(new FindScm(), ...)`
- Use of GitProvenance and initalization of `git`
- Correctness / completeness of the `ScmValues` logic for various Git origin formats

### Checklist


rewrite-maven/src/main/java/org/openrewrite/maven/utilities/ScmValues.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven.utilities;
+
+public class ScmValues{
```
e5LA, line 18: Would be reused by `AddScm` recipe.

Also, once higher version of java is supported, it could be replaced by `Record`

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/ScmValues.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven.utilities;
+
+public class ScmValues {
```
e5LA: Would be used by future `AddScm` recipe.

Also, once higher java is supported, could be replaced by `Record`

e5LA:
Hmm, I'm not sure why these tests have failed. I don't think my changes impact them. 
Are there any known issues or test flakiness that might explain this?

If someone with permissions could help re-trigger the build, that’d be appreciated.

e5LA:
Thanks for triggering the build, @timtebeek.
One of my tests was failing - must have missed it locally after some cleanup and refactoring..

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/ScmValues.java
```diff
@@ -0,0 +1,56 @@
...
+            connection = "scm:git:" + origin;
+            String sshPath = cleanOrigin
+                    .replaceFirst("^https?://", "") // github.com/user/repo
+                    .replaceFirst("/", ":");        // github.com:user/repo
+            developerConnection = "scm:git:git@" + sshPath + ".git";
+        } else {
+            url = cleanOrigin;
+            connection = "scm:git:" + origin;
+            developerConnection = "scm:git:" + origin;
+        }
```
timtebeek: I wonder a bit about the logic here specifically for `developerConnection`, as we already have GitRemote to go in the other direction that seems to differ for some `Service`s. Would it make sense to harmonize these here?

https://github.com/openrewrite/rewrite/blob/9a001ee814bb30ef036c12b4cc7c54ac2ec0fb95/rewrite-core/src/main/java/org/openrewrite/GitRemote.java#L100-L103

timtebeek: Also I feel like maybe we can keep `ScmValues` as private class of `UpdateScmfromGitOrigin` for now, as we're usually a bit hesitant to add public API surface before we've seen a few more use cases.

timtebeek (commented):
Great to see already! Shaping up nicely. I made some small changes already, and a suggestion to reduce the public API we start with; I think once addressed we could add this fairly soon after. Thanks again!

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/ScmValues.java
e5LA: Thanks for reviewing and making the changes @timtebeek !

I have just looked closely at `GitRemote` class and to be honest, I'm not sure how we could harmonize it.

Tbh, more I look at the values, more confused I get.. 
AFAIK, in SCM we need:
### `<url>` 
which usually is `http` or `https`, without `.git` at the end of the path. So e.g.: `<url>https://new-server.example.com/username/repo</url>`

We can't use `git.getOrigin()` directly (because it may contain value like: `git@new-server.example.com:username/repo.git`) so we would need to do some String operations.

Looking at `git.getGitRemote()`: 
```
{GitRemote@5224}
 service = {GitRemote$Service@5225} "Unknown"
 url = "git@new-server.example.com:username/repo.git"
 origin = "new-server.example.com"
 path = "username/repo"
 organization = "username"
 repositoryName = "repo"
```
we could build `url` from `origin` + `path` + `.git`. We would also need to derive the protocol from `url`. 
So it's kind of the same as we do in `ScmValues`, isn't it?

I also wonder if for  `origin` that has `ssh://` as protocol, shall we still use `https` or keep `ssh://` in `url`?

### `<connection>`
starts with `scm:git:` and then we could put value of `<url>` + `.git`. We can't use `url` directly.

### `<developerConnection>`
starts with `scm:git:`. Then I'm not sure if we should always go with e.g. `git@github.com:username/repo.git` even if `git.origin` contains `https` as protocol. Because in fact 'write-access' could also go via https afaik.
For `ssh://` I would assume it's: `scm:git:ssh://git@github.com/username/repo.git`

In any case, I'm not sure if `GitRemote` can help here. For reference, for `https` protocol we get:
```
result = {GitRemote@5222} 
 service = {GitRemote$Service@5223} "Unknown"
 url = "https://new-server.example.com/username/repo.git"
 origin = "new-server.example.com"
 path = "username/repo"
 organization = "username"
 repositoryName = "repo"
```

--
Did I miss anything? Are these SCM values `Service` -specific in any way? Could we easily derive them from `GitRemote`?
Another idea would be to extend `GitRemote` with the logic to derive the values, but not sure if scm-specific logic really belongs there.

Regarding making `ScmValues` class private - happy to do so if we decide to keep it. 
Having that said, I wonder how we’ll reuse this logic when creating a new Recipe to add the <scm> tag (https://github.com/openrewrite/rewrite/issues/5561). (But maybe that’s a tomorrow's problem :) )

e5LA: Hi @timtebeek, I've marked `ScmValues` as private for now.
Is there anything else I can do before we move this forward?

timtebeek:
Great to see! I've pushed up some light polish, and I'm not spot checking the results against Apache projects. I'm finding a few patterns that I wonder if we should support, such as for instance
image

Notice how
- we're stripping off the `.git` suffix on `<connection>`; should we retain the `.git` suffix there if present previously ?
- switching the `<developerConnection>` from `https` to `git@`; should we allow `<developerConnection>scm:git:https://...` ?
- correctly adding the `incubator-` prefix that was missing before.

image

Ideally I think we'd see minimal changes unless there's a host/path that needs to be corrected; that way we can run this recipe broadly and only have to review proper fixes as opposed to a broader set of potentially more contentious changes.

timtebeek:
If you'd like to similarly see the effects at scale, then you can follow these steps:

1. [Download the Moderne CLI](https://docs.moderne.io/user-documentation/moderne-cli/getting-started/cli-intro#installation-and-configuration)
2. [Set your recipe as the active recipe](https://docs.moderne.io/user-documentation/moderne-ide-integration/how-to-guides/debugging-recipes#step-1-set-the-active-recipe) (requires a license :( )
3. `mod git clone moderne apache/ Apache --metadata-only`
4. `mod build apache/ --download-only`
5. `mod run apache/ --active-recipe`

You can also achieve step 5 without step 2 by using `./gradlew pTML` followed by `mod config recipes jar install org.openrewrite:rewrite-maven:LATEST`.

Hope that helps!

e5LA:
Thanks @timtebeek for the feedback and instructions how to run the Recipe on apache repos. Really helpful.

After seeing the changes you got in the apache repo,  I agree with you - we should indeed try to keep the changes minimal..

So, another approach I’m considering, would be: instead of generating the SCM `url`, `connection` and `developerConnect` values entirely from `git.origin` and then replacing the full value in the tags - I would rather extract just the host and path from `git.origin` and use that to replace only the corresponding part of the existing SCM values. This way, we keep the original protocol and `.git` suffix.

With this approach, we may not be able to update the protocol.. But it’s likely a safer option overall.

Would you agree with this approach, or do you see any other potential issues?

timtebeek:
That's indeed likely a safer option. One thing to keep an eye on then still is that some git hosts use a context-path like `/scm` which for instance `github.com` does not have. As always there's more edge cases the closer you look, and scale helps bring those to light. :) Appreciate your help here!

e5LA:
I've updated the logic, now it looks a bit more complex..

I've also run it against few apache projects:
image

--

image

--

image

timtebeek:
Thanks a lot! I'll be camping over the weekend & back Wednesday, so review from my side might take a while longer, but know it's appreciated and I hope to get it in soon after I'm back.

timtebeek (approved):
Great to see! Thanks a lot for a neat addition for the Apache Maven best practices.

PR #5652, opened 2025-06-21
New `VariableScope` trait

knutwannheden:


jkschneider:
I think much of this logic is in the forthcoming rewrite-program-analysis module, plus more sophistication. I would expect a trait to be built on top of that.

PR #5658, opened 2025-06-23
Uniform behaviour for the methodInvocations that are needed for template unsubstitution

Jenson3210:
## What's changed?
Added an overload to `Substitutions` `maybeParameter` that accepts the prefixed element. 
When using a template that passes in a placeholder that must be replaced by an Object / parameter, the resulting template is shows as `__P__.<java.lang.Object>/*__p0__*/p()`

The existing code will pass methodName's `Identifier` (`/*__p0__*/p()`) to the `maybeParameter` call which will cause it to use the prefix of this Identifier to parse the template resulting in trimmed template prefix. 

By overloading and passing the entire methodInvocation for prefix calculations we no longer need to [overcome this in every recipe](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java#L74-L80) that is using templates but we can just rest assured that the prefix of the template will be also present in the result lst objects. 

## What's your motivation?
Stumbled into this investigation with @Laurens-W  after having some other template issues. 

### Checklist


Jenson3210:
Do not merge this one without approval from @jkschneider

rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest.java
```diff
@@ -1524,4 +1526,50 @@ abstract static class Two<TwoT extends Two<TwoT, OneT>, OneT extends One<TwoT, O
           )
         );
     }
+
+    @ParameterizedTest(name = "Keep template prefix for block argument when template is \"{0}\"")
+    @CsvSource({"#{}", "#{any()}"})
```
timtebeek, line 1530: Earlier today @knutwannheden mentioned:
> The `#{any()}` syntax is indeed only applicable to expressions. We would have to come up with something like `#{type()}` if we need to be able to reference types using template parameters

The `if_.getThenPart()` `J.Block` we're passing in here is a `Statement`, not an `Expression` with it's own type.

Without any changes in `rewrite-java/src/main/java/org/openrewrite/java/internal/template/Substitutions.java` the test using `#{}` already passes, whereas the variant with `#{any()}` seems unsupported. I wonder if this is the right change to be making then, as it seems we're fighting a symptom of something not (yet?) supported, and instead perhaps should [make changes downstream](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java#L74-L80).

Any quick thoughts here @Jenson3210 / Knut?

Jenson3210: I do not think this one is related to these statements per se. 
In a template parameter `__P__.p()` methodInvocation, the prefix of the Identifier `p` will always be an empty string. If you put the placeholder in a template `String s = #{any()};`, the prefix of the method invocation is a blank and the rendered result should have that blank (and not the no-space from `p()`) as that is requested in the template.

Jenson3210: You are right. Tried to validate the change using some other expressions and getting other weird behaviors. Better not to touch it for now!

timtebeek: Thanks! Would it make sense to make the same adjustment for the switch template downstream then? Just so we know?

Jenson3210: I am not sure what you mean. There the template is working fine, and we just corrected the prefix manually if I recall correctly. Not having the template over there would result back in printing all the lst elements to Strings as it was before?

timtebeek: I thought something like this might then help remove the awkward prefix handling, but it's not working yet:
image

Issue #4816, opened 2024-12-25
UpdateGradleWrapper fails for a custom wrapperUri with placeholders


bmuschko:
## What version of OpenRewrite are you using?

- OpenRewrite v8.41.0

## How are you running OpenRewrite?

Moderne.io

Is your project a single module or a multi-module project? Single

Can you share your configuration so that we can rule out any configuration issues?

```
type: specs.openrewrite.org/v1beta/recipe
...
recipeList:
  - org.openrewrite.gradle.UpdateGradleWrapper:
       version: 8.9
       distribution: bin
       wrapperUri: https://services.gradle.org/distributions/gradle-${version}-${distribution}.zip
```

## What is the smallest, simplest way to reproduce the problem?

Run the [org.openrewrite.gradle.UpdateGradleWrapper](https://docs.openrewrite.org/recipes/gradle/updategradlewrapper) recipe with the [declarative YAML format](https://docs.openrewrite.org/reference/yaml-format-reference).

## What did you see instead?

An error with a stack trace when executing the recipe.

## What is the full stack trace of any errors you encountered?

```
java.lang.IllegalArgumentException: Illegal character in path at index 50: https://services.gradle.org/distributions/gradle-${version}-${distribution}.zip
  java.base/java.net.URI.create(URI.java:906)
  org.openrewrite.gradle.UpdateGradleWrapper.getGradleWrapper(UpdateGradleWrapper.java:139)
  ...
```

## Are you interested in [contributing a fix to OpenRewrite]

Yes

shanman190:
Welcome @bmuschko!

I think this issue may be one of the documentation being out of sync with expectations, but open to feedback as well.

During the work on https://github.com/openrewrite/rewrite/pull/4491, I had mentioned there that interpolating the exact version and distribution into the wrapperUri felt kinda weird given the only way to define the wrapperUri is directly and in the most general case, it's being used for corporate environments to swap to a privately built Gradle wrapper. So in that PR, the parameter replacements were removed. It seems like the `@Option` values though didn't get updated which then transitively impact the documentation and Moderne SaaS.

I think for this, the best course may be to update the documentation to remove the no longer valid information. What do you think?

bmuschko:
Thanks for clarifying!

There's another use case for this I'd like to mention. A company hosts a Gradle Wrapper binary on an internal repository so that consumers don't need to reach out to the public Wrapper service endpoint (as it would be blocked due to security measures).

[...]

PR #5659, opened 2025-06-23
Fixed UpdateGradleWrapper description to represent behaviour

Jenson3210:
## What's changed?
Just a description change

## What's your motivation?
- Fix: https://github.com/openrewrite/rewrite/issues/4816

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
@@ -97,9 +97,9 @@ public class UpdateGradleWrapper extends ScanningRecipe<UpdateGradleWrapper.Grad
     @Option(example = "https://services.gradle.org/distributions/gradle-8.5-bin.zip",
             displayName = "Wrapper URI",
             description = "The URI of the Gradle wrapper distribution. " +
-                          "When specified, version and distribution CANNOT be specified, and it is assumed that this uri is a reachable/valid wrapper distribution." +
-                          "This is particularly useful in environments where the gradle services url is not reachable. " +
-                          "The recipe will fetch gradle wrapper shell/bat files from the zip location and if the uri specified is not valid, the repo will keep its current files in place as they are mostly compatible.",
+                    "When specified, version and distribution CANNOT be specified, and it is assumed that this uri is a reachable/valid wrapper distribution." +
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
```diff
@@ -94,13 +94,12 @@ public String getDescription() {
...
             displayName = "Wrapper URI",
             description = "The URI of the Gradle wrapper distribution. " +
-                          "Lookup of available versions still requires access to https://services.gradle.org " +
-                          "When this is specified the exact literal values supplied for `version` and `distribution` " +
-                          "will be interpolated into this string wherever `${version}` and `${distribution}` appear respectively. " +
-                          "Defaults to https://services.gradle.org/distributions/gradle-${version}-${distribution}.zip.",
+                    "When specified, version and distribution CANNOT be specified and it is assumed that this uri is a reachable/valid wrapper distribution path." +
+                    "This is particularly useful in environments where the gradle services url is not reachable or where there are custom gradle wrappers in place. " +
+                    "The recipe will fetch gradle wrapper shell/bat files from the location and if the uri specified is not accessible, the repo will keep its current files in place as they are mostly compatible.",
```
jevanlingen: To be honest, I find this documentation still a little vague. I let AI create the following text:
> The URI of the Gradle wrapper distribution.
Specifies a custom location from which to download the Gradle wrapper scripts (gradlew, gradlew.bat, etc.). This is useful for setting up the Gradle wrapper without relying on Gradle's official distribution services.
> 
> When this option is set, the version and distribution fields must not be specified — only one source of truth is allowed. The URI should point to a valid and reachable Gradle wrapper distribution (typically a .zip archive containing the wrapper files).
> 
> This is particularly helpful in environments where access to Gradle's central services is restricted or where custom Gradle wrapper setups are required.
> 
> If the URI is inaccessible, the recipe will leave the existing wrapper files in the repository unchanged, as they are generally compatible with various Gradle versions.

Maybe change it a little bit more, but try to think from a user perspective. Ask yourself: is the text understandable if you read it without prior knowledge?

Jenson3210: stole your text 😄 
I personally think it's saying exactly the same, so no feelings for my own text over this one.

jevanlingen (approved):

PR #5663, opened 2025-06-24
Fix `FindDependency` to exclude constraints

### LINKED ISSUE #5599: `gradle.FindDependency` also finds Gradle Dependency Constraints
I am using current main.

Via the unit tests, https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindDependencyTest.java

With `FindDependency` to me configured to find `com.fasterxml.jackson.core:jackson-databind`

```groovy
plugins { id 'java' }

dependencies {
    constraints {
        implementation('com.fasterxml.jackson.core:jackson-databind:2.12.7.1')
    }
}
```

No Search Result Marker

## What did you see instead?

```groovy
plugins { id 'java' }

dependencies {
    constraints {
        /*~~>*/implementation('com.fasterxml.jackson.core:jackson-databind:2.12.7.1')
    }
}
```

knutwannheden:
## Changes
- ~Modified `FindDependency.java` to skip matching dependencies inside `constraints { }` blocks~
- ~Added `isWithinConstraintsBlock()` method to traverse AST and detect constraint context~
- Added comprehensive tests verifying constraints are ignored while regular dependencies are still found

- Switch implementation to use GradleDependency trait, for consistent match and to remove duplicated logic.

- Fixes #5599


timtebeek (approved):

shanman190:
@knutwannheden , any reason to not refactor this to the `GradleDepencency` trait?

knutwannheden:
That would probably make sense.

timtebeek:
Figured give that a quick shot, as it does clean up the code nicely. But now we seem to fail on .kts. :(

shanman190:
@timtebeek , I think I see what the issue is there. With these test cases, they don't add the `GradleProject` marker which in this case is fine, but it triggers the method matching logic found [here](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2Ftrait%2FGradleDependency.java#L93) which does not match unknown types for kotlin-dsl. That simple fix should restore the test to functioning correctly.

Also note, I believe that this test case is an invalid Gradle Kotlin DSL script, but only marginally so. When applying plugins as a literal I believe it has to be a back ticked string versus a quoted string. I haven't tested this, but I've also never seen a script with a quoted string plugin application like is described here.

timtebeek:
Thanks for the help here @shanman190 ; your suggestions as usual where spot on. With that I think this is good to merge: we still pass all the tests we did before, we pass additional tests, and we now use the `GradleDependency` trait that has also separately passed review and is used in other core recipes to change/upgrade/remove Gradle dependencies, making this nicely consistent.

timtebeek:
Previously we were also able to find some older Gradle version dependencies; should the Gradle Matcher trait be updated to allow for those as well?

shanman190:
@timtebeek , the trait prefers using the Gradle model itself for this exact reason. Now the reason why the test fails when attempting to check for older configurations is because those don't exist within the Gradle version that we use in the tooling API by default and for this reason is exactly why the tooling API helped methods expose a version property to the test author.

https://github.com/openrewrite/rewrite-gradle-tooling-model/blob/08f89b0f74ae1140b4aa54dc4576b099eda973d6/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModelBuilder.java#L83

https://github.com/openrewrite/rewrite-gradle-tooling-model/blob/08f89b0f74ae1140b4aa54dc4576b099eda973d6/model/src/main/java/org/openrewrite/gradle/toolingapi/Assertions.java#L201

Gradle 7.0 removed these old configurations officially, so since when we don't specify the version we will use Gradle 8.12, this is why those test examples fail under the current conditions.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
```diff
@@ -69,9 +71,102 @@ void literal() {
...
+      "testImplementation",
+      "testCompileOnly",
+      "testRuntimeOnly",
+      "debugImplementation",
+      "releaseImplementation",
+      "androidTestImplementation",
+      "featureImplementation",
+      "annotationProcessor",
+      "kapt",
+      "ksp"
```
shanman190, line 130: These are not Gradle native configurations and instead come from a custom configuration that come from a plugin not defined within this test case which explains why they fail.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
```diff
@@ -69,9 +71,102 @@ void literal() {
...
+      "implementation",
+      "compileOnly",
+      "runtimeOnly",
+      "testImplementation",
+      "testCompileOnly",
+      "testRuntimeOnly",
+      "debugImplementation",
+      "releaseImplementation",
+      "androidTestImplementation",
+      "featureImplementation",
```
shanman190, line 126: And these all come from the Android plugin.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
```diff
@@ -69,9 +71,102 @@ void literal() {
...
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+      "compile", // deprecated
+      "runtime", // deprecated
+      "testCompile", // deprecated
+      "testRuntime" // deprecated
+    })
+    void decprecatedMethods(String method) {
```
shanman190, line 174: This is presently failing because Gradle 6.9.4 supports Java 15 at the time of its release... We may need to think about how to deal with these sorts of things...

timtebeek:
Thanks both! Merging this step as we look to expand on the work done here in a subsequent PR, with the below still applicable

> With that I think this is good to merge: we still pass the tests we did before, we pass additional tests, and we now use the `GradleDependency` trait that has also separately passed review and is used in other core recipes to change/upgrade/remove Gradle dependencies, making this nicely consistent.

PR #5674, opened 2025-06-25
Determine the correct configuration for Gradle dependency constraints

bmuschko:
## What's changed?

The logic that determine which Gradle configuration to use based on the originating configuration for which we find the transitive dependency to bump up the version for.

## What's your motivation?

The preexisting logic tried to approximate the correct configuration to use, however, the logic wasn't really correct for multiple reasons:

1. Existing dependency constraints were not updated the correctly, as the incorrect configuration would have been chosen resulting in a duplicate definition of the constraint.
2. The configuration mapping would sometimes pick an inappropriate configuration. For example, say the declared dependency uses `compileOnly` (only for compiling the code but not for running the code) would be mapped to the constraint configuration `implementation` (for compiling and running the code). This declaration would result in adding the constraint to runtime dependencies which is not the intention of `compileOnly`. See the [configuration visualizations](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management) in the Java plugin for reference.
3. Legacy configurations like `compile` used by projects running Gradle < 6.x would map to the constraint configuration `implementation` which doesn't even exist in that version of Gradle resulting in a build failure. You could argue that the Gradle version should be upgraded first and switch to `implementation` in the declared dependency, however, the recipe should still result in the correct mapping.

## Example for incorrect behavior

Given the following build and the usage of [legacy Gradle configurations](https://docs.gradle.org/current/userguide/upgrading_version_6.html#sec:configuration_removal) like `compile` (which requires Gradle < 7).

```
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

dependencies {
    constraints {
        compile 'com.thoughtworks.xstream:xstream:1.4.17'
    }
}

dependencies {
    compile 'org.springframework.cloud:spring-cloud-starter-netflix-eureka-client:4.2.0'
}
```

Running the following recipe.

```
$ mod run . --recipe UpgradeTransitiveDependencyVersion --recipe-option "groupId=com.thoughtworks.xstream" --recipe-option "artifactId=xstream" --recipe-option "version=1.4.21"
```

The recipe [UpgradeTransitiveDependencyVersion](https://docs.openrewrite.org/recipes/gradle/upgradetransitivedependencyversion) will add a duplicate dependency constraint with the `implementation` configuration instead of using `compile`. Before adding a new one, the recipe should identify an existing declarations and update that one.

```
diff --git a/build.gradle b/build.gradle
index 68fd396..67019c8 100644
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ org.openrewrite.gradle.UpgradeTransitiveDependencyVersion
 
 dependencies {
     constraints {
+        implementation('com.thoughtworks.xstream:xstream:1.4.21')
         compile 'com.thoughtworks.xstream:xstream:1.4.17'
     }
 }
```

## Any additional context

The expectations of some of the existing test cases were incorrect. For example, requesting that only the configuration `implementation` dependencies should be inspected also caused `compileOnly` dependencies to add a constraint, however, `compileOnly` does not extend from `implementation` and vice versa.


bmuschko:
@timtebeek @sambsnyd Any chance you review this one? The requestor of the fix is already asking for it. Thanks!

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
```diff
@@ -485,7 +568,7 @@ void addConstraintToNonTransitiveExtendingTransitiveConfiguration() {
                   deploy 'org.openrewrite:rewrite-java:7.0.0'
               
                   constraints {
-                      earlib('com.fasterxml.jackson.core:jackson-core:2.12.5') {
+                      deploy('com.fasterxml.jackson.core:jackson-core:2.12.5') {
```
timtebeek: This seems to change what was explicitly added in
 - https://github.com/openrewrite/rewrite/pull/4107

Note the explicit `configurations.earlib.extendsFrom configurations.deploy`.
I still think it makes sense to use deploy since that was used for the dependency, but perhaps we should use `earlib` for the dependency instead? The test still passes in that case and it's perhaps closer to what was intended here.

bmuschko: Generally speaking, declaring the line `configurations.earlib.extendsFrom configurations.deploy` doesn't make sense. The [two configurations](https://docs.gradle.org/current/userguide/ear_plugin.html#sec:ear_dependency_management) provided by the Ear plugin are treated differently. The point is to have two separated configurations, for two different use cases.

> The Ear plugin adds two dependency configurations: `deploy` and `earlib`. All dependencies in the `deploy` configuration are placed in the root of the EAR archive, and are not transitive. All dependencies in the `earlib` configuration are placed in the 'lib' directory in the EAR archive and are transitive.

When defining a constraint, it should use the originating configuration and not some child configuration.

Maybe @Jenson3210 can comment on the intention of this test case, as he added it.

PR #5676, opened 2025-06-26
Fixing code suggestions

greg-at-moderne:
## What's changed?

Applying code suggestions to one class.

## What's your motivation?
Avoid noise in unrelated PRs.

```
Some suggestions could not be made:

rewrite-core/src/main/java/org/openrewrite/internal/ManagedThreadLocal.java
lines 115-115
lines 185-185
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ManagedThreadLocal.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ManagedThreadLocal.java
@@ -113,7 +113,7 @@ public class ManagedThreadLocal<T extends AutoCloseable> {
         AtomicReference<@Nullable T> resourceRef = new AtomicReference<>();
 
         return new Scope<>(
-                () -> {
+                (Supplier<T>) () -> {
                     T resource = resourceRef.get();
                     if (resource == null) {
                         resource = factory.get();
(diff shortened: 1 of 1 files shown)


greg-at-moderne:
Interesting. The compilation fails with:
```
rewrite/rewrite-core/src/main/java/org/openrewrite/internal/ManagedThreadLocal.java:115: error: reference to Scope is ambiguous
        return new Scope<>(
  both constructor <T>Scope(T,AutoCloseable) in Scope and constructor <T>Scope(Supplier<T>,AutoCloseable) in Scope match
  where T is a type-variable:
    T extends AutoCloseable declared in class Scope
```

knutwannheden:
This is a bug in that `UseDiamondOperator` recipe.

greg-at-moderne:
Yeah. Probably https://github.com/openrewrite/rewrite-static-analysis/issues/214 or similar.

greg-at-moderne:
An extra cast as in 355c78b8b36450e272dde29b310b0e82e54dcca6 should be a good enough workaround until the bug is fixed.

greg-at-moderne:
> This is a bug in that UseDiamondOperator recipe.

Reported https://github.com/openrewrite/rewrite-static-analysis/issues/603

PR #5678, opened 2025-06-26
PreferSatisfiesKeyword

greg-at-moderne:
### Checklist


greg-at-moderne:
I am giving up on this.
Mostly because it's a hard problem to detect what should be the declared type (`$x`) based on the `kind: ...` declarations. There some heuristics, but they might different between projects. And also even in our codebase the conventions are not applied consistently.

rewrite-javascript/rewrite/src/recipe/prefer-satisfies-keyword.ts
```diff
@@ -0,0 +1,78 @@
...
+import {JavaScriptVisitor, JS} from "../javascript";
+import {emptySpace, J, space} from "../java";
+import {randomId} from "../uuid";
+import {emptyMarkers} from "../markers";
+import {produce} from "immer";
+
+export class PreferSatisfiesKeyword extends Recipe {
+    name = "org.openrewrite.PreferSatisfiesKeyword";
+    displayName = "Use `satisfies` keyword instead of `as`";
+    description = "Use `satisfies` keyword instead of `as` when constructing a new object.";
```
greg-at-moderne, line 29: This is wrong. `satisfies` is not a better version of `as`. It may complement `as` type attribution to provide type safety.

The code below has already evolved over that idea. But since I am closing this PR I don't spend time making the text consistent.

PR #5682, opened 2025-06-27
Adjusting the `isValid` check for `XRange` so that it doesn't allow for going from release to pre-release versions when upgrading.

steve-aom-elliott:
## What's changed?
- Changed the `isValid` check for `XRange` back to calling the `super`'s version of the method first, so if you're on `3.5.0` and trying to upgrade to `3.6.x`, if there are `3.6.1-SNAPSHOT` and `3.6.0` available, it will take `3.6.0`.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
@@ -46,7 +46,7 @@ public class XRange extends LatestRelease {
     public boolean isValid(@Nullable String currentVersion, String version) {
         // For X-range patterns, we need to check if the version matches the pattern
         // but we should not exclude it just because it's a pre-release version
-        if (!VersionComparator.checkVersion(version, getMetadataPattern(), false)) {
+        if (!VersionComparator.checkVersion(version, getMetadataPattern(), getMetadataPattern() == null)) {
             return false;
         }
 
--- a/rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
@@ -44,9 +44,7 @@ public class XRange extends LatestRelease {
     @SuppressWarnings("ResultOfMethodCallIgnored")
     @Override
     public boolean isValid(@Nullable String currentVersion, String version) {
-        // For X-range patterns, we need to check if the version matches the pattern
-        // but we should not exclude it just because it's a pre-release version
-        if (!VersionComparator.checkVersion(version, getMetadataPattern(), getMetadataPattern() == null)) {
+        if (!super.isValid(currentVersion, version)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


greg-at-moderne:
Updating your PR for unrelated Junit 5/6 problems.

rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
```diff
@@ -46,7 +46,7 @@ public class XRange extends LatestRelease {
     public boolean isValid(@Nullable String currentVersion, String version) {
         // For X-range patterns, we need to check if the version matches the pattern
         // but we should not exclude it just because it's a pre-release version
-        if (!VersionComparator.checkVersion(version, getMetadataPattern(), false)) {
+        if (!VersionComparator.checkVersion(version, getMetadataPattern(), getMetadataPattern() == null)) {
```
knutwannheden: I think it would make sense to add a test case which demonstrates that a metadata pattern like `-beta.*` could be used to allow upgrades to a beta (non-release) version.

knutwannheden (approved):
Looks good to me. I added a comment regarding an additional test case.

rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
steve-aom-elliott: So if you provide the `metadataPattern`, it's going to ensure any possible upgraded version also has that `metadataPattern`, even if there are higher release versions. Is that what we want to show?

i.e. You're at `3.4.0` and you tell it to upgrade to `3.5.x` with a metadata pattern of `-beta`. If our available versions for `3.5.x` are `3.5.2` and `3.5.1-beta`, it's going to take `3.5.1-beta` over `3.5.2` because of the `metadataPattern`.

knutwannheden: Good question. Would it then pick an older beta if the latest release doesn't have a beta? That wouldn't be what we want. Then we should instead just pass in `true` here.

```suggestion
        if (!VersionComparator.checkVersion(version, getMetadataPattern(), true)) {
```

steve-aom-elliott: @knutwannheden Hmm, so we go back to now allowing it to upgrade to a non-release version at all, yes? I would agree with that, as from what I've seen, the intended usage of the `metadataPattern` is really more for things like `-jre` or `-zulu` suffixes on Java versions and the like anyway.

steve-aom-elliott: Changed this in the most recent push now.

knutwannheden (approved):

rewrite-core/src/main/java/org/openrewrite/semver/XRange.java
knutwannheden: This is an area where Maven versioning clashes a bit with semver, unfortunately.

PR #5690, opened 2025-06-30
Fix AddOrUpdateAnnotation recipe changing implicit `value` into explicit `value` in annotations

JohannisK:
## What's changed?
When updating an existing annotation property the recipe should not change whether `value` is implicit or explicit.

## What's your motivation?
The recipe makes unnecessary changes which is noisy.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
@@ -1481,4 +1481,75 @@ class AddOrUpdateAnnotationAttributeTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void doNotAddOrRemoveValueWhenOldAtrributeValueDoesNotMatchAndAttributeNameIsNotValue() {
+        rewriteRun(
+          spec -> spec.recipe(
+            new AddOrUpdateAnnotationAttribute(
+              "org.example.Foo",
+              "name",
+              "newValue",
+              "oldValue",
+              null,
+              null)),
+          java(
+            """
+              package org.example;
+              public @interface Foo {
+                  String name() default "";
+                  String value() default "";
+              }
+              """,
+            SourceSpec::skip
+          ),
+          java(
+            """
+              import org.example.Foo;
+              
+              @Foo("oldValue")
+              public class A {}
+              
+              @Foo(value = "oldExplicitValue")
+              public class B {}
+              """
+          )
+        );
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -250,7 +250,8 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
                                     return it;
                                 }
                                 return ((J.Literal) it).withValue(newAttributeValue).withValueSource(newAttributeValue);
-                            } else {
+                            } else if (oldAttributeValue == null) {
```
Laurens-W, line 253: The test cases you added only seem to be situations where `oldAttributeValue` is not null, and you haven't adjusted any existing test expectations.
Which makes me wonder if we are actually capturing the behavior that's changed here

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
```diff
@@ -1481,4 +1481,75 @@ public class A {}
...
+
+    @Test
+    void doNotAddOrRemoveValueWhenOldAtrributeValueDoesMatchAndAttributeNameIsValue() {
+        rewriteRun(
+          spec -> spec.recipe(
+            new AddOrUpdateAnnotationAttribute(
+              "org.example.Foo",
+              "value",
+              "newValue",
+              "oldValue",
```
Laurens-W, line 1529: The name of the testcase implies this should be matched, but I don't think that's the case

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
JohannisK: I've added a test to show the change from implicit to explicit `value` is still done when no oldAttributeValue is present.
The original issue was that when an oldAttributeValue was supplied, it would still modify implicit value to explicit value, even though nothing effectively changed.

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
JohannisK: It's not about the value, it's really about the explicit parameter name `value`.

I understand it's a very confusing thing, but I'm not sure how to name this better...

PR #5691, opened 2025-06-30
Create production-package.zip earlier in the process

shanman190:
## What's changed?
Create the production-package.zip earlier in the process and make it available on the classpath for usage.

## What's your motivation?
Improvements upon https://github.com/openrewrite/rewrite/commit/49f19f5b519cb311991d379070d84265f466753a.

N/A

@knutwannheden 

N/A

## Any additional context
N/A

### Checklist

### DIFF
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -157,6 +157,7 @@ val npmInstallTemp = tasks.register<NpmTask>("npmInstallTemp") {
     val tempInstallDir = layout.buildDirectory.file("tmp/npmInstall")
     inputs.files(npmPack)
         .withPathSensitivity(PathSensitivity.RELATIVE)
+    outputs.dir(tempInstallDir)
     dependsOn(npmInitTemp)
 
     // Use a provider to defer evaluation until execution time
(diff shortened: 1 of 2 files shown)


rewrite-javascript/build.gradle.kts
```diff
@@ -157,54 +157,37 @@ val npmInstallTemp = tasks.register<NpmTask>("npmInstallTemp") {
...
 
     // Use a provider to defer evaluation until execution time
     args = provider { listOf("install", npmPack.get().archiveFile.get().asFile.absolutePath, "--omit=dev") }
     workingDir.set(tempInstallDir)
 }
 
-sourceSets {
-    main {
-        resources {
-            srcDir("src/main/generated-resources")
```
shanman190, line 170: callout: Should always avoid writing to the project directory as it can result in triggering tasks to run unnecessarily.

rewrite-javascript/build.gradle.kts
```diff
@@ -157,54 +157,37 @@ val npmInstallTemp = tasks.register<NpmTask>("npmInstallTemp") {
...
-        }
-    }
-}
-
-afterEvaluate {
```
shanman190, line 175: callout: Avoid `afterEvaluate`

rewrite-javascript/rewrite/tsconfig.build.json
```diff
@@ -8,7 +8,6 @@
     "removeComments": true
   },
   "include": [
-    "src/**/*",
-    "test/**/*"
+    "src/**/*"
```
shanman190, line 11: callout: the `tsc` compiler isn't very smart, so what this was resulting in is the `test/**` sources being built twice which isn't desirable.

shanman190 (commented):
I'm double checking the licensing and source jar tasks yet, so that'll be just a moment for an update on those.

rewrite-javascript/build.gradle.kts
```diff
@@ -157,54 +157,42 @@ val npmInstallTemp = tasks.register<NpmTask>("npmInstallTemp") {
...
+}
+
+tasks.named<Jar>("sourcesJar") {
     from(createProductionPackage)
+    exclude("production-package.zip")
+}
+
+tasks.named<ProcessResources>("processResources") {
+    dependsOn(createProductionPackage)
 }
```
shanman190: callout: All of this is really a hack to make the IDE work correctly. The `production-package.zip` is a generated resource that would normally be included into the binary jar itself. Unfortunately, since IDEA doesn't generate the jar, but rather uses the raw classes and resources, we're stuck with forcing the `production-package.zip` to be generated earlier in the process than would typically be necessary, including it as a main resource for IDE support, and then having to filter it out of other tasks for which we don't want the generated binary to be included for.

Overall, not ideal by any means.

shanman190 (commented):
Licensing no longer required any changes because the resource is coming from the `build` directory rather than the project directory.

Source jar still needed it's dependency and exclusion (see further inline comments).

rewrite-javascript/rewrite/tsconfig.build.json
knutwannheden: We'd still need some more refactoring to get rid of this. The problem is that some integration tests fail without this. The tests aren't executed on CI, so I can also fix this afterwards.

rewrite-javascript/build.gradle.kts
knutwannheden: I think we can probably get rid of the hack and just require any IDE tests to specify the path to the dist dir of the project when setting up RPC. I think that is acceptable.

rewrite-javascript/rewrite/tsconfig.build.json
shanman190: Just a clarification, but as in the Java integration tests?

knutwannheden: Yes. They depend on a JS recipe which is in test. I will try to fix that.

shanman190: Alright, so how would you feel about adding an NPM script to compile tests separate from the `test` and `ci:test` phases. Then we can wire a dependency from `integrationTest` to this new task. Thoughts?

shanman190: So we can take this thought a step further by using the jvm-test-suite plugin instead of the Nebula IntegTest plugin. With the jvm-test-suite plugin, we would be declaring a dependency upon the production outgoing artifacts which would result in the jar being built and more closely aligning the integration tests with what the real runtime environment would look like.

shanman190:
I had a couple of integration test failures on Windows, but pushed the changes here as it may be OS specific.

shanman190:
@knutwannheden , is there anything else that we need to do for this one?

shanman190:
@knutwannheden , following back up on this one. I think it's safe to merge though.

knutwannheden (approved):
I had lost track of this one... I ran some simple tests and from what I can tell it looks good!

PR #5693, opened 2025-07-01
Fix AddOrUpdateAnnotation recipe changing implicit `value` into explicit `value` in annotations with arrays

jevanlingen:
## What's changed?
When updating an existing annotation property, the recipe should not change whether value is implicit or explicit.

## What's your motivation?
The recipe makes unnecessary changes which is noisy.

### Checklist


rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
```diff
@@ -536,6 +536,43 @@ public class A {
...
+              import org.example.Foo;
+              
+              @Foo({"oldTest"})
+              public class A {
+              }
+              """,
+            """
+              import org.example.Foo;
+              
+              @Foo({"newTest"})
```
JohannisK, line 568: Can you add a test showing it will also handle `@Foo(value = {"oldTest"})` to `@Foo(value = {"newTest"})` correctly?

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
```diff
@@ -573,6 +610,43 @@ public class A {
...
+              public @interface Foo {
+                  String[] value() default {};
+              }
+              """
+          ),
+          java(
+            """
+              import org.example.Foo;
+              
+              @Foo({"oldTest"})
```
JohannisK, line 746: I'm missing the cases for `@Foo("oldTest")` to `@Foo({"newTest1", "newTest2"})`

jevanlingen: Done!

rewrite-java-test/src/test/java/org/openrewrite/java/AddOrUpdateAnnotationAttributeTest.java
jevanlingen: Done!

PR #5698, opened 2025-07-02
Untangle the `addOnly` and `appendArray` options for AddOrUpdateAnnotationAttribute recipe

jevanlingen:
## What's changed?
- Restore the intention of the `addOnly` property. It's meant to enable a "AddAnnotationAttribute recipe" mode; this mode guarantees existing annotation parameters are not changed. _It should not be mingled with another option, which makes it hard to understand, thus hard to use well._
- With the `appendArray` option:
  -  **off or disabled**: the recipe will replace the current initializer expressions with the new expressions. This is in line with the non-array annotation attributes, which also will totally replace the old value with the new attributeValue (albeit a single value). 
  - **enabled**: the recipe will not replace the current initializer expressions with the new expressions, but will extend the list of expressions with the given attributeValue(s).
- Array annotation attributes will filter duplicates automatically.

## What's your motivation?
The main idea of this PR is to make the recipe more maintainable. The current code quality was lacking a bit, so I improved it. When I was working on it, I stumbled about the weirdness of the `addOnly` and `appendArray` options. With the introduction of array support, the options got intertwined; even from a user perspective it's not really clear how they options should behave. So when refactoring, I changed the scope a bit to make these options usable (and maintainable).

Check if the options feel logical now when you read the description. I tried to improve the code to make it more easy to maintain, but be critical here. If it still needs more improvement, let me know!

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
@@ -84,7 +84,7 @@ public class AddOrUpdateAnnotationAttribute extends Recipe {
 
     @Option(displayName = "Append array",
             description = "If the attribute is an array and attribute is present, setting this option to `true` will append the value(s). Duplicate values will not be added. " +
-                    "Otherwise, the recipe will replace the existing value(s) with the new value(s).",
+                    "If omitted or `false`, the recipe will replace the existing value(s) with the new value(s).",
             required = false)
     @Nullable
     Boolean appendArray;
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
@@ -17,28 +17,23 @@ package org.openrewrite.java;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import lombok.With;
 import org.jetbrains.annotations.Contract;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.search.UsesType;
-import org.openrewrite.java.tree.Expression;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JavaType;
-import org.openrewrite.java.tree.TypeUtils;
-import org.openrewrite.marker.Marker;
-import org.openrewrite.marker.Markers;
+import org.openrewrite.java.tree.*;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
+import static java.lang.Boolean.TRUE;
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.marker.Markers.EMPTY;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -115,117 +110,51 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
-                                    if (Boolean.TRUE.equals(addOnly)) {
-                                        m = 0;
-                                    }
-                                    for (int j = m; j < attributeList.size(); j++) {
-                                        String newAttributeListValue = maybeQuoteStringArgument(attributeName, attributeList.get(j), finalA);
-                                        initializerList.add(j, new J.Literal(randomId(), initializerList.get(j - 1).getPrefix(), Markers.EMPTY, newAttributeListValue, newAttributeListValue, null, JavaType.Primitive.String));
-                                    }
-                                }
-
+                                updateInitializerList(finalA, initializerList, getAttributeValues());
```
knutwannheden: AFAICT the `updateInitializerList()` method updates the list in-place. This means that the `withInitializer()` should typically return `this` again. But since we here under the hood use a `JContainer` a referential equality check is actually not possible, so it appears to be working. Even though we are here working on a copy (as it is based on a `JContainer` list), I think it would still be best to instead create a new copy of the list, as this isn't immediately apparent from looking at the code.

Laurens-W: Inside `updateInitializerList` you could probably use ListUtils.map, returning null when you want elements to be removed and maintaining a separate list of entries to add, which you can concat at the end. Theoretically of course :)

jevanlingen: Ugh, I just moved that part 😆. Well, let me see what I can do!

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -115,117 +110,51 @@ public J.Annotation visitAnnotation(J.Annotation a, ExecutionContext ctx) {
...
-                    // First assume the value exists amongst the arguments and attempt to update it
+
+                    String newAttributeValueResult = attributeIsArray(a) ? getAttributeValuesAsString() : newAttributeValue;
+                    return JavaTemplate
+                            .apply("#{} = #{}", getCursor(), a.getCoordinates().replaceArguments(), attributeName, newAttributeValueResult);
+                } else if (!TRUE.equals(addOnly)) {
+                    // UPDATE the value when the annotation has arguments, like `@Foo(name="example")`
                     AtomicBoolean foundOrSetAttributeWithDesiredValue = new AtomicBoolean(false);
                     final J.Annotation finalA = a;
                     List<Expression> newArgs = ListUtils.map(currentArgs, it -> {
```
JohannisK, line 124: I wonder if this could be further improved by using visitors.
I think we can visit the children of the annotation, and then visit specifically:
J.Assignement -> this represents all (including single) named assignments
J.Literal -> this represents all **implicit** "value = <string>" assignements
J.FieldAccess -> this represents all **implicit** "value = <ENUM>" assignements
J.NewArray -> this represents all **implicit** "value = {}" assignments.
The J.Assignement can use a super call so it will fall through to J.Literal, J.FieldAccess or J.NewArry visitors to modify the actual value.

This subvisitor can keep track of changes it does, so in the end we can still decide if we need to add a new argument.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
jevanlingen: But I agree with you: https://github.com/openrewrite/rewrite/pull/5698/commits/3d263b7fa4c11ff1f82eab8a47c3b9302a1c1c43.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
jevanlingen: I did think about this, but I don't really think the code will get _much_ cleaner that way. Because if you would go down this route, for every of these methods you need to do a check if you are located within an annotation or maybe part of a subtree which is still located in the annotation. I will break the "update stuff" up in different functions though in the next PR, I think that's the nice middle ground.

Note: Normally I would agree with you, but for this case I don't think it's both worth the effort + would really be an improvement.

timtebeek:
Seeing downstream failures that would either need a quick fix, or revert:
1. https://ge.openrewrite.org/s/44buchebuga4u/tests/task/:test/details/org.openrewrite.java.recipes.MissingOptionExampleTest/lacksExample()?top-execution=1
2. https://ge.openrewrite.org/s/7pxvpya7davvy/tests/task/:test/details/org.openrewrite.java.migrate.javax.AddColumnAnnotationTest/updateColumnAnnotationWithExistingAttributes()?top-execution=1

timtebeek:
Reverted for now in https://github.com/openrewrite/rewrite/commit/b23dfc2d5a3187b289eaebd6d973eb3405db118b, as @jevanlingen is out tonight, but snapshots roll out continuously.

PR #5701, opened 2025-07-02
Add a Gradle best practices recipe

timtebeek:
As a start towards
- https://docs.gradle.org/current/userguide/best_practices_general.html
- https://blog.gradle.org/gradle-best-practices

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
@@ -44,7 +44,8 @@ public class DependencyUseStringNotation extends Recipe {
     public String getDescription() {
         return "In Gradle, dependencies can be expressed as a `String` like `\"groupId:artifactId:version\"`, " +
                 "or equivalently as a `Map` like `group: 'groupId', name: 'artifactId', version: 'version'`. " +
-                "This recipe replaces dependencies represented as `Maps` with an equivalent dependency represented as a `String`.";
+                "This recipe replaces dependencies represented as `Maps` with an equivalent dependency represented as a `String`, " +
+                "as recommended per the [Gradle best practices for dependencies to use a single GAV](https://docs.gradle.org/8.14.2/userguide/best_practices_dependencies.html#single-gav-string).";
     }
 
     @Override


rewrite-gradle/src/main/resources/META-INF/rewrite/gradle.yml
```diff
@@ -15,6 +15,14 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.gradle.GradleBestPractices
+displayName: Apply Gradle best practices
+description: Apply a set of [Gradle best practices](https://docs.gradle.org/current/userguide/best_practices_general.html) to the build files, for more efficient and ideomatic builds.
+recipeList:
+  - org.openrewrite.gradle.MigrateToGradle8
+  - org.openrewrite.gradle.DependencyUseStringNotation
```
timtebeek, line 23: I'd wondered if we should be adding a few more here, as for best practices recipes we can go above and beyond.
```suggestion
  - org.openrewrite.gradle.DependencyUseStringNotation
  - org.openrewrite.gradle.EnableDevelocityBuildCache
  - org.openrewrite.gradle.EnableGradleBuildCache
  - org.openrewrite.gradle.EnableGradleParallelExecution
```

shanman190: I'd probably replace the Develocity build cache one with the OSS Gradle property `org.gradle.caching=true`. Rationale is that recipe needs the Develocity plugin to be applied which may not be true.

timtebeek: Ah I had thought that would no-op if develocity wasn't found, but I'll take it out
```suggestion
  - org.openrewrite.gradle.DependencyUseStringNotation
  - org.openrewrite.gradle.EnableGradleBuildCache
  - org.openrewrite.gradle.EnableGradleParallelExecution
```

shanman190: Ah, you are indeed right. It explicitly looks for an existing Develocity section and then updates that in place only if one is present.

timtebeek: Then ok to keep both? Just wondering which of the above blocks to apply, and whether to add/remove any others.

shanman190:
off-topic: Should we consider adding a link in the descriptions of applicable recipes as well? Such as a link to the best practice of using the String notation for the `DependencyUseStringNotation` recipe.

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle.yml
shanman190: Discussed a little with Tim offline, we should go ahead under the assumption of targeting OSS Gradle best practices and plan for Enterprise Gradle best practices to extend off of these within company-specific recipe libraries.

shanman190 (approved):

PR #5704, opened 2025-07-03
Fixes for Java RPC

knutwannheden:
The `JavaSender` and `JavaReceiver` logic contained a few small bugs leading to RPC transmission errors.

Additionally, the JS implementation of `J.VariableDeclarations` doesn't have the property `dimensionsBeforeName`, which was deprecated a long time ago and indeed isn't used anywhere. So removing that field on the Java side too.


knutwannheden:
I did find one usage in `rewrite-static-analysis`, which would have been broken by this change, but I've fixed that here: https://github.com/openrewrite/rewrite-static-analysis/commit/887b95444f0193d68ca1015025a6f440b7c76a47

PR #5706, opened 2025-07-03
Added datatable

JohannisK:
## What's changed?
Added a datatable with the dependencies found

## What's your motivation?
Especially when ran with wildcards in either group or artifact, you might want to know what was found **exactly**

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/FindDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/FindDependency.java
@@ -18,10 +18,7 @@ package org.openrewrite.gradle.search;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Option;
-import org.openrewrite.Recipe;
-import org.openrewrite.TreeVisitor;
+import org.openrewrite.*;
 import org.openrewrite.gradle.trait.GradleDependency;
 import org.openrewrite.marker.SearchResult;
 
@@ -45,6 +42,8 @@ public class FindDependency extends Recipe {
     @Nullable
     String configuration;
 
+    transient FoundDependencyReport foundDependencyTable = new FoundDependencyReport(this);
+
     @Override
     public String getDisplayName() {
         return "Find Gradle dependency";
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/FindDependency.java
```diff
@@ -102,6 +104,8 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
             public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 if (isDependencyTag(groupId, artifactId) &&
                     versionIsValid(version, versionPattern, () -> findDependency(tag))) {
+                    ResolvedDependency rd = findDependency(tag);
```
JohannisK, line 107: 1. Do we want to add the same functionality to Maven?
2. How much would this impact performance? Since `findDependency(tag)` is wrapped in a supplier in other places?

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/FindDependency.java
```diff
@@ -64,10 +63,35 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new GradleDependency.Matcher()
+        GradleDependency.Matcher matcher = new GradleDependency.Matcher();
+        return matcher
```
timtebeek: Perhaps I missed something, but this can just remain as is right?
```suggestion
        return new GradleDependency.Matcher()
```

JohannisK: yes, reverted

jevanlingen (approved):

rewrite-gradle/src/test/java/org/openrewrite/gradle/search/FindDependencyTest.java
```diff
@@ -125,6 +127,85 @@ void findDependencyByGlob() {
         );
     }
 
+    @Test
+    void generatesDataTableForFoundDependencies() {
+        rewriteRun(
+          spec -> {
+              spec.dataTable(FindDependency.Row.class, rows -> {
+                  assertThat(rows).containsExactly(
+                    new FindDependency.Row("project/build.gradle", "org.openrewrite", "rewrite-core", "latest.release"),
```
jkschneider, line 136: Let's use the existing data table from DependencyInsight

timtebeek:
Suggestion indeed to instead use either the Gradle, or the Maven or Gradle recipe to create these insights:
- https://docs.openrewrite.org/recipes/java/dependencies/dependencyinsight#data-tables
- https://docs.openrewrite.org/recipes/gradle/search/dependencyinsight
- https://docs.openrewrite.org/recipes/maven/search/dependencyinsight

These recipes already produce data tables that include GAV as well as project name & source set. Hope that helps!

PR #5710, opened 2025-07-03
fix: remove unused imports after deleting a method argument

pdelagrave:
## What's changed?
An import will now be removed if the argument deletion by the `DeleteMethodArgument` recipe leaves it unused.

## What's your motivation?
So we can simplify recipes using `DeleteMethodArgument`, no need to explicitly call `RemoveUnusedImports` afterwards anymore.

@timtebeek 

## Any additional context
- Came up when working on https://github.com/openrewrite/rewrite-apache/pull/85

### Checklist


timtebeek:
Tests look good so far. Thanks for getting this started.

timtebeek (approved):
Thanks for the fix here! I like the variety you've used in your tests to guarantee we can remove various different styles of imports. The changes here are additive and mostly defensive, for a recipe that's up to now sparingly used in
* Mockito migrations in rewrite-testing-frameworks
   * https://github.com/openrewrite/rewrite-testing-frameworks/blob/79e3b3f46345ea526131fc7fb5791bd94002aad3/src/main/resources/META-INF/rewrite/mockito.yml#L111-L140
* Jakarta 10 servlet migrations
   * https://github.com/openrewrite/rewrite-migrate-java/blob/a141c96ba43b41117c12c5c2a8b7facd0e07bb65/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml#L335-L352

With that I see not reason to hold off on a merge, such that we can unblock and simplify the work done on
 - https://github.com/openrewrite/rewrite-apache/pull/85#discussion_r2182829642

Not merging right now myself, as I think that's best done when everyone is back to work Monday.

okundzich:
what is the reason to not use another recipe that manipulates imports afterwards? i thought this is how many recipes function, like change method requires adding a recipe to manipulate imports afterwards. 

> So we can simplify recipes using DeleteMethodArgument, no need to explicitly call RemoveUnusedImports afterwards anymore.

timtebeek:
The reason we wanted to include this part of the change in `DeleteMethodArgument` is to limit the scope (and potential for impact) of a `RemoveUnusedImports` that would directly follow this in Yaml, as we saw on
- https://github.com/openrewrite/rewrite-apache/pull/85

If we were to use for instance
```yaml
recipeList:
  - org.openrewrite.java.DeleteMethodArgument:
      methodPattern: org.apache.hc.client5.http.impl.auth.BasicSchemeFactory <constructor>(java.nio.charset.Charset)
      argumentIndex: 0
  - org.openrewrite.java.RemoveUnusedImports
```
both would apply to all files, with `RemoveUnusedImports` much more likely to make changes as compared to `DeleteMethodArgument`, perhaps confusingly so for a recipe doing part of an Apache HttpClient migration.

There is a way to limit the scope with preconditions, but that then quickly leads to bloat in defining yaml recipes. E.g.
```yaml
name: org.openrewrite.apache.httpclient5.MigrateBasicSchemeFactory
displayName: Removes the charset argument from `BasicSchemeFactory()` constructor calls
description: RFC 7617 mandates the use of UTF-8. This constructor is now deprecated and httpclient v5 ignores any provided charset, ensuring that only UTF-8 is used.
preconditions:
  - org.openrewrite.java.search.FindMethods:
      methodPattern: org.apache.hc.client5.http.impl.auth.BasicSchemeFactory <constructor>(java.nio.charset.Charset)
recipeList:
  - org.openrewrite.java.DeleteMethodArgument:
      methodPattern: org.apache.hc.client5.http.impl.auth.BasicSchemeFactory <constructor>(java.nio.charset.Charset)
      argumentIndex: 0
  - org.openrewrite.java.RemoveUnusedImports
```

By including the import handling here we can just use the following without a need for `preconditions:` or `RemoveUnusedImports`.
```yaml
recipeList:
  - org.openrewrite.java.DeleteMethodArgument:
      methodPattern: org.apache.hc.client5.http.impl.auth.BasicSchemeFactory <constructor>(java.nio.charset.Charset)
      argumentIndex: 0
```

This is inline with other building block recipes that similarly remove imports themselves, as for instance seen in
https://github.com/openrewrite/rewrite/blob/5e1473e06e3462f302d1794f2ade7e7676005da2/rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodTargetToStatic.java#L128

Hope you agree given that context that we thought this would lead to the most maintainable usage downstream, to for instance also remove the import associated with [`jakarta.servlet.Servlet` as used here in rewrite-migrate-java](https://github.com/openrewrite/rewrite-migrate-java/blob/a141c96ba43b41117c12c5c2a8b7facd0e07bb65/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml#L347-L352).

okundzich:
sounds good, thanks for clarifying

PR #5711, opened 2025-07-03
Support trailing comma in GroovyParser

e5LA:
## What's changed?
Added support for trailing comma in GroovyParser. Using `Marker.TrailingComma` as suggested in the original issue.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/4614

Usage of the `Marker`

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -796,12 +796,11 @@ public class GroovyParserVisitor {
                     }
 
                     Space after = EMPTY;
-                    Markers markers = Markers.EMPTY;
                     if (i == unparsedArgs.size() - 1) {
                         if (hasParentheses) {
                             saveCursor = cursor;
                             Space before = whitespace();
-                            if (sourceStartsWith(",")) {
+                            if (source.charAt(cursor) == ',') {
                                 skip(",");
                                 JRightPadded<Expression> arg = JRightPadded.build(exp)
                                         .withMarkers(exp.getMarkers().add(new TrailingComma(randomId(), sourceBefore(")"))))
@@ -823,7 +822,7 @@ public class GroovyParserVisitor {
                         cursor++;
                     }
 
-                    args.add(JRightPadded.build(exp).withAfter(after).withMarkers(markers));
+                    args.add(JRightPadded.build(exp).withAfter(after));
                 }
             }
 
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -790,24 +790,31 @@ public class GroovyParserVisitor {
                 boolean lastArgumentsAreAllClosures = endsWithClosures(expression.getExpressions());
                 for (int i = 0; i < unparsedArgs.size(); i++) {
                     org.codehaus.groovy.ast.expr.Expression rawArg = unparsedArgs.get(i);
-                    Expression arg = visit(rawArg);
+                    Expression exp = visit(rawArg);
                     if (!hasParentheses) {
-                        arg = arg.withMarkers(arg.getMarkers().add(new OmitParentheses(randomId())));
+                        exp = exp.withMarkers(exp.getMarkers().add(new OmitParentheses(randomId())));
                     }
 
                     Space after = EMPTY;
                     Markers markers = Markers.EMPTY;
                     if (i == unparsedArgs.size() - 1) {
                         if (hasParentheses) {
-                            if (source.charAt(cursor) == ',') {
+                            saveCursor = cursor;
+                            Space before = whitespace();
+                            if (sourceStartsWith(",")) {
                                 skip(",");
-                                TrailingComma t = new TrailingComma(randomId(), sourceBefore(")"));
-                                markers = Markers.build(singletonList(t));
+                                JRightPadded<Expression> arg = JRightPadded.build(exp)
+                                        .withMarkers(exp.getMarkers().add(new TrailingComma(randomId(), sourceBefore(")"))))
+                                        .withAfter(before);
+
+                                args.add(arg);
+                                continue;
                             } else {
+                                cursor = saveCursor;
                                 after = sourceBefore(")");
                             }
                         }
-                    } else if (!(arg instanceof J.Lambda && lastArgumentsAreAllClosures && !hasParentheses)) {
+                    } else if (!(exp instanceof J.Lambda && lastArgumentsAreAllClosures && !hasParentheses)) {
                         after = whitespace();
                         if (source.charAt(cursor) == ')') {
                             // next argument(s), if they exists, are trailing closures and will have an OmitParentheses marker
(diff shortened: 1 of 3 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/GroovyParserTest.java
```diff
@@ -128,4 +128,18 @@ void variableDeclarationWithArrayInstantiationAsTheLastStatement() {
...
     }
+
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite/issues/4614")
+    void trailingCommaInMethodCall() {
+        rewriteRun(
+          groovy(
+            """
+              System.out.println("Hello World", )
+              System.out.println("Hello World with no extra space",)
```
timtebeek, line 138: This made me wonder if we can also handle a space before the comma. Likely yes given the quality of your earlier contributions, but perhaps still good to check here. :)

e5LA: Ah, I missed that this time.. Thanks @timtebeek for spotting it and adding the test case!
I'll take a look.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -796,9 +796,16 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
...
 
                     Space after = EMPTY;
+                    Markers markers = Markers.EMPTY;
                     if (i == unparsedArgs.size() - 1) {
                         if (hasParentheses) {
-                            after = sourceBefore(")");
+                            if (source.charAt(cursor) == ',') {
+                                skip(",");
+                                TrailingComma t = new TrailingComma(randomId(), sourceBefore(")"));
+                                markers = Markers.build(singletonList(t));
```
jevanlingen:
```suggestion
                                markers = Markers.build(singleton(new TrailingComma(randomId(), sourceBefore(")"))));
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -796,9 +796,16 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
                     }
 
                     Space after = EMPTY;
+                    Markers markers = Markers.EMPTY;
                     if (i == unparsedArgs.size() - 1) {
                         if (hasParentheses) {
-                            after = sourceBefore(")");
+                            if (source.charAt(cursor) == ',') {
```
jevanlingen, line 803:
```suggestion
                            if (sourceStartsWith(",")) {
```

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -796,9 +796,16 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
...
+                                skip(",");
```
jevanlingen, line 804: You probably need to take into account there can be whitespace before the comma.

rewrite-groovy/src/test/java/org/openrewrite/groovy/GroovyParserTest.java
jevanlingen: @e5LA just did a quick review with will give you some pointers where/how to fix this.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
e5LA: Thanks for the review! Valid points, I appreciate it.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
e5LA: Actually I've noticed that the `if(source.charAt(cursor) == '')` is used in quite a few places, so I just wonder - is `sourceStartsWith("")` functionally equivalent in this context? 
If so, would it make sense to update the other occurrences as well, for consistency? 

Just want to make sure I'm not missing some details here.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
e5LA: Indeed, added in https://github.com/openrewrite/rewrite/pull/5711/commits/c7df52235501528201438b4af9baf6e6c677c19d
Thanks for your comments.

rewrite-groovy/src/test/java/org/openrewrite/groovy/GroovyParserTest.java
e5LA: Thank you both for the review.
It's covered by https://github.com/openrewrite/rewrite/pull/5711/commits/c7df52235501528201438b4af9baf6e6c677c19d

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
greg-at-moderne: >  if(source.charAt(cursor) == '') is used in quite a few places, so I just wonder - is sourceStartsWith("")

For the context: these two are not always the same, as the latter has a call to`indexOfNextNonWhitespace`.
Not fully sure what @jevanlingen thinks, but from my perspective:
- using `source.charAt()` is better when checking for a single character
- as it's more explicit and easier to understand without looking up some other method.

Having said, if it was up to me, I would start any mass-conversion from `source.charAt` to `indexOfNextNonWhitespace`.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -790,17 +790,31 @@ public void visitArgumentlistExpression(ArgumentListExpression expression) {
...
                     org.codehaus.groovy.ast.expr.Expression rawArg = unparsedArgs.get(i);
-                    Expression arg = visit(rawArg);
+                    Expression exp = visit(rawArg);
                     if (!hasParentheses) {
-                        arg = arg.withMarkers(arg.getMarkers().add(new OmitParentheses(randomId())));
+                        exp = exp.withMarkers(exp.getMarkers().add(new OmitParentheses(randomId())));
                     }
 
                     Space after = EMPTY;
+                    Markers markers = Markers.EMPTY;
```
greg-at-moderne: This one isn't used anywhere, is it?

e5LA: Thanks for the review! 
In fact the variable was used, but after your comment I realized it was actually a leftover from my previous changes, and it's not needed. Therefore I deleted it.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
e5LA: Thanks for your input @greg-at-moderne 
I decided to keep the original `source.charAt(cursor) == ','`, to keep it explicit and consistent.

greg-at-moderne (approved):

Issue #5712, opened 2025-07-03
`AnnotationTemplateGenerator` does not handle nested annotations correctly


steve-aom-elliott:
## What version of OpenRewrite are you using?

- OpenRewrite v8.56.1 (but really latest commit)
- rewrite-openapi v0.21.0 (but really latest commit)

## How are you running OpenRewrite?

I am debugging tests within `rewrite-openapi` directly, but the logic issue appears to be within `rewrite`.

## What is the smallest, simplest way to reproduce the problem?
```java
@Test
    void replaceNestedAnnotation() {
        rewriteRun(
          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
              private AnnotationMatcher ANNOTATION_MATCHER = new AnnotationMatcher("@NestedAnnotation");
              public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ctx) {
                  J.Annotation an = super.visitAnnotation(annotation, ctx);
                  if (!ANNOTATION_MATCHER.matches(an) || an.getArguments().isEmpty()) {
                      return an;
                  }
                  J.Assignment assignment = (J.Assignment) an.getArguments().get(0);
                  if (assignment.getVariable() instanceof J.Identifier && ((J.Identifier) assignment.getVariable()).getSimpleName().equals("a")) {
                      J.Identifier id = (J.Identifier) assignment.getVariable();
                      return JavaTemplate.builder("#{any()}")
                        .build()
                        .apply(
                          getCursor(),
                          an.getCoordinates().replaceArguments(),
                          assignment.withVariable(id.withSimpleName("b"))
                        );
                  return an;
              }
          })).parser(JavaParser.fromJavaVersion()
            .dependsOn(
              """
                import java.lang.annotation.Repeatable;
                @Repeatable(NestedAnnotations.class)
                @interface NestedAnnotation {
                  String a() default "";
                }
                """,
              """
                @interface NestedAnnotations {
                  NestedAnnotation[] value() default {};
                }
                """
            )
          ),
          //language=java
          java(
            """
              class A {
                @NestedAnnotations(value = {
                  @NestedAnnotation(a = "first"),
                  @NestedAnnotation(a = "second")
                })
                void method() {}
              }
              """,
            """
              class A {
                @NestedAnnotations(value = {
                  @NestedAnnotation(b = "first"),
                  @NestedAnnotation(b = "second")
                })
                void method() {}
              }
              """
          )
        );
    }
```
[...]

PR #5713, opened 2025-07-04
Fix `AnnotationTemplateGenerator` to handle nested annotations correctly

knutwannheden:
The `AnnotationTemplateGenerator` was failing with an `IndexOutOfBoundsException` when processing nested
annotations (e.g., repeatable annotations within annotation arrays). This occurred because the template
generator couldn't find a "typical annotation target" for annotations nested inside other annotations' array
values.

The fix adds proper handling for nested annotations by:
1. Walking up the parent hierarchy to find the ultimate annotation target (method, variable, or class)
2. Adding a fallback case that generates `@interface $Placeholder {}` when no typical target can be found
3. Updating both the `template()` and `cacheKey()` methods to handle nested annotation cases consistently

This ensures that `JavaTemplate` can successfully process and transform nested annotations without throwing
exceptions.

Added a test case that demonstrates the fix by transforming nested repeatable annotations' attributes using
`JavaTemplate`.

Fixes:
- #5712

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
@@ -53,6 +53,33 @@ public class AnnotationTemplateGenerator {
             } else {
                 after.insert(0, "static class $Clazz {}");
             }
+        } else if (j instanceof J.Annotation) {
+            // Handle nested annotations by finding their ultimate parent
+            J annotationParent = cursor.getParent() != null ? cursor.getParent().firstEnclosing(J.class) : null;
+            int level = 1;
+            while (annotationParent instanceof J.NewArray || annotationParent instanceof J.Annotation) {
+                level += 1;
+                if (cursor.getParent(level) == null) {
+                    break;
+                }
+                annotationParent = cursor.getParent(level).firstEnclosing(J.class);
+            }
+            
+            if (annotationParent instanceof J.MethodDeclaration) {
+                after.insert(0, " void $method() {}");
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
```diff
@@ -89,12 +89,18 @@ public String template(Cursor cursor, String template) {
...
+                        Cursor classCursor = j instanceof J.ClassDeclaration ? cursor : cursor.getParent(level);
+                        if (classCursor != null && classCursor.getParentOrThrow().getValue() instanceof JavaSourceFile) {
                             after.insert(0, "class $Clazz {}");
                         } else {
                             after.insert(0, "static class $Clazz {}");
                         }
+                    } else if (j instanceof J.Annotation && annotationParent == null) {
+                        // Fallback for annotations that couldn't find a typical target
+                        // This handles cases like nested annotations in annotation arrays
+                        after.insert(0, "@interface $Placeholder {}");
```
timtebeek: When running `JavaTemplateAnnotationTest` with coverage enabled, none of the tests seem to hit this branch. Would we still want to retain this, and maybe add a test?

timtebeek (approved):
Approved already; could merge as is, but wasn't sure given that you had marked this as a draft.

rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
knutwannheden: Removed the branch.

PR #5716, opened 2025-07-04
RemoveDependency recipe removes or alters comments

jevanlingen:
## What's changed?
Comments after dependencies are removed as you would expected.

## What's your motivation?
Comments after a dependency should either stay if dependency is not removed or removed if dependency is not removed. Before the fix in this PR, the comments were put on a different dependency or were removed altogether, because all whitespace data is located in the prefix. By removing a method invocation you basically 'move' the comment to another line.

E.g. when using this recipe prior to this fix and we remove `io.github.classgraph:classgraph`:

**Before**
```groovy
implementation("io.github.classgraph:classgraph:latest.release") // there's something specific about the classgraph
implementation("org.yaml:snakeyaml:latest.release")
```

**After**
```groovy
implementation("io.micrometer:micrometer-core:1.9.+") // there's something specific about the classgraph
implementation("org.yaml:snakeyaml:latest.release")
```

## Any additional context
You could argue this PR tries to fix a flaw of the LST model (as there is no suffix element). I do not suggest we apply this kind of fix to all recipes where we remove method invocations. But as this is one of the core recipes, I think it makes sense to have this fix for this recipe.

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveDependencyTest.java
@@ -146,7 +146,7 @@ class RemoveDependencyTest implements RewriteTest {
                   testImplementation "org.junit.vintage:junit-vintage-engine:5.6.2"        /* comment 1 */ /* comment 2 */ // comment 3
                   // and more
                   /* and more */ implementation "org.springframework.boot:spring-boot-starter-web:2.7.0"
-                  def someStatementWithinDepBlock = 33
+                  def someStatementWithinDepBlock = 33  // some other comment
                   implementation "com.google.guava:guava:29.0-jre"
               }
               """,
@@ -161,7 +161,7 @@ class RemoveDependencyTest implements RewriteTest {
               
               dependencies {
                   testImplementation "org.junit.vintage:junit-vintage-engine:5.6.2"        /* comment 1 */ /* comment 2 */ // comment 3
-                  def someStatementWithinDepBlock = 33
+                  def someStatementWithinDepBlock = 33  // some other comment
                   implementation "com.google.guava:guava:29.0-jre"
               }
               """
@@ -187,7 +187,8 @@ class RemoveDependencyTest implements RewriteTest {
                   // Comment 1
                   implementation "org.springframework.boot:spring-boot-starter-web:2.7.0" /* comment 2 */ /* comment 3 */ // comment 4
                   // and more
-                  implementation "com.google.guava:guava:29.0-jre"
+                  implementation "com.google.guava:guava:29.0-jre" // some other comment
+                  implementation "org.yaml:snakeyaml:latest.release"
               }
               """,
             """
@@ -202,7 +203,8 @@ class RemoveDependencyTest implements RewriteTest {
               dependencies {
                   testImplementation "org.junit.vintage:junit-vintage-engine:5.6.2"
                   // and more
-                  implementation "com.google.guava:guava:29.0-jre"
+                  implementation "com.google.guava:guava:29.0-jre" // some other comment
+                  implementation "org.yaml:snakeyaml:latest.release"
               }
               """
           )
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveDependencyTest.java
```diff
@@ -129,6 +129,116 @@ void removeGradleDependencyUsingStringNotationWithExclusion() {
...
+                  id 'java-library'
+              }
+              
+              repositories {
+                  mavenCentral()
+              }
+              
+              dependencies {
+                  testImplementation "org.junit.vintage:junit-vintage-engine:5.6.2"        /* comment 1 */ /* comment 2 */ // comment 3
+                  def someStatementWithinDepBlock = 33
```
Laurens-W: Honestly I don't think we can explain this behavior to the customer nor will it always be correct. I'd be inclined to keep all comments intact and have the user (during a manual review) choose to remove comments that no longer apply.

Laurens-W: That also makes the logic surrounding comments a lot simpler: prepend them to the comment list of the next element

jevanlingen: I don't agree. At least 95% of all users would expect is that the comments behind a line would apply to the current line:

```java
someStatement(); // something about this line what you should know
```

The oracle docs do follow [this](https://www.oracle.com/java/technologies/javase/codeconventions-comments.html#:~:text=Java%20programs%20can%20have%20two,are%20delimited%20by%20%2F**...) convention and to be honest everywhere I've ever seen code this convention is applied. See also [this](https://chatgpt.com/share/686bc0d5-b2dc-8009-a1a5-62eba2a62482) question.

Though you are right we cannot know for sure, I thinks it's easy to explain, as it's a simple rule from user perspective:
> Comments before a dependency will be removed and comments on the same line after the dependency will removed.

I noticed my code was not good enough for this rule, because I forgot to implement one important case: "What should happen when you remove a dependency with comments?". So I made another test: https://github.com/openrewrite/rewrite/pull/5716/commits/67be0725aab6051b9fe3acbc85be67221f57eaf3. Do you agree this test makes a lot of sense?

Laurens-W: Normally I would agree but in this instance we cannot distinguish between that 5% and 95% as that needs context we do not have.
I agree that test makes sense, but not every customer might unfortunately

jevanlingen: The current behaviour, so before this PR, is also problematic as it cause changes to the line above. If you would take the example of this PR and you would run the recipe, you would end up with following diff:

```diff
- implementation("io.micrometer:micrometer-core:1.9.+")
+ implementation("io.micrometer:micrometer-core:1.9.+") // there's something specific about the classgraph
- implementation("io.github.classgraph:classgraph:latest.release") // there's something specific about the classgraph
implementation("org.yaml:snakeyaml:latest.release")
```

That 95% was just a number I picked, but now I look at this diff I am pretty certain it's closer to 99.5%. I don't know how to convince you, because you are still right. We cannot know what the comment is really about. To me it does make sense the default behaviour of this recipe would create following diff though:

```diff
implementation("io.micrometer:micrometer-core:1.9.+")
- implementation("io.github.classgraph:classgraph:latest.release") // there's something specific about the classgraph
implementation("org.yaml:snakeyaml:latest.release")
```

Laurens-W: I'm just not sure how tightly we want to stick to "do no harm" for comments, I'm convinced that in most cases this would be perfectly fine :)

Laurens-W (approved):

rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveDependencyTest.java
jevanlingen: 👍🏻 let's see what other reviewers have to say about this.

PR #5720, opened 2025-07-04
Changing what `LatestRelease` recognizes as the latest release or latest snapshot keyword versions by expanding it to include those from Ivy/Gradle

steve-aom-elliott:
## What's changed?
- Expanded `LatestRelease` to not only recognize `LATEST` and `RELEASE` from Maven, but also their equivalents from Ivy / Gradle (`latest.integration` and `latest.release`)

## What's your motivation?
We were seeing `UpgradeDependencyVersion` for Gradle change `latest.release` to a specific version number rather than leaving it alone.

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
@@ -987,6 +987,29 @@ class UpgradeDependencyVersionTest implements RewriteTest {
     }
 
     @Test
+    void retainLatestReleaseOrLatestIntegrationIfUsed() {
+        rewriteRun(
+          spec -> spec.recipe(new UpgradeDependencyVersion("org.projectlombok", "lombok", "1.18.*", null)),
+          buildGradle(
+            """
+              plugins {
+                  id 'java'
+              }
+              
+              repositories {
+                  mavenCentral()
+              }
+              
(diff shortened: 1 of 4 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -986,6 +986,29 @@ void disallowDowngrade() {
         );
     }
 
+    @Test
+    void retainLatestReleaseOrLatestIntegrationIfUsed() {
+        rewriteRun(
+          spec -> spec.recipe(new UpgradeDependencyVersion("org.projectlombok", "lombok", "1.18.*", null)),
```
timtebeek, line 1177: Adding the context here that this is [coming from](https://github.com/openrewrite/rewrite-migrate-java/blob/e4e2ecd60ed8f9e013b43b0168f5497471bd59a3/src/main/resources/META-INF/rewrite/lombok.yml#L41-L51) in `org.openrewrite.java.migrate.lombok.UpdateLombokToJava11`.

timtebeek (approved):
This looks good to me already; I'm somewhat surprised we hadn't run up against this before. I'm going to chalk that up to only recently having added support for .kts files which our own projects use alongside `latest.release`.

Given the potential for impact I'm going to ask for a quick review from @jkschneider to resolve this customer issue.

PR #5724, opened 2025-07-07
Added additional tests for nested annotations, and a fix to break out of the `J.Assignment` for finding an annotation parent that is more significant when the nearest desired parent is a not a `J.ClassDeclaration`

steve-aom-elliott:
- Finishes https://github.com/openrewrite/rewrite/issues/5712
- Follow-on to https://github.com/openrewrite/rewrite/pull/5713 to fix case when the nearest desired parent isn't a `J.ClassDeclaration` and your wrapping annotation uses a `J.Assignment` for the `J.NewArray` rather than just a raw `J.NewArray`.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateAnnotationTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateAnnotationTest.java
@@ -16,10 +16,12 @@
 package org.openrewrite.java;
 
 import org.intellij.lang.annotations.Language;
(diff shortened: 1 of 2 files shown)


knutwannheden (approved):
Good to have even more supported cases. Did we already have any issue for this?

steve-aom-elliott:
> Good to have even more supported cases. Did we already have any issue for this?

- https://github.com/openrewrite/rewrite-openapi/issues/16#issuecomment-3014704319 was what spawned https://github.com/openrewrite/rewrite/issues/5712

PR #5725, opened 2025-07-07
Adding recipe for `DoesNotIncludeDependency` for Gradle (Maven had one already), to allow for checking whether a dependency is not included.

steve-aom-elliott:
## What's changed?
- Added recipe to check that a dependency is not included in Gradle

## What's your motivation?
I have a need for being able to detect that a dependency is not included to decide whether to proceed with additional migrations or not. Maven already had a version of this, but Gradle was missing one, and I want to be thorough in this precondition check.

I have not written tests for this quite yet. All of the tests for the Maven version appear to be commented out for some reason, though I do want to add a couple of examples in tests still.

One thing to point out is that it does not include the same `directOnly` option as the Maven version, as the way the implementations of `DependencyInsight` is processing dependencies between Maven and Gradle is slightly different, and that class does not take the option in the Gradle version.

### Checklist


rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
```diff
@@ -0,0 +1,71 @@
...
+            example = "guava")
+    String artifactId;
+
+    @Option(displayName = "Scope",
+            description = "Default any. If specified, only the requested scope's classpaths will be checked.",
+            required = false,
+            valid = {"compile", "test", "runtime", "provided"},
+            example = "compile")
+    @Nullable
+    String scope;
```
Laurens-W: This needs to change to suite Gradle's `configuration`, see the Gradle `DependencyInsight` recipe for inspo

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.not(Preconditions.or(dependencyInsightVisitors()));
+    }
+
+    @SuppressWarnings("unchecked")
+    private TreeVisitor<?, ExecutionContext>[] dependencyInsightVisitors() {
+        if (scope == null) {
+            return new TreeVisitor[] {
+                    new org.openrewrite.gradle.search.DependencyInsight(groupId, artifactId, null, null).getVisitor(),
```
Laurens-W: See if you can import this instead of fully qualifying it :)

steve-aom-elliott:
Still working on the tests to include ones for specific configurations and transitive dependencies, but getting closer.

rewrite-gradle/src/test/java/org/openrewrite/gradle/search/DependencyInsightTest.java
```diff
@@ -65,6 +73,71 @@ void findTransitiveDependency() {
...
+              }
+              """;
+            rewriteRun(
+              spec -> spec.recipe(new DependencyInsight("com.google.guava", "guava", null, checkingConfiguration)),
+              expectationHelper(existingConfiguration, gradleTemplate, matchComment, matched)
+            );
+        }
+
+        @ParameterizedTest
+        @MethodSource("configurationsAndMatches")
```
timtebeek, line 123: Would something like this help readability perhaps? Took me 20 seconds with Claude, so don't go changing things manually:
```
        @CsvSource(
          useHeadersInDisplayName = true, textBlock = """
            configuration,         annotationProcessor, api,   implementation, compileOnly, runtimeOnly, testImplementation, testCompileOnly, testRuntimeOnly
            annotationProcessor,   true,                false, false,          false,       false,       false,              false,           false
            compileClasspath,      false,               true,  true,           true,        false,       false,              false,           false
            runtimeClasspath,      false,               true,  true,           false,       true,        false,              false,           false
            testCompileClasspath,  false,               true,  true,           false,       false,       true,               true,            false
            testRuntimeClasspath,  false,               true,  true,           false,       true,        true,               false,           true
            """)
```

rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
```diff
@@ -15,532 +15,472 @@
...
-                    </dependencies>
-                  </project>
-                  """, scope)
-              )
-            );
-        }
+            @ParameterizedTest
+            @CsvSource({"compile,compile", "compile,test"})
+            @CsvSource({"runtime,compile", "runtime,test",})
+            @CsvSource({"test,test"})
```
timtebeek: I was a bit surprised by the repeated annotation here; is there a logic to the grouping that I'm missing? Might a textBlock make sense?

rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
```diff
@@ -15,532 +15,472 @@
...
-                        <scope>%s</scope>
-                      </dependency>
-                    </dependencies>
-                  </project>
-                  """, scope)
-              )
-            );
+        @Nested
+        class OnlyDirectTrue {
+            @DocumentExample
```
timtebeek: The way `@DocumentExample` works is that we look at the `.java` source file, and extract the values without running the tests. In this case I think it wouldn't work, so then it's maybe best to add an example for the docs at the top that's just a single straightforward two text blocks test.

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private TreeVisitor<?, ExecutionContext>[] dependencyInsightVisitors() {
+        if (configuration == null) {
+            return new TreeVisitor[] {
+                    new DependencyInsight(groupId, artifactId, null, null).getVisitor(),
+            };
+        }
+        return new TreeVisitor[] { new DependencyInsight(groupId, artifactId, null, configuration).getVisitor() };
```
timtebeek: Wouldn't this be equivalent and simpler?
```suggestion
        return Preconditions.not(new DependencyInsight(groupId, artifactId, null, configuration).getVisitor());
```

timtebeek:
Mostly looks good already; couple questions & comments, but nothing blocking. Once addressed we can merge.

rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
steve-aom-elliott: I was grouping them by the existing scope of the dependency per row. It was more for my own legibility than anything else.

rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
steve-aom-elliott: Fair enough. I can add one like that separately.

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
steve-aom-elliott: I think it was more a holdover from the Maven version, but yes, I can make that change

rewrite-gradle/src/test/java/org/openrewrite/gradle/search/DependencyInsightTest.java
steve-aom-elliott: Honestly I've never used the `@CsvSource` with `textBlock` before, but yeah, something like that would also work.

steve-aom-elliott:
I think the only things I didn't end up changing right now were the suggestions for using `@CsvSource` with a `textBlock`, but please feel free to use Claude to convert those if they seem more legible afterwards.

timtebeek (approved):
Approved already. Thanks for the help here!

PR #5727, opened 2025-07-08
Re enable `DoesNotIncludeDependencyTest` tests

Laurens-W:
## What's changed?

Re enabled and brought back into shape the `DoesNotIncludeDependencyTest` tests

## What's your motivation?

Some work is being done on a Gradle variant so good to have a maven example

### Checklist


steve-aom-elliott:
I think the only thing I might consider (I can do it as part of the other PR if that makes more sense) is to reorganize the tests so that we have more nested classes and can reuse the `pom.xml` text block across some tests (given it already has a placeholder for scope).

So top level: TransitivelyPresent, DirectlyPresent, NotPresent and then under those nesting additional condition nested classes (so more layers than current).

I have an idea in my head for how to organize it but I can take that on when I add tests for the Gradle version.

timtebeek (approved):
Thanks for reenabling these tests now that we have scanning recipe preconditions again.

PR #5732, opened 2025-07-09
Fix bug in AnnotationTemplateGenerator relative to finding the correct "target" of the annotation.

jonathannewell:
## What's changed?
AnnotationTemplateGenerator has been updated to correctly determine the annotationParent (target).

## What's your motivation?
This is affecting recipes that rewrite complex nested annotations (such as ApiResponse/(es) in swagger)

I have added two tests to the JavaTemplateAnnotationTest. One that ensures the code continued to work as it did before my changes relative to determining annotationParent when the engineer used default or non specified name for the assignment and one where the subannotation collection (array) is assigned (value = ).  The latter test failed before my code change. The new code will only fall through when no parent is found or a valid one (e.g. variable/method/class declaration).

Whoever is most familiar with this area of the internal templating code.

I exhausted all attempts to work around this problem.

## Any additional context
I changed only these two files and ran all the tests in the JavaTemplateAnnotationTest class but did not run your full suite of tests.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateAnnotationTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateAnnotationTest.java
@@ -16,10 +16,12 @@
 package org.openrewrite.java;
 
 import org.intellij.lang.annotations.Language;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Issue;
+import org.openrewrite.Recipe;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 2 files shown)


timtebeek (changes requested):
Thanks for the effort here! It's a little hard to review due to the many formatting changes; would you mind reverting those such that we can focus on the intentional changes?

jonathannewell:
Ok, I fixed all the formatting issues. Which is weird because I did as requested and defaulted my intellij java format before formatting the files before my initial commit. Just went through each and reverted any format changes from the previous commit. I also noted there was an interim commit that my fork did not have so I merged that stuff in. Diff should be clean now. Sorry about that!

timtebeek:
Thanks a lot for making the change set manageable! With that it's already looking a whole lot better and close to a merge. Unfortunately I'm out camping for the next four days, but I'll ask if colleagues have time for a review & merge.

jonathannewell:
Tim, I hope you have a great time camping! I should point out that the interim commit had a fix for this same problem. However, I chose a different approach to the solution. The fix currently in the codebase extends the original thinking by using non-desired classes in the conditional of the while loop. In my mind that is only semi-deterministic. I flipped that and made the conditional about the classes (states) that the remainder of the code is designed to handle. Therefore filling any potential gaps where other class types could show up in the object graph. My thinking could be muddled based on my nascent understanding of the open-rewrite model and design so I will completely understand if the team just closes this pull request. 

Once again have a great time getting back to nature!

timtebeek:
Thanks for the help here! It looks like there have been some parallel efforts here by @steve-aom-elliott :
- https://github.com/openrewrite/rewrite/pull/5724

If I revert the changes in `AnnotationTemplateGenerator` you've added here, then all the tests still pass, including the ones that you added. That makes me wonder if we should be making these changes at all right now. Any insights from the both of you whether we'd need more?

We can reopen this PR in case we've missed some case that we'd like to support still, preferably by seeing a test that fails on main first.

PR #5736, opened 2025-07-09
refactor: Remove redundant null checks before instanceof

timtebeek:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.RemoveRedundantNullCheckBeforeInstanceof?organizationId=ODQ2MGExMTUtNDg0My00N2EwLTgzMGMtNGE1NGExMTBmZDkw

### DIFF
--- a/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
+++ b/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
@@ -62,7 +62,7 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                 element.acceptChildren(this, data)
                 if (element is FirResolvedTypeRef) {
                     // Do not visit FirUserTypeRef, since it's not mappable to a type.
-                    if (element.delegatedTypeRef != null && element.delegatedTypeRef !is FirUserTypeRef) {
+                    if (element.delegatedTypeRef !is FirUserTypeRef) {
                         // not sure why this isn't taken care of by `FirResolvedTypeRefImpl#acceptChildren()`
                         element.delegatedTypeRef?.accept(this, data)
                     }
--- a/rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java
@@ -372,8 +372,7 @@ public class MethodMatcher {
             return false;
         }
 
-        if (method.getSelect() != null &&
-            method.getSelect() instanceof J.Identifier &&
+        if (method.getSelect() instanceof J.Identifier &&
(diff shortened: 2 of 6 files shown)


rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
```diff
@@ -62,7 +62,7 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                 element.acceptChildren(this, data)
                 if (element is FirResolvedTypeRef) {
                     // Do not visit FirUserTypeRef, since it's not mappable to a type.
-                    if (element.delegatedTypeRef != null && element.delegatedTypeRef !is FirUserTypeRef) {
+                    if (element.delegatedTypeRef !is FirUserTypeRef) {
```
timtebeek: This is perhaps the only one where I wasn't 100% certain of the Kotlin nullability implications.

greg-at-moderne: This one uses `!is` - i.e. in Java terms it would be `!(instanceof FirUserTypeRef)`, right?

So why does the recipe kick in here?

greg-at-moderne: It's because `a !is String` is parsed as `J.InstanceOf` with `org.openrewrite.kotlin.marker.NotIs` marker.
Which on one hand I can understand why, on the other hand it's kind of misleading as it means something exactly opposite.
image

Should the recipe just be fixed to test against the `NotIs` Kotlin marker?

greg-at-moderne: And specifically about this source line I think this change is wrong and shouldn't be applied.
Maybe we know from elsewhere, the value is not null, but it doesn't stand out as obvious to me.

timtebeek:
```suggestion
                    if (element.delegatedTypeRef != null && element.delegatedTypeRef !is FirUserTypeRef) {
```

timtebeek: Indeed best to have the recipe not apply to Kotlin, or check for that marker. The use of a Marker here now makes me doubt any recipes we have that use `J.InstanceOf` with Kotlin as this is quite unexpected.

timtebeek: Added a precondition for now: https://github.com/openrewrite/rewrite-static-analysis/commit/3cf0d8a37d5b016f5ea4680fb5cccf326e029c29

PR #5737, opened 2025-07-09
Do not check all-service endpoint if exact version requested

bmuschko:
## What's changed?

The https://services.gradle.org/versions/all endpoint is really only needed to determine the latest version available if a version wildcard notation (e.g. 8.x) or null has been requested by the end user. In all other cases, we are making an unnecessary HTTPS call.

_Side note:_ The `repositoryUrl` always used the null value. I removed the parameter to make the logic less confusing.

## What's your motivation?

Many declarative recipes add the UpdateGradleWrapper recipe and provide a specific version as a default value. For example, the [MigrateToGradle8](https://docs.openrewrite.org/recipes/gradle/migratetogradle8) recipe sets the version option value to 8.x.

Within enterprises, the https://services.gradle.org endpoint is often times not exposed. Calling the endpoint simply results in a failure after running into a HTTP timeout or by not being able to resolve it. End users usually do not know that the Wrapper distribution needs to be resolved from an in-house binary repository; nevertheless they execute the recipe with the default values. For scenarios where the user provided an exact version, we'll increase the performance of the recipe execution.

No

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
@@ -139,7 +139,7 @@ public class UpdateGradleWrapper extends ScanningRecipe<UpdateGradleWrapper.Grad
                 return gradleWrapper = GradleWrapper.create(URI.create(wrapperUri), ctx);
             }
             try {
-                gradleWrapper = GradleWrapper.create(distribution, version, null, ctx);
+                gradleWrapper = GradleWrapper.create(distribution, version, ctx);
             } catch (Exception e) {
                 // services.gradle.org is unreachable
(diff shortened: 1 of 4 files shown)


okundzich:
how would this help given that most of our recipes have ranges

bmuschko:
> how would this help given that most of our recipes have ranges

It only helps with use cases where you request of an exact version. I know plenty of open source projects and companies that will not necessarily want to the latest if they call the UpdateGradleWrapper recipe. And as mentioned, it will also reduce the number of calls if you "accidentally" make a call to https://services.gradle.org/.

okundzich:
For this to matter it would require changes all the OSS recipes since most of them use latest constraints. 

What about extracting gradle distribution URL from the gradle property as we discussed in slack?

okundzich:
@sambsnyd please review and merge if looks good. We'll have a follow up PR to look at the URL from gradle wrapper properties as well

PR #5741, opened 2025-07-10
Support GString interpolation in dependency version declaration

bmuschko:
## What's changed?

Supports GString interpolation in a dependency declaration.

```
dependencies {
    implementation (group: "com.google.guava", name: "guava", version: "${guavaVersion}")
}
```

## What's your motivation?

Gradle's dependency declaration notation allows for two different ways to specify the GAV: string notation and map notation. When using Groovy, you can wrap a variable into parenthesis and evaluate it with string interpolation.

Generally speaking, this is a misuse of the Groovy language as you can simply reference the variable by name without wrapping it into a string. However, the language does not prevent you from doing so. This change adds support for it.


bmuschko:
Released with rewrite-gradle [8.58.0](https://github.com/openrewrite/rewrite/releases/tag/v8.58.0). Included in POM of rewrite-java-security [3.15.0](https://central.sonatype.com/artifact/org.openrewrite.recipe/rewrite-java-security/3.15.0).

Issue #5729, opened 2025-07-08
Groovy parsing of script that calls a closure fails with `java.lang.StringIndexOutOfBoundsException`


reftel:
## What version of OpenRewrite are you using?
`org.openrewrite.maven:rewrite-maven-plugin:6.12.1` with org.openrewrite.recipe:rewrite-migrate-java:3.12.0

## How are you running OpenRewrite?

`mvn org.openrewrite.maven:rewrite-maven-plugin:6.12.1:run -Drewrite.activeRecipes=org.openrewrite.java.migrate.UpgradeToJava17 -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:3.12.0
`

## What is the smallest, simplest way to reproduce the problem?
Place this in a file in a repo, run the rewrite above.
```groovy
foo { }()
```

## What did you expect to see?
Nothing

## What did you see instead?
```
[WARNING] There were problems parsing fail.groovy
[INFO] Running recipe(s)...
[WARNING] The recipe produced 1 warning(s). Please report this to the recipe author.
[WARNING] Run with `--debug` or `-Drewrite.exportDatatables=true` to see all warnings.
```

## What is the full stack trace of any errors you encountered?
```
org.openrewrite.groovy.GroovyParsingException: Failed to parse fail.groovy, cursor position likely inaccurate.
    at org.openrewrite.groovy.GroovyParserVisitor.visit (GroovyParserVisitor.java:205)
    at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$3 (GroovyParser.java:147)
    at java.util.stream.ReferencePipeline$3$1.accept (ReferencePipeline.java:197)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining (ArrayList.java:1625)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:509)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:499)
    at java.util.stream.StreamSpliterators$WrappingSpliterator.forEachRemaining (StreamSpliterators.java:310)
    at java.util.stream.Streams$ConcatSpliterator.forEachRemaining (Streams.java:735)
    at java.util.stream.Streams$ConcatSpliterator.forEachRemaining (Streams.java:734)
    at java.util.stream.Streams$ConcatSpliterator.forEachRemaining (Streams.java:735)
[...]

PR #5743, opened 2025-07-11
Fix: handle implicit closure call in Groovy

e5LA:
## What's your motivation?
Fixes https://github.com/openrewrite/rewrite/issues/5729

## What's changed?
Added `boolean implicitCall` which checks if the method that is being invoked (e.g. "call") is not actually present at the current cursor position in the source.

## Any additional context
Groovy allows implicit invocation of closures using syntax like `foo { }()`, which look like `foo({}).call()`. 
In these cases, `visitMethodCallExpression(MethodCallExpression call)` gets `call = "call"` and tries to extract the method name from the source using substring, which caused `StringIndexOutOfBoundsException` - because the `call` identifier is not present in the source.

Another way would be just to hardcode the `call` name, like:
```
if("call".equals(methodNameExpression))
```

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1706,14 +1706,23 @@ public class GroovyParserVisitor {
                 }
 
                 Space prefix = whitespace();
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for the fix here @e5LA ! Looks good already, but I can only review from mobile as both Greg and I are out this weekend. Maybe @sambsnyd can fit in a review and merge?

jkschneider (approved):

PR #5747, opened 2025-07-12
DeleteProperty: remove comments associated with the property deleted

xLitil:
## What's changed?

New property named `removePropertyComments` added to `org.openrewrite.properties.DeleteProperty`

## What's your motivation?

Remove comments associated with the property deleted an avoid to keep this obsoletes comments.

Unit test coverage

I search but found no alternative.

### Checklist

### DIFF
--- a/rewrite-properties/src/main/java/org/openrewrite/properties/DeleteProperty.java
+++ b/rewrite-properties/src/main/java/org/openrewrite/properties/DeleteProperty.java
@@ -56,9 +56,15 @@ public class DeleteProperty extends Recipe {
     @Nullable
     Boolean relaxedBinding;
 
+    @Option(displayName = "Remove property comments",
+            description = "Remove all comments found before the property to removed. By convention, empty line is used to indicate start of property comments",
+            required = false)
+    @Nullable
+    Boolean removePropertyComments;
+
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new PropertiesVisitor<ExecutionContext>() {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the nice suggestion @xLitil ! We're a little hesitant to add ever more boolean options to recipes, so I figured I'd make your suggestion the new default going forward. 

From our docs it seems this recipe is only used sparingly:
https://docs.openrewrite.org/recipes/properties/deleteproperty#used-by

Although I know of at least [one other place in rewrite-spring](https://github.com/openrewrite/rewrite-spring/blob/df1b00cd9269369ffdd96c74d7ebbe63262f3b68/src/main/java/org/openrewrite/java/spring/DeleteSpringProperty.java#L63-L64).

I think it's fine to remove preceding comments as you've helpfully illustrated with your tests in those cases.

Do let me know if you agree with my changes before we merge! Wouldn't want to break anything for you specifically.

xLitil:
Hi @timtebeek,

Thank you for the code review, i'm OK 👍

Can you add some documention, maybe in description, to indicate that this recipe now remove previous comment ?
Example : 
```java
    @Override
        return "Deletes key/value pairs from properties files. By convention, all comments found before the property will be removed, empty line is used to indicate start of property comments";
    }
```

timtebeek:
Quick follow up after seeing failures downstream: https://github.com/openrewrite/rewrite/commit/89b1cca9ce3381845499a31b4b02a074f7c4d0e4

Issue #5702, opened 2025-07-02
Issue discovered on `zuul-core/build.gradle` with `Use `String` notation for Gradle dependency declarations` (DependencyUseStringNotation)


timtebeek:
### Problem
Classifier looks to be dropped.

### Expected behavior
Classifier retained and converted.

### Example diff

**From:** `zuul-core/build.gradle`

```diff
api "io.netty:netty-transport"

implementation "io.netty:netty-codec-haproxy"
-    implementation (group: "io.netty", "name": "netty-transport-native-epoll", "classifier": "linux-x86_64")
-    implementation (group: "io.netty", "name": "netty-transport-native-kqueue", "classifier": "osx-x86_64")
+    implementation ("io.netty:netty-transport-native-epoll")
+    implementation ("io.netty:netty-transport-native-kqueue")

implementation "io.netty.incubator:netty-incubator-transport-native-io_uring:${versions_netty_io_uring}:linux-x86_64"

// We are using the long-form dependency syntax here because we want to
// explicitly set the classifier. We do not have the version number so we can't use
// Gradle's short-form dependency notation.
-    runtimeOnly( group: "io.netty", name: "netty-tcnative-boringssl-static", classifier: "linux-x86_64" )
-    runtimeOnly( group: "io.netty", name: "netty-tcnative-boringssl-static", classifier: "linux-aarch_64" )
-    runtimeOnly( group: "io.netty", name: "netty-tcnative-boringssl-static", classifier: "osx-x86_64" )
-    runtimeOnly( group: "io.netty", name: "netty-tcnative-boringssl-static", classifier: "osx-aarch_64" )
+    runtimeOnly( "io.netty:netty-tcnative-boringssl-static")
+    runtimeOnly( "io.netty:netty-tcnative-boringssl-static")
+    runtimeOnly( "io.netty:netty-tcnative-boringssl-static")
+    runtimeOnly( "io.netty:netty-tcnative-boringssl-static")

implementation 'io.perfmark:perfmark-api:0.27.0'
api 'jakarta.inject:jakarta.inject-api:2.0.1'

```

#### Recipes in example diff:
- `org.openrewrite.gradle.DependencyUseStringNotation`

### References:
[...]

PR #5748, opened 2025-07-12
Improve construction of gradle string notation

aet2505:
All elements apart from name are considered optional in gradle string notation. Enhance the conversion logic to handle null elements

## What's changed?
Update the construction of the string literal object to account for null elements

## What's your motivation?
- Resolves: https://github.com/openrewrite/rewrite/issues/5702
Additionally adds support for extension definitions within the dependencies

Testing of the extension branches was not possible as the test fails to instantiate due to an unknown property on `ExternalModuleDependency` I believe this is a limitation of the test setup rather. Unsure on how to correct this
A steer would be greatly appreciated

N/A

N/A

## Any additional context
Referenced https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.dsl.DependencyHandler.html#N174A0 for notation syntax

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
@@ -149,7 +149,7 @@ public class DependencyUseStringNotation extends Recipe {
             }
 
             private J.@Nullable Literal toLiteral(Space prefix, Markers markers, Map<String, Expression> mapNotation) {
-                //Name is the only required key in a dependency map.
+                // Name is the only required key in a dependency map.
                 if (mapNotation.containsKey("name")) {
                     String group = coerceToStringNotation(mapNotation.get("group"));
                     String name = coerceToStringNotation(mapNotation.get("name"));


rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
```diff
@@ -148,28 +148,40 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
...
-                if (mapNotation.containsKey("group") && mapNotation.containsKey("name")) {
-                    String stringNotation = "";
-
+                //Name is the only required key in a dependency map.
+                if (mapNotation.containsKey("name")) {
                     String group = coerceToStringNotation(mapNotation.get("group"));
+                    String name = coerceToStringNotation(mapNotation.get("name"));
+                    String version = coerceToStringNotation(mapNotation.get("version"));
+                    String classifier = coerceToStringNotation(mapNotation.get("classifier"));
+                    String extension = coerceToStringNotation(mapNotation.get("extension"));
```
shanman190:
```suggestion
                    String extension = coerceToStringNotation(mapNotation.get("ext"));
```

The key for extension is `ext` which is probably why your test was failing.

shanman190 (commented):
I wonder if we should also update the `internal.Dependency` helper as well.
https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2Finternal%2FDependency.java#L44-L57

I did verify the behavior upstream in Gradle, so overall this looks good.
https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ParsedModuleStringNotation.java

aet2505:
> I wonder if we should also update the `internal.Dependency` helper as well. https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2Finternal%2FDependency.java#L44-L57
> 
Thank you for pointing this out, I will get this updated. I'll also update the recipe to use this class & logic to save duplication.
Unless there is a convention to not do this?

> I did verify the behavior upstream in Gradle, so overall this looks good. https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/ParsedModuleStringNotation.java
> 
> NOTE: I think the classifier's colon can be omitted as well when extension is in use. I believe the expectation here is that classifier could be null, but extension defined.

I'll review my cases against the gradle test cases here to avoid any extra colons creeping in https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/test/groovy/org/gradle/api/internal/notations/DependencyStringNotationConverterTest.groovy


Thank you for the review

rewrite-gradle/src/main/java/org/openrewrite/gradle/DependencyUseStringNotation.java
aet2505: Thank you, brain must have failed me on this one! Corrected this and added test case covering extension rewrite

rewrite-gradle/src/test/java/org/openrewrite/gradle/internal/DependencyTest.java
```diff
@@ -0,0 +1,56 @@
...
+        assertEquals("com.example:artifact:1.0.0@jar", dep.toStringNotation());
+    }
+
+    @Test
+    void toStringNotationWithOnlyArtifact() {
+        Dependency dep = new Dependency(null, "artifact", null, null, null);
+        assertEquals(":artifact", dep.toStringNotation());
+    }
+
+}
```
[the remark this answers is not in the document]
aet2505: Assume this wants a new line at end of file and it's got confused? I've applied this change manually.
Equally let me know if I'm just being silly!

rewrite-gradle/src/test/java/org/openrewrite/gradle/internal/DependencyTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Test
+    void toStringNotationWithNullVersionAndClassifierButExtensionPresent() {
+        Dependency dep = new Dependency("com.example", "artifact", null, null, "jar");
+        assertEquals("com.example:artifact@jar", dep.toStringNotation());
+    }
+
+    @Test
+    void toStringNotationWithNullVersionAndExtensionButClassifierPresent() {
```
shanman190: Duplicate of `toStringNotationWithNullVersionButClassifierPresent`.

shanman190 (commented):
We should also mark `group` as `@Nullable` in the `internal.Dependency`.

rewrite-gradle/src/test/java/org/openrewrite/gradle/internal/DependencyTest.java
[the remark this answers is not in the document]
shanman190: Yeah, new line at the end of the file is what it's trying to express here.

shanman190 (approved):

jkschneider (approved):

PR #5750, opened 2025-07-14
Add `MergeYamlDocument` recipe

Laurens-W:
## What's changed?

Added the `MergeYamlDocument` recipe (not sure if this name is very good)

## What's your motivation?

Even when `MergeYaml` is used in combination with a `FindKey` precondition if the key passed to `MergeYaml` is very generic it may touch more `Yaml.Document`s than is desirable.

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlDocument.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlDocument.java
@@ -105,7 +105,6 @@ public class MergeYamlDocument extends Recipe {
     }
 
     final static String FOUND_MATCHING_ELEMENT = "FOUND_MATCHING_ELEMENT";
-    final static String REMOVE_PREFIX = "REMOVE_PREFIX";
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {


rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlDocument.java
```diff
@@ -0,0 +1,149 @@
...
+import lombok.Value;
+import org.intellij.lang.annotations.Language;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
+import org.openrewrite.yaml.MergeYaml.InsertMode;
+import org.openrewrite.yaml.tree.Yaml;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+@NoArgsConstructor(force = true) // TODO: remove with @deprecated constructor
```
jevanlingen, line 29: Do you really need this?

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlDocument.java
```diff
@@ -0,0 +1,149 @@
...
+public class MergeYamlDocument extends Recipe {
```
jevanlingen, line 30: I am wondering, would it not make more sense to add another option to the MergeYam recipe:
```java
@Option(displayName = "Document filter",
  description = "A [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expression used to select a subset of document(s)",
  example = "$.kind[?(@ == 'Deployment')]")
@Nullable String documentFilter;
```

I understand there is a balance between more recipes vs more options. But I have the feeling in this case yet again  another option is the better choice from a user perspective 😇.

Laurens-W: @jkschneider what do you think?

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlDocument.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public String getInstanceNameSuffix() {
+        return String.format("at `%s`", key);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Merge a YAML snippet with an existing YAML document.";
```
Laurens-W, line 104: Perhaps in `MergeYaml` this should be `Merge a YAML snippet with an existing YAML file.` I don't know if that semantic difference is clear to the user though

jkschneider:
We need to seek a more general solution to applying preconditions on a per document level rather than source file.

timtebeek:
@Laurens-W will log an issue for applying preconditions on Yaml Document separate; we'll close this PR.

PR #5758, opened 2025-07-15
Fixing the application of `DoesNotIncludeDependency` against files of the incorrect type.

steve-aom-elliott:
## What's changed?
Reworked the visitor for `DoesNotIncludeDependency` (both versions) so that it will first check that the file would have been accepted by `DependencyInsight`'s visitor before trying to visit it, and then only if an acceptable file was not marked, will it return a found result with that file in it.

## What's your motivation?
Previously there was a bug wherein if you applied `DoesNotIncludeDependency` to a file of a conceptually inapplicable type (i.e. a _Java or Gradle file_ for the **Maven** version or a _Java or POM file_ for the **Gradle** version), it would end up marking the file because the visitor for `DependencyInsight` in each case was being run through the precondition `not`.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
@@ -79,15 +79,10 @@ public class DoesNotIncludeDependency extends Recipe {
 
             @Override
             public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
-                if (!(tree instanceof SourceFile)) {
-                    return tree;
-                }
-                if (di.isAcceptable((SourceFile) tree, ctx)) {
-                    Tree t2 = di.visitNonNull(tree, ctx);
-                    // if Gradle file is unchanged, we found something without the dependency
-                    if (t2 == tree) {
-                        return SearchResult.found(tree);
-                    }
+                Tree t2 = di.visit(tree, ctx);
+                // if Gradle file is unchanged, we found something without the dependency
+                if (t2 == tree) {
+                    return SearchResult.found(tree);
                 }
                 return tree;
             }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/DoesNotIncludeDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/DoesNotIncludeDependency.java
@@ -89,15 +89,10 @@ public class DoesNotIncludeDependency extends Recipe {
 
             @Override
             public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
-                if (!(tree instanceof SourceFile)) {
-                    return tree;
-                }
-                if (di.isAcceptable((SourceFile) tree, ctx)) {
-                    Tree t2 = di.visitNonNull(tree, ctx);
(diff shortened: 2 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
```diff
@@ -68,6 +69,28 @@ public Validated<Object> validate() {
...
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                return di.isAcceptable(sourceFile, ctx);
+            }
+
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (!(tree instanceof SourceFile)) {
+                    return tree;
+                }
+                if (di.isAcceptable((SourceFile) tree, ctx)) {
```
jevanlingen: I don't think you need to do this check here, as it's already been used at the

```java
public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
    return di.isAcceptable(sourceFile, ctx);
}
```

(which is the right place).

Laurens-W:
I'm not sure I understand exactly what the issue is here.
https://github.com/openrewrite/rewrite/blob/41ce44412a1b3312d40526a0f4d048ca4ff74b54/rewrite-core/src/main/java/org/openrewrite/Preconditions.java#L65-L67
`Preconditions.not` marks every sourcefile that is not acceptable, but if the recipes(visitors) that come after that have the appropriate `isAcceptable` checks, that's fine right?

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
steve-aom-elliott: Fair. The `isAcceptable` inside the visitor for `DoesNotIncludeDependency` was a later addition, so I hadn't gone back over it to see if I could cut out more yet.

steve-aom-elliott:
> I'm not sure I understand exactly what the issue is here.
> 
> https://github.com/openrewrite/rewrite/blob/41ce44412a1b3312d40526a0f4d048ca4ff74b54/rewrite-core/src/main/java/org/openrewrite/Preconditions.java#L65-L67
> 
> `Preconditions.not` marks every sourcefile that is not acceptable, but if the recipes(visitors) that come after that have the appropriate `isAcceptable` checks, that's fine right?

The problem was that when it would call the `isAcceptable` for `DependencyInsight`, it would say that all `pom.xml` files for example were fine to work on, but that line you referenced negates the `isAcceptable` and then adds the file to the acceptable list which meant that all non-`pom.xml` files were also fine to work on, which was the problem.

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
steve-aom-elliott: Ended up simplifying it in the last push.

timtebeek (approved):
Looks good; thanks for the fix here! We'll keep an eye on the for now only use downstream:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/366deb6292a0e17daf021268e5acfc7550ce859f/src/main/resources/META-INF/rewrite/junit5.yml#L136-L146

PR #5763, opened 2025-07-16
Add a space between `yield` and its expression

pdelagrave:
## What's changed?
`yield "expression"` in a switch case statement requires a space between `yield` and `"expression"`.

## What's your motivation?
Using a JavaTemplate such as:
```
StringBuilder template = new StringBuilder(
        "Object o = switch (#{any()}) {\n" +
        "  default: yield #{any()};\n" +
        "}");
```
would generate `yield"string"` when using a String and `yield55` when using an int.
Interestingly enough, the `yield"string"` syntax is valid and compiles; whereas `yield55` is an error as expected.

The current "fix" works for the tests but I really didn't put time to think about it and it might be inadequate, the main goal was to create a bug report.

## Any additional context
The bug was first discovered when using JavaTemplate to generate the LST for a switch expression. The test case uses `org.openrewrite.java.format.AutoFormat` directly to keep the test small but `JavaTemplate` eventually calls it too. 

### Checklist

### DIFF
--- a/rewrite-java-17/src/test/java/org/openrewrite/java/SwitchEnhancementsTest.java
+++ b/rewrite-java-17/src/test/java/org/openrewrite/java/SwitchEnhancementsTest.java
@@ -1,0 +1,88 @@
+package org.openrewrite.java;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.Cursor;
+import org.openrewrite.DocumentExample;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

timtebeek:
Thanks for the immediate fix! Good to see this resolved downstream.

jkschneider (approved):

PR #5765, opened 2025-07-17
MinimalViableSpacing missed imports and annotated variable declarations

JohannisK:
## What's changed?
- Added tests for missing cases for 
  - import statements
  - annotated fields with (a) modifier(s)
- Added implementation for fields with annotations
- Added implementation for imports

## What's your motivation?
I think these cases might cause issues in other recipess

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
@@ -234,7 +234,7 @@ public class MinimumViableSpacingVisitor<P> extends JavaIsoVisitor<P> {
     @Override
     public J.Import visitImport(J.Import _import, P p) {
         J.Import i = super.visitImport(_import, p);
-        if(i.getQualid().getPrefix().isEmpty()) {
+        if (i.getQualid().getPrefix().isEmpty()) {
             return i.withQualid(i.getQualid().withPrefix(i.getQualid().getPrefix().withWhitespace(" ")));
         }
         return i;


rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
```diff
@@ -231,6 +249,12 @@ public J.Case visitCase(J.Case _case, P p) {
         return c;
     }
 
+    @Override
+    public J.Import visitImport(J.Import _import, P p) {
+        J.Import i = super.visitImport(_import, p);
+        return i.withQualid(i.getQualid().withPrefix(Space.SINGLE_SPACE));
```
JohannisK: Just noticed I used `Space.SINGLE_SPACE` where in other places in this file the existing prefix is modified.
Is one approach preferable to the other?

Laurens-W: Preferably we modify the whitespace of the existing prefix otherwise we risk removing comments (as those can also be part of a prefix)

JohannisK: I would be very surprised if someone put a comment between ‘import’ and the actual ‘fqn ’ but I’ll change it tomorrow :-)

Laurens-W: You never know haha

knutwannheden: I think we should just guard the modification with an isEmpty() check to be on the safe side.

JohannisK: I added the isEmpty, and used the preexisting prefix

rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
```diff
@@ -174,27 +174,44 @@ public J.Return visitReturn(J.Return return_, P p) {
     public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, P p) {
         J.VariableDeclarations v = super.visitVariableDeclarations(multiVariable, p);
 
-        boolean first = v.getLeadingAnnotations().isEmpty();
+        boolean hasLeadingAnnotations = !v.getLeadingAnnotations().isEmpty();
+        boolean hasModifiers = !v.getModifiers().isEmpty();
+
+        /*
+         * We need at least one space between leading annotations, otherwise we could get a run-on like "@NonNull@Generated@Serial".
+         * Note, this is applicable anywhere that annotations can exist, such as class declarations, etc.
```
jevanlingen:
```suggestion
```

Though it is a lot more work, I think it's better to add this leadingAnnotations-fix to all LST elements with leading annotations, then having this comment here.  To show my point, I added another test to prove this problem does indeed exists for the other LST elements (like this comment says): https://github.com/openrewrite/rewrite/pull/5765/commits/bf7619c5993fa413e0187244b92cf766229d8cf4.

timtebeek: As also commented in Slack: is it even really a problem though? My compiler accepts it, and the minimum vaiable spacing visitor need not look good: it's intended to wipe out formatting before the detected or configured style is applied.

timtebeek (approved):

PR #5768, opened 2025-07-18
MinimalViableSpacing recipe does not handle `extends` correctly

JohannisK:
## What's changed?
- Added test for `implements` and `extends`
- Fixed a bug when dealing with classes extending

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
@@ -379,4 +379,59 @@ class MinimumViableSpacingTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void classExtends() {
+        rewriteRun(
+          spec -> spec.expectedCyclesThatMakeChanges(2),
+          java(
+            """
+              import java.lang.Exception;
+
+              class MyException extends Exception {
+              }
+              """,
+            """
+              import java.lang.Exception;class MyException extends Exception{}
+              """
+          )
+        );
+    }
+
+    @Test
+    void classImplements() {
+        rewriteRun(
+          spec -> spec.expectedCyclesThatMakeChanges(2),
+          java(
+            """
+              import java.io.Serializable;
+
+              class Clazz implements Serializable {
+              }
+              """,
+            """
+              import java.io.Serializable;class Clazz implements Serializable{}
+              """
+          )
+        );
+    }
+
+    @Test
+    void classImplementsSeveral() {
+        rewriteRun(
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
```diff
@@ -75,6 +75,9 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, P
             Space before = c.getPadding().getExtends().getBefore();
             if (before.getWhitespace().isEmpty()) {
                 c = c.getPadding().withExtends(c.getPadding().getExtends().withBefore(before.withWhitespace(" ")));
+                if(c.getExtends().getPrefix().isEmpty()) {
```
timtebeek:
```suggestion
                if (c.getExtends().getPrefix().isEmpty()) {
```

rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
```diff
@@ -379,4 +379,59 @@ class Clazz {
...
+
+              class Clazz implements Serializable, Cloneable{
+              }
+              """,
+            """
+              import java.io.Serializable;import java.lang.Cloneable;class Clazz implements Serializable,Cloneable{}
+              """
+          )
+        );
+    }
```
jevanlingen, line 436: Can you add another test with both extends and implements?

JohannisK:
> Thanks for the quick fixes here!
> 
> I wonder if we can now also remove this oddly added whitespace:
> 
> https://github.com/openrewrite/rewrite/blob/6e66e189f8b35af3748ced6103c78c1bfdf6ef63/rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateJavaExtension.java#L168

It seems this takes care of the space before the `extends` keyword itself.

timtebeek:
> > Thanks for the quick fixes here!
> > I wonder if we can now also remove this oddly added whitespace:
> > https://github.com/openrewrite/rewrite/blob/6e66e189f8b35af3748ced6103c78c1bfdf6ef63/rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateJavaExtension.java#L168
> 
> It seems this takes care of the space before the `extends` keyword itself.

Hmm; that's what [line 77](https://github.com/openrewrite/rewrite/pull/5768/files#diff-8335559d839602b398f6e9e39857610e4626e2154a61b88aec9be0b4f492a46eR77) does as well; guess that's a separate code path then. No need to change that here.

PR #5770, opened 2025-07-18
MinimalViableSpacing does not handle throws correctly

JohannisK:
## What's changed?
- Added test for method that throws
- Fixed implementation for throws

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
@@ -66,6 +66,26 @@ class MinimumViableSpacingTest implements RewriteTest {
     }
 
     @Test
+    void methodWithThrows() {
+        rewriteRun(
+          spec -> spec.expectedCyclesThatMakeChanges(2),
+          java(
+            """
+              import java.lang.Exception;
+              
+              class A {
+                  public <T> void foo() throws Exception {
+                  }
+              }
+              """,
+            """
+              import java.lang.Exception;class A{public <T> void foo() throws Exception{}}
+              """
+          )
+        );
+    }
+
+    @Test
     void returnExpression() {
         rewriteRun(
           spec -> spec.expectedCyclesThatMakeChanges(2),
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
```diff
@@ -138,10 +158,10 @@ class Test {
                   void foo(final int[] arr) {
                       for (int n = 0, x = 0; n < 100; n++, x++) {
                       }
-
+
```
timtebeek: This is sometimes annoying about IntelliJ; a quick change on a different line followed by auto format should clear these out.

timtebeek (approved):
Please revert the needless indentation; otherwise approved already, thanks!

rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
```diff
@@ -250,23 +270,23 @@ void yieldReformatted() {
           spec -> spec.recipe(new AutoFormat()),
           java(
             """
-            class Test {
```
JohannisK, line 253: @timtebeek I left this whitespace change, because it makes the indentation consistent with the rest of the file.

rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
```diff
@@ -65,6 +65,26 @@ class A{public <T> void foo(){}}
...
+            """
+              import java.lang.Exception;
+              
+              class A {
+                  public <T> void foo() throws Exception {
+                  }
+              }
+              """,
+            """
+              import java.lang.Exception;class A{public <T> void foo() throws Exception{}}
```
timtebeek: Small proposal to minimize the elements in this test to focus purely on `throws`.
```suggestion
                  void foo() throws Exception {
                  }
              }
              """,
            """
              import java.lang.Exception;class A{void foo() throws Exception{}}
```

PR #5771, opened 2025-07-18
Show more context in error message if parsing a Groovy source fails

jevanlingen:
## What's changed?
If the Groovy Parser fails to parse a source file, you see more context around the cursor

## What's your motivation?
The current implementation shows 10 chars. This does not give enough intel, especially if the code breaks where some whitespace is going bonkers.

## Any additional context
There is no unit test, because it's hard to create a test. In the ideal world, there is no way you ever face a GroovyParsingException. If you know a case, it is valid Groovy but not parsed well, thus that known issue should be fixed... Creating a unit test with nonsense does not work, as you would get a `groovyjarjarantlr4.v4.runtime.InputMismatchException` error.


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -206,8 +206,9 @@ public G.CompilationUnit visit(SourceUnit unit, ModuleNode ast) throws GroovyPar
                 }
                 throw new GroovyParsingException(
                         "Failed to parse " + sourcePath + " at cursor position " + cursor +
-                        ". The next 10 characters in the original source are `" +
-                        source.substring(cursor, Math.min(source.length(), cursor + 10)) + "`", t);
+                        ". The surrounding characters in the original source are:\n" +
+                        source.substring(Math.max(0, cursor - 250), cursor) + "~~cursor~>" +
+                        source.substring(cursor, Math.min(source.length(), cursor + 250)) + "\n", t);
```
timtebeek: I'm not sure I'm fully comfortable with this type of change; if there's not enough context we'd rather ask folks to attach a debugger or share the failing source file to do so ourselves.

jevanlingen: Guess we can find the middle ground, by trimming it down to say, 80 chars?

greg-at-moderne: To be honest, I am not sure why putting more source in the error message would be a problem for OpenRewrite.

I suppose the concern is about Moderne SaaS product. And making sure error messages are not leaking (potentially) proprietary source code. If that's the case, I suppose this leak prevention should happen on Moderne side?

timtebeek:
Ok so from talking over the options, the underlying issue is that we have a data table export that shows differing levels of detail for various failures, based on 
https://github.com/openrewrite/rewrite/blob/6f8e2ed66f33d08727d5d6cb0eb5ad0e621aa005/rewrite-core/src/main/java/org/openrewrite/FindParseFailures.java#L106-L156

In the case of print idempotency issues we show a diff of the problematic lines;
https://github.com/openrewrite/rewrite/blob/0067a102176930469266bfe0b06ba37669a6bc76/rewrite-core/src/main/java/org/openrewrite/Parser.java#L43-L56

In the case of `GroovyParsingException` we for now only show the tiniest snippet.
https://github.com/openrewrite/rewrite/blob/615360e91282cb1563db9ddd688c72595fd68734/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java#L157-L159

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
timtebeek: I wonder if it makes sense to include the `entry.getKey().getLine()` and `entry.getKey().getColumn()` here on such parse failures; those would be easier to correlate for folks than a cursor position.

jevanlingen: You could include the line number and column if you really want to report things thoroughly, but to be honest I don't think it's strictly necessary. What's usually most helpful is a snippet of the code where the parser started to go off track.

Once the parser loses its footing, the cursor position can already be somewhat misleading: it might skip over characters or fail to skip when it should, which throws off its accuracy. So if the message mentions where the error roughly occurs (with both some code prior and after the cursor), that's most of the time enough context to start fixing the issue.

timtebeek: Yeah I'm wondering mostly if instead of the cursor, we should be using the `ASTNode` which often has both `getText()` as well as `toString()` to show more details where needed. I think that might be the better change here particularly since we know our parser is already struggling when we reach this point.

jevanlingen: I don't agree, because the source code is sometimes very different than the Groovy AST node, as the AST does not contain whitespace, braces and all the other goodness the LST does. You really want to see the source code (the raw input) instead of the intermediate representation of the compiler.

timtebeek: How about a smaller adjustment from what we're coming from: showing the next ten non-whitespace characters.
```java
                Matcher matcher = Pattern.compile("\\s*(.{0,10})").matcher(source.substring(cursor));
                int end = matcher.find() ? matcher.end() : Math.min(source.length(), cursor + 10);
                throw new GroovyParsingException(
                        "Failed to parse " + sourcePath + " at cursor position " + cursor +
                        ". The next 10 characters in the original source are `" +
                        source.substring(cursor, end) + "\n", t);
```
I think that would still help pin point the issue in the case of let's say indentation and newlines, without capturing a lot more than we had originally set out to do here.

jevanlingen: A little bit better! But still not good enough if you ask me 😇. Like I said before, you need context from before and after the cursor, as the cursor can be off once a GroovyParsingException occurs. Furthermore, ten characters is in most cases too low to get proper information.

Let me give you an example to make my point clearer. Recently we added method references support to the parser. Before this addition, you would have faced a GroovyParsingException. So let's say you would have written this code:

```groovy
["a", "b", "c"].stream()
    .map(String::getMetaPropertyValue)
    .toList()
```

As method references back then could not be handled, the cursor would already be passed the `::` operator, because it would go into a different code path (and then ultimately error out). Thus the next ten characters would probably have been something like `getMetaPro`. With only this information being shown as failure, the error message does not make any sense if you don't have the source file at hand. 

So my point:
At least take the previous ten non white characters. But to do it properly, you want even more characters to be sure you catch enough information. Or remove this error all together if you don't want to expose this information in the first place.

timtebeek (approved):

PR #5773, opened 2025-07-18
Adding `exclusions` and `applyTo` to `CoalesceProperties`

steve-aom-elliott:
WiP - Feeling like the way it checks the `applyTo` is wrong. Either that or I'm just struggling a bit with what the appropriate JSONPath syntax should be for certain scenarios.

### Checklist


jevanlingen:
Follow up:
- https://github.com/openrewrite/rewrite/pull/5824

PR #5774, opened 2025-07-21
Method references for Groovy Parser

jevanlingen:
## What's changed?
Method references are supported.

## What's your motivation?
Expanding support for the Groovy Parser.

### Checklist


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1877,6 +1876,33 @@ public void visitStaticMethodCallExpression(StaticMethodCallExpression call) {
...
+            Space fmt = whitespace();
+
+            String referenceName = null;
+            if (ref.getMethodName() instanceof ConstantExpression) {
+                referenceName = ((ConstantExpression) ref.getMethodName()).getValue().toString();
+            }
+
+            queue.add(new J.MemberReference(randomId(),
+                    fmt,
+                    isMethodRef ? Markers.EMPTY : Markers.build(singleton(new MethodPointer(randomId()))),
```
timtebeek: While I agree with the change this addition of a new marker likely makes this a core change that needs further review.

jevanlingen: We could also break it up in two PRs. So this PR would only provide support for method references, and the other (still to be created) PR which add support for method pointers with the new marker.

timtebeek: I think that would indeed help have some immediate value now, and not block on a longer review

PR #5777, opened 2025-07-21
Catch block should support modifiers

jevanlingen:
## What's changed?
The `final` keyword can be used.

## What's your motivation?
Support for:

```groovy
try {
} catch (final RuntimeException e) {
}
```

### Checklist

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/TryTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/TryTest.java
@@ -31,13 +31,13 @@ class TryTest implements RewriteTest {
           groovy(
             """
               try {
-               
+              
               } catch (RuntimeException e) {
-               
+              
               } catch (Exception e) {
-                           
+              
               }
-               """
+              """
           )
         );
     }
@@ -51,6 +51,19 @@ class TryTest implements RewriteTest {
               
               } catch (all) {
               
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void modifier() {
+        rewriteRun(
+          groovy(
+            """
+              try {
+              } catch (final RuntimeException e) {
               }
               """
           )
@@ -84,7 +97,7 @@ class TryTest implements RewriteTest {
              } finally {
                  def a = ""
              }
-              """
+             """
           )
         );
     }


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/TryTest.java
```diff
@@ -57,6 +57,19 @@ void catchOmittingType() {
         );
     }
 
+    @Test
+    void modifier() {
+        rewriteRun(
+          groovy(
+            """
+              try {
+              } catch (  final       RuntimeException                           e     ) {
```
MBoegers: why that many blanks?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1108,6 +1108,7 @@ public void visitBlockStatement(BlockStatement block) {
         public void visitCatchStatement(CatchStatement node) {
             Space prefix = sourceBefore("catch");
             Space parenPrefix = sourceBefore("(");
+            List<J.Modifier> modifiers = getModifiers();
```
MBoegers, line 1111: We don't have to guard here against any non `final` (`static` a.o.) modifier because the Groovy toolchain would reject the code as invalid beforehand, right?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1133,8 +1134,8 @@ public void visitCatchStatement(CatchStatement node) {
             cursor += param.getName().length();
             Space rightPad = whitespace();
             skip(")");
-            JRightPadded<J.VariableDeclarations> variable = JRightPadded.build(new J.VariableDeclarations(randomId(), paramType.getPrefix(),
-                    Markers.EMPTY, emptyList(), emptyList(), paramType.withPrefix(EMPTY),
+            JRightPadded<J.VariableDeclarations> variable = JRightPadded.build(new J.VariableDeclarations(randomId(), EMPTY,
+                    Markers.EMPTY, emptyList(), modifiers, paramType,
```
MBoegers: why not call `getModifiers` directly? any further plans with the list from [line 1111](https://github.com/openrewrite/rewrite/pull/5777/files#diff-d900151f599dd809ec1076aca6e0ecb67997b3b5b1e52e435d1c64de318fd834R1111)

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/TryTest.java
jevanlingen: Just to test if all the prefixes are put where they are supposed to be. I can also break it up in two tests if you like that more.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: Because you need to take the whitespace into account. The modifiers come first in the source code, thus you need to call them at first.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: Yes!

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
MBoegers: Yes you are right; forgot the stateful nature of the parser here. Maybe add a brief comment for futur 🧙

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/TryTest.java
MBoegers: I would indeed like, as they are testing 2 different nuances of the same feature and the of cause perfectly choosen name will document it ;)

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: I don't think that's needed. The `getModifiers` call has already been used several times before like this. As we now have the unit tests in place, as soon as you think you are smart and inline them, the unit will break.

At the moment we mostly write comments when the Groovy AST is so different than expected, that we add an explanation.

timtebeek: Hmm; we have a convention of adding whitespace to the outer-most element; here it seems you're moving a prefix that on main is on the variable declarations into the param type. I'd prefer to keep the prefix on the variable declarations, which presumably here means you need to move the prefix from either the modifiers or the paramtype to the outer variable declarations.

jevanlingen: Sure: https://github.com/openrewrite/rewrite/pull/5777/files/ce782be9017a1cb1038ed61c4594d805c6e65eed..1afea0cf84e6f78bbd253caf6a05b91a36d92583!

PR #5778, opened 2025-07-21
Parse dynamic method invocation

jevanlingen:
## What's changed?
Dynamic method calls are _parsed_ as well.

## What's your motivation?
The parser could not handle dynamic method calls like:

```groovy
def xMethod() {}
def prefix = "x"
"${prefix}Method"()
```

## Any additional context
This is more of a workaround than a real fix, as dynamic methods are now stringified and then used as method name. The parameter(s) within the GString will not be visited when used in a visitor, as that information will not be available in the LST (it's just transformed to text). Still I think this is one step forward, as source files that use dynamic methods can now at least be parsed.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -504,8 +504,8 @@ public class GroovyParserVisitor {
                 methodName = method.getName();
             } else {
                 // Method name might be in quotes
-                char openingQuote = source.charAt(cursor);
-                methodName = openingQuote + method.getName() + openingQuote;
+                String delim = source.charAt(cursor) + "";
+                methodName = sourceSubstring(cursor, delim) + delim;
             }
             cursor += methodName.length();
             J.Identifier name = new J.Identifier(randomId(),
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/MethodInvocationTest.java
```diff
@@ -198,6 +198,19 @@ void useClassAsArgumentJavaStyle() {
...
     }
 
+    @Test
+    void dynamicMethodInvocation() {
+        rewriteRun(
+          groovy(
+            """
+              def xMethod() {}
+              def prefix = "x"
+              "${prefix}Method"()
```
timtebeek, line 208: Should we add a test case with an escaped delimiter? I could see that happen with say spock test method names.

jevanlingen: Sure: https://github.com/openrewrite/rewrite/pull/5778/commits/54862c7066f24a5d6aaa0ee7b76d3e9f349a592d! Funnily enough, we did not yet support this either 😄.

greg-at-moderne (approved):

PR #5780, opened 2025-07-21
Null as Identifier in HCL

MBoegers:
## What's changed?
`null` is a valid key for objects and variables in HCL. We blocked it as syntactically error.

## What's your motivation?
allow basic setups as
```hcl
terraform {
  required_providers {
    null = {
      source = "hashicorp/null"
      version = "=3.2.3" 
    }
  }
}
```

`null` seems to be common in TF

### Checklist


MBoegers:
I know we usually don't force push, I just neede to get the PR displaying right

rewrite-hcl/src/main/antlr/HCLParser.g4
```diff
@@ -19,7 +19,7 @@ bodyContent
     ;
 
 attribute
-    : Identifier ASSIGN expression
+    : (Identifier | NULL) ASSIGN expression
     ;
```
timtebeek, line 23: Somewhat surprised not to see any corresponding change in the tree package here around:
https://github.com/openrewrite/rewrite/blob/05795eb525099812887796b5d0a73eda6d7f6673/rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java#L65

Is that just a problem with my expectations? Could you help explain why such a change wasn't needed?

MBoegers: Token `NULL` here is not the OOP `null` as no object. In the G4 the `NULL` is a special identifier with the value `"null"` and will be parsed into a `Hcl.Identifier` with the `name="null"`.

The `NULL`-Token is only used in 2 locations to selectively allow `"null"` as an identifier in these locations (as object keys and variable names). This way we still reject corrupt files at the parser level when `null` is used in a different location or no identifier is provided.
 
You see no change in the `Hcl.Tree` because none is needed to represent an `Hcl.Identifier` with value `"null"`. And I choose the minimal change.
We could think about mirroring this setup with a specialization of `Hcl.Identifier` to explicitly model `"null"` as an identifier in these cases. 
I think the gain would be very low, as null is simply valid in these cases, and the risk of introducing new LST elements is higher.

For example:

```hcl
terraform {
required_providers {
  null = {
    source = "hashicorp/null"
    version = "=3.2.3"
  }
}
```

is parsed into (sorry for no TreeVisitingPrinter):
grafik

MBoegers: While writing these lines, I realize that `NULL` is probably not a great name for the Token.

MBoegers:
agree they look strange let me reset to main

MBoegers:
@greg-at-moderne the answer is, I executd the `./gradlew rewrite-hcl:generateAntlrSources` task which also generated the HCL internal JSON grammar. rolled back because its a disturbing change

timtebeek (approved):

Issue #4061, opened 2024-02-29
Groovy Parser does not support method reference operator ::


blipper:
I am using

- OpenRewrite Latest

## How are you running OpenRewrite?
Gradle

## What is the smallest, simplest way to reproduce the problem?

```
test.gradle

final String[] finalDeps = 
            List.of(new Object()).stream().toArray(String[]::new);

```

## What is the full stack trace of any errors you encountered?

```
org.openrewrite.groovy.GroovyParsingException: Failed to parse buildSrc/src/main/groovy/test.gradle at cursor position 93. The next 10 characters in the original source are `;
`
        at org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:174)
        at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$4(GroovyParser.java:154)
        at java.base@17.0.10/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.
```

## Are you interested in [contributing a fix to OpenRewrite](
Yes

shanman190:
Hi @blipper! 

This one is going to be a little difficult to achieve, but it's possible if we perform a similar thing to what has been done for the major Java versions so far. 

Presently, the rewrite-groovy project is built against Groovy 3.0 which is backwards compatible at the AST level with Groovy 2.x. However, to handle the Java 8 syntax requires the Groovy 4.0 "Parrot" parser. Since the migration of the packages was wrapped up in the 4.0 release as the project moved under the Apache Foundation this means the the AST itself wasn't backwards compatible and would have to be built back up using the new types and compiler.

knutwannheden:
Interesting. I looked at the Groovy 4 release notes, which mention quite a few new language features, but nothing about method references. Qas that maybe already part of Groovy 3?

shanman190:
@knutwannheden, hmm... thats interesting, but you are right. I could have sworn that it was a Groovy 4 thing for method references, but it is actually available in Groovy 3. So that shouldn't be too bad to add in then.

knutwannheden:
We should here also cover the "method pointer" syntax `.&` which predates the `::` syntax but has the same semantics. In both cases we can probably create a `J.MemberReference` and for `.&` we could then just add a dedicated marker to it.

PR #5782, opened 2025-07-22
Support for method pointers and array constructor method references

jevanlingen:
## What's changed?
- Method pointer operator is supported (`.&`)
- Array constructor method references are supported

## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/4061

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -848,8 +848,12 @@ public class GroovyParserVisitor {
 
         @Override
         public void visitClassExpression(ClassExpression clazz) {
-            Space prefix = whitespace();
             ClassNode type = clazz.getType();
+            if (type.getTypeClass().isArray()) {
+                queue.add(arrayType(clazz.getType()));
+                return;
+            }
+            Space prefix = whitespace();
             String name = type.getNameWithoutPackage().replace('$', '.');
             if (!source.startsWith(name, cursor)) {
                 name = type.getUnresolvedName().replace('$', '.');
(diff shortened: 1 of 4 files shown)


jevanlingen:
@jkschneider Added you for review, as this PR also brings in another marker.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1878,21 +1882,22 @@ public void visitStaticMethodCallExpression(StaticMethodCallExpression call) {
...
             queue.add(new J.MemberReference(randomId(),
                     whitespace(),
-                    Markers.EMPTY,
-                    padRight(visit(ref.getExpression()), sourceBefore("::")),
+                    isMethodRef ? Markers.EMPTY : Markers.build(singleton(new MethodPointer(randomId()))),
+                    padRight(visit(ref.getExpression()), sourceBefore(isMethodRef ? "::" : ".&")),
                     null, // not supported by Groovy
                     padLeft(whitespace(), new J.Identifier(randomId(),
-                            sourceBefore(referenceName),
+                            sourceBefore(name),
```
greg-at-moderne, line 2005: I know this was also the case before your changes, but does `sourceBefore()` even support the case when the argument is null?
And if not, we should handle it better when the `name` is null, i.e. when it's not a `ConstantExpression`. Not sure what else could be provided here.

jevanlingen: Good one, I could easily solve this by using the existing superclass method `getText()`: https://github.com/openrewrite/rewrite/pull/5782/commits/31f0cda5cfd54d5ecec94ed33110d58a820c8d2e!

greg-at-moderne (approved):

timtebeek:
There's an alternative implementation not using a marker but using a new LST element `MethodPointer`:
- https://github.com/openrewrite/rewrite/pull/6501

Elsewhere we've tended to reuse existing LST elements where possible to avoid needing dedicated visitors when it comes to recipes like ChangeMethodName.

Can we get this reviewed and merged to avoid further confusion?

PR #5783, opened 2025-07-22
Add support for diamond operator in Groovy Parser

jevanlingen:
## What's changed?
Diamond operators (`<>`) are now correctly supported when creating new generic instances like `new ArrayList<>()`.

## What's your motivation?
Previously, using `new ArrayList<>()` would incorrectly render as `new ArrayList<>()>()`, due to missing type inference information.

## Any additional context
We did already support `new HashMap<>()`; this was only because the Groovy compiler happened to set the `StaticTypesMarker.INFERRED_TYPE` flag in that case. However, this flag is not consistently applied. This change ensures support even when the flag is missing by also checking if the generic type list is empty.

Interestingly, when the flag is present, the generic types are populated, which means we need to handle both cases: either rely on the flag or detect when the generics are empty.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -3095,8 +3095,7 @@ public class GroovyParserVisitor {
 
         ClassNode innerClass = ctor.getType();
         if (!(innerClass instanceof InnerClassNode)) {
-            GenericsType[] generics = innerClass != null ? innerClass.getGenericsTypes() : null;
-            return generics != null && generics.length == 0;
+            return innerClass.getGenericsTypes() != null && innerClass.getGenericsTypes().length == 0;
         }
 
         ClassNode superClass = innerClass.getUnresolvedSuperClass();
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2516,15 +2516,10 @@ private <T extends TypeTree & Expression> T typeTree(@Nullable ClassNode classNo
                     && classNode.getUnresolvedSuperClass().isUsingGenerics()
                     && !classNode.getUnresolvedSuperClass().isGenericsPlaceHolder()
                     && classNode.getGenericsTypes() == null;
-            if (isAnonymousClassWithGenericSuper) {
+            if (isAnonymousClassWithGenericSuper || (classNode.isUsingGenerics() && !classNode.isGenericsPlaceHolder())) {
```
jevanlingen, line 2519: ℹ️ This change is just a small refactor. The implementation was completely the same, except for the argument of the `visitTypeParameterizations` call. I figured it's way easier to understand if we collapse these two branches into one.

PR #5785, opened 2025-07-22
refactor: Inline variable

timtebeek:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.InlineVariable?organizationId=ODQ2MGExMTUtNDg0My00N2EwLTgzMGMtNGE1NGExMTBmZDkw

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -42,7 +42,6 @@ import java.util.function.BiFunction;
 import java.util.function.UnaryOperator;
 
 import static java.util.Collections.unmodifiableList;
-import static java.util.Objects.requireNonNull;
 import static org.openrewrite.ExecutionContext.SCANNING_MUTATION_VALIDATION;
 import static org.openrewrite.Recipe.PANIC;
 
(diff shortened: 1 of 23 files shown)


timtebeek:
@sambsnyd this is a bit of a stylistic preference, but it comes from running `InlineVariable` which now also support local method variable assignments. I found it frequently pops up in code reviews where folks do a quick last assign before a return, which of course is a bit silly and helps to clear out.
- https://github.com/openrewrite/rewrite-static-analysis/pull/647

It's also enforced through rewrite.yml and review bot, so if we opt not to merge this one we should comment out the recipe here:
https://github.com/openrewrite/rewrite/blob/f1dd96fd5651736a856f9bcf77168ac43a6d0f6a/rewrite.yml#L32

greg-at-moderne (approved):
I agree this might be a matter of taste. So let me share my take on this based on scrolling through this PR:
- I can imagine in some very specific cases, the name given to the variable can be really valuable and important, thus removing it might be negative value, but these cases would be super rare and I think we can sacrifice them,
- most (if not all) of the removed lines in the PR are just purely mechanical and I like the change,
- I can also empathise with someone who might prefer the style of keeping one variable to work with through the whole code of the method, and then return the variable. And in that sense the change is slightly moving away from this structure. And this is what I guess might be questioned by some. I like it though and see no problem here.

timtebeek:
Great thanks! I'll merge this after the release such that we have a little time to shake things out, even if I don't expect any issues.

PR #5787, opened 2025-07-22
Groovy Parser supports negative literals with trailing zeros

jevanlingen:
## What's changed?
Even better support for numeric literals.

## What's your motivation?
Literals like `-0.10f` were printed as `-0.1,f`.

## Any additional context
To support numbers with special cases like underscores or trailing zeros, we used the metadata flags like `_FLOATING_POINT_LITERAL_TEXT` to get the correct text. Sadly enough, this flag is not produced by the Groovy compiler for negative numbers. So I removed the dependency on those flags and used the source code itself to get the actual text.

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1330,23 +1330,11 @@ public class GroovyParserVisitor {
                     jType = JavaType.Primitive.Char;
                 } else if (type == ClassHelper.double_TYPE || Double.class.getName().equals(type.getName())) {
                     jType = JavaType.Primitive.Double;
-                    if (expression.getNodeMetaData().get("_FLOATING_POINT_LITERAL_TEXT") instanceof String) {
-                        text = (String) expression.getNodeMetaData().get("_FLOATING_POINT_LITERAL_TEXT");
-                    }
                 } else if (type == ClassHelper.float_TYPE || Float.class.getName().equals(type.getName())) {
                     jType = JavaType.Primitive.Float;
-                    if (expression.getNodeMetaData().get("_FLOATING_POINT_LITERAL_TEXT") instanceof String) {
-                        text = (String) expression.getNodeMetaData().get("_FLOATING_POINT_LITERAL_TEXT");
-                    }
                 } else if (type == ClassHelper.int_TYPE || Integer.class.getName().equals(type.getName())) {
                     jType = JavaType.Primitive.Int;
-                    if (expression.getNodeMetaData().get("_INTEGER_LITERAL_TEXT") instanceof String) {
-                        text = (String) expression.getNodeMetaData().get("_INTEGER_LITERAL_TEXT");
-                    }
                 } else if (type == ClassHelper.long_TYPE || Long.class.getName().equals(type.getName())) {
-                    if (expression.getNodeMetaData().get("_INTEGER_LITERAL_TEXT") instanceof String) {
-                        text = (String) expression.getNodeMetaData().get("_INTEGER_LITERAL_TEXT");
-                    }
                     jType = JavaType.Primitive.Long;
                 } else if (type == ClassHelper.short_TYPE || Short.class.getName().equals(type.getName())) {
                     jType = JavaType.Primitive.Short;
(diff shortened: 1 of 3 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1375,20 +1363,23 @@ public void visitConstantExpression(ConstantExpression expression) {
...
-                }
-                cursor += text.length();
-                // Numeric literals may be followed by "L", "f", or "d" to indicate Long, float, or double respectively
-                if (jType == JavaType.Primitive.Long || jType == JavaType.Primitive.Float || jType == JavaType.Primitive.Double) {
-                    if (source.startsWith("L", cursor) || source.startsWith("f", cursor) || source.startsWith("d", cursor)) {
-                        text += source.charAt(cursor);
-                        cursor++;
+                // Get the string literal from the source, as numeric literals may have a unary operator, underscores, dots and can be followed by "L", "f", or "d"
+                if (jType == JavaType.Primitive.Int || jType == JavaType.Primitive.Long || jType == JavaType.Primitive.Float || jType == JavaType.Primitive.Double) {
+                    int i = cursor;
```
jevanlingen, line 1368: ℹ️ The piece of code is conceptually equal the the name function:
https://github.com/openrewrite/rewrite/blob/25523e727446c3563cbdaf1590088218d05a4032/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L2823-L2835

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -1375,20 +1363,23 @@ public void visitConstantExpression(ConstantExpression expression) {
...
-                    text = "+" + text;
-                }
-                cursor += text.length();
-                // Numeric literals may be followed by "L", "f", or "d" to indicate Long, float, or double respectively
-                if (jType == JavaType.Primitive.Long || jType == JavaType.Primitive.Float || jType == JavaType.Primitive.Double) {
-                    if (source.startsWith("L", cursor) || source.startsWith("f", cursor) || source.startsWith("d", cursor)) {
-                        text += source.charAt(cursor);
-                        cursor++;
+                // Get the string literal from the source, as numeric literals may have a unary operator, underscores, dots and can be followed by "L", "f", or "d"
+                if (jType == JavaType.Primitive.Int || jType == JavaType.Primitive.Long || jType == JavaType.Primitive.Float || jType == JavaType.Primitive.Double) {
```
greg-at-moderne, line 1367: Short and Byte too?
Not sure about char, probably fine to leave it out.

greg-at-moderne (approved):

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
jevanlingen: > Short and Byte too?

Nope, because you can't actually write it like that. For example:

```groovy
def a = 0b0010_1010 as byte
byte b = 0b0010_1010
byte c = 10
```

Are all at ints for the ConstantExpression:
```
Type: int
Value: 42
----
Type: int
Value: 42
----
Type: int
Type: 10
```

Same applies to the short. So these are already taking into account, because `jType == JavaType.Primitive.Int` is one of the conditions! To give a little more context, this is why (ChatGPT):

```
The JVM itself does not have literal representations for byte or short in the .class file. All numeric literals (except for long, float, double) are treated as int:

A byte or short literal in source code is compiled into a constant int value.
When the bytecode needs to store that into a byte or short variable or field, it includes a cast operation (i2b or i2s).
So even though you write byte b = 42;, the constant 42 is an int, and it's cast at runtime to byte.
This means the JVM cannot distinguish the literal 42 in byte b = 42 from the literal 42 in int i = 42.
```

> Not sure about char, probably fine to leave it out.

Char is a special case, as it is also a int at JVM level, but Groovy considers them as a String!

---

So I think in the end the:

```java
 } else if (type == ClassHelper.byte_TYPE) {
   jType = JavaType.Primitive.Byte;
} else if (type == ClassHelper.char_TYPE) {
   jType = JavaType.Primitive.Char;
} else if (type == ClassHelper.short_TYPE || Short.class.getName().equals(type.getName())) {
   jType = JavaType.Primitive.Short;
}
```

could even be removed from the parser. But let's keep 'm there, just in case I am missing something 😅.

PR #5789, opened 2025-07-22
Add Flag.Enum to allow identifying enum constants from other members in a JavaType.Class of Enum Kind

pdelagrave:
## What's changed?
A new flag `Enum` in `org.openrewrite.java.tree.Flag`.

## What's your motivation?
For the new `SwitchCaseAssigningToSwitchExpression` [recipe being developed](https://github.com/openrewrite/rewrite-migrate-java/pull/795), it's necessary to be able to tell if a switch statement applied to an enum is "exhaustive". This implies being able to tell if items of `JavaType.Class.members` are an [enum constant](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-EnumConstant) or not.

Without the present change, there's no way to tell if a member is an [enum constant](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-EnumConstant) or a normal class field in the enum.

* Would that change have an unforseen impact? `JavaType.Variable#getFlagsBitMap()` allows getting direct access to the bitmap but the "unknown" flags are zeroed when a `Variable` is instantiated so no existing code should be using/expecting anything out of bit 14 already.

@jkschneider @timtebeek 

There doesn't seem to be a way to get that information otherwise.

## Any additional context
* When visiting the enum declaration in the tree it's very clear which fields are enum constant or not, but the `JavaType.Class.members` items don't contain that information.
* At [type mapping time](https://github.com/openrewrite/rewrite/blob/45c438972d37db6b1f1903c87c752e141e4cc997/rewrite-java/src/main/java/org/openrewrite/java/internal/JavaReflectionTypeMapping.java#L290-L290) the flags (modifiers but contains more than just the modifiers flags) for the field as passed to construct a `JavaType.Variable` but the ['unknown' flags are zeroed](https://github.com/openrewrite/rewrite/blob/45c438972d37db6b1f1903c87c752e141e4cc997/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L1625-L1625), which includes the `Enum` flag.
* The [downstream change waiting for that fix](https://github.com/openrewrite/rewrite-migrate-java/pull/795/commits/c29b44c6f3760731bc777c9c1e985f7d81e764f8)

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/tree/Flag.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/tree/Flag.java
@@ -52,7 +52,7 @@ public enum Flag {
 
     private final long bitMask;
 
-    public static final long VALID_CLASS_FLAGS = Stream.of(Public, Private, Protected, Static,  Final, Interface, Abstract, Strictfp)
+    public static final long VALID_CLASS_FLAGS = Stream.of(Public, Private, Protected, Static,  Final, Interface, Abstract, Strictfp, Enum)
             .map(Flag::getBitMask).reduce(0L, (m1, m2) -> m1 | m2);
     public static final long VALID_FLAGS = Arrays.stream(Flag.values())
             .map(Flag::getBitMask)
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/EnumTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/EnumTest.java
@@ -230,4 +230,18 @@ class EnumTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void enumClassHasJavaTypeClassEnumFlagEnabled() {
+        rewriteRun(
+          java(
+            """    
+              enum Color {}
+              """,
+            spec -> spec.afterRecipe(cu -> {
+                assertTrue(requireNonNull(cu.getClasses().get(0).getType()).hasFlags(Flag.Enum));
+            })
+          )
+        );
+    }
 }


rewrite-java/src/main/java/org/openrewrite/java/tree/Flag.java
```diff
@@ -33,6 +33,10 @@ public enum Flag {
     Interface(1L << 9),
     Abstract(1L << 10),
     Strictfp(1L << 11),
+    /**
+     * An enumeration type (on a class) or an enumeration constant (on a member)
+     */
+    Enum(1L << 14),
```
sambsnyd, line 39: Should also be added to the `VALID_CLASS_FLAGS` field below

pdelagrave: Right, that'd be more consistent, although the Class' Kind is set using the flag so that information is already available for a Class: https://github.com/openrewrite/rewrite/blob/45c438972d37db6b1f1903c87c752e141e4cc997/rewrite-java/src/main/java/org/openrewrite/java/internal/JavaReflectionTypeMapping.java#L123-L139

sambsnyd: Does the java compiler internally place that flag on enum classes? If so we should list it in our valid flags map. Otherwise leave it omitted.

timtebeek: If I set a breakpoint at line 420 here, then indeed it seems that the bit is set for enum classes as well, as `flagsBitMap & (1L << 14) == 16384`
https://github.com/openrewrite/rewrite/blob/1f4436d4e93633dcbc0fdbb3f4dacbbe096fee57/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L415-L420

That would indeed hint at adding the `Enum` flag to `VALID_CLASS_FLAGS`
https://github.com/openrewrite/rewrite/blob/1a1690fbf53c40d4e092f603c2e02a3fac857258/rewrite-java/src/main/java/org/openrewrite/java/tree/Flag.java#L55-L56

timtebeek: Approved and ok to merge if indeed we add the `Enum` flag to `VALID_CLASS_FLAGS`, and update the test to assert it's there, just to guard against regression.

pdelagrave: Yes, `java.lang.Class#getModifiers()` bit field has the `Enum` flag enabled when the class is an enum.
Just added it with a test: https://github.com/openrewrite/rewrite/pull/5789/commits/bab84bfd16ff76dc2658b10df667bf04cb8a620f

timtebeek (approved):
Thanks! Nice find to make things easier downstream.

PR #5793, opened 2025-07-23
Handling cases for `AddCommentToMethodInvocations` where single line comments would break code and simplifying recipe.

steve-aom-elliott:
## What's changed?
- Will always write out a single-line block comment `/* ... */`, flattening newlines into ` * `

## What's your motivation?
Prior to this simplification, if you were to request a single line comment on a method invocation that was being assigned to something, it would end up commenting out the entire method invocation by mistake.

### Checklist


steve-aom-elliott:
I'm not sure why it's yelling at me about dozens of other unrelated files. Did I do the formatting wrong on the ones I did change and it thinks it needs to align the other files?

PR #5791, opened 2025-07-23
Revert "Support for external properties for UpgradeTransitiveDependencyVersion recipe (#5756)"

jevanlingen:
## What's changed?
Revert https://github.com/openrewrite/rewrite/pull/5756.

## What's your motivation?
Downstream failures.

PR #5756, opened 2025-07-15
Support for external properties for UpgradeTransitiveDependencyVersion recipe

jevanlingen:
## What's changed?
The UpgradeTransitiveDependencyVersion recipe does support properties:
- defined in a properties file
- defined in the `ext` block in the settings.gradle file (Groovy DSL only)

## What's your motivation?
Improvement of thie recipe, just like https://github.com/openrewrite/rewrite/commit/c323d4cced0815c17dbcd9c34af34300c924c81c did for the UpgradeDependencyVersion recipe.

## Additional context
At one point we should consider to either introduce utility methods or merge this recipe with UpgradeDependencyVersion, as both recipes have a lot of common code. For now I just improved this recipe, so it does support external properties without relying on other components.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #5794, opened 2025-07-24
Support for external properties for UpgradeTransitiveDependencyVersion recipe (part 2)

jevanlingen:
## What's changed?
The UpgradeTransitiveDependencyVersion recipe does support properties:
- defined in a properties file
- defined in the `ext` block in the settings.gradle file (Groovy DSL only)

## What's your motivation?
- Improvement of the recipe
- https://github.com/openrewrite/rewrite/pull/5791

## Why part 2?
The previous version, https://github.com/openrewrite/rewrite/pull/5756, was reverted because it failed to handle certain cases correctly.  The core issue lies in how downstream recipes can use its accumulator with uninitialised fields:

```java
new UpgradeTransitiveDependencyVersion("", "", "", null, null, null).getInitialValue(ctx);
```

As a result, this leads to a creation of an empty dependency matcher:

```java
@Override
public DependencyVersionState getInitialValue(ExecutionContext ctx) {
    return new DependencyVersionState("", ""); 
}

// accumulator 
public static class DependencyVersionState {
    DependencyMatcher dependencyMatcher;
    // other fields

    public DependencyVersionState(String groupId, String artifactId) {
       this.dependencyMatcher = new DependencyMatcher(groupId, artifactId, null);
    }
}
```

which in turn will be used in the _entire_ UpgradeTransitiveDependencyVersion recipe. So even if these recipes will call at a later stage:

```java
new UpgradeTransitiveDependencyVersion(<groupId>, <artifactId>, <version>, null, null, null)
     .getScanner(acc.getTransitiveAcc())
     .visit(sourceFile, ctx);
```

it still fails, as the accumulator still holds the empty dependency matcher. So I kept the creation of new dependency matchers instances in the scanner and visit phase, which made the problem disappear!

### Checklist


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -623,8 +623,8 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                             ((J.MethodInvocation) depArgs.get(0)).getSimpleName().equals("enforcedPlatform"))) {
                     m = m.withArguments(ListUtils.mapFirst(depArgs, platform -> updateDependency((J.MethodInvocation) platform, ctx)));
                 }
-            } else if ("ext".equals(method.getSimpleName()) && isSettingsGradle()) {
-                // rare case that gradle versions are set via settings.gradle ext block
+            } else if ("ext".equals(method.getSimpleName()) && getCursor().firstEnclosingOrThrow(SourceFile.class).getSourcePath().endsWith("settings.gradle")) {
+                // rare case that gradle versions are set via settings.gradle ext block (only possible for Groovy DSL)
```
bmuschko, line 621: I'd think you can do this with the Kotlin DSL as well. The Gradle API doesn't limit access to extensions based on the used DSL.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
```diff
@@ -185,18 +212,12 @@ public DependencyVersionState getInitialValue(ExecutionContext ctx) {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getScanner(DependencyVersionState acc) {
-        return Preconditions.check(new FindGradleProject(FindGradleProject.SearchCriteria.Marker), new JavaVisitor<ExecutionContext>() {
+        return Preconditions.check(new IsBuildGradle<>(), new JavaVisitor<ExecutionContext>() {
```
bmuschko, line 214: Is the a reason why you decided to switch to `IsBuildGradle`?

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
```diff
@@ -400,6 +402,188 @@ void updateConstraint() {
         );
     }
 
+    @Test
+    @Disabled("String interpolation with inline properties is not yet supported")
```
timtebeek:
```suggestion
    @Disabled("String interpolation with inline properties is not yet supported")
    @Test
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
jevanlingen: Does the same, but is written shorter => thus easier to maintain!

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
jevanlingen: I asked AI to come up with something multiple times, but all I get is the answer:
>  - ❌ _gradle.ext_ does not work in _settings.gradle.kts_. There is no Gradle plugin or officially supported mechanism that adds an ext-like property system to _settings.gradle.kts_. The limitations are due to how Gradle is architected: the _settings.gradle.kts_ file is evaluated before any _build.gradle.kts_ files, and it runs in a more restricted environment; mostly intended for configuring included builds, root project layout, and plugin management.
>  - ✅ In _build.gradle.kts_, use extra via `extra["foo"] = "bar"` if needed, but not in _settings.gradle.kts_.

---

I also tried Googling about this:
- https://stackoverflow.com/questions/69323109/unresolved-reference-ext-when-convert-gradle-to-gradle-kts
- https://stackoverflow.com/questions/45753733/ext-in-buildscript-can-not-be-recognised-by-gradle-kotlin-dsl

So far I can't find any support for the `gradle.ext` thing. If you can point me to some example for the documentation, I am willing to believe you and create a unit test for it 😉. But until then, I really think setting extra properties in the settings file is not an option for the Kotlin DSL. Also notice, when I create a simple unit test as:

```java
@Test
void kotlinDslGradleExt() {
    rewriteRun(
      settingsGradleKts(
        """
          gradle.ext {
              jacksonCoreVersion = '2.12.0'
          }
          """
      )
    );
}
```

I already stumble upon following error:
```
Caused by: org.gradle.tooling.BuildException: Could not fetch model of type 'OpenRewriteModel' using connection to Gradle distribution 'https://services.gradle.org/distributions/gradle-8.12-bin.zip'.
```

JohannisK (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
jevanlingen: Now this PR has merged, let's just create another PR to support this if I am mistaken. It's better that way, because right now there were no unit tests to prove the recipe could even migrate this for Kotlin DSL anyway.

PR #5809, opened 2025-07-28
Remove UpdateDependencyLock recipe and adapt consuming recipes

Jenson3210:
## What's changed?
Removed the UpdateDependencyLock and adapted its consumers

## What's your motivation?
As this project does not 100% accurately bump the lockfiles, this breaks the "Do No Harm" principle. 
We've [documented](https://docs.moderne.io/administrator-documentation/references/faq#my-build-is-broken-after-dependency-changes-when-using-gradle-lockfiles) how to 100% correctly update the dockerfiles, so no need to have a half-solution kept in place

As this is a core change, not merging until reviewed in depth

### Checklist


timtebeek (approved):
Looks fine to me since all we're doing is removing the parts related to lock files. What made you hesitant to merge this as is?

Jenson3210:
We're touching `UpgradeDependencyVersion` and `UpgradeTransitiveDependencyVersion` accumulators here. 
I also used platform to find any usages of the LockFile recipe as a String (no found), but could not 100% accurately say that the modules field I added for Lockfiles will not be used somewhere. 

As there 2 recipes are core, I just wanted to check first if this is a core-risk change or not by at least mentioning it could be one

PR #5813, opened 2025-07-28
Maven `AddDependency` widens scope instead of duplicating dependency

Laurens-W:
## What's changed?

Adjusted `AddDependency` to widen the scope when called for a dependency that's already present directly but with a narrower scope than requested.

## What's your motivation?

Was seeing dependencies being duplicated with different scopes.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
@@ -31,7 +31,6 @@ import java.util.*;
 import java.util.function.Predicate;
 
 import static java.util.Collections.emptyList;
-import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.groupingBy;
 import static org.openrewrite.internal.StringUtils.matchesGlob;
 import static org.openrewrite.maven.tree.Plugin.PLUGIN_DEFAULT_GROUPID;


rewrite-maven/build.gradle.kts
```diff
@@ -45,6 +45,7 @@ dependencies {
     testRuntimeOnly("org.mapdb:mapdb:latest.release")
     testRuntimeOnly(project(":rewrite-java-21"))
     testRuntimeOnly("org.rocksdb:rocksdbjni:latest.release")
+    testRuntimeOnly("jakarta.annotation:jakarta.annotation-api:2.1.1")
```
timtebeek: We're not using any classes from this dependency right? Why are we adding this to the build gradle file then?

timtebeek:
```suggestion
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -232,6 +233,16 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
                 String resolvedScope = scope == null ? maybeScope : scope;
                 Scope resolvedScopeEnum = Scope.fromName(resolvedScope);
+
+                Scope presentInScope = determineDirectDependencyAlreadyPresent(dependencies);
+                if (presentInScope != null) {
+                    if (Scope.maxPrecedence(presentInScope, resolvedScopeEnum) == resolvedScopeEnum) {
+                        return (Xml) new ChangeDependencyScope(groupId, artifactId, resolvedScopeEnum == Scope.Compile ? null : resolvedScope).getVisitor().visitNonNull(document, ctx);
+                    } else {
+                        return maven; // already present in a wider scope
+                    }
```
timtebeek:
```suggestion
                    }
                    return maven; // already present in a wider scope
```

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -37,7 +37,7 @@ class AddDependencyTest implements RewriteTest {
     @Override
     public void defaults(RecipeSpec spec) {
         spec.parser(JavaParser.fromJavaVersion()
-          .classpath("junit-jupiter-api", "guava", "jackson-databind", "jackson-core"));
+          .classpath("junit-jupiter-api", "guava", "jackson-databind", "jackson-core", "jakarta.annotation-api"));
```
timtebeek: We're not using any classes from the dependency though, right? Then it should not be needed.
```suggestion
          .classpath("junit-jupiter-api", "guava", "jackson-databind", "jackson-core"));
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -253,6 +263,18 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
                         type, classifier, optional, familyPatternCompiled, metadataFailures).visitNonNull(document, ctx);
             }
 
+            private @Nullable Scope determineDirectDependencyAlreadyPresent(Map<Scope, List<ResolvedDependency>> dependencies) {
+                for (Map.Entry<Scope, List<ResolvedDependency>> entry : dependencies.entrySet()) {
```
timtebeek, line 267: I couldn't immediately find whether there's a defined order to the dependencies map here; it seems this method then returns whichever scope is encountered first, not whichever has the highest precedence. Should we adjust this method to return the highest precedence instead?

jevanlingen: It's a little more nuanced than that. @Laurens-W and me actually talked about this 😎.

This method returns the scope for the dependency that matches the groupId and artifactId at root level (direct dependency, which is depth == 0). So 99% this method is fine, as duplicate dependencies at root level are _very_ unlikely. But it's still possible to define something like:

```xml
<dependency>
  <version>33.4.8-jre</version>
</dependency>
<dependency>
  <version>33.4.7-jre</version>
</dependency>
```

In this case, the first dependency should just be _ignored_ (contrary to what you would expect, the [first declaration win](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#:~:text=first%20declaration%20wins)-rule does only apply to transitive depth; try for yourself with a demo project). Our resolution framework cannot handle this though, as it tries to process both of the dependencies, albeit using the first version, which is also wrong:

Image


This problem is clearly something that's beyond the authorship of this recipe. Maybe we should create an issue/PR for this problem?. Whatever we decide, I created a test to show this: https://github.com/openrewrite/rewrite/pull/5813/commits/177981d06f900a7f06d4d7b0323de07038d92e2e. I don't know if we really want to support this at all, so @Laurens-W either think of a solution, remove the test or add `@ExpectedToFail` to the test. At least it's good we think about this for a little while more!

sambsnyd: Yeah it's a bug in the parser. According to Maven a sequence of duplicate declarations last wins. This behavior also got reported as a customer issue recently:
https://github.com/moderneinc/customer-requests/issues/1388

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,156 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
         );
     }
 
+    @Test
+    void addDependencyWithBroaderScopeChangesExistingScopeImplicit() {
+        rewriteRun(
+          spec -> spec.recipe(new AddDependency("jakarta.annotation", "jakarta.annotation-api", "2.1.1", null, null, null, null, null, null, null, null, null)),
```
jkschneider: Seeing the unit test, this just looks dangerous to me. I may be saying I want to upgrade it in the compile scope but when applied in a multi-repository environment, pulling things from provided to compile will be bad. So I don't think I'm on board with the concept of this change.

jkschneider (changes requested):

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
jkschneider: To be clear, I don't think adding a duplicate is necessarily right either.

Maybe we just need to recognize that a provided dependency is still available at compile time? So the request to add a dependency based on a type being in use in the source code is satisfied by provided or optional or compileOnly, etc.

Laurens-W: I've added a testcase to represent the scenario you've described and indeed we'd widen the scope of a consciously overridden dependency in something like a child module.

When I finish up some stuff I'll see if I can make that addition to detect in a multimodule project the compile-provided dynamic

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
jevanlingen: Resolved by 
- https://github.com/openrewrite/rewrite/pull/5872

Maybe it's worth to take another look at this case @Laurens-W, now this issue no longer exists in main?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -978,10 +978,23 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
...
+                case Provided:
+                    if (dScope == scope || dScope.transitiveOf(scope) == scope) {
+                        rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
+                    }
+                    break;
+                case Test:
+                    if (dScope == scope || dScope.transitiveOf(scope) == scope || dScope == Scope.Provided || dScope == Scope.Runtime) {
+                        rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
+                    }
+                    break;
```
Laurens-W, line 1010: I noticed in https://github.com/openrewrite/rewrite/pull/5872 already but `DependencyAndDependent` always got `Scope.Compile` passed to it, can't imagine that's correct

Laurens-W: Also note that this logic is still flawed as can be seen by https://github.com/openrewrite/rewrite/pull/5813/files#diff-b0f5d3f844d974d30e82c9f719745f74b9a63cb3c0546823b542a252621bbd03R1780-R1828
note that even on main currently this test resolves incorrectly: 
image

jevanlingen: This code is hard to read, as much is the same, but that's hard to spot unless you start reading it very carefully. I would do it more like this:



```java
boolean anotherReason = false;
if ((scope == Compile && dScope == Scope.Provided) || (scope == Test && (dScope == Scope.Provided || dScope == Scope.Runtime))) {
  anotherReason = true;
} 

if (dScope == scope || dScope.transitiveOf(scope) == scope || anotherReason) {
  rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
}
```

Jenson3210: I had also doubted about why the compile scope was always used. It was already there [before](https://github.com/openrewrite/rewrite/pull/5872/files#diff-4a72cb0156b15c23f5b5ce6ea41628e13b4e3616c563162ef0d50170792fa395L982)

This triggered me back then to check, but the scope here is never used in this method's result afaics.

Jenson3210: I can be wrong, but I was kinda expecting this to be handled correctly by `dScope.transitiveOf(scope) == scope`

but that reads as `Provided.transitiveOf(Compile)` which returns null. 
If your turn that one around, `Compile.transitiveOf(Provided)` it gives back Provided
Could that be the issue here?

Laurens-W: > I can be wrong, but I was kinda expecting this to be handled correctly by `dScope.transitiveOf(scope) == scope`
> 
> but that reads as `Provided.transitiveOf(Compile)` which returns null. If your turn that one around, `Compile.transitiveOf(Provided)` it gives back Provided Could that be the issue here?

This should only apply when transitive dependencies are at play

Laurens-W: @jevanlingen I took a slightly different approach to leave it a little more explicit imo, let me know what you think :)

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -978,10 +979,20 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
...
+            rootDependencies.remove(d.getGav().asGroupArtifact());
+            // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
+            //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
+            switch (scope) {
+                case Compile:
+                    maybePutRootDependency(scope, requestedDependency, dScope, rootDependencies, d, s -> s == Scope.Provided);
+                    break;
+                case Runtime:
+                case Provided:
+                    maybePutRootDependency(scope, requestedDependency, dScope, rootDependencies, d);
```
jevanlingen: Nitpick, but I think I would just do:


```suggestion
                    maybePutRootDependency(scope, requestedDependency, dScope, rootDependencies, d, null);
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -978,10 +979,20 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
         for (Dependency requestedDependency : getRequestedDependencies()) {
             Dependency d = getValues(requestedDependency, 0);
             Scope dScope = Scope.fromName(d.getScope());
-            if (dScope == scope || dScope.transitiveOf(scope) == scope) {
-                // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
-                //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
-                rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
+            rootDependencies.remove(d.getGav().asGroupArtifact());
```
jevanlingen, line 997:
```suggestion
            // Invoke LAST_WINS/MAVEN strategy, by removing a possible existing entry (which filters away old duplicates)
            rootDependencies.remove(d.getGav().asGroupArtifact());
```

Also, what about LATEST_WINS/GRADLE? Does this change not impact this too much?

Jenson3210: There are more things not okay when we want to correctly support Latest_wins, so I would ignore that for now. This piece of the code only gets used for the root dependencies. Having multiple the same dependencies of different versions is already not a normal thing to have at depth 0.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
jevanlingen: Yeah, that's fine as well, as the current code is also readable 👌🏻.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -154,6 +155,54 @@ public boolean isDependencyTag(String groupId, String artifactId) {
         return false;
     }
 
+    /**
+     * Is a tag a dependency that matches the group and artifact?
+     *
+     * @param groupId    The group ID glob expression to compare the tag against.
+     * @param artifactId The artifact ID glob expression to compare the tag against.
+     * @return true if the tag matches.
+     */
```
jevanlingen, line 183: I think this JavaDoc and the one above are a bit vague, especially since these methods have subtle differences (`isDependencyTag` uses resolved dependencies, while `isRequestedDependencyTag` uses requested dependencies).

I would go for something like:

```java
/**
 * Determines whether the current XML tag represents a dependency matching the given
 * groupId and artifactId patterns, based on the project's resolved dependencies.
 *
 * @param groupId    a glob expression to match against the dependency's {@code groupId}.
 * @param artifactId a glob expression to match against the dependency's {@code artifactId}.
 * @return {@code true} if the current tag matches a resolved dependency with the given criteria, otherwise {@code false}.
 */
public boolean isDependencyTag(String groupId, String artifactId) { .. }
```

and

```java
/**
 * Determines whether the current XML tag represents a dependency matching the given
 * groupId and artifactId patterns, based on the project's requested dependencies
 * (as declared in the POM before resolution).
 *
 * @param groupId    a glob expression to match against the dependency's {@code groupId}.
 * @param artifactId a glob expression to match against the dependency's {@code artifactId}.
 * @return {@code true} if the current tag matches a requested dependency with the given criteria, otherwise {@code false}.
 */
public boolean isRequestedDependencyTag(String groupId, String artifactId) { .. }
```

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -154,6 +155,54 @@ public boolean isDependencyTag(String groupId, String artifactId) {
...
+    public boolean isRequestedDependencyTag(String groupId, String artifactId) {
```
jevanlingen, line 184: This method and the one before also share 95% or the same code base. I would create another function:

```java
private boolean isDependencyTag(String groupId, String artifactId, Map<Scope, List<GroupArtifactVersion>> dependencies) { 
  // exact same implementation without the `Map<Scope, List<Dependency>> dependencies` line
}
```

The only thing you have to do then is map the ResolvedDependencies en Dependencies to the GroupArtifactVersion and call that method.

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -4039,10 +4039,10 @@ void lastListedDependencyIsUsedForScope(String firstVersion, String firstScope,
             spec -> spec.afterRecipe(pom -> {
                 MavenResolutionResult resolution = pom.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow();
                 assertThat(resolution.findDependencies("com.fasterxml.jackson.core", "jackson-core", Scope.fromName(firstScope)))
-                  .hasSize(1)
-                  .extracting(ResolvedDependency::getGav)
-                  .extracting(ResolvedGroupArtifactVersion::getVersion)
-                  .containsExactly(firstVersion);
+                  .hasSize(0);
+//                  .extracting(ResolvedDependency::getGav)
```
jevanlingen: Is this intentional?

Laurens-W: Yes as the first dependency now no longer shows up in the Resolved dependencies (used by `findDependencies`)

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
Laurens-W: I think we should keep these 2 concepts separate explicitly

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
jevanlingen: Then remove the comments.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
jevanlingen: Why? I think they are quite similar, but just use a different set. What's the reason to duplicate code?

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+              <project>
+                  <groupId>com.mycompany.app</groupId>
+                  <artifactId>my-app</artifactId>
+                  <version>1</version>
+                  <dependencies>
+                      <dependency>
+                          <groupId>jakarta.annotation</groupId>
+                          <artifactId>jakarta.annotation-api</artifactId>
+                          <version>2.1.1</version>
+                          <scope>provided</scope>
```
bmuschko, line 1720: We should probably have a test case that does not declare a scope at all in the dependency declaration. You kind of have that in the next test case but there you explicitly redefine the scope from `compile` to `provided`.

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+                  <dependencies>
+                      <dependency>
+                          <groupId>jakarta.annotation</groupId>
+                          <artifactId>jakarta.annotation-api</artifactId>
+                          <version>3.0.0</version>
+                      </dependency>
+                      <dependency>
+                          <groupId>jakarta.annotation</groupId>
+                          <artifactId>jakarta.annotation-api</artifactId>
+                          <version>2.1.1</version>
```
bmuschko, line 1823: This behavior is probably not correct. The logic changes the dependency declaration from the `test` scope to the `compile` scope. It should have preserved the existing `test` scope.

That would mean that the dependency is not available anymore for test compilation and runtime. Say this would have been a JUnit dependency then your code under `src/test/java` would not compile anymore after running the recipe if you use an import from the JUnit API.

Is it supposed to work that way?

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+            """
+              <project>
+                  <groupId>com.mycompany.app</groupId>
+                  <artifactId>my-app</artifactId>
+                  <version>1</version>
+                  <dependencies>
+                      <dependency>
+                          <groupId>jakarta.annotation</groupId>
+                          <artifactId>jakarta.annotation-api</artifactId>
+                          <version>2.1.1</version>
```
bmuschko, line 1861: Same issue as above.

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+                          <scope>provided</scope>
```
bmuschko, line 1900: `test` should not be changed into `provided`. This would cause test dependencies not to be available for test source code compilation anymore.

> `test`: This scope indicates that the dependency is not required for normal use of the application, and is only available for the test compilation and execution phases.

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

Or is the intention of `AddDependency` to override existing scopes as it was explicitly requested by the end user?

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
```
bmuschko, line 1924: `test` and `provided` are not really inheriting from one another. I think "narrower" is an inaccurate description for the test case.

Inheritance hierarchy:

test -> does not inherit from another scope
provided -> compile

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+            pomXml(
+              """
+                <project>
+                    <groupId>com.mycompany.app</groupId>
+                    <artifactId>root</artifactId>
+                    <version>1</version>
+                    <modules>
+                        <module>project1</module>
+                    </modules>
+                    <dependencies>
```
bmuschko, line 1745: Maybe I am not fully understanding what you mean by "transitive" in the test method name here. Which of those cases are you testing for here?

- Scope inheritance
- Dependency inheritance of the submodule from the root project
- Both (if both then I'd separate the test cases)

I don't think that you are overriding the dependency in `project1` from the one declared in the `root` project. In Maven, dependencies declared in the root (parent) module are not automatically inherited by submodules.

Laurens-W: I did add the root project as a parent to the submodule, which I believe is testing scenario 2 (Dependency inheritance of the submodule from the root project)

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
Laurens-W: > That would mean that the dependency is not available anymore for test compilation and runtime.

Are you sure about that?
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
Laurens-W: > Or is the intention of AddDependency to override existing scopes as it was explicitly requested by the end user?

Yes this is the intention, and that is what is being tested here, not the validity of the change on a real world project :)

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
Laurens-W: To prove we don't add a scope when left implicit? We already have a test for that: 
https://github.com/openrewrite/rewrite/blob/b5cb986ff101c7a52bfbfee9e174e6cd9580040b/rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java#L678-L705

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -232,6 +233,15 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
 
                 String resolvedScope = scope == null ? maybeScope : scope;
                 Scope resolvedScopeEnum = Scope.fromName(resolvedScope);
+
+                Scope presentInScope = determineDirectDependencyAlreadyPresent(dependencies);
+                if (presentInScope != null) {
+                    if (Scope.maxPrecedence(presentInScope, resolvedScopeEnum) == resolvedScopeEnum) {
```
Laurens-W, line 239: This follows the order defined here
https://github.com/openrewrite/rewrite/blob/b5cb986ff101c7a52bfbfee9e174e6cd9580040b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Scope.java#L21-L28

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
bmuschko: > To prove we don't add a scope when left implicit?

Yes, that was my thought process here.

> We already have a test for that

OK, great! All good then.

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
bmuschko: I just tried this with an actual Maven project. Dependencies defined in the parent module are inherited by the child module.

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -1699,6 +1700,236 @@ void addDependencyToAggregatingPomAndParentPomWhenAggregatingPomIsParent() {
...
+    void addDependencyWithImplicitAndExplicitCompileScopeDoesNotChangeTransitiveProvidedScopeOverride(String scope) {
+        rewriteRun(
+          spec -> spec.recipe(new AddDependency("jakarta.annotation", "jakarta.annotation-api", "2.1.1", null, scope, null, null, null, null, null, null, null)),
+          mavenProject("root",
+            pomXml(
+              """
+                <project>
+                    <groupId>com.mycompany.app</groupId>
+                    <artifactId>root</artifactId>
+                    <version>1</version>
```
bmuschko, line 1741: Maybe our test harness is adding that automatically but you might be missing `<packaging>pom</packaging>` here.

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
bmuschko: I stand corrected here. The compilation classpath is available to the test compilation classpath.

In the case of a test dependency like JUnit, this would added to the published `pom.xml`. I guess it would be a deliberate decision of the end user though. Do we want to guard against such a situation?

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
bmuschko: Got it. Thanks!

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -382,6 +434,24 @@ protected boolean isProperty(@Nullable String value) {
...
+    public @Nullable Dependency findRequestedDependency(Xml.Tag tag, @Nullable Scope inClasspathOf) {
+        Scope tagScope = Scope.fromName(tag.getChildValue("scope").orElse("compile"));
+        if (inClasspathOf != null && tagScope != inClasspathOf && !tagScope.isInClasspathOf(inClasspathOf)) {
+            return null;
+        }
+        for (Map.Entry<Scope, List<Dependency>> scope : getResolutionResult().getPom().getRequestedDependencies().stream().collect(groupingBy(dep -> Scope.fromName(dep.getScope()))).entrySet()) {
+            if (inClasspathOf == null || scope.getKey() == inClasspathOf || scope.getKey().isInClasspathOf(inClasspathOf)) {
+                for (Dependency d : scope.getValue()) {
+                    if (tag.getChildValue("groupId").orElse(getResolutionResult().getPom().getGroupId()).equals(d.getGroupId()) &&
+                        tag.getChildValue("artifactId").orElse(getResolutionResult().getPom().getArtifactId()).equals(d.getArtifactId())) {
```
Jenson3210: These 2 orElse's confuse me. What if the tag has no artifactId nor a groupId child? This will search for current project as a requested dependency? How will the artifact name of the project matter? 

Also, what If I send a random non-dependency tag to this method. We will fall in all else-blocks of the optionals and potentially still return something? Or am I missing something?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -978,10 +979,21 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
         for (Dependency requestedDependency : getRequestedDependencies()) {
             Dependency d = getValues(requestedDependency, 0);
             Scope dScope = Scope.fromName(d.getScope());
-            if (dScope == scope || dScope.transitiveOf(scope) == scope) {
-                // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
-                //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
-                rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
+            // Apply LATEST_WINS/MAVEN strategy, by removing a possible existing entry (which filters away old duplicates)
```
Jenson3210:
```suggestion
            // Apply NEAREST_WINS/MAVEN strategy, by removing a possible existing entry (which filters away old duplicates)
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -978,10 +979,21 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
...
+            rootDependencies.remove(d.getGav().asGroupArtifact());
+            // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
+            //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
```
Jenson3210: I know its my comment, but till then I had no idea it had official names 😄 
```suggestion
            //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for NEAREST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
```

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
Laurens-W: I think the aim is to have some sensible fallback that never matches, but that can also be achieved by returning an empty string in the `orElse`. I'm not sure what the reasoning is though, This method is a duplicate of the one above it, merely checking the requested dependencies instead of the resolved ones

rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
Laurens-W: It's not necessary for parsing and proving the recipe works as intended, we try to keep the poms in these tests as minimal as possible, notice for example that there's also no namespace and schema attributes, nor the model version tag

steve-aom-elliott:
- Superseded by https://github.com/openrewrite/rewrite/pull/6434

PR #5814, opened 2025-07-28
Dependency resolution failure markers linger after sequential `ChangeDependency` calls that bring file into correct state

steve-aom-elliott:
## What's changed?
Added a test showing the failure markers when running two `ChangeDependency` recipes for dependencies relying on the same `pom.xml` property for their version number. I suppose the problem is that if you run the recipes one after the other, although it reaches the correct state, it's going to still say in the file that there were resolution failures and that would be in the diff.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeManagedDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeManagedDependency.java
@@ -123,7 +123,7 @@ public class ChangeManagedDependency extends Recipe {
             validated = validated.and(Semver.validate(newVersion, versionPattern));
         }
         validated = validated.and(Validated.required("newGroupId", newGroupId).or(Validated.required("newArtifactId", newArtifactId)));
-        validated = validated.and(Validated.test(
+        return validated.and(Validated.test(
                 "coordinates",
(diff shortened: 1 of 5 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -1676,4 +1677,48 @@ void changeDependencyGroupIdAndArtifactIdWithVersionProperty() {
           )
         );
     }
+
+    @Test
+    void shouldUpdateSharedPropertyVersionNumberForChangedDependency() {
+        rewriteRun(
+          spec -> spec.recipes(
+            new ChangeDependencyGroupIdAndArtifactId("io.swagger", "swagger-annotations", "io.swagger.core.v3", null, "2.2.x", null, null, null),
+            new ChangeDependencyGroupIdAndArtifactId("io.swagger", "swagger-models", "io.swagger.core.v3", null, "2.2.x", null, null, null)
```
timtebeek, line 1686: Heads up that this is also explored in
- https://github.com/openrewrite/rewrite-openapi/pull/45

timtebeek:
Closing because of conflicts, and because the linked issue has since been closed.

PR #5815, opened 2025-07-28
`ChangeDependencyGroupIdAndArtifactId` does not update properties in parent pom

steve-aom-elliott:
## What's changed?
The `ChangeDependencyGroupIdAndArtifactId` recipe can also use properties defined a parent pom.

## What's your motivation?
Having version's defined in a parent pom is quite common. Other recipes like `UpgradeDependencyVersion` do already support this.

## Additional context
Most of the changes consist of moving all the visitor code into the scanning phase, so we can determine which actions have to be done (whether it is changing properties, creating, remove or changing values). Then in the actual visitor phase we apply these changes accordingly.

### Checklist


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -1676,4 +1676,96 @@ void changeDependencyGroupIdAndArtifactIdWithVersionProperty() {
...
+                  </modules>
+                </project>
+                """,
+              """
+                <project>
+                  <groupId>com.mycompany.app</groupId>
+                  <artifactId>parent-project</artifactId>
+                  <version>1</version>
+                  <properties>
+                    <activation.api>1.2.2</activation.api>
```
jevanlingen, line 1713:
```suggestion
                    <activation.api.groupid>javax.activation</activation.api.groupid>
                    <activation.api.artifact>javax.activation-api</activation.api.artifact>
                    <activation.api.version>1.2.0</activation.api.version>
```

should be able to work now as well! _But it doen't yet, so I'll work on some more..._

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -170,58 +166,59 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
-                                    t = changeChildTagValue(t, "version", resolvedNewVersion, ctx);
+                                    storeParentPomProperty(currentVersion.orElse(null), resolvedNewVersion);
+                                    acc.changeVersion = resolvedNewVersion;
                                 }
                             } else if (configuredToOverrideManageVersion || !newDependencyManaged) {
-                                //If the version is not present, add the version if we are explicitly overriding a managed version or if no managed version exists.
-                                Xml.Tag newVersionTag = Xml.Tag.build("<version>" + resolvedNewVersion + "</version>");
-                                //noinspection ConstantConditions
-                                t = (Xml.Tag) new AddToTagVisitor<ExecutionContext>(t, newVersionTag, new MavenTagInsertionComparator(t.getChildren())).visitNonNull(t, ctx, getCursor().getParent());
+                                // If the version is not present, add the version if we are explicitly overriding a managed version or if no managed version exists.
```
steve-aom-elliott, line 214: Would it be worth shifting or replicating comments like this to the places in the visitor where the removed code below has been moved to?

steve-aom-elliott:
There's obviously a lot of changes, but in general it seems to be doing what it needs to.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
jevanlingen: I would say: **no**. The logic was in the visitor, now it's in the scanner. The visitor has now become a "dumb" thing, by a "just do it" mentality if the scanner proposes a change. If you would do such logic in the visitor again (even by just adding comments), you would only make it harder, not easier, to figure out what's going on.

timtebeek:
As discussed with Jonathan I'll do a last pass over this one before we merge, since we already use the same pattern elsewhere and can resolve the immediate need with this change.

In the near term we'd want to revisit this recipe and others like it to extract traits, as we're increasingly seeing complicated logic in each of the Maven (and Gradle) dependency recipes, with subtle variations that might best to pulled out of the recipes and handled once.

timtebeek (approved):

Issue #5673, opened 2025-06-25
Illegal character in path at index 76: `~/.m2/repository/org/apache/logging/log4j/log4j-bom/${log4j2.version}/log4j-bom-${log4j2.version}.pom`


xLitil:
## What version of OpenRewrite are you using?
I am using

- rewrite-recipe-bom: 3.9.0
- rewrite-maven: 8.55.0 (bom overrided)

## What is the smallest, simplest way to reproduce the problem?

Recipe: 
```yaml
type: specs.openrewrite.org/v1beta/recipe
name: com.socgen.digital.xbl.recipes.RemoveLog4JBom
description: >-
  Remove log4j-bom dependency because it's already defined un SB's BOM.
recipeList:
  - org.openrewrite.maven.RemoveManagedDependency:
      groupId: org.apache.logging.log4j
      artifactId: log4j-bom
  - org.openrewrite.maven.RemoveDependency:
      groupId: org.apache.logging.log4j
      artifactId: log4j-bom
  - org.openrewrite.maven.RemoveProperty:
      propertyName: log4j2.version
```

Unit test (multi module project):
```java
class RemoveLog4JBomTest implements RewriteTest {

    @Override
    public void defaults(RecipeSpec spec) {
        Recipe recipe = Environment.builder()
                .scanYamlResources()
                .build()
                .activateRecipes("com.socgen.digital.xbl.recipes.RemoveLog4JBom");
        spec.recipe(recipe);
    }

    @Test
    void removeLog4JBom() {
        rewriteRun(
                mavenProject("",
                        srcMainResources(
                                pomXml(
                                        """
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

    <groupId>dummyGroupId</groupId>
    <artifactId>parentArtifactId</artifactId>
    <version>1.0.0-SNAPSHOT</version>
    <packaging>pom</packaging>
    
    <properties>
        <log4j2.version>2.17.1</log4j2.version>
    </properties>

    <modules>
        <module>child-module</module>
    </modules>
  
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.apache.logging.log4j</groupId>
[...]

Issue #5144, opened 2025-03-07
RemoveRedundantDependencyVersions It does not appear to support identifying when a property is declared to drive the dependency version


wac84s:
RemoveRedundantDependencyVersions It does not appear to support identifying when a property is declared to drive the dependency version.  I dont know if this is a specific spring boot thing or a common use case.  Can this support be added?
https://docs.spring.io/spring-boot/appendix/dependency-versions/properties.html (edit

## What version of OpenRewrite are you using?

I am using

- OpenRewrite Bom 3.3.0
- OpenRewrite v8.47.1
- rewrite-maven v8.47.1

## How are you running OpenRewrite?

I am using the Maven project to write a recipie, and my project is a single module project.
```java
public class MyRecipeChain extends Recipe {

    @Override
    public String getDisplayName() {
        return "My Recipe Chain";
    }

    @Override
    public String getDescription() {
        return "A chain of recipes to perform multiple transformations.";
    }

    @Override
    public List<Recipe> getRecipeList() {
        String groupPattern=null;
        String artifactPattern=null;
        RemoveRedundantDependencyVersions.Comparator onlyIfManagedVersionIs=RemoveRedundantDependencyVersions.Comparator.GTE;
        List<String> except=null;

        return Arrays.asList(
            //new ChangeManagedDependencyGroupIdAndArtifactId("org.yaml","snakeyaml","org.yaml","snakeyaml","2.3.0")
                new org.openrewrite.maven.RemoveRedundantDependencyVersions(groupPattern,artifactPattern,onlyIfManagedVersionIs,except)
                //removes stuff it shouldnt and leaves stuff it shouldnt
                //,new org.openrewrite.maven.RemoveUnusedProperties(null)

            //new ChangeTypeRecipe("old.TypeB", "new.TypeB")
            // Add more recipes as needed
        );
    }
}
```

## What is the smallest, simplest way to reproduce the problem?

I picked latest spring parent starter 3.4.3 and purposely picked older dependencies it manages

```pomMaven
<?xml version="1.0" encoding="UTF-8"?>
                             <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                                      xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
                                 <modelVersion>4.0.0</modelVersion>
                                 <parent>
[...]

PR #5821, opened 2025-07-29
Avoid overwriting parent pom from marker when going in child.

Jenson3210:
Thanks for the helpful reproducers @xLitil 

## What's changed?
By only `put`ting if absent, the resulting `Map` now contains the requested pom of the level in the hierarchy and not the deeper nested ones. As we make changes to the pom itself and not to the child's parent references, the update marker was failing to resolve due to a child's requested pom overwriting the parent requested pom.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5673

Should we rework the code to update all the child's parent references when removing dependency? I think this should suffice as the parents references are update in a later stage to the newly calculated parent. We just had to be able to calculate the parent which was failing. 

Now I also wonder if we can ever come in the situation where we do not start from the parent in that marker building that Map as that would then use the nested one and not the parent However, we could easily traverse to the highest parent first before building the map resulting in always the same MAP/pom values to be present iso the entry point where you called the method. Traversing to `parent == null` first would mean some more loops, but also acceptable imho. That would be more correct. 

@timtebeek please see comment above and inform me what you think would be best

## Related issues
- https://github.com/openrewrite/rewrite/issues/4949
- https://github.com/openrewrite/rewrite/issues/5144

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
@@ -230,7 +230,7 @@ public class MavenResolutionResult implements Marker {
     }
 
     private Map<Path, Pom> getProjectPomsRecursive(Map<Path, Pom> projectPoms) {
-        projectPoms.put(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
+        projectPoms.putIfAbsent(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
         if (parent != null) {
             parent.getProjectPomsRecursive(projectPoms);
         }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
@@ -230,10 +230,10 @@ public class MavenResolutionResult implements Marker {
     }
 
     private Map<Path, Pom> getProjectPomsRecursive(Map<Path, Pom> projectPoms) {
-        projectPoms.putIfAbsent(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
-        if (parent != null) {
+        if (parent != null && !projectPoms.containsKey(parent.getPom().getRequested().getSourcePath())) {
             parent.getProjectPomsRecursive(projectPoms);
         }
+        projectPoms.put(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
         for (MavenResolutionResult module : modules) {
             if (!projectPoms.containsKey(module.getPom().getRequested().getSourcePath())) {
                 module.getProjectPomsRecursive(projectPoms);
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
```diff
@@ -230,7 +230,7 @@ public boolean isMultiModulePom() {
     }
 
     private Map<Path, Pom> getProjectPomsRecursive(Map<Path, Pom> projectPoms) {
-        projectPoms.put(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
+        projectPoms.putIfAbsent(requireNonNull(pom.getRequested().getSourcePath()), pom.getRequested());
```
timtebeek: Had a quick look the debugger printing:
`pom.getRequested().getGav() + ": " + projectPoms.containsKey(sourcePath) +  " - " + sourcePath`

```
dummyGroupId:parentArtifactId:1.0.0-SNAPSHOT: false - pom.xml
dummyGroupId:child-artifact:1.0.0-SNAPSHOT: false - child-module/pom.xml
dummyGroupId:parentArtifactId:1.0.0-SNAPSHOT: true - pom.xml
dummyGroupId:parentArtifactId:1.0.0-SNAPSHOT: false - pom.xml
dummyGroupId:child-artifact:1.0.0-SNAPSHOT: false - child-module/pom.xml
dummyGroupId:parentArtifactId:1.0.0-SNAPSHOT: true - pom.xml
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
```diff
@@ -230,7 +230,7 @@ public boolean isMultiModulePom() {
...
         if (parent != null) {
             parent.getProjectPomsRecursive(projectPoms);
         }
```
timtebeek: Hard to say if there could be any negative impact since we seem to be bubbling up; but from the call log above it seems we'd still visit a parent first. Given how minimal the change is here we can monitor for any effects downstream.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
Jenson3210: Not sure I understand the question here but this is correct I think. 

When `RemoveManagedDependency` makes a change, the `maybeUpdateModel` triggers the first 3 lines. 
When later `RemoveProperty` also makes a change, the same `maybeUpdateModel` triggers the next 3 lines

rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenResolutionResult.java
Jenson3210: it seems it is written to be able to go all directions and start at any entrypoint. 
That's indeed the reason of my asking if we should first bubble up so that at least we always have the same result whatever pom you start from

Jenson3210:
@timtebeek As an alternative, bubbling up first would mean this -> also test passes
Screenshot 2025-07-29 at 13 41 21

Laurens-W (approved):

PR #5824, opened 2025-07-29
Add `exclusions` and `applyTo` options to CoalesceProperties

jevanlingen:
## What’s changed?
The CoalesceProperties recipe now supports two additional options: exclusions and applyTo.

## What’s the motivation?
With the introduction and growing use of the UnfoldProperties recipe, which acts as the counterpart to CoalesceProperties, the two recipes became slightly unbalanced in terms of functionality. By adding these two new options, the CoalesceProperties recipe now aligns more closely with its counterpart.

## Additional context
The UnfoldProperties recipe supports certain features of JsonPath Plus syntax. This was essential for that recipe to handle dotted keys correctly, as it would not function without that capability. However, this is not the case for CoalesceProperties, which can operate without dot support. Therefore, I opted to keep this version simpler and limited the enhancement to elevating the JsonPathMatcher.

### Checklist

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
@@ -15,11 +15,37 @@
  */
 package org.openrewrite.yaml;
 
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 
+import java.util.List;
+
+import static java.util.Collections.emptyList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
 public class CoalesceProperties extends Recipe {
+
+    @Option(displayName = "Exclusions",
+            description = "An optional list of [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expressions to specify keys that should not be unfolded.",
+            example = "$..[org.springframework.security]")
+    List<String> exclusions;
+
+    @Option(displayName = "Apply to",
+            description = "An optional list of [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expressions that specify which keys the recipe should target only. " +
+                    "Only the properties matching these expressions will be unfolded.",
+            example = "$..[org.springframework.security]")
+    List<String> applyTo;
+
+    public CoalesceProperties(@Nullable final List<String> exclusions, @Nullable final List<String> applyTo) {
+        this.exclusions = exclusions == null ? emptyList() : exclusions;
+        this.applyTo = applyTo == null ? emptyList() : applyTo;
+    }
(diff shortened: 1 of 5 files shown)


steve-aom-elliott (approved):

timtebeek:
For context: neither recipe seems to be used by other declarative(!) recipes, so we're fairly safe adding these options here in terms of impact:
- https://docs.openrewrite.org/recipes/yaml/coalesceproperties
- https://docs.openrewrite.org/recipes/yaml/unfoldproperties

rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
```diff
@@ -15,12 +15,42 @@
...
+            description = "An optional list of [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expressions that specify which keys the recipe should target only. " +
+                    "Only the properties matching these expressions will be unfolded.",
+            example = "$..[org.springframework.security]")
+    List<String> applyTo;
+
+    public CoalesceProperties() {
+        this(null, null);
+    }
+
+    public CoalesceProperties(@Nullable final List<String> exclusions, @Nullable final List<String> applyTo) {
```
timtebeek: Should either of these (likely the latter) be tagged JsonCreator? Or is that not the latest state?

rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalescePropertiesVisitor.java
```diff
@@ -16,16 +16,29 @@
...
 public class CoalescePropertiesVisitor<P> extends YamlIsoVisitor<P> {
     private final FindIndentYamlVisitor<P> findIndent = new FindIndentYamlVisitor<>();
+    private final List<JsonPathMatcher> exclusionMatchers;
+    private final List<JsonPathMatcher> applyToMatchers;
 
     public CoalescePropertiesVisitor() {
+        this(null, null);
+    }
+
+    public CoalescePropertiesVisitor(@Nullable final List<String> exclusions, @Nullable final List<String> applyTo) {
```
timtebeek: Same as above with regards to ambiguity when instantiating recipes.

rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalescePropertiesVisitor.java
```diff
@@ -41,39 +54,50 @@ public Yaml.Document visitDocument(Yaml.Document document, P p) {
...
 
-        return m;
+    private boolean match(Cursor c, List<JsonPathMatcher> matchers) {
+        while (c != null && !c.isRoot()) {
+            Cursor current = c;
+            if (matchers.stream().anyMatch(it -> it.matches(current))) {
+                return true;
+            } else {
+                c = c.getParent();
+            }
```
timtebeek:
```suggestion
            }
            c = c.getParent();
```

rewrite-yaml/src/main/java/org/openrewrite/yaml/CoalesceProperties.java
jevanlingen: Yes, indeed. Could catch!

Issue #5823, opened 2025-07-29
Rewrite Recipe to remove build cache configuration for Develocity


div-bargali:
## What problem are you trying to solve?
Presently there is a rewrite recipe for Gradle projects that removes the `develocity` or `gradleEnterprise` configuration from `settings.gradle(.kts)` or `build.gradle(.kts) file`. However, users can also have a buildCache method configured inside the settings or build.gradle file.

This new recipe will remove the build cache configuration from the settings or build.gradle file.

## What precondition(s) should be checked before applying this recipe?
The recipe should be only applied on settings or build.gradle file, using the precondition: 
`or(new IsBuildGradle<>(), new IsSettingsGradle<>())`

## Describe the situation before applying the recipe

```groovy
buildCache {
  remote(HttpBuildCache) {
    url = "some-url"
    push = true
    enabled = true
  }
}
```

## Describe the situation after applying the recipe

```groovy
// buildCache configuration removed
```

## Have you considered any alternatives or workarounds?

Currently there does not seem to be any alternative recipes. 

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?

Yes, would like to contribute the recipe in case of no workarounds.

timtebeek:
hi @div-bargali ; thanks for the suggestion! From looking at the existing recipe you referenced it should be fairly straightforward to add a parallel recipe to remove buildCache as well.
https://github.com/openrewrite/rewrite/blob/dd2886c193ea20f416171b5b280a49086e969710/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java#L30-L57

timtebeek:
Here's a quick attempt:
- https://github.com/openrewrite/rewrite/pull/5826

PR #5826, opened 2025-07-29
Remove Gradle buildCache configuration

timtebeek:
- fixes https://github.com/openrewrite/rewrite/issues/5823

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveExtension.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveExtension.java
@@ -35,7 +35,7 @@ public class RemoveExtension extends Recipe {
 
     @Override
     public String getDescription() {
-        return "Remove the Gradle build extension from `settings.gradle(.kts)` or `build.gradle(.kts)` files.";
+        return "Remove a Gradle build extension from `settings.gradle(.kts)` or `build.gradle(.kts)` files.";
     }
 
     @Option(displayName = "Method name",


shanman190 (approved):
We're kinda tiptoeing around the line of removal of an arbitrary extension via specific recipes.

While not necessarily something to be addressed right now, it may be something that we revisit sooner or later.

timtebeek:
> We're kinda tiptoeing around the line of removal of an arbitrary extension via specific recipes.
> 
> While not necessarily something to be addressed right now, it may be something that we revisit sooner or later.

Good call out, thanks! Updated the PR to now add a recipe that takes in an argument method name, and also use that when it comes to removing develocity configuration. Hope you agree! :)

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveBuildConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+
+import static org.openrewrite.Preconditions.or;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class RemoveBuildConfiguration extends Recipe {
```
shanman190: Just a naming thing, but what about:

```suggestion
public class RemoveExtension extends Recipe {
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveBuildConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public String getDisplayName() {
+        return "Remove build configuration method by name";
```
shanman190: After the above, this could become something like:

```suggestion
        return "Remove build extension by name";
```

shanman190 (approved):
Just a possible rename to align more closely with Gradle terminology.

Issue #5829, opened 2025-07-30
Maven recipe to remove redundant properties


tschut:
## What problem are you trying to solve?
In Maven projects it is common practice to use properties for version management. We often see people overriding properties for valid reasons, but then forgetting to remove their overrides when they are no longer needed.

It would be nice to have a recipe, similar to [RemoveRedundantDependencyVersions](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantDependencyVersions.java), but to remove redundant properties.

Note that although version properties is a common usecase, this could just as well apply to other properties.

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?

I'm working on a recipe already, will open draft PR soon.

timtebeek:
- Fixed in https://github.com/openrewrite/rewrite/pull/5831

Thanks again, and enjoy what's left of your holidays!

PR #5831, opened 2025-07-30
Maven recipe to remove redundant properties

tschut:
## What's changed?
Implement https://github.com/openrewrite/rewrite/issues/5829

@timtebeek 

Note: I'll try to add some tests tomorrow or Friday, but not sure I'll have the time.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantProperties.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantProperties.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven;
+
+import lombok.EqualsAndHashCode;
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for the start! Hope you don't mind I pushed some tests to speed things up a bit; Let me know if you agree!

rewrite-maven/src/main/java/org/openrewrite/maven/RemoveRedundantProperties.java
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright 2020 the original author or authors.
```
greg-at-moderne:
```suggestion
 * Copyright 2025 the original author or authors.
```

timtebeek:
One thing I've considered here is to rename the recipe to something like RemovePropertyOverride , and then take an additional argument when lower/equal/higher. That way we can reuse the logic elsewhere more easily.

tschut:
> One thing I've considered here is to rename the recipe to something like RemovePropertyOverride , and then take an additional argument when lower/equal/higher. That way we can reuse the logic elsewhere more easily.

Yes, I've considered that too but did not yet settle on something. Generally I see two possible approaches:
1. The recipe is generic to properties, it doesn't know/understand these might be versions. This is basically the way I've set it up right now.
2. The recipe assumes the properties contain a version, and accept an argument (similar to `RemoveRedundantDependencyVersions`) to configure how the comparison is done.

Maybe there is also 
3. We do 2, but if the property value can't be parsed as semver fallback to generic equals/notEquals.

Anything not 1. does make the recipe a bit more complex though. Wdyt?

timtebeek:
You're right that adding higher/lower does mean we need to be certain they're versions to begin with. Perhaps best then to limit ourselves for now to only remove properties that are overriding with the same value, and make that explicit in the docs.

Want us to take over while you enjoy your holiday? Either way enjoy there!

tschut:
> Want us to take over while you enjoy your holiday? Either way enjoy there!

Yes that would be awesome! I don't think I'll get to it myself this week.

timtebeek (approved):
Thanks @tschut ! Great to be able to clear these out.

PR #5801, opened 2025-07-25
Moved the example of the template compilation issue for easier troubleshooting/testing

Jenson3210:
- reference: https://github.com/openrewrite/rewrite-spring/pull/757

I moved a reproduceable error to this branch. (NOT INTENDED FOR MERGING)
The issue occurs when you run the test added with the currently commented out code in `RewriteJavaTemplate` and does not occur when we add a second compilation trial.

PR #5832, opened 2025-07-30
Retry parsing in JavaTemplateParser

greg-at-moderne:
## What's changed?

Adding a retry in `JavaTemplateParser` parsing the Java code.
The code was taken from 
- #5801.

## What's your motivation?

This is a workaround for a weird case described in https://github.com/openrewrite/rewrite-spring/pull/757.

- No better solution is known at this moment
- Significant time was spent trying to come up with a proper fix
- This unblocks some work depending on it


Jenson3210:
I think we agree that we do not really like the solution, 

The biggest YES for this way would be that it only comes into picture for templates which failed to parse the first time. So normally no of the existing template usages are impacted

Jenson3210 (approved):
I think we agree that we do not really like the solution,

The biggest YES for this way would be that it only comes into picture for templates which failed to parse the first time. So normally no of the existing template usages are impacted

rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateParser.java
```diff
@@ -270,13 +270,34 @@ private JavaSourceFile compileTemplate(@Language("java") String stub) {
...
+            sourceFile = (stub.contains("@SubAnnotation") ?
+                    jp.reset().parse(ctx, stub, SUBSTITUTED_ANNOTATION) :
+                    jp.reset().parse(ctx, stub))
+                    .findFirst()
+                    .filter(JavaSourceFile.class::isInstance) // Filters out ParseErrors
+                    .map(JavaSourceFile.class::cast)
+                    .orElseThrow(() -> new IllegalArgumentException("Could not parse as Java:\n" + stub));
+        }
+
+        return sourceFile;
```
timtebeek: Might it help to extract a method to avoid the duplication and make it clear we're doing the same except for the throw?

```java
    private JavaSourceFile compileTemplate(@Language("java") String stub) {
        ExecutionContext ctx = new InMemoryExecutionContext();
        ctx.putMessage(JavaParser.SKIP_SOURCE_SET_TYPE_GENERATION, true);
        ctx.putMessage(ExecutionContext.REQUIRE_PRINT_EQUALS_INPUT, false);
        Parser jp = parser.build();
        return getJavaSourceFile(stub, jp, ctx)
                // In some specific and rare cases, the parser fails to parse what is a valid program. This has been
                // investigated for several days to no avail, so the workaround is to retry parsing, which is known to
                // address the issue.
                // Context:
                // - https://github.com/openrewrite/rewrite-spring/pull/757
                // - also a thread in private Moderne slack mentioning this PR
                // - https://github.com/openrewrite/rewrite/pull/5801 which is a unit test that reproduces the issue
                // TLDR: I suspect either a bug in Java Compiler, or some fault in how we call its internals.
                .orElseGet(() -> getJavaSourceFile(stub, jp, ctx)
                        .orElseThrow(() -> new IllegalArgumentException("Could not parse as Java:\n" + stub)));
    }

    private static Optional<JavaSourceFile> getJavaSourceFile(@Language("java") String stub, Parser jp, ExecutionContext ctx) {
        return (stub.contains("@SubAnnotation") ?
                jp.reset().parse(ctx, stub, SUBSTITUTED_ANNOTATION) :
                jp.reset().parse(ctx, stub))
                .findFirst()
                .filter(JavaSourceFile.class::isInstance) // Filters out ParseErrors
                .map(JavaSourceFile.class::cast);
    }
```

timtebeek (approved):

Issue #3692, opened 2023-11-15
`MethodMatcher` fails to properly match using standalone `*` wildcard


shanman190:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite latest snapshot

## How are you running OpenRewrite?
While authoring a custom recipe for LaunchDarkly, I was attempting to use the `MethodMatcher` to match on any parameter type in the second method parameter. Essentially, the conversion to the aspectj pattern fails to handle the pure wildcard match.

## What is the smallest, simplest way to reproduce the problem?
```java
class A {
    void foo(String bar) {
    }
    void foo(int i) {
    }
    void foo(double d) {
    }
}
```

```java
MethodMatcher matcher  = new MethodMatcher("A foo(*)");
```

This will match just fine for any pure primitive type, but will fail on any class type or a primitive wrapper type.

## What did you expect to see?
When a complex type and the generic wildcard is used, the matcher should return `true` for matches. This aligns with the current OSS documentation as well as the behavior of Aspectj.

## What did you see instead?
The matcher presently returns `true` for the `int` and `double` methods, but returns `false` for the `java.lang.String` type.

## What is the full stack trace of any errors you encountered?
N/A

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Always. 😄

timtebeek:
Indeed replicated with
```java
@Test
@Issue("https://github.com/openrewrite/rewrite/issues/3692")
void wildcardMatchesTypes() {
    rewriteRun(
      java("""
        class Foo {
            void foo(String bar) {}
            void foo(int i) {}
            void foo(double d) {}
        }
        """),
      java(
        """
          class Bar {
              void bar(Foo f) {
                  f.foo("s");
                  f.foo(1);
                  f.foo(1.0);
              }
          }
          """,
        spec -> spec.afterRecipe(cu -> assertThat(FindMethods.find(cu, "Foo foo(*)")).hasSize(3))
      )
    );
}
```
This only finds 2 instances, not 3.

timtebeek:
Looks like we might not support the wildcard for arguments:
https://github.com/openrewrite/rewrite/blob/6873fbd6d6550d5412db9b448e364832a7a539f5/rewrite-java/src/main/java/org/openrewrite/java/MethodMatcher.java#L48

[...]

PR #5833, opened 2025-07-30
Fixed wildcard matching for elements with type information.

Jenson3210:
## What's changed?
Updated the antlr code to allow for wildcard parameters similarly to dot_dot. 

## What's your motivation?
Before, the existing test used to work due to the following: 
- `*` gets regexed by the `StringUtils.aspectjNameToPattern("*")` to `[^.]+`.
- Foo.bar() was a non existing method => type information `<unknown>` does not contain a `.` so the matcher works. 
- The moment you add type information by returning a string for a static method in the test, the test fails. 

Screenshot 2025-07-30 at 11 50 51

_The current code behavior: see the `<unknown>` and `[^.]+`_

Screenshot 2025-07-30 at 11 58 40

_Changing the test adding type information_

Screenshot 2025-07-30 at 12 00 00

_See how that regex would no longer match?_

@greg-at-moderne 

We could alter the `getRegex` on `FormalType` to avoid changes to anltr code but that's not correct according the semantics of the code.

### Checklist


Jenson3210:
- fixes https://github.com/openrewrite/rewrite/issues/3692

rewrite-java-test/src/test/java/org/openrewrite/java/MethodMatcherTest.java
```diff
@@ -140,15 +140,49 @@ void matchesUnqualifiedJavaLangArguments() {
     void matchesArgumentsWithWildcards() {
         assertTrue(argRegex("A foo(java.util.*)").matcher("java.util.Map").matches());
         assertTrue(argRegex("A foo(java..*)").matcher("java.util.Map").matches());
+        assertTrue(argRegex("A foo(*.util.*)").matcher("java.util.Map").matches());
+        assertTrue(argRegex("A foo(*..*)").matcher("java.util.Map").matches());
+    }
+
+    @Test
+    void matchesExactlyOneWithWildcard() {
```
jevanlingen, line 149: Could you add
- a non primitive example
- a version without spaces
?

Both should pass, but better to have these tests as well 😄.

rewrite-java-test/src/test/java/org/openrewrite/java/MethodMatcherTest.java
```diff
@@ -140,15 +140,49 @@ void matchesUnqualifiedJavaLangArguments() {
...
+        assertFalse(argRegex("A foo(int, *, double)").matcher("int,double").matches());
+        assertFalse(argRegex("A foo(int, *, double)").matcher("double,int,double").matches());
     }
 
     @Test
     void matchesArgumentsWithDotDot() {
         assertTrue(argRegex("A foo(.., int)").matcher("int").matches());
         assertTrue(argRegex("A foo(.., int)").matcher("int,int").matches());
+        assertTrue(argRegex("A foo(.., int)").matcher("double,int").matches());
+        assertFalse(argRegex("A foo(.., int)").matcher("int,double").matches());
```
jevanlingen, line 183: As .. would mean one or more, I guess this works too:

```java
assertTrue(argRegex("A foo(.., int)").matcher("double,double,int").matches());
assertTrue(argRegex("A foo(int, .., double)").matcher("int,double,String,int,double").matches());
```

So add these tests as well!

Jenson3210: this indeed works

jevanlingen (approved):

Jenson3210:
@timtebeek Do I add this one to the list of core changes? As MethodMatcher is quite broadly used. The tests are quite good though, so not sure.

timtebeek:
Yes please add to the list of core issues to review! 🙏🏻

timtebeek (approved):
Approved from my side already; added another test to show the difference pointed out in PR comments when type information for `Foo.bar()` is present, but for `Assert.assertTrue` likely still absent.

rewrite-java-test/src/test/java/org/openrewrite/java/MethodMatcherTest.java
```diff
@@ -404,17 +449,42 @@ void matchUnknownTypesWildcardArguments() {
     void matchUnknownTypesSingleWildcardArgument() {
         var mi = asMethodInvocation("Assert.assertTrue(Foo.bar(), \"message\");");
         assertTrue(new MethodMatcher("org.junit.Assert assertTrue(*, String)").matches(mi, true));
+        assertTrue(new MethodMatcher("org.junit.Assert assertTrue(*, java.lang.String)").matches(mi, true));
+        assertTrue(new MethodMatcher("org.junit.Assert assertTrue(String, *)").matches(mi, true));
+        assertTrue(new MethodMatcher("org.junit.Assert assertTrue(double, *)").matches(mi, true));
+        assertTrue(new MethodMatcher("org.junit.Assert assertTrue(java.lang.String, *)").matches(mi, true));
```
knutwannheden, line 455: These three assertions surprise me. In this test the type of `Foo.bar()` is unknown. So why would we match against `String`, `double`, or `java.lang.String`?

knutwannheden: I guess that is how `matchUnknownTypes` is supposed to work: Just match whatever, as long as the "structure" is the same.

Issue #5843, opened 2025-07-31
RenameVariable is renaming type references (Fizz.class -> fizz.class)


jeanbottein:
## What version of OpenRewrite are you using?
I am using
- OpenRewrite rewrite-static-analysis v2.15.0-SNAPSHOT

But the issue affects the core `RenameVariable` visitor used by multiple recipes

## How are you running OpenRewrite?
Through maven command line, I discover the issue while executing on some repository.
The issue is reproducible in the test suite of the rewrite-static-analysis project.

## What is the smallest, simplest way to reproduce the problem?
```java
class Fizz {}
class Test {
    void method() {
        Fizz Fizz = new Fizz(); // Local variable shadows class name
        // Only the variable should be renamed, not the class reference
        Class<?> clazz = Fizz.class;
    }
}
```

Apply the `RenameLocalVariablesToCamelCase` recipe to this code (to rename Fizz to fizz)

## What did you expect to see?
```java
class Fizz {}
class Test {
    void method() {
        Fizz fizz = new Fizz(); // Variable renamed to camelCase
        // Class literal should remain unchanged
        Class<?> clazz = Fizz.class;
    }
}
```

The class literal `Fizz.class` should remain unchanged because it's a type reference, not a variable reference.

## What did you see instead?
```java
class Fizz {}
class Test {
    void method() {
        Fizz fizz = new Fizz(); // Variable correctly renamed
        // Class literal incorrectly renamed
        Class<?> clazz = fizz.class;
    }
}
```

The class literal `Fizz.class` is incorrectly renamed to `fizz.class`, which breaks the code since `fizz` is a variable, not a type.

## What is the full stack trace of any errors you encountered?
No stack trace - this is a logic error where the recipe produces incorrect output. The failing test is:

```
RenameLocalVariablesToCamelCaseTest > doesNotRenameClassReference() FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "Fizz.java":
    diff --git a/Fizz.java b/Fizz.java
    index c59f565..cc1e49c 100644
    --- a/Fizz.java
    +++ b/Fizz.java
    @@ -3,6 +3,6 @@ 
         void method() {
             Fizz fizz = new Fizz();
             // Only the variable should be renamed, not the class reference
    -        Class<?> clazz = Fizz.class;
    +        Class<?> clazz = fizz.class;
         }
     }
```

## Root Cause Analysis

[...]

PR #5848, opened 2025-08-01
 fix: prevent variable renaming in class literal references

jeanbottein:
## What's changed?
This PR fixes an issue in the `RenameVariable` recipe where class literal references (e.g., `SomeClass.class`) were not correctly detected, potentially causing incorrect renaming. The detection logic now checks if the parent AST node is a `J.FieldAccess` whose name is `"class"`, which is the canonical OpenRewrite representation of a class literal.

## What's your motivation?
Fixes incorrect renaming of variables when encountering class literal references. This ensures that class literals are not affected by the rename operation. (No linked issue, but discovered during recipe usage and test coverage.)

- Closes https://github.com/openrewrite/rewrite/issues/5843

Please review the logic in `RenameVariable.RenameVariableVisitor#isClassLiteralReference` and the associated unit test `RenameVariableTest#doNotRenameClassLiterals`. I'm not sure if you guys prefer to use the getCursor() in the isClassLiteralReference directly.

I considered walking further up the AST or using a utility, but the current approach is the canonical and idiomatic way for OpenRewrite's AST structure.

## Any additional context
- An issue was openned https://github.com/openrewrite/rewrite/issues/5843
A unit test was added in `RenameVariableTest` named `doNotRenameClassLiterals` to demonstrate the issue and verify the fix. All tests pass locally.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RenameVariableTest.java
@@ -957,6 +957,30 @@ class RenameVariableTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/5843")
+    @Test
+    void doNotRenameClassLiterals() {
+        rewriteRun(
+          spec -> spec.recipe(renameVariableTest("FooBarBaz", "fooBarBaz", true)),
+          java(
+            """
+              class FooBarBaz {}
+              class A {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the report & fix here @jeanbottein ! I've slightly changed your approach to override `visitFieldAccess` instead, as we can selectively choose to not visit the identifier there without having to call `getCursor()` for any `J.Identifier` we visit.

jeanbottein:
Thank you @timtebeek 
I will try to understand, so I could maybe help on other issues 🙂

PR #5850, opened 2025-08-02
Ensure rewrite-kotlin uses junit 5

jevanlingen:
## What's changed?
Pin JUnit version to 5.13.4 to avoid newer versions like 6.0.0-M2.

## What's your motivation?
Got a binary error when I tried to run the tests: 
```kotlin
file:///Users/jme/.gradle/caches/modules-2/files-2.1/org.junit.jupiter/junit-jupiter-api/6.0.0-M2/c267b83536ac774a70f37f5fd026c90799363b0/junit-jupiter-api-6.0.0-M2.jar!/META-INF/junit-jupiter-api.kotlin_moduleModule was compiled with an incompatible version of Kotlin.

The binary version of its metadata is 2.1.0, expected version is 1.9.0.
```

### DIFF
--- a/rewrite-kotlin/build.gradle.kts
+++ b/rewrite-kotlin/build.gradle.kts
@@ -1,4 +1,4 @@
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     id("org.openrewrite.build.language-library")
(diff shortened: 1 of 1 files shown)


timtebeek (approved):

jkschneider:
We shouldn't be pinning dependencies like this but rather allow them to be driven by the `rewrite.language-library` plugin. Fixed in https://github.com/openrewrite/rewrite/commit/0e5e51a8f797c1edcb60112d6dbcd0fcdaa8003d.

Issue #2082, opened 2022-07-28
HCL Parser loses newlines in empty object values


rmadamanchi:
Below is a test case to demonstrate the issue:

```kotlin
import org.junit.jupiter.api.Test
import org.openrewrite.InMemoryExecutionContext
import org.openrewrite.hcl.HclParser
import org.openrewrite.hcl.tree.Hcl.ConfigFile
import java.util.function.Consumer
import kotlin.test.assertEquals

class ParserTest {
    @Test
    fun test() {
        val parser = HclParser.builder().build()

        val source = """
            locals {
              myvar = {
            
              }
            }
        """.trimIndent()

        val ctx = InMemoryExecutionContext(Consumer { t: Throwable -> throw t })
        val sources = parser.parse(ctx, source)
        assertEquals(source, (sources[0] as ConfigFile).printAll())
    }
}
```

The printed output loses the newlines in empty object value:
```hcl
locals {
  myvar = {}
}
```

knaumchuk:
Hi. 
I noticed similar behavior for a multiline empty array declaration.  Please see example below: 

```
import org.junit.jupiter.api.Test;
import org.openrewrite.ExecutionContext;
import org.openrewrite.InMemoryExecutionContext;
import org.openrewrite.ParseExceptionResult;
import org.openrewrite.Parser;
import org.openrewrite.SourceFile;
import org.openrewrite.hcl.HclParser;
import org.openrewrite.marker.Marker;

import java.util.List;
import java.util.Optional;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;

public class HclParserTest {
    @Test
    void test() {
        Parser parser = HclParser.builder().build();

        String source = """
              myvar = [
              ]
        """;

        ExecutionContext ctx = new InMemoryExecutionContext();
        Optional<SourceFile> possibleParseResult = parser.parse(ctx, source).findFirst();
        assertTrue(possibleParseResult.isPresent());
        SourceFile parseResult = possibleParseResult.get();
        List<Marker> markers = parseResult.getMarkers().getMarkers();
        assertEquals(1, markers.size());
        Marker marker = markers.get(0);
        assertTrue(marker instanceof ParseExceptionResult);
        String message = ((ParseExceptionResult)marker).getMessage();
        assertTrue(message.contains("java.lang.IllegalStateException"));
        assertTrue(message.contains("is not print idempotent."));
[...]

Issue #5844, opened 2025-07-31
HCL object is not print idempotent


memetron:
## What version of OpenRewrite are you using?

I am using
 
- rewrite-hcl v8.9.9
 
## How are you running OpenRewrite?

I am using the Maven rewrite-hcl dependency.
```xml
<dependency>
<groupId>org.openrewrite</groupId>
<artifactId>rewrite-hcl</artifactId>
<version>8.9.9</version>
</dependency>
```
 
## What is the smallest, simplest way to reproduce the problem?

```hcl
module "foo" {
    bar = {
    }
}
```
The above fragment causes the parser to complain about lack of print idempotency, however the below works fine
```hcl
module "foo" {
    bar = {}
}
```
## What is the full stack trace of any errors you encountered?

```
java.lang.IllegalStateException: 387575330277779\file.tf is not print idempotent. 
diff --git a/387575330277779/file.tf b/387575330277779/file.tf
index 194336f..40d0da1 100644
--- a/387575330277779/file.tf
+++ b/387575330277779/file.tf
@@ -1,4 +1,3 @@ 
module "foo" {
-  bar = {
-  }
+  bar = {}
}
\ No newline at end of file
 
org.openrewrite.Parser.requirePrintEqualsInput(Parser.java:52)
```

timtebeek:
Thanks for the example! Looks like we'd need to add in an empty element to capture the whitespace, but we'd have to dive in. Care to help out? :)

greg-at-moderne:
Duplicate of #2082?

timtebeek:
- Picked up in https://github.com/openrewrite/rewrite/pull/5854

timtebeek:
We can't seem to replicate this with recent versions @memetron ; would you mind trying again with https://github.com/openrewrite/rewrite/releases/tag/v8.59.1 or later?

PR #5854, opened 2025-08-04
Visit `Hcl.Empty` in `HclPrinter` for print equality

timtebeek:
- Fixes https://github.com/openrewrite/rewrite/issues/2082

### DIFF
--- a/rewrite-hcl/src/test/java/org/openrewrite/hcl/tree/HclEmptyTest.java
+++ b/rewrite-hcl/src/test/java/org/openrewrite/hcl/tree/HclEmptyTest.java
@@ -16,6 +16,7 @@
 package org.openrewrite.hcl.tree;
 
 import org.junit.jupiter.api.Test;
+import org.openrewrite.Issue;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.hcl.Assertions.hcl;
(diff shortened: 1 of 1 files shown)


rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/HclPrinter.java
```diff
@@ -195,6 +195,13 @@ public Hcl visitConfigFile(Hcl.ConfigFile configFile, PrintOutputCapture<P> p) {
         return configFile;
     }
 
+    @Override
+    public Hcl visitEmpty(Hcl.Empty empty, PrintOutputCapture<P> p) {
```
greg-at-moderne: Interesting, so we had the parser parse some statements into `Hcl.Empty`, but we didn't have a printer for it?

timtebeek: Hmm no looking again that seems not to make a difference, despite the recent reports. I had another look at https://github.com/openrewrite/rewrite/issues/5844 and it seems that's using `v8.9.9`, whereas the latest is `v8.59.1`. I've added a reproducer to make this a test only change, and with that we'll close both issues.

Issue #5855, opened 2025-08-04
Idempotency issue with generics placeholder (?) in Javadoc


greg-at-moderne:
## What version of OpenRewrite are you using?

Current OpenRewrite release = 8.59.1

## How are you running OpenRewrite?

Moderne SaaS

## What is the smallest, simplest way to reproduce the problem?
```diff
 /**
- * @see org.apache.empire.commons.BeanPropertyUtils#convertPropertyValue(Object value, Class<?> type)
+ * @see org.apache.empire.commons.BeanPropertyUtils#convertPropertyValue(Object value, Class<> type)
  */
```

when running an unrelated recipe (`org.openrewrite.staticanalysis.SimplifyCompoundStatement` in my case).

## OSS repro

- https://github.com/apache/empire-db/blob/aba1c541a032cec8b2573aa9b58820fd0725f096/empire-db/src/main/java/org/apache/empire/commons/BeanPropertyUtils.java#L233
- https://github.com/apache/brooklyn-server/blob/e2a8e0f8a2051644d1dff44b4f3a42fe944e5f80/api/src/main/java/org/apache/brooklyn/api/entity/EntityLocal.java#L28

timtebeek:
Picked up in
- https://github.com/openrewrite/rewrite/pull/5857

PR #5857, opened 2025-08-04
Support Javadoc generic wildcards

timtebeek:
## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/5855

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
@@ -20,6 +20,7 @@ import com.sun.source.tree.ArrayTypeTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.ParameterizedTypeTree;
 import com.sun.source.tree.PrimitiveTypeTree;
+import com.sun.source.tree.WildcardTree;
 import com.sun.source.util.DocTreeScanner;
 import com.sun.source.util.TreePath;
 import com.sun.source.util.TreeScanner;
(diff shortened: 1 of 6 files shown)


rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
```diff
@@ -1254,5 +1255,32 @@ public J visitParameterizedType(ParameterizedTypeTree node, Space fmt) {
...
+            if (wildcard.kind.kind != null) {
+                switch (wildcard.kind.kind) {
+                    case EXTENDS:
+                        bound = JLeftPadded.build(J.Wildcard.Bound.Extends).withBefore(whitespace());
+                        cursor += "extends".length();
+                        break;
+                    case SUPER:
+                        bound = JLeftPadded.build(J.Wildcard.Bound.Super).withBefore(whitespace());
+                        cursor += "super".length();
+                        break;
```
greg-at-moderne, line 1274: Probably worth adding `default` to throw exception or handle it somehow.

greg-at-moderne (approved):

rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
timtebeek: So there is already a `com.sun.tools.javac.code.BoundKind.UNBOUND("?")` that we're not handling here; that would then need a case statement with a `break` as well before we can add a `default` that throws.

When not explicitly handled through `default` we expect to already get a parser to printer idempotency issue reported if any new kind where to be added & encountered in the future. For simplicity I'd rather rely on that as opposed to adding explicit handling here now that complicates the logic for all older Java versions.

Would you agree and be ok to merge as is?

greg-at-moderne: OK

PR #5860, opened 2025-08-05
RemoveImport should look through supertypes for Kotlin

jevanlingen:
## What's changed?
The remove import recipe takes imports of super types into account.

## What's your motivation?
Kotlin can import a static java member of a parent class by importing the child class. The recipe did not take this feature into account. There's also the option of using imports of companion objects with an interface. These kind of setup was not supported as well.

## Any additional context
At the moment I could not find a proper way to add Java source code as dependency of the Kotlin sources, so I added a `@Disabled` test. If there's a way to do this, it should be used.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java
@@ -29,6 +29,7 @@ import org.openrewrite.style.Style;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static java.util.Collections.singleton;
 import static org.openrewrite.Tree.randomId;
 import static org.openrewrite.java.style.ImportLayoutStyle.isPackageAlwaysFolded;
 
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java
```diff
@@ -58,48 +58,85 @@ public RemoveImport(String type, boolean force) {
...
+                        if (owningClass != null) {
+                            Queue<JavaType.FullyQualified> toVisit = new LinkedList<>(owningClass.getInterfaces());
+                            Set<JavaType.FullyQualified> visited = new HashSet<>();
+                            while (!toVisit.isEmpty()) {
+                                JavaType.FullyQualified current = toVisit.poll();
+                                if (!visited.add(current)) {
+                                    continue;
+                                }
+                                toVisit.addAll(current.getInterfaces());
+                            }
```
Laurens-W: Not sure if we'd prefer nested forloops here instead of a Queue

Laurens-W (approved):

rewrite-java/src/main/java/org/openrewrite/java/RemoveImport.java
jevanlingen: I think the queue solution is better, otherwise you need to use recursion (quick ChatGPT generated):
```java
Set<String> types = new HashSet<>(singleton(type));
if (owningClass != null) {
    Set<JavaType.FullyQualified> visited = new HashSet<>();
    collectInterfaces(owningClass, visited);
    for (JavaType.FullyQualified current : visited) {
        types.add(TypeUtils.toFullyQualifiedName(current.getFullyQualifiedName()));
    }
}

private void collectInterfaces(JavaType.FullyQualified type, Set<JavaType.FullyQualified> visited) {
    for (JavaType.FullyQualified iface : type.getInterfaces()) {
        if (visited.add(iface)) {
            // recurse into the interfaces of this interface
            collectInterfaces(iface, visited);
        }
    }
}
```

I think I slightly prefer the current solution to recursion, because you need a separate method to make recursion work¹.. At the moment all the "process more info of imports" is all located  within the `for (J.Import cuImport : cu.getImports()) {` block. As we normally favour over helper utility function when we write a recipe, I kinda like this solution.

 Do you prefer the recursive solution over queues? And so, why?

--
¹ _Recursion could also lead to stack overflows when having many cycles, but in this case I don't see it happening that interfaces that extends interfaces etc. will exceed more than 25 cycles ever._

Laurens-W: I must've misread the code initially, thinking it could be solved with a nested forloop. Your arguments for keeping it like this are strong and valid :)

timtebeek:
### Alternative 1 — extend the `ImportService` SPI (instead of Kotlin-aware logic in `rewrite-java`)

`maybeAddImport` and `maybeRemoveImport` are dispatched differently in `JavaVisitor`:

- `maybeAddImport` → `service(ImportService.class).addImportVisitor(...)` — **polymorphic**. `KotlinImportService extends ImportService` already overrides it to return the Kotlin-specific `org.openrewrite.kotlin.AddImport`.
- `maybeRemoveImport` → `new RemoveImport<>(...)` — **hardcoded** in the Java module. There is no language seam, which is *why* this PR injects Kotlin awareness (path sniffing, `"Kt"`/`"Kts"` facade handling, supertype walking) into the shared `org.openrewrite.java.RemoveImport`.

So `AddImport` already has exactly the separation `RemoveImport` is missing. The alternative:

1. Add `removeImportVisitor(...)` to `ImportService`.
2. Route `JavaVisitor#maybeRemoveImport` through `service(ImportService.class)` instead of `new RemoveImport<>(...)`.
3. Override `removeImportVisitor` in `KotlinImportService` to return a new `org.openrewrite.kotlin.RemoveImport`, mirroring `org.openrewrite.kotlin.AddImport`.

**Pros:** mirrors `AddImport` exactly; removes all `isKotlin` source-path sniffing from `rewrite-java`; uses the type system rather than a filename heuristic; Kotlin-specific logic (facade convention, supertype walk) lives in `rewrite-kotlin`.

**Cons:** more moving parts (new SPI method + new class); `RemoveImport` carries a lot of shared logic, so the Kotlin subclass would either duplicate it or require extracting shared helpers into a common base.

timtebeek:
### Alternative 3 — fix it at type attribution in the Kotlin parser

The root cause this PR works around is how Kotlin attributes the relevant types:

- top-level Kotlin functions get a `null` `owningClass` and a `<File>Kt` facade `declaringType`, even though they are imported by their package-qualified name (`org.example.one`);
- members inherited by a companion object (e.g. from an implemented interface) aren't matched against the import without walking supertypes.

If `KotlinTypeMapping` attributed these so that the **existing** `RemoveImport` logic already matched (e.g. the top-level function carried its package-qualified name, inherited members resolved through the supertype), then no change to `RemoveImport` would be needed at all.

**Pros:** fixes the data rather than every consumer; the most "correct" location — any recipe relying on these types benefits, not just `RemoveImport`.

**Cons:** highest blast radius — touches type attribution used everywhere, so regression risk is much larger; and the facade↔import-name mapping is arguably a presentation concern that may not belong on the type itself.

timtebeek:
I've pushed up the alternative that does not have any Kotlin specific handling built into this recipe that operates across a growing number of languages. 
- https://github.com/openrewrite/rewrite/pull/8159

Thanks for helping explore the problem space!

PR #5862, opened 2025-08-06
Imports from companion objects should have owningClass information

jevanlingen:
## What's changed?
Non-star imports in Kotlin now retain not only the type of the imported object (e.g., a companion object), but also its supertypes. This ensures that full type information is available for the imported member.

## What's your motivation?
When analyzing a Kotlin file, type information such as `cu.getTypesInUse().getUsedMethods()` or `cu.getTypesInUse().getVariables()` will refer to the method or variable as declared on its actual type. This can be the declaring class, a superclass, or an interface.

In Kotlin, this becomes problematic when importing a member from a companion object that implements an interface. The import itself only carries the type of the companion object, not of its owning class. As a result, the type of the used method and the type of the imported symbol don’t match, making it impossible to correlate the import with the usage.

By including the owning class information on imported members, this allows recipes to match imports against used methods and variables, even when they’re inherited from a supertype.

- https://github.com/openrewrite/rewrite/pull/5860/

### Checklist


knutwannheden (approved):
Nice find!

PR #5865, opened 2025-08-06
UpgradeTransitiveDependencyVersion should only compare name when comparing the config with the GradleDependencyConfiguration

jevanlingen:
## What's changed?
When the GradleDependencyConfiguration is compared, it should not take the version into account. This PR replaces by a `c.allExtendsFrom().contains(config)` call to a loop where only the _name_ is used for comparison.

## What's your motivation?
With the change of
- https://github.com/openrewrite/rewrite/pull/5794

the recipe turned into a scanning recipe, which means when in the scanning phase, no changes to the code have been made. If this recipe is run multiple times by another recipe (so still one recipe run), this has an impact on the resolution. Instead of the versions already been upgraded, it can differ between runs and scans. 

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -267,10 +267,16 @@ public class UpgradeTransitiveDependencyVersion extends ScanningRecipe<UpgradeTr
                                                     for (GradleDependencyConfiguration config : all.keySet()) {
                                                         outer:
                                                         for (GradleDependencyConfiguration gc : c.allExtendsFrom()) {
-                                                            if (gc.getName().equals(config.getName()) && gc.getRequested().size() == config.getRequested().size()) {
+                                                            if (gc.getName().equals(config.getName()) &&
+                                                                    gc.getRequested().size() == config.getRequested().size()) {
+                                                                // Check if all dependencies match, only looking at groupId and artifactId
                                                                 for (int i = 0; i < gc.getRequested().size(); i++) {
-                                                                    if (!Objects.equals(gc.getRequested().get(i).getGroupId(), config.getRequested().get(i).getGroupId()) ||
-                                                                        !gc.getRequested().get(i).getArtifactId().equals(config.getRequested().get(i).getArtifactId())) {
+                                                                    if (!Objects.equals(
+                                                                            gc.getRequested().get(i).getGroupId(),
+                                                                            config.getRequested().get(i).getGroupId()) ||
+                                                                        !Objects.equals(
+                                                                                gc.getRequested().get(i).getArtifactId(),
+                                                                                config.getRequested().get(i).getArtifactId())) {
                                                                         continue outer;
                                                                     }
                                                                 }
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -241,7 +241,7 @@ public class UpgradeTransitiveDependencyVersion extends ScanningRecipe<UpgradeTr
                             }
                         }
                         for (ResolvedDependency resolved : configuration.getResolved()) {
-                            if (resolved.getDepth() > 0 && dependencyMatcher.matches(resolved.getGroupId(), resolved.getArtifactId(), resolved.getVersion())) {
+                            if (resolved.isTransitive() && dependencyMatcher.matches(resolved.getGroupId(), resolved.getArtifactId(), resolved.getVersion())) {
                                 try {
                                     String selected = versionSelector.select(resolved.getGav(), configuration.getName(), version, versionPattern, ctx);
                                     if (selected == null || resolved.getVersion().equals(selected)) {
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -265,8 +265,17 @@ public class UpgradeTransitiveDependencyVersion extends ScanningRecipe<UpgradeTr
                                                     }
 
                                                     for (GradleDependencyConfiguration config : all.keySet()) {
-                                                        if (c.allExtendsFrom().contains(config)) {
-                                                            return true;
+                                                        outer:
+                                                        for (GradleDependencyConfiguration gc : c.allExtendsFrom()) {
+                                                            if (gc.getName().equals(config.getName()) && gc.getRequested().size() == config.getRequested().size()) {
+                                                                for (int i = 0; i < gc.getRequested().size(); i++) {
+                                                                    if (!Objects.equals(gc.getRequested().get(i).getGroupId(), config.getRequested().get(i).getGroupId()) ||
+                                                                        !gc.getRequested().get(i).getArtifactId().equals(config.getRequested().get(i).getArtifactId())) {
+                                                                        continue outer;
+                                                                    }
+                                                                }
+                                                                return true;
+                                                            }
                                                         }
 
                                                         // TODO there has to be a better way!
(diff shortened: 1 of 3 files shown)


jevanlingen:
Example of different versions, taken from a test downstream:

image

timtebeek:
Discussed with Jonathan; we're going to look to change the test to use a declarative recipe, document why we chose particular versions and what was expected when combining those, see if we can compare the configurations only by name, as opposed to looking at their contained dependencies.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
```diff
@@ -265,8 +265,23 @@ public TreeVisitor<?, ExecutionContext> getScanner(DependencyVersionState acc) {
...
+                                                                for (int i = 0; i < gc.getRequested().size(); i++) {
+                                                                    if (!Objects.equals(
+                                                                            gc.getRequested().get(i).getGroupId(),
+                                                                            config.getRequested().get(i).getGroupId()) ||
+                                                                        !Objects.equals(
+                                                                                gc.getRequested().get(i).getArtifactId(),
+                                                                                config.getRequested().get(i).getArtifactId())) {
+                                                                        continue extendsFrom;
+                                                                    }
+                                                                }
```
timtebeek:
```suggestion
                                                        for (GradleDependencyConfiguration gc : c.allExtendsFrom()) {
                                                            if (gc.getName().equals(config.getName())) {
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
```diff
@@ -1284,6 +1287,74 @@ void noChangesIfDependencyIsAlsoPresentOnProjectForVersionInSettings() {
...
+                  UpgradeTransitiveDependencyVersion upgradeTransitive2 = new UpgradeTransitiveDependencyVersion("com.fasterxml.jackson.core", "jackson-databind", "2.15.1", null, null, null);
+                  UpgradeTransitiveDependencyVersion.DependencyVersionState accTransitive2 = upgradeTransitive2.getInitialValue(ctx);
+                  upgradeTransitive2.getScanner(accTransitive2).visit(tree, ctx);
+
+                  tree = upgrade2.getVisitor(acc2).visit(tree, ctx);
+                  tree = upgradeTransitive2.getVisitor(accTransitive2).visit(tree, ctx);
+
+                  return super.visit(tree, ctx);
+              }
+          })),
```
timtebeek: This ought to help readability, as well as make it easier to document the scenario we are after, which is hard to deduce from the constructors, and hard to maintain when options change.
```suggestion
          spec -> spec.recipeFromYaml("""
            ---
            type: specs.openrewrite.org/v1beta/recipe
            name: org.openrewrite.gradle.UpgradeDependencyVersionAndTransitiveTwiceTest
            displayName: Upgrade dependency and transitive twice
            description: |-
              TODO Document what scenario we are testing here with these two recipes.
            recipeList:
              - org.openrewrite.gradle.UpgradeDependencyVersion:
                  groupId: org.openrewrite
                  artifactId: rewrite-java
                  newVersion: 8.0.0
              - org.openrewrite.gradle.UpgradeTransitiveDependencyVersion:
                  groupId: com.fasterxml.jackson.core
                  artifactId: jackson-databind
                  newVersion: 2.12.3
              # TODO Document why we're running these again, and why we then expect the configuration change to
              # `implementation('com.fasterxml.jackson.core:jackson-databind:2.15.1')`
              - org.openrewrite.gradle.UpgradeDependencyVersion:
                  groupId: org.openrewrite
                  artifactId: rewrite-java
                  newVersion: 8.0.1
              - org.openrewrite.gradle.UpgradeTransitiveDependencyVersion:
                  groupId: com.fasterxml.jackson.core
                  artifactId: jackson-databind
                  newVersion: 2.15.1
            """,
            "org.openrewrite.gradle.UpgradeDependencyVersionAndTransitiveTwiceTest"),
```

timtebeek: Interestingly when we apply this change, the output changes to
```
implementation('com.fasterxml.jackson.core:jackson-databind:2.19.2')
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
timtebeek: The thinking here is that it's enough to compare just the names of configurations, as those ought to be unique.

timtebeek:
@jevanlingen could you have another look at the suggestions here, and figure out why we see changes to the expected output?

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
jevanlingen: **TLDR**: No, let's keep it in Java code.

---

I tried to do test this scenario with the Yaml declarative recipe as first, just like you suggested. But then it came to me this is not the exactly the setup I am after! If you use a declarative recipe, the recipe are separated entities, which will be executed after each other:

https://github.com/openrewrite/rewrite/blob/24b27e105f9bfa696ca5bf85643cc55158701e06/rewrite-core/src/main/java/org/openrewrite/RecipeScheduler.java#L74-L84

Using one recipe with constructors is the setup we should use, because this is equally to a proprietary recipe downstream which had this failure. It's better to resemble the test scenario as close as the real problem.

Why the outcome of the declarative vs java-inline setups differ though is something I should think about a little (so I am going to do that today). I am with you that "you should expect it to do the same". So I'll come back to you once I have an explanation.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
jevanlingen: Yes, that works as well (also tested this on the failing recipe downstream). Awesome, much more understandable without the extra clutter!

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersionTest.java
jevanlingen: Ah, it got it! In the Java code, I did:
1. scan (UpgradeDependencyVersion)
1. scan (UpgradeTransitiveDependencyVersion)
2. visit (UpgradeDependencyVersion)
3. visit (UpgradeTransitiveDependencyVersion)
4. scan (UpgradeDependencyVersion)
4. scan (UpgradeTransitiveDependencyVersion)
5. visit (UpgradeDependencyVersion)
6. visit (UpgradeTransitiveDependencyVersion)

Now, when you use the declarative recipes, you would end up with:
1. scan (UpgradeDependencyVersion)
2. visit (UpgradeDependencyVersion)
3. scan (UpgradeTransitiveDependencyVersion)
4. visit (UpgradeTransitiveDependencyVersion)
5. scan (UpgradeDependencyVersion)
6. visit (UpgradeDependencyVersion)
7. scan (UpgradeTransitiveDependencyVersion)
8. visit (UpgradeTransitiveDependencyVersion)

Notice the difference. In the Java code the scanning phases apply to the same tree for both UpgradeDependencyVersion and UpgradeTransitiveDependencyVersion. For the declarative recipes, every recipe run scans and changes the tree, before the next recipe runs. This can lead to a different outcome of course!

jevanlingen: Took me a while to figure it really out, but actually the answer lies already in the linked RecipeScheduler. If you look closely, you would see:

```
after = cycle.scanSources(after); 
after = cycle.editSources(after); 
```

That `after` variable is the LargeSourceSet. With other words, in the declarative recipe all the recipes in the recipe-list are first scanned and then the editing phase starts. In the example test this is not the case. It first scans and visits and then uses this output in the second UpgradeTransitiveDependencyVersion run.

Now I really understand what's going on, I could even make the test more concise and clear: https://github.com/openrewrite/rewrite/pull/5865/commits/05b87fe1e0b445fd7850508f1c1270e180540b26. So in the end the "doing it twice" did not have anything to do with it!

jevanlingen:
@timtebeek I think all the things we discussed are in (or out 😇). If more needs to be done, let me know. Otherwise, you may merge this PR!

timtebeek (approved):
Thanks for the changes here; much clearer as compared to the earlier approach we had here, and good to fix the downstream issue (finally).

PR #5872, opened 2025-08-07
Ensure Maven parser uses last-wins strategy for duplicate dependencies

Jenson3210:
## What's changed?
Using a `Map<GroupArtifact, DependencyAndDependent>` instead of a `List<DependencyAndDependent>` we can use the `put` method to overwrite earlier occurring requested `DependencyAndDependent` objects in the pom. 
**TODO**: verify with @sambsnyd if we do not want to use the `merge` method and introduce a ResolutionStrategy to differentiate between Maven's last-wins and Gradle's latest-wins.

## What's your motivation?
Maven resolves duplicate dependencies by using the last declaration. Previously, our parser did not follow this behavior, causing issues where only the first occurrence was updated by recipes, but the effective version was still determined by the last. This change updates the parser logic to ensure that, when multiple dependencies with the same coordinates are declared, the last one takes precedence, matching Maven's resolution strategy.

@sambsnyd 

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -980,7 +980,7 @@ public class ResolvedPom {
             Scope dScope = Scope.fromName(d.getScope());
             if (dScope == scope || dScope.transitiveOf(scope) == scope) {
                 // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
-                //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for HIGHEST_VERSION_WINS/GRADLE
+                //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for LATEST_WINS/GRADLE
                 rootDependencies.put(d.getGav().asGroupArtifact(), new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
             }
         }
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -974,17 +974,20 @@ public class ResolvedPom {
                                                         MavenPomDownloader downloader, ExecutionContext ctx) throws MavenDownloadingExceptions {
         List<ResolvedDependency> dependencies = new ArrayList<>();
 
-        List<DependencyAndDependent> dependenciesAtDepth = new ArrayList<>();
+        Map<GroupArtifact, DependencyAndDependent> rootDependencies = new HashMap<>();
         for (Dependency requestedDependency : getRequestedDependencies()) {
             Dependency d = getValues(requestedDependency, 0);
             Scope dScope = Scope.fromName(d.getScope());
             if (dScope == scope || dScope.transitiveOf(scope) == scope) {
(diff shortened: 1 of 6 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -974,17 +974,20 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
...
         List<ResolvedDependency> dependencies = new ArrayList<>();
 
-        List<DependencyAndDependent> dependenciesAtDepth = new ArrayList<>();
+        Map<GroupArtifact, DependencyAndDependent> rootDependencies = new HashMap<>();
         for (Dependency requestedDependency : getRequestedDependencies()) {
             Dependency d = getValues(requestedDependency, 0);
             Scope dScope = Scope.fromName(d.getScope());
             if (dScope == scope || dScope.transitiveOf(scope) == scope) {
-                dependenciesAtDepth.add(new DependencyAndDependent(requestedDependency, Scope.Compile, null, requestedDependency, this));
+                // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
```
Jenson3210, line 982: @sambsnyd, could you have a look at this TODO and let me know how I should continue (with/without adding a strategy enum method parameter)
If we go for with, do we want to have a new method that uses a default here to avoid breaking compilation. What should be the default: last_wins (=maven) or latest_wins (=gradle)?

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -3970,6 +3971,134 @@ void jaxbRuntime() {
         );
     }
 
+    @ParameterizedTest
```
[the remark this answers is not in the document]
timtebeek: The bot get's confused; I'm guessing it wants you to sort your annotations as per the best practices.

Jenson3210: @timtebeek, Are we always using alphabetic sorting? 
IMO it does not make sense to not put these ParameterizedTest related annotations together. 
Imagine having different annotations. The parameterization context should be grouped together and the value should be in the same location always, no?

Some examples
```java
@CsvSource
@ParameterizedTest
```
```java
@ParameterizedTest
@ValueSource
```
```java
@CsvSource({"value1", "value6"}
@CustomerSpecificAnnotation
@CustomTestFrameworkAnnotation
@DataSet(before = """
    {
        tableName: [
            {col1: value2, col2: value3},
            {col1: value4, col2: value5},
        ]
    }
    """)
@ParameterizedTest
void insertsStateIntoDatabase(String whereAreTheValues)
```

Would be cleaner imo if the stuff for parameterized tests is always the same and have that block as the last annotations in the list (the source is easy to map to the parameters that way)
```java
@CustomerSpecificAnnotation
@CustomTestFrameworkAnnotation
@DataSet(before = """
    {
        tableName: [
            {col1: value2, col2: value3},
            {col1: value4, col2: value5},
        ]
    }
    """)
@ParameterizedTest
@CsvSource({"value1", "value6"}
//@ValueSource
void testMethod(String someCsvValue)
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
sambsnyd: @Jenson3210 I think this change is good, and I am also in favor of adding a strategy enum defaulting to Maven's behavior. I've been planning to lean on Maven dependency resolution even more for Gradle dependency model updates, and being able to opt-in to newest-wins rather than nearest-wins would be helpful.

sambsnyd (approved):

PR #5873, opened 2025-08-07
Read `settings.xml` from disk if no maven settings are defined for testing framework

jevanlingen:
## What's changed?
If no maven settings is defined when Maven dependencies will be downloaded when a `rewriteRun` is performed, the local settings.xml file from the Maven configuration directory (_~/.m2/settings.xml_ by default) is used.

## What's your motivation?
The settings.xml file is often used to define mirrors, which can redirect all artifact requests to an internal repository manager such as Artifactory. For example:

```xml
<mirrors>
  <mirror>
    <id>artifactory</id>
    <mirrorOf>*</mirrorOf>
    <url>https://my.own.artifactory.com</url>
  </mirror>
</mirrors>
```

For some company environments, direct access to external repositories such as https://repo.maven.apache.org/maven2 is blocked (often returning a 403). Without support for settings.xml, recipe runs could fail in such environments because dependencies cannot be resolved.

Additionally, this change supports scenarios where mirrors are used to discard certain repositories, for example:

```xml
<mirror>
  <id>skip-sonatype-nexus-snapshots</id>
  <mirrorOf>sonatype-nexus-snapshots</mirrorOf>
  <url>http://nonexistent.invalid</url>
</mirror>
```

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/Assertions.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/Assertions.java
@@ -124,7 +124,7 @@ public class Assertions {
     }
 
     public static SourceSpecs buildGradleKts(@Language("kotlin") @Nullable String before, @Language("kotlin") @Nullable String after,
-                                             Consumer<SourceSpec<K.CompilationUnit>> spec) {
+                                          Consumer<SourceSpec<K.CompilationUnit>> spec) {
         SourceSpec<K.CompilationUnit> gradle = new SourceSpec<>(K.CompilationUnit.class, "gradle", gradleParser, before, s -> after);
         gradle.path("build.gradle.kts");
         spec.accept(gradle);
@@ -174,7 +174,7 @@ public class Assertions {
     }
 
     public static SourceSpecs settingsGradleKts(@Language("kotlin") @Nullable String before, @Language("kotlin") @Nullable String after,
-                                                Consumer<SourceSpec<K.CompilationUnit>> spec) {
+                                             Consumer<SourceSpec<K.CompilationUnit>> spec) {
         SourceSpec<K.CompilationUnit> gradle = new SourceSpec<>(K.CompilationUnit.class, "gradle", gradleParser, before, s -> after);
         gradle.path("settings.gradle.kts");
         spec.accept(gradle);


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -139,6 +140,9 @@ private MavenPomDownloader(Map<Path, Pom> projectPoms, HttpSender httpSender, Ex
         this.projectPomsByGav = projectPomsByGav(projectPoms);
         this.httpSender = httpSender;
         this.ctx = MavenExecutionContextView.view(ctx);
+        if (ctx.getMessage(MAVEN_SETTINGS_LOAD_FROM_DISK, false)) {
```
jevanlingen: I now made it work with this flag. You could argue though, if there is no maven-settings available in the context, you always want to read the maven settings from local. What do you guys think?

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -378,6 +379,7 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
         recipeCtx = CursorValidatingExecutionContextView.view(recipeCtx)
                 .setValidateCursorAcyclic(TypeValidation.before(testMethodSpec, testClassSpec).cursorAcyclic())
                 .setValidateImmutableExecutionContext(TypeValidation.before(testMethodSpec, testClassSpec).immutableExecutionContext());
+        recipeCtx.putMessage(MAVEN_SETTINGS_LOAD_FROM_DISK, true);
```
Jenson3210: Do we also have to set this in the gradle/maven plugins or do they already do `MavenSettings.readMavenSettingsFromDisk(ctx)`?

jevanlingen: I guess they don't have to, because these plugins will just start running the tests. The tests **_themselves_** call the `rewriteRun` functions, which in turn will just set the `MAVEN_SETTINGS_LOAD_FROM_DISK` property. With other words, it does not really matter how/which build tool starts the tests, as you don't need any configuration from outside the scope of the tests.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
sambsnyd: Let's do that. It's a sensible default behavior and then we don't need to toggle this boolean just right.

rewrite-core/src/main/java/org/openrewrite/ExecutionContext.java
```diff
@@ -40,7 +40,7 @@ public interface ExecutionContext extends RpcCodec<ExecutionContext> {
     String RUN_TIMEOUT = "org.openrewrite.runTimeout";
     String REQUIRE_PRINT_EQUALS_INPUT = "org.openrewrite.requirePrintEqualsInput";
     String SCANNING_MUTATION_VALIDATION = "org.openrewrite.test.scanningMutationValidation";
-
+§
```
timtebeek: Never seen this before; can we revert this?

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -219,6 +219,10 @@ public MavenRepository getMavenLocal() {
         return mavenLocal;
     }
 
+    public boolean hasDefinedConfiguration() {
```
jevanlingen: Maybe there is a better way to do this. If so please inform me!

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -139,10 +139,13 @@ private MavenPomDownloader(Map<Path, Pom> projectPoms, HttpSender httpSender, Ex
         this.projectPomsByGav = projectPomsByGav(projectPoms);
         this.httpSender = httpSender;
         this.ctx = MavenExecutionContextView.view(ctx);
+        if (this.ctx.getSettings() == null) {
+            this.ctx.setMavenSettings(MavenSettings.readMavenSettingsFromDisk(ctx));
+        }
         this.mavenSettings = this.ctx.getSettings();
```
knutwannheden: I am a bit unsure about this part. To me it would seem more natural if the code running the recipe (like the build plugin, CLI, and test harness) were required to initialize the Maven settings properly, rather than default it here. Otherwise other consumers (like [MavenParser](https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java#L103-L106)) would also need to be updated to do this defaulting. Any thoughts on this?

rewrite-maven/src/main/java/org/openrewrite/maven/MavenExecutionContextView.java
```diff
@@ -253,6 +270,10 @@ public MavenExecutionContextView setMavenSettings(@Nullable MavenSettings settin
         return effectiveSettings;
     }
 
+    public boolean hasNoDefinedConfiguration() {
+        return getSettings() == null && getLocalRepository().equals(MAVEN_LOCAL_DEFAULT) && getRepositories().isEmpty() && getActiveProfiles().isEmpty() && getMirrors().isEmpty();
+    }
```
timtebeek: While I understand adding this here, can we instead inline this in MavenPomDownloader? The reason is this class is parent classloaded, so then it's a more complicated coordinated roll out.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
jevanlingen: Yeah, I see your point. It has everything to do with: https://github.com/openrewrite/rewrite/pull/5873#discussion_r2259642565. But I actually think it's a good thing that this default is used, when nothing is set at all. It means that other consumers, like MavenParser, use the settings.xml file, unless something else has been set up. Seems to be the most logical choice.

I cleaned it up a bit though, so this default will only be applied when:
- no settings is giving 
- local repository == .m2
- repo's, active repo's and mirrors are all empty

 So basically when nothing is set at all, the default kicks in. I think by having all these checks, we are good to go. You can also see this by my changes, I did not need to change any unit test anymore.

timtebeek: I believe the use case here was uniquely for tests right? Because we already have some older handling that we could adapt here:
https://github.com/openrewrite/rewrite/blob/bb7e9579d08aa9a4dd8585d80d7135993a6571c9/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java#L154-L157
https://github.com/openrewrite/rewrite/blob/93d992cd46c759cee607ba5e772770fce9e84b1e/rewrite-maven/src/main/java/org/openrewrite/maven/Assertions.java#L34-L43

Might make sense to do away with the System property and (only) in Assertions load settings.xml from disk if nothing has been configured.

jevanlingen: > I believe the use case here was uniquely for tests right? 

Yes, it started this way. Currently it does a lot more, like this thread describes 😇.

But I was not aware of this setting in the tests at all! But to be clear, do you want to remove the added code:

```diff
- if (nothingConfigured) {
-     // Load Maven settings to pick up any mirrors, proxies etc. that might be required in corporate environments
-     this.ctx.setMavenSettings(MavenSettings.readMavenSettingsFromDisk(ctx));
- }
```

and just alter the Maven Assertion? So put this kind of code there?

knutwannheden: I suspect that if the `MavenParser` were to get called directly, then the `MavenPomDowloader` might get called too late.

I think we should either make it the caller's responsibility to initialize the Maven settings or we could instead maybe put the defaulting inside `MavenExecutionContextView`.

jevanlingen: Make sense to let the caller be responsible. Let's tone down this PR again, to only reflect the tests: https://github.com/openrewrite/rewrite/pull/5873/commits/2c461033bb8135fba56b92faf3ea9e8d8a6738f8.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -150,12 +150,6 @@ void maybeDecryptPasswords(ExecutionContext ctx) {
         }
     }
 
-
-    public static boolean readFromDiskEnabled() {
-        final String propertyValue = System.getProperty("org.openrewrite.test.readMavenSettingsFromDisk");
-        return propertyValue != null && !"false".equalsIgnoreCase(propertyValue);
-    }
```
timtebeek: Apologies; on second thought best to keep this one (maybe deprecated), because it's parent class loaded as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/Assertions.java
```diff
@@ -17,27 +17,47 @@
...
+        boolean nothingConfigured = mctx.getSettings() == null &&
+                                    mctx.getLocalRepository().equals(MAVEN_LOCAL_DEFAULT) &&
+                                    mctx.getRepositories().isEmpty() &&
+                                    mctx.getActiveProfiles().isEmpty() &&
+                                    mctx.getMirrors().isEmpty();
+        if (nothingConfigured) {
+            // Load Maven settings to pick up any mirrors, proxies etc. that might be required in corporate environments
+            mctx.setMavenSettings(MavenSettings.readMavenSettingsFromDisk(mctx));
+        }
+    }
```
timtebeek: Why this addition here? I'd aim to keep the changes small, and we've had no requests that I know of to load Maven settings for Gradle, and it seems odd.

timtebeek: I'd revert this change to gradle/Assertions.java for now.

jevanlingen: > Why this addition here? I'd aim to keep the changes small, and we've had no requests that I know of to load Maven settings for Gradle, and it seems odd.

Because Gradle also uses the settings.xml file (and thus also uses the MavenPomDownloader). To be honest, I was kind of surprised this `customizeExecutionContext` only exists for Maven. This does not make any sense to me, as both Gradle and Maven use the same artifactories etc.

timtebeek: That may be, but since this was reported as an issue running tests specifically for Maven let's minimize the changes here. We do not expect our Gradle users to have a requirement of using particular corporate `~/.m2/settings.xml`, nor do we want to suggest they add one for now.

PR #5876, opened 2025-08-07
Prevent dependencies blocks in Gradle kotlin build files

Laurens-W:
## What's changed?

No longer add dependencies blocks to anything that is not inside a method invocation

## What's your motivation?

Saw dependencies block being added to functions defined in gradle kotlin build files.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -241,10 +241,8 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
                     }
 
                     private boolean isTopLevel(Cursor cursor) {
-                        // If we're visiting a block in a Kotlin Gradle build file, we should only
-                        // mark it as top level when the direct parent is the compilation unit.
-                        if (cursor.firstEnclosing(K.CompilationUnit.class) != null && cursor.getValue() instanceof J.Block) {
-                            return cursor.getParentOrThrow().getValue() instanceof K.CompilationUnit;
+                        if (cursor.getValue() instanceof J.Block) {
+                            return cursor.getParentOrThrow().getValue() instanceof JavaSourceFile;
                         }
                         return cursor.getParentOrThrow().firstEnclosing(J.MethodInvocation.class) == null;
                     }


Laurens-W:
Now [this](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/JvmTestSuiteTest.java#L348-L409) test is failing

rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/AddDependencyVisitor.java
```diff
@@ -115,6 +115,19 @@ public class AddDependencyVisitor extends JavaIsoVisitor<ExecutionContext> {
...
+                try {
+                    getCursor().dropParentUntil(value -> value instanceof J.MethodInvocation &&
+                                                         ("testing".equals(((J.MethodInvocation) value).getSimpleName()) ||
+                                                         "sourceSets ".equals(((J.MethodInvocation) value).getSimpleName()) ||
+                                                         "configurations ".equals(((J.MethodInvocation) value).getSimpleName()))
+                            );
+                } catch (IllegalStateException e) {
+                    return b;
+                }
+            }
```
shanman190: So I wonder if it may be better to fine tune the `isTopLevel` check within `gradle.AddDependency` rather than including those additional rules here.

You'll have a bit more control there as well since that recipe is _only_ designed to add a dependency to the immediate project's `dependencies` block or add a block if one is missing.

Laurens-W: That makes sense, I've moved the logic and that also allowed me to get rid of the check on `MethodInvocation` names completely

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1645,6 +1645,67 @@ void kotlinDslMatchesDependencyDeclarationStyle() {
         );
     }
 
+    @Test
+    void doNotAddDependencyToFunctionsInKotlinGradle() {
```
jevanlingen, line 1649: Just to be sure, add another test without a dependencies-block at all.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -240,6 +241,11 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
                     }
 
                     private boolean isTopLevel(Cursor cursor) {
+                        // If we're visiting a block in a Kotlin Gradle build file, we should only
+                        // mark it as top level when the direct parent is the compilation unit.
+                        if (cursor.firstEnclosing(K.CompilationUnit.class) != null && cursor.getValue() instanceof J.Block) {
+                            return cursor.getParentOrThrow().getValue() instanceof K.CompilationUnit;
+                        }
```
jevanlingen: I have been wondering, do we actually want to do a Kotlin check here. This works fine as well:


```suggestion
                        if (cursor.getValue() instanceof J.Block) {
                            return cursor.getParentOrThrow().getValue() instanceof JavaSourceFile;
                        }
```

If you read this code, it does look like a logical check. Plus, it's a little bit faster, as you don't have to do the `firstEnclosing` call.

Laurens-W: I quickly checked whether this "hurts" any of the groovy gradle files and it doesn't seem to, so I agree, let's simplify it further

jevanlingen (approved):

PR #5880, opened 2025-08-07
ChangeDependencyGroupIdAndArtifactId should not limit updating managed dependencies

jevanlingen:
## What's changed?
The ChangeDependencyGroupIdAndArtifactId recipe does always update managed dependencies, regardless of whether a new version is requested or not.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-openapi/issues/46

To be more specific, the [org.openrewrite.openapi.swagger.UseJakartaSwaggerArtifacts](https://github.com/openrewrite/rewrite-openapi/blob/eaac2145713a44dd40a22026f4aca5e85272d15b/src/main/resources/META-INF/rewrite/swagger-2.yml#L52-L63) recipe has listed as one of the steps:

```yaml
  - org.openrewrite.java.dependencies.ChangeDependency:
      oldGroupId: io.swagger.core.v3
      oldArtifactId: swagger-annotations
      newArtifactId: swagger-annotations-jakarta
```

This does not work for managed dependencies without this fix.

--

On top of that, this fix also brings this recipe on par with its Gradle's counterpart:

https://github.com/openrewrite/rewrite/blob/8266018254b1d41045c2b0195cfa82f34b8fbee2/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java#L198-L200

## Any additional context
This  change could lead to downstream failing tests! As the managed dependencies will now also be changed when no version is given, running recipes that use ChangeDependencyGroupIdAndArtifactId / ChangeDependency without a `newVersion` or `versionPattern` will behave a little differently, which could lead to a different outcome. We should fix this by either fixing the tests or setting the `changeManagedDependency` option to false in such cases.

### Checklist


timtebeek:
Is this still relevant given the conflicts we see on those same lines @jevanlingen ?

jevanlingen:
Yep, it's still relevant. These changes you see have everything to do with the recipe flipping between a Recipe and a ScanningRecipe. This PR is just a loosening of an if-check, with results in the ChangeManagedDependencyGroupIdAndArtifactId recipe be called more often.

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -36,7 +36,9 @@ void changeDependencyGroupIdAndArtifactId() {
             "jakarta.activation",
             "jakarta.activation-api",
             null,
-            null
+            null,
+            false,
+            false
```
timtebeek, line 41: What has me slightly worried is that we're changing these values here from what they were (implied `false, true`), to now explicitly setting the second to `false` to keep the test passing. That to me signals that we're changing the behavior in a way that might surprise folks using the explicit constructor, as we for instance do here:
https://github.com/openrewrite/rewrite-migrate-java/blob/05fa3e0a9e228a4d5568702e36ac62ee0d4cf889/src/main/java/org/openrewrite/java/migrate/javax/AddJaxbRuntime.java#L221-L224

Can you explain why this change in default is needed, and what changes we should be making downstream as a result?

timtebeek: What I'm mostly trying to look out for is that we don't inadvertently start to see more changes that we hadn't anticipated or desired following this change; and that's hard to say with just a single test added and the defaults flipped on three more. Just trying to make sure we can move forward with this with confidence.

timtebeek: from what it looks like we're only sparsely using that recipe from Yaml; with no changes there expected (version was already used)
https://github.com/search?q=org%3Aopenrewrite+ChangeDependencyGroupIdAndArtifactId+language%3AYAML&type=code&l=YAML

timtebeek: Looks like this is the only downstream usage of the outdated constructor:
- https://github.com/openrewrite/rewrite-migrate-java/pull/838

jevanlingen: > What has me slightly worried is that we're changing these values here from what they were (implied false, true), to now explicitly setting the second to false to keep the test passing. 

Yes I can explain this. The tests that I changed do not take managed dependencies into account. Thus by setting these managed properties to explicit false, the behaviour stays the same. This is in line with the thing I mentioned in the starting post:

> As the managed dependencies will now also be changed when no version is given, running recipes that use ChangeDependencyGroupIdAndArtifactId / ChangeDependency without a newVersion or versionPattern will behave a little differently, which could lead to a different outcome.

PR #5881, opened 2025-08-08
Add the concept of an "attribute" to Dependency.

sambsnyd:
We use maven Dependency to model Gradle dependencies, but Gradle and Maven allocate information differently. Maven has a separate dependency management section, whereas in Gradle BOMs are intermixed with regular dependencies and differentiated via the "category" attribute.

Being unable to reliably differentiate `platform()` and project dependencies from binary dependencies has limited how accurate our model updates can be. `UpgradeDependencyVersion` and `UpgradeTransitiveDependencyVersion` try their best but without this information inaccuracies add up. Particularly when many dependency version changes are performed together, as is the case with major migrations or dependency vulnerability check.

Right now dependency vulnerability check will tell you a gradle project that uses `platform()` has a vulnerability if the BOM references a vulnerable dependency, even if that dependency isn't actually used in the project. Because to our model a BOM currently just looks like any other dependency.


shanman190:
I think this may be a bit too narrow... Gradle attributes are typically a `Map<String, String>`. I've honestly wondered if we should have specific dependency types and be modeling more closely to Gradle and this feels like it shifts us further away.

As an illustrative example:

* `ProjectDependency` (has project path)
* `ExternalModuleDependency` (GAVC)
* etc

A `platform(..)` is an `ExternalModuleDependency` with the `org.gradle.platform` attribute set.

There are a ton of other attributes though which would certainly be useful for other contexts.

rewrite-gradle/src/main/java/org/openrewrite/gradle/attributes/ProjectAttribute.java
```diff
@@ -0,0 +1,43 @@
...
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.ProjectDependency;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.maven.tree.DependencyAttribute;
+
+/**
+ * Marks a dependency as representing another Gradle project within the same multi-project build
+ */
+@Value
+public class ProjectAttribute implements DependencyAttribute {
```
shanman190: Presently, `project` dependencies aren't even represented in the dependency graph. Should we rather create a `ProjectDependency` with the path directly on it? Then we could introduce a interface to establish the higher order type signature to allow us to have represent both projects and Maven dependencies in configurations.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/DependencyAttribute.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven.tree;
+
+import java.io.Serializable;
+
+public interface DependencyAttribute extends Serializable {
```
shanman190: This should probably just be generalized to `Attribute` because these can also exist on `Configuration` instances.

rewrite-gradle/src/main/java/org/openrewrite/gradle/attributes/ProjectAttribute.java
sambsnyd: I don't want to introduce a subclass of `Dependency` right now. I would have to make `Dependnecy` non-final and a non-value class

rewrite-maven/src/main/java/org/openrewrite/maven/tree/DependencyAttribute.java
sambsnyd: Fair enough

PR #5882, opened 2025-08-08
Prevent duplicate rows in Gradle DependencyInsight

timtebeek:
### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
@@ -19,6 +19,7 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
+import org.gradle.internal.impldep.org.bouncycastle.oer.its.ieee1609dot2.LinkageData;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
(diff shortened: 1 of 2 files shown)


jevanlingen:
Yep, looking good! I am only wondering about the maven counterpart. If maven's recipe can also have this same problem, I would add it in this PR as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
```diff
@@ -19,6 +19,7 @@
 import lombok.EqualsAndHashCode;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
+import org.gradle.internal.impldep.org.bouncycastle.oer.its.ieee1609dot2.LinkageData;
```
jevanlingen:
```suggestion
```

timtebeek:
Adding some context here: the test uses rewrite-core, which indeed has three copies of jackson-core coming in
```
dependencies {
    implementation 'org.openrewrite:rewrite-core:7.39.1'
}
```
image

all three result in the same data table row before the changes made here.

jevanlingen (approved):

jkschneider (changes requested):
I’m pretty sure this is a non-goal in fact. Dependencies can come to be in multiple ways in a file. And there is utility in knowing how many times a dependency is included

timtebeek:
As discussed in Slack: we'll instead change the Maven recipe to match Gradle, as opposed to changing the Gradle recipe to match Maven. We need to update the tests accordingly and the Maven DependencyInsight recipe. In the future we might add a column to distinguish between the data table rows produced.

PR #5886, opened 2025-08-10
add new recipe to add gitignore entries

zieka:


rewrite-core/src/test/java/org/openrewrite/AddToGitignoreTest.java
```diff
@@ -0,0 +1,258 @@
...
+              """,
+            """
+              *.class
+              
+              local.properties
+              """,
+            spec -> spec.path("src/.gitignore")
+          )
+        );
+    }
```
timtebeek: How would we want to handle more than one `.gitignore` file in a repository? Because I could see there being one at the root, and then multiple in separate source sets. I don't think we'd want to just add to all, possibly only at the root.

rewrite-core/src/test/java/org/openrewrite/AddToGitignoreTest.java
```diff
@@ -0,0 +1,258 @@
...
+              *.log
+              
+              # OS files
+              .DS_Store
+              """,
+            spec -> spec.path(".gitignore")
+          )
+        );
+    }
+}
```
timtebeek, line 488: It might be nice to avoid adding entries already matched by existing patterns.
```suggestion
    }

    @Test
    void doNotAddSuperfluousEntries() {
        rewriteRun(
          spec -> spec.recipe(new AddToGitignore("foo.tmp")),
          text(
            """
              *.tmp
              """,
            spec -> spec.path(".gitignore")
          )
        );
    }
}
```

timtebeek:
Let me know if you'd want to finish this yourself, or would rather hand over to me or someone on my team.

rewrite-core/src/test/java/org/openrewrite/AddToGitignoreTest.java
zieka: good point

zieka:
> Let me know if you'd want to finish this yourself, or would rather hand over to me or someone on my team.

Hey just saw this; I think I addressed your comments.  I don't mind if someone from your team wants to continue on whatever would remain.

rewrite-core/src/main/java/org/openrewrite/AddToGitignore.java
```diff
@@ -0,0 +1,293 @@
...
+            private boolean matchesFilePattern(String sourcePath, String pattern) {
+                // Simple pattern matching for common cases
+                if (".gitignore".equals(pattern)) {
+                    // Root .gitignore only
+                    return ".gitignore".equals(sourcePath);
+                }
+                if ("**/.gitignore".equals(pattern)) {
+                    // All .gitignore files
+                    return true;
+                }
```
timtebeek: Looks like this is covered by the conditions below.
```suggestion
```

PR #5887, opened 2025-08-10
Extract common RewriteTest defaults for RecipeSpec

timtebeek:


timtebeek:
Dot test failure unrelated.

Issue #5889, opened 2025-08-11
KotlinParser does not allow for a classpath definition


evanchooly:
## What version of OpenRewrite are you using?

I am using rewrite-recipe-bom version 3.11.1.

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.

```
mvn -e -U \
     org.openrewrite.maven:rewrite-maven-plugin:run \
		 -Drewrite.recipeArtifactCoordinates=dev.morphia.morphia:morphia-rewrite:3.0.0-SNAPSHOT \
		 -Drewrite.activeRecipes=dev.morphia.UpgradeToMorphia30,dev.morphia.InternalOnly
```

## What is the smallest, simplest way to reproduce the problem?

A PR with the fix and a test is following right after this.

## What did you expect to see?

```java
class A {
    void foo(String bar) {
        int i = 5;
    }
}
```

## What did you see instead?

```java
class A {
    void foo(String bar) {
        int i = 5;
    }
}
```

## What is the full stack trace of any errors you encountered?

```
java.lang.IllegalStateException: LST contains missing or invalid type information
Identifier->MemberReference->FieldAccess->MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(Identifier type is missing or malformed)~~>*/Document

MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(MethodInvocation type is missing or malformed)~~>*/AggregationOptions()

	at org.openrewrite.kotlin.Assertions.assertValidTypes(Assertions.java:242)
	at org.openrewrite.kotlin.Assertions.validateTypes(Assertions.java:57)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:512)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:130)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:125)
	at dev.morphia.rewrite.recipes.test.pipeline.kotlin.KotlinDatastoreAggregateStringTest.methodCall(KotlinDatastoreAggregateStringTest.java:102)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
[...]

PR #5890, opened 2025-08-11
Make the parser settable on KotlinTemplate

evanchooly:
## What's changed?
This makes `KotlinTemplate#parser()` public so the parser can be set externally.  Does some, admittedly tangential, clean up to reduce duplication between `KotlinTemplate` and `JavaTemplate`.

## What's your motivation?

This fixes 
- #5889 

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyTemplate.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyTemplate.java
@@ -70,19 +70,10 @@ public class GroovyTemplate extends JavaTemplate {
     @SuppressWarnings("unused")
     public static class Builder extends JavaTemplate.Builder {
 
-        private final String code;
-        private final Set<String> imports = new HashSet<>();
-
         private GroovyParser.Builder parser = GroovyParser.builder();
-
-        private Consumer<String> onAfterVariableSubstitution = s -> {
-        };
-        private Consumer<String> onBeforeParseTemplate = s -> {
-        };
 
         Builder(String code) {
             super(code);
-            this.code = code;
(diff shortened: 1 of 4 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyTemplate.java
```diff
@@ -99,33 +88,11 @@ public Builder imports(String... fullyQualifiedTypeNames) {
...
-            if (StringUtils.isBlank(typeName)) {
-                throw new IllegalArgumentException("Imports must not be blank");
-            } else if (typeName.startsWith("import ")) {
-                throw new IllegalArgumentException("Imports are expressed as fully-qualified names and should not include an \"import \" prefix");
-            } else if (typeName.endsWith(";") || typeName.endsWith("\n")) {
-                throw new IllegalArgumentException("Imports are expressed as fully-qualified names and should not include a suffixed terminator");
-            }
-        }
-
         Builder parser(GroovyParser.Builder parser) {
```
shanman190: We should probably make this public while we're at it.

```suggestion
        public Builder parser(GroovyParser.Builder parser) {
```

shanman190 (commented):
I think the changes overall is good. Initially, the templates were focused on JDK primitives and contexts where type information already wasn't available, if I'm recalling correctly.

shanman190 (approved):

timtebeek:
Thanks for the suggestion and review both! We're exposing mostly the same API here as before, just unblocking a common use case that was skipped previously, but needed for any real work with Kotlin/Groovy. Merging to get this into the next release.

PR #5893, opened 2025-08-11
Stable label order for `PrintMavenAsDot`

timtebeek:
We were seeing some flaky tests lately, possibly following:
- https://github.com/openrewrite/rewrite/pull/5821

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/PrintMavenAsDot.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/PrintMavenAsDot.java
@@ -71,8 +71,13 @@ public class PrintMavenAsDot extends Recipe {
                 dotLabel(dot, 0, root);
                 index.put(root, 0);
 
-                for (List<ResolvedDependency> deps : mrr.getDependencies().values()) {
-                    for (ResolvedDependency dep : deps) {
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/utilities/PrintMavenAsDot.java
```diff
@@ -71,8 +71,13 @@ private String dot(MavenResolutionResult mrr) {
                 dotLabel(dot, 0, root);
                 index.put(root, 0);
 
-                for (List<ResolvedDependency> deps : mrr.getDependencies().values()) {
-                    for (ResolvedDependency dep : deps) {
+                // Build up index of all dependencies, so we can reference them by index in the DOT output
+                for (Scope scope : Scope.values()) {
```
steve-aom-elliott, line 75: Would it make sense to limit it to `Compile`, `Runtime`, `Provided` and `Test` like the loop lower down that is calling `dotEdges(..)` or would that not change things?

timtebeek: Yes sure, makes sense. Also rendered the big disabled test SVG:
sb dot

steve-aom-elliott (approved):

PR #5896, opened 2025-08-11
Fix property resolution for dependencies with remote parent POMs

Jenson3210:
## Summary
Fixes property resolution when upgrading dependency versions that are defined via properties in remote parent POMs.

### MavenVisitor.java
- Enhanced `changeChildTagValue()` method to correctly handle property resolution for dependencies that inherit properties from remote parent POMs
- Added logic to check if a property is defined in the current POM or inherited from a parent POM that is not in the same source tree (i.e., remote parent)
- The fix ensures that when a dependency version is defined via a property in a remote parent POM, the recipe can still update it correctly

### UpgradeDependencyVersionTest.java
- Enabled previously disabled test `upgradeVersionDefinedViaExplicitPropertyInRemoteParent()`
- This test validates that dependencies using properties from remote parent POMs (like Spring Boot's `spring-boot-dependencies`) can be upgraded successfully

## Technical Details
The issue occurred when trying to upgrade dependencies whose versions were controlled by properties defined in remote parent POMs. The previous implementation only checked if the property existed in the requested POM's properties, missing cases where properties were inherited from remote parents. 

The fix now also checks:
1. If the property exists in the resolved POM's properties
2. If the parent POM is remote (no relative path specified)

This allows proper handling of common patterns like Spring Boot's dependency management where version properties are defined in the remote `spring-boot-dependencies` parent POM.

## Testing
- Re-enabled and passing test case that validates upgrading Flyway from version defined in Spring Boot parent
- All existing tests continue to pass

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -980,7 +980,7 @@ public class ResolvedPom {
             Scope dScope = Scope.fromName(d.getScope());
             if (dScope == scope || dScope.transitiveOf(scope) == scope) {
                 // TODO can we always use the Map.put approach? Using the latest one is Maven specific, but this resolving is also used for gradle which does use highest version.
-                //  We could introduce a ResolutionStrategy to handle this and use Map.merge where we take later occurring one for LAST_WINS/MAVEN and higher version one for HIGHEST_VERSION_WINS/GRADLE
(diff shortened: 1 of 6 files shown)


timtebeek:
@Jenson3210 I pressed the merge main button and now this PR is empty; can it be closed?

Jenson3210:
Weird, perhaps I did accept one prompt to much from claude as that was most certainly not my branch :) 
Recreated a correct PR at 
- https://github.com/openrewrite/rewrite/pull/5899

PR #5899, opened 2025-08-11
Fix property resolution for dependencies with remote parent POMs

Jenson3210:
## Summary
Fixes property resolution when upgrading dependency versions that are defined via properties in remote parent POMs.

### MavenVisitor.java
- Enhanced `changeChildTagValue()` method to correctly handle property resolution for dependencies that inherit properties from remote parent POMs

### UpgradeDependencyVersionTest.java
- Enabled previously disabled test `upgradeVersionDefinedViaExplicitPropertyInRemoteParent()`

## Technical Details
The issue occurred when trying to upgrade dependencies whose versions were controlled by properties defined in remote parent POMs. The previous implementation only checked if the property existed in the requested POM's properties, missing cases where properties were inherited from remote parents. 

The fix now also checks:
1. If the property exists in the resolved POM's properties
2. If the parent POM is remote (no relative path specified)


timtebeek:
For that last question
> I am not 100% sure if the null parent pom path is sufficient of a check to know that properties will not be set twice (parent and pom)

Would we be able to try this out with a unit test, or by running the recipe at scale through the CLI before we merge? Just to avoid rework later.

Jenson3210:
For the one with the spring boot parent it is null in the unit test. 

I will do a cli run tomorrow on more projects. What would that look like. 
How would I find projects with a parent outside sourcerepo scope to know the behavior in these ones.

timtebeek:
> I will do a cli run tomorrow on more projects. What would that look like. How would I find projects with a parent outside sourcerepo scope to know the behavior in these ones.

I'd expect the "Spring Guides" org to have a lot of parents defined outside their immediate source sets ; and fairly straightforward to bump some spring dependencies on that set and see how that affects the properties introduced.

timtebeek:
Hmm; I had a bit of an awkward bit of code I was surprised to have to do in a recipe earlier today; I wonder if this PR would have helped, or if we need to look more broadly at what would help to pull properties from external parent pom files:
- https://github.com/openrewrite/rewrite/pull/5831/files#diff-99af8635016d35d218a5c6d08a7c2ced8570ca8a8e413220503a6b39354d3387R89-R108

Jenson3210:
This pr would not have helped. The goal of this pr is :

In an external pom that has a property
```xml
<properties>
    <dep.version>1.2</dep.version>
</properties>
```

And a local pom that refers to this property
```xml
<dependency>
    <>...</>
    <version>${dep.version}</version>
</dependency>
```

The property cannot be bumped in the external one -> we need to add it in the current one, effectively overwriting the parent's property

Jenson3210:
@timtebeek I ran some cli tests on Spring guides but there is one problem there. 
They do not really use properties from the parent pom

Manually modified sources of one of them locally to include a dependency that uses the `lombok` version property and tested that this also works as expected. 
I could not verify the code itself due to MavenVisitor not loading my changes correctly in cli, but I verified using a breakpoint and manually checking the if conditions , then calling `doAfterVisit` that the result would be okay. 

The if statement would evaluate to false as the current requestPom does not have the property
Screenshot 2025-08-13 at 13 30 40

else-if will be true
Screenshot 2025-08-13 at 13 31 27

getParent is null as it is a remote parent
Screenshot 2025-08-13 at 13 32 07

Calling the doAfterVisit in evaluation
Screenshot 2025-08-13 at 13 33 20

results in the change being made
Screenshot 2025-08-13 at 13 28 17

Jenson3210:
But I also feel that the tests added show that we will bubble up to the external one and then add the version if missing. so this test behavior is the same as the cli behavior

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -292,9 +292,24 @@ protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullabl
...
+                        while (resolutionResult.getParent() != null) {
+                            resolutionResult = resolutionResult.getParent();
+                            if (resolutionResult.getPom().getRequested().getProperties().containsKey(propertyName)) {
+                                inheritsPropertyFromExternalParent = false;
+                                break;
+                            }
+                        }
+                        if (inheritsPropertyFromExternalParent) {
+                            doAfterVisit((TreeVisitor<?, P>) new ChangePropertyValue(propertyName, newValue, true, false).getVisitor());
+                        }
```
timtebeek, line 317: Would this be an accurate addition to explain why we are not making changes in this case?
```suggestion
                        }
                        // Otherwise, expect to change the property in the parent project in the same source tree
```

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -292,9 +292,24 @@ protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullabl
             String oldValue = childTag.get().getValue().orElse(null);
             if (newValue != null && !newValue.equals(oldValue)) {
                 if (isProperty(oldValue)) {
+                    MavenResolutionResult resolutionResult = getResolutionResult();
                     String propertyName = oldValue.substring(2, oldValue.length() - 1);
-                    if (getResolutionResult().getPom().getRequested().getProperties().containsKey(propertyName)) {
+                    ResolvedPom pom = resolutionResult.getPom();
+                    //If the current requested pom has the property defined, or if the pom has the property defined and it is inherited from a parent pom that is not in the same source tree
```
timtebeek: I had a little trouble mentally parsing this comment and in the implications for the code below.

Can we instead add comments just before the `doAfterVisit` calls below with something along the lines of
```
// The property is defined in an external parent; we need to override it here to get the changed value we're after
doAfterVist(...)
```
Note that I don't know if the above comment is true, as I had a little trouble following along.

Jenson3210: What do you think about these comments?

timtebeek (approved):
Thanks a lot!

bmuschko:
@Jenson3210 @timtebeek Is this ready to be merged?

Jenson3210:
Not on pc atm, but I thought this was in the list of core changes to review. If you could check, otherwise I will verify tomorrow

bmuschko:
@timtebeek approved the changes end of last week. That's why I am wondering. Tomorrow is fine. Thanks!

Jenson3210:
@bmuschko I verified this issue is in the list of core-changes that needs additional review

PR #5902, opened 2025-08-12
Allow ChangeDependencyGroupIdAndArtifactId to update in a chainable way keeping the ScanningRecipe in place

Jenson3210:
### Summary
This PR enhances the `ChangeDependencyGroupIdAndArtifactId` recipe to correctly handle sequential transformations of the same dependency within a single recipe run. The fix uses the `ExecutionContext` to pass state between different recipe scanning phases, enabling later recipe executions to detect when a dependency will already have been modified by an earlier recipe in the chain by the time this visitor receives it.

### Problem
When multiple `ChangeDependencyGroupIdAndArtifactId` recipes are chained together in a single recipe run (e.g., transforming A→B then B→C), the second recipe would fail to find and transform the dependency because it was still looking for the original coordinates rather than the updated ones from the first transformation.

### Solution
The implementation now leverages the `ExecutionContext` to store transformation state between recipe scanning phases:

1. **Scanning Phase**: During the initial scanning phase, each recipe records which dependencies it will transform in the `ExecutionContext`
2. **State Tracking**: This context is passed between different recipe runs, allowing subsequent recipes to check if a dependency tag will be modified by an earlier recipe

#### ChangeDependencyGroupIdAndArtifactId.java
- Added `ExecutionContext` state management to track transformations across recipe executions
- Implemented logic to detect and handle dependencies that have been modified by earlier recipes in the chain
- Ensures proper coordination between scanning and visitation phases
- Cleaned up the Accumulator a bit to use the same object.

#### ChangeDependencyGroupIdAndArtifactIdTest.java  
- Added comprehensive test `canChangeSameDependencyMultipleTimesInASingleRecipeRun()` that validates a sequential transformation of bouncycastle dependencies: `bcprov-jdk15on` → `bcprov-jdk15to18` → `bcprov-jdk18on`

### Technical Details
The `ExecutionContext` serves as a communication channel between recipe phases, storing metadata about planned transformations. This allows the recipe framework to maintain awareness of in-flight changes, ensuring that subsequent recipes in a chain operate on the most current state rather than the original document state. Note that this is not the change that I originally intended / am very happy with, but other attempts to use cursor messages etc have all proved not to be working. 

### Testing
- New test validates complex transformation chains work correctly
- Ensures recipes can handle both single transformations and sequential multi-step transformations
- All existing tests continue to pass, confirming backward compatibility

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -16,8 +16,7 @@
 package org.openrewrite.maven;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import lombok.EqualsAndHashCode;
-import lombok.Value;
+import lombok.*;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.maven.table.MavenMetadataFailures;
@@ -30,6 +29,8 @@ import org.openrewrite.xml.tree.Xml;
 
 import java.nio.file.Path;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.util.Collections.max;
 import static org.openrewrite.Validated.required;
@@ -40,6 +41,7 @@ import static org.openrewrite.internal.StringUtils.isBlank;
 @EqualsAndHashCode(callSuper = false)
 public class ChangeDependencyGroupIdAndArtifactId extends ScanningRecipe<ChangeDependencyGroupIdAndArtifactId.Accumulator> {
     transient MavenMetadataFailures metadataFailures = new MavenMetadataFailures(this);
+    private static final String CHANGED_DEPENDENCIES = "ChangeDependencyGroupIdAndArtifactId.ScanningMap";
 
     @Option(displayName = "Old groupId",
             description = "The old groupId to replace. The groupId is the first part of a dependency coordinate `com.google.guava:guava:VERSION`. Supports glob expressions.",
@@ -145,6 +147,7 @@ public class ChangeDependencyGroupIdAndArtifactId extends ScanningRecipe<ChangeD
 
     @Override
     public Accumulator getInitialValue(ExecutionContext ctx) {
+        ctx.computeMessageIfAbsent(CHANGED_DEPENDENCIES, key -> new HashMap<Dependency, Dependency>());
         return new Accumulator();
     }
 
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -163,63 +166,79 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
                                 }
                             } else if (configuredToOverrideManageVersion || !newDependencyManaged) {
                                 // If the version is not present, add the version if we are explicitly overriding a managed version or if no managed version exists.
-                                acc.createVersion = resolvedNewVersion;
+                                acc.createVersion = true;
                             }
+                            changedDependency = changedDependency.withVersion(resolvedNewVersion);
                         } catch (MavenDownloadingException e) {
-                            return e.warn(tag);
+                            throw new RuntimeException("Failed to resolve version for " + changedDependency, e);
```
Jenson3210, line 232: Unrelated change -> I adapted this as the e.warn modifies the lst and then I had tests fail with a warning indicating that scanners cannot modify the lst

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -289,6 +308,7 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
                             Optional.ofNullable(newArtifactId).orElse(oldArtifactId),
                             newVersion, versionPattern).getVisitor());
                 }
+                ctx.pollMessage(CHANGED_DEPENDENCIES);
```
Jenson3210, line 310: Is polling the message to clear the context necessary? Is there a better way to clear this message than to run it in every visitor? We can safely remove it after the all-recipe-scanning phase.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -163,63 +165,79 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
             @Override
             public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 Xml.Tag t = (Xml.Tag) super.visitTag(tag, ctx);
-                boolean isOldDependencyTag = isDependencyTag(oldGroupId, oldArtifactId);
+
+                Map<GroupArtifactVersion, GroupArtifactVersion> changedDependencies = ctx.getMessage(CHANGED_DEPENDENCIES);
```
steve-aom-elliott, line 169: Is the context not shared between visiting different POM files? What I was meaning was that if we have a record of a dependency having been changed, we may need distinguish that between different POM files in the same repo so that we don't mistakenly skip changing a dependency in a sibling module's PO.

Jenson3210: I fail to see how that is different from before. 
the accumulator is also shared between different POM files. 
I assumed that if this recipe runs, it will bump on all impacted modules making the same changes there. Or am I misunderstanding?

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -163,63 +165,79 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
+
+                GroupArtifactVersion oldDependency = new GroupArtifactVersion(oldGroupId, oldArtifactId, null);
+                for (Map.Entry<GroupArtifactVersion, GroupArtifactVersion> entry : changedDependencies.entrySet()) {
```
Jenson3210, line 172: @steve-aom-elliott Your comment makes me realize that there might be multiple bumps earlier on so that I will have to change the loop logic still a bit. (luckily the revert gives us this time)

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
steve-aom-elliott: @Jenson3210 I think what I was thinking was that `getInitialValue(..)` was called once when the recipe was invoked, whereas the visitor was instantiated per file being analyzed, and so if you were checking whether you needed to make a change based on the _presence_ of the message for a particular dependency then it might skip changing when it actually needed to do so still in the current file, but I realize now I had misread the checks, and that they should still be safe, other than what you mentioned in the comment below.

timtebeek:
We saw similar explored in
- https://github.com/openrewrite/rewrite/pull/5814
- https://github.com/openrewrite/rewrite-openapi/pull/45

Closing because of the conflicts and likelihood that this is already fixed; if not feel free to reopen / restart/

PR #5905, opened 2025-08-12
Fix whitespace preservation when removing access modifiers

dongwooooooo:
## What's changed?
Fixed whitespace and comment handling when changing method access level to package-private. The removed access modifier's Space (whitespace and comments) is now properly merged with the next modifier to maintain consistent formatting.

## What's your motivation?
Resolves spacing issues where the auto-formatter doesn't influence spacing when comments are involved. This was causing inconsistent formatting when access modifiers were removed, particularly when line comments were present.
Referenced issue pattern can be seen in rewrite-spring's BeanMethodsNotPublicTest where dataSource3 bean formatting was problematic.

- The Space merging logic in lines 118-143 that handles whitespace and comment preservation
- The test case `publicToPackagePrivateMovesLineComment()` which covers the specific scenario where line comments need to be relocated
- Validation that the fix doesn't break existing formatting behavior

Considered relying solely on auto-formatting, but as noted in the issue, auto-formatter doesn't handle spacing when comments are involved. The current approach manually preserves the spacing structure by merging Space objects.

### Checklist


timtebeek (approved):
Thanks for the helpful fix here @dongwooooooo !

PR #5910, opened 2025-08-13
Exclude ibm `icu4j` from antlr

Laurens-W:
## What's changed?

Exclude ibm icu4j dependency, as it looks like we're not using the unicode support it brings to antlr

## What's your motivation?

Vulnerability reporting brought it to attention

We could also override the version, but it seems all versions currently available are vulnerable

### Checklist

### DIFF
--- a/rewrite-java/build.gradle.kts
+++ b/rewrite-java/build.gradle.kts
@@ -37,7 +37,9 @@ dependencies {
     api("io.micrometer:micrometer-core:1.9.+")
     api("org.jetbrains:annotations:latest.release")
 
-    antlrGeneration("org.antlr:antlr4:4.13.2")
+    antlrGeneration("org.antlr:antlr4:4.13.2") {
+        exclude(group = "com.ibm.icu", module = "icu4j")
+    }
     implementation("org.antlr:antlr4-runtime:4.13.2")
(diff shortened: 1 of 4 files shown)


Laurens-W:
@greg-at-moderne am I correct in assuming it would break the build if we'd actually be reliant on this transitive depedency?

timtebeek (approved):
There's likely a gradlely way to avoid the repetition, but I'd say this is good enough for now, as any regression would pop back up in vulnerabilities.

rewrite-java/build.gradle.kts
```diff
@@ -37,7 +37,9 @@ dependencies {
     api("io.micrometer:micrometer-core:1.9.+")
     api("org.jetbrains:annotations:latest.release")
 
-    antlrGeneration("org.antlr:antlr4:4.13.2")
+    antlrGeneration("org.antlr:antlr4:4.13.2") {
```
knutwannheden, line 40: Could we instead use org.antlr:antlr4:4.13.2:complete?

Laurens-W: If I do that without the exclude it surfaces the vulnerability still. When I add the exclusion again it doesn't surface it, which is weird because as far as I can tell `complete` would mean fat jar, so no exclusions?

knutwannheden: Yes, that is odd. But then your PR is probably indeed what is simplest.

PR #5914, opened 2025-08-14
 Fix template generation to preserve statements when adding before/after coordinates

JohannisK:
## What's changed?
This PR fixes an issue where the template generator incorrectly modified statements when using JavaCoordinates to add statements before or after existing code, rather than replacing them.

## What's your motivation?
When using JavaTemplate to add a statement before an existing method invocation (e.g., using `firstStatement()` coordinates), the template generator would incorrectly prefix the new statement with variable declarations like `Object o =`. This happened because the generator didn't distinguish between REPLACEMENT mode (where such prefixes ensure valid compilation) and addition modes (BEFORE/AFTER) where the statement should remain unmodified.

For example, when adding `String newLine = "test";` before `o.get();`, the generator would produce the invalid statement: `Object o = String newLine = "test";`

Solution

- Added the JavaCoordinates.Mode parameter to the contextFreeTemplate method signature
- Modified template generation logic to only add variable declaration prefixes when in REPLACEMENT mode
- Applied the fix consistently across Java, Groovy, and Kotlin template generators


knutwannheden:
As discussed with @JohannisK we want to explore a different approach, as it doesn't seem quite right to link this behavior to the `mode`.

PR #5916, opened 2025-08-14
Improve record parsing

Laurens-W:
## What's changed?

Improve the parsing of records to match closer with the Java compiler output

## What's your motivation?

We've seen annotations with different targets end up on different elements in the compiler, but this is not reflected in the LST

## Any additional context

~Remaining TODOs~
~- [ ] Expand the logic around mapping the members, as currently we only consider `VariableTree` when parsing a record~
~- [ ] Add a marker to the LST so we can preserve the original order of the annotations~
~- [ ] Adjust the JavaPrinter so that we become print idempotent again~
Based on Knut's guidance I've simplified the approach to not require drastic changes to the printer

- Closes https://github.com/openrewrite/rewrite-spring/pull/713

### Checklist


rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -551,7 +592,33 @@ public J visitClass(ClassTree node, Space fmt) {
         if (enumSet != null) {
             members.add(enumSet);
         }
-        members.addAll(convertStatements(membersMultiVariablesSeparated));
+        // We have to determine which annotations to add to the field declarations here.
+        if (kind.getType() == J.ClassDeclaration.Kind.Type.Record) {
+            for (Tree tree : membersMultiVariablesSeparated) {
+                if (tree instanceof VariableTree vt) {
```
Laurens-W: Only checking for `VariableTree` skips a lot of other elements

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -2460,3 +2535,9 @@ private void addPossibleEmptyStatementsBeforeClosingBrace(List<JRightPadded<Stat
         }
     }
 }
+
+record RecordParam(VariableTree varDecl,
```
Laurens-W: Perhaps this record should have utility methods to produce the list of `J.Annotation` applicable to either the `VariableTree` or the `JCVariableDecl`

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -420,19 +422,49 @@ public J visitClass(ClassTree node, Space fmt) {
...
+                }
                 if (member instanceof VariableTree vt) {
                     if (hasFlag(vt.getModifiers(), Flags.RECORD)) {
-                        stateVector.add(vt);
+                        Space varDecsPrefix = whitespace();
+                        List<J.Annotation> recordAnnotations = collectAnnotations(
+                                mapAnnotations(vt.getModifiers().getAnnotations(), recordAnnotationPosTable.getOrDefault(vt.getName(), new HashMap<>()))
+                        );
+                        Space typeExpressionPrefix = whitespace();
+                        JRightPadded<Statement> varDec = this.<Statement>convert(vt, t -> { Space suffix = whitespace(); skip(","); skip(")"); return suffix; })
```
jevanlingen:
```suggestion
                        JRightPadded<Statement> varDec = this.<Statement>convert(vt, __ -> { Space suffix = whitespace(); skip(","); skip(")"); return suffix; })
```

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -420,19 +422,49 @@ public J visitClass(ClassTree node, Space fmt) {
...
+            }
+
             primaryConstructor = JContainer.build(
-                    sourceBefore("("),
-                    stateVector.isEmpty() ?
-                            singletonList(padRight(new J.Empty(randomId(), sourceBefore(")"), Markers.EMPTY), EMPTY)) :
-                            convertAll(stateVector, commaDelim, t -> sourceBefore(")")),
+                    prefix,
+                    ListUtils.map(varDecs, elem -> {
+                        if (elem != null && elem.getElement() instanceof J.VariableDeclarations vd) {
```
jevanlingen:
```suggestion
                        if (elem.getElement() instanceof J.VariableDeclarations vd) {
```

No need (again 😜).

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -420,19 +422,49 @@ public J visitClass(ClassTree node, Space fmt) {
 
         JContainer<Statement> primaryConstructor = null;
         if (kind.getType() == J.ClassDeclaration.Kind.Type.Record) {
-            List<Tree> stateVector = new ArrayList<>();
+            Map<String, List<J.Annotation>> recordParams = new HashMap<>();
+            Space prefix = sourceBefore("(");
+            Map<Name, Map<Integer, JCAnnotation>> recordAnnotationPosTable = new HashMap<>();
+            List<JRightPadded<Statement>> varDecs = new ArrayList<>();
```
jevanlingen: You need to rewrite some more, but I think it's clearer if you use:
```suggestion
            List<JRightPadded<J.VariableDeclarations>> varDecs = new ArrayList<>();
```
here.

Laurens-W: I don't think I can, because there is a path where `varDecs` remains empty and a `J.Empty` is added, which is a `Statement` but not a `J.VariableDeclarations`

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -877,9 +909,17 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
-        cursor(endPos(node));
+        int endPos = endPos(node);
         Object value = node.getValue();
-        String valueSource = source.substring(((JCLiteral) node).getStartPosition(), endPos(node));
+
+        if (endPos == Position.NOPOS) {
+            String str = value + "";
+            int delimLength = source.substring(cursor).indexOf(str);
+            endPos = cursor + delimLength + str.length() + delimLength;
+        }
```
knutwannheden: This hack doesn't always work. For example when you have an `int` or `long` typed literal like `1_000` or `1_000L` it will break. Can we somehow skip past the token?

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -1780,6 +1820,7 @@ public J visitWildcard(WildcardTree node, Space fmt) {
      * Conversion utilities
      * --------------
      */
+    @Contract("null -> null")
```
knutwannheden:
```suggestion
    @Contract("null -> null; !null -> !null")
```

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
knutwannheden: While it is doable it does require quite a few more changes. Questionable if worth it or not.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -420,21 +423,53 @@ public J visitClass(ClassTree node, Space fmt) {
...
+                skip(")");
+                //noinspection unchecked
+                primaryConstructor = JContainer.build(
+                        prefix,
+                        varDecls.stream().map(varDecl ->
+                                (JRightPadded<Statement>) (JRightPadded<?>) varDecl.withElement(varDecl.getElement()
+                                        .withLeadingAnnotations(recordParams.get(varDecl.getElement().getVariables().get(0).getSimpleName())))
+                        ).collect(toList()),
+                        Markers.EMPTY
+                );
```
jevanlingen:
```suggestion
                List<JRightPadded<Statement>> elements = (List<JRightPadded<Statement>>) (List<?>) ListUtils.map(varDecls, it ->
                        it.withElement(it.getElement()
                                .withLeadingAnnotations(recordParams.get(it.getElement().getVariables().get(0).getSimpleName()))
                        ));
                primaryConstructor = JContainer.build(prefix, elements, Markers.EMPTY);
```

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -849,9 +884,31 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
-        String valueSource = source.substring(((JCLiteral) node).getStartPosition(), endPos(node));
+
+        if (endPos == Position.NOPOS) {
+            if (typeMapping.primitive(((JCLiteral) node).typetag) == JavaType.Primitive.String) {
+                if (source.substring(cursor).startsWith("\"\"\"")) {
+                    cursor += 3;
+                    endPos = cursor + source.substring(cursor).indexOf("\"\"\"") + 3; // +3 for the closing delimiter
+                } else {
+                    endPos = cursor + value.toString().length() + 2; // +2 for the quotes
+                }
```
jevanlingen:
```suggestion
                int quote = source.substring(cursor).startsWith("\"\"\"") ? 3 : 1;
                if (quote == 3) {
                    endPos = cursor + quote + source.substring(cursor + 3).indexOf("\"\"\"") + quote;
                } else {
                    endPos = cursor + quote + value.toString().length() + quote;
                }
```

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -849,9 +884,31 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
+            } else {
+                endPos = cursor + indexOf(source.substring(cursor),
+                        ch ->
+                                ch.equals(',') || ch.equals(';') || ch.equals(')') || ch.equals(']') || ch.equals('}') ||
+                                ch.equals('+') || ch.equals('-') || ch.equals('*') || ch.equals('/') || ch.equals('%') ||
+                                ch.equals('=') || ch.equals('!') || ch.equals('<') || ch.equals('>') || ch.equals('&') ||
+                                ch.equals('|') || ch.equals('^') || ch.equals('?') || ch.equals(':') || ch.equals('.') ||
+                                Character.isWhitespace(ch)
+                );
```
jevanlingen:
```suggestion
                endPos = cursor + indexOf(source.substring(cursor), c ->
                        Character.isWhitespace(c) || Set.of(',', ';', ')', ']', '}', '+', '-', '*', '/', '%', '=', '!', '<', '>', '&', '|', '^', '?', ':', '.').contains(c)
                );
```

Probably a little bit slower, but way more readable. I don't think we have to much to think about performance here, as 99% of all the literals do have a defined end position (so `endPos(node) > 1`).

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -2141,8 +2200,16 @@ private int positionOfNext(String untilDelim, @Nullable Character stop) {
         return delimIndex > source.length() - untilDelim.length() ? -1 : delimIndex;
     }
 
-    private final Function<Tree, Space> semiDelim = ignored -> sourceBefore(";");
-    private final Function<Tree, Space> commaDelim = ignored -> sourceBefore(",");
+    private final Function<Tree, Space> semiDelim = ignored -> {
```
jevanlingen:
```suggestion
    private final Function<Tree, Space> semiDelim = __ -> {
```

The `__` is a preference I see being used more and more in our codebase.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
knutwannheden: Avoid calling `String#substring()`. Instead you can probably work with the `String#indexOf(String, int)` method.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
knutwannheden: AFAIK the JVM isn't smart enough to only call `Set.of()` once. Its escape analysis should allow it to allocate on the stack, but if we want to do it this way, I suggest creating a constant. Another simple and readable way would be `",;)]}+-*/%=!<>&|^?:.".indexOf(c) != -1`.

jevanlingen: Your way is even better!

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
Laurens-W: I adopted the ListUtil approach but left it inlined as that seems to be preferred if no further changes are made to the resulting list

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -849,9 +884,26 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
+                }
+            } else {
+                endPos = cursor + indexOf(source.substring(cursor),
+                        ch -> Character.isWhitespace(ch) || ",;)]}+-*/%=!<>&|^?:.".indexOf(ch) != -1
+                );
+            }
+        }
+
+        cursor(endPos);
+        String valueSource = source.substring(((JCLiteral) node).getStartPosition(), endPos);
```
Laurens-W: The reason this had to change at all is because the endpos table is computed in this step https://github.com/openrewrite/rewrite/blob/04628ec7de2429a6d61cc909667aec418d2378fd/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17Parser.java#L224
and the removal of unapplicable annotations is done here https://github.com/openrewrite/rewrite/blob/04628ec7de2429a6d61cc909667aec418d2378fd/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17Parser.java#L236 without updating that endpos table, so we can no longer reliable get that information

rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
```diff
@@ -877,9 +912,26 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
+            if (typeMapping.primitive(((JCLiteral) node).typetag) == JavaType.Primitive.String) {
+                int quote = source.substring(cursor).startsWith("\"\"\"") ? 3 : 1;
+                if (quote == 3) {
+                    endPos = cursor + quote + source.indexOf("\"\"\"", cursor + quote) + quote;
+                } else {
+                    endPos = cursor + quote + value.toString().length() + quote;
+                }
+            } else {
+                endPos = cursor + indexOf(source.substring(cursor),
+                        ch -> Character.isWhitespace(ch) || ",;)]}+-*/%=!<>&|^?:.".indexOf(ch) != -1
```
knutwannheden: If we want to use this `indexOf()` utility, I think we should add an overload that takes an additional `int fromIndex` parameter, so that we can get rid of this `substring()` call.

rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
```diff
@@ -292,9 +292,20 @@ public static int indexOfNonWhitespace(String text) {
...
+     * @param fromIndex Start offset to scan from
+     * @param test The predicate to match
+     * @return The index of the first character for which the predicate returns <code>true</code>,
+     * or <code>-1</code> if no character in the string matches the predicate.
+     */
+    public static int indexOf(String text, int fromIndex, Predicate<Character> test) {
+        for (int i = fromIndex; i < text.length(); i++) {
             if (test.test(text.charAt(i))) {
-                return i;
+                return i - fromIndex;
```
knutwannheden: I think returning `i` would be more appropriate and should typically simplify the calling code.

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -849,9 +884,23 @@ public J visitLambdaExpression(LambdaExpressionTree node, Space fmt) {
...
         Object value = node.getValue();
-        String valueSource = source.substring(((JCLiteral) node).getStartPosition(), endPos(node));
+
+        if (endPos == Position.NOPOS) {
+            if (typeMapping.primitive(((JCLiteral) node).typetag) == JavaType.Primitive.String) {
+                int quote = source.startsWith("\"\"\"", cursor) ? 3 : 1;
+                int elementLength = quote == 3 ? source.indexOf("\"\"\"", cursor + quote) - cursor - quote  : value.toString().length();
+                endPos = cursor + quote + elementLength + quote;
+            } else {
+                endPos = cursor + indexOf(source, cursor,
```
knutwannheden: After the proposed cgange:

```suggestion
                endPos = indexOf(source, cursor,
```

timtebeek:
@Laurens-W should any of the improvements here now be replicated into the Java 25 parser since this wasn't merged yet?

Laurens-W:
> @Laurens-W should any of the improvements here now be replicated into the Java 25 parser since this wasn't merged yet?

Yes, they probably should, I'll update the PR!

rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java
```diff
@@ -26,6 +26,8 @@
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.*;
 import com.sun.tools.javac.util.Context;
+import com.sun.tools.javac.util.Position;
+import org.jetbrains.annotations.Contract;
```
jkschneider, line 30: Is this the right time to introduce `Contract`?

tubbynl:
tnx for the hard work; i'll retest https://github.com/openrewrite/rewrite-spring/issues/714 when this get's released :)

tubbynl:
re-tested with `rewrite-maven-plugin:6.19.0` and `rewrite-spring:6.15.0 `and can confirm it now parses this correctly and does not yield errors :)

Laurens-W:
Thanks for reporting back @tubbynl!
If you find any other issues feel free to report them 👍🏻

PR #5918, opened 2025-08-15
Allow `AddOrUpdateAnnotationAttribute` to be chained after `ChangeType` and other recipes that produce shallow types

BoykoAlex:
If annotation type is `ShallowClass` without methods a product of a type change from a previous recipe than `AddOrUpdateAnnotationAttribute` recipe doesn't add/change the attribute


BoykoAlex:
@steve-aom-elliott Sorry about that. Added the unit test. The fix seemed so obvious that I just felt not wasting time writing a test ;-) I do realize that the value of the test is not to break that in the future - good thing you asked for it :-)

PR #5919, opened 2025-08-15
Gradle model rework

sambsnyd:
Consolidate `GradleProject` marker updates inside of `GradleProject` and `GradleDependencyConfiguration`.
This takes advantage of fairly extensive changes to rewrite-gradle-tooling-model which finally enable us to differentiate between platform dependencies (BOMs), constraints, projects, and regular dependencies. 

Now every recipe which manipulates gradle dependencies will have a standard, terse way of expressing model changes which will all flow through a single, tested code path. This enables the deletion of chunks of buggy model update code from each of the gradle dependency recipes. I've added it to `UpgradeDependencyVersion` and `UpgradeTransitiveDependencyVersion`, the most important ones, but the others could benefit from similar updates. 

With more accurate model updates dependency manipulating recipes will compose together better and the model's accuracy will degrade less over the course of a large migration or dependency vulnerability check run.


shanman190 (commented):
I've skimmed through this and will do a more thorough analysis here after a bit.

Some initial questions/comments:

* It looks like we don't handle the buildscript `classpath` configuration correctly anymore
* Should the logic really be carried on the `GradleProject` marker or should there be something else that carries the logic? Most LST elements haven't carried domain logic while this is introducing quite a bit of it onto the model object.

sambsnyd:
> It looks like we don't handle the buildscript `classpath` configuration correctly anymore

Good point.

> Should the logic really be carried on the `GradleProject` marker or should there be something else that carries the logic? Most LST elements haven't carried domain logic while this is introducing quite a bit of it onto the model object.

`GradleProject` isn't an LST element, it's more like type information. In maven land the model updating is mostly exposed through a method on `MavenVisitor`. But there is no `GradleVisitor`, instead a mix of `GroovyVisitor`, `KotlinVisitor`, and or `JavaVisitor` are used. I can't go put gradle-specific model update methods onto `JavaVisitor` so putting the model update into the model itself feels sensible.

Issue #5400, opened 2025-05-07
Support refactoring based on an annotation that provides the replacement template (e.g. `@InlineMe`)


traceyyoshima:
## What problem are you trying to solve?
Automated transformations via type attribution and annotations.
Tim asked about automating refactoring during the KotlinParser development using deprecated annotations.

I thought it might be possible to create an OR annotation like `DeprecatedX(...)`, where the annotation arguments provide information about how to handle a migration.

## Describe the solution you'd like
- The annotations could be target-site specific, like METHOD, CLASS, etc.
- OR would own the annotation, and its existence could trigger a recipe.
- The type-attribution on a class, method, etc., may have arguments that provide contextual instructions on the migration.

```java
class Example {

    // I don't recall what is available in the annotation arguments, but I imagine something could be sorted out.
    @org.openrewrite.DeprecatedMethod("methodMatcherSig", "someTransformationToMethodB")
    void methodA(...) {...}

    // Method calls of `methodA` would be migrated to `methodB`.
    void methodB(...) {...}
}
```

## Have you considered any alternatives or workarounds?
No, this was just a thought. This may already exist, but I haven't tracked all the changes across OR for awhile.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
I'm pretty busy at the moment, but anything is possible!

timtebeek:
Hi @traceyyoshima ; thanks for the suggestion! There's been a few such discussions in the past indeed, notably around
- https://github.com/openrewrite/rewrite-kotlin/issues/25
- https://errorprone.info/docs/inlineme

The second one is more geared towards Java, but seems to have only seen limited adoption so far.

I like the suggestion to introduce our own, as then we'd be better in control of just what we need to make these work, and potentially could run a processor to generate the recipes. Should also be easier now that we capture annotation argument values as of
- https://github.com/openrewrite/rewrite/pull/4746

traceyyoshima:
> * Add annotation element values to type attribution model #4746

Ah, that's very cool. I didn't know about that update, but it does simplify the concept.

Ownership has many benefits, mainly due to OR's velocity. A lightweight package may also help spread adoption of the framework!

timtebeek:
[...]

PR #5923, opened 2025-08-16
MethodType parameter names not retained for Guava method

timtebeek:
## What's your motivation?
- https://github.com/openrewrite/rewrite/issues/5400
- https://github.com/openrewrite/rewrite-migrate-java/pull/788

### DIFF
--- a/rewrite-java-21/src/test/java/org/openrewrite/java/MethodTypeTest.java
+++ b/rewrite-java-21/src/test/java/org/openrewrite/java/MethodTypeTest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.Issue;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.test.RewriteTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.test.RewriteTest.toRecipe;
+
+class MethodTypeTest implements RewriteTest {
+    @Issue("https://github.com/openrewrite/rewrite/issues/5400")
+    @Test
+    void guavaMethodParameterNamesPresent() {
+        rewriteRun(
+          spec -> spec.parser(JavaParser.fromJavaVersion().classpath("guava")),
+          java(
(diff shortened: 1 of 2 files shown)


rewrite-java-21/src/test/java/org/openrewrite/java/MethodTypeTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openrewrite.java.Assertions.java;
+
+class MethodTypeTest implements RewriteTest {
+    @Issue("https://github.com/openrewrite/rewrite/issues/5400")
+    @Test
+    void guavaMethodParameterNamesPresent() {
+        rewriteRun(
+          spec -> spec.parser(JavaParser.fromJavaVersion().classpath("guava")),
```
timtebeek: I only wanted this test and runtime dependency added to one of the parsers, instead of all of them, hence why this test is here and not in rewrite-java-test/tck.

Since it's such a small change we can also remove this test here completely and rely on a downstream test that uses the parameter names in
- https://github.com/openrewrite/rewrite/pull/5923

knutwannheden:
I think there may have been a bug in the type table writer, so that it didn't write out parameter names. I fixed that when I added annotation support.

timtebeek:
They'd helpful, but this test still failed over the weekend and is using direct classpath dependencies right, not classpathFromResources?

knutwannheden:
> They'd helpful, but this test still failed over the weekend and is using direct classpath dependencies right, not classpathFromResources?

I didn't actually check. Your comment just reminded me of this issue.

rewrite-java-21/src/test/java/org/openrewrite/java/MethodTypeTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.Issue;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.test.RewriteTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.openrewrite.java.Assertions.java;
+
+class MethodTypeTest implements RewriteTest {
```
knutwannheden: Would it make sense to move this test to the TCK, as you change all parsers to support this?

timtebeek: I'd held off on adding it to all because I didn't want to propagate that Guava dependency, as detailed here:
- https://github.com/openrewrite/rewrite/pull/5923#discussion_r2280618501

I'm fine with removing the test here and only having that check downstream where I plan to use this:
- https://github.com/openrewrite/rewrite-migrate-java/pull/788

I don't think it's likely anyone would remove this without us noticing downstream.

knutwannheden: You could also just call `java.util.List.of()` or similar in your test. The parameters would be called `e1` and `e2` in that case. Then no classpath dependencies at all are required.

knutwannheden: Maybe not best example, as it won't work for Java 1.8. But if this isn't strictly about Guava, we can maybe use something from the JDK or perhaps also AssertJ, which is also around already.

timtebeek: Better indeed to test with assertj; this had moved up from my intended use for a Guava method annotated with `@InlineMe`, but is not specific to it.

PR #5926, opened 2025-08-18
Document `JavaParser.Builder` classpath options

timtebeek:
Documents proper use with context as to which to use when, and why we see a lot of older usage still.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
@@ -113,7 +113,7 @@ public interface JavaParser extends Parser {
     }
 
     /**
-     * Load artifacts from packaged resources. This is useful for loading dependencies that are not on the recipe
+     * Load artifacts from packaged resources. This is useful for loading dependencies which are not on the recipe
      * execution classpath, or where you need to load multiple different versions of the same artifact.
      * Supports both {@link TypeTable} `classpath.tsv.gz` and packaged resource jars in `META-INF/rewrite/classpath/`.
      *
@@ -319,8 +319,8 @@ public interface JavaParser extends Parser {
          * {@link #classpathFromResources(ExecutionContext, String...)}, with the latter preferred to limit dependencies
          * needed on the recipe runtime classpath, as the runtime classpath may differ in say the CLI or Platform.
          * <p>
-         * Fine for example when writing tests, which may need to use older versions of dependencies that need not be
-         * added to the recipe runtime classpath through resources.
+         * This is suitable, for example, when writing tests that may require older versions of dependencies,
+         * without needing to add them to the recipe runtime classpath through resources.
          *
          * @return A list of paths to jars on the runtime classpath of the process constructing the parser.
          */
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -310,13 +319,32 @@ public B classpath(String... artifactNames) {
...
+         */
         @SuppressWarnings({"UnusedReturnValue", "unused"})
         public B classpathFromResources(ExecutionContext ctx, String... classpath) {
             this.artifactNames = emptyList();
             this.classpath = dependenciesFromResources(ctx, classpath);
             return (B) this;
         }
 
+        /**
+         * Sets the classpath from runtime classpath dependencies of the process constructing the parser. Predates
```
knutwannheden: Should we maybe deprecate this method? I only found one (test) use: https://app.moderne.io/results/t4hbKe42r

timtebeek: Yes seems like that would make sense to deprecate; that test seems to refer to some old usage patterns
- https://github.com/openrewrite/rewrite/issues/3222#issuecomment-1552682975

I'll deprecate this one instead, thanks!

rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -304,19 +314,43 @@ public B addClasspathEntry(Path entry) {
...
+         * {@link #classpathFromResources(ExecutionContext, String...)}, with the latter preferred to limit dependencies
+         * needed on the recipe runtime classpath, as the runtime classpath may differ in say the CLI or Platform.
+         * <p>
+         * Fine for example when writing tests, which may need to use older versions of dependencies that need not be
+         * added to the recipe runtime classpath through resources.
```
jevanlingen:
```suggestion
         * This is suitable, for example, when writing tests that may require older versions of dependencies,
         * without needing to add them to the recipe runtime classpath through resources.
```

rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -111,6 +112,15 @@ static List<Path> filterArtifacts(String artifactName, List<Path> runtimeClasspa
         return artifacts;
     }
 
+    /**
+     * Load artifacts from packaged resources. This is useful for loading dependencies that are not on the recipe
```
jevanlingen:
```suggestion
     * Load artifacts from packaged resources. This is useful for loading dependencies which are not on the recipe
```

rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -304,19 +314,43 @@ public B addClasspathEntry(Path entry) {
...
+         * @return A list of paths to jars on the runtime classpath of the process constructing the parser.
+         */
         public B classpath(String... artifactNames) {
             this.artifactNames = Arrays.asList(artifactNames);
             this.classpath = emptyList();
             return (B) this;
         }
 
+        /**
+         * Load artifacts from packaged resources. This is useful for loading dependencies that are not on the recipe
```
jevanlingen:
```suggestion
         * Load artifacts from packaged resources. This is useful for loading dependencies which are not on the recipe
```

Issue #5796, opened 2025-07-24
ChangeParentPom calls RemoveRedundantDependencyVersions and without any provision to pass except parameter


dcsekar:
UpgradeParentVersion recipe intern uses the ChangeParentPom. And ChangeParentPom calls the RemoveRedundantDependencyVersions recipe without any exclusions list passed [here](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java#L247). This affects our use case where we are using UpgradeParentVersion, but want to call RemoveRedundantDependencyVersions only with some exclusions list.

## What version of OpenRewrite are you using?
rewrite's main branch

## How are you running OpenRewrite?
IDE - running new test in our multi-module project.

## What is the full stack trace of any errors you encountered?
The issue is straight forward and simple. UpgradeParentVersion recipe uses ChangeParentPom and it calls RemoveRedundantDependencyVersions recipe. In our use project we call UpgradeParentVersion recipe, and we have to call the RemoveRedundantDependencyVersions recipe only with the specific set of arguments (groupPattern, artifactPattern, except). But since UpgradeParentVersion call results in calling RemoveRedundantDependencyVersions with hard-coded null arguments, it brings unintentional changes.
It will be good if it's possible that an exclusion list could be propagated from UpgradeParentVersion to the embedded RemoveRedundantDependencyVersions recipe or if ChangeParentPom get rids of RemoveRedundantDependencyVersions which is not it's direct intended change.

crankydillo:
Not a bug IMO.  This is a feature request.

> RE if ChangeParentPom get rids of RemoveRedundantDependencyVersions which is not it's direct intended change.

Not sure I agree on this part.  I think you can make a case for either of these:

1. As you say, allow passing of an exclusions list from upgradeparent.. to removeredundant..,  or
2. Have a boolean that indicates removeredundant should not be used.  Leaving it to the user to add that if they want.

I think either would work.

crankydillo:
@timtebeek Quick thoughts on this?  Any easy workaround?  If adding a feature, do you prefer 1 or 2 from [this list](https://github.com/openrewrite/rewrite/issues/5796#issuecomment-3113594218)?   Or maybe another option..

crankydillo:
We are currently coding option 1 to see if it works for us.  Hope that will be considered a good feature:)

[...]

PR #5892, opened 2025-08-11
UpgradeParentVersion, ChangeParentPom to provide except to RemoveRedundantDependencyVersions

dcsekar:
## What's changed?
The UpgradeParentVersion recipe accepts the except parameter and passes it to ChangeParentPom
The ChangeParentPom recipe passes the except parameter to RemoveRedundantDependencyVersions
Enabling provision to pass exclusion list of dependencies list to RemoveRedundantDependencyVersions while calling UpgradeParentVersion.

## What's your motivation?
This change is done to address the enhance request described in [#5796](https://github.com/openrewrite/rewrite/issues/5796)

## Anything in particular you'd like reviewers to focus on?
There is requirement to upgrade the parent version in our project and also we should not remove duplicate dependencies for some set of the dependencies. Hence this change is required to handle that situation.

### Checklist
I've added unit tests to cover the cases of passing the except parameter in UpgradeParentVersionTest and ChangeParentPomTest

PR #5927, opened 2025-08-18
Propagate `except` to `RemovedundantDependencyVersions` from `ChangeParentPom`

dcsekar:
## What's changed?
The `UpgradeParentVersion` recipe accepts the except parameter and passes it to `ChangeParentPom`
The `ChangeParentPom` recipe passes the except parameter to `RemoveRedundantDependencyVersions`
Enabling provision to pass exclusion list of dependencies list to `RemoveRedundantDependencyVersions` while calling `UpgradeParentVersion`.

## What's your motivation?
This change is done to address the enhance request described in
- closes https://github.com/openrewrite/rewrite/issues/5796

There is requirement to upgrade the parent version in our project and also we should not remove duplicate dependencies for some set of the dependencies. Hence this change is required to handle that situation.

### Checklist
I've added unit tests to cover the cases of passing the except parameter in `UpgradeParentVersionTest` and `ChangeParentPomTest`

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeParentVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeParentVersion.java
@@ -21,8 +21,6 @@ import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.semver.Semver;
 
-import java.util.List;
-
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class UpgradeParentVersion extends Recipe {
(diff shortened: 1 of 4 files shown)


dcsekar:
Earlier PR [#5892](https://github.com/openrewrite/rewrite/pull/5892) got closed. Hence I created this with the same code.

crankydillo:
@timtebeek If there's any way I can help on this or the related PR (#5894), please let me know!

timtebeek:
Thanks both! I've had a look over the changes, added some polishing and made one substantial change: `UpgradeParentVersion` does (now) not take a new `except` argument, to avoid repeated hand over from one recipe to the next. If you want those semantics use `ChangeParentPom` instead, as `UpgradeParentVersion` is merely a convenience delegate. For more control use `ChangeParentPom` itself. This minimizes the possible impact, and reduces the amount of options we add to recipes, as that tends to complicate the logic over time.

PR #5931, opened 2025-08-19
Fix `ClasspathScanningLoader` when inheritance hierarchy spans jars

knutwannheden:
### Problem

The `ClasspathScanningLoader` was using `ClassGraph.acceptJars()`restrict scanning to only the target jar when loading
pes from a specific jar file. This caused ClassGraph to miss pes that indirectly extend `Recipe` or `ScanningRecipe`
ugh parent classes defined in other jars.ClassGraph's `acceptJars()` method restricts both:
1. **Class discovery** - which classes are found during scanning
2. **Reference resolution** - which classes can be referenced for inheritance checks

### Solution

Remove `acceptJars()` restriction and scan the entire classpath, but filter results during recipe processing.


knutwannheden:
I am wondering if this refactoring perhaps allows to simplify the code in `org.openrewrite.config.Environment.Builder#scanJar()`.

rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -155,37 +154,53 @@ private void scanYaml(ClassGraph classGraph, Properties properties, Collection<?
...
+                }
             }
         }
     }
 
-    private void configureRecipes(ScanResult result, String className) {
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
+    private boolean isFromJar(ClassInfo classInfo, String jarName) {
+        return classInfo.getClasspathElementURI() != null &&
+               classInfo.getClasspathElementURI().toString().contains(jarName);
```
timtebeek: Some of our recipe modules are prefixed of others; like `rewrite-java` and `rewrite-java-dependencies` or `-security`; do we see any issues with that here?

knutwannheden: Good point. I had that on my TODO list, but it got lost somehow.

sambsnyd: If that is not fixed the consequence would be that when loading rewrite-java-security in the saas or CLI it would show all the rewrite-java recipes as part of it. Which could potentially then lead to rewrite-java recipes being unloaded alongside rewrite-java-security recipes if that module is uninstalled.

knutwannheden: Using `Files#isSameFile()` now.

knutwannheden: It is not quite like that, because it includes the whole file name. So the only collision would be if two jars have the same artifact name and version, but a different group. But that is now also handled.

PR #5936, opened 2025-08-20
Upgrade annotation processor versions in Maven when using UpgradeDependencyVersion

Jenson3210:
## Summary

This PR extends Maven's `UpgradeDependencyVersion` recipe to automatically upgrade annotation processor versions in Maven's `<annotationProcessorPaths>` configuration, bringing Maven's behavior in line with how Gradle already handles annotation processor upgrades.

## Problem

Previously, when using `UpgradeDependencyVersion` to upgrade dependencies like MapStruct, the recipe would upgrade:
- Regular dependencies in `<dependencies>`
- Plugin dependencies in `<plugin><dependencies>`
- Gradle annotation processors

However, it would **not** upgrade Maven annotation processors defined in:
```xml
<annotationProcessorPaths>
  <path>
    <groupId>org.mapstruct</groupId>
    <artifactId>mapstruct-processor</artifactId>
    <version>1.4.1.Final</version>
  </path>
</annotationProcessorPaths>
```

This inconsistency meant that users had to manually update annotation processor versions in Maven projects, even when using automated dependency upgrade recipes.

## Solution

This change treats Maven's `<annotationProcessorPaths><path>` elements similarly to plugin dependencies, ensuring they are upgraded alongside regular dependencies. The implementation:

1. Adds detection for annotation processor path tags using a new `ANNOTATION_PROCESSORS_PATH_MATCHER`
2. Reuses the existing upgrade logic (renamed from `upgradePluginDependency` to `upgradeTag`) for both plugin dependencies and annotation processor paths
3. Properly handles version properties used in annotation processor declarations

## Testing

Added comprehensive tests demonstrating:
- Direct version upgrades in annotation processor paths
- Version property resolution for annotation processors
- Verification that Gradle variant of the recipe at the moment does the same thing.

## Impact

This ensures consistent dependency management across build systems, eliminating the need for manual intervention when upgrading libraries that include annotation processors (like MapStruct, Lombok, etc.).

## References

- Related discussion: https://github.com/openrewrite/rewrite-migrate-java/pull/818#issuecomment-3196376207

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
@@ -225,8 +225,8 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
                                 maybeUpdateModel();
                             }
                         }
-                    } else if (isPluginDependencyTag(groupId, artifactId)) {
-                        t = upgradePluginDependency(ctx, t);
+                    } else if (isPluginDependencyTag(groupId, artifactId) || isAnnotationProcessorPathTag(groupId, artifactId)) {
+                        t = upgradeTag(ctx, t);
                     }
                 } catch (MavenDownloadingException e) {
                     return e.warn(t);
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -373,6 +373,15 @@ private String resolveVersion(String version) {
                 return MavenDependency.findNewerVersion(groupId, artifactId, version, getResolutionResult(), metadataFailures,
                         versionComparator, ctx);
             }
+
+            public boolean isAnnotationProcessorPathTag(String groupId, String artifactId) {
```
Jenson3210: I deliberately did not add this to the MavenVisitor like the other methods, as i feel these pollute the MavenVisitor. We can add helper methods there for every kind of tag, but if there is no reuse needed, I feel this is not "big" enough to be part of the MavenVisitor to avoid ending up with a whole lot of different helper methods. 

I did reuse `isTag`, making it protected accessible.

timtebeek:
Great! @Jenson3210 can you resolve the conflict and merge this PR? Then we can update our friends.

PR #5937, opened 2025-08-20
Write Gradle Wrapper scripts with UTF-8 encoding

bmuschko:
## What's changed?

Set the encoding when writing the Gradle Wrapper scripts to UTF-8.

## What's your motivation?

Allow for properly encoding special characters used in the scripts.

@shanman190


bmuschko:
Closing. The encoding was derived of the JDK used to build the LST via `mod build`, and likely the source charset was different than UTF-8.

PR #5938, opened 2025-08-20
Avoid making HTTPS calls to services.gradle.org

shanman190:
## What's changed?
Capture additional data into Gradle versions.csv in order to avoid the early invocation to https://services.gradle.org for resolving a SemVer to a specific version.

## What's your motivation?
By making this change, it ensures that we no longer have the need to make HTTPS calls to https://services.gradle.org which may be blocked within some organizations. Instead, we can discover the information from our internalized CSV file (nightly updates to come in a followup PR) which makes the process faster and less prone to timeout errors which impact users.

Overall, this seems like a more acceptable behavior, but I would like to have the concept at least reviewed and accepted.

@jkschneider @sambsnyd 

When trying to reach a resolved version from a SemVer constraint, there's not really any other place to fetch the data from. Our own versions.csv is the next best spot to look at.

## Any additional context
N/A

### Checklist


sambsnyd (approved):

shanman190:
After feedback, we decided to go ahead and implement a dynamic version resolution directly against Artifactory instead.

PR #5940, opened 2025-08-21
Simplify `Environment#scanJar()`

knutwannheden:
The logic of `Environment#scanJar()` was quite convoluted with a lot of delegate resource loaders. This is now drastically simplified.


knutwannheden:
No longer required now that we have the recipe.csv mechanism.

PR #4618, opened 2024-10-27
Backport DocCommentTree reflection to ReloadableJava17ParserVisitor

timtebeek:
## What's changed?
Backport the changes from the Java 23 support in our Java 21 parser to our Java 17 parser:
- https://github.com/openrewrite/rewrite/pull/4516

## What's your motivation?
As requested via links below, and helpful for folks that only have rewrite-java-17 on their classpath still.

## Anyone you would like to review specifically?
@stephan202

## Any additional context
- https://github.com/PicnicSupermarket/error-prone-support/pull/1329#issuecomment-2350934389
- https://github.com/openrewrite/rewrite/pull/4516#issuecomment-2440037849

PR #5942, opened 2025-08-21
Init of Java 25 parser

Laurens-W:
## What's changed?

Initial setup of Java 25 parser based on the Java 21 variant

### Checklist

### DIFF
--- a/.github/workflows/receive-pr.yml
+++ b/.github/workflows/receive-pr.yml
@@ -14,11 +14,11 @@ concurrency:
 # Since this pull request receives untrusted code, we should **NOT** have any secrets in the environment.
 jobs:
   upload-patch:
-    uses: openrewrite/gh-automation/.github/workflows/receive-pr.yml@java25
+    if: false # Temporarily disable to avoid excessive suggestion spam
+    uses: openrewrite/gh-automation/.github/workflows/receive-pr.yml@main
     with:
       java_version: |
         25-ea
         21
       recipe: 'org.openrewrite.recipes.rewrite.OpenRewriteRecipeBestPracticesSubset'
-      additional_gradle_args: '-Drewrite.exclusion=**/rewrite-java-25/**'
 


.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ concurrency:
 jobs:
   build:
     uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
+    with:
+      java_version: '25'
```
timtebeek: Perhaps we can use 24 on this branch at least until there's an early access version of 25 available within GitHub Actions?
```suggestion
      java_version: '24'
```

rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -162,6 +162,17 @@ static List<Path> dependenciesFromResources(ExecutionContext ctx, String... arti
...
+                return (JavaParser.Builder<? extends JavaParser, ?>) Class
+                        .forName("org.openrewrite.java.Java25Parser")
+                        .getDeclaredMethod("builder")
+                        .invoke(null);
+            } catch (Exception e) {
+                //Fall through, look for a parser on an older version.
+            }
+        }
+
         if (version >= 21) {
```
timtebeek: I wonder here we're doing these comparisons wrong; shouldn't we use something like
```suggestion
        if (version > 21) {
            try {
                return (JavaParser.Builder<? extends JavaParser, ?>) Class
                        .forName("org.openrewrite.java.Java25Parser")
                        .getDeclaredMethod("builder")
                        .invoke(null);
            } catch (Exception e) {
                //Fall through, look for a parser on an older version.
            }
        }

        if (version > 17) {
```
And so on... 

That way folks get the next LTS whenever available as that's more likely to fold in parser fixes than a previous version.

timtebeek: Then for 26+ Shannon recently proposed adding a `rewrite-java-next` if that requires changes, as an on going effort to support the then-latest-not-LTS version going forward.

Laurens-W: Should I make adding `rewrite-java-next` part of this PR? Or should that be a separate effort?

timtebeek: No I'd keep that separate; it won't be relevant until ~March next year, and then only if there are parser changes we need to commodate, so potentially it'll take a whole year.

rewrite-java-25/src/test/java/org/openrewrite/java/Java25JavadocTest.java
timtebeek, line 1: Are these tests specific to 25? Or just a copy of the ones we had for 21? I'd only keep tests here that are very specific to 25, and for whatever reason do not fit into `rewrite-java-test`.

rewrite-java-25/src/test/resources/META-INF/rewrite/classpath/jackson-annotations-2.17.1.jar
timtebeek, line 1: Somewhat surprised with this one; is it needed are just a copy?

Laurens-W: Just a copy, will remove this along with the copied tests

Laurens-W:
@jevanlingen verified the build (aside from npmTest) works on his machine as well
@timtebeek could you give this another look over?
After that we'll need https://github.com/openrewrite/rewrite-build-gradle-plugin/pull/109 to be merged (or another solution to the ambiguous configurations) so we can verify the CI succeeds

Changes other than copying the rewrite-java-21 module primarily reside here:
- https://github.com/openrewrite/rewrite/blob/java25-parser/rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25JavadocVisitor.java#L981-L985
- https://github.com/openrewrite/rewrite/pull/5942/files#diff-2ec72a63602921457b9e2322c2e286adf38f0b62b297e97a2b76117c170e1c3c
- https://github.com/openrewrite/rewrite/pull/5942/files#diff-4a40351046da5b1e475a57481886004f1e685c10cfeb935fb862cc80fbe21454
- https://github.com/openrewrite/rewrite/pull/5942/files#diff-e3a65c8e28605fd5ab2d618f3bfce6968654a65dc58a484bc879879cf707e2a6

Laurens-W:
The Kotlin compiler doesn't play nice with ea versions of Java 25. I did find these open issues;
https://youtrack.jetbrains.com/projects/KT/issues/KT-77518/Kotlin-CLI-JavaVersion.parse-fails-on-25-beta
https://youtrack.jetbrains.com/issue/KT-73967/JDK-25-IllegalArgumentException-25-ea-with-EA-builds

Laurens-W:
I've enabled Java 21 toolchain support for rewrite-kotlin and made that available (here for now https://github.com/openrewrite/gh-automation/compare/main...java25) and that seems to do the trick

Laurens-W:
I think the rewrite-gradle-plugin can't handle Java 25 just yet, hence why it's spewing out so many diffs, there's a concurrency issue. I can reproduce it locally

timtebeek:
There's a bit of a hack still in the Java 21 parser, which we wouldn't need there anymore following the new matching of Java versions with LTS parsers. Can we remove that handling from the Java 21 parser?
https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java#L1791-L1794

And I say that as the one that added it; ideally we push folks to the next LTS parser as opposed to back porting based on what folks happen to have on the classpath.
- https://github.com/openrewrite/rewrite/pull/4618

rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25JavadocVisitor.java
```diff
@@ -0,0 +1,1333 @@
...
+                cursor += escapedCharLength;
+            } else {
+                text.append(c);
+                cursor++;
+            }
+            // The AST contained unnecessary whitespace for Javadoc, and they got rid of this with Java 25.
+            // So now have to manually account for this.
+            if (i+1 <= node.length() -1 && node.charAt(i+1) != source.charAt(cursor) && Character.isWhitespace(source.charAt(cursor))) {
+                text.append(whitespaceBeforeAsString());
+            }
```
timtebeek, line 985: These lines are new as compared to the Java 21 parser; they look fine to me; figured call these out for wider review as the rest of the file is identical between 21 and 25.

timtebeek:
Similarly to above, we can revert this ward that was added to the Java 21 parser to support Java 24:
https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java#L780-L785

We'd want to revert to the simple `type instanceof Type.UnknownType` check that we had before these two PRs:
- https://github.com/openrewrite/rewrite/pull/5309/
- https://github.com/openrewrite/rewrite/pull/5385/

.github/workflows/receive-pr.yml
```diff
@@ -14,6 +14,11 @@ concurrency:
...
 jobs:
   upload-patch:
-    uses: openrewrite/gh-automation/.github/workflows/receive-pr.yml@main
+    uses: openrewrite/gh-automation/.github/workflows/receive-pr.yml@java25
     with:
+      java_version: |
+        25-ea
+        21
       recipe: 'org.openrewrite.recipes.rewrite.OpenRewriteRecipeBestPracticesSubset'
+      additional_gradle_args: '-Drewrite.exclusion=**/rewrite-java-25/**'
```
timtebeek: Wondering if we should temporarily disable this workflow completely, such that we can merge the Java 25 parser and aim to fix the runtime issues downstream that cause the excessive suggstions to be reported.

timtebeek:
Now seeing the same suspicious code suggestions in an unrelated PR, so perhaps we don't have to trouble shoot that here:
- https://github.com/openrewrite/rewrite/pull/5983#pullrequestreview-3173236354

Laurens-W:
> Now seeing the same suspicious code suggestions in an unrelated PR, so perhaps we don't have to trouble shoot that here:
> 
> * [`RewriteTest` should load recipes through `RecipeLoader`, not `RecipeIntrospectionUtils` #5983 (review)](https://github.com/openrewrite/rewrite/pull/5983#pullrequestreview-3173236354)

I was explicitly able to reproduce by changing my system jdk to 25 explicitly and running the plugin as we do in the pipeline though

.github/workflows/receive-pr.yml
Laurens-W: That seems like the only way forward, it is gonna be a challenge though, since the rewrite-gradle-plugin doesn't seem immediately fit for Gradle 9, but I can continue looking into that one :)

timtebeek (approved):
Great to see, thanks for taking the time to remove some oddities from our older parsers, ensure going forward we pick the right next parser, and can now move on to making the other parts work for Java 25+.

PR #5946, opened 2025-08-22
Unwrap else block after return or throw statement

greg-at-moderne:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.UnwrapElseAfterReturn?organizationId=QUxML09wZW4gU291cmNlL09wZW5SZXdyaXRl


rewrite-core/src/main/java/org/openrewrite/MoveFile.java
```diff
@@ -101,11 +102,11 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
                             if (isFileOnRoot) {
                                 return ((SourceFile) tree).withSourcePath(Paths.get(separatorsToSystem(destination.toString())));
-                            } else if (isWindowsPath) {
+                            }
+                            if (isWindowsPath) {
                                 return ((SourceFile) tree).withSourcePath(Paths.get(separatorsToWindows(destination.toString())));
-                            } else {
-                                return ((SourceFile) tree).withSourcePath(Paths.get(separatorsToUnix(destination.toString())));
                             }
+                            return ((SourceFile) tree).withSourcePath(Paths.get(separatorsToUnix(destination.toString())));
```
timtebeek, line 109: Thanks for opening the PR! I had up to now kept this out of the PR checks we run here, as I wasn't sure how others would feel about the relative readability of the chained if-else here, versus the immediate returns. Any thoughts or preferences there?

https://github.com/openrewrite/rewrite/blob/8485f7420e0664635decc4000225bf65fd9c9010/rewrite.yml#L45

greg-at-moderne: Indeed, this looks suboptimal. No strong feelings here, but I suppose we could amend the recipe to cover this scenario slightly better first.

PR #5947, opened 2025-08-22
Correct parsing of TOML multiline strings

knutwannheden:


rewrite-toml/src/test/java/org/openrewrite/toml/TomlParserTest.java
```diff
@@ -23,12 +23,54 @@
 import static org.openrewrite.toml.Assertions.toml;
 
 class TomlParserTest implements RewriteTest {
+
     @Test
-    void keyValueString() {
+    void multiLineString() {
```
shanman190: optional: Did we intend to drop the original test case?

shanman190 (approved):

rewrite-toml/src/test/java/org/openrewrite/toml/TomlParserTest.java
knutwannheden: No. Good catch!

Issue #5929, opened 2025-08-19
Recipe to apply Kotlin's `@Deprecated(replaceWith=...)` suggestions


timtebeek:
## What problem are you trying to solve?
Support trivial replacements built into the Kotlin language, but then scaled up through recipes, as per:
- https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-deprecated/
- https://github.com/openrewrite/rewrite-kotlin/issues/25

## Describe the situation before applying the recipe
Any usage of these deprecated methods
```kotlin
  @Deprecated(
    "orNone is being renamed to getOrNone to be more consistent with the Kotlin Standard Library naming",
    ReplaceWith("getOrNone()")
  )
  public fun orNone(): Option<B> = getOrNone()
```
```kotlin
  @Deprecated(
    DeprMsg + "Use isLeft on Either after refactoring",
    ReplaceWith("toEither().isLeft()")
  )
  public val isInvalid: Boolean =
    fold({ true }, { false })
```
```kotlin
@Deprecated(
  DeprMsg + "Use left instead to construct the equivalent Either value",
  ReplaceWith("this.left()", "arrow.core.left")
)
public inline fun <E> E.invalid(): Validated<E, Nothing> =
  Invalid(this)
```

## Describe the situation after applying the recipe
Those usages converted to their suggested replacements.

## Have you considered any alternatives or workarounds?
Generate recipes based on such patterns identified in well known libraries; not as usable or easily updated.

## Any additional context
We already have similar support for Java's @InlineMe`; we can copy the same approach here.
- https://github.com/openrewrite/rewrite-migrate-java/pull/788#issuecomment-3200452153
- https://github.com/openrewrite/rewrite-migrate-java/blob/d309e2be9380f4a5d03cc36c4a16f998404a8999/src/main/java/org/openrewrite/java/migrate/InlineMethodCalls.java#L40

PR #5953, opened 2025-08-22
Add `@InlineMe` and `InlineMethodCalls` recipe to replace annotated methods

timtebeek:
## What's changed?
Add an annotation and associated recipe, which supports any recipe with the same structure: our own, Google's or a library specific one, much like `@Nullable` can come from a variety of libraries

## What's your motivation?
Allow folks to easily mark methods for replacement with an argument template. We have these use cases internally for our libraries any moment we expand a constructor with a new nullable argument, or when deprecating a method with a clear one to one replacement using (some of) the same arguments.

It's already used in libraries like Guava, where we can now do such replacements just from having the library on the classpath.

Would we want to support and promote this going forward?

- Instead of `@InlineMe` I've considered `@ReplaceWith`, to more clearly state that it's not just for inlining. We can optionally support both.

## Any additional context
- Based on https://errorprone.info/docs/inlineme
- As seen on https://github.com/openrewrite/rewrite-migrate-java/pull/788
- There's a similar request out for Kotlin, where it's natively part of `@Deprecated`: https://github.com/openrewrite/rewrite/issues/5929

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
@@ -1,0 +1,306 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.tree.*;
+
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.util.Collections.emptySet;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toMap;
+import static java.util.stream.Collectors.toSet;
+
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
```diff
@@ -0,0 +1,309 @@
...
+                Template template = values.template(mi);
+                if (template == null) {
+                    return mi;
+                }
+                removeAndAddImports(method, values.getImports(), values.getStaticImports());
+                J replacement = JavaTemplate.builder(template.getString())
+                        .contextSensitive()
+                        .imports(values.getImports().toArray(new String[0]))
+                        .staticImports(values.getStaticImports().toArray(new String[0]))
+                        .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
```
timtebeek, line 74: This `runtimeClasspath` is the only part I'm not quite pleased with, as what's on the runtime classpath might differ in certain environments. Still I wouldn't immediately know of a good alternative here, and we've already seen the recipe work in practice on some projects.

sambsnyd: I think for now `runtimeClasspath()` is as good as you're going to get

rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
```diff
@@ -0,0 +1,309 @@
...
+                return avoidMethodSelfReferences(nc, replacement);
+            }
+
+            private @Nullable InlineMeValues findInlineMeValues(JavaType.@Nullable Method methodType) {
+                if (methodType == null) {
+                    return null;
+                }
+                List<String> parameterNames = methodType.getParameterNames();
+                if (!parameterNames.isEmpty() && "arg0".equals(parameterNames.get(0))) {
+                    return null; // We need `-parameters` before we're able to substitute parameters in the template
```
sambsnyd, line 108: This feels like something we could always enable in `JavaParser` if it isn't by default

sambsnyd (approved):

rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
timtebeek: Yes it's here mostly as a safeguard; it should be set in each of the Java parsers we have:
https://github.com/openrewrite/rewrite/blob/380c94278772192b2c71a02f41dfeee057ec5e29/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21Parser.java#L112

PR #5954, opened 2025-08-22
Rewrite JavaScript uses npx, no zip bundling

jkschneider:


rewrite-javascript/build.gradle.kts
```diff
@@ -83,39 +102,29 @@ tasks.named("check") {
...
     inputs.files(fileTree("rewrite/src"))
         .withPathSensitivity(PathSensitivity.RELATIVE)
-    outputs.dir(file("rewrite/dist/src/"))
+    outputs.dir(file("rewrite/dist/"))
+
+    val versionTxt = file("src/main/resources/META-INF/version.txt")
+    outputs.file(versionTxt)
+    doLast {
+        versionTxt.writeText(datedSnapshotVersion)
+    }
```
shanman190, line 118: 1. Best practice would be to instead create this somewhere in the build directory where it doesn't overlap with another output, then update `processResources` or `jar` to pull the output via the task reference as an input.

If you only want to pull in the version file, then you can specifically call that out via the `include` part of `CopySpec`.

2. The version is presently not an input to this task either. So any time that it changes, but the inputs don't trigger this task, then the file either won't exist or will be incorrect. Should the version writing be it's own task possibly?

rewrite-javascript/src/main/java/org/openrewrite/javascript/rpc/JavaScriptRewriteRpcFactory.java
```diff
@@ -0,0 +1,126 @@
...
+                nodeOptions.add("--inspect-brk=" + inspectBrk);
+            }
+
+            String pkg = "--package=@openrewrite/rewrite@" + StringUtils.readFully(getClass().getResourceAsStream("/META-INF/version.txt"));
+            String[] cmd = Stream.of(
+                    npxPath.toString(), nodeOptions.toString(), pkg, "rewrite-rpc",
+                    log == null ? null : "--log-file=" + log.toAbsolutePath().normalize()
+            ).filter(Objects::nonNull).toArray(String[]::new);
+
+            System.out.println(String.join(" ", cmd));
```
knutwannheden: Looks like a leftover.
```suggestion
```

rewrite-javascript/rewrite/package.json
```diff
@@ -1,29 +1,30 @@
 {
   "name": "@openrewrite/rewrite",
-  "version": "0.0.0",
+  "version": "8.61.0-SNAPSHOT",
```
knutwannheden: Are we going to update this file in place for each release?

shanman190: It looks like the `npmVersion` task is still there, so this looks like leftover from some local testing.

rewrite-javascript/build.gradle.kts
jkschneider: I want it to work in the IDE as well though.

shanman190: The IDE is able to pull resources from the build folder as well when performing interactions. But I get what you're meaning when you may be using the Intellij Builder as well.

PR #5961, opened 2025-08-25
Making the `exclusions` and `applyTo` options for `CoalesceProperties` actually optional when invoking from declarative recipe.

steve-aom-elliott:
## What's changed?
Made the options for `CoalesceProperties` have `required = false`, as there are already constructors provided that work around an absence of parameters being passed in, and the original intent was that they would be optional.

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/CoalescePropertiesTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/CoalescePropertiesTest.java
@@ -288,9 +288,100 @@ class CoalescePropertiesTest implements RewriteTest {
     }
 
     @Test
+    void usingYamlTest() {
+        rewriteRun(
+          spec -> spec.recipeFromYaml(
+            //language=yaml
+            """
+              type: specs.openrewrite.org/v1beta/recipe
+              name: com.testing.TestRecipe
+              displayName: Test Recipe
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/CoalescePropertiesTest.java
```diff
@@ -287,6 +287,52 @@ void doNotCoalesceDocumentsHavingAnchorsAndAliases() {
         );
     }
 
+    @Test
+    void usingYamlTest() {
```
jevanlingen: I wonder if we really need these declarative tests at all. With these tests you are actually testing the Open Rewrite framework, not the options of the recipe. I think in this PR I would just add the `required = false` property and nothing else.

steve-aom-elliott:
- Related PR in `rewrite-spring`: https://github.com/openrewrite/rewrite-spring/pull/773

rewrite-yaml/src/test/java/org/openrewrite/yaml/CoalescePropertiesTest.java
steve-aom-elliott: I've removed said tests now

jevanlingen (approved):

PR #5962, opened 2025-08-25
"value" attr when adding another attribute

BoykoAlex:
Adding attribute `foo = "bar"` with `addOnly` set to `true` before the fix produces:
`@MyAnnotation("q")` -> `@MyAnnotation(foo = "bar", "q")`
Instead it should be:
`@MyAnnotation("q")` -> `@MyAnnotation(foo = "bar", value = "q")`


timtebeek:
Thanks for the fix! I know @steve-aom-elliott is working on a parallel effort in this larger PR:
- https://github.com/openrewrite/rewrite/pull/5911

But given the small scope of your change here and imminent release I think we can get this in already.

timtebeek (approved):

BoykoAlex:
Thank you very much Tim!!!

PR #5969, opened 2025-08-26
InlineMethodCalls should handle delegating constructors

timtebeek:


rewrite-java/src/main/java/org/openrewrite/java/InlineMethodCalls.java
```diff
@@ -255,17 +255,24 @@ Template template(MethodCall original) {
...
             return new Template(templateString, parameters.toArray(new Object[0]));
         }
 
-        private static String createTemplateString(MethodCall original, String replacement, List<String> originalParameterNames) {
-            String templateString = original instanceof J.MethodInvocation &&
+        private static String createTemplateString(MethodCall original, String replacement, JavaType.Method methodType) {
+            String templateString;
+            if (original instanceof J.NewClass && replacement.startsWith("this(")) {
+                // For constructor-to-constructor replacement, replace "this" with "new ClassName"
+                templateString = "new " + methodType.getDeclaringType().getClassName() + replacement.substring(4);
```
greg-at-moderne, line 267: I have a hunch this might break in case of inner classes. Although this might be a super rare case to be used with `@InlineMe`

timtebeek: Thanks yes there's likely a few edge cases like that, depending on where the annotation is added, and how the replacement template would be formatted. If folks were to use for instance `@InlineMe(replacement = "new Outer.Inner(...")` then there would be no issue, but figured we can discover those cases as we adopt this ourselves.

Issue #5921, opened 2025-08-16
Unnamed lambda parameters in java 22+


aleksandrserbin:
[JEP 456](https://openjdk.org/jeps/456) (released in jdk 22) allowed lambda parameter name to be `_`. In JDKs `com.sun.tools.javac.parser.JavacParser` it is then [replaced with an empty string](https://github.com/openjdk/jdk/blob/jdk-26%2B11/src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java#L3967-L3969):
```
if (Feature.UNNAMED_VARIABLES.allowedInSource(source) && name == names.underscore) {
     name = names.empty;
}
```
As a result, `print` idempotence check is not satisfied.

I am not sure if this is in the scope of the project, as Java21 parser clearly does not have to support Java 22 features, but I've seen some of the other PRs enhancing 21+ support.

## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.60.1
- rewrite-java-21

## What is the smallest, simplest way to reproduce the problem?

Unit test:
```java
@Test
void underscoreVariableName() {
       rewriteRun(
         spec -> spec.typeValidationOptions(TypeValidation.none()),
         java(
           """
             import java.util.Function;
             public class A {
                 Function<String, String> f = _ -> "";
             }
             """
         )
       );
}
```

## What did you expect to see?
No parser issues

## What did you see instead?
```
[When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "A.java":
 import java.util.Function;
 public class A {
-    Function<String, String> f = _ -> "";
+    Function<String, String> f =  -> "";
 }
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Yes, I am not sure about the best approach though.

timtebeek:
hi! Thanks for the detailed report; we're planning to start a Java 25 parser soon; that's the likely best place for any such accommondations; welcome to help out: some of the steps there are very straightforward:
- https://github.com/openrewrite/rewrite/issues/5717#issuecomment-3035267942

aleksandrserbin:
Thanks Tim. What do you think about the approach though? Is it okay for parser to check if variable declaration is in lambda parameter and then restore the name?

timtebeek:
[...]

PR #5971, opened 2025-08-28
Support unnamed variables in Java25Parser

timtebeek:
Figured pull this separate from adding the parser, even if we likely need to adjust the parser still.
- https://github.com/openrewrite/rewrite/pull/5942

- Fixes https://github.com/openrewrite/rewrite/issues/5921

Based on 
- https://openjdk.org/jeps/456
- https://docs.oracle.com/en/java/javase/21/language/unnamed-variables-and-patterns.html

Since these tests only activate on Java 25 we could already merge this, but it would then complicate / delay merging the parser.


rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
```diff
@@ -0,0 +1,325 @@
...
+          java(
+            """
+              interface TriConsumer<T, U, V> {
+                  void accept(T t, U u, V v);
+              }
+
+              class Test {
+                  void test() {
+                      TriConsumer<String, String, String> consumer = (_, _, third) ->
+                          System.out.println(third);
```
timtebeek, line 120: I wonder if we should add some after recipe check here to verify that the first and second `_` are not semantically equal.

timtebeek:
We'll want to make sure the tests in rewrite-java-25 run before we merge, as in the latest run to date it was still skipped:
https://ge.openrewrite.org/s/zdegkdjzujfwy/projects

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
Laurens-W: Not a bad idea

.github/workflows/publish.yml
```diff
@@ -14,6 +14,10 @@ concurrency:
 jobs:
   release:
     uses: openrewrite/gh-automation/.github/workflows/publish-gradle.yml@main
+    with:
+      java_version: |
+        25-ea
+        21
```
Laurens-W: Unrelated to this PR but still has to be done

rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
```diff
@@ -811,9 +811,19 @@ public J visitDeconstructionPattern(DeconstructionPatternTree node, Space fmt) {
                 type);
     }
 
+    @Override
+    public J visitAnyPattern(AnyPatternTree node, Space space) {
```
Laurens-W, line 850: This was available as preview feature in Java 21, do we backport it?

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
```diff
@@ -0,0 +1,337 @@
...
+                  void test() {
+                      int _;
+                      _ = 42;
+                      _ = 100;
+                  }
+              }
+              """
+          )
+        );
+    }
```
Laurens-W: I don't think this test makes sense, it doesn't look like reassigning a value is allowed

rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
timtebeek: Nope; we don't support preview features, as we'd then have to enable those when we parse proejcts, and they're not guaranteed worth the effort. As an example StringTemplates were previewed and then dropped again, whereas others have had 10 previews.

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
timtebeek: Then let's remove it!

.github/workflows/publish.yml
timtebeek: https://github.com/openrewrite/rewrite/commit/b5366b3895a5b8b5003da448af9f3829e0d14d37

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
Laurens-W:
```
Declaring an unnamed variable does not place a name in scope, so the variable cannot be written or read after it is initialized.
An initializer must be provided for an unnamed variable declared in a local variable declaration statement or in the resource specification of a try-with-resources statement.
```
from https://openjdk.org/jeps/456

Laurens-W:
> We'll want to make sure the tests in rewrite-java-25 run before we merge, as in the latest run to date it was still skipped: https://ge.openrewrite.org/s/zdegkdjzujfwy/projects

Compatibility tests for Java 25 ran and the project was included in the develocity report!

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
```diff
@@ -694,6 +694,9 @@ public J visitIdentifier(Identifier ident, PrintOutputCapture<P> p) {
         visit(ident.getAnnotations(), p);
         beforeSyntax(ident, Space.Location.IDENTIFIER_PREFIX, p);
         p.append(ident.getSimpleName());
+        if (ident.getSimpleName().isEmpty()) {
+            p.append("_");
```
timtebeek: @jkschneider we're following the pattern here of modeling LST elements the same as they are modeled in the compiler: unnamed variables have no name (empty), which we handle in the printer as seen here, as opposed to using any explicit Marker or `"_"` name. Do you agree with this mapping?

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyPrinter.java
```diff
@@ -392,6 +392,16 @@ public J visitForEachLoop(J.ForEachLoop forEachLoop, PrintOutputCapture<P> p) {
             return forEachLoop;
         }
 
+        @Override
```
jkschneider, line 395: why is this necessary? i believe it extends from `JavaPrinter` and this looks visually similar?

rewrite-java/src/main/java/org/openrewrite/java/JavaPrinter.java
jkschneider: Hard to understand the reasoning why they did that, but I think mapping it to empty and having special treatment in the printer will cause more pain for other languages that extend `J` because they would all have to extend printing of identifiers. Also, I don't know fully unless we checked what for example scala does for this.

Laurens-W: Without looking at the compiler mappings/models it seems in most languages either `_` is a convention to name ignored variables or it's a language construct that avoids creating a variable all together

In Scala `_` has a bunch of use cases, like `list.reduce(_ + _) -> expands to: list.reduce((a, b) => a + b)` or `list.zipWithIndex.map { case (_, index) => index }` which are similar to unnamed/anonymous variables, but they also use it as wildcard, for generic typing and a few more use cases

I'll convert to mapping the `_` as the identifier name, which might also make writing a recipe to start using an unnamed variable a little easier

Laurens-W: Ah, `/*~~(type information has a different variable name '')~~>*/_` that's fair but inconvenient

rewrite-java/src/main/java/org/openrewrite/java/search/FindMissingTypes.java
```diff
@@ -117,7 +117,10 @@ public J.Identifier visitIdentifier(J.Identifier identifier, ExecutionContext ct
                 identifier = SearchResult.found(identifier, "Identifier type is missing or malformed");
             }
             if (identifier.getFieldType() != null && !identifier.getSimpleName().equals(identifier.getFieldType().getName())) {
-                identifier = SearchResult.found(identifier, "type information has a different variable name '" + identifier.getFieldType().getName() + "'");
+                // Allow underscore when field type name is empty (unnamed variable pattern)
+                if (!("_".equals(identifier.getSimpleName()) && identifier.getFieldType().getName().isEmpty())) {
+                    identifier = SearchResult.found(identifier, "type information has a different variable name '" + identifier.getFieldType().getName() + "'");
+                }
```
Laurens-W: Can't say I think this is an elegant solution

rewrite-java-tck/src/main/java/org/openrewrite/java/tree/UnnamedVariableTest.java
timtebeek: I meant a slightly different assertion as now added in https://github.com/openrewrite/rewrite/pull/5971/commits/3081b5daf5a83acb1957a46f60953a769f4c4bb9 ; that's made the test fail, and I think it would be good to at least explore a fix.

Laurens-W:
```
Declaring an unnamed variable does not place a name in scope, so the variable cannot be written or read after it is initialized.
```
https://openjdk.org/jeps/456

@timtebeek With the above statement, it's ok they are semantically equal right?

rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
```diff
@@ -846,10 +846,29 @@ public J visitDeconstructionPattern(DeconstructionPatternTree node, Space fmt) {
...
+        JavaType type = typeMapping.type(node);
+        skip("_");
+        J.Identifier name = new J.Identifier(randomId(), space, Markers.EMPTY, emptyList(), "_",
+                type instanceof JavaType.Variable ? ((JavaType.Variable) type).getType() : type,
+                type instanceof JavaType.Variable ? (JavaType.Variable) type : null);
+        if (name.getFieldType() != null && name.getFieldType().getName().isEmpty()) {
+            name = name.withFieldType(name.getFieldType().withName("_"));
```
timtebeek: I can't get the tests to run locally yet to attach a debugger, but it reads a bit odd to me to create an instance with a specific type and field type, and then directly following change only the field type on the created instance.

Would it not make more sense to either change the `type` variable on line 851 directly, and thereby remove the conditional and partial update, or even move that into `typeMapping.type(node)` for unnamed variables anywhere?

rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
```diff
@@ -846,10 +846,29 @@ public J visitDeconstructionPattern(DeconstructionPatternTree node, Space fmt) {
...
+    public J visitAnyPattern(AnyPatternTree node, Space space) {
+        JavaType type = typeMapping.type(node);
+        if (type instanceof JavaType.Variable var) {
+            type = var.withName("_");
+        }
+        skip("_");
+        J.Identifier name = new J.Identifier(randomId(), space, Markers.EMPTY, emptyList(), "_",
+                type instanceof JavaType.Variable ? ((JavaType.Variable) type).getType() : type,
+                type instanceof JavaType.Variable ? (JavaType.Variable) type : null);
+        return name;
```
timtebeek:
```suggestion
        return new J.Identifier(randomId(), space, Markers.EMPTY, emptyList(), "_",
                type instanceof JavaType.Variable ? ((JavaType.Variable) type).getType() : type,
                type instanceof JavaType.Variable ? (JavaType.Variable) type : null);
```

timtebeek (commented):
Approved from my side already; from Slack it seemed your changes align with what was expected by @jkschneider as well, but let's ask of explicit approval here or in Slack before we merge, just to be sure since these choices are hard to change later.

PR #5977, opened 2025-08-29
Compare `JavaType.Method.name` before `declaringType` in `equals(Object)`

timtebeek:
As discovered on
- https://github.com/openrewrite/rewrite-static-analysis/pull/727#discussion_r2309690738

Before we had
https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L1563-L1572

Which then led to calling equals on the `declaringType` first, which for `FullyQualified.Class` looks like:
https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L576-L583

Whereas the `name` is just a String that we can more cheaply compare and reject, making it unnecessary to do name comparisons first in
https://github.com/openrewrite/rewrite-static-analysis/blob/232339c63ac81a5e47275a093beb845154b6f745/src/main/java/org/openrewrite/staticanalysis/RemoveUnusedPrivateMethods.java#L96-L98

### DIFF
--- a/rewrite.yml
+++ b/rewrite.yml
@@ -68,7 +68,7 @@ recipeList:
   - org.openrewrite.java.recipes.UseTreeRandomId
   - org.openrewrite.java.recipes.migrate.RemoveTraitsUsageRecipes
   - org.openrewrite.staticanalysis.NeedBraces
-  - org.openrewrite.staticanalysis.RemoveSystemOutPrintln
+#  - org.openrewrite.staticanalysis.RemoveSystemOutPrintln
 #  - org.openrewrite.java.RemoveAnnotation:
 #      annotationPattern: '@org.openrewrite.NlsRewrite.DisplayName'
 #  - org.openrewrite.java.RemoveAnnotation:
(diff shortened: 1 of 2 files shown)


rewrite.yml
```diff
@@ -68,7 +68,7 @@ recipeList:
   - org.openrewrite.java.recipes.UseTreeRandomId
   - org.openrewrite.java.recipes.migrate.RemoveTraitsUsageRecipes
   - org.openrewrite.staticanalysis.NeedBraces
-  - org.openrewrite.staticanalysis.RemoveSystemOutPrintln
+#  - org.openrewrite.staticanalysis.RemoveSystemOutPrintln
```
greg-at-moderne, line 71: Was this intended?

timtebeek: Yes, didn't want to keep patching main when folks as print statements to test code.

Issue #5924, opened 2025-08-17
YAML anchored sequences are not working


jimschubert:
## What version of OpenRewrite are you using?

This has never worked afaik. Can be verified on latest master 501f54e27b92d54fcd9059c666b9d297b7058824.

## How are you running OpenRewrite?

I usually run via the maven plugin.

Can you share your configuration so that we can rule out any configuration issues?

The issue can be verified via this test in MappingTest:

```
    @Test
    void mappingSequenceWithAnchor() {
        rewriteRun(
          yaml(
            """
              defaults:
                - &first
                  A: 1
                  B: 2
              mapping:
                - *first
                - <<: *first
                  A: 23
                  C: 99
              """
          )
        );
    }
```
Errors, resulting in this diff:
<img width="790" height="326" alt="Image" src="https://github.com/user-attachments/assets/a3446d56-c0d3-43e1-a7d6-2c8bbcd4218c" />

Also via minimal repro in SequenceTest:

```
    @Test
    void sequenceWithAnchors() {
        rewriteRun(
          yaml(
            """
            - &first
              A: 1
              B: 2
            - <<: *first
              A: 23
              C: 99
            """
          )
        );
    }
```

Errors, resulting in this diff:

<img width="790" height="326" alt="Image" src="https://github.com/user-attachments/assets/0e686bcc-73b5-4f71-9ab7-5f5b41b61bc5" />

Is your project public? If so, can you share a link to it?
n/a

## What is the smallest, simplest way to reproduce the problem?

```yaml
- &first
  A: 1
  B: 2
- <<: *first
  A: 23
  C: 99
```

## What did you expect to see?

Print fidelity when provided this valid YAML.

Here's an example unit test showing that SnakeYAML supports this syntax, which is admittedly a bit ambiguous in the YAML specification:

```
   @Test
    void parsesAnchorsAndAliases() {
        String yamlStr = """
                     - &first
                       A: 1
                       B: 2
                     - <<: *first
                       A: 23
                       C: 99
                     """;

        org.yaml.snakeyaml.Yaml yaml = new org.yaml.snakeyaml.Yaml(new SafeConstructor(new LoaderOptions()));
        Node root = yaml.compose(new StringReader(yamlStr));
        assertThat(root).isInstanceOf(SequenceNode.class);

        List<Node> items = ((SequenceNode) root).getValue();
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < items.size(); i++) {
[...]

PR #5978, opened 2025-08-29
fix(yaml): handle anchored map within sequence

jimschubert:
## What's changed?
rewrite-yaml's parsing of anchored maps within a sequence, as described by my issue 
- #5924

## What's your motivation?
The parsing of valid yaml containing an anchored map within a sequence would result in the `-` being dropped in the output - which is invalid yaml in some cases.

I was unsure why `fmt` was being reevaluated within the anchoring block, as the motivation was not commented and was the source of this bug. I've removed the fmt and resolved the handling of the `-` and anchor prefix.

n/a

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
@@ -248,9 +248,9 @@ class MappingTest implements RewriteTest {
         rewriteRun(
           yaml(
             """
-                    -data: &first
-                      test: 0
-              """
+            -data: &first
+              test: 0
+            """
           )
         );
     }
@@ -452,7 +452,7 @@ class MappingTest implements RewriteTest {
     void mappingSequenceWithAnchor() {
         rewriteRun(
           yaml(
-            """
+              """
               defaults:
                 - &first
                   A: 1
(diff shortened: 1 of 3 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/MappingTest.java
```diff
@@ -408,6 +448,27 @@ void mappingAnchor() {
...
+                - &first
+                  A: 1
+                  B: 2
+                - <<: *first
+                  A: 23
+                  C: 99
+              """
+          )
+        );
+    }
```
jimschubert: In case this is helpful for review… here's a screenshot of debugging this test, showing  <kbd>Sequence</kbd> -> <kbd>Sequence.Entry</kbd> -> <kbd>Mapping</kbd> with anchor and entries in the expected location. Expand the image to show the anchor prefix is the space between the `-` and `&`. I hope this is acceptable.

image

jimschubert:
I didn't check the last item in the checklist on the PR description. I've auto-formatted all new code, but some existing code in YamlParser is not formatted according to IntelliJ details. I'm assuming this project is like many others and would prefer a formatting-only change in history. Let me know if that's not the case and I'll reformat that file fully in this PR.

The latest IntelliJ update seems to cause poor formatting in the test's text blocks. I had an issue with this before at work and had to define an IntelliJ entry in my editor config there.

PR #5979, opened 2025-08-29
UpgradeTransitiveDependencyVersion should not downgrade a transitive dependency version if lower than resolved version

bmuschko:
## What's changed?

Changes the logic in the AddManagedDependency recipe to prevent version downgrade for transitive dependencies.

## What's your motivation?

A Maven build defines two dependencies with transitive dependencies on com.fasterxml.jackson.core:jackson-databind. The transitive dependency versions are different.

- jackson-dataformat-xml:2.18.0 → jackson-databind:2.18.0
- jackson-datatype-jsr310:2.15.3 → jackson-databind:2.15.3

Requesting a lower version than 2.18.0, downgrades the transitive dependency com.fasterxml.jackson.core:jackson-databind which should not be allowed.

The same behavior cannot be observed for a Gradle build.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
@@ -30,9 +30,6 @@ import org.openrewrite.xml.tree.Xml;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
-
-import static org.openrewrite.internal.StringUtils.matchesGlob;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
```diff
@@ -231,6 +242,96 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
             }
 
+            /**
+             * Find the current highest version of a dependency in the project.
+             * For transitive dependencies, finds the highest among all transitive versions.
+             * For direct dependencies, uses the first found version.
+             */
+            private String findCurrentHighestVersion(String groupId, String artifactId, 
+                                                     @Nullable Scope scope, VersionComparator versionComparator) {
+                List<ResolvedDependency> dependencies = getResolutionResult()
```
sambsnyd: In maven the conflict resolution strategy is "nearest wins", not "newest wins" as it is with Gradle. This is already baked into our model... so I think you could probably simplify this down to a single `findDependencies` invocation.

bmuschko: Thanks, I made the relevant change.

I noticed that the Maven model doesn't properly represent the use case though. When running a Maven build with the following dependencies:

```
<dependency>
    <groupId>com.fasterxml.jackson.dataformat</groupId>
    <artifactId>jackson-dataformat-xml</artifactId>
    <version>2.18.0</version>
</dependency>
<dependency>
    <groupId>com.fasterxml.jackson.datatype</groupId>
    <artifactId>jackson-datatype-jsr310</artifactId>
    <version>2.15.3</version>
</dependency>
```

The resolution result looks as follows. The transitive dependency `com.fasterxml.jackson.core:jackson-databind` resolves to 2.18.0.

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ my-app ---
[INFO] com.mycompany.app:my-app:jar:1
[INFO] +- com.fasterxml.jackson.dataformat:jackson-dataformat-xml:jar:2.18.0:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-core:jar:2.18.0:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.18.0:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.18.0:compile
[INFO] |  +- org.codehaus.woodstox:stax2-api:jar:4.2.2:compile
[INFO] |  \- com.fasterxml.woodstox:woodstox-core:jar:7.0.0:compile
[INFO] \- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.15.3:compile
```

When using the method `getResolutionResult().findDependencies`, it resolves `com.fasterxml.jackson.core:jackson-databind:2.15.3` which is not correct. I think here the following dependency resolution rule for transitive dependency should apply:

> First declaration wins for equal distance: If two dependency versions are at the same depth in the dependency tree, Maven uses the version from the dependency that was declared first in your POM file.

This is different from the usual "last declaration wins" for direct dependency declarations.

PR #5980, opened 2025-08-29
Use flyweights for classes representing GAV coordinates.

sambsnyd:
We have seen high GC overhead / memory pressure related to these objects. They are created and discarded frequently both during parsing and recipe execution.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/GroupArtifact.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/GroupArtifact.java
@@ -16,14 +16,89 @@
 
 package org.openrewrite.maven.tree;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.google.errorprone.annotations.InlineMe;
 import lombok.Value;
-import lombok.With;
+import lombok.experimental.NonFinal;
+import org.jspecify.annotations.Nullable;
 
 import java.io.Serializable;
+import java.lang.ref.WeakReference;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
 
 @Value
-@With
 public class GroupArtifact implements Serializable {
+    private static final LinkedHashMap<String, WeakReference<GroupArtifact>> CACHE = new LinkedHashMap<String, WeakReference<GroupArtifact>>() {
+        @Override
+        protected boolean removeEldestEntry(Map.Entry eldest) {
+            return size() > 10_000;
+        }
+    };
+
     String groupId;
     String artifactId;
+
+    @NonFinal
+    transient int hash;
+
+    /**
+     * Static factory method of() should be used instead. This is temporarily still public for minimally-disruptive deprecation.
+     */
+    @Deprecated
+    @InlineMe(replacement = "GroupArtifact.of(groupId, artifactId)", imports = "org.openrewrite.maven.tree.GroupArtifact")
+    public GroupArtifact(String groupId, String artifactId) {
+        this.groupId = groupId;
+        this.artifactId = artifactId;
+    }
+
+    @JsonCreator
+    public static GroupArtifact of(
+            @Nullable String groupId,
+            String artifactId) {
+        String finalGroup = groupId == null ? "" : groupId;
+        String key = finalGroup + ":" + artifactId;
+
+        synchronized (CACHE) {
+            WeakReference<GroupArtifact> ref = CACHE.get(key);
+            GroupArtifact instance = ref != null ? ref.get() : null;
+
+            if (instance == null) {
+                instance = new GroupArtifact(finalGroup, artifactId);
+                CACHE.put(key, new WeakReference<>(instance));
+            }
+
+            return instance;
+        }
+    }
+
+    public GroupArtifact withGroupId(String groupId) {
+        return groupId.equals(this.groupId) ? this : of(groupId, artifactId);
+    }
+
+    public GroupArtifact withArtifactId(String artifactId) {
+        return artifactId.equals(this.artifactId) ? this : of(groupId, artifactId);
+    }
+
+    @Override
+    public int hashCode() {
+        int h = hash;
+        if (h == 0) {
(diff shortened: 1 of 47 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedGroupArtifactVersion.java
```diff
@@ -16,16 +16,26 @@
...
 import java.io.Serializable;
+import java.lang.ref.WeakReference;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.Objects;
 
 @Value
-@With
 public class ResolvedGroupArtifactVersion implements Serializable {
+    private static final LinkedHashMap<String, WeakReference<ResolvedGroupArtifactVersion>> CACHE = new LinkedHashMap<String, WeakReference<ResolvedGroupArtifactVersion>>() {
```
knutwannheden: Since we have Caffeine anyway, I would just use that, as it will perform better. I would probably also skip the weak references as the objects aren't that heavy (especially with string deduplication by the JVM).

sambsnyd: Are we actually taking advantage of JVM string de-duplication? I was just looking through the CLI codebase and I don't see `-XX:+UseStringDeduplication` being used anywhere we start up a JVM

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedGroupArtifactVersion.java
```diff
@@ -40,23 +46,66 @@ public class ResolvedGroupArtifactVersion implements Serializable {
...
+    }
+
+    @JsonCreator
+    public static ResolvedGroupArtifactVersion of(
+            @Nullable String repository,
+            String groupId,
+            String artifactId,
+            String version,
+            @Nullable String datedSnapshotVersion) {
+        String key = (repository == null ? "" : repository) + "::" + groupId + ":" + artifactId + ":" + version + ":" + (datedSnapshotVersion == null ? "" : datedSnapshotVersion);
```
pstreef, line 67: I presume the double `::` is intentional?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/GroupArtifactVersion.java
```diff
@@ -16,16 +16,22 @@
...
 
 import java.io.Serializable;
 import java.util.Objects;
 
 @Value
-@With
 public class GroupArtifactVersion implements Serializable {
+    private static final Cache<String, GroupArtifactVersion> CACHE = Caffeine.newBuilder()
+            .maximumSize(10_000)
+            .build();
```
pstreef, line 33:
```suggestion
private static final Cache<String, GroupArtifactVersion> CACHE = CaffeineCacheMetrics.monitor(Metrics.globalRegistry, Caffeine.newBuilder()
                    .maximumSize(10_000)
                    .recordStats()
                    .build(),
            "FlyweightCache - GroupArtifactVersion");
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedGroupArtifactVersion.java
```diff
@@ -16,16 +16,22 @@
...
-import lombok.With;
 import org.jspecify.annotations.Nullable;
 
 import java.io.Serializable;
 import java.util.Objects;
 
 @Value
-@With
 public class ResolvedGroupArtifactVersion implements Serializable {
+    private static final Cache<String, ResolvedGroupArtifactVersion> CACHE = Caffeine.newBuilder()
```
pstreef, line 31:
```suggestion
private static final Cache<String, ResolvedGroupArtifactVersion > CACHE = CaffeineCacheMetrics.monitor(Metrics.globalRegistry, Caffeine.newBuilder()
                    .maximumSize(10_000)
                    .recordStats()
                    .build(),
            "FlyweightCache - ResolvedGroupArtifactVersion");
```

rewrite-maven/src/main/java/org/openrewrite/maven/tree/GroupArtifact.java
```diff
@@ -16,14 +16,50 @@
...
 import lombok.Value;
-import lombok.With;
+import org.jspecify.annotations.Nullable;
 
 import java.io.Serializable;
 
 @Value
-@With
 public class GroupArtifact implements Serializable {
+    private static final Cache<String, GroupArtifact> CACHE = Caffeine.newBuilder()
```
pstreef, line 30:
```suggestion
private static final Cache<String, GroupArtifact > CACHE = CaffeineCacheMetrics.monitor(Metrics.globalRegistry, Caffeine.newBuilder()
                    .maximumSize(10_000)
                    .recordStats()
                    .build(),
            "FlyweightCache - GroupArtifact");
```

pstreef (commented):
For some reason GitHub does not like to remove multi lines in suggestions.. but you need imports too so I guess it's not a simple apply anyway..

not sure about the cache names, but having the metrics will likely be helpful in the future.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedGroupArtifactVersion.java
shanman190: Most String usage tends to land in the JVM string pool which automatically handles deduplication. Essentially, `"value"` lands in the string pool and `new String(..)` lands on heap. The later aren't automatically deduped, but they can also be converted to interned strings (the JVM flag tries to do this behind the scenes during GC for you for the type that lands on heap).

Frameworks, such as Jackson, put a lot of care into making sure to take advantage of this fact in both directions.

jkschneider:
I'm getting increasingly skeptical of this until we've determined why something like `DependencyInsight` could allocate so heavily these types.

PR #5985, opened 2025-09-01
Init of Java 25 parser

Laurens-W:
## What's changed?

Initial setup of Java 25 parser based on the Java 21 variant

### Checklist


rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21ParserVisitor.java
timtebeek, line 1: The reflection here was a source-compatible change, but not interchangeable between the Java 21 and 23+ compilers.

rewrite-javascript/rewrite/package.json
timtebeek, line 1: I'm having a bit of trouble judging this particular change; any details as to what made this necessary, and the impact of this change?

timtebeek (approved):
Approved aside from the open questions on the JS build changes.

rewrite-javascript/rewrite/package.json
Laurens-W: It's basically a circular reference which (on Windows) was causing the npm install to fail, the local and CI builds succeed after this change

Issue #5988, opened 2025-09-02
TOML parser fails with trailing commas in arrays


greg-at-moderne:
## What version of OpenRewrite are you using?

Current main = 03fdec3980cce56b892843aa9ef71a25e3304fd4

## What is the smallest, simplest way to reproduce the problem?

The following test in `TomlParserTest`:
```java
    @Test
    void trailingComma() {
        rewriteRun(
          toml(
            """
tsfresh = [
    "featuretools-tsfresh-primitives >= 1.0.0",
]
            """
          )
        );
    }
```
fails with:
```
Failed to parse sources or run recipe
java.lang.AssertionError: Failed to parse sources or run recipe
	at org.openrewrite.test.RewriteTest.lambda$defaultExecutionContext$14(RewriteTest.java:646)
	at org.openrewrite.test.RewriteTest$$Lambda/0x000000070117dc78.accept(Unknown Source)
	at org.openrewrite.toml.TomlParser.lambda$parseInputs$0(TomlParser.java:59)
	at org.openrewrite.toml.TomlParser$$Lambda/0x00000007011c0ce0.apply(Unknown Source)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1939)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:283)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:130)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:125)
	at org.openrewrite.toml.TomlParserTest.t2(TomlParserTest.java:417)
[...]

PR #5989, opened 2025-09-02
Allow trailing comma in TOML

timtebeek:
- Fixes https://github.com/openrewrite/rewrite/issues/5988


rewrite-toml/src/main/antlr/TomlParser.g4
```diff
@@ -114,7 +114,7 @@ commentOrNl
 
 array
     : L_BRACKET commentOrNl* R_BRACKET
-    | L_BRACKET commentOrNl* value (COMMA commentOrNl* value COMMA?)* commentOrNl* R_BRACKET
+    | L_BRACKET commentOrNl* value (COMMA commentOrNl* value)* COMMA? commentOrNl* R_BRACKET
```
greg-at-moderne, line 117: Hmm. Just by looking at this grammar production, I fail to understand why wouldn't the comma be allowed before?
Something about the greediness of the `*` operator?

rewrite-toml/src/test/java/org/openrewrite/toml/TomlParserTest.java
```diff
@@ -401,6 +401,47 @@ void trailingEmptyComment() {
         );
     }
 
+    @Test
+    void arrayWithTrailingComma() {
+        // Test for issue #5988 - TOML 1.0 spec allows trailing commas in arrays
```
greg-at-moderne: This should go to `@Issue` annotation if we really need it.

rewrite-toml/src/test/java/org/openrewrite/toml/TomlParserTest.java
```diff
@@ -401,6 +401,47 @@ void trailingEmptyComment() {
...
+              # Inline array with trailing comma  
+              inline = ["value1", "value2",]
+              """
+          )
+        );
+    }
+
+    @Test
+    void inlineTableWithTrailingComma() {
+        // Test that inline tables also support trailing commas
```
greg-at-moderne: Useless comment.

rewrite-toml/src/main/antlr/TomlParser.g4
timtebeek: Intuitively to me it's clear when I read it like this:
1. an optional number of values _additional_ values each starting with a comma
2. which may end in a final trailing comma, even if there was only one value to begin with

Not sure if you have that same reading, but at least the tests indicate it wasn't working before, and does now.

greg-at-moderne: > the tests indicate it wasn't working before, and does now.

Yeah. Hard to question that.

It's just I must be missing something in how these grammar rules are interpreted.

shanman190: The case here is that lists and inline-tables of size 1 with a trailing comma failed. Lists and inline-tables with size 2 or more, should have worked just fine though.

shanman190 (approved):

rewrite-toml/src/main/antlr/TomlParser.g4
greg-at-moderne: And now it makes sense! Thanks.

PR #5991, opened 2025-09-02
Introduce support for pattern matching within the `Semver` type classes

shanman190:
## What's changed?
- `CaretRange` incorrect upper bound
- Add support for comparison of a `CaretRange` boundary when within an existing `CaretRange`
- Add support for comparison of a `HyphenRange` boundary when within an existing `HyphenRange`
- Add support for comparison of a `SetRange` boundary when within an existing `SetRange`
- Add support for comparison of a `TildeRange` boundary when within an existing `TildeRange`
- Add support for comparison of an `XRange` boundary when within an existing `XRange`

## What's your motivation?
I'm reanalyzing the `SemVer` semantics because I have a use case within `UpdateGradleWrapper` in which I want to take the user supplied pattern and use that as a comparison against the current Gradle version discovered from the `BuildTool` marker. By enabling patterns to be compared within the same class, this will allow me to effectively define a precondition such that running `UpdateGradleWrapper(version=7.x)` will be skipped when the current version is 8.2, as an example.

N/A

N/A

N/A

## Any additional context
None

### Checklist

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/semver/CaretRangeTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/semver/CaretRangeTest.java
@@ -95,7 +95,7 @@ class CaretRangeTest {
     }
 
     /**
-     * ^0.0.x := >=0.0.0 <1.0.0
+     * ^0.0.x := >=0.0.0 <0.1.0
      */
     @Test
     void desugarPatchWildcard() {
@@ -104,8 +104,7 @@ class CaretRangeTest {
         assertThat(caretRange).isNotNull();
         assertThat(caretRange.isValid("1.0", "0.0.0")).isTrue();
         assertThat(caretRange.isValid("1.0", "0.0.1")).isTrue();
-        assertThat(caretRange.isValid("1.0", "0.1.0")).isTrue();
-        assertThat(caretRange.isValid("1.0", "1.0.0")).isFalse();
+        assertThat(caretRange.isValid("1.0", "0.1.0")).isFalse();
     }
 
     @Test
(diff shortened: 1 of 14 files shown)


rewrite-core/src/test/java/org/openrewrite/semver/CaretRangeTest.java
```diff
@@ -106,7 +95,7 @@ void desugarMinorWildcard() {
     }
 
     /**
-     * ^0.0.x := >=0.0.0 <0.1.0
+     * ^0.0.x := >=0.0.0 <1.0.0
```
shanman190:
```suggestion
     * ^0.0.x := >=0.0.0 <0.1.0
```

rewrite-core/src/test/java/org/openrewrite/semver/CaretRangeTest.java
```diff
@@ -115,6 +104,24 @@ void desugarPatchWildcard() {
         assertThat(caretRange).isNotNull();
         assertThat(caretRange.isValid("1.0", "0.0.0")).isTrue();
         assertThat(caretRange.isValid("1.0", "0.0.1")).isTrue();
-        assertThat(caretRange.isValid("1.0", "0.1.0")).isFalse();
+        assertThat(caretRange.isValid("1.0", "0.1.0")).isTrue();
+        assertThat(caretRange.isValid("1.0", "1.0.0")).isFalse();
```
shanman190:
```suggestion
        assertThat(caretRange.isValid("1.0", "0.1.0")).isFalse();
```

PR #5997, opened 2025-09-03
Implement module imports in Java 25 parser

Laurens-W:
## What's changed?

Add support for module imports in the Java 25 parser

## What's your motivation?

Making the parser idempotent

### Checklist


Laurens-W:
```
We extend the grammar of import declarations ([JLS §7.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-7.html#jls-7.5)) to include import module clauses:

ImportDeclaration:
  SingleTypeImportDeclaration
  TypeImportOnDemandDeclaration
  SingleStaticImportDeclaration
  StaticImportOnDemandDeclaration
  ModuleImportDeclaration

ModuleImportDeclaration:
  import module ModuleName;
  ```
  https://openjdk.org/jeps/511
  
  This concerns me a little, as that means `ImportTree` (input of `visitImport` in our [parser](https://github.com/openrewrite/rewrite/pull/5997/files#diff-de62bb89edd174c2045453a8a97eb4faad09b9af137e911e06a9b9bfdde13f98R780-R792)) can now be two things; `JCImport` or `JCModuleImport`.
If we introduce a new LST element ([`J.ModuleImport`](https://github.com/openrewrite/rewrite/pull/5997/files#diff-580e185ad228d76309d57b80586c4b258446de177a08e20ba79b7d34fbdaa9f5R3098-R3239)) to mimic this, that means we need to add commonality ([`Import`](https://github.com/openrewrite/rewrite/pull/5997/files#diff-11aee9de4831a92201f2deac5d43c28da1e911220c3fce9afe9b6a444fec8d02) interface, AST defines `boolean isStatic();` `boolean isModule();` and `Tree getQualifiedIdentifier();` the latter i've shortened to `Qualid` like we do in other places) between that and `J.Import` and do a massive refactoring of any places using `J.Import`, as a `CompilationUnit` will then return the interface and not the implementation.

On the other hand we can choose to add it to `J.Import` (`module` like [`statik`](https://github.com/openrewrite/rewrite/pull/5997/files#diff-580e185ad228d76309d57b80586c4b258446de177a08e20ba79b7d34fbdaa9f5L2886) which is modeled in the AST somewhat like that, handle the prefix of the keyword etc) but this leaves surface for issues later on

File: rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ImportTest.java
Comment (Laurens-W) [inline rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ImportTest.java:1]: Note these are entirely Claude, I'll go over how valid these all are in a bit

File: rewrite-java/src/main/java/org/openrewrite/java/tree/J.java [hunk 1]
```diff
@@ -2866,6 +2866,18 @@ public If withThenPart(JRightPadded<Statement> thenPart) {
     @RequiredArgsConstructor
     @AllArgsConstructor(access = AccessLevel.PRIVATE)
     final class Import implements Statement, Comparable<Import> {
+
+        @JsonCreator
+        public Import(UUID id, Space prefix, Markers markers, JLeftPadded<Boolean> statik, FieldAccess qualid, @Nullable JLeftPadded<Identifier> alias) {
```
knutwannheden: Typically we don't add constructors for backwards compatibility. So I think we should probably remove this constructor and instead require updating client code everywhere. Alternatively, we could consider adding the `@ToBeRemoved` annotation, so we don't forget about this (we still need a recipe that we can run to report these!). What are your thoughts, @timtebeek ?

I take it the reason for adding this constructor and then also annotating it as `@JsonCreator` was to allow Jackson to deserialize old LSTs. That's an interesting approach, which we would have to test and make sure that Jackson then indeed also does set the `module` property using reflection, when it is present in the LST. But if that is the purpose of this constructor, then we should declare it as package private and also annotate it as `@ToBeRemoved`.

Laurens-W: There's always the `@RequiredArgsConstructor` which should cover cases where the `module` property is present, right?
It is indeed supposed to cover backwards compatibility when the `module` property is not present in the LST; We could probably also cover that by marking it `@Nullable` though, which is probably the better approach

timtebeek:
As a quick thought: once we have merged the parser support, we'll likely also want to adapt the style detection to figure out at which point to switch from explicit imports to module imports, much like we do for wildcard imports. That tally can then factor into `OrderImports` to adopt module imports there.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3071,15 +3082,23 @@ public JLeftPadded<Boolean> getStatic() {
             }
 
             public Import withStatic(JLeftPadded<Boolean> statik) {
-                return t.statik == statik ? t : new Import(t.id, t.prefix, t.markers, statik, t.qualid, t.alias);
+                return t.statik == statik ? t : new Import(t.id, t.prefix, t.markers, statik, t.module, t.qualid, t.alias);
+            }
+
+            public JLeftPadded<Boolean> getModule() {
```
knutwannheden: I think at least the getter here also needs to be `@Nullable`.

sambsnyd (approved):

sambsnyd:
Since this is a model change be sure to queue up a saas deployment after you merge it

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
Laurens-W: ~The getter does a null check and defaults to false~ ah nvm, the getter on the padding!

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3070,9 +3070,7 @@ public Import withStatic(JLeftPadded<Boolean> statik) {
                 return t.statik == statik ? t : new Import(t.id, t.prefix, t.markers, statik, t.module, t.qualid, t.alias);
             }
 
-            @Nullable
-            public JLeftPadded<Boolean> getModule() {
-                return t.module;
+            public @Nullable JLeftPadded<Boolean> getModule() {
             }
```
timtebeek, line 3075:
```suggestion
            public @Nullable JLeftPadded<Boolean> getModule() {
                return t.module;
            }
```

jkschneider:
I don't approve of the approach. The JLS and compiler itself differentiate between module imports and non-imports, and I think it will add complexity to every import-related features if J.Import represents both. Closing and will redo at a later time.

PR #5998, opened 2025-09-03
Restore pull request receive workflow to apply best practices

Laurens-W:
### Checklist

### DIFF
--- a/.github/workflows/receive-pr.yml
+++ b/.github/workflows/receive-pr.yml
@@ -14,7 +14,7 @@ concurrency:
 # Since this pull request receives untrusted code, we should **NOT** have any secrets in the environment.
 jobs:
   upload-patch:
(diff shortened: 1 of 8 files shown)


Laurens-W:
@timtebeek it doesn't seem as bad as before, since we decided not to go for Gradle 9.
Let me check locally tomorrow if I can remediate some of those findings so we can have it enabled again

.github/workflows/receive-pr.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 # Since this pull request receives untrusted code, we should **NOT** have any secrets in the environment.
 jobs:
   upload-patch:
-    if: false # Temporarily disable to avoid excessive suggestion spam
+#    if: false # Temporarily disable to avoid excessive suggestion spam
```
timtebeek:
```suggestion
```

PR #6004, opened 2025-09-05
Add tests for flexible constructor bodies

Laurens-W:
## What's changed?

Added testcases to cover flexible constructor bodies

## Any additional context

https://openjdk.org/jeps/513

### Checklist

### DIFF
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.tree;
 
+import org.junit.jupiter.api.Nested;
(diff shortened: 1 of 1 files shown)


rewrite-java-tck/src/main/java/org/openrewrite/java/tree/ConstructorTest.java
```diff
@@ -173,4 +174,127 @@ public A(String a) {
           )
         );
     }
+
+    @Issue("https://openjdk.org/jeps/513")
+    @MinimumJava25
```
timtebeek, line 180: Could it make sense to add these five new tests to a nested test class that has these two annotations? That way we're not repeating ourselves and can run those selectively.

timtebeek (approved):
Otherwise approved already, thanks! Nice to know nothing needs to change here.

PR #6009, opened 2025-09-08
Groovy - `getDelimiter` handling ~/ regex in enums

### LINKED ISSUE #6003: GroovyParser goes into an infinite loop with regular expression pattern

Current main = 4999c71bc42d5b56948333e9c4ed51c3cddcd2d5

This test:
```java
    @Test
    void regexpPattern() throws IOException {
        rewriteRun(
                groovy(
                    """
                    enum E {
                        OPTION1(~/alef/)
                    }
                    """
                )
        );
    }
```
when added in `GroovyParserTest` never stops (or at least until my patience runs out - i.e. for several minutes)

## OSS repro

- https://github.com/joschi/licenser/blob/c9a9017f6a186955437416206a24de8527b0ff55/src/main/groovy/org/cadixdev/gradle/licenser/header/HeaderStyle.groovy#L31

greg-at-moderne:
## What's changed?

Fix in the Groovy parser logic to handle
```
            enum E {
                OPTION1(~/alef/)
            }
```

## What's your motivation?

- Fixes #6003.

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2911,7 +2911,7 @@ public class GroovyParserVisitor {
 
     private boolean validateIsDelimiter(@Nullable ASTNode node, int c) {
         if (node == null) {
-            return false;
+            return true;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks!

PR #6018, opened 2025-09-09
Groovy - special handling of double-quote values in enum arguments

### LINKED ISSUE #6015: Groovy parsing infinite loop with URLs used in Enum arguments

Current main = 9f1f7c3301288cdcb7c297d4762077fdfde1794c

```java
    @Test
    void uris() {
        rewriteRun(
          groovy(
            """
            enum E {
              LOCAL("http://localhost:8080/api/v1/clusters")

              E(String uri) {
              }
            }
            """
          )
        );
    }
```
in `org.openrewrite.groovy.tree.EnumTest`

never stops. I.e. goes into an infinite loop.

## Context

Similar to:
- #6009 
- #6014 

I suspect this is because `sourceBefore(")` is used and on observing `//` it considers the rest of the line as line-comment, which is not appropriate.

## OSS repro

- https://github.com/hortonworks/ambari-rest-client/blob/d6dd0f25ac1e528a72346a7f5479dc509986de55/src/test/groovy/com/sequenceiq/ambari/client/TestResources.groovy#L21

greg-at-moderne:
## What's changed?

In Groovy handling double-quoted arguments (i.e. Strings) in enum arguments in a different way than the other arguments. Here we shouldn't use the `sourceBefore()` hack which is designed to handle whitespace. Manual `cursor` advancement instead.

## What's your motivation?

- fixes #6015

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -470,6 +470,9 @@ public class GroovyParserVisitor {
                         if ("\"".equals(delimiter.close)) {
                             // This is to prevent sourceBefore interpreting // in strings as comments
                             cursor = source.indexOf("\"", cursor) + 1;
+                            while (source.charAt(cursor - 2) == '\\') {
+                                cursor = source.indexOf("\"", cursor) + 1;
+                            }
                         } else {
                             sourceBefore(delimiter.close);
                         }
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -467,7 +467,7 @@ public class GroovyParserVisitor {
                     Delimiter delimiter = getDelimiter(null, cursor);
                     if (delimiter != null) {
                         cursor += delimiter.open.length();
-                        if (delimiter.close == "\"") {
+                        if ("\"".equals(delimiter.close)) {
                             // This is to prevent sourceBefore interpreting // in strings as comments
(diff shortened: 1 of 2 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -467,7 +467,12 @@ private J.EnumValue visitEnumField(FieldNode field) {
                     Delimiter delimiter = getDelimiter(null, cursor);
                     if (delimiter != null) {
                         cursor += delimiter.open.length();
-                        sourceBefore(delimiter.close);
+                        if (delimiter.close == "\"") {
```
timtebeek: Did you mean the following? Or was this intentional? Rarely a problem in practice I think, but it looked out of place.
```suggestion
                        if ("\"".equals(delimiter.close)) {
```

rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/EnumTest.java
```diff
@@ -277,4 +277,22 @@ private TagNameSerializer(String type, Function<String, String> fun) {
...
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/6015")
+    @Test
+    void uris() {
+        rewriteRun(
+          groovy(
+            """
+            enum E {
+              LOCAL("http://localhost:8080/api/v1/clusters")
```
timtebeek: Looks like we might have another edge case with
```
              ESCAPED_QUOTE("http://localhost:8080/\"api/v1/clusters")
```

greg-at-moderne: Good point.
aff2617e7

timtebeek (approved):
Thanks for the fixes here!

PR #6020, opened 2025-09-09
Yaml newline after end of document

timtebeek:
## What's changed?
Add a trailing content marker for content after document end `...` (https://yaml.org/spec/1.2.2/#22-structures).

## What's your motivation?
This wasn't captured, and thus led to failure to faithfully reproduce the file, which lead to parse failures and missed changes.

I'm not a huge fan of the new marker added. Open to alternatives.

We could model the whitespace as prefix on the next element, or failing that, as suffix on the `Yaml.Documents`. Neither seemed too appealing for what is a rare edge case that one is unlikely to target with recipes.

https://github.com/openrewrite/rewrite/blob/9d7fe7709a41e66816033b29786deab0173b3091/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java#L179-L206

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -27,6 +27,7 @@ import org.openrewrite.marker.Markers;
 import org.openrewrite.tree.ParseError;
 import org.openrewrite.tree.ParsingEventListener;
 import org.openrewrite.tree.ParsingExecutionContextView;
+import org.openrewrite.yaml.marker.TrailingContent;
 import org.openrewrite.yaml.tree.Yaml;
 import org.openrewrite.yaml.tree.YamlKey;
 import org.yaml.snakeyaml.LoaderOptions;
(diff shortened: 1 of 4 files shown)


Laurens-W:
Ah I didn't check if anyone had picked this up just yet, had Claude chew on it a bit after adding an explicit postfix to the Yaml model and it came up with https://github.com/openrewrite/rewrite/compare/main...yaml-triple-dot-file-end
Not sure if we can take any inspiration or merge this solution

timtebeek:
Ah yes that's pretty much the alternative I had outlined above, except you've added it to the `End` instead of `Documents`. My thought was to add it to `Documents`, as we have a tendency to assign to the next outermost element where possible, whereas for two consecutive documents the prefix would be on the next document, with only the whitespace after the last `...` stored in a suffix.

Laurens-W:
I looked into it a little more, [the streamEnd event indicates the end of the file](https://github.com/snakeyaml/snakeyaml/blob/master/src/main/java/org/yaml/snakeyaml/parser/ParserImpl.java), therefore the logic in that case seems lacking, it only really handles an empty yaml file.
If we add an else case we can compute and store the suffix and use it when the `Yaml.Documents` is created or we can create the `Yaml.Documents` inside of that case

I feel like that might be a more natural way of handling things, I've updated my branch to reflect storing the suffix and using that when the `Yaml.Documents` is created

timtebeek:
Turns out there (also?) was a missing call to visit the document end, added in
- https://github.com/openrewrite/rewrite/pull/6030

Not yet clear what the impact here would be but I suggest you open a PR for your approach, since we weren't capturing that suffix yet, and I think that makes most sense. 

I'll close this PR.

PR #6023, opened 2025-09-10
Fixed a bug where root entries with newlines in front would have the entire entry be indented after unfolding.

Jenson3210:
Fixed the following bug.

When unfolding yaml: 
```yaml
logging.level.com.example: DEBUG
              
management.test:
  a.b:
    value: c
```

expected
```yaml
logging:
  level:
    com:
      example: DEBUG

management:
  test:
    a:
      b:
        value: c
```

actual
```yaml
logging:
  level:
    com:
      example: DEBUG

  management:
    test:
      a:
        b:
          value: c
```


Jenson3210:
The maybeAutoformat returns the entry with 2 spaces. This was the only way I could bypass that keeping the newlines and other prefix the same without overly complexifying the prefix calculation. As we are already shifting left in all cases, I figured this is the cleanest bypass.

timtebeek:
Thanks for that context, the detailed look and minimal fix then.

PR #6024, opened 2025-09-10
Add cycle detection to prevent StackOverflow in recipe loading

kmccarp:
## What's changed?
Prevent cycles when loading DeclarativeRecipes. Error out when a cycle is detected in order to prevent unexpected behavior, such as missing a certain recipe within a recipe jar.

## What's your motivation?

In the Moderne CLI, Moderne SaaS, and OpenRewrite build plugins, if a recipe jar load is attempted and there is an invalid DeclarativeRecipe (either self-referential, or cyclically referred to by one of the recipes in its list), this can currently cause a StackOverflowError.

N/A

@jkschneider 

We are going to guard against said StackOverflowError in the SaaS and CLI as well.

## Any additional context
N/A

### Checklist


rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
```diff
@@ -135,13 +137,30 @@ private void initialize(List<Recipe> uninitialized, List<Recipe> initialized, Fu
...
 
+    private void initializeDeclarativeRecipe(DeclarativeRecipe declarativeRecipe, String recipeIdentifier,
+                                              Function<String, @Nullable Recipe> availableRecipes, Set<String> initializingRecipes) {
+        String recipeName = declarativeRecipe.getName();
+        if (initializingRecipes.contains(recipeName)) {
+            // Cycle detected - throw exception to fail fast
+            String cycle = String.join(" -> ", initializingRecipes) + " -> " + recipeName;
+            throw new RecipeIntrospectionException(
+                    "Recipe '" + recipeIdentifier + "' creates a cycle: " + cycle +
+                    " (in " + source + ")");
```
Laurens-W, line 155: Can we make this assumption when recipes are parameterized?

kmccarp: `DeclarativeRecipe`s can't be parameterized, and non-`DeclarativeRecipes` won't show up in `initializingRecipes`. Do you have a specific example in mind?

Laurens-W: I was thinking of a mix of declarative and imperative recipes where somehow there's a cyclic dependency in declarative recipes containing imperative ones, but that doesn't seem to apply then!

PR #6028, opened 2025-09-11
Validate the syntax of `MethodMatcher` and `AnnotationMatcher` signatures at instantiation

pdelagrave:
## What's changed?
Validate the syntax of `MethodMatcher` and `AnnotationMatcher` signatures at instantiation

Fail fast instead of silently creating a matcher with an invalid syntax that would do matches inconsistent with the provided pattern.

## What's your motivation?
When syncing the recipes using the Moderne CLI with the `mod config recipes moderne sync` command, its logs would contain context-less lines suchs as:
```
line 1:94 missing ')' at '<EOF>'
line 1:88 no viable alternative at input ', String, String'
line 1:88 no viable alternative at input ', String'
line 1:88 missing ')' at '<EOF>'
```
They aren't helpful per se and require a good amount of digging to figure out what's wrong where and why.
They come from the `AnnotationMatcher` and `MethodMatcher` signature expression lexer and parser.
For example:
```java
private MethodMatcher matcher1 = new MethodMatcher("org.springframework.jdbc.core.JdbcTemplate queryForLong(..)");
private AnnotationMatcher matcher2 = new AnnotationMatcher("@org.springframework.beans.factory.annotation.Value");
```
If the pattern used in the constructor argument has a syntax error, there will be no exception, just a message sent to stdout (and depending on how stdout is sent/redirected, it might be invisible). Depending on the nature of the syntax error, it will either:
* Be lenient and filter in result that shouldn't. For example with this AnnotationMatcher before that [patch](https://github.com/openrewrite/rewrite/pull/5939) was applied:
   `new Annotated.Matcher("@org.springframework.data.jpa.repository.Query(nativeQuery=true)");` ([source](https://github.com/openrewrite/rewrite-spring/blob/f9500bb1f85be1e82815bd7fa4aa8216dbf1606c/src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java#L37-L37))
  This would print an error on stdout about the syntax but would still let in `Query` annotations no matter its `nativeQuery` value. This required this [extra check](https://github.com/openrewrite/rewrite-spring/blob/e92b3e8cf648cc63f16e044c267128cec067636d/src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java#L67-L69) in the visitor because the matcher couldn't be trusted. It has since [been fixed](https://github.com/openrewrite/rewrite-spring/commit/754ad593396b05cda072826133989808f1b9f89a) but such situations will continue to exist.
* Fail to be parsed enough so that nothing will be filtered in.

Both scenarios will cause bugs and waste developer time by lengthening the feedback loop leading to discover that a matcher signature expression isn't correct. 

### Checklist


pdelagrave:
@knutwannheden from an earlier discussion:
> Knut: Failing hard could be an issue, because the input string could sometimes maybe be a user input. But in this case the recipe should maybe validate it. I suggest we compare with MethodMatcher to see how it was addressed there.

`MethodMatcher` has a `validate()` that recipes are calling from their own overridden `Recipe.validate()` method when they need to validate a pattern coming from a user input. This makes for a cleaner UX as the error is sent back in a more controlled way. This PR also adds the specific lexing/parsing syntax error message in the returned failed validation message.

`AnnotationMatcher` has no such mechanism and the recipes using it are currently assuming it's valid, so failing fast in this case shouldn't be worse off because a failure to instantiate the matcher already fails hard (although syntax errors weren't causing it so far). The changes from this PR will just make it easier to spot issues with the syntax, just not in a nicely formatted way like `MethodMatcher` can provide.

A generic solution that'd save boilerplate validation code for both Method and Annotation matcher would be to re-use the existing  `@Option.description` value already used on method matcher pattern fields and have `Recipe.validate()` process them depending on their value. For example:

([source](https://github.com/openrewrite/rewrite/blob/c856f931e5442100e1d2b559b2e5c3bfb8948dc4/rewrite-java/src/main/java/org/openrewrite/java/AddCommentToMethodInvocations.java))
```java
public class AddCommentToMethodInvocations extends Recipe {
    @Option(displayName = "Method pattern",
            description = "A pattern to match methods to add the comment to. " + MethodMatcher.METHOD_PATTERN_INVOCATIONS_DESCRIPTION,
            example = "java.util.List add*(..)")
    String methodPattern;

    @Override
        return super.validate().and(MethodMatcher.validate(methodPattern));
    }
```

`MethodMatcher.METHOD_PATTERN_INVOCATIONS_DESCRIPTION` (and its deprecated version `MethodMatcher.METHOD_PATTERN_DESCRIPTION`) is already used in most if not all recipes requiring a method matcher pattern as input.

The `validate()` method in that recipe would be removed and the `Recipe` abstract class would scan for a matching `@Option.description` and validate the pattern automatically.
The same would work for `AnnotationMatcher` but there's no such standard description being used yet.

It might be better to create a new different `@Option` field for that purpose, it's not necessarily very clean nor intuitive to piggyback off the description field to know how to automatically validate a field. Or a new annotation itself if necessary.

This could be refactored with a best practice recipe to help us and existing custom recipe users to migrate their recipes.

knutwannheden:
I don't know if it is worth it to have a generic solution. Personally I would like to be able to declare fields of type `MethodMatcher` and annotate them directly as `@Option`, but that would be more work. But one possibility would be to add a `validator` member to the `Option` annotation type, which would be of type `Class<? extends Validator>` and then we could have `AnnotationMatcher` implement that interface.

PR #6030, opened 2025-09-12
Fixed bug in Yaml parser

Jenson3210:
Fixed a bug where possible helm template parameters in yaml document end would not be replaced back from uuid with their actual content.

Screenshot 2025-09-12 at 15 38 56
Screenshot 2025-09-12 at 15 39 14
Screenshot 2025-09-12 at 15 39 56

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -481,6 +481,11 @@ public class YamlParser implements org.openrewrite.Parser {
                 }
 
                 @Override
+                public Yaml.Documents visitDocuments(Yaml.Documents documents, Integer integer) {
+                    return unwrapPrefixedMappings(super.visitDocuments(documents, integer));
+                }
+
+                @Override
                 public Yaml.Mapping visitMapping(Yaml.Mapping mapping, Integer p) {
                     Yaml.Mapping m = super.visitMapping(mapping, p);
                     String opening = m.getOpeningBracePrefix();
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -474,7 +474,11 @@ private String restoreUuidPlaceholders(String text) {
                 @Override
                 public @Nullable Yaml visit(@Nullable Tree tree, Integer integer) {
                     Yaml visit = super.visit(tree, integer);
-                    if (visit == null || StringUtils.isBlank(visit.getPrefix())) {
+                    if (visit instanceof Yaml.Document && !StringUtils.isBlank(((Yaml.Document) visit).getEnd().getPrefix())) {
+                        Yaml.Document d = (Yaml.Document) visit;
+                        visit = d.withEnd(visitDocumentEnd(d.getEnd(), integer));
+                    }
+                    if (visit == null || (StringUtils.isBlank(visit.getPrefix()))) {
```
timtebeek:
```suggestion
                    if (visit == null || StringUtils.isBlank(visit.getPrefix())) {
```

Jenson3210:
Still one issue with the bigger things (also related to helm templates, but not this one Document.end. 
i will add it here

rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -474,7 +474,11 @@ private String restoreUuidPlaceholders(String text) {
                 @Override
                 public @Nullable Yaml visit(@Nullable Tree tree, Integer integer) {
                     Yaml visit = super.visit(tree, integer);
-                    if (visit == null || StringUtils.isBlank(visit.getPrefix())) {
+                    if (visit instanceof Yaml.Document && !StringUtils.isBlank(((Yaml.Document) visit).getEnd().getPrefix())) {
+                        Yaml.Document d = (Yaml.Document) visit;
+                        visit = d.withEnd(visitDocumentEnd(d.getEnd(), integer));
```
timtebeek: I'm slightly puzzled by the need for this separate call; should that not already be handled by the `super.visit` above? Or might the call to `visitDocumentEnd` be missing from the super visitor?

rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -474,7 +474,11 @@ private String restoreUuidPlaceholders(String text) {
...
+                    }
```
timtebeek: It looks like `visitDocument` in `YamlVisitor` is indeed missing a `withEnd` call; I propose we add that here and remove this conditional
```suggestion
```

based on
https://github.com/openrewrite/rewrite/blob/b39966691388776d428c2db76ddc3f737ce404eb/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlVisitor.java#L73-L80

Jenson3210: Oh... nice. I never looked into these kind of issues before so did not make the connection. I just thought the visiting here was not done due to that earlier return.  Thanks!

Jenson3210:
@timtebeek I know you approved already but not merging as I did change more than only that comment so would appreciate another round of review from you ( I know you are out, so next week 😉 ) or @sambsnyd

timtebeek:
Thanks both, looks good!

PR #6032, opened 2025-09-12
Groovy: Allow parsing named arguments in enum constants

knutwannheden:


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -477,6 +483,27 @@ private J.EnumValue visitEnumField(FieldNode field) {
...
+                            if (delimiter != null) {
+                                cursor += delimiter.open.length();
+                                if ("\"".equals(delimiter.close)) {
+                                    cursor = source.indexOf("\"", cursor) + 1;
+                                    while (source.charAt(cursor - 2) == '\\') {
+                                        cursor = source.indexOf("\"", cursor) + 1;
+                                    }
+                                } else {
+                                    sourceBefore(delimiter.close);
+                                }
```
greg-at-moderne, line 502: I am wondering if there's a nice way to deduplicate this over the similar lines just above, but no easy win comes to my mind.

rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -489,6 +516,13 @@ private J.EnumValue visitEnumField(FieldNode field) {
                         cursor++;
                     }
                     argCount++;
+                    // Safety check: if cursor didn't advance, throw an exception to avoid infinite loop
+                    if (cursor == cursorBeforeIteration) {
+                        throw new IllegalStateException(
+                                "Parser error: unable to parse enum constructor arguments at position " + cursor +
+                                        " near '" + source.substring(Math.max(0, cursor - 20), Math.min(source.length(), cursor + 20)) + "'"
+                        );
```
greg-at-moderne, line 524: That's a good one. Should make it easier to catch all these cases instead of us just identifying some mysteriously hanging build.

PR #6043, opened 2025-09-16
UpgradeDependencyVersion needs to support common Groovy property access notations

bmuschko:
## What's changed?

Adds support for parsing the property name when using the following APIs:

- [Project#property(String)](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#property(java.lang.String))
- [Project#findProperty(String)](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#findProperty(java.lang.String))
- [Project#getProperties()](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getProperties())

Using the `project` reference is implied in the Groovy DSL so we need to support the use with and without it. We also needed to make sure that GString interpolation works (while poor practices if just wrapped into double quotes without adding anything to the String).

## What's your motivation?

A property defined in `gradle.properties` can be accessed in a variety of ways in the Gradle Groovy DSL. This change supports all common notations.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -31,6 +31,7 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaSourceFile;
@@ -119,6 +120,8 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
     }
 
     private static final String UPDATE_VERSION_ERROR_KEY = "UPDATE_VERSION_ERROR_KEY";
+    private static final MethodMatcher PROPERTY_METHOD = new MethodMatcher("* property(String)");
+    private static final MethodMatcher FIND_PROPERTY_METHOD = new MethodMatcher("* findProperty(String)");
 
     @Value
     public static class DependencyVersionState {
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -416,6 +429,37 @@ private boolean shouldResolveVersion(String declaredGroupId, String declaredArti
                        new DependencyMatcher(groupId, artifactId, null).matches(declaredGroupId, declaredArtifactId);
             }
 
+            /**
+             * Extract property name from method invocation patterns like {@code property('guavaVersion')}
+             * or {@code findProperty('guavaVersion')}.
+             */
+            private @Nullable String extractPropertyNameFromMethodInvocation(J.MethodInvocation mi) {
+                String methodName = mi.getSimpleName();
+                if ("property".equals(methodName) || "findProperty".equals(methodName)) {
```
bmuschko: I had an issue with using `MethodMatcher`. It seems that `MethodMatcher` doesn't work well with Groovy method invocations that lack full type attribution. Is there a better way to achieve this?

timtebeek: Unfortunately no; @shanman190 has wrestled with this a lot as well. Tagging in case he has any new insights to share.

timtebeek (approved):
Good addition, thanks!

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
shanman190: Yeah, the closest that you can get is to define a `MethodMatcher` the way that you would want it, then use the `matches` override that allows setting `unknownTypes` to true.

bmuschko: Thanks! Pushed the change.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -119,6 +120,8 @@ public Validated<Object> validate() {
     }
 
     private static final String UPDATE_VERSION_ERROR_KEY = "UPDATE_VERSION_ERROR_KEY";
+    private static final MethodMatcher PROPERTY_METHOD = new MethodMatcher("* property(String)");
+    private static final MethodMatcher FIND_PROPERTY_METHOD = new MethodMatcher("* findProperty(String)");
```
shanman190, line 124: The class reference here can even be the fqcn as well. That way if/when these become type attributes, they'll match more accurately.

shanman190 (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
bmuschko: It looks like the matcher expression`org.gradle.api.Project property(String)` works for `findProperty('guavaVersion')` but it doesn't work for `project.findProperty('guavaVersion')`.

shanman190: That's peculiar. Hmm.

PR #6048, opened 2025-09-18
Groovy: Guard against NPE when coming across a multi variable assignment

Laurens-W:
## What's changed?

Added a guard for now -> a solution for proper parsing is being worked on 
- https://github.com/openrewrite/rewrite/pull/5293

## What's your motivation?

Instead of throwing a `GroovyParsingException` the code will now properly flow into this handling https://github.com/openrewrite/rewrite/blob/765ad965581a5868e341d13096092a3e17e8dbe3/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java#L1649-L1652 which gives a informational message on what's going on

## Any additional context

I did verify with a test but that test remains failing so there's no sense in adding it
```
Caused by: org.openrewrite.groovy.GroovyParsingException: Failed to parse file.groovy at cursor position 17. The surrounding characters in the original source are:
tag = '0.0.0'
def~cursor~> (major, minor, patch) = tag.tokenize('.')
	at org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:209)
	at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$3(GroovyParser.java:149)
	... 15 more
Caused by: java.lang.UnsupportedOperationException: Parsing multiple assignment (e.g.: def (a, b) = [1, 2]) is not implemented
```

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2338,7 +2338,7 @@ public class GroovyParserVisitor {
             queue.add(new G.Unary(randomId(), fmt, Markers.EMPTY, JLeftPadded.build(G.Unary.Type.Spread), visit(spreadExpression.getExpression()), null));
         }
 
-        public TypeTree visitVariableExpressionType(VariableExpression expression) {
+        public TypeTree visitVariableExpressionType(@Nullable VariableExpression expression) {
             if (expression == null) {
                 return null;
             }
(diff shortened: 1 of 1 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2339,6 +2339,9 @@ public void visitSpreadExpression(SpreadExpression spreadExpression) {
         }
 
         public TypeTree visitVariableExpressionType(VariableExpression expression) {
```
timtebeek:
```suggestion
        public TypeTree visitVariableExpressionType(@Nullable VariableExpression expression) {
```

timtebeek: If we're doing a null check the variable should be marked as nullable right?

timtebeek (approved):
Approved apart from the minor suggestion.

PR #6055, opened 2025-09-19
Parsers should target the previous LTS to support non-LTS versions

timtebeek:
## What's your motivation?
- In https://github.com/openrewrite/rewrite/pull/5985 we changed which parser is selected based on the active Java version.
- Now we're seeing folks continuing to run on non-LTS versions being unable to build their projects, as these parsers were compiled for a newer JVM version:
- https://github.com/openrewrite/rewrite-maven-plugin/issues/1038

Do we foresee any issues with targeting lower Java versions for the compiler specifically?

- We can't have them all target Java 8, as some parser functionality is only available in more recent versions. Hence why for instance the Java17Parser needs at least Java 15+.
- We could partially revert the Java version selection we had added in #5985, to again pick the LTS _before_ rather than _after_ a non-LTS release, with an optional additional restore of the reflection warts added to parsers to support features introduced before the next LTS release.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
@@ -168,7 +168,7 @@ public interface JavaParser extends Parser {
                         .forName("org.openrewrite.java.Java25Parser")
                         .getDeclaredMethod("builder")
                         .invoke(null);
-            } catch (Exception e) {
+            } catch (Throwable e) {
                 //Fall through, look for a parser on an older version.
             }
         }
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -168,7 +168,7 @@ static List<Path> dependenciesFromResources(ExecutionContext ctx, String... arti
                         .forName("org.openrewrite.java.Java25Parser")
                         .getDeclaredMethod("builder")
                         .invoke(null);
-            } catch (Exception e) {
+            } catch (Throwable e) {
```
timtebeek, line 171: This now also catches `UnsupportedClassVersionError`, meaning while'll continue down to try other parsers as well.

timtebeek:
Tried this locally with `pTML` on openrewrite/rewrite, followed by `mvn install` on rewrite-maven-plugin, and then upgrade Spring Boot Starter project using Java 24 from a Java 24 JDK to Java 25; that all worked.

PR #6056, opened 2025-09-19
`JavaTemplate.Matcher` mistakenly matches `0.0` to `0`

timtebeek:


timtebeek:
Any immediate insights come to mind @knutwannheden ?

knutwannheden:
I believe that was done on purpose: https://github.com/openrewrite/rewrite/blob/b91cf626b96a5e77169aecd7149b108082d4ef85/rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java#L894-L911

But I am sure that could be refined somehow.

knutwannheden:
What about comparing 0.0 against 0.0f?

timtebeek:
> What about comparing 0.0 against 0.0f?

Updated the handling to
- use `doubleValue()` when both left and right are either a double or a float.
- use `longValue()` when both left and right are any of byte/short/int/long
- return false otherwise, to fix the underlying issue where `0.0` != `0`

Ok to merge?

PR #6059, opened 2025-09-21
Inline annotated methods through generated recipes to support multiple versions

timtebeek:
## What's changed?
Change the relatively new `InlineMethodCalls` to take arguments, and provide a generator to generate declarative recipes based on annotations found in TypeTables.

## What's your motivation?
We want to support multiple (major) versions of Guava and other libraries that might use the annotations only in specific versions, before methods are dropped in subsequent versions. That means we would only be able to support the version that a project was compiled against, but not a chain of versions that might follow for those same methods. Additionally we were not able to use the `classpathFromResources` for such replacements, leaving us `.classpath` which does not work as well in all environments, as types might not be on the classpath there.

- I've placed the generator in the same package as the TypeTable itself, such that I could use package-private for some TypeTable nested classes, as opposed to `public`. I didn't want to expose those types publicly to avoid locking ourselves in too much.

We could move the generator to `rewrite-build-gradle-plugin` instead as a separate task.

## Any additional context
- https://github.com/openrewrite/rewrite/pull/5953


timtebeek:
On rewrite-migrate-java this generated 47 recipes:
```yaml
name: org.openrewrite.java.InlineMethodCallsGenerated
displayName: Inline methods annotated with `@InlineMe`
description: >
  Automatically generated recipes to inline method calls based on `@InlineMe` annotations
  discovered in the type table.
recipeList:

  # From com.google.guava:guava:33.4.8-jre
  - org.openrewrite.java.InlineMethodCalls:
      methodPattern: 'com.google.common.primitives.Booleans compare(boolean, boolean)'
      replacement: 'Boolean.compare(a, b)'
      classpathFromResources:
        - 'guava-33.4.8-jre'
  - org.openrewrite.java.InlineMethodCalls:
      methodPattern: 'com.google.common.primitives.Longs compare(long, long)'
      replacement: 'Long.compare(a, b)'
      classpathFromResources:
        - 'guava-33.4.8-jre'
...
```

timtebeek:
Here's what usage downstream looks like as a branch; to be opened and merged if this is approved and merged:
https://github.com/openrewrite/rewrite-migrate-java/compare/generate-guava-inlineme-recipes?

rewrite-java/src/test/java/org/openrewrite/java/internal/parser/InlineMethodCallsRecipeGeneratorTest.java
```diff
@@ -0,0 +1,166 @@
...
+            Automatically generated recipes to inline method calls based on `@InlineMe` annotations
+            discovered in the type table.
+          recipeList:
+
+            # From com.example:test-lib:1.0.0
+            - org.openrewrite.java.InlineMethodCalls:
+                methodPattern: 'com.example.TestClass annotatedMethod()'
+                replacement: 'replacement()'
+                classpathFromResources:
+                  - 'test-lib-1.0.0'
```
knutwannheden, line 163: I think it would be really nice if we could change the `JavaParser.Builder#classpathFromResources()` API a bit to work with GAVs. My proposal would be that we in a first step leave the signature as is, but change it so that if the value contains a `:`, then it is interpreted as a GA or GAV (depending on if there is one or two colons). For the version part we could then also allow version ranges like `1.0.+` or `~1.0.0`

```suggestion
                  - 'com.example:test-lib:1.0.0'
```

I realize this isn't directly related to this PR.

knutwannheden (approved):
I assume these recipes will then be part of a corresponding framework upgrade recipe, which will then make sure that we don't run one of these recipes, when they aren't yet applicable.

I think it is a bit of a shame that we need to do it this way, but I see why we need to do something like this, as we never recompile the source code against new dependencies.

PR #6065, opened 2025-09-23
Fix RemoveRedundantDependencyVersions comment handling with new GradleDependencies Trait

Jenson3210:
### Summary
This PR fixes issues with `RemoveRedundantDependencyVersions` where removing dependency statements would incorrectly handle comments, leading to comment displacement, loss of dependency block comments, and other formatting issues. The solution introduces a new reusable `GradleDependencies` Trait that properly manages statement removal while preserving surrounding comments and whitespace.

### Problem
Previously, `RemoveRedundantDependencyVersions` was not really taking comments and `gradle` formatting into account.
- Comments that were part of the next statement's prefix would shift to the previous element (resulting in CVE descriptions being linked to wrong dependencies
- Dependency block comments would be incorrectly removed
- Whitespace and formatting would be corrupted

### Solution
Introduced a new `GradleDependencies` Trait that:
- Properly handles comment preservation when filtering/removing statements from dependency blocks
- Maintains correct whitespace and formatting when statements are removed
- Provides a reusable pattern for similar Gradle dependency manipulations

The trait implements sophisticated logic to:
- Track and preserve comments attached to removed statements
- Correctly merge comments when adjacent statements are filtered
- Handle edge cases like end-of-line comments and block-level comments
- Maintain proper indentation and whitespace

### Changes
- **New**: `GradleDependencies` trait in `rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/`
  - Implements `withStatements(Predicate<Statement>)` for safe statement filtering
  - Handles complex comment preservation scenarios
  - Provides reusable functionality for Gradle dependency manipulation

- **Updated**: `RemoveRedundantDependencyVersions` recipe
  - Now uses `GradleDependencies` trait for clean statement removal
  - Collects statements to remove and applies filtering in a single pass
  - Preserves all surrounding comments and formatting correctly

- **Added**: Comprehensive test coverage
  - `GradleDependenciesTest` to validate the trait's behavior
  - Updated `RemoveRedundantDependencyVersionsTest` with comment preservation scenarios

### Testing
All existing tests pass, with new tests added to verify:
- Comment preservation when removing dependency statements
- Correct handling of multi-line comments
- Proper whitespace management
- Edge cases with various comment positions

### Impact
This change ensures that automated dependency cleanup operations preserve developer comments and maintain clean, readable build files. The new trait can be reused for other Gradle-related recipes that need to manipulate dependency blocks while preserving formatting.


Jenson3210:
Thanks for pushing the comments @timtebeek, you beat me to it :)

rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -736,6 +736,179 @@ void removeRepeatedDependency() {
         );
     }
 
+    @Test
+    void complexExample() {
+        rewriteRun(
+          spec -> spec.recipe(new RemoveRedundantDependencyVersions(null, null, RemoveRedundantDependencyVersions.Comparator.GTE)),
+          buildGradle(
+            """
+            group = 'io.moderne'
```
Jenson3210:
```suggestion
            group = 'org.openrewrite'
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveRedundantDependencyVersionsTest.java
```diff
@@ -736,6 +736,179 @@ void removeRepeatedDependency() {
...
+                // Test dependencies
+                testImplementation('org.springframework.boot:spring-boot-starter-test')
+                testImplementation('org.wiremock:wiremock-standalone:3.13.0')
+                testImplementation('io.projectreactor:reactor-test:3.7.6')
+                testImplementation('org.hamcrest:hamcrest:3.0')
+                testImplementation('org.springframework.boot:spring-boot-starter-web')
+            }
+            """,
+            """
+            group = 'io.moderne'
```
Jenson3210:
```suggestion
            group = 'org.openrewrite'
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
```diff
@@ -0,0 +1,333 @@
...
+@RequiredArgsConstructor
+public class GradleDependencies implements Trait<J.MethodInvocation> {
+    private static final MethodMatcher DEPENDENCY_DSL_MATCHER = new MethodMatcher("DependencyHandlerSpec *(..)");
+
+    @Getter
+    private final Cursor cursor;
+
+    private final List<Statement> statements;
+
+    public @Nullable GradleDependencies withStatements(Predicate<Statement> filter) {
```
Jenson3210: If we would prefer a `Function<Statement, @Nullable Statement>` we could also allow for mapping (null would remove) + the comments handling could be done with the mapping results as well as with the current approach. 
Small change that would open us up to a broader usage, but also to a broader ab-usage. 
We can also always add an overload later if we would ever feel the need for the function.

Jenson3210: I locally tested this and it would be a very easy change to get this working. 

The only thing that is not possible "easily" is to have a `Function<Statement, @Nullable Object>` to correspond with the `ListUtils.flatmap` as then we would have a harder time handling the comments addition to the correct items (eol comments of the current statement are in the next statements prefix -> keep some counter of how many items where added and add the aggregated eol comments to the second list item added , but keep the other comments at the prefix of the newly adding statement etc...)

sambsnyd: We use `with` methods all through the codebase to mean a specific thing. The semantics of this method will clash with the expectations set by the name. 

The concept of filtering a  list of LST elements and transferring comments around those that remain is more general than dependency manipulation. I wonder if that should go into some other trait or utility and be used elsewhere as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+
+import java.util.stream.Stream;
+
+@RequiredArgsConstructor
+public class GradleDependencies implements Trait<J.MethodInvocation> {
```
sambsnyd, line 46: A `dependencies` block is logically, more or less, a list of Gradle dependencies and constraints. So I'm a bit surprised to see a `GradleDependencies` trait that has no interoperability or relationship with the `GradleDependency` trait.

I've been thinking that many dependency management recipes might be simplified if we had a `GradleBuildFile` trait which contained `GradleDependencies` and `GradlePlugins` and other such utilities, centralized in one place. I would be interested to see if you can make these two traits compose well together and use them to simplify any of the other existing recipes help put this notion to the test.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
Jenson3210: I also wondered the same. But I did not want to introduce a more generic thing to void too many "core" changes introduced. 
I was thinking we could later on still move this out if we feel the same need occur somewhere else. 

I agree with the name, had already thought about that. Would you think `map` is a better name?

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
Jenson3210: The Dependencies block can contain `constraints` for example which the `GradleDependency` trait does not handle (and should not). 
Linking them tightly coupled will make them harder to use in "all" cases was my thinking. 
The only gradle specific thing here is the fact that it should be in a `dependencies {}` block. 

i agree with the Idea of a broader `GradleBuildFile` but then again that one would grow very complex very fast, so thinking this could act as a building block to One such trait. 

That one could offer `RemoveDependency` which then could use the combination of the 2 current Traits to do just that [like so](https://github.com/openrewrite/rewrite/pull/6065/files#diff-f89206fb46483522d6b44a9cf445b92b6ee9e499e7f03ba77dacbc5fe631f0caR45). 
If we do not entangle these togethere, the GradleDependencies one can also do stuff on constraint blocks etc if the user chooses to do something with these statements and the overlaying `GradleBuildFile` Trait could be the combiner of these functional things.

Jenson3210:
Marking as draft to avoid merge as there are some open items

shanman190 (commented):
So far everything looks good to me. There are few cases for each of the new traits that we should probably do something with.

1. `GradleDependencies` could use some treatment for nested `dependencies` blocks. The `GradleProject` marker only carries information about the current project and not any references to `allprojects`, `subprojects`, and `configure(project(..))`. So some specialized handling may be needed here. You can kinda glean what is possible on the current project via a union of `GradleDependency` and the `JvmTestSuite` traits.

2. For the `Block` trait, we likely also need to have some specialized handling for methods that return a value where the removed method is the return statement.

3. A case for both is also what the behavior should be when all statements have been removed and comments remain.

sambsnyd (approved):
I agree with Shannon's feedback, after that it looks good

rewrite-java/src/test/java/org/openrewrite/java/trait/BlockTest.java
```diff
@@ -0,0 +1,889 @@
...
+                      //comment at end
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void emptyBlockIsRemoved() {
```
Jenson3210:
```suggestion
    void emptyBlockIsRetained() {
```

Jenson3210:
Thanks! 

> 1. GradleDependencies could use some treatment for nested dependencies blocks. The GradleProject marker only carries information about the current project and not any references to allprojects, subprojects, and configure(project(..)). So some specialized handling may be needed here. You can kinda glean what is possible on the current project via a union of GradleDependency and the JvmTestSuite traits.

I did not even know that that was possible in gradle. Will for sure have a look at it. But as it just delegates the work to GradleDependency, and you mention that one already supports this, is that not sufficient? I was thinking this GradleDependencies is just a warpper around few other Traits (like doing the Groovy Method Invocation firs arugment mapping, removing the block if the Block trait returned an empty block (for Blocks we do not want that behavior as it would break compilation in java cases.

> 2. For the Block trait, we likely also need to have some specialized handling for methods that return a value where the removed method is the return statement.

I think I already [have this](https://github.com/openrewrite/rewrite/pull/6065/files#diff-15f585efab18869e6e3cbcb22074678595c0050407bff768c9f70720ec0b9732R56) and [this](https://github.com/openrewrite/rewrite/pull/6065/files#diff-15f585efab18869e6e3cbcb22074678595c0050407bff768c9f70720ec0b9732R142-R148) that takes this into account. I could not find a solution for `return 42;` as that is not a Statement but an Expression in the return. Any suggestions to cover for that would be great! Perhaps the Function should act on `Expression`s? But then we might be missing other elements that are Statement but no Expression?
The current behavior is tested in a few tests (these are not all): 
- [return gets removed (no eol)](https://github.com/openrewrite/rewrite/pull/6065/files#diff-59d8c58ac56795f0317723467d30df82a3d5a7a3d95f235bdfe42b930165867aR347)
- [return gets removed](https://github.com/openrewrite/rewrite/pull/6065/files#diff-59d8c58ac56795f0317723467d30df82a3d5a7a3d95f235bdfe42b930165867aR446)
- [return all but one (also last one is gone)](https://github.com/openrewrite/rewrite/pull/6065/files#diff-59d8c58ac56795f0317723467d30df82a3d5a7a3d95f235bdfe42b930165867aR635)

> 3. A case for both is also what the behavior should be when all statements have been removed and comments remain.

This should also be [covered](https://github.com/openrewrite/rewrite/pull/6065/files#diff-59d8c58ac56795f0317723467d30df82a3d5a7a3d95f235bdfe42b930165867aR144-R203) already I guess for the Block. 
In the GradleDependencie block, we do [remove if no comments remaining](https://github.com/openrewrite/rewrite/pull/6065/files#diff-f89206fb46483522d6b44a9cf445b92b6ee9e499e7f03ba77dacbc5fe631f0caR145-R209) when the [Block trait has returned an empty block](https://github.com/openrewrite/rewrite/pull/6065/files#diff-5b9edb1e63a1b0863af862bc6fe4632b46f1068b6b5bb8e85e71f7f9c4ef511dR69).

Jenson3210:
@shanman190 Could you clarify the first one. Googled / chatGpt'ed a bit and could not find an example :/ 

I was interpreting your comment that 
```
dependencies {
    dependencies {
    }
}
```
should need special handling. 

If you meant: 
```
subProjects {
        dependencies {
    }
}
```
needs handling, I think we already do that[ by calling super.visit](https://github.com/openrewrite/rewrite/pull/6065/files#diff-5b9edb1e63a1b0863af862bc6fe4632b46f1068b6b5bb8e85e71f7f9c4ef511dR105)

shanman190:
@Jenson3210 , given it's just a wrapper around the other traits then things probably will work just fine and avoid the `subprojects`, etc cases (because `GradleDependency` does internally). So we should be good here.

It does seem like you're handling the return case, but maybe not in the new last statement has a different return type than the original?

I still need to follow up on the other comments.

Jenson3210:
> maybe not in the new last statement has a different return type than the original?

Hmmm.. But how would we go around that? We could pass the return statement to the mapper iso unwrapping its expression. Then the responsability of handling falls out of the Block Trait (which makes sense as that one does not know if we are in a Method bode, class body, just a block...) (meaning we then need to check for return statements in all the consumers, but still that is cleaner than making some behavior that breaks in certain cases)

fwiw, `ListUtils` also does not "unbox" return statements.

I was actually thinking, maybe we do not even need a J.Block Trait and could just go for `StatementListUtils`?
This one could then offer a `filter(List<Statement>, Predicate)`, `map(List<Statement>, Function<Statement, Statement>)`

sambsnyd:
> maybe not in the new last statement has a different return type than the original?

That might be asking too much of the `map` function. I don't think we can eliminate every possible mistake in `map`, not removing necessary statements is at least partially the responsibility of the caller.

shanman190:
> That might be asking too much of the map function. I don't think we can eliminate every possible mistake in map, not removing necessary statements is at least partially the responsibility of the caller.

I agree with you. The problem that I see here is that the caller isn't aware of where the statement is. `filter` and `map` just pass in a single statement without any additional information then except to get back a boolean / statement or `null` respectively. The caller doesn't have the ability to make the judgement of what should be done.

Jenson3210:
I added a method that allows for the user to know this. By default we will just wrap the new last value in the return statement. This could indeed change the method's return type but that is not up to the block trait to decide but up to the user. There is also a method available that accepts a BiFunction that receives the old return, new last statement and should give back the resulting return. 
This could result in cyclic changes if that one returns null so I have also build in a validation for that. If the new return statement would be returning null after the mapping, the validation phase of tests will try to remove that one again.

Jenson3210:
Merged as discusses offline with @sambsnyd. 
Opening a followup PR where we will tackle extracting the block trait into a more generic Block trait where we have time for debate on the contract

Issue #6067, opened 2025-09-24
java.lang.ArrayIndexOutOfBoundsException in org.openrewrite.groovy.GroovyParserVisitor.determineParenthesisLevel


jochenberger:
## What version of OpenRewrite are you using?

I am using

- Gradle plugin 7.16.0

## What is the smallest, simplest way to reproduce the problem?

```groovy
package org.example

import spock.lang.Specification

class GroovyTest extends Specification {

    def "formatting test"() {
        expect:
        1 == 2
    }
}

```
## What is the full stack trace of any errors you encountered?

```
Error during rewrite run
org.openrewrite.groovy.GroovyParsingException: Failed to parse veo-adapter/src/test/groovy/GroovyTest.groovy at cursor position 161. The surrounding characters in the original source are:
package org.example

import spock.lang.Specification

class GroovyTest extends Specification {

    def "formatting test"() {
        expect:
        1 == 2
    ~cursor~>}
}

        at org.openrewrite.groovy.GroovyParserVisitor.visit(GroovyParserVisitor.java:209)
        at org.openrewrite.groovy.GroovyParser.lambda$parseInputs$3(GroovyParser.java:149)
        at java.base@21.0.9-ea/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base@21.0.9-ea/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
        at java.base@21.0.9-ea/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base@21.0.9-ea/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base@21.0.9-ea/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.base@21.0.9-ea/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.base@21.0.9-ea/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base@21.0.9-ea/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
        at java.base@21.0.9-ea/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276)
[...]

PR #6069, opened 2025-09-24
`ArrayIndexOutOfBoundsException` in `GroovyParserVisitor.determineParenthesisLevel` for Spock test

timtebeek:
- Closes https://github.com/openrewrite/rewrite/issues/6067


timtebeek:
My potentially lying assistant tells me:
> The test is using spock-core on the classpath. The issue is that when Spock is on the classpath, the Groovy compiler applies Spock's AST transformations automatically. This is a fundamental limitation - we can't parse Spock tests exactly as written when Spock is on the classpath because the transformations happen at compile time.

PR #6076, opened 2025-09-25
Fix HCL parser idempotency issue with spaces before closing brace

### LINKED ISSUE #6073: HCL Parser idempotency issue with spaces before closing brace

Current main = c295c36d255907c648fa98d10adda48b8f0e0418

The following test:
```java
    @Test
    void spaceInBraces() {
        rewriteRun(
          hcl("""
            foo {
              default = "${ var.resource_group_name }"
            }
          """));
    }
```
when added to `HclQuotedTemplateTest` fails with:
```hcl
expected: 
  "foo {
    default = "${ var.resource_group_name }"
  }"
 but was: 
  "foo {
    default = "${ var.resource_group_name}"
  }"
```

## OSS repro

-  https://github.com/Azure/aztack/blob/e35cd7e582384c9e1419a0f2ea956c6c7c25b1a8/terraform/modules/rg/rg.tf

timtebeek:
## Summary
- Fixes an HCL parser issue where spaces before the closing brace in template interpolations were being dropped
- Resolves the test failure when parsing `"${ var.resource_group_name }"` which was incorrectly printed as `"${ var.resource_group_name}"`

## What's changed
- Modified `TemplateInterpolation` class to use `HclRightPadded<Expression>` instead of just `Expression` to preserve trailing whitespace
- Added proper padding support following the established pattern in other HCL tree classes
- Updated the parser visitor to capture spaces before the closing brace
- Updated the printer to properly output the preserved spaces
- Added necessary enum entries to `Space.Location` and `HclRightPadded.Location`
- Updated `HclVisitor` to handle the new padding structure

## Test plan
- [x] Added test `spaceInBraces()` to `HclQuotedTemplateTest` to reproduce the issue
- [x] Verified test passes with the fix
- [x] Ran all HCL tests to ensure no regressions

- Fixes #6073


rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
```diff
@@ -1531,20 +1531,35 @@ public enum Type {
...
+        @Getter
         Space prefix;
 
         @With
+        @Getter
         Markers markers;
 
-        @With
-        Expression expression;
+        HclRightPadded<Expression> expression;
```
timtebeek: This change likely has an effect on serialization; if any whitespace were present we would not have been able to parse those files previously, but if there was no whitespace we might not be able to deserialize existing LTS. Thoughts?

knutwannheden: I am thinking we could address this by having a package private `@JsonCreator`, which declares the `expression` parameter as of type `Object` and then you can do an `instanceof` and wrap it into a `HclRightPadded` if it's an `Expression`.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
```diff
@@ -1531,20 +1533,51 @@ public enum Type {
...
+            this.id = id;
+            this.prefix = prefix;
+            this.markers = markers;
+            if (expression instanceof Expression) {
+                this.expression = HclRightPadded.withElement(null, (Expression) expression);
+            } else {
+                //noinspection unchecked
+                this.expression = (HclRightPadded<Expression>) expression;
+            }
+        }
```
timtebeek, line 1572: @knutwannheden hope I got this right based on your comment above in https://github.com/openrewrite/rewrite/pull/6076/files#r2391939772

rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/HclPrinter.java
```diff
@@ -343,7 +343,7 @@ public Hcl visitQuotedTemplate(Hcl.QuotedTemplate template, PrintOutputCapture<P
     public Hcl visitTemplateInterpolation(Hcl.TemplateInterpolation template, PrintOutputCapture<P> p) {
         beforeSyntax(template, Space.Location.TEMPLATE_INTERPOLATION, p);
         p.append("${");
-        visit(template.getExpression(), p);
+        visitRightPadded(singletonList(template.getPadding().getExpression()), HclRightPadded.Location.TEMPLATE_INTERPOLATION, "", p);
```
knutwannheden: This should also work:
```suggestion
        visitRightPadded(template.getPadding().getExpression(), HclRightPadded.Location.TEMPLATE_INTERPOLATION, p);
```

knutwannheden (commented):
I think it looks right. To make extra sure we should perhaps test it.

rewrite-hcl/src/main/java/org/openrewrite/hcl/tree/Hcl.java
knutwannheden: It doesn't work unfortunately, as Jackson passes it in as a `Map`... We need to change the name of the field and then this constructor needs to declare parameters for both the old and the new field.

Issue #6044, opened 2025-09-16
Credentials from configured artifact repositories are not populated when loading repositories


martamedio:
## What version of OpenRewrite are you using?
I'm using latest releases

- OpenRewrite v8.62.1
- Gradle Plugin 7.16.0

## How are you running OpenRewrite?
I'm using Gradle plugin to run `org.openrewrite.gradle.AddDependency` recipe using [`latest.release` or XRange version selector](https://docs.openrewrite.org/reference/dependency-version-selectors):
```
---
type: specs.openrewrite.org/v1beta/recipe
name: dynamic.custom.org.openrewrite.java.dependencies.AddDependency
recipeList:
  - org.openrewrite.gradle.AddDependency:
      groupId: com.project.custom
      artifactId: custom-project-starter
      version: '1.5.X'
```

The dependency is located at a remote repository that needs authentication, so I'm configuring the repo and its credentials on my `init.gradle`:

```gradle
apply plugin: EnterpriseRepositoryPlugin

class EnterpriseRepositoryPlugin implements Plugin<Gradle> {

    void apply(Gradle gradle) {
        gradle.allprojects { project ->
            project.repositories {
                mavenLocal()
                maven {
                    name "MY_ENTERPRISE_REPO"
                    url "https://myenterprise.com/artifactory/enterprise"

                    credentials {
                        username = 'username'
                        password = 'custom-password'
                    }
                    authentication {
                        basic(BasicAuthentication)
                    }
                }
           }
        }
    }
}
```

## What is the smallest, simplest way to reproduce the problem?

I've detected that is not possible to pull the `maven-metadata.xml` file from the repository to discover the latest release because the credentials are ignored when loading the available repositories.

I think the easiest way for the OpenRewrite team to test it, is to [modify `AddDependencyTest.addDynamicVersionDependency` trying to add a dependency located in a Repository needing authentication](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java#L1120), and configure that Repository in your Gradle settings like described above.
You'll see an error:
```
Unable to download metadata. Tried repositories:
  https://myenterprise.com/artifactory/enterprise/: HTTP 401
....
```

Another way is just executing the recipe like described above and you'll see that is not possible to resolve the latest version of the dependency and won't be added to the `build.gradle` of the project.

## What is the full stack trace of any errors you encountered?

[...]

PR #6096, opened 2025-10-01
Use Maven repository credentials in `GradleProjectBuilder`

sullis:
## What's changed?

This PR changes the logic in GradleProjectBuilder mapRepositories

We are now mapping the Maven Repository "username" and "password" if they exist.

## What's your motivation?

- Closes https://github.com/openrewrite/rewrite/issues/6044

### Checklist

### DIFF
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
@@ -94,8 +94,10 @@ public final class GradleProjectBuilder {
                             .uri(repo.getUrl().toString())
                             .releases(true)
                             .snapshots(true);
-                    Optional.ofNullable(repo.getCredentials().getUsername()).ifPresent(builder::username);
-                    Optional.ofNullable(repo.getCredentials().getPassword()).ifPresent(builder::password);
+                    if (GradleVersion.current().compareTo(GradleVersion.version("6.6")) >= 0) {
+                        Optional.ofNullable(repo.getCredentials().getUsername()).ifPresent(builder::username);
+                        Optional.ofNullable(repo.getCredentials().getPassword()).ifPresent(builder::password);
+                    }
                     return builder.build();
                 })
                 .collect(toList());
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
@@ -90,6 +90,8 @@ public final class GradleProjectBuilder {
                 .map(repo -> MavenRepository.builder()
                         .id(repo.getName())
                         .uri(repo.getUrl().toString())
+                        .username(repo.getCredentials().getUsername())
+                        .password(repo.getCredentials().getPassword())
                         .releases(true)
                         .snapshots(true)
                         .build())
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
@@ -88,19 +88,16 @@ public final class GradleProjectBuilder {
         return repositories.stream()
                 .filter(MavenArtifactRepository.class::isInstance)
                 .map(MavenArtifactRepository.class::cast)
-                .map(repo -> {
-                    MavenRepository.Builder builder = MavenRepository.builder()
-                            .id(repo.getName())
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
```diff
@@ -76,6 +76,48 @@ void noopUpgrade() {
...
+            "implementation",
+            (original, updated) -> assertThat(updated)
+              .isSameAs(original)
+              .extracting(GradleProject::getMavenRepositories)
+              .satisfies(list -> assertThat(list)
+                .singleElement()
+                .satisfies(repo -> {
+                      assertThat(repo.getUri()).isEqualTo("https://example.com/maven2");
+                      assertThat(repo.getUsername()).isEqualTo("dummyuser");
+                      assertThat(repo.getPassword()).isEqualTo("dummypass");
```
timtebeek: Somehow these assertions still fail locally; figured it'd be good to verify these are picked up by the builder, but I'm not sure if the builder is even exercised for this test path.

rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
```diff
@@ -94,15 +94,22 @@ static List<MavenRepository> mapRepositories(List<ArtifactRepository> repositori
...
                 .collect(toList());
     }
 
+    private static void mapAuthentication(MavenArtifactRepository repo, MavenRepository.Builder builder) {
+        try {
+            PasswordCredentials passwordCredentials = repo.getCredentials(PasswordCredentials.class);
+            Optional.ofNullable(passwordCredentials.getUsername()).ifPresent(builder::username);
+            Optional.ofNullable(passwordCredentials.getPassword()).ifPresent(builder::password);
+        } catch (IllegalArgumentException e) {
+            // We're not using password credentials
```
shanman190, line 106: Interestingly enough, `MavenRepository` hasn't been updated to support HTTP Header authentication which was quite peculiar, but I know we've seen other Gradle users using that type, so we have to guard against it here.

There's also AWS credentials, but these are further out of scope right now.

rewrite-gradle-tooling-model/plugin/src/main/java/org/openrewrite/gradle/toolingapi/ToolingApiOpenRewriteModelPlugin.java
```diff
@@ -71,4 +73,12 @@ public Object buildAll(String modelName, Project project) {
...
         }
     }
+
+    interface MavenRepositoryMixin {
+        @JsonProperty(access = JsonProperty.Access.READ_WRITE)
+        String getUsername();
+
+        @JsonProperty(access = JsonProperty.Access.READ_WRITE)
+        String getPassword();
+    }
```
shanman190, line 83: In the `MavenRepository`, we prevent username and password from being written into the LST while here we are attempting to transfer it from the Gradle process to the tooling API where it is acceptable to serialize it.

rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
shanman190: Fixed via https://github.com/openrewrite/rewrite/commit/aa9c44d7cd6b036bf1a80167a183f62fe667b03b

rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
```diff
@@ -76,6 +79,195 @@ void noopUpgrade() {
...
+                    assertThat(repo.getPassword()).isNull();
+                }))
+          )),
+          properties(
+            """
+            mySecureRepositoryUsername=dummyuser
+            mySecureRepositoryPassword=dummypass
+            """,
+            spec -> spec.path("gradle.properties")
+          ),
```
shanman190:
```suggestion
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
```diff
@@ -76,6 +78,184 @@ void noopUpgrade() {
...
+            new GroupArtifactVersion("org.openrewrite", "rewrite-java", "8.56.0"),
+            "implementation",
+            (original, updated) -> assertThat(updated)
+              .isSameAs(original)
+              .satisfies(gp -> assertThat(gp.getMavenRepositories())
+                .singleElement()
+                .satisfies(repo -> {
+                    assertThat(repo.getUri()).isEqualTo("https://example.com/maven2");
+                    assertThat(repo.getUsername()).isNull();
+                    assertThat(repo.getPassword()).isNull();
```
shanman190, line 230: Ideally, we'd check for the header with to be present here instead, but we don't yet have the necessary LST elements for that.

PR #6098, opened 2025-10-01
Fix Java parser for annotated varargs by adding marker on `J.ArrayType`, as opposed to field on `J.VariableDeclarations`

### LINKED ISSUE #3881: Parser: not idempotent on annotated varargs
I am using

- Maven/Gradle plugin v5.17.1

```java
class ParserError {
    private String method(@NonNull final String @NonNull... args) {
        return "";
    }
}
```

```diff
java.lang.IllegalStateException: ParserError.java is not print idempotent.
(several lines removed)
-    private String method(@NonNull final String @NonNull... args) {
+    private String method(@NonNull final String[] @NonNull... args) {
         return "";
     }
 
  org.openrewrite.Parser.requirePrintEqualsInput(Parser.java:52)
```
The NonNull is from org.eclipse.jdt.annotations, but that should work for any type annotation which can be applied to arrays and types. When a normal array is used instead of the varargs, the parser doesn't produce an error.

probably not, I only feel confident in the recipes area, not the parser

### LINKED ISSUE #5742: Parse error in Java parser when using type annotation with varags
## Used version
rewrite-java 8.56.1

## The gist
`Object @Nullable ... other` is interpreted as `Object @Nullable[] ... other`

while e.g. the following is interpreted correctly:
`Object @Nullable[] ... other`
`Object ... other`

## Issue fully replicated in test:

```
package de.domagoj.or.test;

import static org.openrewrite.java.Assertions.java;

import org.openrewrite.java.JavaParser;
import org.openrewrite.test.RecipeSpec;

public class MyTest implements RewriteTest {
        @Test
        void shouldParse() {
            rewriteRun(
                java(
                    """
package de.domagoj.or.test.example;

import static java.lang.annotation.ElementType.TYPE_USE;

import java.lang.annotation.Documented;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

public interface Itf {
    @Documented
    @Retention(RetentionPolicy.CLASS)
    @Target({ TYPE_USE })
    public @interface NonNull {
    }
    @Documented
    @Retention(RetentionPolicy.CLASS)
    @Target({ TYPE_USE })
    public @interface Nullable {
    }


    @NonNull
    Object exampleMethod(@NonNull Object @Nullable ... other);
}
                    """
                )
            );
        }

}
```

knutwannheden:
Using a new `Varargs` marker.

TODO:
* [x] Fix all parsers
* [x] Phase out `varargs` field

- Fixes https://github.com/openrewrite/rewrite/issues/3881
- Fixes https://github.com/openrewrite/rewrite/issues/5742


timtebeek:
Let me know if you'd like help replicating this pattern to the Java 8, 11, 17 and 25 parsers. Would be good to see this resolved.

timtebeek (approved):
Approved from my side; anything that had made you held off on merging?

I'd seen this fail in a number of projects, including when serializing the LST, so figured good to clear out.

rewrite-java/src/test/java/org/openrewrite/java/format/SpacesVisitorTest.java
```diff
@@ -373,7 +373,7 @@ void method(  String   first  ,   int  ...   values  ) {
             """,
             """
             class Test {
-                void method(String first, int... values) {}
+                void method(String first, int ... values) {}
```
timtebeek, line 376: Above we had two spaces; after normalization we are now left with one instead of zero. I think that's fine?

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -6109,6 +6109,8 @@ final class VariableDeclarations implements J, Statement, TypedTree {
         @With
         @Nullable
         @Getter
+        @Deprecated
+        @ToBeRemoved(after = "2026-04-01")
         Space varargs;
```
timtebeek, line 6175: It's still used once with a non-null value in the Groovy parser, but figured we can revise that if/when we decide to drop the field. Most other cases it's `null` already, and then there will be the long tail of deserialization support.

timtebeek:
Tagging Jonathan for review since we're adding a marker to the LST model, and deprecating a field (not yet removed).

rewrite-java/src/main/java/org/openrewrite/java/marker/Varargs.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Indicates that a {@link org.openrewrite.java.tree.J.ArrayType} represents a varargs parameter type.
+ * When present, the array type should be printed using {@code ...} syntax instead of {@code []}.
+ * <p>
+ * For example, {@code String...} instead of {@code String[]}.
+ */
+@Value
+@With
+public class Varargs implements Marker {
+    UUID id;
```
timtebeek, line 33: This marker could use a review before we merge, since we're adding a marker for what used to be a field.

sambsnyd (commented):
Basically all languages have some concept of varargs. Can we standardize this across all of them? Assuming it's agreed this is the way to go.

greg-at-moderne:
> all languages have some concept of varargs. Can we standardize this across all of them? 

Just to help with the discussion, I think below is a fair description of current state of affairs. TLDR: it's fairly consistent.

- **Java**: `varargs` field of type `Space` on `J.VariableDeclarations`, this PR aims to change it to a marker
- **JS/TS**: same `varargs` field of type `Space` on `J.VariableDeclarations`, plus custom `JS.Spread` for `...` call-site syntax
- **Python**: same `varargs` field of type `Space` on `J.VariableDeclarations` (`*args`, `**kwargs`), plus custom `Py.Star` for call-sites
- **C#**: the language has different syntax: `params`, stored as a `J.Modifier`
- **Go**: same `varargs` field of type `Space` on `J.VariableDeclarations`
- **Scala**: not implemented yet (?)
- **Kotlin**: the language has different syntax: `vararg`, stored as a `J.Modifier`
- **Groovy**: `varargs` field of type `Space` on `J.VariableDeclarations`


Comment:
- I think it's fair to call it fairly consistent, even if not ideal
- I understand Sam's comment as a call for changing other parsers to be in line with Java, if we decide to make the change for Java

knutwannheden:
Thanks @greg-at-moderne for the survey — couple of corrections after digging into each.

**Scala** is implemented, but not using the `varargs` field for varargs. It repurposes the `Space` slot to store whitespace before the colon in field type ascription (`ScalaTreeVisitor.scala:2772`, e.g. `given IntSchema : SchemaFor[Int]`). Cleaning that up is independent of varargs and can be its own small PR (probably an `S.SpaceBeforeColon` marker on `J.VariableDeclarations`).

The other languages share a more interesting pattern: **JS/TS, Python, and Go all actively read/write this same field for their own native varargs syntax**, not just relay it over RPC:

- JS/TS prints `...args` rest parameters (`print.ts:497-499`)
- Python prints `*args`/`**kwargs` (`printer.py:1852-1854` — the `Space` is the whitespace before `*`)
- Go prints variadic `...T` (`go_printer.go:332-334`)

So this isn't a Java field that other languages happen to relay — it's a shared whitespace bucket every variable-declaration-having frontend is squatting on. Only C# is pure relay (it uses a `params` modifier natively).

That reshapes the plan. I think the right path is: land this PR as Java-only now, then track the cross-language work separately:

1. **Java** (this PR) — `Varargs` marker on `J.ArrayType`
2. **Groovy** — same pattern (also `String...` syntax)
3. **JS/TS** — marker on the rest-parameter representation
4. **Python** — dedicated field/marker for `*` / `**`
5. **Go** — marker on variadic-param representation
6. **Scala** — orthogonal: stop misusing the field for `:` whitespace

Once 1–5 land we can drop the field; until then it stays. I've also dropped the `@ToBeRemoved(after = "2026-04-01")` annotation in the latest push — no realistic removal date until the cross-language migration completes, and a stale date is worse than none. I'll open a tracking issue for 2–6.

rewrite-java/src/test/java/org/openrewrite/java/format/SpacesVisitorTest.java
knutwannheden: Yep, intentional — `SpacesVisitor` doesn't have a rule that strips the space, just normalizes it. Going from two spaces to one is the expected behavior.

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
knutwannheden: Updated based on the cross-language audit (see [main thread comment](https://github.com/openrewrite/rewrite/pull/6098#issuecomment-4307970752)): keeping `@Deprecated` (right signal for Java recipe authors — use the marker), but removed `@ToBeRemoved` since JS/TS, Python, and Go are all actively using the field for their own native varargs syntax. Removal needs the cross-language migration tracked above.

rewrite-java/src/main/java/org/openrewrite/java/marker/Varargs.java
knutwannheden: Sam reviewed on Apr 1 — marking resolved.

PR #6099, opened 2025-10-01
ChangeValue for Json should use JsonParser to interpret the target value argument

timtebeek:
We saw folks needing to escape arguments awkwardly, especially when the target value contains quotes. By handing off to the parser we hope to see this resolved.

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
@@ -17,14 +17,14 @@ package org.openrewrite.json;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.openrewrite.*;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
 import org.openrewrite.json.tree.Json;
 import org.openrewrite.json.tree.JsonValue;
-import org.openrewrite.marker.Markers;
 
 import java.util.Optional;
-
-import static org.openrewrite.Tree.randomId;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
@@ -66,12 +66,23 @@ public class ChangeValue extends Recipe {
                 }
 
                 String targetValue = value;
-                if (m.getValue() instanceof Json.Literal && ((Json.Literal) m.getValue()).getSource().equals(targetValue)) {
-                    return m;
+                String withoutQoutes = targetValue;
+                if (targetValue.startsWith("\"") || targetValue.startsWith("'")) {
+                    withoutQoutes = withoutQoutes.substring(1, withoutQoutes.length() - 1);
+                }
+                String withQuotes = targetValue;
+                if (!(targetValue.startsWith("\"") || targetValue.startsWith("'"))) {
+                    withQuotes = "\"" + withQuotes + "\"";
                 }
 
+                if (m.getValue() instanceof Json.Literal &&
+                        (((Json.Literal) m.getValue()).getSource().equals(targetValue) ||
+                                ((Json.Literal) m.getValue()).getSource().equals(withoutQoutes) ||
+                                ((Json.Literal) m.getValue()).getSource().equals(withQuotes))) {
+                    return m;
+                }
                 Optional<JsonValue> jsonValue = JsonParser.builder().build()
-                        .parse(targetValue, '"' + targetValue + '"')
+                        .parse(withoutQoutes, targetValue, withQuotes)
                         .filter(it -> it instanceof Json.Document)
                         .findFirst()
                         .map(Json.Document.class::cast)
--- a/rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
@@ -17,12 +17,12 @@ package org.openrewrite.json;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Option;
-import org.openrewrite.Recipe;
-import org.openrewrite.TreeVisitor;
+import org.openrewrite.*;
 import org.openrewrite.json.tree.Json;
+import org.openrewrite.json.tree.JsonValue;
 import org.openrewrite.marker.Markers;
+
+import java.util.Optional;
 
 import static org.openrewrite.Tree.randomId;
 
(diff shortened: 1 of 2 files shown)


rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
```diff
@@ -61,17 +59,33 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                    if (!(m.getValue() instanceof Json.Literal) || !((Json.Literal) m.getValue()).getSource().equals(ChangeValue.this.value)) {
-                        m = m.withValue(new Json.Literal(randomId(), m.getValue().getPrefix(), Markers.EMPTY, ChangeValue.this.value, source));
-                    }
+                if (!matcher.matches(getCursor()) || m.getMarkers().findFirst(Changed.class).isPresent()) {
+                    return m;
                 }
-                return m;
+
+                String withQuotes = value;
+                String withoutQoutes = value;
```
zieka: typo Quotes

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +84,169 @@ void changeArrayValue() {
...
+          json(
+            """
+              { "apiVersion": "123" }
+              """,
+            """
+              { "apiVersion": 123 }
+              """
+          )
+        );
+    }
```
zieka, line 266: I imagine we also want like a null test case

```suggestion
    }
    
  @Test                                                                                                                                                                                                                                                                          
  void changeToNull() {                                                                                                                                                                                                                                                          
      rewriteRun(                                                                                                                                                                                                                                                                
        spec -> spec.recipe(new ChangeValue("$.apiVersion", "null")),                                                                                                                                                                                                            
        json(                                                                                                                                                                                                                                                                    
          """                                                                                                                                                                                                                                                                    
          { "apiVersion": "v1" }                                                                                                                                                                                                                                                 
          """,                                                                                                                                                                                                                                                                   
          """                                                                                                                                                                                                                                                                    
          { "apiVersion": null }                                                                                                                                                                                                                                                 
          """                                                                                                                                                                                                                                                                    
        )                                                                                                                                                                                                                                                                        
      );                                                                                                                                                                                                                                                                         
  }                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                 
  @Test                                                                                                                                                                                                                                                                          
  void changeToStringLiteralNull() {                                                                                                                                                                                                                                             
      rewriteRun(                                                                                                                                                                                                                                                                
        spec -> spec.recipe(new ChangeValue("$.apiVersion", "\"null\"")),                                                                                                                                                                                                        
        json(                                                                                                                                                                                                                                                                    
          """                                                                                                                                                                                                                                                                    
          { "apiVersion": "v1" }                                                                                                                                                                                                                                                 
          """,                                                                                                                                                                                                                                                                   
          """                                                                                                                                                                                                                                                                    
          { "apiVersion": "null" }                                                                                                                                                                                                                                               
          """                                                                                                                                                                                                                                                                    
        )                                                                                                                                                                                                                                                                        
      );
  }

```

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +85,215 @@ void changeArrayValue() {
...
+    @Test
+    void changeToStringSingleQuotes() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeValue("$.apiVersion", "'v2'")),
+          json(
+            """
+              { "apiVersion": "v1" }
+              """,
+            """
+              { "apiVersion": 'v2' }
```
timtebeek: This one is a little questionable; our parser tolerates it, which is why it's passed through.

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +85,215 @@ void changeArrayValue() {
...
+    @Test
+    void changeToStringSingleQuotesAroundDoubleQuotes() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeValue("$.apiVersion", "'\"v2\"'")),
+          json(
+            """
+              { "apiVersion": "v1" }
+              """,
+            """
+              { "apiVersion": '"v2"' }
```
timtebeek: Same here; tolerated by our parser and hence passed through.

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +84,214 @@ void changeArrayValue() {
...
+    @Test
+    void changeToNumberAsString() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeValue("$.apiVersion", "\"123\"")),
+          json(
+            """
+              { "apiVersion": "v1" }
+              """,
+            """
+              { "apiVersion": 123 }
```
timtebeek: Since we only have special handling for `"null"` this `"123"` String gets turned into an number. This would once again require double quoting I believe to prevent that conversion, or a change in the parsing logic.

rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
```diff
@@ -57,21 +60,38 @@ public String getDescription() {
...
-                    }
-                    if (!(m.getValue() instanceof Json.Literal) || !((Json.Literal) m.getValue()).getSource().equals(ChangeValue.this.value)) {
-                        m = m.withValue(new Json.Literal(randomId(), m.getValue().getPrefix(), Markers.EMPTY, ChangeValue.this.value, source));
+                if (matcher.matches(getCursor()) && !m.getMarkers().findFirst(Changed.class).isPresent()) {
+                    if (!jsonValue.isPresent()) {
+                        return SearchResult.found(m, "Could not parse value: " + value);
                     }
+                    JsonValue parsedValue = jsonValue.get();
+                    return m.withValue(parsedValue.withPrefix(m.getValue().getPrefix()))
+                            .withMarkers(parsedValue.getMarkers().add(new Changed(Tree.randomId())));
```
zieka: should this be the member itself getting the marker?

```suggestion
                    JsonValue parsedValue = jsonValue.get();
                    return m.withValue(parsedValue.withPrefix(m.getValue().getPrefix()))
                            .withMarkers(m.getMarkers().add(new Changed(Tree.randomId())));
```

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
zieka: ya since 
```json
{ "apiVersion": 'v2' }
```
 is invalid JSON i would have expected the input to be mapped to a JSON string literal resulting in:
```json
{ "apiVersion": "'v2'" }
```

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +84,214 @@ void changeArrayValue() {
...
+    @Test
+    void changeToStringSingleQuotesAroundDoubleQuotes() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeValue("$.apiVersion", "'\"v2\"'")),
+          json(
+            """
+              { "apiVersion": "v1" }
+              """,
+            """
+              { "apiVersion": '"v2"' }
```
zieka:
```json
{ "apiVersion": '"v2"' }
```

this is invalid JSON and I would have expected the parsing to fail and then this inputed value to be treated as a string:
```json
{ "apiVersion": "'\"v2\"'" }
```

rewrite-json/src/test/java/org/openrewrite/json/ChangeValueTest.java
```diff
@@ -84,4 +84,214 @@ void changeArrayValue() {
...
+    @Test
+    void changeToStringWithDoubleQuotesAroundSingleQuotes() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeValue("$.apiVersion", "\"'v2'\"")),
+          json(
+            """
+              { "apiVersion": "v1" }
+              """,
+            """
+              { "apiVersion": "'v2'" }
```
zieka, line 172: questionable assumption but I think this might be fine

rewrite-json/src/main/java/org/openrewrite/json/ChangeValue.java
timtebeek: Yes thanks! Didn't show in tests, but we should retain markers from the original member, not the parsed value.

zieka (approved):
Approving

At a later point we might consider adding these tests:

```java
@Test
void intentionalSingleQuotedNullAsString() {
    rewriteRun(
      spec -> spec.recipe(new ChangeValue("$.apiVersion", "'null'")),
      json(
        """
        { "apiVersion": "v1" }
        """,
        """
        { "apiVersion": "'null'" }
        """
      )
    );
}

@Test
void intentionalSingleQuotedNumberAsString() {
    rewriteRun(
      spec -> spec.recipe(new ChangeValue("$.apiVersion", "'123'")),
      json(
        """
        { "apiVersion": "v1" }
        """,
        """
        { "apiVersion": "'123'" }
        """
      )
    );
}

@Test
void  intentionalSingleQuotedBooleanAsString() {
    rewriteRun(
      spec -> spec.recipe(new ChangeValue("$.apiVersion", "'true'")),
      json(
        """
        { "apiVersion": "v1" }
        """,
        """
        { "apiVersion": "'true'" }
        """
      )
    );
}
```

PR #6101, opened 2025-10-02
Fix MergeYaml handling of multi-document YAML files

### LINKED ISSUE #5972: MergeYaml does not handle multi-document yaml file
I am using

- OpenRewrite v8.61.1

MergeYaml doesn't handle multi-document correctly. 
Simply test to reproduce, test failed:

```java
    @Test
    void insertAtRootMultiDocumentYaml() {
        rewriteRun(
          spec -> spec.recipe(new MergeYaml(
            "$",
            //language=yaml
            "spec: 0",
            true,
            null,
            null,
            null,
            null,
            null
          )),
          yaml(
            """
              apiVersion: policy/v1beta1
              kind: PodSecurityPolicy
              ---
              configuration: prod
              """,
            """
              apiVersion: policy/v1beta1
              kind: PodSecurityPolicy
              spec: 0
              ---
              configuration: prod
              spec: 0
              """
          )
        );
    }
```

```yaml
"apiVersion: policy/v1beta1
kind: PodSecurityPolicy
spec: 0
---
configuration: prod
spec: 0"
```

## What did you see instead?

```yml
"apiVersion: policy/v1beta1
kind: PodSecurityPolicy

spec: 0---
configuration: prod
spec: 0"
```

Short research shows that it could be related to the auto-format in the `MergeYamlVisitor` visitor

timtebeek:
- Fixes #5972

## Problem
When merging YAML at the root (`$`) into a multi-document YAML file (documents separated by `---`), the document separator was incorrectly placed on the same line as the merged content, and extra blank lines were added.

**Before:**
```yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy

spec: 0---
configuration: prod
spec: 0
```

**Expected:**
```yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
spec: 0
---
configuration: prod
spec: 0
```

## Root Cause
1. The document end prefix (containing the newline before `---`) was treated as a comment and moved to the last merged entry
2. Autoformat then added another newline, causing a double blank line
3. The document end prefix was removed, losing the newline before `---`

## Solution
The fix involves two changes:

### In `MergeYaml.java`:
- Added a `preserveDocumentSeparator()` helper method that detects multi-document YAML
- When removing the document end prefix, preserve a newline if there's a following explicit document (with `---`)

### In `MergeYamlVisitor.java`:
- Added `isMultiDocumentSeparator()` helper method to detect multi-document context  
- Modified the comment handling logic to NOT treat document end prefix as a comment when it's actually a document separator

## Testing
- Added test case `insertAtRootMultiDocumentYaml()` that reproduces the issue from #5972
- All existing tests continue to pass

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -195,17 +195,14 @@ public class MergeYaml extends Recipe {
                 Yaml.Documents documents = getCursor().firstEnclosing(Yaml.Documents.class);
                 if (documents != null) {
                     int currentIndex = documents.getDocuments().indexOf(document);
-                    if (currentIndex >= 0 && currentIndex < documents.getDocuments().size() - 1) {
-                        Yaml.Document nextDoc = documents.getDocuments().get(currentIndex + 1);
-                        if (nextDoc.isExplicit()) {
-                            // Preserve a newline before the document separator
-                            return "\n";
-                        }
+                    // Preserve a newline before the next document separator
+                    if (0 <= currentIndex && currentIndex < documents.getDocuments().size() - 1 &&
+                            documents.getDocuments().get(currentIndex + 1).isExplicit()) {
+                        return "\n";
                     }
-                    if (currentIndex == documents.getDocuments().size() - 1) {
-                        if (document.getEnd().isExplicit()) {
-                            return "\n";
-                        }
+                    // Or if this is the last document and it is explicit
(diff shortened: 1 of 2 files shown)


steve-aom-elliott:
Still needs a bit of a fix for `...` indicating end of documents. Ends up inlining the final `...` right now. Going to see if I can just make that adjustment quickly, unless Claude has an idea for that.

steve-aom-elliott:
Specifically if it looks like:

```yaml
---
abc: 123
...
---
def: 456
...
```
So the last document ending with the explicit document end

timtebeek (commented):
Thanks for spotting that `...` handling and the fix there.

PR #6103, opened 2025-10-03
chore: add `.vscode/` to `.gitignore`

qxo:
### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ out/
 .project
 .classpath
 .settings/
+.vscode/
 .DS_Store
 **/node_modules/
 IDE.properties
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -16,3 +16,4 @@ IDE.properties
 .aider*
 /worktrees
 rewrite-gradle-tooling-model/model/src/main/resources/test-manifest.txt
+[.]*
```
greg-at-moderne: That's too broad. It would filter out all files starting with `.`, right?
We don't want that.

qxo: Or we just ignore .vscode ?

PR #6104, opened 2025-10-03
Use internal method to get credentials to avoid accidentally setting an invalid state

shanman190:
## What's changed?
Use internal `getConfiguredCredentials()` to avoid eager initialization of an empty credentials object in Gradle.

## What's your motivation?
Internal to `getCredentials()` and `getCredentials(Class<?>)`, it would create an instance of `PasswordCredentials` with null values. As a result of `mavenLocal()` being a `MavenArtifactRepository` and deeply internal checks, credentials existing for a repository with type `file` is invalid and we were incidentally triggering credentials to be populated thus causing errors later on that were not observed at first.

N/A

N/A

N/A

## Any additional context
- Related to https://github.com/openrewrite/rewrite/pull/6096

### Checklist

### DIFF
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
@@ -116,15 +116,18 @@ public final class GradleProjectBuilder {
 
     @SuppressWarnings("unchecked")
     private static @Nullable Credentials getCredentials(AuthenticationSupportedInternal authenticatedRepo) {
-        Object configuredCredentials = authenticatedRepo.getConfiguredCredentials();
-        if (configuredCredentials instanceof Property) {
-            // Gradle 6.6+ returns a Property<Credentials> instance
-            return ((Property<Credentials>) configuredCredentials).getOrNull();
-        }
-        if (configuredCredentials instanceof Credentials) {
-            // Gradle < 6.6 returns a Credentials instance
-            return (Credentials) configuredCredentials;
-        }
+        try {
+            Method getConfiguredCredentialsMethod = AuthenticationSupportedInternal.class.getDeclaredMethod("getConfiguredCredentials");
+            Object configuredCredentials = getConfiguredCredentialsMethod.invoke(authenticatedRepo);
+            if (configuredCredentials instanceof Property) {
+                // Gradle 6.6+ returns a Property<Credentials> instance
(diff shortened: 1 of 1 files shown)


rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
```diff
@@ -98,16 +102,32 @@ static List<MavenRepository> mapRepositories(List<ArtifactRepository> repositori
...
+        Object configuredCredentials = authenticatedRepo.getConfiguredCredentials();
+        if (configuredCredentials instanceof Property) {
+            // Gradle 6.6+ returns a Property<Credentials> instance
+            return ((Property<Credentials>) configuredCredentials).getOrNull();
+        } else if (configuredCredentials instanceof Credentials) {
+            // Gradle < 6.6 returns a Credentials instance
+            return (Credentials) configuredCredentials;
+        }
+        return null;
+    }
```
shanman190, line 132: We can inline this if we'd prefer, it just seemed like something simple to externalize and keep the comments closely related/clean.

timtebeek (approved):

rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
```diff
@@ -98,16 +102,32 @@ static List<MavenRepository> mapRepositories(List<ArtifactRepository> repositori
...
         return builder;
     }
 
+    @SuppressWarnings("unchecked")
+    private static @Nullable Credentials getCredentials(AuthenticationSupportedInternal authenticatedRepo) {
+        Object configuredCredentials = authenticatedRepo.getConfiguredCredentials();
+        if (configuredCredentials instanceof Property) {
+            // Gradle 6.6+ returns a Property<Credentials> instance
+            return ((Property<Credentials>) configuredCredentials).getOrNull();
+        } else if (configuredCredentials instanceof Credentials) {
```
timtebeek:
```suggestion
        }
        if (configuredCredentials instanceof Credentials) {
```

rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
timtebeek: Fine indeed as it is; I like the early returns that would otherwise have to be assignments to the same variable.

PR #6105, opened 2025-10-03
Fixed situation where a changing a managed dependency that does not have an explicit `version` tag would fail to add a `version` tag when providing the `newVersion` option.

steve-aom-elliott:
## What's changed?
- Previously, when you had a POM with a _managed_ dependency that did not have an explicit `version` tag (because the version was applied to the managed dependency in an external parent artifact), calling `ChangeManagedDependencyGroupIdAndArtifactId` would fail to make a change. Now it will add the `version` tag to the managed dependency.

## What's your motivation?
This was preventing managed dependencies from being updated correctly.

## Still TODO
- A situation that I haven't managed to get correct yet is when you have a _managed_ dependency in a POM (let's call it project `A`) w/o a `version` tag, and a _direct_ dependency in a *child* module (let's call it project `B`)'s POM also w/o a `version` tag. In this situation, what appears to be happening right now is that the POM for project `A` is getting the version number added correctly to the _managed_ dependency, but at the same time, the POM for project `B` is getting a version number added incorrectly to the _direct_ dependency. `RemoveRedundantDependencyVersions` in theory should have been able to handle this, but I believe it to not be detecting the change that occurred in project `A`'s POM due to not having re-evaluated the `MavenResolutionResult` again after the change.
  - Might need some help figuring this one out.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
@@ -26,6 +26,7 @@ import org.openrewrite.maven.tree.ResolvedManagedDependency;
 import org.openrewrite.maven.tree.ResolvedPom;
 import org.openrewrite.semver.Semver;
 import org.openrewrite.semver.VersionComparator;
+import org.openrewrite.xml.AddToTagVisitor;
 import org.openrewrite.xml.RemoveContentVisitor;
 import org.openrewrite.xml.tree.Xml;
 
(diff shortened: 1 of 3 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -248,6 +262,23 @@ private boolean isDependencyManaged(Scope scope, String groupId, String artifact
                 return false;
             }
 
+            private boolean canAffectManagedDependency(MavenResolutionResult result, Scope scope, String groupId, String artifactId) {
```
steve-aom-elliott, line 265: I think one potential situation I'm unsure how we'd prefer to handle is if your parent POM had a managed dependency `A` and your child POM has a managed dependency on a BOM import that also has a managed dependency on `A`. Would we in that case actually want to indicate that we can't affect the managed dependency? Also, are we adequately able to detect this scenario in general?

sambsnyd: I think you're saying if we're importing a bom or parent from outside the repository which manages a version, well it's not part of the LST/repo so we can't edit it. Leaving a situation where the recipe makes no changes and if the build requires there to be a managed version for the new coordinates it will fail.

But this recipe could still add a _new_ dependency management entry to the current pom with the updated coordinates, though. Then, even if the parent or bom is out of reach, the new dependency coordinates get a managed version.

PR #6106, opened 2025-10-03
Added method wrapping style

Jenson3210:
## Summary
This PR adds support for IntelliJ's method chaining wrapping style configuration, allowing AutoFormat to wrap method chains according to specified style rules.

Screenshot 2025-10-02 at 23 24 09

### New Features
- **`AlwaysWrapMethodChains` visitor**: New visitor that wraps chained method calls based on configured method names
  - Identifies chain starter methods from a configurable list
  - Adds line breaks after chain starter methods to wrap subsequent calls (indentation will be added by another visitor that runs later in the formatting process)

- **Enhanced `WrappingAndBracesStyle`**: Added `ChainedMethodCalls` configuration
  - `wrap`: Line wrap setting for chained method calls  
  - `builderMethods`: List of method names that start a chain (e.g., builder methods)

- **Enhanced `AutoFormatVisitor`**: 
  - Now accepts `NamedStyles` varargs to run AutoFormat with specific style configuration
  - Prevents redundant formatting by checking if file was already formatted with the same styles using `AutoFormatted` marker
  - Integrates `AlwaysWrapMethodChains` into the formatting pipeline via `WrappingAndBracesVisitor`

- **Updated IntelliJ defaults**: Added default configuration for chained method calls wrapping


rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
```diff
@@ -97,32 +117,59 @@ public J visit(@Nullable Tree tree, P p) {
...
+    @AllArgsConstructor
+    private static class AutoFormatted<P> implements Marker {
+        UUID id;
+
+        List<NamedStyles> styles;
+
+        public AutoFormatted(List<NamedStyles> styles) {
+            this.id = randomUUID();
+            this.styles = styles;
+        }
```
sambsnyd: Using markers as shorthand for a completed formatting operation should be avoided. 
It indicates a lack of stability in the composed visitors.
If we've correctly separated concerns between the formatting visitors they should not stomp on eachother's changes. If we haven't then, even with a marker like this, we'll get different results depending on the order the visitors are run in.

A marker like this could also make for bugs where a recipe changes an LST element, invokes the formatter, then a subsequent change is made and the formatter is invoked again, but it makes no changes the second time because the marker erroneously signals to it that no formatting is required.

rewrite-java/src/test/java/org/openrewrite/java/format/AlwaysWrapMethodChainsTest.java
```diff
@@ -0,0 +1,2104 @@
...
+              public class Test {
+                  {
+                      int n = 0;
+                      n++;
+                  }
+              }
+              """
+          )
+        );
+    }
```
sambsnyd: Many of these tests seem to be about aspects of formatting other than method chaining?

sambsnyd (changes requested):

rewrite-java/src/test/java/org/openrewrite/java/format/AlwaysWrapMethodChainsTest.java
Jenson3210: Oh boy... I restored the wrong class from IntelliJ local history probably. Will update tomorrow!

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
Jenson3210: That's indeed why I only take markers into account if the styles input variable was not empty. The formatting runs from recipes would have that as empty is what i thought. 

I will seee what I can do to fix these other visitors!

rewrite-java/src/test/java/org/openrewrite/java/format/AlwaysWrapMethodChainsTest.java
Jenson3210: edit: I see what happened. 
Kept only the relevant classes for this purpose of this visitor here.

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
Jenson3210: Fixed the other visitors to be only adding newlines etc if not already containing them to not have cyclic changes. Turned out to be less impact then expected :)

Jenson3210:
@timtebeek Marking as ready for review as I have implemented the change requested. If there was another reason for putting it as Draft, feel free to put it back to "Draft" 😉

rewrite-java/src/main/java/org/openrewrite/java/format/WrappingAndBracesVisitor.java
```diff
@@ -136,6 +132,14 @@ public J.If.Else visitElse(J.If.Else else_, P p) {
         return e;
     }
 
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, P p) {
+        if (style.getChainedMethodCalls().getWrap() == WrapAlways) {
+            method = new AlwaysWrapMethodChains<>(style.getChainedMethodCalls().getBuilderMethods()).visitMethodInvocation(method, p);
+        }
```
sambsnyd: When appropriate use of a component should always be gated behind a particular condition, that condition should be part of the component.
This is a good design principle for a few reasons:

- The condition exists in one place, inside the component, rather than many places at each call-site
- It's more difficult to accidentally misuse the component omitting the guard
- We will no longer need to use "Always" in the name to allude to the danger of misuse.
 
We should move the check inside the component and give it the shorter, better name `WrapMethodChains`

Jenson3210: Thanks, this indeed is an improvement when I look at it 👍

PR #6108, opened 2025-10-05
Drop sparsely used commons-lang3

timtebeek:
Turns out we were mostly using it for the wrong reasons; other existing classes work just as well, or the more targeted commons-text. That last one we also only have for a single use case, but it's a nice bit to keep. If we want to further reduce the dependencies then we can inline that one algorithm.

Prompted by seeing a warning about: https://nvd.nist.gov/vuln/detail/CVE-2025-48924
Note that the fix has already been out in [3.18.0 since Jul 09, 2025](https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.18.0), so we're not vulnerable.
Just thought it better to remove the dependency when it's not needed at all.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/Find.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/Find.java
@@ -17,7 +17,6 @@ package org.openrewrite.text;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.apache.commons.lang3.BooleanUtils;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.binary.Binary;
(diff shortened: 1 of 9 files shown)


rewrite-core/src/main/java/org/openrewrite/text/Find.java
```diff
@@ -156,7 +155,7 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                     int matchStart = matcher.start();
                     snippets.add(snippet(rawText.substring(previousEnd, matchStart)));
                     String text = rawText.substring(matchStart, matcher.end());
-                    snippets.add(SearchResult.found(snippet(text), BooleanUtils.isTrue(description) ? text : null));
+                    snippets.add(SearchResult.found(snippet(text), Boolean.TRUE.equals(description) ? text : null));
```
greg-at-moderne, line 158: An idea for recipe maybe?

greg-at-moderne (approved):

rewrite-core/src/main/java/org/openrewrite/text/Find.java
timtebeek: Yep; although I'd hope this one is rare. I already added this one as well
- https://github.com/openrewrite/rewrite-apache/pull/103

PR #6110, opened 2025-10-06
Add Kotlin settings file support to AddSettingsPluginRepository

Jenson3210:
## Summary

Extends the `AddSettingsPluginRepository` recipe to support Kotlin DSL settings files (`settings.gradle.kts`) in addition to the existing Groovy settings file (`settings.gradle`) support.

### Core Recipe Updates
- **Refactored visitor hierarchy**: Changed from `GroovyIsoVisitor` to `JavaIsoVisitor` to handle both Groovy and Kotlin ASTs
- **Added K.CompilationUnit support**: Implemented separate visitor methods for both `G.CompilationUnit` (Groovy) and `K.CompilationUnit` (Kotlin)
- **Improved code generation**: Updated template code generation to produce syntax appropriate for each language (e.g., `url = "..."` for Groovy vs `url = uri("...")` for Kotlin)

### Repository Search Improvements
- **Enhanced FindRepository**: Updated to work with both Groovy and Kotlin file types
- **Kotlin string template handling**: Added `TemplateAsString` helper class to properly match URLs in Kotlin string templates (`K.StringTemplate`) alongside existing Groovy GString support (removing the `print` need)

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddDevelocityGradlePlugin.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddDevelocityGradlePlugin.java
@@ -25,13 +25,14 @@ import org.openrewrite.gradle.IsBuildGradle;
 import org.openrewrite.gradle.IsSettingsGradle;
 import org.openrewrite.gradle.marker.GradleProject;
 import org.openrewrite.gradle.marker.GradleSettings;
-import org.openrewrite.groovy.GroovyIsoVisitor;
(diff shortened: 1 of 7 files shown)


sambsnyd (approved):

shanman190:
Just a note, but the `uri` form is valid for both Kotlin and Groovy DSLs. 

https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#uri(java.lang.Object)

Jenson3210:
Thanks @shanman190 for the insight. In this case however, there are too many other differences in the template that it makes more sense to me to keep them separated methods.

PR #6112, opened 2025-10-08
ChopIfTooLong for methodInvocation chains

Jenson3210:
Also support ChopIfTooLong for the recently added WrapMethodChains formatter.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/style/StyleHelper.java
+++ b/rewrite-core/src/main/java/org/openrewrite/style/StyleHelper.java
@@ -15,13 +15,19 @@
  */
 package org.openrewrite.style;
 
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.SourceFile;
 import org.openrewrite.internal.StringUtils;
+import org.openrewrite.marker.Markers;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
+import java.util.function.Supplier;
 
 public class StyleHelper {
 
@@ -75,7 +81,9 @@ public class StyleHelper {
                 if (rightValue != null) {
                     if (!isPrimitiveOrWrapper(rightValue) && !isEnum(rightValue)) {
                         Object leftValue = getter.invoke(left);
-                        rightValue = merge(leftValue, rightValue);
+                        if (leftValue instanceof Collection && !((Collection<?>) leftValue).isEmpty()) {
+                            rightValue = merge(leftValue, rightValue);
+                        }
                     }
                     //noinspection unchecked
                     left = (T) wither.invoke(left, rightValue);
@@ -85,5 +93,46 @@ public class StyleHelper {
             }
         }
         return left;
+    }
+
+    public static <T extends SourceFile> T addStyleMarker(T t, List<NamedStyles> styles) {
+        if (!styles.isEmpty()) {
+            Set<NamedStyles> newNamedStyles = new HashSet<>(styles);
+            boolean styleAlreadyPresent = false;
+            for (NamedStyles namedStyle : t.getMarkers().findAll(NamedStyles.class)) {
+                styleAlreadyPresent = !newNamedStyles.add(namedStyle) || styleAlreadyPresent;
+            }
+            // As the order or NamedStyles matters, we cannot simply use addIfAbsent.
+            if (!styleAlreadyPresent) {
+                Markers markers = t.getMarkers().removeByType(NamedStyles.class);
+                for (NamedStyles namedStyle : newNamedStyles) {
+                    markers = markers.add(namedStyle);
+                }
+
+                return t.withMarkers(markers);
+            }
+        }
+        return t;
+    }
+
+    public static <S extends Style, T extends SourceFile> @Nullable S getStyle(Class<S> styleClass, List<NamedStyles> styles, T sourceFile) {
+        S projectStyle = Style.from(styleClass, sourceFile);
+        S style = NamedStyles.merge(styleClass, styles);
+        if (projectStyle == null) {
+            return style;
+        }
+        if (style != null) {
+            return StyleHelper.merge(projectStyle, style);
+        }
+        return projectStyle;
+    }
+
+    public static <S extends Style, T extends SourceFile> S getStyle(Class<S> styleClass, List<NamedStyles> styles, T sourceFile, Supplier<S> defaultStyle) {
+        S projectStyle = Style.from(styleClass, sourceFile, defaultStyle);
+        S style = NamedStyles.merge(styleClass, styles);
+        if (style != null) {
+            return StyleHelper.merge(projectStyle, style);
+        }
+        return projectStyle;
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/text/Find.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/Find.java
@@ -20,6 +20,7 @@ import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.binary.Binary;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.marker.Markers;
 import org.openrewrite.marker.SearchResult;
 import org.openrewrite.quark.Quark;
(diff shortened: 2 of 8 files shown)


rewrite-java/src/main/java/org/openrewrite/java/format/WrapMethodChains.java
```diff
@@ -73,6 +79,21 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, P ctx
...
+        Object cursorValue = cursor.getValue();
+        if (cursorValue instanceof J) {
+            J j = (J) cursorValue;
+            boolean hasNewLine = j.getPrefix().getWhitespace().contains("\n") || j.getComments().stream().anyMatch(c -> c.getSuffix().contains("\n"));
+            Cursor parent = cursor.getParentTreeCursor();
+            boolean isRoot = Cursor.ROOT_VALUE == parent.getValue();
+            if (!hasNewLine && !isRoot) {
+                return calculateLineLength(parent);
+            }
+            return j.print(cursor).replaceAll("^.*\\n", "").length() + 1;
```
Jenson3210: - adding 1 to count the semicolon
- do we know of a better way to do this that does not require `print`?

rewrite-java/src/main/java/org/openrewrite/java/format/LengthCalculator.java
```diff
@@ -0,0 +1,68 @@
...
+            J j = (J) cursorValue;
+            boolean hasNewLine = j.getPrefix().getWhitespace().contains("\n") || j.getComments().stream().anyMatch(c -> c.getSuffix().contains("\n"));
+            Cursor parent = cursor.getParentTreeCursor();
+            boolean isCompilationUnit = parent.getValue() instanceof J.CompilationUnit;
+            if (!hasNewLine && !isCompilationUnit) {
+                return computeTreeLineLength(parent.getValue(), parent);
+            }
+        } else {
+            throw new RuntimeException("Unable to calculate length due to unexpected cursor value: " + cursorValue.getClass());
+        }
```
sambsnyd: We have a service mechanism for supplying language-specific implementations of concepts that appear in multiple languages. This strikes me as a good candidate for that pattern. See [JavaSourceFile.service()](https://github.com/openrewrite/rewrite/blob/f0b1d51c5efe74958a16c7f2ccf3a64e230cb43f/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaSourceFile.java#L68). 
This is how a `JavaVisitor` can visit a kotlin source file and have `maybeAddImport()` use the kotlin version of `AddImport` to do it. 

I see this as the default `SourcePositionService`.  While you don't need to do it right now, it could be a reasonable place to put a `getLineNumber()` or `getStartColumn()`/`getEndColumn()` methods which are sometimes requested. Methods like these could be relevant in any language or data format.

There could be Groovy/Kotlin variations which, for example, return `false` more often in `needsSemicolon` as those languages only require them to separate multiple statements on the same line.

rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
```diff
@@ -97,6 +97,14 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, P
         return c;
     }
 
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, P p) {
+        if (method.getName().getComments().isEmpty() && method.getName().getPrefix().getWhitespace().contains("\n")) {
+            return method.withName(method.getName().withPrefix(Space.EMPTY));
+        }
+        return super.visitMethodInvocation(method, p);
```
sambsnyd, line 105: A method invocation meeting this condition could be contained within another method invocation which meets this condition. In that circumstance the inner method would not be reformatted as no traversal into sub-elements happens when this condition is matched.

rewrite-java/src/main/java/org/openrewrite/java/format/LengthCalculator.java
Jenson3210: I took this approach from `ColumnPositionCalculator` already present. Is that then also a good candidate? 

This method though does not `getLineNumber` but gets the length of the current element (including its newlines) starting from the first element that starts on it's line. 

Counting the prefix  (indentation only) of that element + the length. I struggled to find an explanatory name for that. Any suggestions?

rewrite-java/src/main/java/org/openrewrite/java/format/MinimumViableSpacingVisitor.java
Jenson3210: good catch, thanks!

rewrite-java/src/test/java/org/openrewrite/java/format/WrapMethodChainsTest.java
```diff
@@ -1129,4 +1141,71 @@ MyObject process(List<String> items) {
...
+              """,
+            """
+              package com.example;
+              
+              class Test {
+                  void test() {
+                      MyObject obj = MyObject.builder()
+              .name("test")
+              .age(25)
+              .build();
```
timtebeek, line 1176: Somewhat surprised by the left aligned new lines here; is the expectation that a formatter is run afterwards?

Jenson3210: Wrapping decides that the calls end up on the new line. The `TabsAndIndentsVisitor` which runs after the `WrappingAndBracesVisitor` (which runs this WrapMethodChains) in the `AutoFormatVisitor` make sure that indentation will afterwards be correct. 
You wil find almost exactly the same tests in there, but with correct indentation.

rewrite-java/src/main/java/org/openrewrite/java/format/LengthCalculator.java
Jenson3210: @sambsnyd, I refactored this one to be what I think you meant. Could you please have another look at it?

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -0,0 +1,70 @@
...
+import org.openrewrite.java.tree.Statement;
+
+import static java.util.Collections.emptyList;
+
+@Incubating(since = "8.63.0")
+public class SourcePositionService {
+
+    public int computeTreeLength(J tree, Cursor cursor) {
+        Object cursorValue = cursor.getValue();
+        if (cursorValue instanceof J) {
```
sambsnyd: Why does this accept both a `J` and a `Cursor` ? Shouldn't the `Cursor` point to the `J` making it redundant to require the `J` as a parameter?

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -0,0 +1,70 @@
...
+    private int getSuffixLength(J tree) {
+        if (tree instanceof Statement && needsSemicolon((Statement) tree)) {
+            return 1;
+        }
+        return 0;
+    }
+
+    private boolean needsSemicolon(Statement statement) {
+        return statement instanceof J.MethodInvocation || statement instanceof J.VariableDeclarations || statement instanceof J.Assignment || statement instanceof J.Package;
+    }
```
sambsnyd, line 68: What about `Return`, `Import`, `Assert`?

What about `J.MethodInvocation` which appear in an expression context, like when they are in the argument list of another method invocation? 
```java
// No semicolon allowed on the inner bar()
foo(bar());
```

Jenson3210: As we are counting line length, we look at the outer method invocation if needed(the inner has no new line before it) so we would either count the inner one or the outer one.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: Done!

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: The "line-"length of both `foo(bar())` and `bar()` in that example is the same (prefix of foo + length(`foo(bar())`) + 1 semicolon)

PR #6114, opened 2025-10-09
Don't use `parentPomIsProjectPom()` to determine same repo for parent…

### LINKED ISSUE #6113: `UpdateMavenProjectPropertyJavaVersion` not working with parent POM via Maven plugin
8.63 snapshot but this is a very old bug that comes starting `rewrite-migrate-java` 2.30.x

To reproduce clone pet clinic from https://github.com/BoykoAlex/spring-petclinic and switch to `2.7.x` branch. Then use the following CLI command:
```
./mvnw org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:3.18.0 -Drewrite.activeRecipes=org.openrewrite.java.migrate.UpgradeBuildToJava11 -Drewrite.failOnInvalidActiveRecipes=true -Drewrite.recipeChangeLogLevel=INFO
```

Expected result is `<java.version>1.8<java.version>` -> `<java.version>11<java.version>`

This doesn't happen. The culprit is `UpdateMavenProjectPropertyJavaVersion` namely this call:
`getResolutionResult().parentPomIsProjectPom()`

The map of project poms in the case of the project parsed via maven rewrite plugin has entry for `spring-boot-dependencies` and `spring-boot-parent` coming from the local `./m2` repo via a path such as `../../../.m2/repository/...`. The `parentPomIsProjectPom()` method isn't really checking if parent lives in the same repo. The junit won't show this issues since parent is always null due to this line:
https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/MavenParser.java#L115 and no modules.

Seems that with maven repo MavenProject with poms come for all resolved poms found including the parents.

I'm not sure how to set the parent properly yet with the unit test for the petclinic example...
Nevertheless, I think the fix for this issue is simple: use scanning recipes to keep a set of parsed maven pom GAVs as this would be all poms in the current repo

BoykoAlex:
- Fixes #6113

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -17,7 +17,10 @@ package org.openrewrite.maven;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.openrewrite.*;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.TreeVisitor;
 import org.openrewrite.maven.tree.Plugin;
 import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
 import org.openrewrite.xml.XPathMatcher;


timtebeek:
- Fixed in https://github.com/openrewrite/rewrite/pull/6119

PR #6115, opened 2025-10-09
Attempt to fix type information when unfolding wildcard imports

Laurens-W:
## What's changed?

Wildcard imports receive type Unknown however after we unfold them they should be specific imports that have type information

## What's your motivation?

When a `org.junit.*` import is unfolded during the junit 4 to 5 migration and attempt is made to move some imports to their respective JUnit 5 variant ambiguity occurs because the type information is invalid

### Checklist


Laurens-W:
> That looks sensible but the "attempt" in the title and no test showing it works doesn't inspire great confidence

I did write a test to validate my hypothesis, but the limited setup I have succeeds regardless of this change, I'll work on adding a testcase that proves the point and remove Attempt from the title :)

MBoegers:
Is there a reason why the recipe does 3 things in one visitor run? I understand that we don't want 3 recipes, as none of them is reasonable on its own, but we could create 3 visitors and chain them together in one recipe.
The upside here would be that we can write tests per visitor and probably replace old implementations with now existing atomic recipes.

PR #6124, opened 2025-10-10
Add TOML recipes to change keys and values, find and delete keys, and create new TOML files

zieka:
## Solution

Added foundational TOML refactoring recipes:

- **ChangeKey** - Rename TOML keys while preserving values
- **ChangeValue** - Change TOML values with automatic type detection
- **DeleteKey** - Remove key-value pairs from documents and tables
- **FindKey** - Search for keys using path expressions with wildcard support
- **CreateTomlFile** - Create new TOML files with overwrite option
- **TomlPathMatcher** - Path matching utility supporting wildcards and dotted notation

These recipes provide base refactoring capabilities for pyproject.toml and Cargo.toml files.

### DIFF
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/TomlPathMatcher.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/TomlPathMatcher.java
@@ -21,7 +21,6 @@ import org.openrewrite.toml.tree.Toml;
 import org.openrewrite.toml.tree.TomlKey;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.regex.Pattern;
 


zieka:
the error in the CI `[receive-pr / upload-patch / upload-patch (pull_request)` looks unrelated to this pr:

- `UpdateMavenProjectPropertyJavaVersion.java`

rewrite-toml/src/main/java/org/openrewrite/toml/DeleteKey.java
```diff
@@ -0,0 +1,82 @@
...
+                    if (value instanceof Toml.KeyValue) {
+                        Cursor valueCursor = new Cursor(getCursor(), value);
+                        if (matcher.matches(valueCursor)) {
+                            return null;
+                        }
+                    }
+                    return value;
+                }));
+
+                return t;
```
timtebeek:
```suggestion
                return t.withValues(ListUtils.map(t.getValues(), value -> {
                    if (value instanceof Toml.KeyValue) {
                        Cursor valueCursor = new Cursor(getCursor(), value);
                        if (matcher.matches(valueCursor)) {
                            return null;
                        }
                    }
                    return value;
                }));
```

Jenson3210:
Should we similarly also add a `DeleteTable` which removes an entire table?

timtebeek:
Yes makes sense; could you ask Claude? :)

PR #6129, opened 2025-10-13
Prepare chopping and wrapping parameters in methodDeclarations.

Jenson3210:
Previously we used to calculate the indentation of the later arguments based on the first arguments indentation. 
If that elements was wrapped/chopped, its indentation is 0 in length. 

If the indentation of the argument would print the element in a column that is before the indentation of the methodDeclaration, this code change will use the methodDeclarations block to calculate the indentation by adding the continuationSpacing to the declaration prefix length.


rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -397,7 +397,13 @@ private int computeFirstParameterColumn(J.MethodDeclaration method) {
         }
 
         if (firstArg.getPrefix().getLastWhitespace().contains("\n")) {
-            return getLengthOfWhitespace(firstArg.getPrefix().getLastWhitespace());
+            int declPrefixLength = getLengthOfWhitespace(method.getPrefix().getLastWhitespace());
+            int argPrefixLength = getLengthOfWhitespace(firstArg.getPrefix().getLastWhitespace());
+            //noinspection ConstantConditions to be backwards compatible with older style versions
+            if (style != null && style.getContinuationIndent() != null && declPrefixLength >= argPrefixLength) {
```
Jenson3210: I did not introduce these properties so they should exist already. Perhaps sufficient to do 
```suggestion
            if (declPrefixLength >= argPrefixLength) {
```

or `if (style.getContinuationIndent() != null && declPrefixLength >= argPrefixLength) {`

sambsnyd: The `style` field is not something new you've introduced and is not nullable so there should be no need for null check

sambsnyd (approved):

PR #6130, opened 2025-10-13
Render Gradle 9.1.0 wrapper templates without classpath conditional

timtebeek:
- Based on what we saw in https://github.com/openrewrite/rewrite-spring-to-quarkus/pull/54

### DIFF
--- a/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
+++ b/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
@@ -1,8 +1,10 @@
 version,gradlew,gradlewBat
+9.3.0-20251012005045+0000,59e5bb1c,c518efba
 9.2.0-rc-1,59e5bb1c,c518efba
 9.2.0-milestone-2,59e5bb1c,c518efba
 9.2.0-milestone-1,59e5bb1c,c518efba
-9.2.0-20251011014600+0000,59e5bb1c,c518efba
+9.2.0-20251013021226+0000,59e5bb1c,c518efba
+9.1.0,59e5bb1c,c518efba
 9.1.0-rc-4,59e5bb1c,c518efba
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
```diff
@@ -1092,6 +1098,23 @@ void usesExecutableJarFrom8_14() {
...
+        Collection<SourceFile> generate = new UpdateGradleWrapper(version, "bin", true, null, null)
+          .generate(state, new InMemoryExecutionContext());
+        assertThat(generate)
+          .asInstanceOf(InstanceOfAssertFactories.LIST)
+          .filteredOn(PlainText.class::isInstance)
+          .map(PlainText.class::cast)
+          .allSatisfy(plainText -> assertThat(plainText.getText())
+            .contains("-classpath")
+            .doesNotContain("<%", "%>"));
+    }
```
timtebeek, line 1116: Figured this would be relatively straightforward way to test a few templates and assert some properties of each/none of those. Open to other options.

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
shanman190, line 1: Might run the downloader a couple more times until there aren't any failures encountered (probably need to add retries at some point), but the note here is that 9.1.0 release is still missing from the list.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
shanman190: Yeah, this should be enough to get started. My longer term thought was to perform a rendering validation between our `renderTemplate` and Groovy `SimpleTemplateEngine` which will be able to catch these for all versions. But that can come later.

PR #6137, opened 2025-10-14
Prerender Gradle wrapper script templates

shanman190:
## What's changed?
Add validation against the upstream behavior and our own internal rendering logic to ensure that all templates render out fully and correctly.

## What's your motivation?
- Improvement upon https://github.com/openrewrite/rewrite/pull/6130

N/A

@timtebeek 

N/A

## Any additional context
N/A

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/UpdateGradleWrapperTest.java
@@ -1098,23 +1098,6 @@ class UpdateGradleWrapperTest implements RewriteTest {
         );
     }
 
-    @ParameterizedTest
-    @ValueSource(strings = {"7.6", "8.14.3", "9.1.0"})
-    void renderUnixScript(String version) {
-        UpdateGradleWrapper.GradleWrapperState state = new UpdateGradleWrapper.GradleWrapperState();
-        state.setGradleProject(true);
-        state.setAddGradleShellScript(true);
-        Collection<SourceFile> generate = new UpdateGradleWrapper(version, "bin", true, null, null)
-          .generate(state, new InMemoryExecutionContext());
-        assertThat(generate)
-          .asInstanceOf(InstanceOfAssertFactories.LIST)
-          .filteredOn(PlainText.class::isInstance)
-          .map(PlainText.class::cast)
-          .allSatisfy(plainText -> assertThat(plainText.getText())
-            .contains("-classpath")
-            .doesNotContain("<%", "%>"));
-    }
-
     private <S extends SourceFile> S result(RecipeRun run, Class<S> clazz, String endsWith) {
         return run.getChangeset().getAllResults().stream()
           .map(Result::getAfter)


rewrite-gradle/build.gradle.kts
```diff
@@ -69,8 +69,8 @@ dependencies {
     testImplementation(project(":rewrite-gradle-tooling-model:model"))
     "pluginLocalTestClasspath"(project(mapOf("path" to ":rewrite-gradle-tooling-model:model", "configuration" to "pluginLocalTestClasspath")))
     testImplementation("com.squareup.okhttp3:mockwebserver:4.+")
+    testImplementation(localGroovy())
```
shanman190, line 72: When running `SimpleTemplateEngine`, it was complaining that Gradle's internal Groovy 3.0.24 and our Groovy 3.0.25 were in conflict.

timtebeek (approved):
Nice!

.github/workflows/gradle-wrapper-update.yml
```diff
@@ -0,0 +1,37 @@
+---
+name: Update Gradle wrapper scripts
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: 0 10 * * MON
```
shanman190: @timtebeek, any opinions on when and how often to run this workflow?

rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/GradleWrapperScriptGenerator.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.util.Objects.requireNonNull;
+
+public class GradleWrapperScriptGenerator {
+    private GradleWrapperScriptGenerator() {}
+
+    public static Map<String, String> unixBindings(GradleWrapper gradleWrapper) {
```
timtebeek: We only seem to be using `GradleWrapper.getVersion()`; do we foresee a need to pass in the whole object or can we reduce that down to just the version?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
```diff
@@ -402,77 +403,13 @@ private static <T extends SourceFile> T setExecutable(T sourceFile) {
...
-            binding.put("mainClassName", "");
-        } else {
-            binding.put("classpath", "$APP_HOME/gradle/wrapper/gradle-wrapper.jar");
-            binding.put("entryPointArgs", "");
-            binding.put("mainClassName", "org.gradle.wrapper.GradleWrapperMain");
-        }
-
         String gradlewTemplate = StringUtils.readFully(gradleWrapper.gradlew().getInputStream(ctx));
-        return renderTemplate(gradlewTemplate, binding, "\n");
+        return renderTemplate(gradlewTemplate, unixBindings(gradleWrapper), "\n");
```
timtebeek: One thing I'm wondering is whether we should even be rendering these at runtime, or whether we can prerender as we download, to minimize the logic here. Presumably those files will compress quite well, which even if we double what we store should then still not be too much in the grand scheme of things. What was the reason you kept rendering at runtime? Do you foresee additional options to make `classpath` optional for instance?

.github/workflows/gradle-wrapper-update.yml
timtebeek: We tend to release on wednesdays, so I'd say `TUE` would mean we're usually up to date by the time we release.

rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/GradleWrapperScriptGenerator.java
shanman190: I don't think that we'll need any more of it at the moment, so we can minimize it. It's also an internal class, so if we need to adjust it later it's less of a problem.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpdateGradleWrapper.java
shanman190: That's true. If we have all versions, then every template can be fully materialized ahead of time. Originally-originally, we downloaded the distributions and did everything at runtime. Then there was a change to cache the scripts in rewrite-gradle to which I think the rendering at runtime remained to account for gaps in the scripts that we have internally cataloged, but now those gaps are pretty well eliminated. So there's probably no reason to not prerender out the templates ahead of time.

`classpath` is already optional in the current versions, but it only switches between the wrapper jar being passed via `-classpath` vs `-jar`.

shanman190: Addressed via https://github.com/openrewrite/rewrite/pull/6137/commits/22aa9a81b491e63f86195ae57449ecf41c041a38

PR #6139, opened 2025-10-15
Gradle dependency management refactoring

sambsnyd:
Gradle dependency declarations can take many forms. To list a subset of the variations:
```groovy
dependencies {
        // Single-quoted and double quoted literal strings, with and without parentheses
    implementation 'com.google.guava:guava:29.0-jre'
    implementation "com.google.guava:guava:29.0-jre"
    compileOnly('com.google.guava:guava:29.0-jre')
    runtimeOnly("com.google.guava:guava:29.0-jre")


    // Map notation without parentheses
    implementation group: "com.google.guava", name: "guava", version: '29.0-jre'
    implementation(group: "com.google.guava", name: "guava", version: '29.0-jre')

    // interpolated
    def guavaVersion = '29.0-jre'
    implementation "com.google.guava:guava:$guavaVersion"
    implementation "com.google.guava:guava:${guavaVersion}"
    implementation group: "com.google.guava", name: "guava", version: guavaVersion
    implementation group: "com.google.guava", name: "guava", version: "$guavaVersion"
    implementation group: "com.google.guava", name: "guava", version: "${guavaVersion}"

    // Via project properties or ext properties
    implementation "com.google.guava:guava:${property('guavaVersion')}"
    implementation "com.google.guava:guava:${project.property('guavaVersion')}"n
    implementation "com.google.guava:guava:${project.findProperty('guavaVersion')}"
    implementation "com.google.guava:guava:${project.properties['guavaVersion']}"

    // Concatenation
    implementation 'com.google.guava:guava:' + guavaVersion

    // Multiple dependencies in single call (varargs, groovy only)
    implementation(
        'com.google.guava:guava-gwt:29.0-jre',
        'com.google.guava:guava:29.0-jre'
    )

    // platform() or enforcedPlatform() 
    implementation platform("com.google.guava:guava:29.0-jre")

    // version declared in platform/BOM, dependency substitution rule, etc.
    implementation("com.google.guava:guava")
}
```

The logic to handle these different syntax is scattered around the various dependency management recipes, which each of them covering incompletely-overlapping subsets. 
This PR centralizes this syntactical knowledge into the `GradleDependency` trait and the `GradleMultiDependency` trait.

`GradleDependency` should now be a one-stop-shop for reading and refactoring dependency declarations of all kinds.
`GradleMultiDependency` exists only to abstract away the groovy-specific varargs version of a dependency declaration. This is necessary 

Not every recipe has been updated to fully take advantage of this, but this PR is sprawling enough as-is. 

Other enhancements include:
- `StringUtils.matchesGlob()` no longer requires a `null` check to get the most commonly desired behavior
- Remove need for `org.openrewrite.gradle.internal.Dependency` and its parser separate from `org.openrewrite.maven.Dependency`
- Adoption of trait has slimmed a substantial number of lines from `RemoveRedundantDependencyVersion` and `UpgradeDependencyVersion`

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
@@ -35,6 +35,7 @@ import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.kotlin.tree.K;
+import org.openrewrite.marker.Markers;
 import org.openrewrite.maven.tree.Dependency;
 import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedDependency;
@@ -49,6 +50,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.*;
 import static java.util.stream.Collectors.toList;
 
 @Value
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -17,64 +17,1291 @@
...
+        }
+
+        return null;
+    }
+
+    /**
+     * Helper method to extract a string value from various expression types.
+     * Handles literals, identifiers, field access, method invocations, and GStrings.
+     */
+    private @Nullable String extractValueAsString(Expression value) {
```
shanman190, line 468: I'm a little worried about the prints contained in here for a couple of reasons:

* Sustainability/Incorrect results
* If a marker exists on one of these elements from a previous recipe, then the marker information will end up being printed here

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -93,96 +1320,83 @@ public J visitMethodInvocation(J.MethodInvocation method, P p) {
...
-                }
+            if(!isDependencyDeclaration(cursor)) {
+                return null;
+            }
+            J.MethodInvocation methodInvocation = cursor.getValue();
+            GradleProject gradleProject = getGradleProject(cursor);
+            GradleDependencyConfiguration gdc = getConfiguration(gradleProject, methodInvocation);
+            if (gdc == null && !(DEPENDENCY_DSL_MATCHER.matches(methodInvocation) && !"project".equals(methodInvocation.getSimpleName()))) {
+                return null;
+            }
```
shanman190: callout: We're already performing this same check within `isDependencyDeclaration` above.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
sambsnyd: I've added in a `withMarkers(Markers.EMPTY)` to handle the marker problem. Otherwise leaving it mostly the same for now... agree there's possibility for error in there but it's just pretty convenient to be able to get a string out of whatever format the expression is.

PR #6142, opened 2025-10-15
Added more toml recipes to handle tables

Jenson3210:
## Summary

Added TOML table row manipulation recipes to enable programmatic modification of TOML array tables, with support for matching, merging, replacing, and deleting rows.

  ## New Recipes

  - **`ChangeTableRowValue`** - Change a specific property value in a matching table row
  - **`DeleteTableRow`** - Delete table rows matching specified criteria with regex support
  - **`MergeTableRow`** - Merge key-value pairs into existing rows or insert new ones, with null value support for removal of keys
  - **`ReplaceTableRow`** - Replace entire table rows with other rows

  ## Utilities

  - **`TableRowMatcher`** - Shared utilities for matching and extracting values from table rows with regex support
  - **`SemanticallyEqual`** - Recursive equality checking for TOML elements that ignores whitespace and formatting

### DIFF
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/ReplaceTableRow.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/ReplaceTableRow.java
@@ -18,7 +18,10 @@ package org.openrewrite.toml;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.intellij.lang.annotations.Language;
-import org.openrewrite.*;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.toml.tree.Space;
 import org.openrewrite.toml.tree.Toml;
@@ -57,14 +60,6 @@ public class ReplaceTableRow extends Recipe {
     @Override
     public String getDescription() {
         return "Replace a TOML table row with new content. If a row with the same identifying property exists, replace it entirely.";
-    }
-
-    @Override
-    public Validated<Object> validate() {
-        return super.validate()
-                .and(Validated.required("tableName", tableName))
-                .and(Validated.required("row", row))
-                .and(Validated.required("identifyingKey", identifyingKey));
     }
 
     @Override
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/MergeTableRow.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/MergeTableRow.java
@@ -63,14 +63,6 @@ public class MergeTableRow extends Recipe {
     }
 
     @Override
-    public Validated<Object> validate() {
-        return super.validate()
-                .and(Validated.required("tableName", tableName))
-                .and(Validated.required("row", row))
-                .and(Validated.required("identifyingKey", identifyingKey));
-    }
-
-    @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new TomlVisitor<ExecutionContext>() {
             @Override


rewrite-toml/src/main/java/org/openrewrite/toml/DeleteTableRow.java
```diff
@@ -0,0 +1,96 @@
...
+            @Override
+            public @Nullable Toml visitTable(Toml.Table table, ExecutionContext ctx) {
+                Toml.Table t = (Toml.Table) super.visitTable(table, ctx);
+
+                if (t.getName() != null && tableName.equals(t.getName().getName())) {
+                    // Check if this table contains a matching key-value pair
+                    if (TableRowMatcher.hasMatchingKeyValue(t, identifyingKey, identifyingValue, useRegex)) {
+                        return null; // Delete the table
+                    }
+                }
```
Jenson3210:
```suggestion
                if (t.getName() != null && tableName.equals(t.getName().getName()) && TableRowMatcher.hasMatchingKeyValue(t, identifyingKey, identifyingValue, useRegex)) {
                    return null; // Delete the table
                }
```

rewrite-toml/src/main/java/org/openrewrite/toml/ChangeTableRowValue.java
```diff
@@ -0,0 +1,157 @@
...
+                }
+
+                // Check if this table row matches the identifying criteria
+                if (!TableRowMatcher.hasMatchingKeyValue(table, identifyingKey, identifyingValue, useRegex)) {
+                    return super.visitTable(table, ctx);
+                }
+
+                // Update the property value
+                List<Toml> updatedValues = new ArrayList<>(table.getValues());
+                boolean hasChanges = false;
```
timtebeek: This type of pattern makes me wonder if we could/should be using `ListUtils`. Was that something you've looked at?

Jenson3210: applied ListUtils wherever possible

rewrite-toml/src/main/java/org/openrewrite/toml/SemanticallyEqual.java
```diff
@@ -0,0 +1,226 @@
...
+        public Toml visitDocument(Toml.Document document, Toml other) {
+            if (document == other) {
+                return null;
+            }
+            if (!(other instanceof Toml.Document)) {
+                areEqual = false;
+                return null;
+            }
+            Toml.Document otherDocument = (Toml.Document) other;
+            visitList(document.getValues(), otherDocument.getValues());
```
timtebeek, line 54: I'm suprised here by the explicit calls to `visitList` rather than the more established `super.visit` which would then also call `visitValues`. What was the thinking here? I'm inclined to think using established patterns of super visits would make it easier to reason about which methods get called when.

Jenson3210: Took this concept from Java's https://github.com/openrewrite/rewrite/blob/39c864b5a328a2a26ab9c183394c9d98d4aa1183/rewrite-java/src/main/java/org/openrewrite/java/search/SemanticallyEqual.java#L333

Jenson3210: Not sure how we can do it differently when you have to compare the entire list of values. 
`super.visit(document, other)` would visit `document.getValues()`, but would use `other` without sending it's correct value to each separate compare visitor.

timtebeek: Ah yes you're right, thanks!

rewrite-toml/src/main/java/org/openrewrite/toml/ReplaceTableRow.java
```diff
@@ -0,0 +1,154 @@
...
+        return "Replace a TOML table row with new content. If a row with the same identifying property exists, replace it entirely.";
+    }
+
+    @Override
+    public Validated<Object> validate() {
+        return super.validate()
+                .and(Validated.required("tableName", tableName))
+                .and(Validated.required("row", row))
+                .and(Validated.required("identifyingKey", identifyingKey));
+    }
```
timtebeek: None of these fields are marked as nullable or optional; do we then really need this explicit validation?

timtebeek: Same applies to `MergeTableRow`

timtebeek (approved):
Nice work! Approved already, but it looks like we could remove some of those explicit validation methods still, as I think we validate that when assigning values to option annotated fields.

PR #6148, opened 2025-10-16
Map JS `BigInt` literals to string values

knutwannheden:
`BigInt` can't be serialized as JSON.


jkschneider (approved):

jkschneider:
@knutwannheden It's reasonable to map it to long, since that would otherwise be unused. In this case, perhaps we change the name of the Primitive type to `BigInt` (while preserving `long` as the keyword).

image

PR #6150, opened 2025-10-17
Recipe marketplace CSV

jkschneider:
## What's changed?

Revamping the concept of a Recipe Marketplace, eventually consolidating the functionality of `Environment` and `CategoryTree`, based on everything we've learned about marketplaces and their shape over the last several years.


doc/adr/0006-recipe-marketplace-csv-format.md
```diff
@@ -0,0 +1,142 @@
...
+## Decision
+
+We will use a CSV format for recipe marketplace data with the following structure:
+
+### Required Columns
+
+- **`name`**: The fully qualified recipe name (e.g., `org.openrewrite.java.cleanup.UnnecessaryParentheses`)
+- **`category1, category2, ..., categoryN`**: Zero or more category columns, read **left to right** with **left representing the deepest level category**
+
+### Optional Recipe Columns
```
Jenson3210, line 30: Will recipes still have tags? 
Can I add tags to recipes through (a) csv column(s)? 
This would be a requested feature for teams that want to tag certain existing recipes with custom tags. Or would they have to categorize these recipes in a specific category instead?

jkschneider: Tags still exist, and are available on the `RecipeDescriptor` after a describe operation. There is no need for them to be present in the CSV.

Jenson3210: So there is no way to add a custom tag to a recipe which is shipped by a recipe jar without creating a custom wrapper over that recipe?

PR #6156, opened 2025-10-18
UnnecessaryParentheses must not remove parentheses around chained switch statement

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/534


knutwannheden:
Maybe we can at some point merge it with the https://github.com/openrewrite/rewrite/blob/748bb9bad72b743acf81863261fe04615cc6ddf5/rewrite-java/src/main/java/org/openrewrite/java/ParenthesizeVisitor.java#L28

timtebeek:
Indeed could be good, but seen how this visitor is tied into the recipe in rewrite-static-analysis I figured merge this increment already, thanks!

PR #6157, opened 2025-10-20
Fix PlainTextParser conversion marker handling

Jenson3210:
## Summary

Fixes an issue where converting a `SourceFile` to `PlainText` would incorrectly print markers to the text result and lose `Markup` markers in the process.

## Problem

When `PlainTextParser.convert()` was used to convert a `SourceFile` to `PlainText`:
1. Markers were being serialized into the text output because `printAll()` was called without sanitization
2. `Markup` markers attached to elements in the original source tree were not being preserved in the converted `PlainText` result

### `PlainTextParser.java` (rewrite-core:28)
- Changed `sourceFile.printAll()` to use `PrintOutputCapture.MarkerPrinter.SANITIZED` to prevent markers from being printed to text output
- Added `gatherMarkers()` method that traverses the source tree and collects all `Markup` markers
- Applied gathered markers to the converted `PlainText` result via `.withMarkers(gatherMarkers(sourceFile))`

### `FindAndReplace.java` (rewrite-core:3)
- Fixed bug where `sourceFile.getMarkers()` was used instead of `plainText.getMarkers()` when adding `AlreadyReplaced` marker
- Minor cleanup (removed trailing blank line)

### `PlainTextParserConversionTest.java` (rewrite-java:97)
- Added comprehensive test coverage demonstrating markers are properly preserved during conversion
- Test validates that `Markup`, `AlreadyReplaced`, and `RecipesThatMadeChanges` markers are all present after conversion

## Test Plan

- [x] Added test `markersAreAddedWhenConvertingToTextLst` verifying markers are preserved when conversion happens
- [x] Added test `markersAreKeptWhenNoConversionToTextLst` verifying markers are retained when no conversion occurs
- [x] Existing tests pass

### DIFF
--- a/rewrite-java/src/test/java/org/openrewrite/java/PlainTextParserConversionTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/PlainTextParserConversionTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2025 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


timtebeek:
Nice work! Should lead to fewer surprises for folks that leverage our search recipes.

Jenson3210:
Search Recipes I do not know. Unless you want me to also include the SearchResult Markers in the `gather` method?
Atm, we typically use `Markup` markers more for information to the UI and `SearchResult` for search recipes. Should we also pull these ones?

timtebeek:
Ah I already saw this in your PR:
```
                        if (marker instanceof SearchResult || marker instanceof Markup) {
```

But more so I meant the find and replace text recipes, which had bled printed markers into the commit diffs.

No need for further changes if you ask me!

Jenson3210:
You are right. I re-added this condition for SearchResults also and forgot about it. 🤦

PR #6158, opened 2025-10-20
Improved support for indentation formatting using records and single arg method declarations

Jenson3210:
## Summary

This PR improves indentation formatting support for Java records and method declarations, particularly for cases with single or multiple parameters that span multiple lines. The changes ensure consistent continuation indentation and proper alignment of parameters in both record components and method declarations.

#### `TabsAndIndentsVisitor.java`
- Extended indentation logic to support record components (primary constructor parameters) using the same alignment rules as method parameters
- Added support for single-parameter indentation when the parameter is on a new line
- Improved `computeFirstParameterColumn()` to handle both `J.MethodDeclaration` and `J.ClassDeclaration` (for records)
- Added `IndentType` message passing to maintain proper indentation context
- Fixed alignment calculation to consistently use `declPrefixLength + continuationIndent` when first parameter is on a new line
- Added `RECORD_STATE_VECTOR` and `METHOD_DECLARATION_PARAMETER` to argument alignment cases

#### `MinimumViableSpacingVisitor.java`
- Added spacing normalization for record primary constructor parameters
- Added spacing normalization for method declaration parameters
- Ensures parameters on the same line have proper spacing (no space before first param, single space before subsequent params)
- Preserves newlines when parameters are already on separate lines

#### `TabsAndIndentsStyle.java`
- Added new `RecordComponents` configuration class with `alignWhenMultiple` setting
- Mirrors the existing `MethodDeclarationParameters` structure for consistency

#### `Autodetect.java`
- Updated style auto-detection to include `RecordComponents` configuration
- Uses the same detection logic as method parameters for now (aligned vs non-aligned)

#### `IntelliJ.java`
- Updated IntelliJ style defaults to include `RecordComponents` with alignment enabled

#### `AutoFormatTest.java`
Added comprehensive test coverage for:
- **Record single parameter indentation** - Handles various incorrect indentation scenarios and preserves correct formatting
- **Record multiple parameter indentation** - Ensures consistent continuation indent for multiple record components
- **Method single parameter indentation** - Fixes incorrect indentation for single parameters on new lines
- **Method multiple parameter indentation** - Ensures proper alignment for multiple method parameters
- **Nested method invocations** - Tests indentation when method calls are nested as arguments with multiple parameters

#### `StyleHelperTest.java`
- Updated test to verify `RecordComponents` style merging works correctly

### Record with single parameter
```java
// Before
record someRecord(
String name) {
}

// After
record someRecord(
        String name) {
}
```

### Record with multiple parameters
```java
// Before
record someRecord(
String name,
int age) {
}

// After
record someRecord(
        String name,
        int age) {
}
```

### Method with single parameter
```java
// Before
void someMethod(
String name) {
}

// After
void someMethod(
        String name) {
}
```

## Technical Notes

- The implementation uses a `NoSuchMethodError` catch block for backward compatibility when `RecordComponents` is not available in older runtime environments/lsts
- Indentation calculations now properly account for the declaration prefix length plus continuation indent
- The changes maintain consistency between record component formatting and method parameter formatting

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/format/MinimumViableSpacingTest.java
@@ -142,7 +142,7 @@ class MinimumViableSpacingTest implements RewriteTest {
               }
               """,
             """
-              class A{public void foo(int paramA, final int paramB){int unassigned;int a=1;int b,c=5;final int d=10;final int e,f=20;}}
+              class A{public void foo(int paramA,final int paramB){int unassigned;int a=1;int b,c=5;final int d=10;final int e,f=20;}}
               """
           )
         );


rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -232,27 +234,52 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
...
-                                if (alignTo != -1) {
-                                    getCursor().getParentOrThrow().putMessage("lastIndent", alignTo - style.getContinuationIndent());
-                                    elem = visitAndCast(elem, p);
-                                    getCursor().getParentOrThrow().putMessage("lastIndent", indent);
-                                    after = indentTo(right.getAfter(), t == lastArg ? indent : alignTo, loc.getAfterLocation());
+                        //noinspection ConstantConditions
+                        J tree = null;
+                        if (loc == JRightPadded.Location.METHOD_DECLARATION_PARAMETER) {
+                            tree = getCursor().firstEnclosing(J.MethodDeclaration.class);
+                        } else if (loc == JRightPadded.Location.RECORD_STATE_VECTOR) {
```
timtebeek: My IDE gives a warning that this statement is always true, as we already `switch` on `loc` above.
```suggestion
                        } else {
```

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -232,27 +234,52 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
...
+                            tree = getCursor().firstEnclosing(J.ClassDeclaration.class);
+                            getCursor().getParentOrThrow().putMessage("indentType", IndentType.CONTINUATION_INDENT);
+                        }
+                        if (elements.size() > 1) {
+                            try {
+                                if ((loc == JRightPadded.Location.METHOD_DECLARATION_PARAMETER && style.getMethodDeclarationParameters().getAlignWhenMultiple()) ||
+                                        ((loc == JRightPadded.Location.RECORD_STATE_VECTOR && style.getRecordComponents() != null && style.getRecordComponents().getAlignWhenMultiple()))) {
```
timtebeek: `style.getRecordComponents()` is not marked as `@Nullable`; should it be? Or can we remove this guard since we seem to rely on the `NoSuchMethodError` below.

Jenson3210: as with the try/catch, we had NPE's, No SuchMethodErrors last time when introducing new style classes. So added like this to be safe. We should then later be able to remove the `try/catch`'s and the `if != null` statements

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
Jenson3210: Primarily had it like this as the existing code was only written for METHOD_DECLARATION_PARAMETER and then the case `RECORD_STATE_VECTOR` was added which was actually not working. So wanted to avoid that when we would introduce a new 3rd case.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -389,23 +418,27 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
...
-        List<JRightPadded<Statement>> arguments = method.getPadding().getParameters().getPadding().getElements();
+    private int computeFirstParameterColumn(J tree) {
+        List<JRightPadded<Statement>> arguments;
+        if (tree instanceof J.MethodDeclaration) {
+            arguments = ((J.MethodDeclaration) tree).getPadding().getParameters().getPadding().getElements();
+        } else if (tree instanceof J.ClassDeclaration) {
+            JContainer<Statement> primaryConstructorArgs = ((J.ClassDeclaration) tree).getPadding().getPrimaryConstructor();
+            arguments = primaryConstructorArgs == null ? emptyList() : primaryConstructorArgs.getPadding().getElements();
+        } else {
+            throw new IllegalStateException("Unexpected tree type: " + tree.getClass().getSimpleName());
```
timtebeek: This `throw` threw me off; are we sure we want to keep that in or can we more safely fall back to for instance `-1` ?

timtebeek: Alternatively can we pull out the type checks here near to where we already have type checks above in
https://github.com/openrewrite/rewrite/pull/6158/files#diff-cefab22c95ea0364230fa1c46a1c5eb0e71db2456b6d22486f01ae9434b80c6cR238-R244 ?

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
sambsnyd: I think it makes more sense for these temporary backwards compatibility null checks to exist within the relevant style class. You can always provide an implementation for `getRecordComponents()` which returns the default value when the underlying field is null. Then when the null check is no longer relevant there is only one place to remove it from, rather than from every call site.

Jenson3210: That makes sense in this occasion as we will introduce both the style and the default getter in a single go. It will hence be either a NoSuchMethod or a defaulted getter. Last time we also had the possibilitiy of an old (non-defaulted) getter being parent loaded. 
Adapted!

Jenson3210:
ran with cli against apache org -> no errors datatable

PR #6163, opened 2025-10-21
Fix `JavaScriptReceiver#visitJsxTag()` in `rpc.ts`

knutwannheden:
The receiver should always process all properties of the `JSX.Tag` objects. Otherwise, the alignment with the sender will become a problem and cause RPC errors.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/rpc.ts
+++ b/rewrite-javascript/rewrite/src/javascript/rpc.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import {JavaScriptVisitor} from "./visitor";
-import {asRef, RpcCodec, RpcCodecs, RpcReceiveQueue, RpcSendQueue} from "../rpc";
+import {asRef, RpcReceiveQueue, RpcSendQueue} from "../rpc";
 import {isJavaScript, JS, JSX} from "./tree";
 import {Expression, J, Statement, Type, TypedTree, TypeTree} from "../java";
 import {JavaReceiver, JavaSender, registerJLanguageCodecs} from "../java/rpc";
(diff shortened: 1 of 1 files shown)


rewrite-javascript/rewrite/src/javascript/rpc.ts
```diff
@@ -1005,29 +1005,17 @@ class JavaScriptReceiver extends JavaScriptVisitor<RpcReceiveQueue> {
...
                 afterName: await q.receive(tag.afterName, space => this.visitSpace(space, q)),
                 attributes: await q.receiveListDefined(tag.attributes, attr => this.visitRightPadded(attr, q)),
+                selfClosing: await q.receive(tag.selfClosing, space => this.visitSpace(space, q)),
                 children: await q.receiveListDefined(tag.children, child => this.visit(child, q)),
                 closingName: await q.receive(tag.closingName, el => this.visitLeftPadded(el, q)),
                 afterClosingName: await q.receive(tag.afterClosingName, el => this.visitSpace(el, q))
             };
-            return updateIfChanged(tag, updates);
-        }
+            return updateIfChanged(tag, updates as any);
```
greg-at-moderne, line 1019: I understand this `any` comes from the fact that we have `readonly children?: undefined;` in one of the union variants.

greg-at-moderne (approved):

rewrite-javascript/rewrite/src/javascript/rpc.ts
knutwannheden: There might indeed be a cleaner way to do this.

knutwannheden: We could for example add `as any` on the last three properties.

PR #6164, opened 2025-10-21
Simplify numeric literal comparisons in SimplifyBooleanExpression

timtebeek:
## Summary

Enhanced `SimplifyBooleanExpressionVisitor` to evaluate and simplify compile-time constant numeric comparisons. The visitor now simplifies expressions like:
- `1 < 2` → `true`
- `3 == 4` → `false`
- `5 > 3` → `true`
- `2 <= 2` → `true`
- `10 >= 20` → `false`
- `7 != 7` → `false`

This improvement allows for more aggressive constant folding of numeric literal comparisons across all comparison operators (`==`, `!=`, `<`, `>`, `<=`, `>=`), resulting in cleaner and more readable code after recipe application.

### Changes Made

- Added `isNumericLiteral()` helper method to detect numeric literal expressions
- Added `getNumericValue()` to extract numeric values from literals
- Added `compareNumericLiterals()` to evaluate compile-time constant comparisons
- Extended `visitBinary()` to handle numeric literal comparisons for all comparison operators
- Updated existing test `doubleNegatedBinaryWithParentheses()` to reflect improved simplification behavior
- Added new test `simplifyNumericLiteralComparisons()` covering various numeric types (int, long, float, double) and all comparison operators

## Test plan

- [x] Compilation successful
- [x] All 73 tests in `SimplifyBooleanExpressionVisitorTest` pass
- [x] New test `simplifyNumericLiteralComparisons()` verifies numeric comparison simplification
- [x] Updated test `doubleNegatedBinaryWithParentheses()` validates behavior with negated numeric comparisons
- [x] No regressions in existing functionality

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
@@ -304,7 +304,7 @@ public class SimplifyBooleanExpressionVisitor extends JavaVisitor<ExecutionConte
         return expression instanceof J.Literal && ((J.Literal) expression).getType() != JavaType.Primitive.Null;
     }
 
-    private static boolean isNumeric(Expression expression) {
+    private static boolean isNumericLiteral(Expression expression) {
         return expression instanceof J.Literal &&
                 ((JavaType.Primitive) expression.getType()).isNumeric();
     }
@@ -321,7 +321,7 @@ public class SimplifyBooleanExpressionVisitor extends JavaVisitor<ExecutionConte
     }
 
     private @Nullable Boolean compareNumericLiterals(J.Binary binary) {
-        if (!isNumeric(binary.getLeft()) || !isNumeric(binary.getRight())) {
+        if (!isNumericLiteral(binary.getLeft()) || !isNumericLiteral(binary.getRight())) {
             return null;
         }
 


timtebeek:
While on the surface thee might seem unlikely & silly, in practice they happen when we do other replacements as seen in
- https://github.com/openrewrite/rewrite-feature-flags/pull/54
- https://github.com/openrewrite/rewrite-feature-flags/pull/55

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitorTest.java
```diff
@@ -836,4 +836,46 @@ void doSome(String o1, String o2) {
...
+    @Test
+    void simplifyNumericLiteralComparisons() {
+        rewriteRun(
+          java(
+            """
+              class A {
+                  {
+                      boolean a = 1 < 2;
+                      boolean b = 3 == 4;
+                      boolean c = 5 > 3;
```
greg-at-moderne, line 849: I'd add some tests for Double/Float literals, especially the strange ones like `-0.0` or `NaN, etc.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitorTest.java
```diff
@@ -836,4 +836,46 @@ void doSome(String o1, String o2) {
...
     }
+
+    @Test
+    void simplifyNumericLiteralComparisons() {
+        rewriteRun(
+          java(
+            """
+              class A {
+                  {
+                      boolean a = 1 < 2;
```
greg-at-moderne, line 847: I think we should add the test for non-literals, e.g.
```
int i = 1;
bolean eq = i == 1;
```
and other combinations (two non-literals, etc.)

AFAICT we don't expect any changes here.

rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
```diff
@@ -281,6 +304,45 @@ private boolean isNonNullLiteral(Expression expression) {
...
     }
 
+    private static boolean isNumeric(Expression expression) {
+        return expression instanceof J.Literal &&
+                ((JavaType.Primitive) expression.getType()).isNumeric();
+    }
+
+    private @Nullable Double getNumericValue(Expression expression) {
+        Object value = ((J.Literal) expression).getValue();
+        return value instanceof Number ? ((Number) value).doubleValue() : null;
```
greg-at-moderne: `Double` doesn't represent all `Long` values equally, so I think this might introduce undesired changes to the code.
E.g. `new Long(9007199254740992L).doubleValue() < new Long(9007199254740993L).doubleValue()` returns false, while `9007199254740992L < 9007199254740993L` returns true.

greg-at-moderne (changes requested):
Requesting changes for the `double` forced cast.

rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitorTest.java
```diff
@@ -878,4 +878,83 @@ class A {
...
+                  }
+              }
+              """,
+            """
+              class A {
+                  {
+                      // These large long values lose precision when cast to double
+                      // The current implementation incorrectly simplifies these
+                      boolean i = false;
+                      boolean j = true;
```
timtebeek: TODO

rewrite-java/src/main/java/org/openrewrite/java/cleanup/SimplifyBooleanExpressionVisitor.java
timtebeek: Thanks for knowing and pointing out these subtle edge cases! I've added additional tests and reworked the implementation.

PR #6166, opened 2025-10-22
Fix protobuf parser to handle files without syntax declaration

### LINKED ISSUE #6165: Failure to parse Proto files with no `syntax` declaration

Current main = 54657a0964f17471d59626305d9b26f52f07d9c9

The following test added to `ProtoParserTest`
```java
    @Test
    void point() {
        rewriteRun(
          proto(
            """
            message Point {
              required int32 x = 1;
              required int32 y = 2;
              optional string label = 3;
            }
            """
          )
        );
    }
```
fails with:
```
Syntax error in file.proto at line 1:0 mismatched input 'message' expecting 'syntax'.
org.openrewrite.protobuf.ProtoParsingException: Syntax error in file.proto at line 1:0 mismatched input 'message' expecting 'syntax'.
	at org.openrewrite.protobuf.ProtoParser$ForwardingErrorListener.syntaxError(ProtoParser.java:110)
[...]
	at org.antlr.v4.runtime.DefaultErrorStrategy.recoverInline(DefaultErrorStrategy.java:483)
```

## Context

I don't have experience with ProtoBuf, but I think this is a valid proto file.

https://protobuf.dev/reference/protobuf/proto2-spec/ says:

> The syntax statement is used to define the protobuf version. If syntax is omitted, the protocol compiler will use proto2

which allows for no `syntax` line.

## OSS repro

https://github.com/ajaxorg/ace-builds/blob/56a46777e92abd648dce1c29069c13a74480f9a5/demo/kitchen-sink/docs/protobuf.proto

The error message reported by Moderne CLI for this file is:
```
"java.lang.NullPointerException: Cannot invoke "org.openrewrite.protobuf.internal.grammar.Protobuf2Parser$StringLiteralContext.StringLiteral()" because the return value of "org.openrewrite.protobuf.internal.grammar.Protobuf2Parser$SyntaxContext.stringLiteral()" is null
  org.openrewrite.protobuf.internal.ProtoParserVisitor.visitSyntax(ProtoParserVisitor.java:422)
  org.openrewrite.protobuf.internal.ProtoParserVisitor.visitProto(ProtoParserVisitor.java:320)"
```

timtebeek:
## Summary

- Fixes #6165

The protobuf parser was failing to parse proto files without a `syntax` declaration. According to the [protobuf spec](https://protobuf.dev/reference/protobuf/proto2-spec/), the syntax statement is optional:

> The syntax statement is used to define the protobuf version. If syntax is omitted, the protocol compiler will use proto2

This fix allows the parser to correctly handle legacy proto files and OSS projects that don't explicitly declare the syntax version.

### 1. ANTLR Grammar (`Protobuf2Parser.g4`)
- Made the `syntax` rule optional by changing `syntax` to `syntax?` in the proto rule

### 2. Parser Visitor (`ProtoParserVisitor.java`)
- Added logic to detect when no syntax declaration is present
- Creates a synthetic proto2 syntax node with an `ImplicitProto2Syntax` marker
- Adjusts the parsing index to handle both cases (with/without syntax)

### 3. Implicit Syntax Marker (`marker/ImplicitProto2Syntax.java`)
- New marker class to track synthetic syntax nodes that weren't in the original source
- Used to maintain lossless round-trip parsing

### 4. Printer (`ProtoPrinter.java`)
- Updated `visitSyntax` to skip printing syntax nodes marked as implicit
- Ensures that files parsed without syntax are printed back without syntax

### 5. Test Coverage (`ProtoParserTest.java`)
- Added `protoWithoutSyntaxDeclaration()` test case from the issue

## Testing

The fix handles the exact test case from the issue:
```protobuf
message Point {
  required int32 x = 1;
  required int32 y = 2;
  optional string label = 3;
}
```

All existing tests pass, and the new test verifies that:
- Files without syntax declarations parse successfully
- The parser defaults to proto2 (as per spec)
- Round-trip parsing is lossless (source → AST → source)

## OSS Impact

This fix will enable parsing of real-world proto files like [ace-builds/protobuf.proto](https://github.com/ajaxorg/ace-builds/blob/56a46777e92abd648dce1c29069c13a74480f9a5/demo/kitchen-sink/docs/protobuf.proto) that omit the syntax declaration.

### DIFF
--- a/rewrite.yml
+++ b/rewrite.yml
@@ -21,6 +21,9 @@ type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.recipes.rewrite.OpenRewriteRecipeBestPracticesSubset
 displayName: OpenRewrite best practices
 description: Best practices for OpenRewrite recipe development.
+preconditions:
+  - org.openrewrite.FindSourceFiles:
+      filePattern: "!(**/internal/grammar/*.*)"
 recipeList:
   - org.openrewrite.recipes.JavaRecipeBestPracticesSubset
   - org.openrewrite.recipes.RecipeTestingBestPracticesSubset


timtebeek:
FYI: Now that we have our own fork of https://github.com/moderneinc/googleapis-code-suggester I've stopped applying the code suggestions to generated files.

PR #6167, opened 2025-10-22
Add custom style configuration option to AutoFormat recipe

Jenson3210:
## Summary

Adds a `style` option to the `AutoFormat` recipe that accepts YAML configuration for custom named styles. This allows users to specify formatting styles declaratively when using the AutoFormat recipe, without needing to pre-configure styles in the Environment or attach them as markers to source files.

## Changes

- **AutoFormat recipe**: Added optional `String style` parameter that accepts YAML in OpenRewrite's style specification format
- **Style parsing**: Implemented `computeNamedStyles()` method that uses `YamlResourceLoader` to parse the YAML string and convert it to `NamedStyles` objects

## Usage Example

```yaml
name: com.example.CustomAutoFormat
recipeList:
  - org.openrewrite.java.format.AutoFormat:
      style: |
        type: specs.openrewrite.org/v1beta/style
        name: CustomStyle
        displayName: Custom Formatting Style
        styleConfigs:
          - org.openrewrite.java.style.TabsAndIndentsStyle:
              useTabCharacter: false
              tabSize: 4
          - org.openrewrite.java.style.WrappingAndBracesStyle:
              chainedMethodCalls:
                wrap: WrapAlways
                builderMethods:
                  - builder
                  - newBuilder
```

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
@@ -23,21 +23,17 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.style.*;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaSourceFile;
-import org.openrewrite.marker.Markers;
 import org.openrewrite.style.GeneralFormatStyle;
 import org.openrewrite.style.NamedStyles;
-import org.openrewrite.style.Style;
-import org.openrewrite.style.StyleHelper;
 
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
-import java.util.function.Supplier;
 
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 import static org.openrewrite.java.format.AutodetectGeneralFormatStyle.autodetectGeneralFormatStyle;
+import static org.openrewrite.style.StyleHelper.addStyleMarker;
+import static org.openrewrite.style.StyleHelper.getStyle;
 
 public class AutoFormatVisitor<P> extends JavaIsoVisitor<P> {
     @Nullable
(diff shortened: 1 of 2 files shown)


Jenson3210:
Added an additional merge so that the intelliJ defaults are taken by default and the mentioned style is only overwriting deviating properties

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormat.java
```diff
@@ -32,6 +56,14 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new AutoFormatVisitor<>(null);
+        return new AutoFormatVisitor<>(null, computeNamedStyles());
+    }
```
sambsnyd, line 60: Allowing `AutoFormat` to override the detected/configured styles with a parameter is good and useful. 
Part of that should be overriding the attached style markers on the LST with the style provided to `AutoFormat`. Otherwise in a composite recipe the style-overriding `AutoFormat` could have its changes overridden by a subsequent visitor which continues to refer to the styles attached to the LSTs.

Jenson3210: Adding the style marker now (if not present yet ~ cycles)

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
```diff
@@ -121,25 +131,52 @@ public J visit(@Nullable Tree tree, P p) {
...
+                Markers markers = t.getMarkers().removeByType(NamedStyles.class);
+                for (NamedStyles namedStyle : newNamedStyles) {
+                    markers = markers.add(namedStyle);
+                }
+
+                return t.withMarkers(markers);
+            }
+        }
+        return t;
+    }
```
sambsnyd: Nothing about this is really specific to `JavaSourceFile` or the java-specific version of `AutoFormat`. This could be a method in `StyleHelper` which accepts a `Markers` or a `SourceFile`. That would then be available to all other implementations of `AutoFormat` which may want to adopt similar functionality for similar reasons.

PR #6177, opened 2025-10-23
Always wrap builder methods and support align when multiline

Jenson3210:
## Summary

This PR implements two key improvements to method chain wrapping based on IntelliJ IDEA's code style documentation:

1. **Always wrap builder methods** - Builder methods are now wrapped regardless of wrap settings
2. **Add `alignWhenMultiline` support** - Method chains can now be vertically aligned when wrapped across multiple lines

### 1. Always Wrap Builder Methods

The IntelliJ docs (copied below) mention that builder methods are **always wrapped**. The previous implementation only wrapped them when explicitly specified. This PR fixes that behavior to match IntelliJ.

### 2. Align When Multiline Support

Added a new `alignWhenMultiline` boolean flag to `WrappingAndBracesStyle.ChainedMethodCalls` that controls whether wrapped method chains should be vertically aligned.

**When `alignWhenMultiline: false` (default):** the continuationIndent is used
```java
MyObject obj = MyObject.builder()
        .name("test")
        .age(25)
        .build();
```

**When `alignWhenMultiline: true`:**
```java
MyObject obj = MyObject.builder()
                       .name("test")
                       .age(25)
                       .build();
```

## IntelliJ IDEA Documentation Reference

https://www.jetbrains.com/help/idea/2025.1/code-style-java.html?#chained-method-calls

### Chained method calls

Use the following options to format chained method calls and make them easier to read. **Note that builder method calls are always wrapped regardless of the settings for chained calls.**

Item | Description
-- | --
Wrap first call | Allow wrapping the first method call in chained methods.
**Align when multiline** | **Align several method calls.**
Builder methods | Specify comma-separated names (identifiers) of methods that you want to be treated as builder methods. For example: withTitle,withYear,addAuthor,withName,withLastName,end,build
Keep builder methods indents | Keep additional indents that you insert manually intact as you reformat code.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
@@ -258,6 +258,7 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
                                         try {
                                             alignTo = sourceFile.service(SourcePositionService.class).computeColumnToAlignTo(new Cursor(getCursor(), elem), style.getContinuationIndent());
                                         } catch (UnsupportedOperationException e) {
+                                            // Sourcefile$service(...) might give UnsupportedOperationException depending on the runtime, the lst build date... we use deprecated behavior in that case
                                             alignTo = computeFirstParameterColumn(tree);
                                         }
                                         if (alignTo != -1) {
@@ -287,6 +288,7 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
                                 try {
                                     alignTo = sourceFile.service(SourcePositionService.class).computeColumnToAlignTo(new Cursor(getCursor(), elem), style.getContinuationIndent());
                                 } catch (UnsupportedOperationException error) {
+                                    // Sourcefile$service(...) might give UnsupportedOperationException depending on the runtime, the lst build date... we use deprecated behavior in that case
                                     alignTo = computeFirstParameterColumn(tree);
                                 }
                                 if (alignTo != -1) {
@@ -384,6 +386,7 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
                             try {
                                 alignTo = sourceFile.service(SourcePositionService.class).computeColumnToAlignTo(new Cursor(getCursor(), elem), style.getContinuationIndent());
                             } catch (UnsupportedOperationException e) {
+                                // Sourcefile$service(...) might give UnsupportedOperationException depending on the runtime, the lst build date... we do not align in that case.
                                 alignTo = -1;
                             }
                             if (alignTo != -1) {
@@ -451,6 +454,7 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
         return (after == right.getAfter() && t == right.getElement()) ? right : new JRightPadded<>(t, after, right.getMarkers());
     }
 
+    @Deprecated
     private int computeFirstParameterColumn(J tree) {
         List<JRightPadded<Statement>> arguments;
         if (tree instanceof J.MethodDeclaration) {


timtebeek:
Still [one test failure](https://ge.openrewrite.org/s/6jo6w2b4wi35o/tests/task/:rewrite-java-test:test/details/org.openrewrite.java.format.TabsAndIndentsTest/alignMethodDeclarationParamsWhenMultiple()?top-execution=1) it seems.

Jenson3210:
Any reason why these tests for a lot of these formatting recipes are not in the test package of the code itself? I typically just run the current package's unit tests locally before pushing.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -247,7 +249,8 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
                                 if ((loc == JRightPadded.Location.METHOD_DECLARATION_PARAMETER && style.getMethodDeclarationParameters().getAlignWhenMultiple()) ||
                                         (loc == JRightPadded.Location.RECORD_STATE_VECTOR && style.getRecordComponents().getAlignWhenMultiple())) {
                                     if (tree != null) {
-                                        int alignTo = computeFirstParameterColumn(tree);
+                                        JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
+                                        int alignTo = sourceFile.service(SourcePositionService.class).computeColumnToAlignTo(new Cursor(getCursor(), elem), style.getContinuationIndent());
```
Jenson3210: SourceFile is parent loaded -> `java.lang.UnsupportedOperationException: Service class org.openrewrite.java.service.SourcePositionService not supported`

We need a cli release + workers in order for this to work.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -247,7 +252,13 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
                                 if ((loc == JRightPadded.Location.METHOD_DECLARATION_PARAMETER && style.getMethodDeclarationParameters().getAlignWhenMultiple()) ||
                                         (loc == JRightPadded.Location.RECORD_STATE_VECTOR && style.getRecordComponents().getAlignWhenMultiple())) {
                                     if (tree != null) {
-                                        int alignTo = computeFirstParameterColumn(tree);
+                                        JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
+                                        int alignTo;
+                                        try {
```
Jenson3210, line 258: SourceFile is parent loaded -> java.lang.UnsupportedOperationException: Service class org.openrewrite.java.service.SourcePositionService not supported

We need a cli release + workers in order for this to work.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -270,7 +281,13 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
                             }
                         } else {
                             if (elem.getPrefix().getLastWhitespace().contains("\n") && tree != null) {
-                                int alignTo = computeFirstParameterColumn(tree);
+                                JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
+                                int alignTo;
+                                try {
```
Jenson3210, line 288: SourceFile is parent loaded -> java.lang.UnsupportedOperationException: Service class org.openrewrite.java.service.SourcePositionService not supported

We need a cli release + workers in order for this to work.

rewrite-java/src/main/java/org/openrewrite/java/format/WrapMethodChains.java
```diff
@@ -63,48 +61,45 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, P ctx
...
+                        after = after.withComments(ListUtils.mapLast(after.getComments(), comment -> comment == null ? null : comment.withSuffix("\n")));
+                    }
+                    if (after != m.getPadding().getSelect().getAfter()) {
+                        m = m.getPadding().withSelect(m.getPadding().getSelect().withAfter(after))
+                                .withArguments(ListUtils.map(m.getArguments(), arg -> arg.withPrefix(Space.EMPTY)));
+                    }
                 }
             }
-        } catch (NoSuchMethodError ignore) {
+        } catch (NoSuchMethodError | NoSuchFieldError ignore) {
```
Jenson3210, line 87: Added Wrapping type is parent loaded

We need a cli release + workers in order for this to work.

Jenson3210:
Tested in cli with Apache and see no errors. But with all these try/catches in place the code is also not in effect there. 
Impossible to test further this without having the releases of worker / cli that support the new service added in a previous pr and the enum value added in this PR.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -361,7 +378,17 @@ public Space visitSpace(Space space, Space.Location loc, P p) {
                         break;
                     default:
                         elem = visitAndCast(elem, p);
-                        after = visitSpace(right.getAfter(), loc.getAfterLocation(), p);
+                        if (right.getAfter().getLastWhitespace().contains("\n") && alignWhenMultiple(loc)) {
+                            JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
+                            int alignTo = sourceFile.service(SourcePositionService.class).computeColumnToAlignTo(new Cursor(getCursor(), elem), style.getContinuationIndent());
```
timtebeek: No try/catch on that `UnsupportedOperationException` here?

rewrite-java/src/main/java/org/openrewrite/java/format/WrapMethodChains.java
```diff
@@ -63,48 +61,45 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, P ctx
...
             // Styles are parent-first loaded and this can happen if the style is from a older version of the runtime. Can be removed in future releases.
         }
 
         return m;
     }
 
-    private J.@Nullable MethodInvocation findChainStarterInChain(J.MethodInvocation method, List<MethodMatcher> matchers) {
+    private J findChainStarterInChain(J.MethodInvocation method) {
```
timtebeek, line 94: Wouldn't this be clearer as just
```java
    private @Nullable J findChainStarterInChain(J.MethodInvocation method) {
        Expression current = method.getSelect();
        while (current instanceof J.MethodInvocation) {
            current = ((J.MethodInvocation) current).getSelect();
        }
        return current;
    }
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -68,9 +122,47 @@ private boolean needsSemicolon(Statement statement) {
...
+        }
+        return cursor;
+    }
+
+    private @Nullable Cursor alignsWith(Cursor cursor) {
+        J cursorValue = cursor.getValue();
+        SourceFile sourceFile = cursor.firstEnclosing(SourceFile.class);
+        return new JavaIsoVisitor<AtomicReference<Cursor>>() {
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicReference<Cursor> ctx) {
```
timtebeek:
```suggestion
            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicReference<@Nullable Cursor> ctx) {
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -68,9 +122,47 @@ private boolean needsSemicolon(Statement statement) {
...
+                    method = super.visitMethodInvocation(method, ctx);
+                    if (!(method.getSelect() instanceof J.MethodInvocation) && getCursor().getPathAsStream(o -> o instanceof J.MethodInvocation).anyMatch(value -> SemanticallyEqual.areEqual((J) value, cursorValue))) {
+                        ctx.set(getCursor());
+                    }
+                }
+                return method;
+            }
+
+            @Override
+            public @Nullable <J2 extends J> JContainer<J2> visitContainer(@Nullable JContainer<J2> container, JContainer.Location loc, AtomicReference<Cursor> ctx) {
```
timtebeek:
```suggestion
            public @Nullable <J2 extends J> JContainer<J2> visitContainer(@Nullable JContainer<J2> container, JContainer.Location loc, AtomicReference<@Nullable Cursor> ctx) {
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -68,9 +122,47 @@ private boolean needsSemicolon(Statement statement) {
...
+                if (ctx.get() == null) {
+                    if (container.getElements().stream().anyMatch(e -> e == cursorValue)) {
+                        ctx.set(new Cursor(getCursor(), container.getElements().get(0)));
+                    }
+                }
+                return super.visitContainer(container, loc, ctx);
+            }
+        }.reduce(sourceFile, new AtomicReference<>(), new Cursor(cursor.getRoot(), sourceFile)).get();
```
timtebeek:
```suggestion
        }.reduce(sourceFile, new AtomicReference<@Nullable Cursor>(), new Cursor(cursor.getRoot(), sourceFile)).get();
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -15,38 +15,92 @@
...
+                @Override
+                public J visitMethodInvocation(J.MethodInvocation method, PrintOutputCapture<TreeVisitor<?, ?>> p) {
+                    if (SemanticallyEqual.areEqual(method, alignWith.getValue())) {
+                        beforeSyntax(method, Space.Location.METHOD_INVOCATION_PREFIX, p);
+                        visitRightPadded(method.getPadding().getSelect(), JRightPadded.Location.METHOD_SELECT, "", p);
+                        indentation.set(p.getOut().length());
+                        return method;
+                    } else {
+                        return super.visitMethodInvocation(method, p);
+                    }
```
timtebeek:
```suggestion
                    }
                    return super.visitMethodInvocation(method, p);
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -15,38 +15,92 @@
...
+                    return super.append(text);
+                }
+            };
+            javaPrinter.visit(j, printLine, cursor.getParentOrThrow());
+
+            return indentation.get();
+        } else {
+            throw new RuntimeException("Unable to calculate length due to unexpected cursor value: " + newLinedElementCursor.getValue().getClass());
+        }
```
timtebeek:
```suggestion
        }
        throw new RuntimeException("Unable to calculate length due to unexpected cursor value: " + newLinedElementCursor.getValue().getClass());
```

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
Jenson3210: we would normally never fall in this branch as other methods are also only now introduced to fall into it. But better safe than sorry! Adding it.

rewrite-java/src/main/java/org/openrewrite/java/format/WrapMethodChains.java
Jenson3210: We do not want to return null from this one.  The proposed method will return null if the select is a static imported method or an reference to a in-class method or ... (in which case the starter should be that first method itself)
What do you think about 
```
private J findChainStarterInChain(J.MethodInvocation method) {
        J.MethodInvocation chainStarter = method;
        Expression select = method.getSelect();
        while (select instanceof J.MethodInvocation) {
            chainStarter = (J.MethodInvocation) select;
            select = chainStarter.getSelect();
        }
        return chainStarter;
    }
```

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: Thanks! I did not know how to solve that warning.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
sambsnyd: Let's throw a comment in there so later on maybe someone eventually remembers to remove this

sambsnyd (approved):
:+1:

PR #6178, opened 2025-10-23
JavaScript: Properly map literal method names

knutwannheden:
The `J.MethodDeclaration#name` is expected to be of type `J.Identifier`. So when we in JS have a method with a name like `2` or `'foo bar'`, we need to map it to an identifier, where we set that source text to the identifier's `simpleName`.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/parser.ts
@@ -968,11 +968,10 @@ export class JavaScriptParserVisitor {
             };
         }
 
-        let name: J.Identifier = !node.name
-            ? this.mapIdentifier(node, "")
-            : ts.isStringLiteral(node.name)
-                ? this.mapIdentifier(node.name, node.name.getText())
-                : this.visit(node.name);
+        let name = this.mapMethodName(node) as J.Identifier;
+        name = produce(name, draft => {
+            draft.markers = this.maybeAddOptionalMarker(draft, node);
+        });
 
         return {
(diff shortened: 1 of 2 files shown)


jkschneider:
I would instead do the same thing we did on J.VariableDeclarations.NamedVariable here and allow us to widen the name. Not sure if I'd necessarily choose the same `VariableDeclarator` interface since I suppose in this case there cannot be multiple names. Or can there?

image

rewrite-javascript/rewrite/src/javascript/parser.ts
```diff
@@ -968,11 +968,10 @@ export class JavaScriptParserVisitor {
...
         }
 
-        let name: J.Identifier = !node.name
-            ? this.mapIdentifier(node, "")
-            : ts.isStringLiteral(node.name)
-                ? this.mapIdentifier(node.name, node.name.getText())
-                : this.visit(node.name);
+        let name = this.mapMethodName(node) as J.Identifier;
+        name = produce(name, draft => {
+            draft.markers = this.maybeAddOptionalMarker(draft, node);
```
jkschneider, line 973: Consider widening name on `MethodDeclaration` rather than a marker here.

knutwannheden: Just to be clear: The marker being added here has nothing to do with this PR. That was already present before.

But I get your point. I will try to see if we can do something along the lines of `VariableDeclarator`. I just figured that we might as well just "squash" it into a `J.Identifier`.

PR #6192, opened 2025-10-24
Add annotation search methods to AnnotationService

### LINKED ISSUE #6181: Add method(s) to AnnotationService method to "discover" annotations using the type system
It would be very helpful to extend the `AnnotationService` such that you could query an element to determine if it is annotated with a specific type. Currently, the methods only consider annotations present on the LST. 

As a concrete example, lets assume a recipe is attempting to determine if a `J.ClassDeclaration` is annotated with `RestController` and you run into the following scenario:

```
@RestController
public class BaseRestController {
...
}
```

```
public class CustomerController extends BaseRestController {
...
}
```

What can make this more complicated is if the base class is a binary dependency from a different project/module.

Ideally, this new functionality would operate on the type information to search for such annotations.

I am suggesting to add additional methods to the AnnotationService such as `isAnnotatedWith()` or `hasAnnotation()` that would navigate the JavaType information to ensure base classes and binary dependencies are also considered.

I think these could be targeted at specific, common use cases: 

`J.ClassDeclaration`
`J.MethodDeclaration`
`J.Annotation` (where an annotation is a composed annotation, `@SpringBootApplication` should match on `SpringBootConfiguration`)

sambsnyd:
- Fixes #6181

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/service/AnnotationServiceTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/service/AnnotationServiceTest.java
@@ -215,7 +215,7 @@ class AnnotationServiceTest implements RewriteTest {
                         // Direct getAllAnnotations should not find inherited annotations
                         assertThat(service.getAllAnnotations(getCursor())).isEmpty();
 
-                        // But hasAnnotation should find it through the type hierarchy
+                        // But isAnnotatedWith should find it through the type hierarchy
                         assertThat(service.isAnnotatedWith(classDecl, "BaseAnnotation")).isTrue();
                         assertThat(service.annotatedWith(classDecl, "BaseAnnotation")).hasSize(1);
                     }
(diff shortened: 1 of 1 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/service/AnnotationServiceTest.java
```diff
@@ -190,4 +190,117 @@ public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, Integer integer
...
+                    if (parentCursor.getValue() instanceof J.ClassDeclaration) {
+                        J.ClassDeclaration classDecl = parentCursor.getValue();
+                        if ("ChildClass".equals(classDecl.getSimpleName()) &&
+                            "doSomething".equals(method.getSimpleName())) {
+                            // Direct getAllAnnotations should only show @Override
+                            List<J.Annotation> directAnnotations = service.getAllAnnotations(getCursor());
+                            assertThat(directAnnotations).hasSize(1);
+                            assertThat(directAnnotations.get(0).getSimpleName()).isEqualTo("Override");
+
+                            // But hasAnnotation should find @MethodAnnotation from parent
```
tkvangorder: Looks great, note the comment says `hasAnnotation` but the actual method is `isAnnotatedWith`. This is my fault I suggested both names in the issue. 8) 
The previous test has a similar comment

tkvangorder (commented):
Looks great Sam, thanks for turning this around so quickly.

rewrite-java-test/src/test/java/org/openrewrite/java/service/AnnotationServiceTest.java
sambsnyd: It's my fault for being indecisive about what to call it. Did it one way, then changed it to the other

PR #6200, opened 2025-10-27
Fix calculation non-aligned cases (first variable on new line)

Jenson3210:
## Summary

Fixes position calculation in `SourcePositionService` for non-aligned cases where the first variable or element appears on a new line instead of being aligned with subsequent elements.

## Problem

The previous implementation of `computeColumnToAlignTo()` didn't properly handle cases where:
1. The first element in a container (e.g., method parameters, chained method calls) was already on a new line
2. Chained method invocations where alignment logic needed to traverse to the root invocation

This led to incorrect indentation calculations for formatting, particularly when:
- Method parameters start on a new line after the opening parenthesis
- Method chain calls starter calls like `String.valueOf()` or `new StringBuilder().append()` span multiple lines

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
@@ -25,9 +25,26 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.util.Collections.emptyList;
 
+/**
+ * Service for computing source code position metrics such as column alignment positions and tree element lengths.
+ * <p>
+ * This service is useful for formatting and layout calculations, particularly when determining how to align
+ * elements in chained method calls, multi-line variable declarations, and method parameters.
+ */
 @Incubating(since = "8.63.0")
 public class SourcePositionService {
 
+    /**
+     * Computes the column position where an element should be aligned to.
+     * <p>
+     * For elements that should align with a previous element (e.g., in method chains or parameter lists),
+     * this calculates the column position of that alignment point. For elements that don't align,
+     * it returns the parent's indentation plus the continuation indent.
+     *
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -138,31 +138,39 @@ private Cursor computeNewLinedCursorElement(Cursor cursor) {
 
     private @Nullable Cursor alignsWith(Cursor cursor) {
         J cursorValue = cursor.getValue();
-        SourceFile sourceFile = cursor.firstEnclosing(SourceFile.class);
```
Jenson3210, line 141: looking at the parent cursor, it's value is not updated with any recipe's changes. 

So a recipe that introduces newlines, then calls AutoFormat (which will trigger this one), will not correctly calculate the value here

rewrite-java/src/test/java/org/openrewrite/java/format/SourcePositionServiceTest.java
```diff
@@ -202,4 +202,79 @@ public record Record6(
...
+                  return super.visitVariableDeclarations(multiVariable, ctx);
+              }
+          })),
+          java(
+            """
+            package com.example;
+
+            public class Test {
+                public void /* multiline comments can impact though */ example(
+            int p,
```
Jenson3210, line 351: The indentation here is intentionally weird as we want to verify that the service calculates based on the previous newlined indent + continuation rather than expecting the indentation is correct. The previous element should be correctly indented.

timtebeek: Perhaps this would be a helpful comment just above the text block?

Issue #5653, opened 2025-06-22
Collaboration request: extending on rewrite-docker


jimschubert:
Hi. I recently created https://github.com/jimschubert/rewrite-docker, which is available via jitpack. This project started as a pet project on my Twitch stream so I could learn ANTLR, but then I realized either the Docker parser might be difficult to do with ANTLR _or_ (more likely) I don't know enough ANTLR to write a Docker parser with it. So the project sat in a private repo until recently when I was on staycation and spent some time cleaning it up and writing a parser from scratch. I tried to match Docker's documentation closely for the parsed AST, and I'm able to parse everything I've tried (example: [MySQL 8.4 Dockerfile](https://github.com/docker-library/mysql/blob/1b9d0c10ad8569c9419d76aacec9dbab23c48e9e/8.4/Dockerfile.oracle)).

I shared the project on social media and Tim said I should create an issue here to discuss the possibility of collaboration. I discovered https://github.com/openrewrite/rewrite-docker about halfway through creating my rewrite-docker, but only looked at the official repo briefly because it appeared to only use regex for modifying the FROM instruction.

My library parses everything I've thrown at it so far. But I don't know if the AST is aligned with that of other rewrite libraries. I modeled loosely after rewrite-yaml with some ideas taken from rewrite-json.

Let me know if there's interest and what a collab would look like.

---

## Some parser details

I used the *RightPadded decorator from the example project, but I found it to be cumbersome to update, so pretty much any text in the AST is a [Literal](https://github.com/jimschubert/rewrite-docker/blob/8c0f418a32d9864cff28ea099eb03305f7fa7d19/src/main/java/com/github/jimschubert/rewrite/docker/tree/Docker.java#L87) which includes prefix and trailing "Space" (related: https://github.com/openrewrite/rewrite/issues/4668 - "Proposal: Improve whitespace handling"). The space can be whitespace or newline continuation characters, managed by the escape character `\`.

Docker supports _directive_ comments to allow tweaking the escape character, frontend, and a few other things. My library supports this dynamic escape character logic, but most of the other stuff is related to frontend. I don't raise any errors if the Dockerfile refers to another frontend which might invalidate the parsing logic. Directives parse into a `Directive` type and all other comments are a `Comment` type. Multiline comments are parsed as a single comment instruction.

[...]

PR #6203, opened 2025-10-27
Add Dockerfile parser based on ANTLR

timtebeek:
- Closes https://github.com/openrewrite/rewrite-docker/pull/21
- Closes https://github.com/openrewrite/rewrite/issues/5653


sambsnyd (approved):

timtebeek:
Naming things is hard; I'd been considering whether it should be:
- rewrite-dockerfile - more specific that this is only the parser, for now
- rewrite-containerfile - broader standard also compatible
- rewrite-docker - already exist as https://github.com/openrewrite/rewrite-docker/, but possibly merged here

Open to suggestions; mostly looking to not have to change this later, and not to conflict if possible.

PR #6204, opened 2025-10-28
Target matching Java versions after seeing IDE failures

timtebeek:
- Relates to https://github.com/openrewrite/rewrite/pull/6055

Historically, we'd been hesitant to create non-LTS parsers, so after 21 but before 25 we patched the Java 21 parser to support 21+; this has lead folks to now continue to use non-LTS versions and expect support, which isn't as tenable now that 25 is out as the next LTS, and all previous non-LTS versions are out of support.

Targeting Java 21 for the Java 25 parser helped us in the initial days after Java 25 came out to support running from those non-LTS versions still, as it took another week or so before SDKMan and Temurin Java 25 versions to become available. Now that those are out we can bump the requirements to use LTS versions.

We only aimed to support 22, 23 and 24 before 25 was out; going forward we have a better strategy with a new rewrite-java-next parser. I think it's fine to expect/require the next LTS for ingestion for at-this-moment unsupported non-LTS versions.


timtebeek:
Failure unrelated network issue; merging as discussed.

PR #6205, opened 2025-10-28
Upgrade to Gradle 9.1.0 running on Java 21

jkschneider:
## What's changed?

Consolidating on Java 25 so that we have a consistent development experience across recipe modules.

### DIFF
--- a/rewrite-test/build.gradle.kts
+++ b/rewrite-test/build.gradle.kts
@@ -17,5 +17,5 @@ dependencies {
 
     testImplementation(project(":rewrite-groovy"))
     testRuntimeOnly("org.antlr:antlr4-runtime:4.13.2")
-    testRuntimeOnly(project(":rewrite-java-21"))
+    testRuntimeOnly(project(":rewrite-java-25"))
 }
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -22,7 +22,6 @@ jobs:
     with:
       java_version: |
         25
-        21
```
shanman190, line 25: I think we'll need to keep 21 available for the Java toolchains to utilize for rewrite-groovy and rewrite-kotlin as otherwise, those versions are incompatible with JDK 25 at present.

rewrite-test/build.gradle.kts
```diff
@@ -17,5 +17,5 @@ dependencies {
 
     testImplementation(project(":rewrite-groovy"))
     testRuntimeOnly("org.antlr:antlr4-runtime:4.13.2")
-    testRuntimeOnly(project(":rewrite-java-21"))
+    testRuntimeOnly(project(":rewrite-java-25"))
```
timtebeek, line 20: Some downstream projects rely on this transitive dependency for their Java parser. I wonder what using 25 here will do for their tests when those concern using gradle, but I'll guess we'll see over the next hours and otherwise change this back.

PR #6208, opened 2025-10-28
Added MethodDeclaration wrapping options

Jenson3210:
Screenshot 2025-10-28 at 16 51 38

## What's changed?
Support all the options for Wrapping and Braces present for method declaration parameters

## What's your motivation?
Need to enhance the formatting features

Ran with cli -> no failures, mainly due to the try catches present. 
The SourcePositionService has a new method that calculates the "header" of a declaration (variabledecl, methodDecl and class decl. It expects the annotations to be on the correct line already and will not count new lines in the middle of a declaration (to identify if we could "fold" back a declaration to a single line / split it over multiple lines.

### DIFF
--- a/rewrite-java/src/test/java/org/openrewrite/java/format/SourcePositionServiceTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/format/SourcePositionServiceTest.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.Cursor;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.service.SourcePositionService;
 import org.openrewrite.java.service.Span;
@@ -574,6 +575,63 @@ class SourcePositionServiceTest implements RewriteTest {
 
               public class Test {
                   public void method1() {
+                      //Some comment
+                      int x = 1;
+                  }
+
+                  public void method2() {
+                      int x = 1;
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void updatingSomeElementCanImpactPositions() {
+        rewriteRun(
+          spec -> spec.recipe(RewriteTest.toRecipe(() -> new JavaIsoVisitor<>() {
+
+              @Nullable
+              SourcePositionService service;
+
+              @Override
+              public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+                  service = cu.service(SourcePositionService.class);
+                  return super.visitCompilationUnit(cu, ctx);
+              }
+
+              @Override
+              public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+                  //minimize the first method (all declarations should have an updated position)
+                  super.visitClassDeclaration(classDecl.withBody(classDecl.getBody().withStatements(ListUtils.mapFirst(classDecl.getBody().getStatements(), SourcePositionServiceTest::minimize))), ctx);
+                  return classDecl;
+              }
+
+              @Override
+              public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                  if ("method1".equals(method.getSimpleName())) {
+                      //This is the modified method, so doing getCursor is enough here.
+                      assertThat(service.positionOf(getCursor())).isEqualTo(Span.builder().startLine(4).startColumn(5).endLine(7).maxColumn(40).endColumn(6).build());
+                  } else if ("method2".equals(method.getSimpleName())) {
+                      //The parent cursor (= block of class declaration) contains the modified method1 and in order to be used by the position calculation, you must pass that one.
+                      assertThat(service.positionOf(getCursor().getParentTreeCursor(), method)).isEqualTo(Span.builder().startLine(9).startColumn(5).endLine(11).maxColumn(28).endColumn(6).build());
+                      //Just passing this cursor is not enough for the accurate positioning as the modified element in not in this cursor and the service will start top down from JavaSourceFile -> no updated method is used.
+                      assertThat(service.positionOf(getCursor(), method)).isEqualTo(Span.builder().startLine(12).startColumn(5).endLine(14).maxColumn(28).endColumn(6).build());
+                  }
+                  return super.visitMethodDeclaration(method, ctx);
+              }
+          })),
+          java(
+            """
+              package com.example;
+
+              public class Test {
+                  public void method1(
+                    int a,
+                    int b
+                  ) {
                       //Some comment
                       int x = 1;
                   }


rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -125,6 +128,67 @@ public int computeTreeLength(Cursor cursor) {
         throw new RuntimeException("Unable to calculate length due to unexpected cursor value: " + newLinedElementCursor.getValue().getClass());
     }
 
+    public int computeDeclarationLength(Cursor cursor) {
```
sambsnyd: I'm not sure if I understand what is going on in this method. Could probably use an explanatory comment or two describing the intent and appropriate usage. Is the idea to compute the distance between the argument list and the beginning of whatever line the argument list appears on?

When I have wanted to compute the textual length of any LST element in the past I've called `print()`, looked at the size, and that alone has gotten the job done. This seems much more involved but I don't quite understand why. Is all of the complexity just to account for annotations potentially being on the same line or not? I would probably apply a visitor that removes all annotations or other elements _not_ contributing to the relevant length, then print as normal.

Jenson3210: So the goal here is to calculate line length of a declaration. 
The other method introduced for method chains was counting length of an entire lst element. These declarations typically have a block which we do not want to count / print. 
That's what this method tries to avoid. 
It will also not count annotations which are on their own line, but it will count annotations on return type "inside" the declaration. 

`    private static final class A {` would be the easiest example of the values that we count to see if we could/should wrap a declaration over multiple lines or not. The way how this wrapping happens is not in this services, but in the visitors with the wrapping purpose. 

```java
        @Bean //<-- not counted
        @Qualifier("named") //<-- not counted
        public @Nullable BeanClass createBean(Service a, OtherService b) { //<-- indent + line counted
            ...
        }
```
This is a bigger example. I wrote this method to count the "header" of a declaration to be able to know if we should wrap the implements,extends (classes/interfaces), variables (methods), primary constructor(Records)...

Jenson3210: When I used `print` in previous recipes, @timtebeek would typically mention that we would try to avoid using that one. 
Also, that uses the `Default` Marker printer which will print markers which will then be counted in the length.

Jenson3210: As we span multiple lst elements, It would be a lot of `with...` calls to make these elements into printable lst elements which print something to count the length of the objects inbetween, I guess.

sambsnyd: `SourcePositionService` is part of our public API so it's always worth considering whether we're putting out there an API which has a robust interface.
From an API design perspective, `computeDeclarationLength` strikes me as highly and unnecessarily specific.
Wanting to know what column number an element begins (or ends) at is a more general question which could be asked of any LST element.
I lean towards naming the method something like `column()` or `startColumn()` and then striving to make the implementation more generic.

As for making the implementation more terse and generic, I'm pretty sure we can determine the start column of any element with a visitor that overrides only `visitSpace()`, since all we need to do is count columns resetting the count every time we see a `\n` until we discover the relevant element. 

Then there's no need to muck around with anything specific to annotations or method declarations or whatever

Jenson3210:
tests failing is issue unrelated

rewrite-java/src/main/java/org/openrewrite/java/format/WrapMethodDeclarationParameters.java
```diff
@@ -0,0 +1,180 @@
...
+        } catch (NoSuchMethodError | NoSuchFieldError ignore) {
+            // Styles are parent-first loaded and this can happen if the style is from a older version of the runtime. Can be removed in future releases.
+        }
+
+        return m;
+    }
+
+    public Cursor minimize(Cursor cursor) {
+        if (cursor.getValue() instanceof J) {
+            InMemoryExecutionContext ctx = new InMemoryExecutionContext();
```
sambsnyd: We shouldn't be instantiating execution context within a recipe. Only an execution context passed into the recipe run will have the appropriate network configuration / paths to use for caches and temporary files / etc.

In this case it doesn't have any repercussions, but I don't want people (or LLMs) seeing it and thinking "I can do that, too" and producing recipes that _do_ have a problem.

Either pass in a real execution context or use a visitor with a type parameter set to something inconsequential like `Integer`.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -125,6 +132,42 @@ public int computeTreeLength(Cursor cursor) {
...
+    }
+
+    /**
+     * Computes the position span of a J element.
+     *
+     * @see #positionOfChild(Cursor, Object)
+     */
+    public Span positionOf(Cursor cursor, J child) {
+        return positionOfChild(cursor, child);
+    }
```
sambsnyd, line 166: Why are these overloads necessary? Cursors can point to padding. So either the cursor points to a `JRightPadded`, `JContainer`, or regular element.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -223,4 +266,237 @@ private Cursor computeNewLinedCursorElement(Cursor cursor) {
...
+
+    /**
+     * Computes the position span of a child element within the source code by printing the source
+     * file and tracking line/column positions until the target element is reached.
+     * <p>
+     * This method uses a custom {@link JavaPrinter} to traverse the AST from the source file root,
+     * tracking line and column positions as it prints. When the target child element is encountered,
+     * it marks the position and calculates the span including:
+     * <ul>
+     *   <li>Start line and column (1-indexed for line, 0-indexed for column)</li>
```
sambsnyd: I believe most tools 1-index column numbers as well.  IntelliJ does. Note how it cites the very beginning of this empty line as "1":

Image

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -223,4 +266,237 @@ private Cursor computeNewLinedCursorElement(Cursor cursor) {
...
+     * containers, and {@link JRightPadded} wrapped elements.
+     *
+     * @param cursor the cursor providing context for the position calculation, must point to or contain
+     *               a {@link JavaSourceFile} ancestor for printing
+     * @param child  the child element to locate (can be same as cursor element) or any child({@link J} element, {@link JContainer} or {@link JRightPadded})
+     * @return a {@link Span} containing the computed position information
+     * @throws IllegalArgumentException if the child is a raw {@link Collection} (use padding accessor methods instead),
+     *                                  if the child type is not supported, or if the child cannot be found in the source file
+     */
+    private Span positionOfChild(Cursor cursor, Object child) {
```
sambsnyd, line 290: The cursor can point to the "child" element directly, so I'm not sure why we need a second parameter for it.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
```diff
@@ -223,4 +266,237 @@ private Cursor computeNewLinedCursorElement(Cursor cursor) {
...
+                if (tree == foundJ) {
+                    found.set(true);
+                }
+                return super.postVisit(tree, p);
+            }
+
+            private J startFind(J tree) {
+                return tree
+                        .withPrefix(tree.getPrefix()
+                                .withWhitespace(tree.getPrefix().getWhitespace().replace("\n", "\n" + START_FIND))
```
sambsnyd: So if there are multiple newlines in the prefix the START_FIND comment will be added multiple times. 
And it is being placed into the whitespace part of the space - which really is supposed to be _only_ whitespace - despite being a comment. I suppose it's not too problematic since this is a temporary thing which is thrown away at the end, but again don't want people/LLMs thinking it's okay to put non-whitespace into whitespace.

rewrite-java/src/test/java/org/openrewrite/java/format/AutoFormatTest.java
```diff
@@ -2718,4 +2718,784 @@ class Test1 {
...
+                  package com.example;
+
+                  class Test {
+                      void method(
+                              String name,
+                              int age,
+                              boolean active) {
+                      }
+                  }
+                  """
```
sambsnyd, line 2854: A test which asserts no change should only have a `before`.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: Imagine you want to find the position deeply inside an object you modified (eg. the minimization in the Wrapping minimizes the `MethodDeclaration` and asks the position of the parameters `JCollection`. )

If we would have the same cursor for this, that would not be possible. As we first go up to the JavaSourceFile then go down, we would start printing the unminimized MethodDeclaration and then find the parameters at another location than they are in the minimized variant.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: These multiple START_FIND comments are indeed removed as we only use the last one. 
I can come up with some mechanism to circumvent this.

rewrite-java/src/test/java/org/openrewrite/java/format/AutoFormatTest.java
Jenson3210: There is a change in this case as we are adding the style marker as part of the AutoFormat recipe when the style marker is not on the sourceFile (which it is not in these unit tests)

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: I could also make the below method public instead (as that method will fail on any other value), but I figured as this is a public API, we want to be as clear as possible on what is allowed and how this method will behave.

rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
Jenson3210: I now print the prefix without printing active, then call the normal visitor with an overridden Empty prefix. This also allows for even less code in the String handling as the printed value will now really only be the content of the searched element.

sambsnyd (approved):

PR #6212, opened 2025-10-29
Use Kotlin plugin version 2.1 but target 1.9

timtebeek:
- Follows on from #6205 after failing to to run `:rewrite-kotlin:generatePomFileForNebulaPublication` due to `NoSuchMethodError: 'org.gradle.api.Project org.gradle.api.artifacts.ProjectDependency.getDependencyProject()'`

### DIFF
--- a/rewrite-kotlin/build.gradle.kts
+++ b/rewrite-kotlin/build.gradle.kts
@@ -15,7 +15,6 @@ dependencies {
 
     implementation(project(":rewrite-java"))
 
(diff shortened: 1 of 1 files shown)


timtebeek:
Quick reference to the docs used:
- https://kotlinlang.org/docs/gradle-compiler-options.html#centralize-compiler-options-and-use-types
- https://kotlinlang.org/docs/gradle-compiler-options.html#example-of-setting-languageversion

PR #6213, opened 2025-10-29
JavaScript - handling `function()` as expressions in autoformatting

greg-at-moderne:
## What's changed?

Another amendment to JavaScript autoformatting. Handling cases where `function() { ...}` is an expression, not a standalone statement on its own.

## What's your motivation?

Handle more real-life JS/TS code.

### DIFF
--- a/rewrite-javascript/rewrite/test/javascript/format/format.test.ts
+++ b/rewrite-javascript/rewrite/test/javascript/format/format.test.ts
@@ -186,4 +186,22 @@ describe('AutoformatVisitor', () => {
                  `import {delta, gamma} from 'delta.js'`)
             // @formatter:on
         )});
+
+    test('anonymous function expression', () => {
+        return spec.rewriteRun(
+            // @formatter:off
+            //language=typescript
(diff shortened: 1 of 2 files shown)


rewrite-javascript/rewrite/test/javascript/format/format.test.ts
```diff
@@ -185,4 +186,22 @@ describe('AutoformatVisitor', () => {
...
+
+    test('anonymous function expression', () => {
+        return spec.rewriteRun(
+            // @formatter:off
+            //language=typescript
+            typescript(
+                `const fn = function () {return 99;};`,
+                 // TODO the space after `const fn=` is excessive
+                 `
+                const fn =
```
knutwannheden, line 198: If we can get rid of that trailing space on this line, that would also be nice. I think we have a setting that requires spaces around `=`, but I guess a `\n` should also be accepted.

greg-at-moderne: #6215

PR #6216, opened 2025-10-29
Use a 9.1.0 connector for Gradle when no version is specified

timtebeek:
We're seeing some issues downstream when running on Java 25; seeing if this factors in as we explore.


Laurens-W:
Hmm still the same amount of failed tests

timtebeek:
Thanks; yes just did a new release of rewrite-build-gradle-plugin; turns out we were prescribing Java 25 toolchain usage:
https://github.com/openrewrite/rewrite-build-gradle-plugin/releases/tag/v2.6.6

That's been revert and we also now by default no longer have Java 25 available at all, so we should see builds return to normal.

PR #6224, opened 2025-10-30
Add recipe for migrating Gradle dependency declarations to the version catalog

bmuschko:
## What's changed?

Introduces a new recipe for migrating Gradle dependency declarations to the [version catalog](https://docs.gradle.org/current/userguide/version_catalogs.html).

## What's your motivation?

The version catalog is the idiomatic feature in Gradle to declare dependencies in one central location. End users do not have to come up with their unique and sometimes creative way to define dependencies and their versions anymore, e.g. gradle.properties, extra properties, etc.

This recipe inspects dependency declarations defined by the Groovy or Kotlin DSL. As of now, it only derives the version declaration directly from the dependency definition as an inline string or as a reference from a property defined in gradle.properties. Other locations, e.g. extra properties in the build script are not supported.

No


rewrite-gradle/src/test/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalogGroovyDSLTest.java
```diff
@@ -0,0 +1,403 @@
...
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.beforeRecipe(withToolingApi())
+          .recipe(new MigrateDependenciesToVersionCatalog());
+    }
+
+    @Test
+    void doesNotProcessFilesExcludedByPreconditions() {
+        rewriteRun(
+          spec -> spec.expectedCyclesThatMakeChanges(0),
```
sambsnyd:
```suggestion
```

It isn't necessary to set the cycle count low to test that a recipe does not make changes.

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
```diff
@@ -0,0 +1,1164 @@
...
+                " * Migrate version properties from `gradle.properties` to the version catalog\n" +
+                " * Preserve project dependencies unchanged\n\n" +
+                "**Note:** If a version catalog already exists, the recipe will not modify it.";
+    }
+
+    static class DependencyAccumulator {
+        final Map<String, DependencyInfo> dependencies = synchronizedMap(new LinkedHashMap<>());
+        final Set<String> configurations = synchronizedSet(new LinkedHashSet<>());
+        final Set<String> propertyNamesToRemove = synchronizedSet(new LinkedHashSet<>());
+        final Map<String, String> propertyValues = synchronizedMap(new LinkedHashMap<>());
```
sambsnyd: This synchronization shouldn't be necessary, recipe execution is single threaded.

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
```diff
@@ -0,0 +1,1164 @@
...
+
+    static class DependencyCoordinates {
+        String group;
+        String artifact;
+        String version;
+
+        boolean isComplete() {
+            return group != null && artifact != null && version != null;
+        }
+    }
```
sambsnyd, line 100: The `GroupArtifactVersion` class already exists and can be used here

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
```diff
@@ -0,0 +1,1164 @@
...
+                                        for (Expression arg : m.getArguments()) {
+                                            extractDependency(arg, methodName, acc);
+                                        }
+                                    }
+                                }
+                            }
+
+                            return m;
+                        }
+                    }.visitNonNull(sourceFile, ctx);
```
sambsnyd: Most of the recipe beyond this point is duplicating functionality built into the `GradleDependency` trait. Try using that instead.

rewrite-gradle/src/test/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalogGroovyDSLTest.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/MigrateDependenciesToVersionCatalog.java
bmuschko: `GroupArtifactVersion` is an immutable class. The values for fields in `DependencyCoordinate` are populated after instantiating the class. So this wouldn't work without either 1) making `GroupArtifactVersion` mutable or 2) introducing a lot of local fields.

sambsnyd (approved):

PR #6227, opened 2025-10-31
Fix Gradle 9 compatibility by using explicit Gradle 6.1.1 dependencies

knutwannheden:
Gradle 9 embeds Groovy 4.0.28, while RewriteGradleProject.groovy is compiled against Gradle 6.1.1 (Groovy 3.x). The Groovy version mismatch prevented proper type attribution during parsing.

**Solution:**
1. Replace `localGroovy()` with explicit `groovy-all:3.0.25` to maintain Groovy 3.x consistency
2. Replace `gradleApi()` with explicit Gradle 6.1.1 module dependencies to avoid pulling in Gradle 9's API
3. Add `gradle-tooling-api:6.1.1` for GradleConnector support without Gradle 9's full API
4. Change `rewrite-gradle-tooling-model/model` to use `compileOnly(gradleApi())` to prevent leaking Gradle 9 API

### DIFF
--- a/rewrite-gradle-tooling-model/model/build.gradle.kts
+++ b/rewrite-gradle-tooling-model/model/build.gradle.kts
@@ -13,7 +13,9 @@ dependencies {
         // last version which supports java 8, which testGradle4 runs on
         testImplementation("org.assertj:assertj-core:3.27.4!!")
     }
-    implementation(gradleApi())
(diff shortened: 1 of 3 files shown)


timtebeek:
I don't think we have any remaining Gradle 9 compatibility issues at the moment, also considering the recent Groovy 4:
- https://github.com/openrewrite/rewrite/pull/6228

PR #6232, opened 2025-10-31
Convert JaCoCo config when used as method invocation

timtebeek:


rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
```diff
@@ -458,4 +459,199 @@ void alsoWorkOnKotlinDsl() {
...
+        void deprecationsInMethodInvocationSyntax() {
+            rewriteRun(
+              buildGradle(
+                """
+                  plugins {
+                      id "java"
+                      id "jacoco"
+                  }
+
+                  jacocoTestReport {
```
bmuschko, line 475: We should probably cover all possible notations available to Groovy here.

```
jacocoTestReport {
    reports {
       xml.enabled = true
       xml.setEnabled(true)
       xml.setEnabled true
       xml.enabled(true)
       xml.enabled true
    }
}
```

It would probably make sense long-term to create a trait for parsing property value assignments for the Groovy and Kotlin DSL.

rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
```diff
@@ -458,4 +459,199 @@ void alsoWorkOnKotlinDsl() {
           )
         );
     }
+
+    @Nested
+    class AsMethodInvocation {
```
bmuschko, line 464: It's likely a good idea to have the same test case for the Kotlin DSL as well.

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
```diff
@@ -82,6 +114,22 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
 
+            private boolean isReportType(String name) {
+                return "xml".equalsIgnoreCase(name) ||
+                       "csv".equalsIgnoreCase(name) ||
+                       "html".equalsIgnoreCase(name);
+            }
+
+            private J.MethodInvocation replaceDeprecatedMethodName(J.MethodInvocation method) {
+                String methodName = method.getSimpleName();
+                if ("enabled".equalsIgnoreCase(methodName) || "isEnabled".equalsIgnoreCase(methodName)) {
```
bmuschko: Shouldn't this be `setEnabled` instead of `isEnabled`?

bmuschko (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
timtebeek: Hmm for the Kotlin cases it does appear to be `isEnabled`, as seen here:
https://docs.gradle.org/current/userguide/jacoco_plugin.html#default_values_of_the_jacoco_task_extension

Also based on this earlier unchanged test:
https://github.com/openrewrite/rewrite/blob/b72785ecc4c78c1d2c3a4c48cba09197d9f1c88c/rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java#L428-L461

PR #6233, opened 2025-10-31
Update rendering logic to ensure 100% match in official Gradle wrapper scripts

shanman190:
## What's changed?
Generated wrappers locally for all released Gradle versions, then went through a systematic refactor comparing the official `gradlew` and `gradlew.bat` scripts to those that we were generating here until we had achieves a 100% success rate for all Gradle wrappers (exclusions: nightlies and Gradle 5.1-rc-1 and 5.1-rc-2 had broken wrapper jars resulting in those not generating).

## What's your motivation?
We recently started to prerender the scripts after which it was noticed that in certain situations values such as `DEFAULT_JVM_OPTS` would have erroneous values.

N/A

@timtebeek 

N/A

## Any additional context
N/A

### Checklist


rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/GradleWrapperScriptLoader.java
```diff
@@ -25,15 +25,16 @@
...
 import java.util.NavigableMap;
 import java.util.TreeMap;
 
 import static java.util.Objects.requireNonNull;
 
 public class GradleWrapperScriptLoader {
     @Getter
-    private final NavigableMap<String, Version> allVersions = new TreeMap<>(
-            new LatestRelease(null));
+    private final Map<String, Version> allVersions = new HashMap<>();
```
shanman190, line 37: Switched the implementation here because of the following:

* `LatestRelease` considers milestones and release candidates to be equivalent which results in the `TreeMap` excluding all but the first one from the data set
* We no longer have gaps with the exception of 5.1-rc-1 and 5.1-rc-2 in the data set. All Gradle wrappers are present and we have a workflow that updates for new wrappers weekly.

sambsnyd (approved):

rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/unix/dbb485b7.txt
```diff
@@ -65,7 +65,7 @@ cd "`dirname \"$PRG\"`/"
 APP_HOME="`pwd -P`"
 cd "$SAVED"
 
-CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
+CLASSPATH=$APP_HOME/gradle\wrapper\gradle-wrapper.jar
```
timtebeek, line 68: This seems odd for a unix classpath entry; was it also that way in the reference?

shanman190: Yes, this is actually what was used for `1.0-milestone-5`, `1.0-milestone-6`, and `1.0-milestone-7`.

PR #6240, opened 2025-11-02
JavaScript: Templating engine enhancements

knutwannheden:
### 1. Variadic Captures
Capture zero or more elements in sequences with optional min/max constraints. Access the individual elements in templates.
```typescript
const args = capture('args', { variadic: true });
const pat = pattern`foo(${args})`;  // Matches: foo(), foo(a), foo(a,b,c)
const tmpl = template`bar(${args[0]}, ${args.slice(1)})`;  // Reorder/slice arguments

const match = await pat.match(method);
return match && await tmpl.apply(this.cursor, method, match) || method;
```

### 2. Capture Property Access

Navigate properties of captured nodes in templates using dot notation.
```typescript
const method = capture('method');
template`console.log('${method.name.simpleName}')`;  // Access nested properties
```

### 3. Capture Constraints

Validate captured nodes with predicate functions. Regular captures receive single nodes, variadic captures receive arrays.
```typescript
const value = capture<J.Literal>('value', {
  constraint: (node) => typeof node.value === 'number' && node.value > 100
});
const args = capture<J.Literal>('args', {
  variadic: true,
  constraint: (nodes) => nodes.every(n => typeof n.value === 'number')
});
```

### 4. Non-Capturing Matches with any()

Match AST structure without binding values. Use when you need to validate structure but don't need access to the matched nodes.
```typescript
// Match method calls regardless of arguments
const pat = pattern`foo(${any()})`;  // Clearer than capture('ignored')

// Mix captures and any() - only bind what you need
const body = capture();
const pat2 = pattern`if (${any()}) { ${body} }`;  // Match any condition, capture body

// Variadic any() for remaining arguments
const first = capture();
const rest = any({ variadic: true });
const pat3 = pattern`bar(${first}, ${rest})`;  // Only first arg is captured
```

### 5. Builder API

Construct patterns and templates programmatically when structure isn't known at compile time.
```typescript
const pat = Pattern.builder()
  .code('foo(')
  .capture(capture('arg'))
  .code(')')
  .build();
```

### 6. Lenient Type Matching

Patterns without type annotations match typed code by default (configurable via lenientTypeMatching option).
```typescript
pattern`(props, ref) => body`  // Matches: (props: Props, ref: Ref) => body
```

### 7. Context Statements

Provide type attribution context via `context` option (renamed from `imports`).
```typescript
pattern`forwardRef(${arg})`.configure({
  context: [`import { forwardRef } from 'react'`, 'const c = 1;']
});
```

### 8. Simplified `capture()` API

When creating a capture with options the name is now part of that construct rather than making it required.
```typescript
const args = capture({
    name: 'args',
    variadic: { separator: ' | ', min: 2, max: 5 }
});
```

### 9. New `param()` Function

The `param()` name makes more sense when having a plain `template` without `pattern`.
```typescript
const value = param('value');
template`return ${value} * 2`;
```


rewrite-javascript/rewrite/src/javascript/comparator.ts
```diff
@@ -4761,12 +4761,33 @@ export class JavaScriptComparatorVisitor extends JavaScriptVisitor<J> {
...
+
     /**
      * Checks if two types are semantically equal.
      * For method types, this checks that the declaring type and method name match.
+     * With lenient type matching, undefined types on either side are considered compatible.
      */
     private isOfType(target?: Type, source?: Type): boolean {
         if (!target || !source) {
+            // Lenient mode: if either type is undefined, allow the match
+            if (this.lenientTypeMatching && (!target || !source)) {
```
greg-at-moderne: The `!target || !source` check here is redundant over the line above, right?

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+                override async visitLiteral(literal: J.Literal, p: any): Promise<J | undefined> {
+                    if (literal.valueSource === '1') {
+                        return builderTmpl.apply(this.cursor, literal);
+                    }
+                    return literal;
+                }
+            });
+
+            return spec.rewriteRun(
+                typescript('const a = 1', 'const a = const result = 42'),
```
greg-at-moderne: The "after" part of the test doesn't make sense.
Shouldn't we replace the test with something more realistic?

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        test('creates template equivalent to template literal', async () => {
+            // Using builder
+            const builderTmpl = Template.builder()
+                .code('const result = ')
+                .param(42)
+                .code(';')
+                .build();
+
+            // Using template literal
+            const literalTmpl = template`const result = ${42};`;
```
greg-at-moderne: This is not used anywhere, is it?

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+            function createWrapper(innerBody: string): Template {
+                return Template.builder()
+                    .code('function wrapper() { try { ')
+                    .code(innerBody)
+                    .code(' } catch(e) { console.error(e); } }')
+                    .build();
+            }
+
+            const tmpl = createWrapper('return 42;');
+            expect(tmpl).toBeInstanceOf(Template);
```
greg-at-moderne, line 138: In the "then" part of the test shouldn't we assert more here? I.e. the content of the whole source code produced?

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        });
+
+        test('creates pattern equivalent to pattern literal', async () => {
+            const left = capture('left');
+            const right = capture('right');
+
+            // Using builder
+            const builderPat = Pattern.builder()
+                .capture(left)
+                .code(' + ')
```
greg-at-moderne: Would probably be interesting to have the test not use spaces, to show the matching is not textual.

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+                .capture(right)
+                .build();
+
+            // Using pattern literal
+            const literalPat = pattern`${left} + ${right}`;
```
greg-at-moderne: Not used anywhere?

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+            const builder = Pattern.builder().code('myFunction(');
+            const argCount = 3;
+            for (let i = 0; i < argCount; i++) {
+                if (i > 0) builder.code(', ');
+                builder.capture(capture(`arg${i}`));
+            }
+            builder.code(')');
+
+            const pat = builder.build();
+            expect(pat).toBeInstanceOf(Pattern);
```
greg-at-moderne, line 218: Similarly as before, probably worth to assert more here. E.g. give it a piece of TS code and expect it finds the 3-arg method invocation.

rewrite-javascript/rewrite/test/javascript/templating/builder-api.test.ts
```diff
@@ -0,0 +1,344 @@
...
+            const builder = Pattern.builder().code('foo(');
+            builder.capture(capture('first'));
+            if (needsSecondArg) {
+                builder.code(', ').capture(capture('second'));
+            }
+            builder.code(')');
+
+            const pat = builder.build();
+            expect(pat).toBeInstanceOf(Pattern);
+        });
```
greg-at-moderne: Not sure what's the point of these conditional tests. The condition is evaluated in the test code, so not sure what is it testing?

rewrite-javascript/rewrite/test/javascript/templating/capture-constraints.test.ts
```diff
@@ -0,0 +1,295 @@
...
+            // 202: > 50, > 200, even, not divisible by 10 ✓
+            const match2 = await pat.match(await parseExpression('202'));
+            expect(match2).toBeDefined();
+
+            // 60: > 50, even, but divisible by 10 ✗
+            const match3 = await pat.match(await parseExpression('60'));
+            expect(match3).toBeUndefined();
+
+            // 45: not > 50 ✗
+            const match4 = await pat.match(await parseExpression('45'));
```
greg-at-moderne: Minor thing, this is overly verbose.
Probably worth simplifying these 4 assertions to 4 one-liners, with no comments, just the 
`expect(await pat.match(await parseExpression('60'))).toBeUndefined();`

rewrite-javascript/rewrite/test/javascript/templating/param-function.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    test('param() with unnamed parameter', async () => {
+        const value = param(); // Unnamed
+        const tmpl = template`${value} + 1`;
+
+        spec.recipe = fromVisitor(new class extends JavaScriptVisitor<any> {
+            override async visitLiteral(literal: J.Literal, p: any): Promise<J | undefined> {
+                if (literal.valueSource === '7') {
+                    return tmpl.apply(this.cursor, literal, new Map([[value.getName(), literal]]));
+                }
```
greg-at-moderne, line 76: BTW, does it handle paranthezing when needed?

knutwannheden: No, that would still be a feature to add. In `JavaTemplate` we didn't have that for the longest time either: https://github.com/openrewrite/rewrite/pull/5279

I think we should however look into this in the future.

PR #6246, opened 2025-11-03
Expanding TypeMatcher to actually use the matcher for `JavaType.Primitive`s. `FindFieldsOfType` should properly match against primitives now.

steve-aom-elliott:
## What's changed?
- Previously when you tried to use `FindFieldsOfType` for `boolean` or similar primitive types that aren't considered `JavaType.FullyQualified`, it would not end up marking the fields. This change updates the `hasElementType` method to include `JavaType.Primitive`s and allows the `matches(..)` method in `TypeMatcher` to try to use the matcher it configured on `JavaType.Primitive` instances.

## What's your motivation?
It was making it impossible to find fields of primitive (non-object) types.

The change in the `TypeMatcher`, I'm unsure if I should be opting for something safer / more thorough from the `JavaType.Primitive` methods or if this is fine.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/TypeMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/TypeMatcher.java
@@ -97,6 +97,10 @@ public class TypeMatcher implements Reference.Matcher {
             );
         }
 
+        if (type instanceof JavaType.Primitive && typeNameMatcher.matches(type.toString())) {
+            return true;
+        }
+
         return false;
     }
 
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/FindFieldsOfType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/FindFieldsOfType.java
@@ -139,7 +139,7 @@ public class FindFieldsOfType extends Recipe {
                                           boolean matchOverrides) {
         if (type instanceof JavaType.Array) {
             return hasElementType(((JavaType.Array) type).getElemType(), fullyQualifiedName, matchOverrides);
-        } else if (type instanceof JavaType.FullyQualified) {
+        } else if (type instanceof JavaType.FullyQualified || type instanceof JavaType.Primitive) {
             return new TypeMatcher(fullyQualifiedName, matchOverrides).matches(type);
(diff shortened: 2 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/TypeMatcher.java
```diff
@@ -97,6 +97,10 @@ public boolean matches(@Nullable JavaType type) {
             );
         }
 
+        if (type instanceof JavaType.Primitive && typeNameMatcher.matches(type.toString())) {
```
steve-aom-elliott, line 100: This is the the check where I'm not sure if I should be trying to use something more thorough (also this statement can probably be collapsed down into a ternary with the lines below it following review)

rewrite-java/src/main/java/org/openrewrite/java/search/FindFieldsOfType.java
```diff
@@ -139,7 +139,7 @@ private static boolean hasElementType(@Nullable JavaType type, String fullyQuali
                                           boolean matchOverrides) {
         if (type instanceof JavaType.Array) {
             return hasElementType(((JavaType.Array) type).getElemType(), fullyQualifiedName, matchOverrides);
-        } else if (type instanceof JavaType.FullyQualified) {
+        } else if (type instanceof JavaType.FullyQualified || type instanceof JavaType.Primitive) {
```
timtebeek, line 142: Instead of adjusting TypeMatcher (potentially larger more impactful change) could we change the matching here as we already do for Arrays and GenericTypeVariables ? So something like
```
        } else if (type instanceof JavaType.Primitive) {
            return fullyQualifiedName.equals(((JavaType.Primitive) type).getKeyword());
```

steve-aom-elliott: For this iteration, yes, we can reduce the scope of the change to fix this one recipe, but more broadly, we should probably correct the fact that `TypeMatcher` creates an `ExactTypeNameMatcher` but never uses it for trying to match when you're dealing with a primitive type like `boolean`.

PR #6248, opened 2025-11-04
Gradle `UpgradePluginVersion` doesn't update the model (`GradleProject`)

BoykoAlex:
Failure to add xml bind dependency via Rewrite Gradle plugin for Spring Boot 2.x -> 3.x upgrade

- https://github.com/openrewrite/rewrite/issues/6411


timtebeek:
Thanks for the runnable report! Did you already explore why this might be occurring as well?

BoykoAlex:
I was only able to see when this got broken but I wasn't able to see what exactly is broken... Still investigating... 

I've created this PR to run tests via CI... I'm having a lot of issues running tests that make gradle tooling api connections on my machine... I constantly run into some lock issues in gradle cache or class file version issues...

BoykoAlex:
Let me explain a bit what I'm trying to fix.
I'm upgrading spring-petclinic 2.7.x to 3.0.x. The upgrade process should add the dreaded XML Bind API dependency. It works fine for Maven but for Gradle it does not add the dependency. It used to work fine before Rewrtite Gradle Plugin 7.9.0 and `rewrite-spring` 6.9.0. The last version combination where the dependency was added is gradle plugin 7.8.0 and `rewrite-spring` 6.8.2.
I'm sure the issue is in `rewrite-gradle` 8.55.2 and above... I'm thinking that after upgrading gradle plugin version the dependencies graph isn't updated hence the recipe doesn't know the XML Bind API dependency needs to be added. The dependencies graph might not be updated due to the 3.0.x boot plugin not working with 7.x gradle (???). Investigating.
@timtebeek feel free to give a direction to my investigation if you feel I'm going off the rails ;-)

BoykoAlex:
@timtebeek this is the change in 8.55.2 that breaks the 8.55.1 behaviour: https://github.com/openrewrite/rewrite/compare/v8.55.1...v8.55.2#diff-743cf2f048fdba39228e626dd60acbdf76b3f256061b826bffd26106e95338abL241

I've ported that change in the PR. However, I think what the code was doing in `main` is correct... I think the actual problem is that after updating gradle plugin version (boot plugin 2.7.x -> 3.0.x) the `GradleProject` marker isn't updated. The `MavenResolutionResult` is updated if dependency management bit changes via a recipe. The Maven marker is re-resolved "manually", however, I don't see this in the Gradle variant and not sure if this is possible at all... It's okay for something small like a dependency add/change/remove and this is implemented but for a plugin (i.e. dependency management) I'd involve the gradle to do the job somehow.

timtebeek:
Thanks for digging in! Let me tag @shanman190 for his insights here while I'm out.

rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleConfigurationFilter.java
```diff
@@ -59,7 +59,7 @@ public void removeConfigurationsContainingTransitiveDependency(GroupArtifact dep
             GradleDependencyConfiguration gdc = requireNonNull(gradleProject.getConfiguration(tmpConfiguration));
             for (GradleDependencyConfiguration transitive : gradleProject.configurationsExtendingFrom(gdc, true)) {
                 if (transitive.findResolvedDependency(dependency.getGroupId(), dependency.getArtifactId()) != null) {
-                    filteredConfigurations.remove(tmpConfiguration);
+                    filteredConfigurations.remove(transitive.getName());
```
timtebeek, line 62: @shanman190 should we also still remove `tmpConfiguration`?

Or only remove `tmpConfiguration` as before with a `continue` to the next `tmpConfiguration` in the outer loop?

shanman190: So I'm not sure this logic here is actually necessary. It was something that I meant to come back and check in on. `tmpConfigurations` comes from `filteredConfigurations` which are those that we desire to potentially add the configuration to. So in the cases where a transitive configuration includes the requested dependency, I think we want to be removing the `tmpConfiguration` rather than the `transitive.getName()`.

BoykoAlex: This change in the PR is just making the test work. This is what has changed in 8.55.2.
I also agree that it should be `tmpConfiguration` rather than `transitive.getName()`. Therefore, to make the test pass I think the `GradleProject` model should have been updated after the run of `UpgradePluginVersion` to calculate new transitive dependencies as new transitive dependencies wouldn't include the xml bind transitive dependency if I recall this correctly...

oQue: @timtebeek there is indeed a problem that is far more complex than the issue I've created and current changes in this PR. The problem is, as @BoykoAlex has mentioned, not in a particular recipe, but in the way `GradleProject` is (not) updated when plugins and dependencies are updated.

E.g. if we have a dependency A with transitive dependency B and we run recipe with 2 steps:
1. upgrade A from v1 to v2 without transitive dependency B
2. add explicit declaration of dependency B

Second step is always ignored as there are currently 2 issues:
1. The problem @BoykoAlex has found: `GradleProject` seems to be not aware about changes in step (1), so it still thinks that there is a transitive dependency to B
2. The problem with recipe that I've described in #6411 : even with default settings to not accept transitive dependencies step (2) will fail as this logic was dropped in https://github.com/openrewrite/rewrite/pull/5371 . And as `GradleProject` keeps seeing B as transitive, recipe never adds dependency

I haven't confirmed and localized issue with `GradleProject` myself, but from what I see in our ignored steps in recipes, it seems exactly like that.

timtebeek: Thanks all! Let me link some related relevant work, such that we can explore a possible solution.

Firstly it's good to know there's ongoing work in UpgradePluginVersion that might conflict; good to wait for that to merge:
- https://github.com/openrewrite/rewrite/pull/6549

Secondly, we had a similar problem with Maven up until recently, and there we reworked the recipe to do more work in the scanner, such that we could update all sub projects in the visitor:
- https://github.com/openrewrite/rewrite/pull/6536

Perhaps that same approach can be applied here as well; it helps that the recipe is already a scanning recipe, but it doesn't help that the complexity is already relatively high with having to update multiple files based on where one can hide versions in Gradle. 🙃 Something to consider, but not an immediate quick fix just yet.

PR #6258, opened 2025-11-05
Instantiate using reflection to overcome classloading issue

Jenson3210:
As discussed with @pstreef, changing this one to have no ClassCastException


okundzich:
any background on this @Jenson3210? fyi @jkschneider

PR #6262, opened 2025-11-05
Introduce merge strategy to AppendToTextFile recipe

bmuschko:
## What's changed?

Introduces merge strategy to AppendToTextFile recipe.

## What's your motivation?

The recipe doesn't support a merge strategy which some end users may find helpful. This considers the following use case: You run the recipe against hundreds or thousands of recipes as a bulk operation. Some repositories may already contain the text partially (e.g. because developers made a change toward a solution in parts) but you still want to ensure that the full text is appended.

No

## Any additional context

I am going to check with the end user to see if the proposed solution is suitable.


timtebeek:
Thanks! I'll leave it up to you to merge since it seemed you were still checking the fit for purpose over there.

bmuschko:
I checked with the end user.

Issue #1295, opened 2021-12-25
XmlParsingException: mismatched input '﻿' expecting '<'. with file endoded in UTF-8 with BOM


murdos:
> org.openrewrite.xml.XmlParsingException: Syntax error in logback-test.xml at line 1:0 mismatched input '﻿' expecting '<'.

Content of logback-test.xml is attached: [logback-test.xml.gz](https://github.com/openrewrite/rewrite/files/7775736/logback-test.xml.gz)

I suspect that this is because the xml file is encoded in UTF-8 with BOM:
> $file logback-test.xml
> logback-test.xml: XML 1.0 document, Unicode text, UTF-8 (with BOM) text

traceyyoshima:
Hi @murdos, happy holidays! Thanks, for reporting the issue, we'll take a look!

Issue #1268, opened 2021-12-08
xml parser cant handle byte order mark


delanym:
The org.openrewrite.xml.XmlParsingException can't handle UTF8 files with a BOM

sambsnyd:
Thanks for reporting, we'll take a look

traceyyoshima:
Hi @delanym,
I'll gladly look into the parsing issue - would you please share what the XML looks likes?
Thanks in advance!

delanym:
[AndroidManifest.zip](https://github.com/openrewrite/rewrite/files/7682565/AndroidManifest.zip)

traceyyoshima:
Thanks for sample XML @delanym!
I took a look at the examples and the parsing issues are fixed on the main branch.
We're working really hard on getting a release out of 7.17.
Thanks again for the report

PR #6266, opened 2025-11-06
Restore UTF-8 BOM during printing

pstreef:
## What's changed?

OpenRewrite now restores UTF-8 BOMs during printing when `charsetBomMarked=true`. The BOM restoration happens in `Tree.print()` - the single convergence point for all printing paths. Also updated the RPC Print handler to use `tree.print()` instead of direct `printer.visit()` for consistent behavior.

## What's your motivation?

OpenRewrite's `EncodingDetectingInputStream` strips UTF-8 BOMs during parsing and sets `charsetBomMarked=true`, but printers never checked this flag to restore the BOM. This caused:

1. Files with BOMs lost them after OpenRewrite processing (not byte-identical)
2. Patch application failures when git has BOM but LST doesn't (context lines don't match)

The BOM restoration logic checks if a BOM is already present to avoid duplication. This is a defensive measure since in theory the printer should never add a BOM to the output.
There could be other places where printer.visit is used directly, the BOM will not be added there.

Knut

Alternative approach would be to add BOM restoration in each individual printer (JavaPrinter, XmlPrinter, etc.), but that would be error-prone and require updating many printers. The chosen approach puts it in one central location that all printing paths use.

## Any additional context

Historical context:
- [Issue #1268](https://github.com/openrewrite/rewrite/issues/1268): XML parser couldn't handle BOMs
- [Issue #1295](https://github.com/openrewrite/rewrite/issues/1295): XmlParsingException with UTF-8 BOM
- [PR #1707](https://github.com/openrewrite/rewrite/pull/1707): Added BOM detection
- [PR #4139](https://github.com/openrewrite/rewrite/pull/4139): Fixed BOM detection
- [PR #4546](https://github.com/openrewrite/rewrite/pull/4546): Made BOM stripping intentional for parser compatibility

The metadata (`charsetBomMarked`) was added but never used for restoration - this PR completes the round-trip implementation.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/RemoveByteOrderMark.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/RemoveByteOrderMark.java
@@ -46,8 +46,12 @@ public class RemoveByteOrderMark extends Recipe {
                     return sourceFile;
                 }
 
-                if (sourceFile.isCharsetBomMarked()) {
-                    return sourceFile.withCharsetBomMarked(false);
+                try {
+                    if (sourceFile.isCharsetBomMarked()) {
+                        return sourceFile.withCharsetBomMarked(false);
+                    }
+                } catch (UnsupportedOperationException e) {
+                    // Defensive fallback for any other SourceFile implementations that don't support charset operations
                 }
 
                 return sourceFile;


rewrite-core/src/main/java/org/openrewrite/Tree.java
```diff
@@ -79,6 +79,17 @@ default String print(Cursor cursor) {
...
+
+        // Restore BOM if the source file originally had one
+        if (this instanceof SourceFile) {
+            SourceFile sourceFile = (SourceFile) this;
+            if (sourceFile.isCharsetBomMarked() &&
+                out.out.length() > 0 &&
+                out.out.charAt(0) != '\uFEFF') {
+                out.out.insert(0, '\uFEFF');
+            }
+        }
```
knutwannheden, line 99: This is a really clever trick, as in the end all the various `print()` calls (including `printAll()` and `printAllTrimmed()`, etc.) end up calling this method.

I am thinking we could perhaps add the BOM marker before calling the visitor. If the concern is that the visitor also adds the BOM again, we could delete the duplicate afterwards (basically like you are doing here). But that is a minor detail and I am unsure which I think is better.

knutwannheden (approved):

rewrite-core/src/main/java/org/openrewrite/Tree.java
pstreef: this leaves it in a single block, which I kinda. Also the printer visitor can now assume the output is empty still, which might matter/be useful (although it should not I guess).

PR #6269, opened 2025-11-07
Fix UpgradePluginVersion to handle plugins with 'apply false'

### LINKED ISSUE #6268: UpgradePluginVersion fails to upgrade plugins with 'apply false'
## Description

The `UpgradePluginVersion` recipe currently fails to upgrade plugin versions when the `apply false` clause is present in the plugin declaration.

## Current Behavior

When a Gradle plugin is declared with `apply false`, the recipe does not detect it as a plugin version that should be upgraded:

```kotlin
// Kotlin DSL - Not upgraded
plugins {
    id("com.github.johnrengelman.shadow") version "7.1.0" apply false
}
```

```groovy
// Groovy DSL - Not upgraded
plugins {
    id 'com.github.johnrengelman.shadow' version '7.1.0' apply false
}
```

## Expected Behavior

The recipe should upgrade the plugin version regardless of the presence of `apply false`:

```kotlin
// Should be upgraded to:
plugins {
    id("com.github.johnrengelman.shadow") version "7.1.2" apply false
}
```

## Root Cause

The current implementation in `isPluginVersion()` only checks the immediate parent of the `version()` method call for the `plugins()` method. When `apply false` is present, the chain becomes:
```
plugins() -> ... -> apply() -> version()
```

The method needs to walk up the entire method invocation chain to find `plugins()`.

## Environment

- **OpenRewrite version**: main branch (latest)
- **Build tool**: Gradle
- **DSL**: Both Kotlin DSL and Groovy DSL affected
- **Recipe**: `org.openrewrite.gradle.plugins.UpgradePluginVersion`

## Proposed Solution

Modify `isPluginVersion()` to walk up the method invocation chain until it finds the `plugins()` method or reaches the root, rather than only checking the immediate parent.

I have a fix ready and will submit a PR shortly with test coverage for both Kotlin DSL and Groovy DSL cases.

flex-gwanbin:
- Fixes #6268

## Changes

- Modified `isPluginVersion()` to walk up the entire method invocation chain to find the `plugins()` method
- Previously only checked the immediate parent, which failed when `apply false` was between `version()` and `plugins()`
- Now correctly handles patterns like: `id(...) version "x.y.z" apply false`

## Testing

Added test coverage for both DSLs:
- `kotlinDslWithApplyFalse`: Tests Kotlin DSL with apply false
- `groovyDslWithApplyFalse`: Tests Groovy DSL with apply false

### Before
```kotlin
plugins {
    id("com.github.johnrengelman.shadow") version "7.1.0" apply false
}
```
No upgrade happened ❌

### After
```kotlin
plugins {
    id("com.github.johnrengelman.shadow") version "7.1.2" apply false
}
```
Successfully upgraded ✅

## Root Cause

The method invocation chain with `apply false` is:
```
plugins() → ... → apply() → version()
```

The previous implementation only checked one level up from `version()`, finding `apply()` instead of `plugins()`.

## Technical Details

Changed in `UpgradePluginVersion.java:116-141`:
- Added a while loop to traverse the method invocation chain
- Continues until finding `plugins()` or reaching the root
- Properly handles intermediate method calls like `apply()`

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
@@ -146,7 +146,7 @@ public class UpgradePluginVersion extends ScanningRecipe<UpgradePluginVersion.De
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
 
-                GradlePlugin plugin = new GradlePlugin.Matcher().pluginIdPattern(effectivePluginIdPattern()).get(getCursor()).orElse(null);
+                GradlePlugin plugin = new GradlePlugin.Matcher().pluginIdPattern(pluginIdPattern).get(getCursor()).orElse(null);
                 if (plugin == null || plugin.getPluginId() == null) {
                     return m;
                 }
@@ -293,7 +293,7 @@ public class UpgradePluginVersion extends ScanningRecipe<UpgradePluginVersion.De
             @Override
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 // Match the trait before super to ensure the cursor is unmodified
-                GradlePlugin plugin = new GradlePlugin.Matcher().pluginIdPattern(effectivePluginIdPattern()).get(getCursor()).orElse(null);
+                GradlePlugin plugin = new GradlePlugin.Matcher().pluginIdPattern(pluginIdPattern).get(getCursor()).orElse(null);
 
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
 
@@ -333,11 +333,6 @@ public class UpgradePluginVersion extends ScanningRecipe<UpgradePluginVersion.De
             }
         };
         return Preconditions.or(propertiesVisitor, Preconditions.check(Preconditions.or(new IsBuildGradle<>(), new IsSettingsGradle<>()), javaVisitor));
-    }
-
-    private String effectivePluginIdPattern() {
-        // Translate the short "kotlin" DSL name to the full plugin ID pattern
-        return "kotlin".equals(pluginIdPattern) ? "org.jetbrains.kotlin.*" : pluginIdPattern;
     }
 
     private @Nullable String literalValue(Expression expr) {
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/UpgradePluginVersionTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/UpgradePluginVersionTest.java
@@ -62,7 +62,7 @@ class UpgradePluginVersionTest implements RewriteTest {
     @Test
     void upgradeKotlinPluginLiteralVersion() {
         rewriteRun(
-          spec -> spec.recipe(new UpgradePluginVersion("kotlin", "2.3.0", null)),
+          spec -> spec.recipe(new UpgradePluginVersion("org.jetbrains.kotlin.*", "2.3.0", null)),
           buildGradleKts(
             """
               plugins {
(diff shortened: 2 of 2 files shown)


shanman190:
This recipe predates several new enhancements, one of which I think could excellently be deployed here and that's the [`GradlePlugin`](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradlePlugin.java) trait. It understands the apply syntax already as well as a few more syntactic possibilities.

In it's current form, it'd help to find the plugin declaration while also handling several more cases. Furthermore, like we've done with other traits could be updated to handle the basic changes that one may want to apply to a plugin declaration.

If you feel up to this @flex-gwanbin , then I can help coach you through various aspects of the above. Alternatively, if that feels a bit too much Tim or I can jump in and apply this improvement.

timtebeek:
Thanks both! I've gone ahead and made the changes to use the new trait here ; would appreciate another quick round of review from @shanman190  before we merge.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -164,37 +145,25 @@ public J visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionC
             @Override
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (!isPluginVersion(getCursor())) {
+
+                if (!"version".equals(m.getSimpleName())) {
```
shanman190: Are we needing to stop short here for some reason? The trait will select the outer most element of the plugin chain (apply -> version -> plugin).

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -164,37 +145,25 @@ public J visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionC
...
-                String pluginId;
-                if ("kotlin".equals(((J.MethodInvocation) m.getSelect()).getSimpleName())) {
-                    pluginId = "kotlin";
-                } else {
-                    pluginId = literalValue(pluginArgs.get(0));
-                }
-                if (pluginId == null || !StringUtils.matchesGlob(pluginId, pluginIdPattern)) {
+
+                String pluginId = plugin.getPluginId();
+                if (!matchesPluginId(pluginId, pluginIdPattern)) {
```
shanman190: Plugin Id matching is built into the trait.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -338,31 +299,24 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
...
-                    pluginId = "kotlin";
-                } else {
-                    pluginId = literalValue(pluginArgs.get(0));
-                }
-                if (pluginId == null || !StringUtils.matchesGlob(pluginId, pluginIdPattern)) {
-                    return m;
+                // Match the trait before super to ensure the cursor is unmodified
+                GradlePlugin plugin = "version".equals(method.getSimpleName()) ?
+                        new GradlePlugin.Matcher().get(getCursor()).orElse(null) : null;
+                if (plugin != null && !matchesPluginId(plugin.getPluginId(), pluginIdPattern)) {
```
shanman190: Plugin id matching is built into the trait.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -391,6 +335,11 @@ public J visitVariable(J.VariableDeclarations.NamedVariable variable, ExecutionC
         return Preconditions.or(propertiesVisitor, Preconditions.check(Preconditions.or(new IsBuildGradle<>(), new IsSettingsGradle<>()), javaVisitor));
     }
 
+    private String effectivePluginIdPattern() {
+        // Translate the short "kotlin" DSL name to the full plugin ID pattern
+        return "kotlin".equals(pluginIdPattern) ? "org.jetbrains.kotlin.*" : pluginIdPattern;
+    }
```
shanman190: I don't think this is necessary or the places where we are doing this aren't correct.

In particular, the trait expands out the Kotlin plugin alias to its full plugin id automatically -- `kotlin("jvm")` -> `org.jetbrains.kotlin.jvm`. Since the pluginIdPattern is matching on the effective plugin id, then it stands to reason that if an `UpgradePluginVersion` is trying to target all Kotlin plugins it should instead be doing a pattern of `org.jetbrains.kotlin.*` as shown. `kotlin` standalone doesn't really mean anything from a Gradle standpoint.

PR #6278, opened 2025-11-10
Reorganize RecipeMarketplace to better support reading CSVs without l…

jkschneider:
## What's changed?

After working to integrate the new `RecipeMarketplace` a few places downstream, some API changes were in order.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeMarketplace.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeMarketplace.java
@@ -25,6 +25,8 @@ import org.openrewrite.NlsRewrite;
 
 import java.util.*;
 
+import static java.util.Collections.addAll;
+
 @Incubating(since = "8.66.0")
 public class RecipeMarketplace {
     private final @Getter(AccessLevel.PACKAGE) Category root = new Category("Root");
@@ -32,7 +34,7 @@ public class RecipeMarketplace {
 
     public RecipeMarketplace setResolvers(RecipeBundleResolver... resolvers) {
         this.resolvers.clear();
-        Collections.addAll(this.resolvers, resolvers);
+        addAll(this.resolvers, resolvers);
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/Recipe.java
```diff
@@ -227,7 +225,7 @@ protected RecipeDescriptor createRecipeDescriptor() {
```
timtebeek, line 227:
```suggestion
        URI recipeSource;
        try {
            recipeSource = getClass().getProtectionDomain().getCodeSource().getLocation().toURI();
        } catch (URISyntaxException e) {
            throw new RuntimeException(e);
        }
```

rewrite-core/src/main/java/org/openrewrite/Recipe.java
```diff
@@ -37,8 +37,6 @@
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.time.Duration;
```
timtebeek, line 42:
```suggestion
import java.net.URI;
import java.net.URISyntaxException;
import java.time.Duration;
```

rewrite-core/src/main/java/org/openrewrite/Recipe.java
```diff
@@ -227,7 +225,7 @@ protected RecipeDescriptor createRecipeDescriptor() {
 
         return new RecipeDescriptor(getName(), getDisplayName(), getInstanceName(), getDescription(), getTags(),
                 getEstimatedEffortPerOccurrence(), options, recipeList1, getDataTableDescriptors(),
-                getMaintainers(), getContributors(), getExamples(), null);
+                getMaintainers(), getContributors(), getExamples());
```
timtebeek:
```suggestion
                getMaintainers(), getContributors(), getExamples(), recipeSource);
```

PR #6282, opened 2025-11-11
JacocoReportDeprecations should replace deprecated methods with assignments to make it work with Gradle 8

akiraly:
- Refactor `JacocoReportDeprecations` method handling to replace deprecated invocations (original support added in #6232) with assignment statements instead of method calls that do not exist in Gradle 8.

## What's changed?
The previous implementation replaced calls like `enabled(true)` with `required(true)` but the latter doesn't work in Gradle 8. The new implementation replaces them with `required = true` instead.

## What's your motivation?
To make the Gradle 7 -> 8 upgrade recipe work, which depends on the `JacocoReportDeprecations` recipe.

### Checklist

Legal disclaimer:

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
@@ -461,7 +461,7 @@ class JacocoReportDeprecationsTest implements RewriteTest {
     }
 
     @Nested
-    class AsMethodInvocation {
+    class MethodInvocationToAssignments {
         @Test
         void deprecationsInMethodInvocationSyntax() {
             rewriteRun(
@@ -511,17 +511,11 @@ class JacocoReportDeprecationsTest implements RewriteTest {
                       reports {
                           xml {
                               enabled false
-                              enabled true
-                              enabled false
                           }
                           csv {
                               enabled true
-                              enabled false
-                              enabled true
                           }
                           html {
-                              enabled false
-                              enabled false
                               enabled false
                           }
                       }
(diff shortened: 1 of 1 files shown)


akiraly:
The build error doesn't seem to be related to my change:

> * Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':rewrite-javascript:npmTest'.

rewrite-gradle/src/test/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecationsTest.java
```diff
@@ -511,18 +511,12 @@ void deprecationsInMethodInvocationInsideNestedClosure() {
                       reports {
                           xml {
                               enabled false
-                              enabled true
-                              enabled false
```
akiraly: @timtebeek FYI I added these duplicate lines to the tests to make sure I get the paddings right in all cases. I am not saying it is necessary here, just explaining why I added them originally.

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle8/JacocoReportDeprecations.java
akiraly, line 1: @timtebeek thanks for formatting the files. Out of curiosity what are the official code formatting rules and how do you apply them? I noticed that there are some `.editorconfig` files in the code base but they are scattered around, for example there is one in `rewrite-gradle/src/test/java/org/openrewrite/gradle/.editorconfig` but I think that's only applied to the files in that part of the file tree.

timtebeek: Thanks for asking! We typically use the IntelliJ defaults, with `.editorconfig` to ensure we limit continued indentation to two spaces in tests, to avoid running off the screen. Hope that helps!

akiraly: Well that explains some of my confusion. Because I saw 4 spaces in used in some files and then Intellij automatically inserting 2 spaces. And then I thought I need to fix those manually to 4 :)

timtebeek: Hah, good to have that cleared up! I look forward to what else you'll do :)

PR #6289, opened 2025-11-14
Add `@CheckReturnValue` annotations to tree packages

timtebeek:
And make a few adjustments based on new warnings seen.

A quick caveat: this adds the `javax.annotation.CheckReturnValue` which comes from `findbugs:jsr305`; That has the effect of showing warnings in IntelliJ when annotated on the package, where others failed to show, including an internal one I briefly tried.

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/tree/K.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/tree/K.java
@@ -132,8 +132,7 @@ public interface K extends J {
         @Transient
         @Override
         public long getWeight(Predicate<Object> uniqueIdentity) {
-            AtomicInteger n = new AtomicInteger();
-            new KotlinVisitor<AtomicInteger>() {
+            return new KotlinVisitor<AtomicInteger>() {
                 final JavaTypeVisitor<AtomicInteger> typeVisitor = new JavaTypeVisitor<AtomicInteger>() {
                     @Override
                     public JavaType visit(@Nullable JavaType javaType, AtomicInteger n) {
@@ -166,8 +165,7 @@ public interface K extends J {
                     }
                     return markers;
                 }
-            }.visit(this, n);
-            return n.get();
+            }.reduce(this, new AtomicInteger()).get();
         }
 
         @Override


rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
```diff
@@ -446,7 +446,7 @@ public Pom toPom(@Nullable Path inputPath, @Nullable MavenRepository repo) {
                 .profiles(mapProfiles(getProfiles()))
                 .subprojects(mapSubProjects(getModules(), getSubprojects()));
         if (StringUtils.isBlank(pomVersion)) {
-            builder.dependencies(mapRequestedDependencies(getDependencies()))
+            builder = builder.dependencies(mapRequestedDependencies(getDependencies()))
```
timtebeek, line 449: As far as I can tell this is not a fluent builder, so this is fixing an actual issue where the return value was ignored.

timtebeek:
We're also still seeing false positive warnings on these two lines:
https://github.com/openrewrite/rewrite/blob/1ebb4a20a27b6f218da9e271d036e0d55fd2050c/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java#L111-L112

Because the underlying builders do not adhere to the expected pattern by returning `this` instead of a new copy:
https://github.com/openrewrite/rewrite/blob/9d7fe7709a41e66816033b29786deab0173b3091/rewrite-maven/src/main/java/org/openrewrite/maven/tree/MavenRepository.java#L155-L173

I think it's fine to live with these two warnings for now; Alternatively we can restructure that builder, or add explicit suppressions.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
timtebeek: This has been fixed separately in https://github.com/openrewrite/rewrite/pull/6290, just so this PR can be evaluated without some of the resulting fixes.

rewrite-protobuf/src/main/java/org/openrewrite/protobuf/tree/package-info.java
```diff
@@ -13,7 +13,11 @@
...
  * limitations under the License.
  */
+@CheckReturnValue
 @NullMarked
 package org.openrewrite.protobuf.tree;
 
 import org.jspecify.annotations.NullMarked;
+
+import javax.annotation.CheckReturnValue;
+import javax.annotation.CheckReturnValue;
```
knutwannheden:
```suggestion
```

Issue #6287, opened 2025-11-14
Allow `AddMethodParameter` to operate on overridden methods


dsgrieve:
## What problem are you trying to solve?
Say I have
  `interface A { int count(Sheep flock); }`
and an implementation of that
  `class Sleep implements A { @Override public int count(Sheep flock) { return  flock.size(); }`

If interface A changes the count method to `int count(Sheep flock, Wolf wolf);`, I'd like to be able to add the new method parameter to classes that inherit/extend from the super class. I'd like to be able to use
```
recipeList:
  - org.openrewrite.java.AddMethodParameter
    methodPattern: "A count(Sheep)"
    parameterType: "Wolf"
    parameterName: "wolf"
    parameterIndex: 1
```
As `AddMethodParameter` is implemented now, it will only match the specific type. For this to work, then
I would have to have `methodPattern: "Sleep count(Sheep)"`. But what if I have many implementations of `A`? I would have to write an imperative recipe that pulls the type of the overriding class to properly configure and call AddMethodParameter on the fly. 
## Describe the solution you'd like
Before:
  `class Sleep implements A { @Override public int count(Sheep flock) { return  flock.size(); }`
After
  `class Sleep implements A { @Override public int count(Sheep flock, Wolf wolf) { return  flock.size(); }`

## Have you considered any alternatives or workarounds?
Create an imperative recipe that does something like this in a `visitMethodDeclaration`
```
                                String receiverType = enclosingClass.getType().getFullyQualifiedName();
                                String methodPattern = receiverType + " count(Sheep)";

                                doAfterVisit(new AddMethodParameter(
                                    methodPattern,
                                    "Wolf",
                                    "wolf",
                                    1
                                ).getVisitor());

```

Anmol202005:
I am on it :)

PR #6294, opened 2025-11-14
Allow `AddMethodParameter` to match overridden methods

Anmol202005:
Improves AddMethodParameter so that methodPattern resolution includes overridden methods from interfaces and superclasses.

- Fixes: #6287

### Checklist


Anmol202005:
@timtebeek kindly review :)

timtebeek:
Thanks for the runnable example! I'm off here today, but will try to circle back next week.

I had thought we could clear this up with just changing the expected method pattern instead.

Anmol202005:
ping @timtebeek

timtebeek:
Will likely move into next week before I can get to this; I've got some online trainings to give and a weekend ahead first. 🙏🏻

PR #6297, opened 2025-11-15
Add find managed dependency version where specified via property

kbrockhoff:
## What's changed?

Added handling for version set via property in the same pom.xml to FindManagedDependency. This change makes the recipe functionally equivalent to FindDependency which already supports version set via property.

## What's your motivation?

It is a common practice to set dependency management dependency versions via a property. In this use case, FindManagedDependency does not work as expected without this change.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindManagedDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/search/FindManagedDependency.java
@@ -121,10 +121,9 @@ public class FindManagedDependency extends Recipe {
         }
         String actualVersion = resolvedDependency.getVersion();
         Validated<VersionComparator> validate = Semver.validate(desiredVersion, versionPattern);
-        if (validate.isInvalid()) {
-            return false;
-        }
-        assert validate.getValue() != null;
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/search/FindManagedDependency.java
```diff
@@ -99,23 +101,25 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
             return true;
         }
-        if (actualVersion == null) {
-            // rare but technically valid for a dependencyManagement entry to have no version; bail if so
+        ResolvedManagedDependency resolvedDependency = resolvedDependencySupplier.get();
+        if (resolvedDependency == null) {
+            // shouldn't happen, but if it does, fail the condition
             return false;
         }
+        String actualVersion = resolvedDependency.getVersion();
```
timtebeek, line 122: There's a slight difference here compared to `FindDependency` in that `actualVersion` here is nullable, as `resolvedDependency.getVersion()` can return null; Would be good to handle that properly, as the next line might otherwise throw an NPE.

PR #6317, opened 2025-11-20
Avoid `String` allocations in `Version`

knutwannheden:
By not calling `String#substring()` we can avoid unnecessary `String` allocations. As `Integer#parseInt(CharSequence, int, int, int)` is only available in Java 9, I inlined a simplified version of it.


jkschneider:
Let's see if the complexity here is worthwhile after we see the effect of Shannon's earlier optimizations around `Version` are fully rolled out.

PR #6325, opened 2025-11-22
WIP: Use IR for KotlinParser instead of PSI/FIR

Crustack:
## What's changed?
Instead of relying on PSI/FIR for Kotlin Parsing, this proposes to leverage `KotlinToJVMBytecodeCompiler.INSTANCE.analyze` + `JvmIrCodegenFactory.convertToIr` to generate the `IrFile`s.

## What's your motivation?
Improve Kotlin support

Since I am fairly new to the entire Kotlin Compiler eco-system and have only worked with PSI before, some input on the used methods to get the IrFiles is welcome.
Also since I haven't looked deeply into the `KotlinTreeParserVisitor` implementation yet, it would be good to know if we would still need PSI/FIR for some information or would IR be sufficient?

@shanman190 

No

## Any additional context
TODOs:
- [x] Successfully generate `IrFile`s for the given source kotlin files (working `KotlinTypeIrSignatureBuilderTest`)
- [ ] Remove `firFile` + `nodes` from `KotlinSource` (or will they be necessary still?)
- [ ] Test `KotlinTypeMapping` with generated `FirFile`s
- [ ] Remove `firSessions` from `CompiledSource`
- [ ] Rewrite `KotlinTreeParserVisitor` via custom  `IrTreeVisitor`?

### Checklist

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -25,6 +25,7 @@ import org.jetbrains.kotlin.cli.common.messages.AnalyzerWithCompilerReport;
 import org.jetbrains.kotlin.cli.common.messages.MessageCollector;
 import org.jetbrains.kotlin.cli.common.messages.PrintingMessageCollector;
 import org.jetbrains.kotlin.cli.pipeline.jvm.JvmFir2IrPipelineArtifact;
+import org.jetbrains.kotlin.cli.pipeline.jvm.JvmFrontendPipelineArtifact;
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable;
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer;
 import org.jetbrains.kotlin.com.intellij.openapi.util.text.StringUtilRt;
@@ -171,6 +172,7 @@ public class KotlinParser implements Parser {
                                         parsingListener.parsed(kotlinSource.getInput(), cu);
                                         return requirePrintEqualsInput(cu, kotlinSource.getInput(), relativeTo, ctx);
                                     } catch (Throwable t) {
+                                        t.printStackTrace();
                                         ctx.getOnError().accept(t);
                                         return ParseError.build(this, kotlinSource.getInput(), relativeTo, ctx, t);
                                     }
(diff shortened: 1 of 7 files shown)


shanman190:
@PhilKes , thanks for the work here. However, this seems to be using the K1 compiler which would limit options toward Kotlin 2.x work and be going in reverse of where the new development is happening by the Kotlin team in the K2 compiler.

Crustack:
> @PhilKes , thanks for the work here. However, this seems to be using the K1 compiler which would limit options toward Kotlin 2.x work and be going in reverse of where the new development is happening by the Kotlin team in the K2 compiler.

Hi thanks for the feedback, I now updated the code to leverage the same logic as `K2JvmCompiler` does.
The `KotlinTypeIrSignatureBuilderTest` succeeds as it was before (`IrProperty` instead of `IrField` for KtFile's `field`).

My next step would be to try to update the `build.gradle.kts` `kotlinVersion` to `2.2.21`, since I saw that the JVM Compiler Pipeline package was refactored once more (?).
I guess that would mean implementing a custom `AbstractCliPipeline` that uses the phases:
 JvmConfigurationPipelinePhase then JvmFrontendPipelinePhase then  JvmFir2IrPipelinePhase, which returns `JvmFir2IrPipelineArtifact` which includes the `IrFile`s

Crustack:
- I updated the `kotlinVersion` for the dependenies to `2.2.21`
-  Refactored the compilation code to use a custom `JvmFir2IrPipeline` which yields the `FirFile` + `IrFile`s.
- Since I have not yet changed the `KotlinTreeParserVisitor` from PSI/FIR to IR I wanted to make sure all the tests still pass.
- `KotlinIrTypeMappingTest` + `KotlinParserTest` 
- I noticed that e.g. for `EqualsMethodUsageTest` the compiler throws errors like: `openRewriteFile0.kt:3:1: error: missing return statement`
https://github.com/openrewrite/rewrite/blob/522827a66bf49799cc7c15d4bf1549eb2f5e7870/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/cleanup/EqualsMethodUsageTest.java#L38-L40
Are these error supposed to be ignored, or do we always expect valid/complete Kotlin code?

Other than that I am checking what differences are in the generated `FirFile` before and after the compiler changes, since most recipe specific tests currently do not succeed, I am guessing there are some smaller changes to the structure and perhaps some of the `K2JVMCompilerArguments` still have to be fine-tuned but in general I think it looks promising so far.
Something that differs from before is that the `FirSimpleFunctionImpl` objects have `controlFlowGraphReference`= null and some other small differing fields that leads to the `PsiElementAssociations` having less entries in `elementMap`

Crustack:
Since its not prudent trying to update Kotlin-Compiler and move from PSI/FIR to IR in one PR, I am closing this in favor of: https://github.com/openrewrite/rewrite/pull/6338
After it is finished I will probably open another PR to start working on moving to IR based parsing

PR #6332, opened 2025-11-24
Add recipe RemoveMethodThrows

### LINKED ISSUE #6331: Remove method throws
I need to remove some exceptions from the method throws clause (see [JLS spec](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.4.6)).
That's why I'd like to add this recipe. 
The recipe is planned to be used on method declaration.

Precondition `DeclaresMethod` using a method pattern.

```java
import java.io.IOException;

class A {
    public void foo() throws IOException {
        // no-op
    }
}
```

```java
class A {
    public void foo() {
        // no-op
    }
}
```

None found.

## Any additional context

The options planned for this recipe, so far:
- `methodPattern`
- `exceptionType` (FQN)
- `matchOverrides` (true by default), useful for `MethodMatcher`
- `removeAll` (false by default) -- if true, `exceptionType` is ignored

Yes, I'll submit a PR.

ghusta:
- Fixes #6331

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/RemoveMethodThrows.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/RemoveMethodThrows.java
@@ -22,6 +22,7 @@ import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.search.DeclaresMethod;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
 
 @EqualsAndHashCode(callSuper = false)
 @Value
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/RemoveMethodThrows.java
```diff
@@ -0,0 +1,145 @@
...
+            description = "Whether to match overridden forms of the method on subclasses of typeMatcher. Default is true.",
+            required = false)
+    @Nullable
+    Boolean matchOverrides;
+
+    @Option(displayName = "Remove all throws declarations",
+            description = "Remove all throws declarations. Default is false.",
+            required = false)
+    @Nullable
+    Boolean removeAll;
```
timtebeek: Would it make sense to recommend wildcards here instead of a dedicated option?

ghusta: OK I see what you mean. 
If it must use wildcards, maybe the option `exceptionType` could use a _type pattern_, like in the recipe https://docs.openrewrite.org/recipes/java/search/findimports.
I'll see if it can use a `org.openrewrite.java.TypeMatcher` or something like this.
What do you think ?

timtebeek: Indeed a type pattern marcher would support to remove all as well, without the need for an explicit option. If needed I can try to fit that in as well. Your help so far is already appreciated!

ghusta: I have tested a refactoring, so I've pushed a last commit. Then I let you do a final rewiew if you want to change some things.
Thank you for your help 👍

ghusta:
> Thanks again! I've both simplified the implementation a bit, and expanded it to also change the type information we have on method invocations and method references, such that if you were to run `org.openrewrite.staticanalysis.UnnecessaryCatch` right after this one, it would also clear out catch blocks that can no longe r be reached.
> 
> I look forward to seeing this used downstream!

Thank you !
Great work.

PR #6333, opened 2025-11-24
Reduce GC pressure in dependency graph traversal

pstreef:
## Problem

The `DependencyInsight` recipe exhibits high GC pressure when analyzing Gradle projects with complex dependency graphs. Extended benchmark runs show the original code spending 71% of execution time in GC, with worst-case scenarios triggering catastrophic GC collapse (28+ seconds of GC time in a 5-second measurement window).

## Solution

Optimized `AbstractDependencyGraphBuilder.collectDependencyPathsRecursive` to reduce allocation pressure through three changes:

1. **Backtracking pattern** - Reuse a single `ArrayList` during traversal instead of creating new copies at each recursion level
2. **End-based operations** - Add/remove at end of list (O(1)) instead of index 0 (O(n))
3. **Loop-based cycle detection** - Replace stream operations with direct loops to eliminate intermediate allocations

**Benchmark results** (Spring Boot project, wildcard dependency search, 10 iterations, 8GB heap):

| Metric | Original | Optimized | Improvement |
|--------|----------|-----------|-------------|
| Total GC time | 35,784 ms (71%) | 8,127 ms (16%) | **-77%** |
| Worst iteration | 6,890 ops/s | 54,718 ops/s | **8x better** |
| Max GC per iteration | 28,448 ms | 1,345 ms | **-95%** |
| Stability | Catastrophic collapse | Stable | ✓ |

The optimization prevents catastrophic GC thrashing that occurs in extended runs without the changes, providing more stable and predictable performance.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/marketplace/RecipeClassLoader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/marketplace/RecipeClassLoader.java
@@ -65,6 +65,7 @@ public class RecipeClassLoader extends URLClassLoader {
             "org.openrewrite.FileAttributes",
             "org.openrewrite.PrintOutputCapture",
             "org.openrewrite.ipc.http.HttpSender",
+            "org.openrewrite.java.internal.TypesInUse",
             "org.openrewrite.maven.MavenDownloadingException",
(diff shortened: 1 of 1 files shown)


bmuschko (approved):
Nice! I was kind of concerned about potential memory issues when adding this feature.

Issue #5179, opened 2025-03-14
Idempotency issue with YAML tags used in sequences


greg-at-moderne:
## What version of OpenRewrite are you using?

Current main = 46a4761b1833cc82e2c1c594c6e0933adcf2d1e8

## What is the smallest, simplest way to reproduce the problem?

The following unit test case:
```java
    @Test
    void tagsInSequences() {
        rewriteRun(
          yaml(
            """
            Conditions:
              IsPollingFrequencyInMinutesSingular: !Equals [!Ref PollingFrequencyInMinutes, 1]
            """
          )
        );
    }
```
fails with:
```diff
  Conditions:
-   IsPollingFrequencyInMinutesSingular: !Equals [!Ref PollingFrequencyInMinutes, 1]
+   IsPollingFrequencyInMinutesSingular: !Equals [ !Equals !Ref PollingFrequencyInMinutes, 1]
```

## Context

Must be some case not taken into account in #4934.

aet2505:
As far as I can tell this is caused by the tag being both attached to the sequence but also included in the startBracketPrefix.
The startBracketPrefix renders the tag in the expected location but the tag attached to the sequence renders inside the square brackets.

The below code change to YamlParser L380 resolves the test provided with no regression within rewrite-yaml
`if (sse.getTag() != null && startBracketPrefix == null) {`

However this feels incorrect and a bit of a bodge and not confident enough in my Yaml syntax knowledge to spot the edge case!
Any pointer much appreciated

Vishalup29:
Hello @greg-at-moderne and all,

I’ve submitted a PR addressing the idempotency issue with YAML tags in sequences (#5179). This is my first contribution to OpenRewrite, so I’m very open to guidance and feedback. I tried to follow the existing conventions and the discussion here to ensure the fix is in the right direction.

I would greatly appreciate it if reviewers could point out any improvements, best practices, or adjustments needed. I’m happy to revise as required to align with the project’s standards.

Thank you for your time and for helping me get started with contributions to OpenRewrite!

PR #6334, opened 2025-11-25
Fixing parsing of YAML tags in sequences

Vishalup29:
fix(yaml): prevent duplicate tags in sequences

In some YAML sequences, tags were being rendered twice: once in the startBracketPrefix and once attached to the sequence. This caused idempotency issues when formatting YAML.

- Update YamlParser to only attach a tag if startBracketPrefix is null, resolving the failing test case in #5179.
## What's changed?
A minor update to `YamlParser` to prevent duplicate tags in sequences. The parser now only attaches a tag to a sequence if `startBracketPrefix` is null, resolving the idempotency issue.

## What's your motivation?
- Fixes a failing test case (#5179) where YAML sequences with tags were incorrectly rendered with the tag appearing twice: once in the prefix and once inside the brackets.

- The logic change at the `SequenceStartEvent` handling in `YamlParser`:
  ```java
  if (sse.getTag() != null && startBracketPrefix == null) {
      ...
  }

Reviewers

@greg-at-moderne, @aet2505

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -399,7 +399,7 @@ public class YamlParser implements org.openrewrite.Parser {
                             startBracketPrefix = fullPrefix.substring(startIndex, openingBracketIndex);
                         }
                         Yaml.Tag tag = null;
-                        if (sse.getTag() != null && startBracketPrefix == null) { // <-- add check
+                        if (sse.getTag() != null && startBracketPrefix == null) {
                             String prefixAfterColon = fullPrefix.substring(startIndex);
                             final int tagStartIndex = prefixAfterColon.indexOf('!');
                             String tagPrefix = prefixAfterColon.substring(0, tagStartIndex);


greg-at-moderne:
Please add a test - e.g. the one mentioned in the issue description.

Vishalup29:
@greg-at-moderne Thanks for taking a look, I see the test was added by @timtebeek 
Thanks for the test. Pls review/merge when possible

rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -399,7 +399,7 @@ private Yaml.Documents parseFromInput(Path sourceFile, EncodingDetectingInputStr
                             startBracketPrefix = fullPrefix.substring(startIndex, openingBracketIndex);
                         }
                         Yaml.Tag tag = null;
-                        if (sse.getTag() != null) {
+                        if (sse.getTag() != null && startBracketPrefix == null) { // <-- add check
```
greg-at-moderne:
```suggestion
                        if (sse.getTag() != null && startBracketPrefix == null) {
```

greg-at-moderne:
Thanks for coming up with the PR.
I remember the @aet2505 had a comment about this feeling incorrect, but I fail to understand why.

greg-at-moderne (approved):
I suggest we merge it. I consider it an improvement. And if new information to light, we might find a better solution.

aet2505:
My vague memory is that it felt like a bodge to me because it didn't parse the value as a 'Tag' correctly so potentially would have cases where if you tried to rewrite tags it might not behave as expected.

It definitely felt like it worked for me and I couldn't find the case where it failed!

PR #6335, opened 2025-11-25
Add version support for DoesNotIncludeDependency (Gradle and Maven)

DidierLoiseau:
## What's changed?
Added `version` as an optional parameter to `DoesNotIncludeDependency`, for both Gradle and Maven. This is a quick win since both simply rely on `DependencyInsight` under the hood.

## What's your motivation?
- In openrewrite/rewrite-testing-frameworks#850, I noted that the exclude for JUnit 4 should only apply when Testcontainers 1.x is not present. If we don’t want to force an upgrade to Testcontainers 2, `DoesNotIncludeDependency` must allow to specify the version.

I didn’t do extensive tests on different combination of parameters since that would actually be testing `DependencyInsight` itself.

- Possibly @timtebeek as this is a prerequisite for a possible fix for openrewrite/rewrite-testing-frameworks#850, but this PR would also be useful on its own.

n/a

## Any additional context
- openrewrite/rewrite-testing-frameworks#850

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/DoesNotIncludeDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/search/DoesNotIncludeDependencyTest.java
@@ -124,7 +124,8 @@ class DoesNotIncludeDependencyTest implements RewriteTest {
         void withNonMatchingVersionMarked() {
             rewriteRun(
               spec -> spec.recipe(defaultRecipeWithVersion("5.x")),
-              buildGradle(directDependencyTemplate.formatted("api"),
+              buildGradle(
+                directDependencyTemplate.formatted("api"),
                 String.format(marker + directDependencyTemplate, "api"))
             );
         }
@@ -197,7 +198,8 @@ class DoesNotIncludeDependencyTest implements RewriteTest {
         void withNonMatchingVersionMarked() {
             rewriteRun(
               spec -> spec.recipe(defaultRecipeWithVersion("5.x")),
-              buildGradle(transitiveDependencyTemplate.formatted("api"),
+              buildGradle(
+                transitiveDependencyTemplate.formatted("api"),
                 String.format(marker + transitiveDependencyTemplate, "api"))
             );
         }
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/search/DoesNotIncludeDependencyTest.java
@@ -186,7 +186,8 @@ class DoesNotIncludeDependencyTest implements RewriteTest {
             void withNonMatchingVersionMarked() {
                 rewriteRun(
                   spec -> spec.recipe(defaultRecipeWithOnlyDirectAndVersion(onlyDirect, "5.x")),
-                  pomXml(directDependencyTemplate.formatted(""),
+                  pomXml(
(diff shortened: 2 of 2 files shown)


DidierLoiseau:
Don’t know why the build failed there, it seems unrelated 🤔 (some JS tests)

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DoesNotIncludeDependency.java
```diff
@@ -37,6 +38,15 @@ public class DoesNotIncludeDependency extends Recipe {
             example = "guava")
     String artifactId;
 
+    @Option(displayName = "Version",
+            description = "Match only dependencies with the specified resolved version. " +
+                    "Node-style [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors) may be used." +
```
timtebeek:
```suggestion
                    "Node-style [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors) may be used. " +
```

rewrite-maven/src/main/java/org/openrewrite/maven/search/DoesNotIncludeDependency.java
```diff
@@ -53,6 +54,15 @@ public class DoesNotIncludeDependency extends Recipe {
     @Nullable
     String scope;
 
+    @Option(displayName = "Version",
+            description = "Match only dependencies with the specified version. " +
+                    "Node-style [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors) may be used." +
```
timtebeek:
```suggestion
                    "Node-style [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors) may be used. " +
```

timtebeek:
Thanks a lot for the measured change here @DidierLoiseau ; I think this looks mostly good to adopt downstream too indeed.

Still seeing some unrelated test failures, but leaning towards fixing those on main
```
AddPlatformDependencyTest > addPlatformWithExplicitConfiguration(String) > [1] configuration=implementation FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "project/build.gradle":
    diff --git a/project/build.gradle b/project/build.gradle
    index 74140a5..f534d39 100644
    --- a/project/build.gradle
    +++ b/project/build.gradle
    @@ -1,4 +1,5 @@ 
    -plugins {
    +/*~~(org.springframework.pulsar:spring-pulsar-bom:1.0.4 failed. Unable to download POM: org.springframework.pulsar:spring-pulsar-bom:1.0.4. Tried repositories:
    +https://repo.maven.apache.org/maven2/: HTTP 502)~~>*/plugins {
         id "java-library"
     }
 
    ] 
```

timtebeek:
Thanks again! Also already applied in rewrite-java-dependencies: https://github.com/openrewrite/rewrite-java-dependencies/commit/848bc867daad55c6f345a5c335616a46a3209545
Not yet applied in rewrite-testing-frameworks for testcontainers; help appreciated there, otherwise I'll try in the coming days where travel allows.

DidierLoiseau:
@timtebeek thanks a lot, but shouldn’t you release rewrite first? Your change on rewrite-java-dependencies broke the build there.

timtebeek:
No internally we depend on snapshots, until the moment where we start a new release train. That usually works, unless like right now we have a temporary issue publishing snapshots. We'll get that sorted and then at that point the builds should start to succeed as well. 🙏🏻

PR #5963, opened 2025-08-25
Kotlin 2

jkschneider:
## What's changed?

Kotlin 2 compiler internals were substantially changed. Taking another pass at Kotlin language support based on Kotlin 2.

PR #6338, opened 2025-11-25
Use K2Compiler via JvmFirPipeline to support Kotlin v2

### LINKED ISSUE #6007: Support for Kotlin 2.2.0 Projects
So with Openrewrite Gradle Plugin 7.13.0 / 7.15.0 + Kotlin 2.2.0 + Java 21
with Kotlin 2.0 Projects, I am getting multiple errors like below, telling that it cannot parse these files.

So digging in a little deeper with chatgpt ... we've found out that rewrite-kotlin is using Kotlin 1.9
https://github.com/openrewrite/rewrite/blob/main/rewrite-kotlin/build.gradle.kts

**And in turn, downgrading the Project to Kotlin 1.9, makes Openrewrite to work again.**
Unfortunately i cannot share the file(s) in question here .. but it seems like there is an obvious connection.

More simpler Projects with less files work.
But in this Spring BATCH Project Openrewrite basically complains about all files.

So is it possible that current Openrewrite only works with Kotlin 1.9 Projects ?
And is there an Upgrade planned to Kotlin 2.2.0 ?

What is mostly used in this receipe is ChangeType like this

*Changetype*

`  - org.openrewrite.java.ChangeType:
      oldFullyQualifiedTypeName: org.springframework.batch.core.Step
      newFullyQualifiedTypeName: org.springframework.batch.core.step.Step
`

*Error*
`
[org.openrewrite.gradle.isolated.DefaultProjectParser] org.jetbrains.kotlin.util.FileAnalysisException: While analysing somefile.kt:12:51: 
org.jetbrains.kotlin.utils.exceptions.KotlinIllegalArgumentExceptionWithAttachments: 
Unexpected FirClassLikeSymbol null for class org.jetbrains.kotlin.fir.symbols.impl.ConeClassLikeLookupTagImpl
  org.jetbrains.kotlin.util.AnalysisExceptionsKt.wrapIntoFileAnalysisExceptionIfNeeded(AnalysisExceptions.kt:56)
  org.jetbrains.kotlin.fir.FirCliExceptionHandler.handleExceptionOnFileAnalysis(Utils.kt:171)
`

Crustack:
## What's changed?
- [x] Update kotlin-compiler dependencies to `2.2.21`
- [x] Implement `JvmFirPipeline` to use current [K2 phased compiler API](https://github.com/JetBrains/kotlin/tree/1d471395dccaa748f1ae7ea6d02bb0461e57cca8/compiler/cli/src/org/jetbrains/kotlin/cli/pipeline) (with disabled fail on compile errors) to generate `FirFile`s
- [x] Update `KotlinIrTypeMapping`, `IrTreeVisitor`, `KotlinTypeIrSignatureBuilder` to be compilable, but they are not used atm
- [x] Update `KotlinTypeMapping`, `KotlinTypeSignatureBuilder`, `PsiElementAssociations` for new compiler dependency versions to be compilable
- [x] Check all `rewrite-kotlin` and `rewrite-gradle` tests still work and adjust `KotlinTypeMapping`, `KotlinTypeSignatureBuilder`, `PsiElementAssociations` if necessary
- [x] For synthetic source files the contents are pasted into a temp folder to provide compiler with actual file path 

## What's your motivation?
- Closes #6007 
- Groundwork for https://github.com/openrewrite/rewrite/pull/6325 to implement proper kotlin gradle support

@shanman190 
@jkschneider 

### Checklist


Crustack:
@timtebeek @shanman190 @jkschneider PR is now ready for proper review

FYI: I also tested adding support for properly compiling `build.gradle.kts` with Kotlin DSL, which works great and enables proper Kotlin support for many gradle recipes.
I will add another PR for it if/after this has been merged

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -380,153 +348,98 @@ public KotlinParser.Builder clone() {
...
+        Path tempFile = TEMP_SOURCE_DIR.resolve(fileName);
+        try {
+            Files.createDirectories(tempFile.getParent());
+            Files.copy(source.getSource(ctx), tempFile, StandardCopyOption.REPLACE_EXISTING);
+        } catch (FileAlreadyExistsException ignored) {
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return tempFile.toAbsolutePath().toString();
+    }
```
timtebeek, line 377: Calling this out for my colleagues as creating files on disk might be a little different when it concerns the CLI / workers.

rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/KotlinIrTypeMapping.kt
```diff
@@ -550,6 +570,24 @@ class KotlinIrTypeMapping(private val typeCache: JavaTypeCache) : JavaTypeMappin
...
+                    PrimitiveType.BYTE -> JavaType.Primitive.Byte
+                    PrimitiveType.CHAR -> JavaType.Primitive.Char
+                    PrimitiveType.DOUBLE -> JavaType.Primitive.Double
+                    PrimitiveType.FLOAT -> JavaType.Primitive.Float
+                    PrimitiveType.LONG -> JavaType.Primitive.Long
+                    PrimitiveType.SHORT -> JavaType.Primitive.Short
+                    else -> when {
+                        type.isStringClassType() -> JavaType.Primitive.String
+                        type.isNothing() || type.isNullableNothing() -> JavaType.Primitive.None
+                        else -> JavaType.Primitive.None
```
timtebeek, line 587: Love that you've thought to do this as well; it's been a bit of a sore point to me that some Spring migration method patterns failed to match because they were using Java primitives instead of Kotlin primitives.

timtebeek:
> Thanks for your continued efforts here @Crustack ! For the uninitiated (me) could you explain a bit about the what the FIR processing is / adds, and any trade offs as compared to what we had before, and impact on say Kotlin v1 support? That would mean we're better able to review this work, as it seems promising but it's a bit hard to say without more context.

Not sure if you'd seen my earlier comment @Crustack , but it would help us better judge the scope & impact of the work done here, and how it compares to our support for Kotlin v1. Is v1 still supported with your changes for instance?

Crustack:
> Not sure if you'd seen my earlier comment @Crustack , but it would help us better judge the scope & impact of the work done here, and how it compares to our support for Kotlin v1. Is v1 still supported with your changes for instance?

Hi @timtebeek , sorry for not responding earlier.
With the changes of this PR:
- `KotlinParser` still uses PSI+FIR as before, but just with the newer Kotlin Compiler API
- Uses the new K2 Compiler and supports Kotlin v2 (which was already attempted by https://github.com/openrewrite/rewrite/pull/5963, https://github.com/openrewrite/rewrite/pull/5575)
- All Kotlin Tests also pass for e.g. Kotlin v1.9 
- I also figured out how to utilize gradle Kotlin Script Templates with the new compiler API to better parse `build.gradle.kts` files and improve method/field matching for them (which I can submit in a future PR)
- In a next step you could improve Kotlin parsing further by compiling to IR by adding the [JvmFir2IrPipelinePhase](https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/pipeline/jvm/JvmFir2IrPipelinePhase.kt) to `JvmFirPipeline` (see [JvmCliPipeline](https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/pipeline/jvm/JvmCliPipeline.kt#L31))

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
Crustack: I was trying to use the `WorkingDirectoryExecutionContextView`'s working directory to create these temp files (which is only necessary for synthetic temp files in Unit Tests), but executing:
```kotlin
WorkingDirectoryExecutionContextView.view(ctx).getWorkingDirectory()
```
In Unit Tests throws `NullPointerException, because the context has no `CURRENT_CYCLE` message. Not sure how to proceed on this.

timtebeek:
Thanks again for your help here @Crustack ! Youl'll find some of the same changes you've suggested merged in this PR
- https://github.com/openrewrite/rewrite/pull/6766

Hope that has you going with 2.2 support for your use cases there!

Crustack:
To be honest I'm a bit disappointed that over the course of 3 months this PR has not been actually reviewed and then to see an internal PR being created separately and merged immediately.
I would have gladly provided more info or respond to any given feedback.
I understand that internal contributions are generally preferred over outsiders', but since this took a lot of effort, I feel like I could have worked on something else instead if I knew there was no actual interest in it.
But, I will also admit that I should have probably reached out sooner on Slack, I just thought since there was no activity in https://github.com/openrewrite/rewrite/pull/5963 and https://github.com/openrewrite/rewrite/pull/5575 that it would be welcomed...

Thanks to @timtebeek for the initial feedback, and I'm glad to see K2 support being added

Have a nice week.

timtebeek:
Hi @Crustack ; understandable that you're disappointed to not see this explicitly evaluated in depth or merge;d know that it's never easy to decide which of competing PRs to continue from, as we had in this case. You might have noticed that in this case Marius' original work in https://github.com/openrewrite/rewrite/pull/5575 was reflected in [the commits](https://github.com/openrewrite/rewrite/pull/6766/commits). Know that your explorations did help validate the work done in the PR we merged, and was appreciated.

With larger efforts like these it's often hard to get alignment among all involved, and different folks have different styles of working through changes. Pulling work from different branches can than complicate matters when we're looking to get this delivered, but know that it's not our standard way of working.

Should you have any subsequent work you're considering do please reach out, via Slack or here, and I'll do my best to see things through where I can. In this case I myself wasn't confident enough to continue your work, as also evidenced by my earlier comments. But not that support for v2 is here, it should be easier to extend should you find room for improvement. 🙏🏻

timtebeek:
@Crustack based on your hints above I figured explore Kotlin script templates for .kts; figured you'd be interested
- https://github.com/openrewrite/rewrite/pull/6792 
- https://github.com/openrewrite/rewrite/pull/6793

Thanks again!

PR #6339, opened 2025-11-25
Precondition recipe to make a recipe run only once

Laurens-W:
## What's changed?

Introduce a recipe that can be used as precondition to prevent a recipe from running multiple times

## What's your motivation?

Nested recipe cycles may make recipes significantly slower than they have to be

##

### Checklist


rewrite-core/src/main/java/org/openrewrite/RunOnce.java
```diff
@@ -0,0 +1,59 @@
...
+        return "This recipe can be used as a precondition to ensure that the specified recipe only makes changes once per execution cycle.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new TreeVisitor<Tree, ExecutionContext>() {
+
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                for (Recipe recipe : ctx.getCycleDetails().getMadeChangesInThisCycle()) {
```
Laurens-W, line 50: This way if the recipe doesn't make any changes it will still be re attempted.
The recipe stack is not public API of the `RecipeRunCycle`, if it were we could modify the stack to only contain a single instance of the recipe so it's ran exactly once.

jkschneider: I think the more typical desired state with this recipe is run once regardless of whether it made changes.

It is the case of something like removing unnecessary parentheses which might make changes as the result of other recipes that makes this behavior not desirable in every case. But opting in to the singleton-like recipe is an explicit indication of intent.

sambsnyd: As soon as a single file is edited by a recipe it immediately goes into the "made changes this cycle" list.
Consequently, that "highlander" precondition will prevent a recipe from making edits to more than a single file in a run.
It seems like what we're looking for here is a precondition which allows only a single instance of a recipe to make changes for a single cycle. So tracking the fully qualified name is insufficient, it must track the identity of the recipe instance which is allowed to make changes so that other instances may be disallowed.

Image

Issue #4455, opened 2024-08-29
Validation error messages can be ambigious


Bananeweizen:
## What is the smallest, simplest way to reproduce the problem?
Enable these rules, all without parameters
```
org.openrewrite.java.logging.log4j.ParameterizedLogging
org.openrewrite.java.logging.ParameterizedLogging
org.openrewrite.java.logging.slf4j.ParameterizedLogging
```

## What did you see?

```
[ERROR] Recipe validation error in ParameterizedLogging.methodPattern: is required
```

This can be very confusing, when looking up the recipe by (short) name. Only 1 of 3 has parameters. Therefore the user may end up adding the parameters to the wrong recipe, reading the wrong documentation etc.

Even though avoiding the full qualified names is generally a nice idea, it can lead to additional trouble.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Maybe. Depends on which direction a fix would go.

timtebeek:
Hmm; that's a bit unfortunate indeed, but hopefully rare; In this case it's the shortest one that's missing its options; glad at least that we correctly reject that in the validation.

Indeed not sure what to do specifically here, but fully qualifying the recipe name would only help if there are more such identical class names, which I think is reasonably rare, especially not expecting those to be used together (folks use one or the other).

timtebeek:
I could see a case though to always fully qualify; especially with a Spring Boot 3.x best practices containing 2556 recipes, it might make sense to always fully qualify the recipe name.

junseoparkk:
Hi, I’d like to work on this issue.

From the comment, it seems that always showing the fully qualified recipe name in validation error messages might be the preferred direction.  
Would it be okay if I proceed with that approach? If so, could you please assign this issue to me?

Thank you.

timtebeek:
hi @junseoparkk ; appreciate the offer to help! Welcome to get started; we'll assign the issue to you as soon as there is a draft PR. :)

junseoparkk:
Thanks @timtebeek, I’ll start working on this and will open a draft PR soon!

Attyuttam:
Hi, I see this issue to be still open. if this is still an issue, I would like to pick it up.

timtebeek:
Thanks! The validation error is printed in the Maven plugin:
[...]

PR #6340, opened 2025-11-25
fix(validation): use fully-qualified recipe name in validation errors

Vishalup29:
## What's changed?
Validation error messages now display the fully qualified recipe class
name instead of the simple class name.

This resolves ambiguity when multiple recipes have the same short name
(e.g., the various `ParameterizedLogging` recipes), ensuring that users
can clearly see which recipe is missing required parameters.

## What's your motivation?
- Fixes the confusing validation message described in #4455:

Since multiple recipes share the same simple name, users may incorrectly
modify unrelated recipes. Using the fully qualified class name avoids
this ambiguity.

Please confirm this change does not affect any existing expectations
around error formatting, especially in the Maven and Gradle plugins.

@timtebeek (as they provided the suggested fix)

An alternative would be to make error messages include both simple and
fully-qualified names, but using the fully qualified name is aligned
with maintainer guidance and keeps error output clear.

## Any additional context
This is a minimal change (single-line update) but provides clearer,
less ambiguous validation failures.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases (if required)
- [x] I've read and applied the recipe conventions and best practices


timtebeek:
Thanks a lot @Vishalup29 ! We'll merge this as soon as all the tests pass. 🙏🏻

PR #6347, opened 2025-11-26
Track additional Git object information

shanman190:
## What's changed?
Create a new marker to track additional Git object information, presently scoped to Object ID. 

Alternate name: `GitObjectId`.

## What's your motivation?
For Git binary patch support, we have a need to go from `Quark` to `Remote`. In order to generate the `index` line, this requires the old object id and the new object id. For other LST elements, we generate the object id on the fly by generating the sha1 of the raw printed bytes. For `Quark`, this isn't possible and needs a different solution.

N/A

Anybody

We could have added the object id as a field on `Quark` or `SourceFile`, but it was preferred to utilize a dedicated marker instead.

## Any additional context
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/422
- https://github.com/openrewrite/rewrite-maven-plugin/pull/1081

Another long standing issue is related to file mode. The current implementation of `FileAttributes` is very lossy with respect to this information. A possibility would be to capture the `FileMode#getBits()` from JGit since it would cost O(1) to retrieve and add to this marker. Doing so, I'd probably suggest that we change the name of the marker to something else, such as `GitTreeEntry` to be more accurate of it's purpose.

Additionally, on Windows, `Files#isExecutable(Path)` always returns true. So this can manifest in the Git patch later as a file being marked as executable when it normally wouldn't be due to this JVM quirk.

### Checklist


rewrite-core/src/main/java/org/openrewrite/marker/FileMode.java
```diff
@@ -1,17 +1,25 @@
...
  */
+package org.openrewrite.marker;
 
-export {UseObjectPropertyShorthand} from "./use-object-property-shorthand";
+import lombok.Value;
+import lombok.With;
+
+@Value
+@With
+public class FileMode {
```
shanman190: Idea with introducing this now is the following:

1. Store unwrapped as an int in it's serialized form
2. Have a value object here such that when we want to begin manipulating the mode, we can centralize the logic here on this value object.

shanman190:
These JS tests are known to be flaky, but are unrelated to the changes here.

```
FAIL test/javascript/add-import.test.ts (122.281 s)
  ● AddImport visitor › object imports (non-function references) › should add import for vitest vi object when referenced

    thrown: "Exceeded timeout of 30000 ms for a test.

    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      1351 |
      1352 |     describe('object imports (non-function references)', () => {
    > 1353 |         test('should add import for vitest vi object when referenced', async () => {
           |             ^
      1354 |             const spec = new RecipeSpec();
      1355 |             spec.recipe = fromVisitor(createRemoveThenAddImportVisitor("vitest", "vi"));
      1356 |

      at test/javascript/add-import.test.ts:1353:13
      at test/javascript/add-import.test.ts:1352:13
      at Object.<anonymous> (test/javascript/add-import.test.ts:163:9)

FAIL test/javascript/node-resolution-result.test.ts (96.013 s)
  ● NodeResolutionResult marker › should attach NodeResolutionResult marker to package.json

    thrown: "Exceeded timeout of 30000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      33 | describe("NodeResolutionResult marker", () => {
      34 |
    > 35 |     test("should attach NodeResolutionResult marker to package.json", async () => {
         |     ^
      36 |         const spec = new RecipeSpec();
      37 |         await withDir(async (repo) => {
      38 |             await spec.rewriteRun(

      at test/javascript/node-resolution-result.test.ts:35:5
      at Object.<anonymous> (test/javascript/node-resolution-result.test.ts:33:9)
```

shanman190:
Interestingly enough, apparently Git only supports a couple of the posix file modes:

* Regular non-executable file (100644)
* Regular executable file (100755)
* Symbolic link (120000)
* Directory (040000)
* Gitlink/submodule (160000)

All other variations are reduced to their closest match.

This makes the file mode portion a little less useful, but does situate us slightly better with respect to being able to understand the difference between a regular file and a symlink. It also could be the potential foundation needed to model submodules.

rewrite-core/src/main/java/org/openrewrite/marker/FileMode.java
jkschneider: I think the JGit design makes most sense to me, to have `rawMode` stored on the tree and a separate enum to break it down into its 5 or 6 coarse types:

```
int bits = treeWalk.getRawMode(0);
FileMode mode = FileMode.fromBits(bits);
```

I suppose a lot of use cases involve only switching over that enum but the raw mode is needed to preserve the original value in granular detail.

What I'm thinking is even if we introduce `FileMode` here, we'd wind up having `FileMode.Type` enum or something that's converting its single raw field into an enum and then much of business logic would wind up being there, leaving no real logic to remain on the outer class. So we might as well just have the raw mode int stored on the tree and a `FileMode` enum similar to the JGit design. Or folks just use the JGit `FileMode` enum when they need it, which is created from the raw mode anyway.

PR #6348, opened 2025-11-27
Fix Groovy/Gradle parsing for GString-s where the interpolation prefix `'$'` is preceded by an even number of backslashes

akiraly:
Fix Groovy/Gradle parsing for GString-s where the interpolation prefix `'$'` is preceded by an even number of backslashes like in `def a = "\\${System.getProperty('user.name')}"` (in practice this happens often in gradle scripts dealing with windows paths).

## What's changed?
GroovyParserVisitor was changed to fix a GString related parsing issue related to escaping.
Some explicit exceptions were also added to help investigations in the future.

## What's your motivation?
To unblock Gradle recipes from being applied on Gradle build files that show this issue.

Note that I changed the semantics of the `isEscaped()` private method slightly. This change allowed me to move an index check into the method as well (which previously had to be done at the calling side). This method was originally only used from one place, which was updated. This PR calls it from a second place as well.

### Checklist

#### Legal disclaimer:

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1748,7 +1748,7 @@ public class GroovyParserVisitor {
                     String value = sourceSubstring(cursor, delimiter.close);
                     // There could be a closer GString before the end of the closing delimiter, so shorten the string if needs be
                     int indexNextSign = source.indexOf("$", cursor);
-                    while (indexNextSign > 0 && source.charAt(indexNextSign - 1) == '\\') {
+                    while (isEscaped(indexNextSign)) {
                         indexNextSign = source.indexOf("$", indexNextSign + 1);
                     }
                     if (indexNextSign != -1 && indexNextSign < (cursor + value.length())) {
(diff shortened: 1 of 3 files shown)


rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
```diff
@@ -2763,12 +2763,23 @@ private boolean isEscaped(int index) {
      * The cursor will not be moved.
      */
     private String sourceSubstring(int beginIndex, String untilDelim) {
-        int endIndex = source.indexOf(untilDelim, Math.max(beginIndex, cursor + untilDelim.length()));
-        // don't stop if last char is escaped.
-        // Fixed potential infinite loop by correctly handling escaped delimiters in the source string.
-        while (endIndex > 0 && isEscaped(endIndex - 1)) {
+        int fromIndex = Math.max(beginIndex, cursor + untilDelim.length());
+        int endIndex = source.indexOf(untilDelim, fromIndex);
+        if (endIndex < 0) {
```
akiraly, line 2768: Previously a negative `endIndex` triggered a `StringIndexOutOfBoundsException` at the `.substring` call. IMHO an explicit exception would have been more helpful in this case.

PR #6350, opened 2025-11-27
Fixed autoformatting of groovy / js / kotlin files

Jenson3210:
MergeSpacesVisitors -> need to call to their own visitContainers and not super as they are responsible of going into the "ExecutionContext" lst element

JS -> do not use java formatter as JS should have its own formatting service

### DIFF
--- a/rewrite-javascript/src/main/java/org/openrewrite/javascript/service/JavaScriptAutoFormatService.java
+++ b/rewrite-javascript/src/main/java/org/openrewrite/javascript/service/JavaScriptAutoFormatService.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2023 the original author or authors.
+ * Copyright 2025 the original author or authors.
  * <p>
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Moderne Source Available License (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  * <p>
- * https://www.apache.org/licenses/LICENSE-2.0
+ * https://docs.moderne.io/licensing/moderne-source-available-license
  * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,


rewrite-javascript/src/main/java/org/openrewrite/javascript/service/JavaScriptAutoFormatService.java
```diff
@@ -0,0 +1,30 @@
...
+import org.openrewrite.Tree;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.service.AutoFormatService;
+
+public class JavaScriptAutoFormatService extends AutoFormatService {
+
+    @Override
+    public <P> JavaVisitor<P> autoFormatVisitor(@Nullable Tree stopAfter) {
+        return new JavaIsoVisitor<>();
```
knutwannheden: We might as well override the `visit()` method here so that it returns the input tree immediately without traversing it.

rewrite-javascript/src/main/java/org/openrewrite/javascript/service/JavaScriptAutoFormatService.java
```diff
@@ -0,0 +1,30 @@
...
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
knutwannheden, line 15: Let's make sure we use the source available license here. At least I think we do that for all files in this subproject.

Jenson3210: Thanks, I had no idea we had different licenses in use in the core module!

rewrite-javascript/src/main/java/org/openrewrite/javascript/service/JavaScriptAutoFormatService.java
Jenson3210: adapted!

Issue #6053, opened 2025-09-18
Add support for "contains" in XML XPath matching


BrendanHart:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.62.2 (latest in main)

## What is the smallest, simplest way to reproduce the problem?

```java
    @Test
    void searchWithContains() {
        rewriteRun(
          spec -> spec.recipe(new FindTags("/dependencies/dependency[contains(artifactId, 'abc')]")),
          xml(
            """
            <dependencies>
                <dependency>
                    <artifactId>org.openrewrite</artifactId>
                </dependency>
            </dependencies>
            """,
            """
              <dependencies>
                  <!--~~>--><dependency>
                      <artifactId>org.openrewrite</artifactId>
                  </dependency>
              </dependencies>
              """
          )
        );
    }
```

## What did you expect to see?

Test fails since artifactId does not contain abc.

## What did you see instead?

Test passes.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes

timtebeek:
Thanks for the offer to help! Indeed for the moment we support a limited set of XPath methods, as seen here:
https://github.com/openrewrite/rewrite/blob/8ce5308a9557dd6ca91253a5c8fb7e6c34a0ef5a/rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java#L32-L41

If we can safely and without a performance hit add support for `contains` then that could be a good addition indeed.

BrendanHart:
@timtebeek do we feel like generally enhancing the XPath implementation might be a good idea? Right now it's heavily based on regular expressions, maybe spending the time to use ANTLR or something to parse the XPath expressions would make implementation simpler, without introducing extra dependencies since ANTLR is already being used for XML?

timtebeek:
hi! Thanks for the suggestion; I feel @knutwannheden might best weigh in here about the relative merits and expected performance, as he's recently gone in the opposite direction for our method matchers in
- https://github.com/openrewrite/rewrite/pull/6082

knutwannheden:
[...]

Issue #5028, opened 2025-02-12
Support positional elements in XPathMatcher


alvarosanchez:
## What problem are you trying to solve?

In a pom.xml, remove a duplicate `jacoco-maven-plugin` build plugin definition with `org.openrewrite.xml.RemoveXmlTag`.

## Describe the solution you'd like

Use an XPath expression like `/project/build/plugins/plugin[artifactId='jacoco-maven-plugin'][2]` or `/project/build/plugins/plugin[artifactId='jacoco-maven-plugin'][last()]`.

`/project/build/plugins/plugin[artifactId='jacoco-maven-plugin']` works and matches both definitions. I want to match just one of them.

## Have you considered any alternatives or workarounds?

There aren't, to the best of my knowledge

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

`XPathMatcher` has some pretty complicated logic, but if pointed to the appropriate place, I can try.

PR #6351, opened 2025-11-27
Allow subselectors in XPath Function Conditions

Laurens-W:
## What's changed?

Update XPathMatcher to accommodate subselectors on FunctionConditions

## What's your motivation?

- CI for rewrite-spring failing after https://github.com/openrewrite/rewrite/pull/6315

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/format/MergeSpacesVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/format/MergeSpacesVisitor.java
@@ -1605,6 +1605,27 @@ public class MergeSpacesVisitor extends GroovyVisitor<Object> {
     }
 
     @Override
+    public <T> JRightPadded<T> visitRightPadded(@Nullable JRightPadded<T> right, GRightPadded.Location loc, Object p) {
+        return visitRightPadded(right, JRightPadded.Location.LANGUAGE_EXTENSION, p);
+    }
+
+    @Override
+    public <T> JLeftPadded<T> visitLeftPadded(JLeftPadded<T> left, GLeftPadded.Location loc, Object p) {
+        return visitLeftPadded(left, JLeftPadded.Location.LANGUAGE_EXTENSION, p);
+    }
+
+    @Override
+    public Space visitSpace(Space space, GSpace.Location loc, Object p) {
+        return visitSpace(space, Space.Location.LANGUAGE_EXTENSION, p);
+    }
+
+    @Override
+    public <J2 extends J> JContainer<J2> visitContainer(JContainer<J2> container,
+                                                        GContainer.Location loc, Object p) {
+        return visitContainer(container, JContainer.Location.LANGUAGE_EXTENSION, p);
+    }
+
+    @Override
     public <T> @Nullable JRightPadded<T> visitRightPadded(@Nullable JRightPadded<T> right, JRightPadded.Location loc, @Nullable Object ctx) {
         JRightPadded<T> newRight = (JRightPadded<T>) ctx;
         if (right == newRight) {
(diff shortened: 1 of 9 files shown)


rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
```diff
@@ -275,11 +275,11 @@ private boolean matchesWithoutDoubleSlashesAt(Cursor cursor, int doubleSlashInde
             while (condition.find() && (stillMatchesConditions || orCondition)) {
                 boolean matchCurrentCondition = false;
 
-                boolean isAttributeCondition = condition.group(4) != null;
```
Jenson3210, line 278: Open suggestion and definitely not required to adapt, just want your 2 cents on this. 

Instead of updating all these groups, we could also make the groups we do not need non-capturing groups and then have only the group indexes of capturing groups. 

Another option would be to use named groups which I tend to like when dealing with multiple parts of interest. By naming these groups in the code, the `condition.group(int)` could become `condition.group("somedescriptivename")`

rewrite-xml/src/test/java/org/openrewrite/xml/XPathMatcherTest.java
```diff
@@ -94,7 +94,7 @@ class XPathMatcherTest {
     ).toList().getFirst();
 
     private final SourceFile namespacedXml = new XmlParser().parse(
-      """
+      """
```
Jenson3210: This eol space might not be something you intended?

rewrite-xml/src/main/java/org/openrewrite/xml/XPathMatcher.java
Laurens-W: In this case since I added a group near the front it would have still shifted a few indexes, but I agree only capturing what we actually need would make this easier to follow.

I like the idea of named groups but that'll bloat the regex quite a bit if we give them sensical names.

My preference would be to reduce the amount of groups captured and leave it at that as there are tools like regex101 to view what the intent of the groups is for a given string

rewrite-xml/src/test/java/org/openrewrite/xml/XPathMatcherTest.java
```diff
@@ -421,12 +421,22 @@ void matchTextFunctionCondition() {
...
+                </bar>
+              </foo>
               <foo/>
             </test>
             """
         ).toList().getFirst();
 
         // text() predicate should only match element with specific text content
+        assertThat(match("/test/foo[bar/baz/text()='bla']", xml)).isTrue();
+        assertThat(match("/test/foo[bar/text()='bla']", xml)).isTrue();
```
timtebeek, line 439: If I understand correctly previously we were a little too eager to match, then too restrictive, and now it should be just right. It's hard to distinguish those cases here though with the repeated use of `<foo>`. Would it make sense to use more descriptive and separate tags for the various cases? some tag names that show what's at the top level, vs nested for instance, and then patterns that use that descriptively.

timtebeek:
Looks like there's a parallel effort as well:
- https://github.com/openrewrite/rewrite/pull/6353

Jenson3210:
I guess we would still want to get this fix / test merged as u assume getting the antlr approach in will take bit longer? To unblock spring releases.

rewrite-xml/src/test/java/org/openrewrite/xml/XPathMatcherTest.java
greg-at-moderne: Also I suggest to add an assertion for `text()='notbla' - it shouldn't be a match.

knutwannheden:
> Looks like there's a parallel effort as well:
> 
> * [`XPath` ANTLR grammar #6353](https://github.com/openrewrite/rewrite/pull/6353)

I worked a bit more on this and I think the implementation should be in a fairly good state. In my tests the matching is now around 3-10x faster and practically allocation free (not causing any GC pressure). At the same time it also supports a lot more XPath features and makes it fairly easy to add other features. These should be covered:
- https://github.com/openrewrite/rewrite/issues/6053
- https://github.com/openrewrite/rewrite/issues/5028

Laurens-W:
https://github.com/openrewrite/rewrite/pull/6353 is more comprehensive and covers all XPath functionality, the tests added here were transferred to the other PR as well. I'll adjust them there with more distinguishable names and negative tests

rewrite-xml/src/test/java/org/openrewrite/xml/XPathMatcherTest.java
Laurens-W: https://github.com/openrewrite/rewrite/pull/6353/commits/f31ef7da12d53761b5789390da9f919eaccd492e

Issue #6322, opened 2025-11-21
UpgradeDependencyVersion does not correctly update BOM imports if they overlap


MrThaler:
## What version of OpenRewrite are you using?

I am using

- rewrite-recipe-bom 3.18.0

## How are you running OpenRewrite?

Writing my own recipes, but using JUnit tests to confirm the behavior observed "in the field".

## What is the smallest, simplest way to reproduce the problem?

In this test we have two runs:

- In the first run we only upgrade `camunda-bom` to 7.3.0 where it has no overlap with `camunda-engine-dmn-bom`
- In the second run we upgrade `camunda-bom` to 7.20.0 where it has 4 entries in common with `camunda-engine-dmn-bom` (camunda-engine-dmn, camunda-engine-feel-api, camunda-engine-feel-juel and camunda-engine-feel-scala).

```java
  @ParameterizedTest
  @ValueSource(strings={"7.3.0","7.20.0"})
  void sequentialRuns(final String targetVersion){
      rewriteRun(
        spec ->
            spec.recipes(
                new UpgradeDependencyVersion(
                    "org.camunda.bpm", "camunda-bom", targetVersion, null, null, null),
                new UpgradeDependencyVersion(
                    "org.camunda.bpm.dmn", "camunda-engine-dmn-bom", "7.20.0", null, null, null)),
        pomXml("""
              <project>
                <groupId>group.id</groupId>
                <artifactId>artifactId</artifactId>
                <version>1.0.0</version>
  
                <dependencyManagement>
                  <dependencies>
                    <dependency>
                      <groupId>org.camunda.bpm</groupId>
                      <artifactId>camunda-bom</artifactId>
                      <version>7.2.0</version>
                      <type>pom</type>
                      <scope>import</scope>
                    </dependency>
                    <dependency>
                      <groupId>org.camunda.bpm.dmn</groupId>
                      <artifactId>camunda-engine-dmn-bom</artifactId>
                      <version>7.10.0</version>
                      <type>pom</type>
                      <scope>import</scope>
                    </dependency>
                  </dependencies>
                </dependencyManagement>
              </project>
              """,
              """
              <project>
                <groupId>group.id</groupId>
                <artifactId>artifactId</artifactId>
                <version>1.0.0</version>
  
                <dependencyManagement>
                  <dependencies>
                    <dependency>
                      <groupId>org.camunda.bpm</groupId>
                      <artifactId>camunda-bom</artifactId>
                      <version>%s</version>
[...]

PR #6355, opened 2025-11-29
Fix UpgradeDependencyVersion stabilization with multiple managed dependencies

### LINKED ISSUE #6321: UpgradeDependencyVersion does not correctly stabilize when using Wildcards with multiple dependencyManagement entries

I am using

- rewrite-recipe-bom 3.18.0

Writing my own recipe, trying to reproduce odd behavior in JUnit tests.

```java
@Test
    void multipleDependencyManagementEntriesTest() {
    rewriteRun(
        spec ->
            spec.recipe(new UpgradeDependencyVersion("*", "*", "latest.minor", null, null, null)),
        pomXml(
            """
                          <project>
                              <groupId>org.openrewrite.example</groupId>
                              <artifactId>my-app-server</artifactId>
                              <version>1</version>
                              <dependencyManagement>
                                  <dependencies>
                                      <dependency>
                                          <groupId>org.junit.jupiter</groupId>
                                          <artifactId>junit-jupiter-api</artifactId>
                                          <version>5.2.0</version>
                                      </dependency>
                                      <dependency>
                                          <groupId>org.slf4j</groupId>
                                          <artifactId>slf4j-api</artifactId>
                                          <version>1.7.33</version>
                                      </dependency>
                                  </dependencies>
                              </dependencyManagement>
                          </project>
                          """,
            """
                          <project>
                              <groupId>org.openrewrite.example</groupId>
                              <artifactId>my-app-server</artifactId>
                              <version>1</version>
                              <dependencyManagement>
                                  <dependencies>
                                      <dependency>
                                          <groupId>org.junit.jupiter</groupId>
                                          <artifactId>junit-jupiter-api</artifactId>
                                          <version>5.14.1</version>
                                      </dependency>
                                      <dependency>
                                          <groupId>org.slf4j</groupId>
                                          <artifactId>slf4j-api</artifactId>
                                          <version>1.7.36</version>
                                      </dependency>
                                  </dependencies>
                              </dependencyManagement>
                          </project>
                          """));
    }
```

The test should succeed. Both dependencyManagement entries should be updated to the most recent version within in their major version.

## What did you see instead?

The tests fail because UpgradeDependencyVersion takes multiple cycles to stabilize.
However, when increasing the cycles the resulting pom.xml does seem to be correct.

## What steps are needed to reproduce the behavior?

As far as I can see this issue does not prevent when the dependencyManagement consists of (for example) an imported BOM and a regular entry [such as in the "upgradeAllManagedDependenciesToPatchReleases" test](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java#L1247).

timtebeek:
## Summary
- Fixes #6321
- Removes premature `maybeUpdateModel()` call when upgrading non-BOM managed dependencies
- The call was happening before the scheduled visitor had run, causing state inconsistencies when processing multiple managed dependencies with wildcards

## What's the problem?

When using `UpgradeDependencyVersion("*", "*", "latest.minor", ...)` with a POM containing multiple `dependencyManagement` entries, the recipe required multiple cycles to stabilize. This was because `maybeUpdateModel()` was called immediately after scheduling the version change via `doAfterVisit()`, but before the actual change was applied.

## The fix

Remove the `maybeUpdateModel()` call from the managed dependency handling block. The model will be updated naturally when the scheduled visitors run and modify the tree.

## Test plan
- [x] Added test `upgradeMultipleManagedDependenciesWithWildcards()` that verifies multiple managed dependencies can be upgraded with wildcards in a single cycle
- [x] All existing `UpgradeDependencyVersionTest` tests pass


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -255,7 +255,6 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                             TreeVisitor<Xml, ExecutionContext> upgradeManagedDependency = upgradeManagedDependency(tag, ctx, t);
                             if (upgradeManagedDependency != null) {
                                 doAfterVisit(upgradeManagedDependency);
-                                maybeUpdateModel();
```
timtebeek, line 258: It's also called on line 268 below.

timtebeek:
This PR has been updated to also 
- fix #6322.

**Additional changes for #6322:**
- Handle BOM imports directly from the tag rather than through `findManagedDependency()`
- This prevents incorrect matching when a BOM import has the same groupId/artifactId as a dependency managed by another BOM
- Added test `upgradeMultipleBomImportsWithWildcards()` to verify the fix

The key insight for #6322 was that `findManagedDependency()` could incorrectly match a dependency managed by the first BOM when processing the second BOM import tag, if both have overlapping dependencies. By checking for BOM imports first and reading the GAV directly from the tag, we avoid this incorrect matching.

timtebeek:
Reverted to the earlier more minimal change here to make it easier to get this small fix in already.

PR #6360, opened 2025-12-01
Rework dependency graph to be more efficient

shanman190:
## What's changed?
Update dependency graph generation to reduce overhead and only generate for what is needed.

## What's your motivation?
Reduce GC pressure when generating the dependency graph.

N/A

@pstreef @bmuschko 

N/A

## Any additional context
- Additionally observed that some dependencies that should have matched were not being recorded. This has now been fixed.
- Because we were using a `HashSet` to hold the discovered results, this meant that the order could vary. To resolve this, I went ahead and sorted the results to restore consistent behavior.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/InsertDependencyComparator.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/InsertDependencyComparator.java
@@ -25,29 +25,50 @@ public class InsertDependencyComparator implements Comparator<Content> {
     private final Map<Content, Float> positions = new LinkedHashMap<>();
 
     public InsertDependencyComparator(List<? extends Content> existingDependencies, Xml.Tag dependencyTag) {
-        for (int i = 0, existingDependenciesSize = existingDependencies.size(); i < existingDependenciesSize; i++) {
-            positions.put(existingDependencies.get(i), (float) i);
-        }
+        // Group comments with their following dependency tags
+        Map<Xml.Tag, List<Content>> dependencyWithPrecedingComments = new LinkedHashMap<>();
+        List<Content> currentComments = new ArrayList<>();
 
-        // Make a copy of statements with only dependencies
-        List<Xml.Tag> dependencies = new ArrayList<>();
-        for (Content c : existingDependencies) {
-            if (c instanceof Xml.Tag) {
-                dependencies.add((Xml.Tag) c);
+        for (int i = 0; i < existingDependencies.size(); i++) {
+            Content content = existingDependencies.get(i);
+            if (content instanceof Xml.Tag) {
+                dependencyWithPrecedingComments.put((Xml.Tag) content, new ArrayList<>(currentComments));
+                currentComments.clear();
+            } else {
+                currentComments.add(content);
             }
         }
 
+        // Assign initial positions based on current order, keeping comments with their dependencies
+        float position = 0f;
+        for (Map.Entry<Xml.Tag, List<Content>> entry : dependencyWithPrecedingComments.entrySet()) {
+            // Position the comments just before their dependency
+            for (Content comment : entry.getValue()) {
+                positions.put(comment, position);
+                position += 0.1f;
+            }
+            positions.put(entry.getKey(), position);
+            position += 1.0f;
+        }
+
+        for (Content comment : currentComments) {
+            positions.put(comment, position);
+            position += 0.1f;
+        }
+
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
```diff
@@ -36,13 +36,14 @@
...
 import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
+import org.openrewrite.semver.DependencyMatcher;
 import org.openrewrite.semver.Semver;
 import org.openrewrite.semver.VersionComparator;
 
 import java.util.*;
 import java.util.function.Function;
 
-import static java.util.Collections.singletonList;
+import static java.util.Collections.newSetFromMap;
```
pstreef: seems unused

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
```diff
@@ -172,231 +167,86 @@ private boolean matchesConfiguration(GradleDependencyConfiguration c) {
...
+                    String sourceSetName,
+                    String configurationName,
+                    DependencyMatcher dependencyMatcher,
+                    ResolvedDependency resolvedDependency,
+                    Map<String, Set<GroupArtifactVersion>> configurationToDirectDependency,
+                    Map<GroupArtifactVersion, Set<GroupArtifactVersion>> directDependencyToTargetDependency,
+                    Deque<ResolvedDependency> deque,
+                    ExecutionContext ctx
+            ) {
+                if (deque.contains(resolvedDependency)) {
```
pstreef: hash comparison might be a little faster here, but probably neglectable.

rewrite-maven/src/main/java/org/openrewrite/maven/search/DependencyInsight.java
```diff
@@ -114,95 +121,158 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                        dependencyPath.size() - 1,
+                        dependencyGraph
                 ));
+            }
+        };
+    }
+
+    @EqualsAndHashCode(callSuper = false)
+    @RequiredArgsConstructor
+    @Data
```
pstreef: Seems like `@Value` would fit here better?

knutwannheden:
> `rewrite-java-security` is reusing the graph building logic. By removing those classes you deleted (by inlining the logic), it will automatically break the consuming module.

Perhaps some kind of `DependencyTreeWalker` utility could be added on top of the changes made here:
- https://github.com/openrewrite/rewrite/pull/6363

For `rewrite-java-security` the usage would be something like:

```java
  // Store paths as Deque (preserves leaf-to-root order)
  Map<ResolvedGroupArtifactVersion, List<Deque<ResolvedDependency>>> paths = new LinkedHashMap<>();

  // Walk all dependencies (matcher=null), copy path since deque is transient
  DependencyTreeWalker.walk(dependencies, null, (dep, path) ->
      paths.computeIfAbsent(dep.getGav(), __ -> new ArrayList<>())
           .add(new ArrayDeque<>(path)));

  // Later, render the path as an inverse tree for display
  Deque<ResolvedDependency> firstPath = paths.get(gav).get(0);
  String dependencyGraph = DependencyTreeWalker.renderPath("compile", firstPath);
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/search/DependencyInsight.java
knutwannheden: Using `LinkedHashSet` would indeed be an option. But as we are on Java 1.8 we can't change the APIs to return a `SequencedCollection` which would have a `reversed()` view, so getting a reverse iterator is not really possible and we can only append to the end with `LinkedHashSet`.

Since dependencies typically don't run very deep (20 levels?), using an ArrayDeque should be fine, I think.

shanman190: `contains` using a Hash implementation would probably be a little bit faster for the purpose of checking for a recursive dependency path. However, the tradeoffs aren't very ideal for this particular use case.

There's also getting the head and tail as well when creating the dependency path to record additional information. A `Linked*` structure generally has worse performance for reading through the elements (and adds additional GC pressure for removes which would occur frequently during the dependency graph traversal itself).

For less than 10k total items, the performance profile is approximately the same. And with a dependency graph, we're unlikely to see that kind of depth.

PR #6366, opened 2025-12-02
Changing `doAfterVisit(new ChangeTagValueVisitor<>(..)` in Maven `AddProperty` to be direct invocation instead to trigger model update

steve-aom-elliott:
## What's changed?
Changed the usage of the `ChangeTagValueVisitor` to be directly invoking it rather than using `doAfterVisit(..)` when changing of a tag value in `AddProperty`.

## What's your motivation?
Previously, if you were to run the `AddProperty` recipe in a way that would update a value rather than add one, subsequent `AddProperty` or other recipes would still see the old value in the resolution result, causing recipes that need to see the result of the initial `AddProperty` to receive inaccurate information.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddPropertyVisitor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddPropertyVisitor.java
@@ -60,8 +60,7 @@ public class AddPropertyVisitor extends MavenIsoVisitor<ExecutionContext> {
             String tagValue = tag.getValue().orElse(null);
             // Only update if values are different, considering null values
             if (!Objects.equals(value, tagValue)) {
-                doAfterVisit(new ChangeTagValueVisitor<>(tag, value));
-                maybeUpdateModel();
+                return (Xml.Tag) new ChangeTagValueVisitor<>(tag, value).visitNonNull(tag, ctx);
             }
         }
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddPropertyVisitor.java
```diff
@@ -61,6 +61,7 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
             // Only update if values are different, considering null values
             if (!Objects.equals(value, tagValue)) {
                 doAfterVisit(new ChangeTagValueVisitor<>(tag, value));
+                maybeUpdateModel();
```
sambsnyd: Bugs like the one fixed in this PR are why I have a preference for directly applying visitors inline rather than calling `doAfterVisit()`. If this visitor were applied immediately, then the `if (d != document) { maybeUpdateModel() }` in `visitDocument` would be sufficient

steve-aom-elliott: Just pushed a change for this. I think I considered it at one point but had ended up falling back to the smaller code change when I was trying to get the test setup correct so it didn't find extra cycles.

PR #6368, opened 2025-12-03
Added SearchResults as a global datatable

Jenson3210:
# Add SearchResults as a Global DataTable

## Summary
This PR introduces `SearchResults` as a new global data table that captures search markers added by recipes during execution, providing visibility into what search results were found across the codebase during a recipe run.

## Changes
- **New `SearchResults` data table**: Captures search markers added during recipe runs, recording:
  - Source path where the result was found
  - The printed LST element with the search marker
  - Parent recipe and recipe that added the marker

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -245,11 +245,9 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                 0L : recipe.getEstimatedEffortPerOccurrence().getSeconds();
 
         String parentName = "";
-        String parentInstanceName = "";
         boolean hierarchical = recipeStack.size() > 1;
         if (hierarchical) {
             parentName = recipeStack.get(recipeStack.size() - 2).getName();
-            parentInstanceName = recipeStack.get(recipeStack.size() - 2).getInstanceName();
         }
         sourcesFileResults.insertRow(ctx, new SourcesFileResults.Row(
                 beforePath,
@@ -261,28 +259,25 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
 
         List<String> searchMarkers = collectSearchResults(before, after);
         for (String searchResult : searchMarkers) {
-            searchResults.insertRow(ctx, new SearchResults.Row(beforePath, afterPath, searchResult, parentInstanceName, recipe.getInstanceName()));
+            searchResults.insertRow(ctx, new SearchResults.Row(beforePath, afterPath, searchResult, recipe.getInstanceName()));
         }
 
         if (hierarchical) {
-            recordSourceFileResult(beforePath, afterPath, recipeStack.subList(0, recipeStack.size() - 1), effortSeconds, searchMarkers, ctx);
+            recordSourceFileResult(beforePath, afterPath, recipeStack.subList(0, recipeStack.size() - 1), effortSeconds, ctx);
         }
     }
 
-    private void recordSourceFileResult(@Nullable String beforePath, @Nullable String afterPath, List<Recipe> recipeStack, Long effortSeconds, List<String> searchMarkers, ExecutionContext ctx) {
+    private void recordSourceFileResult(@Nullable String beforePath, @Nullable String afterPath, List<Recipe> recipeStack, Long effortSeconds, ExecutionContext ctx) {
         if (recipeStack.size() <= 1) {
             // No reason to record the synthetic root recipe which contains the recipe run
             return;
         }
         String parentName;
-        String parentInstanceName;
         if (recipeStack.size() == 2) {
             // Record the parent name as blank rather than CompositeRecipe when the parent is the synthetic root recipe
             parentName = "";
-            parentInstanceName = "";
         } else {
             parentName = recipeStack.get(recipeStack.size() - 2).getName();
-            parentInstanceName = recipeStack.get(recipeStack.size() - 2).getInstanceName();
         }
         Recipe recipe = recipeStack.get(recipeStack.size() - 1);
         sourcesFileResults.insertRow(ctx, new SourcesFileResults.Row(
@@ -292,10 +287,7 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                 recipe.getName(),
                 effortSeconds,
                 cycle));
-        for (String searchResult : searchMarkers) {
-            searchResults.insertRow(ctx, new SearchResults.Row(beforePath, afterPath, searchResult, parentInstanceName, recipe.getInstanceName()));
-        }
-        recordSourceFileResult(beforePath, afterPath, recipeStack.subList(0, recipeStack.size() - 1), effortSeconds, searchMarkers, ctx);
+        recordSourceFileResult(beforePath, afterPath, recipeStack.subList(0, recipeStack.size() - 1), effortSeconds, ctx);
     }
 
     private @Nullable SourceFile handleError(Recipe recipe, SourceFile sourceFile, @Nullable SourceFile after,
--- a/rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
+++ b/rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
@@ -44,15 +44,8 @@ public class SearchResults extends DataTable<SearchResults.Row> {
                 description = "The trimmed printed tree of the LST element that the marker is attached to.")
         String result;
 
-        @Column(displayName = "Parent of the recipe that had the search marker added",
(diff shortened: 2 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -347,4 +362,36 @@ private static boolean isScanningRequired(Recipe recipe) {
         }
         return false;
     }
+
+    private List<String> collectSearchResults(@Nullable SourceFile before, @Nullable SourceFile after) {
+        if (after == null) {
+            return emptyList();
+        }
+        Set<SearchResult> alreadyPresentMarkers = new TreeVisitor<Tree, Set<SearchResult>>() {
```
Jenson3210, line 359: Do we see any other option instead of going over every sourcefile twice (before and after) PER recipe to get the same results here?

rewrite-core/src/test/java/org/openrewrite/table/SearchResultsTest.java
```diff
@@ -0,0 +1,248 @@
...
+import static org.assertj.core.groups.Tuple.tuple;
+import static org.openrewrite.test.SourceSpecs.text;
+
+class SearchResultsTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipeFromYaml(
+          //language=yml
+          """
```
jkschneider, line 32: Template string indentation is off here. Terminating """ should be one indent left to the right of starting """

rewrite-core/src/test/java/org/openrewrite/table/SearchResultsTest.java
```diff
@@ -0,0 +1,248 @@
...
+              - org.openrewrite.text.Find:
+                  find: ([A-Z])\\w+
+                  regex: true
+                  caseSensitive: true
+            """,
+          "test.SearchMarkerScraping"
+        );
+    }
+
+    @DocumentExample
```
jkschneider: This is not valid here.

rewrite-core/src/test/java/org/openrewrite/table/SearchResultsTest.java
```diff
@@ -0,0 +1,248 @@
...
+            .containsExactlyInAnyOrder(
+              tuple("different-recipes-matching", "We", "Find text and add the results to the global datatable", "Find text `([A-Z])\\w+`"),
+              tuple("different-recipes-matching", "SearchMarkers", "Find text and add the results to the global datatable", "Find text `([A-Z])\\w+`"),
+              tuple("different-recipes-matching", "some other text", "Find text and add the results to the global datatable", "Find text `SOME OTHER TEXT`")
+            )),
+          text(
+            """
+              We add all SearchMarkers that are found during recipe run.
+              file contains some other text somewhere in the middle resulting in 2 different recipes matches.
+              """, """
```
jkschneider: As an OR standard we put new """ on a separate line.

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
jkschneider: I'm not concerned with this, since this should only be called if the recipe made a change. Even in a deeply nested composite of thousands of recipes, most will not make changes.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
```diff
@@ -0,0 +1,58 @@
...
+public class SearchResults extends DataTable<SearchResults.Row> {
+
+    public SearchResults(Recipe recipe) {
+        super(recipe, "Source files that had search results",
+                "Search results that were found during the recipe run.");
+    }
+
+    @Value
+    public static class Row {
+        //TODO: Question for Jonathan: should we not (similar to SourcesFileErrors) have a afterSourcePath and represent both?
```
jkschneider: Sure, I consider this a harmless add.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
```diff
@@ -0,0 +1,58 @@
...
+        // What if a recipe changes the path and adds markers? Do we need both?
+        // If a later recipe removed the sourcefile, we will not have the markers anymore. Not foreseeing that as an issue as typically search and edit/delete will not go together.
+        @Column(displayName = "Source path of search result",
+                description = "The source path of the file with the search result markers present.")
+        @Nullable
+        String sourcePath;
+
+        //TODO: Question for Jonathan: what should the truncation length be?
```
jkschneider: No truncation noted here in this column. If we decide to add truncation at all, we'll do so with an `ExecutionContext` message likely. For now, would assume it shouldn't be necessary. Even `SourcesFileErrors` has cell values with >1000 characters (the stacktrace) that can produce many rows.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
```diff
@@ -0,0 +1,58 @@
...
+        @Column(displayName = "Result",
+                description = "The trimmed printed tree of the LST element that the marker is attached to. Truncated after 1000 chars if longer.")
+        String result;
+
+        //TODO: Question for Jonathan: Should we not report both instance name and recipe name? Searching for the recipe id from the String is not always "easy" (see also the unit tests)
```
jkschneider: Personally I don't see the utility of the recipe ID in this situation. We can always add later in a backwards compatible way. If there are many recipes of the same ID with different options in a composite, the ID doesn't get me any closer to understanding which one produced the result anyway. The parent will always be the top-level recipe anyway.

rewrite-core/src/test/java/org/openrewrite/table/SearchResultsTest.java
```diff
@@ -0,0 +1,248 @@
...
+          text(
+            "hello",
+            spec -> spec.path("non-matched")
+          ),
+          text(
+            "hi",
+            "~~>hi",
+            spec -> spec.path("nested/matched")
+          ),
+          text(
```
jkschneider: This seems to be testing the functionality of `org.openrewrite.text.Find` not the data table entry, so would simplify this unit test by removing this case.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
Jenson3210: > The parent will always be the top-level recipe anyway.

What do you mean by this? Currently the table will contain multiple rows with every level in there (similar to the SourcesFileResults Datatable)
This means that if I have a composite recipe of:
```text
A
  B
    C <- this one puts the marker
```
the datatable will contain the "same" search result multiple times: 

```text
| parentRecipe | recipe | result     |
+--------------+--------+------------+
| null         |A       | "result"   |
| A            |B       | "result"   |
| B            |C       | "result"   |
```
What I understand from your comment is that you  only want (which would not leave the possibility of adding pivot tables on intermediate recipes which resulted in that marker): 
```text
| A            |C       | "result"   |
```

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
Jenson3210: > Sometimes this can be an entire file which can be arbitrarily large. Will probably want to truncate past some length.

@sambsnyd mentioned this as a comment on the doc, so that's why I asked it here.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
```diff
@@ -0,0 +1,58 @@
...
+                        "this is the root of a hierarchy or if the recipe is not hierarchical at all.")
+        String parentRecipe;
+
+        @Column(displayName = "Recipe that added the search marker",
+                description = "The specific recipe that added the Search marker.")
+        String recipe;
+
+    }
+    //TODO: Question for Jonathan: in `SourcesFileResults`, I see a `build` method which is not used in the OR codebase. Perhaps by our other system? Do we need to have this build method or not? `SourcesFileErrors` does not have it.
+}
```
Jenson3210, line 51:
```suggestion
    }
}
```

jkschneider: No, whether the data table is built by a regular constructor or a builder isn't super important to me.

rewrite-core/src/main/java/org/openrewrite/table/SearchResults.java
jkschneider: Correct, I would like the latter. `SourcesFileResults` is used to build the sankey diagram, but I don't think we have such needs here. I'm fine if we remove `parentRecipe` altogether in this initial PR.

PR #6371, opened 2025-12-03
add RemoveConfigurationFromGradleTask

dsgrieve:
## What's changed?

Adding a recipe and test for removing a configuration from a gradle task. 

## What's your motivation?

Trying to handle [classic uber-jar loader removal](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#classic-uber-jar-loader-support) from Spring Boot 4.0. 

### Checklist

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTaskTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTaskTest.java
@@ -1,0 +1,373 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.gradle;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.gradle.Assertions.buildGradle;
+import static org.openrewrite.gradle.Assertions.buildGradleKts;
+
+class RemoveConfigurationFromGradleTaskTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new RemoveConfigurationFromGradleTask(
+          "bootJar",
+          "loaderImplementation = org.springframework.boot.loader.tools.LoaderImplementation.CLASSIC"
+        ));
+    }
+
+    @DocumentExample
+    @Test
+    void removeLoaderImplementationFromBootJarTask() {
+        rewriteRun(
+          //language=gradle
+          buildGradle(
+            """
+              plugins {
+                  id 'org.springframework.boot' version '4.0.0'
+              }
+              
+              bootJar {
+                  loaderImplementation = 'org.springframework.boot.loader.tools.LoaderImplementation.CLASSIC'
+              }
+              """,
+            """
+              plugins {
+                  id 'org.springframework.boot' version '4.0.0'
+              }
+              
+              bootJar {
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void removeLoaderImplementationFromBootJarTaskUsingGlob() {
+        rewriteRun(
+          spec -> spec.recipe(new RemoveConfigurationFromGradleTask("bootJar", "loaderImplementation*")),
+          //language=gradle
+          buildGradle(
+            """
+              plugins {
+                  id 'org.springframework.boot' version '4.0.0'
+              }
+              
+              bootJar {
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTask.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsBuildGradle<>(), new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
+                // Only process Groovy and Kotlin Gradle build files
+                return sourceFile instanceof G.CompilationUnit || sourceFile instanceof K.CompilationUnit;
+            }
```
sambsnyd, line 69: Gradle build files can only be Groovy or Kotlin so `IsBuildGradle<>()`  should be sufficient, the `isAcceptable()` is redundant

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTask.java
```diff
@@ -0,0 +1,173 @@
...
+                // Normalize both the statement and the configuration to remove for comparison
+                String statementStr = normalizeStatement(statement);
+                String configStr = normalizeStatement(configuration);
+                return StringUtils.matchesGlob(statementStr, configStr);
+            }
+
+            private String normalizeStatement(Object obj) {
+                String str;
+                if (obj instanceof Statement) {
+                    str = ((Statement) obj).print(getCursor());
```
timtebeek, line 153: Generally we try not to use print much in recipes. Comparing to one specific notation seems fragile, as folks might use an import instead for instance. Can we match on a type, or something more closely matching an LST element at least?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTask.java
```diff
@@ -0,0 +1,173 @@
...
+                    List<Statement> newStatements = new ArrayList<>();
+                    boolean modified = false;
+
+                    for (Statement statement : body.getStatements()) {
+                        if (!matchesConfigurationToRemove(statement)) {
+                            newStatements.add(statement);
+                        } else {
+                            modified = true;
+                        }
+                    }
```
sambsnyd, line 89: Filtering statements like this should generally be accomplished using `ListUtils.map()`. Return null from `map()` to remove an element. Disregard the nullability warning about returning `null`. (It's because if we remove the `UnaryOperator` overloads the remaining `Function` overloads, while source-compatible, would be binary-incompatible with older jars compiled to target `UnaryOperator`. So they stay for now despite causing nullability warnings)

`ListUtils` functions only allocate new arrays when the elements have definitely changed. 
This preservation of identity when the array is unchanged helps to avoid excess cycles due to referential equality being used to determine if an LST has changed. A newly allocated `ArrayList` will obviously never be referentially equal to the one it replaced even if their elements are the same.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveConfigurationFromGradleTask.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Option(displayName = "Task name",
+            description = "The name of the Gradle task to modify.",
+            example = "bootJar")
+    String taskName;
+
+    @Option(displayName = "Configuration",
+            description = "The configuration statement to remove from the task. Supports glob expressions.",
+            example = "loaderImplementation = org.springframework.boot.loader.tools.LoaderImplementation.CLASSIC")
+    String configuration;
```
sambsnyd, line 49: The configuration of task or plugin can take many forms. These can be nested arbitrarily deeply within lamdas. For example, configuring one or more `Test` tasks to use Junit 5 can be accomplished in any of these substantially equivalent ways:
```groovy
test {
    useJUnitPlatform()
}
test {
    useJUnitPlatform { }
}
tasks.withType(Test) {
    useJUnitPlatform()
}
tasks.withType(Test).configureEach {
    useJUnitPlatform()
}
tasks.named('test', Test) {
    useJUnitPlatform()
}
```
Configuration DSLs are often multiple levels deep:
```groovy
publishing {
  publications {
    myPublication(MavenPublication) {
      from components.java
      artifact sourceJar
      pom {
        name = "Demo"
        description = "A demonstration of Maven POM customization"
        url = "http://www.example.com/project"
        licenses {
          license {
            name = "The Apache License, Version 2.0"
            url = "http://www.apache.org/licenses/LICENSE-2.0.txt"
          }
        }
        developers {
          developer {
            id = "johnd"
            name = "John Doe"
            email = "john.doe@example.com"
          }
        }
        scm {
          connection = "scm:svn:http://subversion.example.com/svn/project/trunk/"
          developerConnection = "scm:svn:https://subversion.example.com/svn/project/trunk/"
          url = "http://subversion.example.com/svn/project/trunk/"
        }
      }
    }
  }
}
```

Terminology note: "taskName" isn't precise. Often it is DSLs (domain specific lanugages) which receive user configuration. Directly configuring tasks not via plugin-provided configuration DSLs is considered less clean, but is of course very common. 

For this to be a generally useful utility I would expect it to accept something like an `XPath` but geared towards method chains/lambdas rather than XML. See `https://github.com/openrewrite/rewrite/commit/5224e77f7822ca25a8112f4983534b0781b3c19a`.
I anticipate that if we have more than a few use cases for this the current interface it will not be able to adequately express them, requiring effectively a new recipe. 

In its current form this isn't a good general-purpose "remove configuration" recipe.
Short of that, something like this could work as a more specific "Remove spring boot jar configuration" in rewrite-spring.

timtebeek:
Based on the feedback above let's close this effort.

PR #6372, opened 2025-12-03
Fix `NoSuchMethodError` during Java 11 parsing when the LST generation is running on an older Java 11 version which doesn't have `JCTree.JCVariableDecl.declaredUsingVar()`

akiraly:
Instead of calling `com.sun.tools.javac.tree.JCTree.JCVariableDecl.declaredUsingVar()` (which was [added](https://github.com/openjdk/jdk11u/commit/1db4ed1a62e1c93c4f32d51685723b3ec99aae03#diff-1fa76229dfd36a4b7f96a49c14157c3eac397346221b159f82c9963df8903fb5R967) in 2022, around [11.0.15](https://bugs.openjdk.org/browse/JDK-8261205)) figure out manually whether the lambda parameter was declared with the `var` keyword or not

## What's changed?
`org.openrewrite.java.isolated.ReloadableJava11ParserVisitor.visitVariables()` was changed to handle lambda parameters without the need to call `node.declaredUsingVar()`.
Also new tests were added to cover the issue.

## What's your motivation?
This issue currently causes thousands of java file parse errors in our organization and so are excluded from any Moderne recipe run. My goal is to make them available.

Is there is a better way to detect that the next token is the `var` keyword? The current approach
1. Skips any whitespace
2. Checks that the next token starts with `var`
3. Checks the character after `var` is whitespace

Alternative is to update all impacted projects to the latest Java 11 version which is definitely something we want to do. But fixing the parser eliminates the root problem.

### Checklist

#### Legal disclaimer:


akiraly:
@timtebeek Thanks for moving the test to the right place and merging the PR!

PR #6380, opened 2025-12-04
Add recipe for cleaning up unnecessary Gradle dependency substitution rules

bmuschko:
## What's changed?

Adds a new recipe RemoveRedundantSecurityResolutionRules that removes `resolutionStrategy.eachDependency` rules from Gradle build files when they are no longer needed because a platform/BOM already manages the dependency to an equal or newer version.

## What's your motivation?

When security vulnerabilities are discovered (CVEs, GHSAs), teams often add resolution rules to pin dependencies to fixed versions. 

Over time, these rules become redundant when:

  1. The platform/BOM (e.g., Spring Boot) is upgraded to a version that manages the dependency to a newer version
  2. The pinned version is now older than what the BOM provides

This recipe automates the cleanup by comparing the pinned version against the BOM-managed version and removing rules where the BOM already provides adequate protection.

I kept the recipe small in scope on purpose (security-specific resolution rules). If we'd make it too configurable, the logic could become multiple factors more complex.

End users can only manually identify those resolution rules and verify against the platform/BOM to identify if it is still needed.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
@@ -43,9 +43,7 @@ import org.openrewrite.semver.LatestIntegration;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 
@@ -105,37 +103,35 @@ public class RemoveRedundantSecurityResolutionRules extends Recipe {
         JavaIsoVisitor<ExecutionContext> removeRulesVisitor = new JavaIsoVisitor<ExecutionContext>() {
             @Nullable
             GradleProject gradleProject;
-            @Nullable
-            UUID currentSourceFileId;
             boolean insideBuildscript;
 
-            private void maybeInitialize(JavaSourceFile sourceFile) {
-                // Reset state when visiting a new source file
-                if (currentSourceFileId != null && currentSourceFileId.equals(sourceFile.getId())) {
-                    return;
+            private void maybeInitialize() {
+                // Initialize once per source file (getVisitor() is called once per source file)
+                if (gradleProject == null) {
+                    JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
+                    if (sourceFile != null) {
+                        gradleProject = sourceFile.getMarkers().findFirst(GradleProject.class).orElse(null);
+                    }
                 }
-                currentSourceFileId = sourceFile.getId();
-                gradleProject = sourceFile.getMarkers().findFirst(GradleProject.class).orElse(null);
             }
 
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
-                JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
-                if (sourceFile != null) {
-                    maybeInitialize(sourceFile);
-                }
+                maybeInitialize();
 
-                // Track when we enter/exit buildscript block
-                boolean wasInsideBuildscript = insideBuildscript;
-                if (METHOD_BUILDSCRIPT.equals(method.getSimpleName())) {
+                // Track when we enter buildscript block
+                boolean enteredBuildscript = METHOD_BUILDSCRIPT.equals(method.getSimpleName());
+                if (enteredBuildscript) {
                     insideBuildscript = true;
                 }
 
+                boolean currentlyInBuildscript = insideBuildscript;
                 J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
 
-                // Restore state after visiting children
-                boolean currentlyInBuildscript = insideBuildscript;
-                insideBuildscript = wasInsideBuildscript;
+                // Reset after exiting buildscript block
+                if (enteredBuildscript) {
+                    insideBuildscript = false;
+                }
 
                 if (!isEachDependency(m, getCursor())) {
                     return m;


rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
```diff
@@ -0,0 +1,669 @@
...
+                buildscriptPlatforms.clear();
+                projectPlatforms.clear();
+
+                Optional<GradleProject> maybeGp = sourceFile.getMarkers().findFirst(GradleProject.class);
+                if (maybeGp.isPresent()) {
+                    gradleProject = maybeGp.get();
+
+                    // Find all platforms and download their POMs to get managed versions
+                    // Use a custom visitor that properly tracks buildscript context
+                    MavenPomDownloader mpd = new MavenPomDownloader(ctx);
```
sambsnyd: Any dependency could be bringing in newer versions of the pinned dependencies, not just platforms. I suspect the only comprehensive way to determine if a rule can be removed is to emulate dependency resolution without the rule. 

Fortunately for us, I have made this [easy to simulate](https://github.com/openrewrite/rewrite/commit/92a9c9559941ab16ec3d287a4329a6f720f5218d#diff-b4404f951774c0fb08faec1218167a7c77d371225cfdd54daa895b0f49df4897).
Now you might look at that example and say "That's dealing with constraints, and these are dependency resolution rules, which are not constraints" and you would be right. But in an attempt to make our model more useful I have modeled the [consequences of dependency resolution rules](https://github.com/openrewrite/rewrite/blob/1ebb4a20a27b6f218da9e271d036e0d55fd2050c/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java#L277) as constraints.

So you should be able to ask the question "what happens if I remove this rule" by calling `GradleProject.removeConstraints()` and inspecting the resulting configurations.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
```diff
@@ -0,0 +1,669 @@
...
+
+    @Option(displayName = "Security pattern",
+            description = "A regular expression pattern to identify security-related resolution rules by matching " +
+                          "against the `because` clause. Rules matching this pattern will be considered for removal. " +
+                          "Default pattern matches CVE identifiers (e.g., `CVE-2024-1234`) and GitHub Security " +
+                          "Advisory identifiers (e.g., `GHSA-xxxx-xxxx-xxxx`).",
+            example = "(CVE-\\d|GHSA-[a-z0-9])",
+            required = false)
+    @Nullable
+    String securityPattern;
```
sambsnyd, line 77: Note that multiple identifiers may be included in a single "because"

bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
bmuschko: I worked in the use of `GradleProject.removeConstraints()` but keep the platform handling as fallback. Let me know if I am still missing something.

sambsnyd: You can similarly use `GradleProject.removeDependency()` to simulate the effect of removing a `platform()` dependency.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
```diff
@@ -0,0 +1,683 @@
...
+            boolean insideBuildscript;
+
+            private void maybeInitialize(JavaSourceFile sourceFile) {
+                // Reset state when visiting a new source file
+                if (currentSourceFileId != null && currentSourceFileId.equals(sourceFile.getId())) {
+                    return;
+                }
+                currentSourceFileId = sourceFile.getId();
+                gradleProject = sourceFile.getMarkers().findFirst(GradleProject.class).orElse(null);
+            }
```
sambsnyd, line 116: The recipe scheduler calls `getVisitor()` once per source file. So you don't need to worry about the previous source file still being cached. You shouldn't need to track or check the ID. 

I would move this conditional from `visitMethodInvocation()` into `maybeInitialize`:
```
                JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
                if (sourceFile != null) {
                    maybeInitialize(sourceFile);
                }
```
Since we don't actually need to traverse up the cursor stack from every single method invocation, we only need to do it once when these fields are still `null`

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
```diff
@@ -0,0 +1,683 @@
...
+                boolean wasInsideBuildscript = insideBuildscript;
+                if (METHOD_BUILDSCRIPT.equals(method.getSimpleName())) {
+                    insideBuildscript = true;
+                }
+
+                J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+
+                // Restore state after visiting children
+                boolean currentlyInBuildscript = insideBuildscript;
+                insideBuildscript = wasInsideBuildscript;
```
sambsnyd: If you make `insideBuildscript` a field on the visitor class this can be as simple as
```java
if (METHOD_BUILDSCRIPT.equals(method.getSimpleName())) {
    insideBuildscript = true;
}
J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
insideBulidscript = false;
```

sambsnyd (approved):

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantSecurityResolutionRules.java
bmuschko: done

PR #6383, opened 2025-12-05
Add recipe to add bulk read method to InputStream implementations

pstreef:
Java's default `InputStream.read(byte[], int, int)` calls the single-byte `read()` in a loop, causing up to 350x slower performance for bulk reads.

This recipe detects InputStream subclasses that:
- Override `read()` but not `read(byte[], int, int)`
- Delegate to another `InputStream`

For simple delegation patterns, it adds the missing bulk read method. For complex bodies (side effects, transformations), it adds a search marker for manual review.

Handles both anonymous and named classes, ternary and if-statement null check styles, and skips FilterInputStream subclasses.

## What's changed?

Adding a recipe to detect and update (simple cases of) InputStream implementations that are missing the bulk read method.

## What's your motivation?
I found 3 such instances in OpenRewrite itself which caused huge performance gains when fixed.

Is this recipe useful in its current form, should it maybe only detect? or is only fixing simple cases good too?

@timtebeek 

Not creating this recipe

## Any additional context

- https://github.com/openrewrite/rewrite/pull/6381
- https://github.com/openrewrite/rewrite/pull/6377

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/cleanup/AddInputStreamBulkReadMethod.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/cleanup/AddInputStreamBulkReadMethod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2025 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.cleanup;
 
+import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.Cursor;
 import org.openrewrite.ExecutionContext;
@@ -27,7 +28,9 @@ import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.SearchResult;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/cleanup/AddInputStreamBulkReadMethod.java
```diff
@@ -0,0 +1,499 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.cleanup;
```
timtebeek, line 16: This package only contains a few visitors at this moment, as we've moved the majority of the recipes that were formerly here into rewrite-static-analysis.

For this recipe I'm doubting on where to place it, but I feel perhaps https://github.com/openrewrite/rewrite-migrate-java/blob/b566417e5ef76dfa6b4026cb2c08b9a5b335cbe7/src/main/java/org/openrewrite/java/migrate/io is more fitting than having it here.

If we keep it here, it would also be included with any other language module as those tend to use `rewrite-java`, which feels a bit off.

pstreef:
Moved to openrewrite/rewrite-migrate-java#961 per @timtebeek's review feedback.

pstreef:
Closing in favor of openrewrite/rewrite-migrate-java#961

Issue #6379, opened 2025-12-04
Return type of the method declared is not changed


cmoulliard:
## What version of OpenRewrite are you using?

I am using

- rewrite-core module: 8.67
- rewrite-java module: 8.67

## Issue description

Here is the scenario tested which is not working. The recipe should allow to replace the return type of the method declared within the JavaIsoVisitor using the `visitMethodDeclaration()` to check if there a match with the pattern of the method to search and next to set the new type

When we run the test, we can see that the type of the method changed (see system output - AFTER) but the test  is complaining (= AssertionFailedError) as it is waiting about the new type but got the original type. Why ?

```
package dev.snowdrop.openrewrite.quarkus.spring;

import org.junit.jupiter.api.DisplayName;
import org.junit.jupiter.api.Test;
import org.openrewrite.ExecutionContext;
import org.openrewrite.config.CompositeRecipe;
import org.openrewrite.java.JavaIsoVisitor;
import org.openrewrite.java.MethodMatcher;
import org.openrewrite.java.tree.J;
import org.openrewrite.java.tree.JavaType;
import org.openrewrite.test.RewriteTest;

import java.util.List;

import static org.openrewrite.java.Assertions.java;
import static org.openrewrite.test.RewriteTest.toRecipe;

public class ReplaceMethodReturnTypeTest implements RewriteTest {

    @Test
    @DisplayName("Replace the return type of a method.")
    void replaceReturnTypeOfMethodDeclared() {
        String methodPattern = "TaskController addMessage(..)";
        String newReturnType = "Object";

        rewriteRun(
            spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
                final MethodMatcher methodMatcher = new MethodMatcher(methodPattern, false);

                @Override
                public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
                    J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
                    JavaType.Method type = m.getMethodType();

                    if (methodMatcher.matches(m.getMethodType())) {
                        System.out.println("========== BEFORE ==========");
                        System.out.printf("Method name: %s \n",m.getSimpleName());
                        System.out.printf("Method modifiers: %s \n",m.getModifiers());
                        System.out.printf("Return Type: %s \n",m.getType());
                        System.out.printf("Declaring type: %s \n",m.getMethodType().getDeclaringType());

[...]

PR #6390, opened 2025-12-08
Add new recipe able to change for a method definition the return type

cmoulliard:
## What's changed?

- See issue: #6379
- Add a new recipe able to change for a method definition the return type

### Checklist


cmoulliard:
So based on the slack discussing you are suggesting to create a new Recipe able to add the missing imports able to support typically this use case where we change the return type of a method (declared/defined) ? 

@timtebeek

timtebeek:
My suggestion is not to make this a generic recipe with options provided via Yaml, but to make it a dedicated recipe for the use case that you're after. That way you'll be better able to provide the necessary classpath dependencies to resolve any new return types, able to plug those in without (likely) having to account for parameterized types, and able to additionally handle any usages that likely also need to be updated as opposed to just the declaration.

With that let's close this PR then as something we don't see working in the general case.

PR #6393, opened 2025-12-08
Kotlin missing type information

Laurens-W:
## What's changed?

Added a test for missing Kotlin Gradle DSL

## What's your motivation?

Running into missing type info with Kotlin projects

### Checklist


rewrite-java/src/main/java/org/openrewrite/java/search/FindMissingTypes.java
```diff
@@ -181,12 +181,18 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
         @Override
         public J.MemberReference visitMemberReference(J.MemberReference memberRef, ExecutionContext ctx) {
             J.MemberReference mr = super.visitMemberReference(memberRef, ctx);
-            JavaType.Method type = mr.getMethodType();
-            if (type != null) {
```
knutwannheden: Should we instead change the `else` on line 197 to an `else if (mr.getVariableType() != null)` and then add this new logic as a final `else`? We can then there add a comment explaining that this is used for Kotlin class literals. I guess we should then also call `isWellFormedType()` for that type.

Laurens-W: Applied here https://github.com/openrewrite/rewrite/pull/6402/files#diff-9789da1530726d3f8a2b631e9912c6320af2d46ef462af6c3b1e064a8516cb8c

timtebeek:
The K2 upgrade just merged in PR #6766 — it brings significantly improved type attribution (e.g., generic method invocation types now resolve to concrete types like `kotlin.String` instead of `Generic{E}`). Worth rebasing on main to check if the type information gaps this test captures are now resolved by K2.

PR #6399, opened 2025-12-08
Add annotation matcher support for Kotlin's class reference

Laurens-W:
## What's changed?

Introduced support for Kotlin's `Class::class` class reference notation

## What's your motivation?

`AddMockitoExtensionIfAnnotationsUsed` is adding a duplicate annotation with Java style class reference in Kotlin tests

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AnnotationMatcher.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AnnotationMatcher.java
@@ -169,6 +169,25 @@ public class AnnotationMatcher {
                     return false;
                 }
 
+                // Handle Kotlin's ::class.java syntax
+                if ("java".equals(fa.getSimpleName()) && fa.getTarget() instanceof J.MemberReference) {
+                    J.MemberReference mr = (J.MemberReference) fa.getTarget();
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/AnnotationMatcher.java
```diff
@@ -177,6 +177,25 @@ private boolean argumentValueMatches(String matchOnArgumentName, Expression arg,
                     }
                 }
             }
+            // Handle Kotlin's ::class syntax which is parsed as MemberReference
+            if (arg instanceof J.MemberReference) {
+                J.MemberReference mr = (J.MemberReference) arg;
+                if ("class".equals(mr.getReference().getSimpleName()) && (matchText.endsWith(".class") || matchText.endsWith("::class"))) {
+                    JavaType argType = mr.getContaining().getType();
+                    if (argType instanceof JavaType.FullyQualified) {
+                        // Handle both Java's ".class" and Kotlin's "::class" syntax
```
knutwannheden, line 206: Also note that when a Java annotation is used in Kotlin, it will look like this:

```kt
@SomeJavaAnnotation(String::class.java)
class Example
```

Laurens-W: Added support and a test case for that as well :)

PR #6406, opened 2025-12-09
Avoid duplicate dependencies following one or more `ChangeDependency` recipes

BoykoAlex:
`ChangeDependency` can lead to duplicate dependencies if a a number of dependencies is changed to the same new dependency.

1. Maven recipe has a bug for version ranges, i.e. `2.0.X`. Doesn't match versions with wildcards, only expect exact version
2. Gradle recipe doesn't check if dependency exists.

As a consequence the `ChangeDependency` recipes in `rewrite-java-dependencies` would have this feature working. See 
- https://github.com/openrewrite/rewrite-java-dependencies/pull/165

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -183,35 +183,8 @@ public class ChangeDependency extends Recipe {
             public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
                 if (tree instanceof JavaSourceFile) {
                     JavaSourceFile sourceFile = (JavaSourceFile) tree;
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Thanks a lot @BoykoAlex ! I recall wrestling with this for indeed some Jakarta dependencies, as well as some Jackson and Spring libs where reshuffling of classes has seen some artifacts effectively merged. Good not to have any duplicates in the output, and thanks for tackling both Maven, Gradle and rewrite-java-dependencies at once.

timtebeek:
I'll ask colleagues to review as well given that this is changing (improving!) two core recipes; One application that comes to mind is
https://github.com/openrewrite/rewrite-jackson/blob/c8d2e4ff42f2f59de6624f7fba5bb5df9d99024b/src/main/resources/META-INF/rewrite/jackson-2-3.yml#L172-L208

Issue #6408, opened 2025-12-10
Assertions.validateTypes can falsely report a missing type for a Kotlin method invocation


greg-dennis:
## What version of OpenRewrite are you using?

I am using commit d9be625d432c78398b13070e60cdf6ff8d7a8628, an 8.69.0-SNAPSHOT

## How are you running OpenRewrite?

Developing a new recipe in rewrite-testing-frameworks.

## What is the smallest, simplest way to reproduce the problem?

- Patch the test in [this PR](https://github.com/openrewrite/rewrite-testing-frameworks/pull/870) and run it. It should pass, and the equivalent test does pass if added to the rewrite repo.

## What did you see instead?

The test fails because the type of the method invocation is determined to be null.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes.

I tracked it down to the `FirFunctionCall` having a calleeReference that is found to be a `FirErrorNamedReference` in [PsiElementAssociations here](https://github.com/openrewrite/rewrite/blob/e118585d5a7081be8315c669127177e8b100002f/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt#L360). And a null type is apparently acknowledged as expected for some method invocations by [KotlinTreeParseVisitor here](https://github.com/openrewrite/rewrite/blob/e118585d5a7081be8315c669127177e8b100002f/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java#L2018).

timtebeek:
Figured this might be interesting to @Laurens-W given his ongoing work on
- https://github.com/openrewrite/rewrite/pull/6393

timtebeek:
Closing as resolved by the K2 upgrade in PR #6766. The root cause was `FirErrorNamedReference` in PSI element associations causing null types for Kotlin method invocations. K2 rewrote the named reference handling (using `FirResolvedArgumentList.mapping` instead of index-based argument lookup) and added an `isNamedArgument()` allowlist in `Assertions.java` that correctly exempts the left-hand side of named arguments from type validation.

PR #6409, opened 2025-12-10
Allow Kotlin method invocation type to be considered valid if null

greg-dennis:
## What's changed?

This allows the type of a Kotlin `MethodInvocation` to be  considered valid if null, as already anticipated by the [by the KotlinTreeParseVisitor](https://github.com/openrewrite/rewrite/blob/e118585d5a7081be8315c669127177e8b100002f/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java#L2018).

## What's your motivation?

- Fix/workaround for https://github.com/openrewrite/rewrite/issues/6408

@timtebeek has context from Slack
@Laurens-W

I tried to find a way to force the parser to resolve the type, but the issue appears to be deep in the bowels of the Jetbrains FIR.

### Checklist

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/Assertions.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/Assertions.java
@@ -377,7 +377,10 @@ public final class Assertions {
             // Avoid over-reporting the same problem by checking the invocation only when its elements are well-formed
             if (mi == method) {
                 JavaType.Method type = mi.getMethodType();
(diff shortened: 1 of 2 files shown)


timtebeek:
As discussed on https://github.com/openrewrite/rewrite-testing-frameworks/pull/871 the preferred way of ignoring some type validation issues is through `.typeValidationOptions(...`. Let's close this effort as it's aimed at working around rather than fixing the underlying issue.

PR #6420, opened 2025-12-11
Fix error when parsing xml-s (and plain texts) encoded with UTF-8 BOM

akiraly:
Fix error when parsing xml-s (and plain texts) encoded with UTF-8 BOM.
`Parser.requirePrintEqualsInput()` restores the BOM while printing the parsed `SourceFile` (in `Tree.print()`) but it doesn't restore the BOM on the reread input (in `SourceFile.printEqualsInput()`). This causes `requirePrintEqualsInput()` (and `XmlParser.parseInputs()` which calls the equal check) to always return a `ParsingError` for XML-s encoded with UTF-8 BOM because the 2 String-s won't be equal (one has the BOM and the other doesn't).
`PlainTextParser` suffers from the same issue, but there the parsing doesn't fail because that parser doesn't invoke the equal check during parsing.

## What's changed?
The `SourceFile.printEqualsInput()` was changed to make sure BOM is restored on the reread Input as well not just on the printed SourceFile.
The existing BOM restore logic (in `Tree.print()`) was refactored into a new method to make it reusable from both sides.

## What's your motivation?
We see several hundred XML parsing errors (during LST building) because of this issue in Moderne. I would like to eliminate those errors.
While BOM-s are optional they are very frequently used by some tools/frameworks (mostly from the Microsoft world).

1. Should the plain text parsing also automatically invoke `requirePrintEqualsInput()` like the xml parsing? Or is there a reason why the 2 parser work differently in this aspect?
2. The `XmlParserTest` already included 2 BOM related tests but those are missing to detect this issue because those kind of tests seem to disable the `requirePrintEqualsInput()` part of the parsing for the duration of the tests.
Additionally one of the tests (`XmlParserTest.utf8BOMCharacters()`) uses special characters to represent the BOM part in he input (`ï»¿`) and those don't seem to be detected as BOM characters at all.
I left those tests alone and instead added new tests that also explicitly verify the BOM related parsing as well (and the equality check). But maybe those 2 existing tests should be revisited as well in the future.

### Checklist

#### Legal disclaimer:

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Tree.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Tree.java
@@ -83,27 +83,22 @@ public interface Tree {
         this.<P>printer(cursor).visit(this, out, cursor);
 
         // Restore BOM if the source file originally had one
-        restoreBOM(out.out);
-
-        return out.getOut();
-    }
-
-    default void restoreBOM(StringBuilder out) {
         if (this instanceof SourceFile &&
             !(this instanceof Quark) &&
             !(this instanceof Binary)) {
             SourceFile sourceFile = (SourceFile) this;
             try {
-                if (sourceFile.isCharsetBomMarked()) {
-                    if (out.length() > 0 &&
-                        out.charAt(0) != '\uFEFF') {
-                        out.insert(0, '\uFEFF');
-                    }
+                if (sourceFile.isCharsetBomMarked() &&
+                        out.out.length() > 0 &&
+                        out.out.charAt(0) != '\uFEFF') {
+                    out.out.insert(0, '\uFEFF');
                 }
             } catch (UnsupportedOperationException e) {
                 // Defensive fallback for any other SourceFile implementations that don't support charset operations
             }
         }
+
+        return out.getOut();
     }
 
     default String print(TreeVisitor<?, PrintOutputCapture<Integer>> printer) {
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -16,7 +16,9 @@
 package org.openrewrite;
 
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.binary.Binary;
 import org.openrewrite.internal.StringUtils;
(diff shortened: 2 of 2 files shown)


timtebeek:
Tagging @pstreef for his recent work on
- https://github.com/openrewrite/rewrite/pull/6266

rewrite-core/src/main/java/org/openrewrite/Tree.java
```diff
@@ -83,22 +83,27 @@ default <P> String print(Cursor cursor, PrintOutputCapture<P> out) {
         this.<P>printer(cursor).visit(this, out, cursor);
 
         // Restore BOM if the source file originally had one
+        restoreBOM(out.out);
+
+        return out.getOut();
+    }
+
+    default void restoreBOM(StringBuilder out) {
```
timtebeek: I understand wanting to reuse this functionality. Let me just say that we're fairly cautious adding new API surface area, and a change like this would need some careful deployment on our end before use. Figured point that out for my colleagues mostly.

akiraly: I can perfectly understand that. I also realized this is very core change that effects way more than just the 2 parsers in question.
What do you suggest? Maybe a static utility method with limited visibility?

timtebeek: Given how I don't expect this logic to change much anymore, nor needing it elsewhere, I think even just duplicating the lines we'd need in `printEqualsInput` might be tolerable over reuse.

timtebeek: I've duplicated that logic just now. Hope you agree!

timtebeek (approved):

timtebeek:
Thanks a lot for the help here! I'll tag Peter for a final review before we merge, but I can see this land in the next release.

rewrite-core/src/main/java/org/openrewrite/Tree.java
akiraly: Looks good to me!

pstreef (approved):
LGTM

Thanks you for this contribution!

PR #6421, opened 2025-12-11
Add description column to global SearchResults Datatable

Jenson3210:
Added the description column to the global datatable. 

eg.
A recipe that runs DependencyInsight tags gradle dependencies in the project.:
```yaml
DependencyInsight:
  group: org.springframework.boot
  artifact: *
```
results in 
```groovy
dependencies {
  /*~~(org.springframework.boot:spring-boot-dependencies:2.6.15,org.springframework.boot:spring-boot:2.6.6)~~>*/implementation 'io.pivotal.cfenv:java-cfenv-boot:2.5.0'
}
```

Which would only result in a datatable:
```
     build.gradle     |     build.gradle     |     implementation 'io.pivotal.cfenv:java-cfenv-boot:2.5.0'     |
```
Which then would still require people to go to the dependencies of that dependencies to see which ones are in. 
If we include the description, we get these insights in a single go: 
```
     build.gradle     |     build.gradle     |     implementation 'io.pivotal.cfenv:java-cfenv-boot:2.5.0'     |     org.springframework.boot:spring-boot-dependencies:2.6.15,org.springframework.boot:spring-boot:2.6.6     |
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -26,6 +26,7 @@ import org.openrewrite.config.DeclarativeRecipe;
 import org.openrewrite.internal.ExceptionUtils;
 import org.openrewrite.internal.FindRecipeRunException;
 import org.openrewrite.internal.RecipeRunException;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.marker.*;
 import org.openrewrite.quark.Quark;
 import org.openrewrite.table.RecipeRunStats;
@@ -383,7 +384,7 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                         ctx.add(new SearchResults.Row(
                                 (before == null) ? "" : before.getSourcePath().toString(),
                                 after.getSourcePath().toString(),
-                                ((Tree) cursor.getValue()).print(getCursor(), new PrintOutputCapture<>(0, PrintOutputCapture.MarkerPrinter.SANITIZED)),
+                                StringUtils.trimIndent(((Tree) cursor.getValue()).print(getCursor(), new PrintOutputCapture<>(0, PrintOutputCapture.MarkerPrinter.SANITIZED))),
                                 ((SearchResult) marker).getDescription(),
                                 recipeName));
                     }


timtebeek (approved):

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -352,30 +352,40 @@ private static boolean isScanningRequired(Recipe recipe) {
...
                     Cursor cursor = getCursor();
                     if (!(cursor.getValue() instanceof Tree)) {
                         cursor = cursor.getParentTreeCursor();
                     }
                     if (cursor.getValue() instanceof Tree) {
-                        ctx.add(((Tree) cursor.getValue()).print(getCursor(), new PrintOutputCapture<>(0, PrintOutputCapture.MarkerPrinter.SANITIZED)));
+                        ctx.add(new SearchResults.Row(
+                                (before == null) ? "" : before.getSourcePath().toString(),
+                                after.getSourcePath().toString(),
+                                ((Tree) cursor.getValue()).print(getCursor(), new PrintOutputCapture<>(0, PrintOutputCapture.MarkerPrinter.SANITIZED)),
```
timtebeek: Would we want to immediately trim this value as well? I feel that might be expected by folks processing the Excel file.

Jenson3210: Probably we do want to trim this indeed. I was just not sure how to do this here. 

- `printTrimmed`: none of the existing methods seems to accept the `Sanitized` PrintOutputCapture.
- `withPrefix`: prefix only available on J,Yaml,Properties...
- `trim()`: will only trim leading spaces (eg. comments etc will still have unexpected results)
- `StringUtils.trimIndent`: this could work, but not sure if this will work with comments on elements:
    ```groovy
    dependencies {
        implementation('somedependency') // foobar
        implementation('somemarkeddependency')
    }
    ``` 
    would print as `// foobar\n   implementation('somemarkeddependency')` as the prefix first comment has no whitelines
- introduce `PrintOutputCapture.MarkerPrinter.SANTIZED_COMMENTS` or something else that prints no markers AND no comments. 

So I did not put something in this PR as this trimming is probably a different discussion, but open to adding it here!

jkschneider: @Jenson3210 printTrimmed is just `StringUtils.trimIndent(print(..))` so I think you can just wrap the `print` call you have with `StringUtils.trimIndent`.

Jenson3210: @jkschneider 
I tested that while writing my comment. Unfortunately, that does not work with the example given above. 
StringUtils checks every line and searches for a common "indentation". For the prefix/comment here, the whitespaces is a single space. 
As both lines in that prefix share at most 1 space the result becomes: 
```
//foobar
   implementation('somemarkeddependency')
```
instead of (the spaces are hard to see here)
```
 //foobar
    implementation('somemarkeddependency')
```

Jenson3210:
```java
    @Test
    void testing() {
        assertThat(StringUtils.trimIndent(" //foobar\n    implementation('somemarkeddependency')")).isEqualTo("//foobar\n   implementation('somemarkeddependency')");
    }
```

jkschneider: I don't think you need to fix this edge case right now. Most of the time we won't have a comment at the beginning. It isn't specific to comments particularly. We could later update `trimIndent` to have some heuristic to disregard the first line if every other line has a consistent minimum indent greater than the first line.

PR #6422, opened 2025-12-11
Fixed groovy parser issue to handle slashy string delimters as escape characters.

### LINKED ISSUE #6395: Groovy Parser failure with dollar signs used both in regexp and as variable references

```java
    @Test
    void dollarSomethingDollar() {
        rewriteRun(
          groovy(
            """
            def text = "irrelevant"
            def glassfishVersion = "1"
            println text =~ /^.*ClassPath Element.*glassfish-embedded-all-${glassfishVersion}.jar.*$/
            """
          )
        );
    }
```
when added to one of the Groovy tests fails with:
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "file.groovy":
```

```diff
 def glassfishVersion = "1"
-println text =~ /^.*ClassPath Element.*glassfish-embedded-all-${glassfishVersion}.jar.*$/
+println text =~ /^.*ClassPath Element.*glassfish-embedded-all-${glassfishVersion}.jar.*/$/
```

## OSS repro

- https://github.com/eclipse-ee4j/glassfish-maven-embedded-plugin/blob/9dbeec67deab49d6f76416423317c575acb78fde/src/it/start-with-dependency/postbuild.groovy#L20

sudouser777:
## What's changed?
- Updated groovy parser to to handle salshy string delimiters as non-escaping characters.
- closes #6395 

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -1748,7 +1748,7 @@ public class GroovyParserVisitor {
                     String value = sourceSubstring(cursor, delimiter.close);
                     // There could be a closer GString before the end of the closing delimiter, so shorten the string if needs be
                     int indexNextSign = source.indexOf("$", cursor);
-                    while (isEscaped(indexNextSign)) {
+                    while (isEscaped(indexNextSign, delimiter)) {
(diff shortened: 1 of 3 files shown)


greg-at-moderne (approved):
Lovely. Thanks one more time!

PR #6423, opened 2025-12-11
Find Maven dependencies using exactly LATEST or RELEASE as version

timtebeek:
Before this change we were unable to find dependencies using `<version>LATEST</version>` or `<version>RELEASE</version>`, which is somewhat problematic when trying to find and remove such cases.
- https://docs.openrewrite.org/reference/dependency-version-selectors#hyphenated-range


steve-aom-elliott:
@timtebeek I'm wondering if we should actually be changing `LatestRelease.buildLatestRelease(..)`([here](https://github.com/openrewrite/rewrite/blob/8f3c3a9dd504f198be0c67f412e6deb22874e3b1/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java#L220-L224)) and `LatestIntegration.build(..)`([here](https://github.com/openrewrite/rewrite/blob/8f3c3a9dd504f198be0c67f412e6deb22874e3b1/rewrite-core/src/main/java/org/openrewrite/semver/LatestIntegration.java#L48-L52)) to allow for `RELEASE` and `LATEST` respectively given I had previously added comparisons for them in `LatestRelease.compare(..)`([here](https://github.com/openrewrite/rewrite/blob/8f3c3a9dd504f198be0c67f412e6deb22874e3b1/rewrite-core/src/main/java/org/openrewrite/semver/LatestRelease.java#L90-L96))

Also don't quite understand the `=` part

rewrite-maven/src/test/java/org/openrewrite/maven/search/FindDependencyTest.java
```diff
@@ -334,4 +336,112 @@ void versionInProperties() {
...
+                <dependencies>
+                  <!--~~>--><dependency>
+                      <groupId>javax.annotation</groupId>
+                      <artifactId>javax.annotation-api</artifactId>
+                      <version>%s</version>
+                  </dependency>
+                  <!--~~>--><dependency>
+                      <groupId>javax.servlet</groupId>
+                      <artifactId>javax.servlet-api</artifactId>
+                      <version>3.1.0</version>
```
timtebeek, line 439: This currently matches with `LatestIntegration` and `LatestRelease`, but likely should not as `3.1.0` is not the latest version; that's `4.0.1` used above(!). That seems like a preexisting problem with those two version modes.

timtebeek:
Thanks; adjusted just now so we can judge the effects; I've left a column where I don't like the effect of those version selectors.

Also the `=` is a shorthand to say you want an `ExactVersion`, as seen in the parser here:
https://github.com/openrewrite/rewrite/blob/8f3c3a9dd504f198be0c67f412e6deb22874e3b1/rewrite-core/src/main/java/org/openrewrite/semver/ExactVersion.java#L44-L46

With the change you suggested one would now have to use `=LATEST` or `=RELEASE` to very specifically find `<version>LATEST</version>` or `<version>RELEASE</version>`. That's not terrible, but might require an update to the docs:
https://docs.openrewrite.org/reference/dependency-version-selectors#exact-version

steve-aom-elliott:
@timtebeek I added
- https://github.com/openrewrite/rewrite/pull/6427

Which is me investigating what can be done to resolve that `3.1.0`  being considered something for marking when looking for `RELEASE`

timtebeek:
With https://github.com/openrewrite/rewrite/pull/6430 already merged I'll close my effort here as there's no immediate need to support this; wasn't sure on your WIP branch, but figured that could be closed as well. Feel free to reopen if you think there's value there.

PR #6428, opened 2025-12-12
Update version in the Gradle model

BoykoAlex:
Update new dependency version in the model


timtebeek:
Thanks a lot for helping explore these failures to detect duplicates. Taking into account [your comment here](https://github.com/openrewrite/rewrite-jackson/pull/62#issuecomment-3644385206) I'll also tag Shannon and Sam for review.
> This helps: https://github.com/openrewrite/rewrite/pull/6428
However, I'm somewhat uncomfortable about computing resolvedVersion for every config... wonder if this needed at all. Wonder what would be the argument against keeping one cached resolvedVersion for all configs...

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -799,4 +800,53 @@ void makeChangesInKotlinDependencyManagementImports() {
...
+              }
+              dependencies {
+                  implementation("com.fasterxml.jackson.core:jackson-databind:2.19.0")
+                  implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.19.0")
+              }
+              """,
+            spec -> spec.after(buildGradle ->
+              assertThat(buildGradle)
+                .containsOnlyOnce("tools.jackson.core:jackson-databind:3")
+                .doesNotContain("datatype")
```
timtebeek, line 847: For context: without the above changes we saw `jackson-databind:3` duplicated twice, as both map to the new artifact & version.

shanman190:
So most of the recipes compute the new version during the actual edit itself, then reuses the resolved version to modify all configurations.

Otherwise, the fix seems reasonable.

BoykoAlex:
@timtebeek I think this should be okay to go in as is more or less. The AST changes are in visitMethodInvocation etc hence we cannot update the GradleProject model right away. Changes can be accumulated and then applied at the sourcefile visit then we'd need to have calls to record the updates etc. I think it is fine the way it is implemented now (other ChangeXXX recipe behave the same) the versions are cached since pom metadata is cached

timtebeek (approved):
Thanks for the help here @BoykoAlex ! Seems reasonable from my side, as the version resolves you've added are conditional, and wrapped in an update to a cached transient field to limit how often these get called. I've asked Shannon whether that's acceptable to him as well.

Update: discussed in Slack; decided to move forward. Thanks both!

PR #6429, opened 2025-12-12
Wrapping and braces style

Jenson3210:
## What's changed?
Introduced all properties in the WrappingAndBraces style and partially support their wrapping. 
Added additional support for a bunch of properties. 
Refactored the existing support in the WrappingAndBracesVisitor to all do their logic through the same "visitSpace" concept rather than dedicated visitors for every type of J element. 

## What's your motivation?
Not all of the style properties are supported yet, but this will allow us to clean up the `evaluate()` pattern sooner for all properties as the parent loading will have all properties for all lst elements when we've add support for these wrappings. The unsupported ones are also ones that are less likely to be used in initial usages. This is already a big improvement compared to the before-state. 

There is still a small need for the TabsAndIndentsVisitor currently, but I feel we can get rid of this one in a next iteration as only very little amount of indents are now not immediately correct after WrappingAndBraces -> planning to add the little missing support to that one and then deprecate TabsAndIndents to save a traversal (actually, make up for the one that got introduced by the MergeSpacesVisitor)

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/service/SourcePositionService.java
@@ -304,7 +304,7 @@ public class SourcePositionService {
                         }
                     } else {
                         startCol.incrementAndGet();
-                        if (indenting.get() && c == ' ') {
+                        if (indenting.get() && (c == ' ' || c == '\t')) {
                             rowIndent.incrementAndGet();
                         } else {
                             indenting.set(false);
@@ -329,7 +329,7 @@ public class SourcePositionService {
                     indenting.set(true);
                 } else {
                     startCol.incrementAndGet();
-                    if (indenting.get() && c == ' ') {
+                    if (indenting.get() && (c == ' ' || c == '\t')) {
                         rowIndent.incrementAndGet();
                     } else {
                         indenting.set(false);
--- a/rewrite-java/src/test/java/org/openrewrite/java/format/SourcePositionServiceTest.java
+++ b/rewrite-java/src/test/java/org/openrewrite/java/format/SourcePositionServiceTest.java
@@ -296,6 +296,47 @@ class SourcePositionServiceTest implements RewriteTest {
         );
     }
 
+    @Test
+    void correctlyCalculatesIndentWithTabs() {
+        rewriteRun(
+          spec -> spec.recipe(RewriteTest.toRecipe(() -> new JavaIsoVisitor<>() {
+
+              @Nullable
+              SourcePositionService service;
+
+              @Override
+              public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+                  service = cu.service(SourcePositionService.class);
+                  return super.visitCompilationUnit(cu, ctx);
+              }
+
+              @Override
+              public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
+                  if (multiVariable.getVariables().stream().anyMatch(v -> "x".equals(v.getSimpleName()))) {
+                      // Tab-indented variable should report indent of 2 (two tabs)
+                      Span span = service.positionOf(getCursor());
+                      assertThat(span.getColSpan().getIndent()).isEqualTo(2);
+                  }
+                  if (multiVariable.getVariables().stream().anyMatch(v -> "y".equals(v.getSimpleName()))) {
+                      // Single tab indent should report indent of 1
+                      Span span = service.positionOf(getCursor());
+                      assertThat(span.getColSpan().getIndent()).isEqualTo(1);
+                  }
+                  return super.visitVariableDeclarations(multiVariable, ctx);
+              }
+          })),
+          java(
+            // Using tabs for indentation (represented as \t)
+            "class Test {\n" +
+            "\tvoid method() {\n" +
+            "\t\tint x = 1;\n" +
+            "\t}\n" +
+            "\tint y = 2;\n" +
+            "}\n"
+          )
+        );
+    }
+
     private static  <T extends J> T minimize(T tree) {
         tree = (T) new MinimumViableSpacingVisitor<>(null).visit(tree, -1);
         return (T) new SpacesVisitor<>(IntelliJ.spaces(), null, null, IntelliJ.wrappingAndBraces().withKeepWhenFormatting(IntelliJ.wrappingAndBraces().getKeepWhenFormatting().withLineBreaks(false)), null).visit(tree, -1);


knutwannheden:
This is a really impressive piece of work! I particularly like the architectural approach of routing formatting logic through `visitSpace()`, `visitContainer()`, and `visitRightPadded()` rather than having dedicated visitor methods for every AST element type. The use of `Space.Location` and `JContainer.Location` to dispatch formatting decisions keeps the code centralized and maintainable.

The addition of comprehensive `WrappingAndBracesStyle` properties (force braces, alignment options, open/close newline controls, etc.) fills important gaps in the formatter's capabilities. The test coverage is thorough and demonstrates the flexibility of the new approach.

## Performance Concern

I do have a concern about the performance characteristics of the current implementation. The `SourcePositionService.positionOf()` method prints the source file from the root up to the target element each time it's called:

```java
Cursor printCursor = cursor;
if (!(cursor.getValue() instanceof JavaSourceFile)) {
    printCursor = cursor.dropParentUntil(c -> c instanceof JavaSourceFile);
}
javaPrinter.visit(printCursor.getValue(), printLine, printCursor.getParent());
```

In `WrappingAndBracesVisitor`, this is called for:
- `alignWhenMultiline` calculations (method parameters, arguments, implements lists, try-with-resources, etc.)
- `WrapIfTooLong` / `ChopIfTooLong` checks against `hardWrapAt`
- Text block and method chain alignment

With default IntelliJ settings where several `alignWhenMultiline` options are enabled, a typical source file with many methods could trigger hundreds of these calls. Each call traverses all AST nodes from the file start up to the target element, resulting in O(n^2) complexity in the worst case, where n is the file size.

## Potential Solutions

A few ideas that might help:

### 1. Print from a local anchor point
Instead of printing from the file root, print from a nearby anchor (e.g., the enclosing statement or method). Column position within a statement is what matters for most wrapping decisions, and statements are typically short enough that printing them is cheap.

### 2. Track column position incrementally
Similar to how `JavaPrinter` tracks its output position as it traverses, a formatting visitor could maintain column position as state:
- Reset column on newlines
- Add to column count when visiting content
- Use tracked position directly for alignment/wrapping decisions

This would give O(n) traversal instead of O(n^2).

### 3. Hybrid approach
Keep the centralized `visitSpace()`/`visitContainer()` architecture (avoiding per-element overrides), but combine it with:
- Incremental indentation tracking (like `TabsAndIndentsVisitor`'s `lastIndent` message)
- Local printing only when column position is needed for alignment

This would preserve the PR's architectural elegance while avoiding the performance pitfall.

---

Overall this is heading in a great direction - I just want to flag the performance aspect before it becomes an issue on larger codebases.

Jenson3210:
Thanks @knutwannheden for this! 
That is indeed the reason for this one being a draft.

I am looking at the TabsAndIndentsVisitor right now to see how I could reuse patterns from over there to not have to print entire file all the time. 

The other approach of anchor-printing also popped up in my mind. Printing as of the first element that starts on a new line would already prevent the entire file of being printed all the time. 
That would have to export another kind of `Span` object as it won't have accurate line numbers (which we do not need for formatting)

I also considered just adding the new line only and leaving indentation to the TabsAndIndentsVisitor, but that will give issues later as the WrapIfLong needs to know the indent of other container elements before it can decide if it is long or now.

rewrite-groovy/src/main/java/org/openrewrite/groovy/format/AutoFormatVisitor.java
```diff
@@ -27,13 +27,20 @@ public class AutoFormatVisitor<P> extends GroovyIsoVisitor<P> {
     @Nullable
     private final Tree stopAfter;
 
+    private final boolean removeCustomLineBreaks;
+
```
sambsnyd: This should be a field on a style like all the other style configuration. Unless there's truly something special about this setting? I don't think there is... but enlighten me if I'm missing something

Jenson3210: Thanks. Totally agree. I went for uniformity with Java visitor, but now also moved that one to the WrappingAndBracesStyle 

Screenshot 2025-12-19 at 11 08 24

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
```diff
@@ -956,33 +992,1829 @@ void annotationWrappingWithNulls() {
...
+              }
+              """,
+            """
+              import java.io.*;
+
+              class Test {
+                  void testNameWhichIsReallyWayTooLongForThisLimitOf120SoWraps() throws
+                          FileNotFoundException,
+                          IOException,
+                          RuntimeException {
```
timtebeek: Slightly surprised me that we place all three each on a new line when we only configure `wrap.getThrowsList().withWrap(WrapIfTooLong)`, but fine if that's intentional.

More general I'm only seeing two tests use `WrapIfTooLong`; are we confident enough in the usage elsewhere?

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
```diff
@@ -956,33 +992,1829 @@ void annotationWrappingWithNulls() {
...
+          java(
+            """
+              class Test {
+                  int[] numbers = {1, 2, 3, 4, 5};
+              }
+              """,
+            """
+              class Test {
+                  int[] numbers = {1,
+                          2,
```
timtebeek, line 2245: Do we indeed expect the first element to be on the original line still? Mostly wondering, not telling to change just yet.

timtebeek: Ah I'd missed the `withNewLineAfterOpeningCurly(true)` when reviewing top down.

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
```diff
@@ -956,33 +992,1829 @@ void annotationWrappingWithNulls() {
...
+                          5
+                  };
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void arrayInitializerDoNotWrap() {
```
timtebeek, line 2388: Elsewhere we seem to use
```suggestion
    void arrayInitializerDoNotWrapAlreadyCorrect() {
```
Or would we want to add a before text block here that's not already as we want it?

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
```diff
@@ -956,33 +992,1829 @@ void annotationWrappingWithNulls() {
...
+          ),
+          java(
+            """
+              @SuppressWarnings(value = "all", justification = "test")
+              class Test {
+              }
+              """
+          )
+        );
+    }
```
timtebeek, line 2654: Should we add a before text block here that's not conforming and then gets adjusted?

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
Jenson3210: you are right here. This should not be in the PR anymore. 
Due to the fact that a position of a previous throws expressions gets wrapped, the location of the next one might change. I planned to remove WrapIfTooLong for now again (just pushed this)

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
Jenson3210: added a before block

rewrite-java-test/src/test/java/org/openrewrite/java/format/WrappingAndBracesTest.java
Jenson3210: Added a before block

timtebeek:
There's one thread still unresolved it seems, but other than that what I've seen looks good to me; thanks for the massive effort here!
- https://github.com/openrewrite/rewrite/pull/6429/files#r2632544939

rewrite-java/src/main/java/org/openrewrite/java/style/TabsAndIndentsStyle.java
```diff
@@ -33,28 +33,8 @@ public class TabsAndIndentsStyle implements JavaStyle {
...
-    @With
-    public static class MethodDeclarationParameters {
-        Boolean alignWhenMultiple;
-    }
-
-    @Value
-    @With
-    public static class RecordComponents {
-        Boolean alignWhenMultiple;
-    }
```
timtebeek, line 49: Since these classes are in a style package, I think we'd need a coordinated release right? Given that `if ("tree".equals(part) || "marker".equals(part) || "style".equals(part))` we spotted earlier today

timtebeek (approved):
Approved from my side already. Might need a coordinated merge & release given the packages changed.

Jenson3210:
We should be good i guess. The evaluate method handles the package changes in cli and it also handled it for previous changes. But better safe than sorry 😊

rewrite-java/src/main/java/org/openrewrite/java/style/IntelliJ.java
```diff
@@ -100,16 +98,40 @@ public static SpacesStyle spaces() {
...
-                new WrappingAndBracesStyle.ChainedMethodCalls(DoNotWrap, emptyList(), false),
-                new WrappingAndBracesStyle.MethodDeclarationParameters(DoNotWrap, false, false, false),
-                new WrappingAndBracesStyle.MethodCallArguments(DoNotWrap, false, false, false),
-                new WrappingAndBracesStyle.Annotations(WrapAlways),
-                new WrappingAndBracesStyle.Annotations(WrapAlways),
-                new WrappingAndBracesStyle.Annotations(WrapAlways),
-                new WrappingAndBracesStyle.Annotations(DoNotWrap),
-                new WrappingAndBracesStyle.Annotations(DoNotWrap),
-                new WrappingAndBracesStyle.Annotations(DoNotWrap)
+                new WrappingAndBracesStyle.KeepWhenFormatting(true, true, true, true, false, false, false, false),
```
knutwannheden: Please verify:
```suggestion
                new WrappingAndBracesStyle.KeepWhenFormatting(true, true, true, false, false, false, false, false),
```

rewrite-java/src/main/java/org/openrewrite/java/style/IntelliJ.java
```diff
@@ -100,16 +98,40 @@ public static SpacesStyle spaces() {
...
+                new WrappingAndBracesStyle.ExtendsImplementsPermitsList(LineWrapSetting.DoNotWrap, false),
+                new WrappingAndBracesStyle.ExtendsImplementsPermitsKeyword(LineWrapSetting.DoNotWrap),
+                new WrappingAndBracesStyle.ThrowsList(LineWrapSetting.DoNotWrap, false, false),
+                new WrappingAndBracesStyle.ThrowsKeyword(LineWrapSetting.DoNotWrap),
+                new WrappingAndBracesStyle.MethodDeclarationParameters(LineWrapSetting.DoNotWrap, true, false, false),
+                new WrappingAndBracesStyle.MethodCallArguments(LineWrapSetting.DoNotWrap, false, false, false, false),
+                new WrappingAndBracesStyle.MethodParentheses(false, false),
+                new WrappingAndBracesStyle.ChainedMethodCalls(LineWrapSetting.DoNotWrap, false, false, emptyList(), false, false),
+                new WrappingAndBracesStyle.IfStatement(WrappingAndBracesStyle.ForceBraces.DoNotForce, false, true),
+                new WrappingAndBracesStyle.ForStatement(LineWrapSetting.DoNotWrap, false, false, false, WrappingAndBracesStyle.ForceBraces.DoNotForce),
```
knutwannheden: Please verify:

```suggestion
                new WrappingAndBracesStyle.ForStatement(LineWrapSetting.DoNotWrap, true, false, false, WrappingAndBracesStyle.ForceBraces.DoNotForce),
```

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormat.java
```diff
@@ -46,21 +45,8 @@ public class AutoFormat extends Recipe {
     @Nullable
     String style;
 
-    @Option(displayName = "Remove custom line breaks",
-            description = "Do you want to remove custom line breaks? (default false)",
-            required = false)
-    @Nullable
-    Boolean removeCustomLineBreaks;
```
sambsnyd, line 53: I am glad to see this go 👍

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
```diff
@@ -122,7 +119,7 @@ public J visit(@Nullable Tree tree, P p) {
         return (J) tree;
     }
 
-    @ToBeRemoved(after = "30-11-2025", reason = "Replace me with org.openrewrite.style.StyleHelper.addStyleMarker now available in parent runtime")
+    @ToBeRemoved(after = "30-01-2025", reason = "Replace me with org.openrewrite.style.StyleHelper.addStyleMarker now available in parent runtime")
     private static <T extends SourceFile> T addStyleMarker(T t, List<NamedStyles> styles) {
         if (!styles.isEmpty()) {
```
sambsnyd, line 124: Why did this date go back in time? Probably safe to remove this now

rewrite-java/src/main/java/org/openrewrite/java/service/Span.java
```diff
@@ -29,4 +29,26 @@ public class Span {
...
+
+    /**
+     * Tracks the beginning and ending of a given LST element in source code
+     */
+    @Value
+    @Builder
+    public static class ColSpan {
+        int startColumn;
+        int endColumn;
+        int maxColumn;
```
sambsnyd, line 39: What is the difference between "end column" and "max column" ? I would have assumed these were the same... for that matter why do we need this as a separate concept from `Span` ?

Jenson3210: There are 2 question in this comment afaics. 

- **Difference between endColumn and maxColumn**: 
  Some elements will span multiple lines. The endColumn is the column on which the entire lst element printed is ending. The maxColumn is the width of the printed block. Whichever line is the longest will have its length in the maxColumn. 
  
- **Difference between Span and ColSpan**:
For Span (which has line information), we have to print the file as of the SourceFile's beginning. For Colspan, we only have to print the lst element which starts on a new line (as we do not care about the lines). 
During calculations in WrappingAndBraces / Indents ... we only need column information. This saves a lot of printing during the calculation and then also reduces the runtime of these visitors.

rewrite-java/src/main/java/org/openrewrite/java/format/AutoFormatVisitor.java
Jenson3210: I missed the year. I would not like to introduce the risk of NoSuchMethodErrors in this PR. 
Subsequent PR will follow that will remove this one. Updated the dates of these annotations so we can remove all of them at once.

timtebeek:
what's the plan here folks? I worry about the amount of conflicts we're quickly gathering here.

sambsnyd (approved):

Issue #6391, opened 2025-12-08
Maven `UpgradeParentVersion` incorrectly attempts & fails to download a parent pom.xml when using `<version>${revision}${sha1}${changelist}</version>` in multi module project


elmuerte:
When using CI friendly versions in a multi module maven project, where at least one of the variable is an XML nil value, the upgrade parent recipe fails.

## What version of OpenRewrite are you using?
I am using

- Maven plugin v6.25.0

## How are you running OpenRewrite?

`mvn org.openrewrite.maven:rewrite-maven-plugin:6.25.0:dryRun`

I am using the Maven plugin, and my project is a multi module project.
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>

	<groupId>example.group</groupId>
	<artifactId>project-parent</artifactId>
	<version>${revision}${sha1}${changelist}</version>
	<packaging>pom</packaging>

	<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>3.5.0</version>
	</parent>

	<properties>
		<revision>1.2.3</revision>
		<changelist>-SNAPSHOT</changelist>
		<sha1 /> <!-- nil property value -->
	</properties>

	<modules>
		<module>module1</module>
	</modules>

	<build>
		<plugins>
			<plugin>
				<groupId>org.openrewrite.maven</groupId>
				<artifactId>rewrite-maven-plugin</artifactId>
				<version>6.25.0</version>
				<configuration>
					<activeRecipes>
						<recipe>example.recipe</recipe>
					</activeRecipes>
				</configuration>
			</plugin>
		</plugins>
	</build>
</project>
```

rewrite.yml in the project basedir

```yml
---
type: specs.openrewrite.org/v1beta/recipe
name: example.recipe
recipeList:
  - org.openrewrite.maven.UpgradeParentVersion:
      groupId: org.springframework.boot
      artifactId: spring-boot-starter-parent
      newVersion: 3.5.x
```

submodule:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

PR #6431, opened 2025-12-12
Resolve existing maven properties with a null value to an empty string

elmuerte:
## What's changed?

Defining a maven property as `<key />` would produce a property with the value `null`. During property replacement where it is used in a string value this would result in an empty string.

For the placeholder helper the `getProperty` method should return an empty string for the key, otherwise it is regarded as a non-existing property as the replacement would not happen and the placeholder is kept.

## What's your motivation?

- Fixes issue #6391.

- See also this comment for an analysis of the issue: https://github.com/openrewrite/rewrite/issues/6391#issuecomment-3645757057

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -26,6 +26,7 @@ import org.openrewrite.config.DeclarativeRecipe;
 import org.openrewrite.internal.ExceptionUtils;
 import org.openrewrite.internal.FindRecipeRunException;
 import org.openrewrite.internal.RecipeRunException;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.marker.*;
 import org.openrewrite.quark.Quark;
 import org.openrewrite.table.RecipeRunStats;
@@ -257,9 +258,9 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                 effortSeconds,
                 cycle));
 
-        List<String> searchMarkers = collectSearchResults(before, after);
-        for (String searchResult : searchMarkers) {
-            searchResults.insertRow(ctx, new SearchResults.Row(beforePath, afterPath, searchResult, recipe.getInstanceName()));
+        List<SearchResults.Row> searchMarkers = collectSearchResults(before, after, recipe.getInstanceName());
+        for (SearchResults.Row searchResult : searchMarkers) {
+            searchResults.insertRow(ctx, searchResult);
         }
 
         if (hierarchical) {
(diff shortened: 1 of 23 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -325,6 +325,9 @@ public String getPackaging() {
             } else {
```
timtebeek: Turns out the `propVal` just above here is already a dead store in the original, so I'll refactor a little to clean that out as well.

rewrite-maven/src/test/java/org/openrewrite/maven/tree/ResolvedPomTest.java
```diff
@@ -366,6 +366,33 @@ void resolveExecutionsFromDifferentParents() {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/6391")
+    @Test
+    void propertyWithNullValue() {
```
timtebeek, line 371: This test seems to pass even with the early return above removed on my machine; did you try it before making your change?

timtebeek: Turns out we weren't guarding against such accidental misuse:
- https://github.com/openrewrite/rewrite/pull/6432

elmuerte: No sorry, I did not test if it failed without the fix. I was working off the failing example project. I will address it.

elmuerte: This is interesting... The test fails as the resolved poms don't have the versions replaced. This is because in `ResolvedPom.resolve` it does not compare if the version was changed. As that's the only change in the pom, it would return `this` instead. However, running the minimal example from the issue, it did work correctly.

I've added a new test case which also mimics this minimal example from the issue (with a parent+child pom). Which currently fails as it tried to download the parent pom (with the correct version string). I see it's also trying to do that when I run the recipe against the actual example, but it's not a failure in that case.

So funnily enough it managed to properly resolve the parent GAV based on data form the parent pom. But still wants to download it.

Guess I need to dig a bit further to find out why it still wants to download the pom. But at least I fixed the issue where it tried to use an invalid URL.

elmuerte: I found the issue. The `projectPomsByGav` in the `MavenPomDownloader` has versions not properly replaced. Updating the `mergeProperties` to change null values to an empty string solves the unneeded download issue.

The `MavenParserTest` needed two cases to be updated as the property substitution now happens where it previously did not.

I checked if there are more places where maven properties might not be properly interpolated, and there are quite a few places where this might be the case. For example in repository URLs, but maybe also in the maven property recipe. Fixing all those places is a bit out of scope for this PR. It also requires some investigation in where null values are interpolated to an empty string in maven and where not, and how this affects OpenRewrite actions. The [DefaultInterpolator](https://github.com/apache/maven/blob/master/impl/maven-impl/src/main/java/org/apache/maven/impl/model/DefaultInterpolator.java) is the place where most of the property interpolation is done.

PR #6432, opened 2025-12-12
Prevent unexpected but not technically invalid paths in tests

timtebeek:
- We saw this pop up in https://github.com/openrewrite/rewrite/pull/6431/commits/00a76efc170381d41edafc16c55d56b5130883c9 ; there `mavenProject` was accidentally passed a first argument of XML, which then is treated as a directory by that name. It also means there were no further files in that Maven project, leading to a silent failure to execute the test.

Linux allows just about any character to be used in file paths, so the `Paths.get(String)` we use in `Dir` just always returns a `Path` instance, even for weird cases like the above.

I figured lock this down a bit, not from a security point of view, but to guard against unexpectedly skipped tests as described. Figured we should allow directory separators, regular characters, dashes and spaces, but not much else that I could immediately think of.

### DIFF
--- a/rewrite-test/src/main/java/org/openrewrite/test/Dir.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/Dir.java
@@ -28,14 +28,14 @@ public class Dir implements Iterable<SourceSpec<?>>, SourceSpecs {
     private final Consumer<SourceSpec<SourceFile>> spec;
     private final SourceSpecs[] sourceSpecs;
 
-    public Dir(String dir, SourceSpecs[] sourceSpecs, Consumer<SourceSpec<SourceFile>> spec) {
+    public Dir(String dir, Consumer<SourceSpec<SourceFile>> spec, SourceSpecs... sourceSpecs) {
         // Prevent invalid paths such as `<project>` passed into `mavenProject(...)` where `pomXml` should be used
         if (!dir.matches("[- \\w/\\\\]+")) {
             throw new IllegalArgumentException("Invalid directory: " + dir);
         }
         this.dir = dir;
-        this.sourceSpecs = sourceSpecs;
         this.spec = spec;
+        this.sourceSpecs = sourceSpecs;
     }
 
     @Override


rewrite-test/src/main/java/org/openrewrite/test/Dir.java
```diff
@@ -24,12 +23,21 @@
...
 
-@RequiredArgsConstructor
 public class Dir implements Iterable<SourceSpec<?>>, SourceSpecs {
     private final String dir;
     private final Consumer<SourceSpec<SourceFile>> spec;
     private final SourceSpecs[] sourceSpecs;
 
+    public Dir(String dir, Consumer<SourceSpec<SourceFile>> spec, SourceSpecs... sourceSpecs) {
+        // Prevent invalid paths such as `<project>` passed into `mavenProject(...)` where `pomXml` should be used
+        if (!dir.matches("[- \\w/\\\\]+")) {
```
honnix, line 33: This doesn't allow empty string anymore. Is that a intended change? It seems we can no longer simulate a "root" project.

timtebeek: If you find you need that we could turn that + into a *. Welcome to push that up. On our end we just use "project" everywhere we used to have an empty string.

honnix: Yeah. We are having trouble with https://github.com/openrewrite/rewrite-migrate-java/blob/7592e247e3ba4fa915d42847b3b37338d39b8bff/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java#L126, which assumes the file at the root of a project, IIUC.

timtebeek: https://github.com/openrewrite/rewrite/commit/c2a48f519569ce57deec7102b91b9ab185ffdba9

honnix: Nice! I was writing a PR description :D. I will drop that then.

timtebeek: Hah thanks, yes was quick enough from mobile (and bed: out sick). Hope that helps. The next planned full release is two/three weeks away though, but I imagine a patch release will trickle through over Christmas.

honnix: Thank you so much for the quick fix. Have a good rest and I wish you great holidays.

timtebeek: Thanks! All the best there as well.

PR #6440, opened 2025-12-15
Followup to BOM handling changes in #6420: `isCharsetBomMarked()` call should be inside the try block as that is the only method that can throw `UnsupportedOperationException`

akiraly:
- Followup to BOM handling changes in #6420: `isCharsetBomMarked()` call should be inside the try block as that is the only method that can throw `UnsupportedOperationException`.
Also, some minor test improvements.

## What's changed?
`SourceFile.printEqualsInput()` was changed to follow more closely the BOM restore implementation in `Tree.print()`

## What's your motivation?
- Make sure the change done in #6420 doesn't introduce any regressions where `SourceFile.isCharsetBomMarked()` thows `UnsupportedOperationException`.

@timtebeek 

### Checklist

## Legal disclaimer:

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/SourceFile.java
@@ -48,14 +48,18 @@ public interface SourceFile extends Tree {
         }
 
         // Restore BOM if the source file originally had one
-        if (isCharsetBomMarked() && !(this instanceof Quark) && !(this instanceof Binary)) {
-            try {
-                if (!readFromInput.isEmpty() && readFromInput.charAt(0) != '\uFEFF') {
-                    readFromInput = '\uFEFF' + readFromInput;
-                }
-            } catch (UnsupportedOperationException e) {
-                // Defensive fallback for any other SourceFile implementations that don't support charset operations
+        // should be in sync with the BOM restore logic in Tree.print()
+        try {
+            if (!(this instanceof Quark) &&
+                !(this instanceof Binary) &&
+                isCharsetBomMarked() &&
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/SourceFile.java
```diff
@@ -48,14 +48,18 @@ default boolean printEqualsInput(Parser.Input input, ExecutionContext ctx) {
         }
 
         // Restore BOM if the source file originally had one
-        if (isCharsetBomMarked() && !(this instanceof Quark) && !(this instanceof Binary)) {
-            try {
-                if (!readFromInput.isEmpty() && readFromInput.charAt(0) != '\uFEFF') {
-                    readFromInput = '\uFEFF' + readFromInput;
-                }
-            } catch (UnsupportedOperationException e) {
```
akiraly, line 56: @timtebeek after #6420 was merged I realized that this try-catch here doesn't do anything and in `Tree.print()` the only method call that was covered by the try-catch there was the `isCharsetBomMarked()`. So I updated the logic here to make sure that call happens inside the try-catch; also to follow the original logic as close as possible.

rewrite-xml/src/test/java/org/openrewrite/xml/XmlParserTest.java
```diff
@@ -500,17 +500,6 @@ void utf8BOMCharacters() {
     }
 
     @Issue("https://github.com/openrewrite/rewrite/issues/1382")
-    @Test
```
akiraly, line 503: I have combined this test with the next one, hope that is OK.

PR #6441, opened 2025-12-15
Allow specifying environment variables to `JavaScriptRewriteRpc.Builder`

pdelagrave:
## What's changed?
`JavaScriptRewriteRpc.Builder` now allows custom environment variables to be set.

## What's your motivation?
When the Moderne CLI has a proxy and trust store config set, it has to be propagated to sub processes.

The new test `environmentVariableIsSetRemotely` is using the `ReplaceAssignment` test recipe to check if the environment variable was passed to the RPC process running the recipe, would there be a simpler way to test that?

### Checklist


rewrite-javascript/build.gradle.kts
```diff
@@ -195,6 +195,7 @@ testing {
                 implementation(project(":rewrite-test"))
                 implementation(project(":rewrite-json"))
                 implementation(project(":rewrite-java-tck"))
+                implementation(project(":rewrite-yaml"))
```
greg-at-moderne, line 198: Not sure why do we need it?

rewrite-javascript/rewrite/fixtures/replace-assignment.ts
```diff
@@ -0,0 +1,66 @@
...
+                    valueSource: envVarValue,
+                    type: Type.Primitive.String,
+                };
+                const leftPaddedLiteral: J.LeftPadded<J.Literal> = {
+                    kind: J.Kind.LeftPadded,
+                    markers: emptyMarkers,
+                    element: theLiteral,
+                    before: singleSpace,
+                }
+                draft.initializer = createDraft(leftPaddedLiteral);
```
greg-at-moderne: Not sure why would `createDraft` be needed in this line.

rewrite-javascript/build.gradle.kts
pdelagrave: The `integTest` isn't fully running yet but that's fixing one of its issue, JUnit can't load the test suite correctly without that dependency.

greg-at-moderne: OK. Must've been some other change which made it required. Thanks for fixing this.

PR #6444, opened 2025-12-16
Fix Java parser for record component annotations

### LINKED ISSUE #6401: Java Parser might fail with annotations and records

Current main = c9d48d4c333d41b0d2072c000a558d959dbc122e

```java
    @Test
    void annotationsAndRecords() {
        rewriteRun(
          java(
            """
            import java.lang.annotation.*;
            
            @Target(ElementType.RECORD_COMPONENT)
            @interface Select {
                String value();
            }
            
            public record File(String name, @Select("native") String value) {
            }
            """
          )
        );
    }
```
when added to `JavaParserTest` fails with:
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "File.java":
```

```diff
 
-public record File(String name, @Select("native") String value) {
+public record File(String name, Stringt("native") String value) {
```

## OSS repro

- https://github.com/eclipse-jnosql/jnosql/blob/7cf45e9c7c35a3173222f0bbffbb46099064fefa/jnosql-mapping/jnosql-mapping-reflection/src/test/java/org/eclipse/jnosql/mapping/reflection/entities/PCView.java

sudouser777:
## What's changed?
- Updated Java parsers to handle record component annotations.
- Made the changes for Java17, 21 and 25 parsers.
- closes #6401 

### Checklist

### DIFF
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/RecordTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/RecordTest.java
@@ -124,6 +124,7 @@ class RecordTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/6401")
     @Test
     void annotationsAndRecords() {
         rewriteRun(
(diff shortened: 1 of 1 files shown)


greg-at-moderne:
Makes sense to me. Thanks for fixing this!

sambsnyd (approved):
Looks reasonable. Thank you!

PR #6448, opened 2025-12-16
Add createNewKeys option for CopyValue

dsgrieve:
## What's changed?

Add an optional `createNewKeys` parameter for org.openrewrite.yaml.CopyValue
## What's your motivation?

By default, CopyValue creates new keys in the destination but this is sometimes undesirable. 
## Any additional context

`createNewKeys` parameter was copied from MergeYaml. CopyValue delegates to MergeYaml.

### Checklist

### DIFF
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/CopyValueTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/CopyValueTest.java
@@ -139,15 +139,6 @@ class CopyValueTest implements RewriteTest {
               source:
                 foo: bar
                 key: value
-              destination:
-                foo: baz
-              """,
-            """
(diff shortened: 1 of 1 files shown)


rewrite-yaml/src/test/java/org/openrewrite/yaml/CopyValueTest.java
```diff
@@ -127,4 +127,29 @@ void copyToOtherFile() {
...
+                key: value
+              destination:
+                foo: baz
+              """,
+            """
+              source:
+                foo: bar
+                key: value
+              destination:
+                foo: bar
```
timtebeek: If we're not expecting changes then this would only need a single text block right?
```suggestion
```

PR #6449, opened 2025-12-16
Add CopyValue for JSON

dsgrieve:
## What's changed?

Add a CopyValue recipe for JSON
## What's your motivation?

Need a CopyValue recipe similar to the YAML CopyValue recipe, but for JSON files.
## Any additional context

This started life as a copy of the YAML CopyValue recipe. 

### Checklist

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
@@ -123,11 +123,10 @@ public class CopyValue extends ScanningRecipe<CopyValue.Accumulator> {
                 return source;
             }
         };
-        if (sourceFilePath != null) {
-            visitor = Preconditions.check(new FindSourceFiles(sourceFilePath), visitor);
+        if (sourceFilePath == null) {
+            return visitor;
         }
-
-        return visitor;
+        return Preconditions.check(new FindSourceFiles(sourceFilePath), visitor);
     }
 
     @Override
@@ -135,7 +134,8 @@ public class CopyValue extends ScanningRecipe<CopyValue.Accumulator> {
         if (acc.snippet == null) {
             return TreeVisitor.noop();
         }
-        return Preconditions.check(new FindSourceFiles(destinationFilePath == null ? acc.path.toString() : destinationFilePath),
+        return Preconditions.check(
+                new FindSourceFiles(destinationFilePath == null ? acc.path.toString() : destinationFilePath),
                 new AddKeyValue(destinationKeyPath, destinationKey, acc.snippet, false).getVisitor());
     }
 }
--- a/rewrite-json/src/test/java/org/openrewrite/json/CopyValueTest.java
+++ b/rewrite-json/src/test/java/org/openrewrite/json/CopyValueTest.java
@@ -66,6 +66,15 @@ class CopyValueTest implements RewriteTest {
               }
               """,
             spec -> spec.path("a.json")
+          ),
+          // Unrelated unchanged file
+          json(
+            """
+              {
+                "source": "value"
+              }
+              """,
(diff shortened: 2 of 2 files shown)


rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
```diff
@@ -0,0 +1,152 @@
...
+                          "If `null` then the value will be copied from any JSON file it appears within.",
+            example = "src/main/resources/application.json",
+            required = false)
+    @Nullable
+    String oldFilePath;
+
+    @Option(displayName = "New key path",
+            description = "A [JsonPath](https://docs.openrewrite.org/reference/jsonpath-and-jsonpathmatcher-reference) expression defining where the value should be written.",
+            example = "$.dest.kind")
+    String newKey;
```
sambsnyd: According to `AddKeyValue` this parameter is just a key name, not a path as the description of this parameter suggests.

rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
```diff
@@ -0,0 +1,152 @@
...
+        // For example, "$.dest.kind" should return "$.dest"
+        // For "$.destination" should return "$"
+        // For bracket notation like "$['dest']['kind']" should return "$['dest']"
+        int index = keyPath.lastIndexOf(".");
+        if (index == -1) {
+            index = keyPath.lastIndexOf("[");
+        }
+        keyPath = index > 0 ? keyPath.substring(0, index) : keyPath;
+        return "$".equals(keyPath) ? "$." : keyPath;
+    }
```
sambsnyd, line 140: if the old key path is a dynamic selector of some kind this logic wont work.
Why don't you record the actual matching path from the scanner into the accumulator and use that rather than doing string manipulation

rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
dsgrieve: I tried to make the parameter names more self-documenting by adopting 'copy source to destination' semantics. For example, I replaced "newKey" with "destinationKey"

rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
dsgrieve: I was not able to find a way to say "give me the parent of this path" or to construct one. So now there is a "destinationKeyPath" and a separate "destinationKey". This is what the AddKeyValue recipe does as well.

rewrite-json/src/main/java/org/openrewrite/json/CopyValue.java
```diff
@@ -0,0 +1,142 @@
...
+                (destinationFilePath == null) ? "" : destinationFilePath + ":",
+                destinationKeyPath);
+    }
+
+    @Override
+    public String getDescription() {
+        return "Copies a JSON value from one key to another. " +
+               "The existing key/value pair remains unaffected by this change. " +
+               "Attempts to merge the copied value into the new key if it already exists. " +
+               "Attempts to create the new key if it does not exist.";
```
timtebeek: From what I see in AddKeyValue no attempt is made to merge the contents:
https://github.com/openrewrite/rewrite/blob/3d4c23352c3b7589e6dc822bb7dbc6235c7b65e9/rewrite-json/src/main/java/org/openrewrite/json/AddKeyValue.java#L83
```suggestion
               "Attempts to create the new key if it does not exist.";
```

timtebeek (approved):
Thanks! I've pushed up a small polishing commit as there were some slightly confusion "destination" fields used in tests that in reality where unrelated to what was being tested. With that I think we're ready to merge the first increment.

PR #6450, opened 2025-12-17
Properly construct nested type

BoykoAlex:


timtebeek:
Thanks both! Looking at `RecipeClassLoader` (internal) it looks like `JavaType` ought not to be parent loaded, so merging for now.

PR #6457, opened 2025-12-18
Make assertion future proof for newer parent pom versions

Jenson3210:
Cleaning up after the quick fix.

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
@@ -1451,13 +1451,16 @@ class ChangeParentPomTest implements RewriteTest {
                     </parent>
                   </project>
                   """,
-                spec -> spec.after(actual -> {
+                spec -> spec.after(actual ->
                     assertThat(actual)
+                      .as("Does not bring in Maven arguments.")
+                      .contains("<maven.repo.local>${user.home}/.m2/repository</maven.repo.local>")
+                      .contains("<owaspDb>${maven.repo.local}/blah</owaspDb>")
                       .as("Any version of Spring boot parent above `3.5.7`")
                       .doesNotContain("3.5.7")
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
```diff
@@ -1450,22 +1451,13 @@ void doesNotBringInMavenArguments() {
...
-                      <version>3.5.9</version>
-                    </parent>
-                  </project>
-                  """
+                spec -> spec.after(actual -> {
+                    assertThat(actual)
+                      .as("Any version of Spring boot parent above `3.5.7`")
+                      .doesNotContain("3.5.7")
+                      .containsPattern("<version>3\\.5\\.\\d+</version>");
+                    return actual;
```
timtebeek: Instead of `return actual` you can tag `.actual()` at the end of the chain and return that. Saves a few lines.

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
```diff
@@ -1450,22 +1451,13 @@ void doesNotBringInMavenArguments() {
...
-                      <groupId>org.springframework.boot</groupId>
-                      <artifactId>spring-boot-starter-parent</artifactId>
-                      <version>3.5.9</version>
-                    </parent>
-                  </project>
-                  """
+                spec -> spec.after(actual -> {
+                    assertThat(actual)
+                      .as("Any version of Spring boot parent above `3.5.7`")
+                      .doesNotContain("3.5.7")
```
timtebeek, line 1460: The intention of this test wasn't the dependency upgrade, but to not swap in the actual property value. We should add assertions that these are maintained: 
```
<maven.repo.local>${user.home}/.m2/repository</maven.repo.local>
<owaspDb>${maven.repo.local}/blah</owaspDb>
```

PR #6462, opened 2025-12-22
Allow for upgrading direct and transitive dependencies added by Gradle plugins

bmuschko:
## What's changed?

Enhance the logic of the UpgradeDependencyVersion and UpgradeTransitiveDependencyVersion recipes to allow for upgrading plugin-provided dependencies.

## What's your motivation?

Downstream recipes, e.g. DependencyVulnerabilityCheck, need to address dependencies provided by Gradle plugins.

No

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
@@ -138,12 +138,6 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
         Map<GroupArtifact, @Nullable Object> gaToNewVersion = new HashMap<>();
 
         Map<String, Map<GroupArtifact, Set<String>>> configurationPerGAPerModule = new HashMap<>();
-
-        /**
-         * All dependencies found in AST across all build scripts (including freestanding scripts).
-         * Used to detect if a resolved dependency was declared somewhere.
-         */
-        Set<GroupArtifact> allDeclaredDependencies = new HashSet<>();
     }
 
     @Override
@@ -196,9 +190,6 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
                 // Record the dependency and resolve its version if needed
                 GroupArtifact ga = new GroupArtifact(groupId, artifactId);
                 String configName = gradleDependency.getConfigurationName();
-
-                // Track all declared dependencies across all build scripts
-                acc.allDeclaredDependencies.add(ga);
 
                 if (gradleProject != null) {
                     acc.getConfigurationPerGAPerModule()
@@ -416,11 +407,10 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
 
                     GroupArtifact ga = new GroupArtifact(resolved.getGroupId(), resolved.getArtifactId());
 
-                    // Skip if declared in any build script or in this configuration's requested list
-                    if (acc.allDeclaredDependencies.contains(ga) ||
-                            configuration.getRequested().stream().anyMatch(req ->
-                                    ga.getGroupId().equals(req.getGroupId()) &&
-                                    ga.getArtifactId().equals(req.getArtifactId()))) {
+                    // Skip if declared in this configuration's requested list
+                    if (configuration.getRequested().stream().anyMatch(req ->
+                            ga.getGroupId().equals(req.getGroupId()) &&
+                            ga.getArtifactId().equals(req.getArtifactId()))) {
                         continue;
                     }
 


rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -368,11 +382,132 @@ public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
                 JavaSourceFile sourceFile = (JavaSourceFile) tree;
                 gradleProject = sourceFile.getMarkers().findFirst(GradleProject.class)
                         .orElse(null);
+
+                // Apply plugin-provided direct dependencies
+                if (gradleProject != null) {
+                    sourceFile = applyPluginProvidedDependencies(sourceFile, ctx);
+                }
```
sambsnyd: This null check is repeated inside `applyPluginProvidedDependencies()`. Recommend keeping that one and dropping this. 
```suggestion
                sourceFile = applyPluginProvidedDependencies(sourceFile, ctx);
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
```diff
@@ -418,11 +443,20 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
...
                         }
                     }
+                    // Only check all ancestors when there are no user-declared deps (plugin-provided deps case)
+                    if (declaredConfigurationsWithDeps.isEmpty()) {
+                        for (GradleDependencyConfiguration extended : config.allExtendsFrom()) {
+                            if (extended.getName().equals(constraintConfigName)) {
+                                return extended;
+                            }
+                        }
+                    }
```
sambsnyd: I'm pretty sure that `allExtendsFrom()` is going to return the direct parents first. So you can get the same behavior dropping the unnecessary loop over `getExtendsFrom()`.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
```diff
@@ -595,6 +628,84 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
         }
     }
 
+    @Value
+    @EqualsAndHashCode(callSuper = false)
+    private static class AddDependenciesBlockVisitor extends JavaVisitor<ExecutionContext> {
```
shanman190: `gradle.internal.AddDependencyVisitor` will add the dependencies block when it's missing. So you don't need to do this as well.

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -2148,4 +2148,86 @@ void upgradeAnnotationProcessors() {
...
+              """,
+            //language=groovy
+            """
+              plugins {
+                  id 'org.jetbrains.kotlin.jvm' version '1.7.21'
+              }
+              repositories { mavenCentral() }
+
+              dependencies {
+                  api "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.0"
```
shanman190, line 2176: optional: I can't remember for sure, but something about this specific example makes me think that it doesn't work for Kotlin. Either the plugin gets agitated about the version mismatch or the Gradle platform dependency overrides it.

bmuschko: Picking a potentially incompatible Kotlin library version is happening purely for the purpose of demonstrating the functionality of the plugin. Thanks for the pointer though.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
bmuschko: Thanks! Done.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
bmuschko: done

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -134,6 +138,12 @@ public static class DependencyVersionState {
         Map<GroupArtifact, @Nullable Object> gaToNewVersion = new HashMap<>();
 
         Map<String, Map<GroupArtifact, Set<String>>> configurationPerGAPerModule = new HashMap<>();
+
+        /**
+         * All dependencies found in AST across all build scripts (including freestanding scripts).
+         * Used to detect if a resolved dependency was declared somewhere.
+         */
+        Set<GroupArtifact> allDeclaredDependencies = new HashSet<>();
```
shanman190: This feels a little weird. The added test cases only cover the immediate build script, so it leads me to believe that we don't really need to know all of the declared dependencies across the entire Gradle build.

bmuschko: done

PR #6464, opened 2025-12-23
Resolve GAVs, classifiers and types of transitive dependencies via the containing POM

steve-aom-elliott:
The reason for this is that the overall resolved POM only contains the amalgamation of properties directly contained within our POM and those in the _parent_ POM. It was not taking into account properties of our dependencies for resolving their transitive dependencies. 

## What's changed?
- When resolving values for GAV & classifier / type on dependencies, process them through the POM that has that dependency as a direct dependency rather than through the overall resolved POM. This means that properties in our POM aren't used for resolving transitive dependency `X` and that the properties of the POM containing dependency `X` _directly_ are used instead.

## What's your motivation?
- There's a particular situation with the usage of `com.azure:azure-core-http-netty` wherein it defines a property that is used for a classifier of one of its dependencies, `io.netty:netty-tcnative-boringssl-static`, but because of the way the dependency resolution works in OpenRewrite, it was only using the resolved / cascaded properties of the current POM to try to resolve the classifier placeholder `${boring-ssl-classifier}`, rather than making use of the property definition in `com.azure:azure-core-http-netty`. This lead to a failure to resolve the dependency due to the placeholder persisting in the URL.

- @timtebeek
- @sambsnyd 

- Still working through ideas of how to better approach this

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java
@@ -62,18 +62,7 @@ public class RemoveDevelocityConfiguration extends Recipe {
                             J.Lambda lambda = (J.Lambda) m.getArguments().get(0);
                             J.Block body = (J.Block) lambda.getBody();
 
-                            List<Statement> filteredStatements = ListUtils.filter(body.getStatements(),
-                                    stmt -> {
-                                        if (stmt instanceof J.Return && ((J.Return) stmt).getExpression() instanceof J.MethodInvocation) {
-                                            // Unpack J.Return wrapping last statement in closure
-                                            stmt = (J.MethodInvocation) ((J.Return) stmt).getExpression();
-                                        }
-                                        if (!(stmt instanceof J.MethodInvocation)) {
-                                            return true;
-                                        }
-                                        // Remove only 'remote' method invocations
-                                        return !"remote".equals(((J.MethodInvocation) stmt).getSimpleName());
-                                    });
+                            List<Statement> filteredStatements = ListUtils.filter(body.getStatements(), this::retainStatement);
                             if (filteredStatements.isEmpty()) {
                                 return null;
                             }
@@ -81,6 +70,29 @@ public class RemoveDevelocityConfiguration extends Recipe {
                         }
 
                         return m;
+                    }
+
+                    private boolean retainStatement(Statement stmt) {
+                        if (stmt instanceof J.Return && ((J.Return) stmt).getExpression() instanceof J.MethodInvocation) {
+                            // Unpack J.Return wrapping last statement in closure
+                            stmt = (J.MethodInvocation) ((J.Return) stmt).getExpression();
+                        }
+                        if (!(stmt instanceof J.MethodInvocation)) {
+                            return true;
+                        }
+                        J.MethodInvocation remote = (J.MethodInvocation) stmt;
+                        if (!"remote".equals(remote.getSimpleName())) {
+                            return true;
+                        }
+                        if (remote.getArguments().isEmpty() || !(remote.getArguments().get(0) instanceof J.FieldAccess)) {
+                            return true;
+                        }
+                        J.FieldAccess fieldAccess = (J.FieldAccess) remote.getArguments().get(0);
+                        if (!(fieldAccess.getTarget() instanceof J.Identifier)) {
+                            return true;
+                        }
+                        J.Identifier target = (J.Identifier) fieldAccess.getTarget();
+                        return !"develocity".equals(target.getSimpleName()) && !"gradleEnterprise".equals(target.getSimpleName());
                     }
                 });
     }
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/RemoveDevelocityTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/RemoveDevelocityTest.java
@@ -158,76 +158,6 @@ class RemoveDevelocityTest implements RewriteTest {
     }
 
     @Test
-    void removeHttpBuildCache() {
-        rewriteRun(
-          settingsGradle(
-            """
-              plugins {
-                  id 'com.gradle.develocity' version '3.17'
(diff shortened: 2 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -1081,18 +1081,33 @@ public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifa
                     MavenPomCache cache = MavenExecutionContextView.view(ctx).getPomCache();
                     ResolvedPom resolvedPom = cache.getResolvedDependencyPom(dPom.getGav());
                     if (resolvedPom == null) {
-                        resolvedPom = new ResolvedPom(dPom, getActiveProfiles(), emptyMap(),
-                                emptyList(), true, initialRepositories, emptyList(), emptyList(),
-                                emptyList(), emptyList(), emptyList(), emptyList());
+                        resolvedPom = new ResolvedPom(
+                                dPom,
+                                getActiveProfiles(),
+                                dPom.getProperties(),
```
timtebeek: It looks like the specific test we added still passes even with the original argument passed in here, possibly because of the `getValue(...)` calls below; do we still need this change or can it be reverted?
```suggestion
                                emptyMap(),
```

rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
```diff
@@ -1352,28 +1352,61 @@ public MockResponse dispatch(RecordedRequest recordedRequest) {
...
+        InMemoryExecutionContext ctx = new InMemoryExecutionContext();
+        MavenResolutionResult resolutionResult = doc.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow()
+          .resolveDependencies(new MavenPomDownloader(emptyMap(), ctx, null, null), ctx);
+        List<ResolvedDependency> deps = resolutionResult.getDependencies().get(Scope.Compile);
+        assertThat(deps)
+          .filteredOn(rd -> "io.netty".equals(rd.getGroupId()))
+          .filteredOn(rd -> "netty-tcnative-boringssl-static".equals(rd.getArtifactId()))
+          .isNotEmpty()
+          .extracting(ResolvedDependency::getClassifier)
+          .doesNotContain("${boring-ssl-classifier}");
```
timtebeek: Was surprised to learn there's actually two dependencies with different classifiers in here: `["", "linux-x86_64"]`
Was that correct & expected?

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -1179,17 +1194,20 @@ private Scope getDependencyScope(Dependency d2, ResolvedPom containingPom) {
         Scope scopeInContainingPom;
         //noinspection ConstantConditions
         if (d2.getScope() != null) {
+            // TODO: Unsure if should be resolving scope similarly to classifier below
             scopeInContainingPom = Scope.fromName(getValue(d2.getScope()));
         } else {
+            // TODO: Unsure if should be resolving group id / artifact id / version similarly to classifier
```
timtebeek: I think it would indeed make sense to have similar handling for the other GAV coordinates, as they might similarly be using properties (although perhaps less likely so).

steve-aom-elliott:
So based on what I've been able to see with raw Maven and on our discussion earlier, I do not believe dependencies make use of properties defined in your POM, so I can simplify a few things.

rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
```diff
@@ -1352,28 +1352,63 @@ public MockResponse dispatch(RecordedRequest recordedRequest) {
...
+        MavenResolutionResult resolutionResult = doc.getMarkers().findFirst(MavenResolutionResult.class).orElseThrow()
+          .resolveDependencies(new MavenPomDownloader(emptyMap(), ctx, null, null), ctx);
+        List<ResolvedDependency> deps = resolutionResult.getDependencies().get(Scope.Compile);
+        assertThat(deps)
+          .filteredOn(rd -> "io.netty".equals(rd.getGroupId()))
+          .filteredOn(rd -> "netty-tcnative-boringssl-static".equals(rd.getArtifactId()))
+          .isNotEmpty()
+          .extracting(ResolvedDependency::getClassifier)
+          .doesNotContain("${boring-ssl-classifier}")
+          .anyMatch(""::equals)
```
timtebeek: Would this be simpler to achieve the same?
```suggestion
          .contains("")
```

steve-aom-elliott: I wanted to check that there was both one that was an empty string and one that was not an empty string, but also not `"${boring-ssl-classifier}"`. I would imagine `contains("")` would match either of those cases and not necessarily prove both are there, unless I'm mistaken.

Ah, wait, misread, nevermind

rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
steve-aom-elliott: I've also added to the test a value for the placeholder in the properties and a check that it does not get used to be a bit more thorough. I believe we discussed this separately previously, but for the sake of seeing this later, it has both because the non-classifier version is an uber JAR that has a dependency on the `linux-x86_64` version as well as other platform ones. With the fixes included below, you'll actually see all the classifier flavours for the different platforms in the list of resolved dependencies, as the map storing them has been updated to use `groupId`, `artifactId`, `classifier` and `type` as opposed to only `groupId` and `artifactId`. The two mentioned in your comment both showed up because they were at different depths, but the old logic was constraining dependencies at the same depth to only the first with a particular `groupId` & `artifactId`, meaning only the first defined, `linux-x86_64` was appearing.

steve-aom-elliott:
Title of PR could possibly be made more concise for the sake of patch notes later, so feel free to rephrase it if something better strikes you.

PR #6465, opened 2025-12-26
Fix Recipe to remove build cache configuration for Develocity

div-bargali:
## What's changed?
Added a unit test that fails with the current recipe to remove Develocity Build Cache Configuration.

Edit - Added a boolean flag (optional) removeBuildCache to `RemoveDevelocityConfiguration` recipe. When passed as true, it adds the `buildCache` extension in the `settings.gradle(.kts)` file for removal.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite/issues/6466

Adding a test case that fails with the current implementation. Would like to contribute a fix that would remove the buildCache extension correctly.

There are no workarounds.

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java
@@ -89,7 +89,6 @@ public class RemoveDevelocityConfiguration extends Recipe {
     public List<Recipe> getRecipeList() {
         return Arrays.asList(
                 new RemoveExtension("develocity"),
-                new RemoveExtension("gradleEnterprise")
-        );
+                new RemoveExtension("gradleEnterprise"));
     }
 }


div-bargali:
Hi Team,
Opening a PR to add unit test scenarios where the current recipe fails. Will capture the details in an issue.

div-bargali:
Fixes Issue - https://github.com/openrewrite/rewrite/issues/6466

div-bargali:
Added the fix and updated the PR description to explain the fix.

I noticed that there is a composite recipe `RemoveDevelocity` that uses this recipe as well. (https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle.yml#L47)
Did not make any changes to it as I reckon it is best for users to explicitly pass the flag's value as `true` as there can be a case when Develocity's build cache is not configured through Gradle's `buildCache` extension but the extension is still present. Open to discuss on this.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/RemoveDevelocityConfiguration.java
```diff
@@ -15,13 +15,19 @@
...
 
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
 
 public class RemoveDevelocityConfiguration extends Recipe {
+    @Option(displayName = "Also remove buildCache configuration",
+            description = "Also remove buildCache configuration when removing Develocity Configurations. Defaults to false.",
+            required = false)
+    public Boolean removeBuildCache = false;
```
shanman190: Would it make more sense to just remove it automatically by detecting the `remote(develocity.buildCache)` directly? Then we wouldn't need the additional option which could lead to build errors when the plugin was removed.

div-bargali: I considered automatic detection of `remote(develocity.buildCache)`, but Develocity build cache can also be configured using generic `HttpBuildCache` with URLs pointing to Develocity servers or separate cache nodes. Reliably detecting all these patterns (URL matching, distributed nodes, custom configurations) would be complex and error-prone. 
The explicit boolean flag provides a safer, more predictable approach while still addressing the use case. Users can set `removeBuildCache: true` when they know their build cache is Develocity-specific.

div-bargali: If required, I can enhance the documentation and recipe description to guide users.

shanman190: So I don't think that we need to try to be exhaustive in covering the cases, but I think just removing the `buildCache` extension is a bit more than desirable for the base case. I'd expect the recipe to remove the `remote` section when it's known develocity is in use -- which we can glean by the specific argument usage. For the more course use case, I'd almost expect a user to utilize a declarative recipe in which they combine this recipe with the `gradle.RemoveExtension` recipe in order to perform their more desired, but courser update.

To illustrate a bit more specifically, running this recipe on one repository may generally produce the desired results. Even on many repositories within a specific organization it _may_ produce the right results, but it _may_ also cause harm in this situation if the user has modified other build cache settings and wants to retain those. As you've mentioned, build cache is a Gradle core feature that Develocity enhances, but can be utilized entirely in the absence of Develocity.

div-bargali: @shanman190 Great feedback! It makes sense to limit the recipe to just remove the `remote` method from the `buildCache` extension. I can create a separate recipe that will handle removing the `remote` method and we can remove the boolean flag also. Before implementing, I wanted to get your input on the approach:

**Option A (Focused)**: Create `RemoveRemoteBuildCache` specifically for removing build cache configurations by provider pattern (develocity.buildCache, gradleEnterprise.buildCache, HttpBuildCache etc.)

- ✅ Clear purpose and easy to understand
- ✅ Directly solves the immediate problem
- ✅ Simpler to maintain and test
- ✅ Natural fit alongside existing `RemoveExtension` recipe
- ✅ Users can immediately understand what it does from the name

**Option B (Generic)**: Create a generic recipe that can remove any method invocation from any closure (buildCache, plugins, dependencies, etc.)

- ✅ More reusable across different DSL constructs
- ⚠️ More complex API 
- ⚠️ Might be over-engineering for current needs
- ⚠️ Could be extracted later if we see the pattern repeat

Both correctly handle preserving local cache config and removing empty blocks. 

I'm inclined toward **Option A** since it keeps the recipe catalog focused. We can always extract commonality later if similar needs emerge.

Does this align with OpenRewrite's philosophy? I noticed recipes like `RemoveExtension` and `RemoveBuildPlugin` are also focused on specific use cases rather than being maximally generic.

timtebeek: hi @div-bargali ; thanks for your enthusiasm here! I'd lean towards just removing that method invocation within the existing `RemoveDevelocityConfiguration` recipe, as I don't expect folks to separately want to run any new recipe specifically for that use case of removing that line from `buildCache`. If you'd want to keep the functionality isolated you could use a separate visitor, but I don't think that's necessary here as I'd expect the overall complexity to remain low.

div-bargali: @timtebeek I have updated the recipe to handle the removal of `remote` method from the `buildCache` extension automatically.

div-bargali:
> Thanks for the help here @div-bargali ! I've condensed your work a bit by overriding `getVisitor` and applying some other best practices like `ListUtils.filter`. Great to see this resolved.

Thanks @timtebeek. Glad I could help in contributing to the project!

timtebeek:
All good; after discussing with Shannon we've added a small update to further limit what's removed vs retained.
- https://github.com/openrewrite/rewrite/pull/6476

div-bargali:
> All good; after discussing with Shannon we've added a small update to further limit what's removed vs retained.
> 
> * [Only remove `gradleEnterprise` and `develocity.buildCache` #6476](https://github.com/openrewrite/rewrite/pull/6476)

Thanks for sharing @timtebeek . It makes sense to restrict removal of `remote(HttpBuildCache)` block, however I'd like to point out that I have seen Develocity's Build Cache being configured using the HTTP connector more often than the Develocity's connector.
Maybe in the future we can add an optional flag for users who want to also remove the HttpBuildCache's configuration after evaluation of any reported issue. Thanks!

PR #6471, opened 2025-12-30
Document usages of the various `ClasspathScanningLoader` methods

timtebeek:
No longer contains any functional changes; just documents how the various methods are being used.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
@@ -52,7 +52,8 @@ public class ClasspathScanningLoader implements ResourceLoader {
     private @Nullable Runnable performScan;
 
     /**
-     * Construct a ClasspathScanningLoader scans the runtime classpath of the current java process for recipes
+     * Construct a ClasspathScanningLoader that scans the runtime classpath of the current java process for recipes,
+     * mostly for use in tests.
      *
      * @param properties     Yaml placeholder properties
      * @param acceptPackages Limit scan to specified packages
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -93,6 +93,34 @@ public ClasspathScanningLoader(Properties properties, ClassLoader classLoader) {
...
+     * @param properties    YAML placeholder properties
+     * @param onlyAcceptJar Limit scan to this jar file
+     * @param classLoader   Limit scan to classes loadable by this classloader
+     */
+    public ClasspathScanningLoader(Properties properties, Path onlyAcceptJar, ClassLoader classLoader) {
+        this.classLoader = classLoader;
+        this.recipeLoader = new RecipeLoader(classLoader);
+        this.performScan = () -> {
+            scanClasses(new ClassGraph()
+                            .acceptJars(onlyAcceptJar.getFileName().toString())
```
timtebeek: This line ensures only we only load from the single recipe jar we're interested in from:
- https://github.com/openrewrite/rewrite-build-gradle-plugin/pull/149

rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
```diff
@@ -112,19 +116,50 @@ public ClasspathScanningLoader(Path jar, Properties properties, Collection<? ext
         };
     }
 
-    public static ClasspathScanningLoader onlyYaml(Properties properties) {
+    /**
+     * Construct a ClasspathScanningLoader to load Yaml categories and recipes from the runtime classpath, as part of
+     * running tests or inferring local recipe categories.
+     */
+    public static ClasspathScanningLoader onlyYaml(Properties properties) { // Environment.scanYamlResources
```
timtebeek:
```suggestion
    public static ClasspathScanningLoader onlyYaml(Properties properties) {
```

jkschneider (approved):

PR #6473, opened 2025-12-31
Allow 'if' as HCL attribute name

### LINKED ISSUE #6446: HclParser might fail if you use `if` as property name

Current main = 9e52e9987a45d3310795a9c5834783f9e8b53216

The following test when added to `HclParserTest`:
```java
    @Test
    void usingIf() {
        rewriteRun(
          hcl(
            """
            resource "not_a_real_one_for_sure" deny {
              if = "a"
            }
            """
          )
        );
    }
```

fails with:
```
Caused by: org.openrewrite.hcl.HclParsingException: Syntax error in file.tf at line 2:2 mismatched input 'if' expecting {'null', '}', Identifier}.
	at org.openrewrite.hcl.HclParser$ForwardingErrorListener.syntaxError(HclParser.java:108)
	... 26 more
	at org.antlr.v4.runtime.DefaultErrorStrategy.recoverInline(DefaultErrorStrategy.java:485)
	at org.antlr.v4.runtime.Parser.match(Parser.java:208)
	at org.openrewrite.hcl.internal.grammar.HCLParser.blockExpr(HCLParser.java:1111)
	... 21 more
```

while I think this is a valid HCL source.
Terraform doesn't seem to have a problem with it.

## OSS Repro

- https://github.com/Azure/ArtistAnywhere/blob/bb6b98724607a36589469911a7531f4d677b6880/0.Foundation/policy.tf#L33

Gautam-aman:
Fixes parsing failure when if is used as an attribute name in HCL.

Terraform allows keywords as attribute names, but the current grammar only accepts Identifier, causing valid HCL to fail parsing.

This change allows if in attribute position only and adds a regression test.

- Fixes #6446


timtebeek:
Thanks for looking into this @Gautam-aman ; are you sure you pushed all the changes you were after? As I'm seeing no reference that `identifierLike`, and a failure in running the tests.

Gautam-aman:
This PR is now ready for review.

The grammar and visitor changes allow `if` to be used as an HCL attribute key
(e.g. `if = true`) while preserving its role in conditional and for expressions.

Local build and tests pass:
- :rewrite-hcl:compileJava
- :rewrite-hcl:test

Happy to address any feedback.

rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/grammar/JsonPathParserVisitor.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Generated from ~/git/rewrite/rewrite-hcl/src/main/antlr/JsonPathParser.g4 by ANTLR 4.13.2
+// Generated from /Users/rigelsmacbook/PR/rewrite/rewrite-hcl/src/main/antlr/JsonPathParser.g4 by ANTLR 4.13.2
```
timtebeek: Can we revert these accidental changes here and elsewhere above?
```suggestion
// Generated from ~/git/rewrite/rewrite-hcl/src/main/antlr/JsonPathParser.g4 by ANTLR 4.13.2
```

rewrite-hcl/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ dependencies {
     compileOnly(project(":rewrite-test"))
 
     implementation("org.antlr:antlr4-runtime:4.13.2")
+    implementation("org.antlr:antlr4:4.13.2")
```
timtebeek: From the changes I'm seeing I'd be surprised if we need this addition.
```suggestion
```

rewrite-hcl/src/main/antlr/HCLParser.g4
```diff
@@ -31,6 +31,13 @@ blockLabel
     | Identifier
     ;
 
+identifierLike
+    : Identifier
+    | IF
```
timtebeek, line 36: @greg-at-moderne any idea whether `if` is the only keyword that would need special handling? Or could there be others too?

Gautam-aman:
All review feedback has been addressed and the PR is ready.
Thanks for taking a look!

rewrite-hcl/src/main/java/org/openrewrite/hcl/internal/grammar/JsonPathParserVisitor.java
timtebeek: Would be good to revert this same change elsewhere as well; did you look over https://github.com/openrewrite/rewrite/pull/6473/files ?

Gautam-aman:
Thanks for the review and merge , appreciate it!

timtebeek:
Thanks! There's one open question that I'd like @greg-at-moderne to weigh in on; please don't resolve that comment before then:
- https://github.com/openrewrite/rewrite/pull/6473#discussion_r2655801436

Mostly I'm wondering if the fix here is too specific for the problem that was reported, whereas I think there might be similar cases that we'd want to tackle here as well.

Gautam-aman:
I agree it’s worth considering whether this should be generalized beyond the reported case. I’m happy to wait for @greg-at-moderne’s input and adjust the approach if needed.

Let me know how you’d like to proceed.

rewrite-hcl/src/main/antlr/HCLParser.g4
greg-at-moderne: That's a good point. I think it's not the only one.
Another real-life example is `in`, see this real-world mention:
https://github.com/terraform-aws-modules/terraform-aws-fsx/blob/ac8fe39a0d2428c30be6ed96a2d13026e0306e41/examples/lustre/main.tf#L73

So I think eventually we need to allow all (or at least nearly all) keywords.

Having said that, I think this PR is a net improvement.
@Gautam-aman - would you like to continue the effort and change the impl to support keywords too, not `if` specifically?

timtebeek: As a potentially larger change we can optionally rework the parser to specify where it should expect identifiers using boolean fields as also seen on
- https://github.com/openrewrite/rewrite/pull/6509

I've found Claude to be quite efficient to pick up this work, so you might not want to tackle all of that manually.

Gautam-aman: Thanks, that makes sense.
I agree the broader keyword handling and parser rework would be a larger change. For this PR, I’d prefer to keep the scope limited to if so it remains small and safe.

I’m happy to look into a follow-up PR for broader keyword support or parser refactoring if that would be useful.

timtebeek: It's more that your current change would likely result in an out-of-place keyword `IF` and LST element, where an identifier is expected. Let me try quickly if I can change your test to show that issue.

Gautam-aman: Thanks for checking that makes sense.
If the test exposes an unintended case where IF is accepted outside identifier positions, I’m happy to adjust the implementation accordingly once we see the result.

PR #6478, opened 2026-01-03
Bug: Incorrect plugin insertion order in settings.gradle with pluginManagement and buildscript

div-bargali:
## What's changed?

Added failing unit tests to demonstrate incorrect plugin ordering in `settings.gradle[.kts]` files when both `pluginManagement` and `buildscript` blocks are present.

**New tests added:**
- `addPluginWithPluginManagementAndBuildscriptBlocks()` - Groovy DSL test
- `addPluginWithPluginManagementAndBuildscriptBlocksKotlin()` - Kotlin DSL test

These tests currently fail because `AddPluginVisitor` incorrectly inserts the `plugins` block between `pluginManagement` and `buildscript`, violating Gradle's required ordering.

## What's your motivation?

- Closes https://github.com/openrewrite/rewrite/issues/6479

When `AddSettingsPlugin` is applied to a `settings.gradle` file containing both `pluginManagement` and `buildscript` blocks, it produces invalid Gradle syntax that causes build failures with the error:
only buildscript {} and other plugins {} script blocks are allowed before plugins {} block

## Anything in particular you'd like reviewers to focus on?

- Verify that the test cases accurately represent the expected Gradle ordering requirements
- Confirm that both Groovy and Kotlin DSL tests properly cover the edge case
- Review the `interpolateResolvedVersionKotlin()` helper method for Kotlin tests

## Any additional context

This PR includes the failing tests to document the bug before implementing the fix. The fix will be submitted in a follow-up commit.

**Impact:** This issue affects any project using `AddSettingsPlugin` on settings files that contain both `pluginManagement` and `buildscript` blocks.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #6480, opened 2026-01-05
Fix Add Plugin visitor

### LINKED ISSUE #6479: Bug: Incorrect plugin insertion order in settings.gradle with pluginManagement and buildscript

I am using:
- rewrite-gradle v8.x (latest)

This issue affects the `AddSettingsPlugin` recipe (and the underlying `AddPluginVisitor` class). The bug occurs when running the recipe via Maven plugin, Gradle plugin, or Moderne CLI on projects with `settings.gradle` or `settings.gradle.kts` files.

Added unit tests to cover the case in PR - https://github.com/openrewrite/rewrite/pull/6478

## Additional Context
- This issue only affects settings.gradle[.kts] files. The build.gradle[.kts] handling is correct since pluginManagement cannot appear in build files.
- The same bug exists in both the Groovy and Kotlin code paths.
- Gradle documentation states that in settings files, pluginManagement must be first, and buildscript (if present) must come before plugins.

## Are you interested in contributing a fix to OpenRewrite?
Yes, I'm working on a fix and will submit a PR with:
- The corrected logic to check for both pluginManagement and buildscript blocks
- Unit tests covering this scenario for both Groovy and Kotlin DSL (addPluginWithPluginManagementAndBuildscriptBlocks and addPluginWithPluginManagementAndBuildscriptBlocksKotlin)

Laurens-W:
## What's changed?

Check for the right file to be present

## What's your motivation?

The plugins block was being added as the first entry in a gradle kotlin settings file which left it non compiling when plugin management is also defined

- Closes #6479
- Closes #6478

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddPluginVisitor.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddPluginVisitor.java
@@ -360,7 +360,7 @@ public class AddPluginVisitor extends JavaIsoVisitor<ExecutionContext> {
         }.reduce(cu, new AtomicBoolean());
         if (!hasPluginsBlock.get()) {
             J.Block block = (J.Block) cu.getStatements().get(0);
-            if (cu.getSourcePath().endsWith(Paths.get("settings.gradle")) &&
+            if (cu.getSourcePath().endsWith(Paths.get("settings.gradle.kts")) &&
                     !block.getStatements().isEmpty() &&
                     block.getStatements().get(0) instanceof J.MethodInvocation &&
                     "pluginManagement".equals(((J.MethodInvocation) block.getStatements().get(0)).getSimpleName())) {
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/AddSettingsPluginTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/AddSettingsPluginTest.java
@@ -18,6 +18,7 @@ package org.openrewrite.gradle.plugins;
 import org.intellij.lang.annotations.Language;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.groovy.tree.G;
+import org.openrewrite.kotlin.tree.K;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 import org.openrewrite.test.SourceSpec;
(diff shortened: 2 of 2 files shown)


timtebeek:
Anything worth copying over from 
- https://github.com/openrewrite/rewrite/pull/6478

Laurens-W:
> Anything worth copying over from
> 
> * [[WIP] Bug: Incorrect plugin insertion order in settings.gradle with pluginManagement and buildscript #6478](https://github.com/openrewrite/rewrite/pull/6478)

Perhaps the Gradle Groovy version of the test, other than that this PR embodies the problem at hand and solves it as well

Laurens-W:
Actually a Groovy Gradle test variant already exists, and the buildscript section does not appear relevant to the issue

timtebeek (approved):

div-bargali:
Hi @Laurens-W
While I appreciate the efforts, this PR does not resolve the bug https://github.com/openrewrite/rewrite/issues/6479 at all.
It is entirely missing the tests included in the initial PR which has both `pluginManagement` and `buildscript` blocks in `settings.gradle(.kts)` file. Also there is not fix in this PR for the issue I had highlighted.

Although you have also noticed the issue with (.kts) files, but even that fix is incomplete as the Kotlin AST is not parsed correctly. Please check the original PR for the fix.

P.S. - Have been working on this issue for a couple of days, a heads-up or a comment on the issue would have helped if you're interested in collaborating on the fix, we could've avoided duplicate efforts 🙂

Laurens-W:
> Hi @Laurens-W While I appreciate the efforts, this PR does not resolve the bug #6479 at all. It is entirely missing the tests included in the initial PR which has both `pluginManagement` and `buildscript` blocks in `settings.gradle(.kts)` file. Also there is not fix in this PR for the issue I had highlighted.
> 
> Although you have also noticed the issue with (.kts) files, but even that fix is incomplete as the Kotlin AST is not parsed correctly. Please check the original PR for the fix.
> 
> P.S. - Have been working on this issue for a couple of days, a heads-up or a comment on the issue would have helped if you're interested in collaborating on the fix, we could've avoided duplicate efforts 🙂

Hey @div-bargali! I hadn't checked whether there was an open effort on this front yet, and luckily I don't have much time invested yet! Glad we're in touch now! It was indeed pointed out to me that the test case I've added already worked, something to do with `pluginManagement` being the first block.

Do you have a branch with your progress that I can take a look at? That way we can collaborate and resolve this :)

timtebeek:
I see the same change has gone into https://github.com/openrewrite/rewrite/pull/6478 as well
- https://github.com/openrewrite/rewrite/pull/6478

Let's continue there based on the feedback above.

PR #6488, opened 2026-01-06
Fix issue where BOMs/platforms defined in Gradle Module Metadata could end up erroneously listed as requested dependencies in our maven model

sambsnyd:
This prevented `RecipeBundleResolver` / `Retriever` from installing rewrite-core, which led me to notice and track down the inaccuracies in the dependency model.


sambsnyd:
Shannon reviewed this with me over slack

PR #6489, opened 2026-01-07
Ensure `plugins` block is added after `pluginManagement`, even when `pluginManagement` was added just before

Laurens-W:
## What's changed?

Instead of adding pluginManagement as a block it now get's added as a methodInvocation

## What's your motivation?

When a pluginManagement block was added by `AddSettingsPluginRepository` the `AddSettingsPlugin` would not identify it correctly causing the settings file to get into an invalid state

### Checklist

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
@@ -157,9 +157,9 @@ public class AddSettingsPluginRepository extends Recipe {
                     }));
                     statements.set(0, m);
                 } else {
-                    Statement pluginManagementStatement = pluginManagement instanceof J.Block
-                            ? ((J.Block) pluginManagement).getStatements().get(0)
-                            : (J.MethodInvocation) pluginManagement;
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
```diff
@@ -157,7 +157,10 @@ private List<Statement> addPluginManagementRepos(List<Statement> statements, J p
                     }));
                     statements.set(0, m);
                 } else {
-                    statements.add(0, pluginManagement instanceof J.Block ? (J.Block) pluginManagement : (J.MethodInvocation) pluginManagement);
+                    Statement pluginManagementStatement = pluginManagement instanceof J.Block ?
+                            ((J.Block) pluginManagement).getStatements().get(0) :
```
timtebeek, line 161: I'd briefly explored whether this should be part of the `methodExtractor` passed into `generatePluginManagementBlock`, but couldn't immediately get it to work that way. Did you consider / explore that option?

Laurens-W: I hadn't actually

PR #6495, opened 2026-01-08
UUID generation fallback for Node version pre-14.17.0

bmuschko:
## What's changed?

`crypto.randomUUID()` cannot be accessed by projects with Node version before 14.17.6.

## What's your motivation?

`crypto.randomUUID()` cannot be accessed by projects with Node version before 14.17.6. Build the LST with an older version provides the following stack trace:

```
Error: Exit code 1
[eval]:3
console.log(randomUUID());
         ^

TypeError: randomUUID is not a function
    at [eval]:3:13
    at Script.runInThisContext (vm.js:131:20)
    at Object.runInThisContext (vm.js:297:38)
    at Object.<anonymous> ([eval]-wrapper:10:26)
```

The proposed logic looks as follows:

1. Checks if native `crypto.randomUUID()` is available (Node 14.17.0+)
2. If available, uses the native implementation for best performance
3. If not, falls back to generating a UUID v4 using `crypto.randomBytes(16)` with proper version and variant bits set

This logic avoids having to pull in an external dependency, e.g. a polyfill.

No

@knutwannheden @greg-at-moderne 

There's no real workaround for the issue.

## Any additional context

Organizations with a large suite of Node application will have projects that still use an older version of Node. We'll need to support them with generating the LST for the legacy project.

It looks like we don't have a GitHub Actions workflow in place for multi-version testing. I would suggest to add that at a later time.

### DIFF
--- a/rewrite-javascript/rewrite/src/uuid.ts
+++ b/rewrite-javascript/rewrite/src/uuid.ts
@@ -42,5 +42,5 @@ function fallbackRandomId(): UUID {
  * selected once at module load time to avoid per-call overhead.
  */
 export const randomId: () => UUID = typeof crypto.randomUUID === 'function'
-    ? () => crypto.randomUUID()
+    ? crypto.randomUUID
     : fallbackRandomId;


rewrite-javascript/rewrite/test/uuid.test.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2025 the original author or authors.
```
timtebeek:
```suggestion
 * Copyright 2026 the original author or authors.
```
Happy new year!

rewrite-javascript/rewrite/src/uuid.ts
```diff
@@ -13,10 +13,26 @@
...
+    // Set version 4 bits (0100xxxx in byte 6)
+    bytes[6] = (bytes[6] & 0x0f) | 0x40;
+
+    // Set variant bits (10xxxxxx in byte 8)
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
+
+    // Convert to hex string with dashes
+    const hex = bytes.toString('hex');
+    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
 }
```
knutwannheden, line 35: Since `randomId()` is on the hot path and called frequently, we should move the feature detection to module load time to avoid the conditional check on every call.

```suggestion
import * as crypto from "crypto";

export type UUID = string;

function nativeRandomId(): UUID {
    return crypto.randomUUID();
}

function fallbackRandomId(): UUID {
    const bytes = crypto.randomBytes(16);

    // Set version 4 bits (0100xxxx in byte 6)
    bytes[6] = (bytes[6] & 0x0f) | 0x40;

    // Set variant bits (10xxxxxx in byte 8)
    bytes[8] = (bytes[8] & 0x3f) | 0x80;

    // Convert to hex string with dashes
    const hex = bytes.toString('hex');
    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
}

// Select implementation once at module load time
export const randomId: () => UUID = typeof crypto.randomUUID === 'function'
    ? nativeRandomId
    : fallbackRandomId;
```

This way the conditional is evaluated only once when the module is loaded, and subsequent calls to `randomId` go directly to the appropriate implementation without any runtime branching.

PR #6496, opened 2026-01-08
Add style parameter to OrderImports.

sambsnyd:
### Checklist


Jenson3210:
I am thinking creating a recipe that puts a style marker might be more flexible. 
Otherwise we might end up with a bunch of other locations where suddenly style could become a property.

This way we can have a recipe with recipelist:
- AddStyleMarker (specify the imports style order in this caser)
- OrderImports (will then detect/use the last added style marker)

Or also any other recipe that uses the style detection can have modified stylistic behavior without needing to format the entire file as that recipe would "ONLY" add the marker to the sourcefile.

PR #6497, opened 2026-01-08
RPC performance: sync receive queue and compact array format

knutwannheden:
Optimizes RPC communication between Java and TypeScript for improved performance when processing large ASTs.

### Wire format change

Before (JSON objects):
```json
{
  "state": "ADD",
  "valueType": "org.openrewrite.java.tree.J$Identifier",
  "value": null,
  "ref": 42
}
```

After (compact arrays):
```json
[1, "org.openrewrite.java.tree.J$Identifier", null, 42]
```

### Key changes

- **Sync receive queue**: Convert `RpcReceiveQueue` from async to sync by pre-fetching all data batches upfront, eliminating async/await overhead
- **Compact wire format**: Use compact arrays `[state, valueType, value, ref?, trace?]` instead of JSON objects for RPC messages in both directions
- **Benchmark support**: Add `reset()` methods for cache clearing between benchmark iterations


rewrite-core/src/main/java/org/openrewrite/rpc/RpcObjectData.java
```diff
@@ -15,28 +15,41 @@
...
  * A single piece of data in a tree, which can be a marker, leaf value, tree element, etc.
+ * <p>
+ * Serialized as a compact array: [state, valueType, value, ref?, trace?]
+ * - 3 elements when ref is null and trace is null
+ * - 4 elements when ref is present but trace is null
+ * - 5 elements when trace is present
  */
 @Value
-@NoArgsConstructor(force = true, access = AccessLevel.PRIVATE, onConstructor_ = @JsonCreator)
+@NoArgsConstructor(force = true, access = AccessLevel.PRIVATE)
```
jkschneider, line 44: Rather than changing this at RpcObjectData and in the specification itself, can we move the optimization to compact arrays to only affect the interaction between Java and JS because JS is slow at Object.keys() used for serializing in an object rather than a compact array?

jkschneider (changes requested):

PR #6499, opened 2026-01-08
In yaml `CopyValue`, invoke `UnfoldProperties` after `MergeYaml`

dsgrieve:
## What's changed?

Call `org.openrewrite.yaml.UnfoldProperties` after MergeYaml in `org.openrewrite.yaml.CopyValue`

## What's your motivation?

CopyValue adds new keys to the end of the yaml file. While this is syntacticly correct, it would be better to fold those new keys into the existing keys where appropriate. 

### Checklist


timtebeek (approved):
Great to see this fixed, thanks!

PR #6501, opened 2026-01-09
Add support for method pointers in Groovy.

### LINKED ISSUE #6373: Groovy mishandling `.&` method references (dot ampersand)

Current main = 52f563be74b56be499de3f531ac01cc11014905b

```java
    @Test
    void dotAmpersand() {
        rewriteRun(
          groovy(
            """
            class Validator {
                void validate(Closure logger) {
                    logger("Something went wrong")
                }
            }
        
            def v = new Validator()
        
            // Use Groovy method-pointer syntax on a Java object:
            v.validate(System.err.&println)
            """
          )
        );
    }
```
case added to one of the Groovy unit tests fails with:
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "file.groovy":
```

```diff
 // Use Groovy method-pointer syntax on a Java object:
-v.validate(System.err.&println)
+v.validate(System.err::.&println)
```

## OSS repro
- https://github.com/apache/creadur-rat/blob/4e6919e0f66539579014120c4852fc3e77910c60/apache-rat-core/src/it/resources/ReportTest/RAT_14/verify.groovy#L69C1-L70C1

BrendanHart:
## What's changed?

Add support for Groovy's method pointers.

## What's your motivation?

- Fixes #6373 

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2000,7 +2000,7 @@ public class GroovyParserVisitor {
                 referenceName = ((ConstantExpression) ref.getMethodName()).getValue().toString();
             }
 
-            if (operation.equals(".&")) {
+            if (".&".equals(operation)) {
                 queue.add(new G.MethodPointer(randomId(), fmt, Markers.EMPTY,
                         padRight(expression, opPrefix),
                         padLeft(whitespace(), new J.Identifier(randomId(),


timtebeek:
Thanks for the help here @BrendanHart ! Unfortunately the issue that was opened failed to reference this open PR
- https://github.com/openrewrite/rewrite/pull/5782

That PR has been open for some time as we're waiting for a final approval on the Marker added there, where your approach adds a new LST element MethodPointer. Wherever possible we strive to using existing LST elements with markers for the printer, as opposed to new LST elements. That way we can have maximum effect from recipes that target existing LST elements without needing dedicated visitors for specific elements. As such I'm afraid I have a slight preference for the existing PR, but I'll link yours on the PR for comparison in that final review.

Sorry for the duplicate effort here! Wasn't aware you were considering to work on this, otherwise I'd have been able to reference this earlier work.

BrendanHart:
Ah, no problem at all, I just thought I'd pick an open issue to try out Gemini :) I will close this! Thanks @timtebeek

timtebeek:
Much appreciated; feel free to tag me on any issue that you're considering for guidance.

PR #6505, opened 2026-01-09
Altering check for `AddDependency` to compare versions differently when direct vs indirect dependency.

steve-aom-elliott:
## What's changed?
Changing the version check for initial compile scope check in `AddDependency` wherein now, if a direct dependency, it will check version vs requested, as `AddDependencyVisitor` can sort out the actual version needed, while if a transitive dependency, it will allow it, which can mean things like `3.x` will allow non-latest from transitive dependencies.

## What's your motivation?
Failure in rewrite-spring pipeline

### Checklist


steve-aom-elliott:
> If you can confirm it's fixed in rewrite-spring with this change then OK to merge, thanks! :)

I've verified it does pass the test with this change now

PR #6509, opened 2026-01-10
Add DockerParser and initial recipes

timtebeek:
## What's changed?
- Added rewrite-docker, with a DockerParser, Docker interface and nested types, as well as some initial recipes
- Verified locally against 1492 docker images from these repositories:
```
bitnami/containers@main
CentOS/CentOS-Dockerfiles@master
chainguard-images/images@main
docker-library/official-images@master
docker/awesome-compose@master
dotnet/dotnet-docker@main
fedora-cloud/Fedora-Dockerfiles@master
GoogleContainerTools/distroless@main
jessfraz/dockerfiles@master
kubernetes/kubernetes@master
linuxserver/docker-baseimage-alpine@3.22
moby/moby@master
vimagick/dockerfiles@master
```
This only leaves one type of issue with backticks and newlines in some 20 dotnet images that I think will be rare in practice.
- The lexer now maintains a field whether we are at the start of a _logical_ (non-heredocs / multiline) line before parsing keywords, which simplifies a lot of the logic we had for escapes/keywords/state tracking.

## What's your motivation?
We've had some use cases pop up recently that aim to modify Dockerfiles; while that's possible with text based replacements, it would be better with a structured parser and model to modify.

Are we confident enough now in the current tree classes and ANTLR based parser to roll this out?

It's been proposed to write a manual parser, but not yet clear what the benefit would be there.

## Any additional context
There was an earlier attempt that had been reverted in 2eae35abde29f12fa9fe3130544ee12caa026f23 at the time over some concerns, and because we were just before a release.
- https://github.com/openrewrite/rewrite/pull/6203

Before that there was an externally developed manual parser, with a different model
- Closes https://github.com/openrewrite/rewrite-docker/pull/17


rewrite-docker/src/main/java/org/openrewrite/docker/tree/Docker.java
```diff
@@ -0,0 +1,975 @@
...
+    @With
+    class EnvironmentVariable implements ArgumentContent {
+        @EqualsAndHashCode.Include
+        UUID id;
+
+        Space prefix;
+        Markers markers;
+
+        String name;
+        boolean braced;
```
timtebeek, line 1045: Right now we capture everything as if it were the name, but technically this could be split out into a name and say a fallback value, as documented here: https://docs.docker.com/reference/dockerfile/#environment-replacement
Would we want to make that distinction here?

rewrite-docker/src/main/java/org/openrewrite/docker/tree/Docker.java
```diff
@@ -0,0 +1,975 @@
...
+    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    @With
+    class ShellForm implements CommandForm {
+        @EqualsAndHashCode.Include
+        UUID id;
+
+        Space prefix;
+        Markers markers;
+
+        List<Argument> arguments;
```
timtebeek: This could likely be a list of literals if I'm reading the docs here correctly, with substitution done by the shell, not the builder
- https://docs.docker.com/reference/dockerfile/#environment-replacement
- https://docs.docker.com/reference/dockerfile/#variable-substitution
```suggestion
        List<Literal> arguments;
```

rewrite-docker/src/main/java/org/openrewrite/docker/search/FindDockerBaseImages.java
```diff
@@ -0,0 +1,140 @@
...
+            private @Nullable String extractText(Docker.@Nullable Argument arg) {
+                if (arg == null) {
+                    return null;
+                }
+                StringBuilder builder = new StringBuilder();
+                for (Docker.ArgumentContent content : arg.getContents()) {
+                    if (content instanceof Docker.Literal) {
+                        builder.append(((Docker.Literal) content).getText());
+                    } else if (content instanceof Docker.EnvironmentVariable) {
+                        Docker.EnvironmentVariable env = (Docker.EnvironmentVariable) content;
```
timtebeek, line 108: We can optionally try to weave in the default values for global args, stage args up to this point, and env defaults to get a more complete picture here.

rewrite-docker/src/test/java/org/openrewrite/docker/tree/RunTest.java
```diff
@@ -0,0 +1,539 @@
...
+
+                // Second argument includes comma in its prefix
+                Docker.Literal second = execForm.getArguments().get(1);
+                assertThat(second.getPrefix().getWhitespace()).contains(",");
+                assertThat(second.getText()).isEqualTo("update");
+
+                // Third argument includes comma in its prefix
+                Docker.Literal third = execForm.getArguments().get(2);
+                assertThat(third.getPrefix().getWhitespace()).contains(",");
+                assertThat(third.getText()).isEqualTo("-y");
```
timtebeek, line 513: We'll want to ensure there's no non-whitespace in the whitespace elements here, or generically in Assertions. Exploring that final option now.

rewrite-docker/src/main/java/org/openrewrite/docker/tree/Docker.java
sambsnyd: Are any spaces allowed if there are braces?  All of the examples show `${foo}` and never `${ foo }`, but do we know that docker refuses to accept the latter? Might need more than a boolean if docker does not fail the latter with a terminal error

timtebeek: It really does have to be either `$str` or `${str}`, otherwise you get
```
Sending build context to Docker daemon  2.048kB
Step 1/3 : FROM alpine
 ---> 7e01a0d0a1dc
Step 2/3 : ENV str="String"
 ---> Running in cbf326bf64e4
 ---> Removed intermediate container cbf326bf64e4
 ---> 2ac27d5d5cfa
Step 3/3 : RUN echo ${ str }
 ---> Running in dacc101eb319
/bin/sh: syntax error: bad substitution
The command '/bin/sh -c echo ${ str }' returned a non-zero code: 2
```
https://docs.docker.com/reference/dockerfile/#environment-replacement

PR #6516, opened 2026-01-12
Add `new JsonMessageFormatter` argument to `new JsonRpc`

timtebeek:
Fix compilation issue following upstream change.


rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpcProcess.java
```diff
@@ -139,7 +139,7 @@ public synchronized void start() {
         if (trace) {
             handler = new TraceMessageHandler("client", handler);
         }
-        this.rpcClient = new JsonRpc(handler);
+        this.rpcClient = new JsonRpc(handler, new JsonMessageFormatter(module));
```
timtebeek, line 142: This was the only usage where I could find a `module` to pass in. The others were test usages.

timtebeek:
Reverted upstream for now.

PR #6519, opened 2026-01-12
Retain imported annotations on `package-info.java`

timtebeek:
We saw this fail in our review bot, leading to spurious comments.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/RemoveUnusedImportsTest.java
@@ -560,22 +560,7 @@ class RemoveUnusedImportsTest implements RewriteTest {
         );
     }
 
-    @Test
-    void retainPackageInfoAnnotation() {
-        rewriteRun(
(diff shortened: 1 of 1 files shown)


rewrite-java/src/main/java/org/openrewrite/java/RemoveUnusedImports.java
```diff
@@ -125,6 +125,15 @@ public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionCon
...
             }
 
+            // Collect simple names of package annotations for fallback when type info is missing
+            // (TypesInUse now handles the case when type info IS available)
+            Set<String> packageAnnotationSimpleNames = new HashSet<>();
+            if (cu.getPackageDeclaration() != null) {
+                for (J.Annotation annotation : cu.getPackageDeclaration().getAnnotations()) {
+                    packageAnnotationSimpleNames.add(annotation.getSimpleName());
+                }
+            }
```
timtebeek: Turns out `FindMissingTypes` is failing to pick up on the missing annotation type, so we should fix that and remove this.

PR #6522, opened 2026-01-13
Fix `KotlinTemplate` building for annotations

Laurens-W:
## What's changed?

Introduces extensibility to the AnnotationTemplateGenerator so that other languages (like Kotlin) can intercept certain steps
Introduced Kotlin specific AnnotationTemplateGenerator so that KotlinTemplate can create valid snippets for annotations

## What's your motivation?

`AddMockitoExtensionIfAnnotationsUsed`  currently adds the annotation with a Java style class reference, I'm working on expanding that to work for Kotlin

## Any additional context

Without this change a KotlinTemplate like this `KotlinTemplate.builder("@ExtendWith(MockitoExtension::class)")` will complain about the `$` before Class and Placeholder
```
\607541286333800\openRewriteFile.kt:6:7: error: name expected
class $Clazz {}
      ^
\607541286333800\openRewriteFile.kt:7:1: error: expected annotation identifier after '@'
@interface $Placeholder {}
^
\607541286333800\openRewriteFile.kt:7:12: error: name expected
@interface $Placeholder {}
           ^
```

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
@@ -17,7 +17,6 @@ package org.openrewrite.java.internal.template;
 
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Timer;
-import lombok.NoArgsConstructor;
 import lombok.RequiredArgsConstructor;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.Cursor;


rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
```diff
@@ -92,16 +90,26 @@ public String template(Cursor cursor, String template) {
...
-                            after.insert(0, "static class $Clazz {}");
+                        if (classCursor != null) {
+                            addDummyClass(classCursor, after);
                         }
                     }
                     return before + "/*" + TEMPLATE_COMMENT + "*/" + template + "\n" + after;
                 });
     }
 
+    public void addDummyClass(Cursor cursor, StringBuilder after) {
```
timtebeek: Does it make sense to reduce the visibility here to protected, to avoid exposed API confusing users?
```suggestion
    protected void addDummyClass(Cursor cursor, StringBuilder after) {
```

rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
```diff
@@ -92,16 +90,26 @@ public String template(Cursor cursor, String template) {
...
 
+    protected void addDummyClass(Cursor cursor, StringBuilder after) {
+        if (cursor.getParentOrThrow().getValue() instanceof JavaSourceFile) {
+            after.insert(0, "class $Clazz {}");
+        } else {
+            after.insert(0, "static class $Clazz {}");
+        }
+    }
+
+    protected void addDummyInterface(StringBuilder after) {
```
knutwannheden: `addDummyAnnotationType()`?

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinAnnotationTemplateGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+        super(imports);
+    }
+
+    @Override
+    protected void addDummyClass(Cursor cursor, StringBuilder after) {
+        if (cursor.getParentOrThrow().getValue() instanceof JavaSourceFile) {
+            after.insert(0, "class Clazz {}");
+        } else {
+            after.insert(0, "static class Clazz {}");
+        }
```
knutwannheden, line 34: Classes in Kotlin are static by default, so no need to differentiate here. The `static` modifier would actually result in a parse error.

As we in the Java code check for a class named `$Clazz` we should also here use that name. Identifiers with `$` are reserved in Kotlin, why you need to escape them. E.g.

```kt
class `$Clazz` {}
```

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinAnnotationTemplateGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Override
+    protected void addDummyInterface(StringBuilder after) {
+        after.append("\nannotation class Placeholder {}");
```
knutwannheden: Also here:

```kt
annotation class `$Placeholder` {}
```

rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
```diff
@@ -34,6 +35,7 @@
 import static java.util.Collections.newSetFromMap;
 
 @RequiredArgsConstructor
+@NoArgsConstructor(force = true)
```
timtebeek: Do we still need this? Looks like the subclass is using a delegate that uses the single argument form already.
```suggestion
```

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinAnnotationTemplateGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+package org.openrewrite.kotlin.internal.template;
+
+import lombok.RequiredArgsConstructor;
+import org.openrewrite.Cursor;
+import org.openrewrite.java.internal.template.AnnotationTemplateGenerator;
+import org.openrewrite.java.tree.JavaSourceFile;
+
+import java.util.Set;
+
+@RequiredArgsConstructor
```
timtebeek: I don't think we'd need this right? If there's no fields?
```suggestion
```

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinAnnotationTemplateGenerator.java
Laurens-W: I didn't realize we could do that, TIL!

knutwannheden: Kotlin and quite a few other languages allow quoting identifiers in this way. Java unfortunately doesn't have anything like this.

knutwannheden (approved):

PR #6525, opened 2026-01-13
Add XML recipe `ChangeTagAttributeKey`

MBoegers:
## What's changed?

Added a recipe to change the key of an XML tags attribute.

## What's your motivation?

enable declarative chnages like:

```diff
- <example name="test" ../>
+ <example title="test" ../>
```

Okay to keep the glob as a filter in the recipe? The other ones don't have this ability and one would have to archive the same with a preconfition. 

Implementing a imperative recipe every time or in spring itself. 
It makes more sense to be central.

## Any additional context

I need this kind of changes for Richface migration in spring

### Checklist

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttributeKey.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttributeKey.java
@@ -42,8 +42,8 @@ public class ChangeTagAttributeKey extends Recipe {
         XPathMatcher xPathMatcher = new XPathMatcher(elementXPath);
         return new XmlIsoVisitor<ExecutionContext>() {
             @Override
-            public Xml.Attribute visitAttribute(Xml.Attribute attribute, ExecutionContext executionContext) {
-                Xml.Attribute a = super.visitAttribute(attribute, executionContext);
+            public Xml.Attribute visitAttribute(Xml.Attribute attribute, ExecutionContext ctx) {
+                Xml.Attribute a = super.visitAttribute(attribute, ctx);
 
                 if (!xPathMatcher.matches(getCursor())) {
                     return a;


rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttributeKey.java
```diff
@@ -0,0 +1,72 @@
...
+            description = "The new name for the attribute.",
+            example = "render")
+    String newAttributeName;
+
+    @Option(displayName = "File matcher",
+            description = "Glob pattern for files to process.",
+            example = "**/*.xhtml",
+            required = false)
+    @Nullable
+    String fileMatcher;
```
timtebeek: This seems like something that should be handled by a precondition instead.
```suggestion
```

timtebeek:
The test failure I've tackled separately in https://github.com/openrewrite/rewrite/commit/4a4663273c8f74018903207a1cfcdaec6fee5feb on main; probably rebase after making the above change.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttributeKey.java
MBoegers: done and added to precondition at the usages of the recipe.

greg-at-moderne:
Not fully sure yet, if this should change anything in the recipe interface, but FWIW the XPath language allows for specifying attributes, not only elements, in the expressions, e.g.
```
/dependencies//dependency//@scope
```

MBoegers:
Using an XPath to match the attribute feels great. It is different from the API of the other attribute-related recipes.

rewrite-xml/src/main/java/org/openrewrite/xml/ChangeTagAttributeKey.java
```diff
@@ -0,0 +1,56 @@
...
+            example = "render")
+    String newAttributeName;
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        XPathMatcher xPathMatcher = new XPathMatcher(elementXPath);
+        return new XmlIsoVisitor<ExecutionContext>() {
+            @Override
+            public Xml.Attribute visitAttribute(Xml.Attribute attribute, ExecutionContext executionContext) {
+                Xml.Attribute a = super.visitAttribute(attribute, executionContext);
```
timtebeek:
```suggestion
            public Xml.Attribute visitAttribute(Xml.Attribute attribute, ExecutionContext ctx) {
                Xml.Attribute a = super.visitAttribute(attribute, ctx);
```

PR #6527, opened 2026-01-13
Update version defined in the gradle.properties referenced by the Spring Dependency Management plugin

bmuschko:
## What's changed?

Resolves a version reference in the `dependencyManagement.imports.mavenBom` introduced by the Spring Dependency Management plugin, and updates the version in the `gradle.properties` file that defines it.

## What's your motivation?

Recipes like ["Migrate to Spring Cloud 2023"](https://docs.openrewrite.org/recipes/java/spring/cloud2023/upgradespringcloud_2023) recipe try to upgrade Spring dependency versions. If the end user decides to use the Spring Dependency Management plugin (legacy) instead of Gradle's platform concept (recommended), the version is not upgraded properly.

No

No

There's no workaround available as of now except for manually making the change.


timtebeek:
@kdvolder FYI; curious to hear what other issues you had.

PR #6529, opened 2026-01-14
Support for directives in Yaml

### LINKED ISSUE #6396: Yaml Parser seems to swallow version declaration directive

```java
%YAML 1.2
---
a:
  - b
```
when added to `YamlParserTest`
fails with:
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "file.yaml":
```

```diff
-%YAML 1.2
 ---
 a:
   - b
```

## OSS repro

- https://github.com/microprofile/microprofile/blob/96302a2943f7c27c171239ce4d27bd136d5138a8/site.yaml#L19C1-L23C60

greg-at-moderne:
## What's changed?

Amending the Yaml support with missing element, the directives, e.g.
```yaml
%YAML 1.2
---
key: value
```

## What's your motivation?

- fixes #6396

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
@@ -160,6 +160,7 @@ public interface Yaml extends Tree {
          * Optional list of directives that precede this document.
          * Directives include %YAML and %TAG.
          */
+        @Nullable
         List<Directive> directives;
 
         boolean explicit;
@@ -177,7 +178,7 @@ public interface Yaml extends Tree {
                     randomId(),
                     prefix,
                     Markers.EMPTY,
-                    directives.stream().map(Directive::copyPaste).collect(toList()),
+                    directives == null ? null : directives.stream().map(Directive::copyPaste).collect(toList()),
                     explicit,
                     block.copyPaste(),
                     end.copyPaste()


rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
```diff
@@ -155,6 +155,13 @@ class Document implements Yaml {
 
         String prefix;
         Markers markers;
+
+        /**
+         * Optional list of directives that precede this document.
+         * Directives include %YAML and %TAG.
+         */
+        List<Directive> directives;
```
knutwannheden, line 164: Note: For deserialization we need to either declare this field as nullable or add an `@JsonCreator` which initializes it if necessary. (A setter called by the deserializer might also be possible.)

greg-at-moderne: Thanks. Made it Nullable in 32666f737.
But for the context - I fail to see similar precautions in some of other `List` fields in `Yaml.java`, e.g. `Yaml.Documents.documents`.

knutwannheden: This is only relevant because it is a new property. So any LST that was ingested before this gets released, will get a null value when deserialized.

PR #6532, opened 2026-01-14
Fix HCL parser for comments with CR-only line endings

timtebeek:
## Summary

- Fix HCL lexer to support CR-only (`\r`) line endings in line comments, in addition to LF and CRLF
- Add comprehensive test coverage for HCL comments at file boundaries (60+ new tests)

### Bug Fixed
The HCL lexer's `LINE_COMMENT` rule only supported LF (`\n`) or CRLF (`\r\n`) line endings. Files with CR-only line endings (classic Mac-style) after comments would fail to parse with errors like:
```
token recognition error at: '# comment\rlocals'
```

### Changes
- **HCLLexer.g4**: Updated `LINE_COMMENT` rule to accept `\r` as a standalone line ending
- **HclCommentTest.java**: Added comprehensive tests for:
  - Comments as the only file content
  - Comments at file start/end (with and without trailing newlines)
  - CR-only, LF, and CRLF line endings
  - Block comments at file boundaries
  - Unicode in comments
  - Realistic Terraform patterns with header/footer comments

### Known Issue (Deferred)
One test is disabled (`emojiInComment`) - emoji (surrogate pairs) in comments at the start of a file causes cursor tracking issues due to a mismatch between ANTLR character indices and the parser visitor's code point tracking. This requires a deeper refactoring and is tracked for future investigation.

## Test plan
- [x] All existing HCL tests pass
- [x] New CR-only line ending tests pass
- [x] New comment boundary tests pass
- [x] Unicode test (`HclParserTest.unicode`) continues to pass


rewrite-hcl/src/main/antlr/HCLLexer.g4
```diff
@@ -58,7 +58,7 @@ Identifier
 // https://github.com/hashicorp/hcl2/blob/master/hcl/hclsyntax/spec.md#comments-and-whitespace
 WS              : [ \t\r\u000C]+                           -> channel(HIDDEN);
 COMMENT         : '/*' .*? '*/'                            -> channel(HIDDEN);
-LINE_COMMENT    : ('//' | '#') ~[\r\n]* '\r'? ('\n' | EOF) -> channel(HIDDEN);
+LINE_COMMENT    : ('//' | '#') ~[\r\n]* ('\r\n' | '\r' | '\n' | EOF) -> channel(HIDDEN);
 NEWLINE         : '\n'                                     -> channel(HIDDEN);
```
mtthwcmpbll: I'm curious why we handle a variety of end of line characters for comments but not broadly as part of the NEWLINE pattern here.  The linked doc on HCL syntax doesn't tie handling newlines only to comments (and it refers to newlines as "Newline sequences (either U+000A or U+000D followed by U+000A)" so we shouldn't actually see a solitary `\r`)

Does it make sense to update the NEWLINE pattern to match the new flexible one `('\r\n' | '\r' | '\n' | EOF)` and then use NEWLINE in our LINE_COMMENT pattern?

PR #6058, opened 2025-09-20
fix: ChangePackage from packageName is part of the submodule dir name issue

qxo:
ie:
```
submodule name: p1-core and ChangePackage from: p1.sub1-> to: p2.sub1

before this pr result ->  p2-core/src/java/sub1/xx
after  this pr result   -> p1-core/src/java/p2/sub1/xx
```

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

PR #6535, opened 2026-01-14
Fix for `ChangePackage` being able to create ambiguous imports for types in use upon reparsing

steve-aom-elliott:
Adding simple test showcasing recipe not failing, but result not being possible to parse again.

## What's changed?
- So far just a failing test

## What's your motivation?
Change package can result in type ambiguity upon reparsing, leading to compilation failures.

### Checklist


rewrite-java-test/src/test/java/org/openrewrite/java/ChangePackageTest.java
```diff
@@ -665,6 +665,119 @@ <T extends Test> T method(T t) {
         );
     }
 
+    @Test
+    void ambiguousImportIssue() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangePackage("origpkg.validation", "newpkg.validation", true)),
```
timtebeek, line 671: Looks like we had a similar PR posted with a potential fix:
- https://github.com/openrewrite/rewrite/pull/6058

steve-aom-elliott: From the looks of that PR, that's just dealing with getting the name right. I think the issue here is not the names themselves but rather that if you change a package name, there may then be two potential sources for the same class, etc. that you are using in your code if you're referencing it in a not fully-qualified type of way.

timtebeek:
Coul you check whether https://github.com/openrewrite/rewrite/pull/6614 might have already fixed this without further changes needed here?
- https://github.com/openrewrite/rewrite/pull/6614

steve-aom-elliott:
@timtebeek It does not fix it, no.

If you comment out the example in the `ambiguousImportIssue()` test, then it still passes, meaning it's still changing
```java
package xyz;

import origpkg.validation.*;
import otherpkg.validation.*;

class A {
  @NotBlank
  private String someField;
  @ExtraneousAnnotation
  private String otherField;
}
```
to
```java
package xyz;

import newpkg.validation.*;
import otherpkg.validation.*;

class A {
  @NotBlank
  private String someField;
  @ExtraneousAnnotation
  private String otherField;
}
```
despite `@NotBlank` existing in both `newpkg.validation` and `otherpkg.validation`, which, if the last example in the test is not commented out, is shown to be not parseable by our JavaParser

steve-aom-elliott:
Closing in favor of #7202 which takes a different approach (ScanningRecipe-based ambiguity detection).

PR #6537, opened 2026-01-15
Add Kotlin DSL support to RemovePluginVisitor

div-bargali:
## What's changed?

Added Kotlin DSL support to `RemovePluginVisitor` to enable plugin removal from both Groovy (`build.gradle`) and Kotlin (`build.gradle.kts`) build files. The visitor now extends `JavaIsoVisitor` instead of `GroovyIsoVisitor` and implements dual-mode matching logic that detects the source language at runtime and applies appropriate matching strategies.

## What's your motivation?

The existing `RemovePluginVisitor` only worked with Groovy build files and failed to process Kotlin DSL files. The root cause was that `MethodMatcher` patterns rely on type attribution, which is not fully available in Kotlin AST nodes, causing matches to fail. This prevented the `RemoveBuildPlugin` recipe from working on Kotlin-based Gradle projects, limiting its utility for projects using the modern Kotlin DSL.

1. **The language detection mechanism**: Using `getCursor().firstEnclosing(K.CompilationUnit.class) != null` to differentiate between Kotlin and Groovy at runtime
2. **The dual-mode matching helpers**: `matchesIdCall()`, `matchesVersionCall()`, and `matchesApplyCall()` that abstract the difference between name-based matching (Kotlin) and type-based matching (Groovy)
3. **Backward compatibility**: Ensuring all existing Groovy test cases continue to pass without modification
4. **The cursor context handling**: How the outermost `J.Block` wrapper in Kotlin files is naturally filtered by the `m != null` check

**Considered Alternatives:**

1. **Using wildcard matchers with relaxed matching** (`"*..* plugins(..)", true`) - Rejected because it could cause false positives and doesn't address the fundamental type attribution issue
2. **Creating separate visitors for Groovy and Kotlin** - Rejected due to massive code duplication; the transformation logic is identical, only matching differs
3. **Extending GroovyIsoVisitor and adding special Kotlin handling** - Rejected because GroovyIsoVisitor doesn't properly traverse Kotlin compilation units

**Chosen Approach Rationale:**

Switching to `JavaIsoVisitor` as the base class is architecturally sound because both Groovy and Kotlin LSTs inherit from the Java LST. This provides a unified traversal mechanism while allowing language-specific matching logic through helper methods.

## Any additional context

**Technical Implementation Details:**

- For Kotlin: Uses simple method name matching (`"plugins".equals(m.getSimpleName())`) since DSL keywords are consistent
- For Groovy: Preserves all original `MethodMatcher` patterns which rely on full type information
- Both languages handle their specific syntax naturally: `G.MapEntry` for Groovy's `apply plugin: "x"` and `J.Assignment` for Kotlin's `apply(plugin = "x")`
- Zero performance impact on Groovy processing; Kotlin path actually faster due to string comparison vs regex matching

**Test Coverage:**

All new Kotlin test cases pass including:
- Plugin removal with/without version
- Unapplied plugins (`apply false`)
- Top-level `apply(plugin = "...")` syntax
- Multiple plugin occurrences
- Empty `plugins{}` block removal

### Checklist

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2429,7 +2429,7 @@ public class GroovyParserVisitor {
             Space space = whitespace();
             J.FieldAccess qualid = TypeTree.build(name()).withPrefix(space);
             JLeftPadded<J.Identifier> alias = null;
-            if (sourceStartsWith("as")) {
+            if (sourceStartsWith("as", "\n", " ")) {
                 alias = padLeft(sourceBefore("as"), new J.Identifier(randomId(), whitespace(), Markers.EMPTY, emptyList(), name(), null, null));
             }
             return maybeSemicolon(new J.Import(randomId(), importPrefix, Markers.EMPTY, statik, qualid, alias));
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/AssertTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/tree/AssertTest.java
@@ -48,4 +48,19 @@ class AssertTest implements RewriteTest {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/6374")
+    @Test
+    void withMessageAndSpaceBeforeColon() {
+        rewriteRun(
+          groovy(
+            """
+              import java.util.Collection
+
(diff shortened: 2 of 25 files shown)


div-bargali:
> Thanks a lot for the help here @div-bargali ! The changes you made were very good already, and I especially like the extensive tests you provided too. Those gave me the confidence to refactor a bit more to add some early returns and extract common logic to reduce indentation a bit. With that we're good to go!
> 
> We'll fold this into a new full release by next Wednesday; possible into a patch release earlier.

Thanks @timtebeek. Glad I could help 😄

PR #6549, opened 2026-01-16
Add `AddExplicitDependencyVersion` visitor to explicitly set or remove dependency versions

Jenson3210:
## What's changed?
Add Visitor to explicitly set or remove dependency versions for dependencies which become (un-)unmanaged during recipes like UpgradeDependencyVersion and UpgradePluginVersion in Gradle. Maven has an equivalent built in when the parent pom changes. 

## What's your motivation?
Sometimes during a Spring Boot migration, versions become managed/unmanaged. The compilation breaks as the dependency no longer receives a version. This PR aims at adding version tags and removing redundant version tags when the platform dependency changes OR the spring boot plugin version is changed. 
We could later similarly add something for dependencies which are managed by the spring dependencies plugin and its corresponding `dependencyManagement.imports`, but that's a different usecase

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
@@ -451,7 +451,7 @@ public class RemoveRedundantDependencyVersions extends Recipe {
                                     return constraint.getStrictVersion();
                                 }
                             }
-                            for (ResolvedDependency dependency : gdc.getDirectResolved()) {
+                            for (ResolvedDependency dependency : gdc.getResolved()) {
                                 if (dependency.getRequested().getVersion() == null && "org.springframework.boot".equals(dependency.getGroupId())) {
                                     return dependency.getVersion();
                                 }


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -508,6 +508,46 @@ implementation platform("com.google.guava:guava:30.1.1-jre")
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite/issues/3227")
```
Jenson3210:
```suggestion
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
```diff
@@ -0,0 +1,169 @@
...
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.maven.tree.ResolvedDependency;
+import org.openrewrite.semver.DependencyMatcher;
+
+import java.util.Optional;
+
+/**
+ * A recipe that adds an explicit version to dependencies that don't have one declared.
```
timtebeek:
```suggestion
 * A visitor that adds an explicit version to dependencies that don't have one declared.
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -168,11 +175,9 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
 
                             MavenPomDownloader mpd = new MavenPomDownloader(ctx);
                             GradleMultiDependency.matcher()
-                                    .groupId(groupPattern)
-                                    .artifactId(artifactPattern)
```
timtebeek, line 172: What's the context behind dropping these constraints? What are the implications?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
timtebeek, line 1: I don't immediately see the link between the newly added `AddExplicitDependencyVersion` and the changes in this file; are these unrelated cleanups or improvements?

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
```diff
@@ -255,9 +259,10 @@ public boolean isAcceptable(SourceFile sf, ExecutionContext ctx) {
             }
 
             @Override
-            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+            public @Nullable Tree preVisit(@Nullable Tree tree, ExecutionContext ctx) {
                 Tree t = tree;
                 if (t instanceof SourceFile) {
+                    stopAfterPreVisit();
```
timtebeek: Why is the `stopAfterPreVisit` only inside the conditional here and not above? There doesn't appear to be any actions we take for any non-source files in the same visitor, so then I'd suspect we can stop traversal for other code paths too right?

rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/UpgradePluginVersionTest.java
```diff
@@ -297,4 +297,83 @@ void upgradePluginVersionInBuildGradleNotProperties() {
...
+                  id 'java'
+                  id 'org.springframework.boot' version '2.7.0'
+              }
+
+              repositories {
+                  mavenCentral()
+              }
+
+              dependencies {
+                  implementation 'javax.servlet:javax.servlet-api:4.0.1'
```
timtebeek, line 317: Should we similar to below also feature this dependency here and assert the version number is not dropped? Because for javax.servlet-api it would (not) have been added.
```suggestion
                  implementation 'javax.servlet:javax.servlet-api:4.0.1'
                  implementation 'org.apache.activemq:activemq-client-jakarta:5.18.2'
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddExplicitDependencyVersionTest.java
```diff
@@ -0,0 +1,662 @@
...
+                  implementation 'com.fasterxml.jackson.core:jackson-databind:2.17.2'
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void noChangeWhenDependencyNotResolved() {
+        // When the dependency can't be resolved (no version and no BOM management), no change should be made
```
timtebeek, line 196: Right now we silently drop errors; would we want to add a marker perhaps about the failure seen? Or not yet you think?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
Jenson3210: improvement to version detection when plugin is active. 
I discovered by running a test with spring boot version but no spring boot dependencies, spring boot version was not detectable. The plugin version exists on the marker though -> added `getSpringBootVersionFromPlugin()` as I figured this will be the most common usage. Obviously, the `platform/enforcedplatform` will also be used, but most typically people will have the plugin enabled. 

Coming to the link to this file: 
Both `UpdateDependencyVersion` and `UpdatePluginVersion` calculate:
- newly managed versions -> delegate the removal of version which are now managed to this one. 
- no longer managed versions -> delegate the addition of version to the newly added one.

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
Jenson3210: This block in the visitor does not modify tree elements. 
It is gathering platforms/enforcedPlatforms Poms active in the build gradle file and fills the `platforms` Map. This Map later gets used for detecting if a dependency is managed. 

When you provide this recipe with okhttp parameters (which is managed by `org.spring...` platform), the platforms map remains empty and the cleanup does not happen as the filter of okhttp group/artifact is filtering out the `spring-boot-dependencies` platform dependency. 
The filtering still happens at line 395, but we need all platforms in order to succesfully check in that editing visitor if we should remove the declared version or not as otherwise platforms will be empty.

rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeDependencyVersion.java
Jenson3210: sure!

Jenson3210:
```suggestion
                stopAfterPreVisit();
                Tree t = tree;
                if (t instanceof SourceFile) {
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/AddExplicitDependencyVersionTest.java
Jenson3210: Added an info Marker. 👍🏼

rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/UpgradePluginVersionTest.java
Jenson3210: Added this one. I never hurts to verify both 😄

Jenson3210:
Thanks for the review here @timtebeek! 
Added comments on the questions and resolved the ones I felt certain I covered.

timtebeek:
Thanks a lot! Since these are quite common cases, and we use `UpgradeDependencyVersion` a lot let me just tag Shannon for additional review in case I had missed something.

Jenson3210:
I also wanted to raise in this PR that I doubted in the beginning to add the logic to UpdatePluginVersion as that one suddenly got a "hint" of Spring boot (dependency management) specific code. 
So if we would prefer moving this to rewrite-spring, that could work. However, then a lot of the existing spring recipes would probably need to call that one iso the existing one (and also existing consumers would need to change)

timtebeek:
I think it's common enough to warrant having that here, as otherwise we'd need the same in modules like rewrite-hibernate depending on whether that's called through a Spring composite or not.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
```diff
@@ -0,0 +1,168 @@
...
+                J.Assignment assignment = (J.Assignment) arg;
+                if (assignment.getVariable() instanceof J.Identifier &&
+                        "version".equals(((J.Identifier) assignment.getVariable()).getSimpleName())) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
```
timtebeek: The GradleDependency trait also comes with a `getDeclaredVersion()` which seems similar to the logic used here:
https://github.com/openrewrite/rewrite/blob/c4c7a8d7821097fe8a91e4f32b98b61d472c9ed8/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java#L391-L457

Does it make sense to try to use that instead of duplicating part of that logic here? Or was it intentional to only look at part of the structure here?

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        // Use the GradleDependency trait to match dependencies
+        return new GradleDependency.Matcher()
+                .matcher(dependencyMatcher)
+                .get(getCursor())
+                .map(gradleDep -> {
+                    // Skip if version is already declared
+                    String declaredVersion = gradleDep.getResolvedDependency().getVersion();
+                    if (!StringUtils.isBlank(declaredVersion) && hasExplicitVersion(gradleDep)) {
```
timtebeek: See below as well; the tests we have still pass if we swap in
```suggestion
                    if (!StringUtils.isBlank(declaredVersion) && gradleDep.getDeclaredVersion() != null) {
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
timtebeek:
```suggestion
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
```diff
@@ -0,0 +1,131 @@
...
+                .matcher(dependencyMatcher)
+                .get(getCursor())
+                .map(gradleDep -> {
+                    // Skip if version is already declared
+                    String declaredVersion = gradleDep.getResolvedDependency().getVersion();
+                    if (!StringUtils.isBlank(declaredVersion) && gradleDep.getDeclaredVersion() != null) {
+                        return m;
+                    }
+
+                    // Look up the resolved version from the GradleProject marker
```
shanman190, line 92: `GradleDependency#getResolvedDependency` should already have the version applied on it, so that you can simplify and not have to search through the configurations again. When it's constructed, it should be pulling the specific `ResolvedDependency` instance from the current configuration name as well that's being visited (in the rare case there are different versions in different configurations).

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/UpgradePluginVersion.java
```diff
@@ -228,6 +229,62 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
...
+            private GradleProject gradleProject;
+
+            @Override
+            public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (tree instanceof SourceFile) {
+                    gradleProject = tree.getMarkers().findFirst(GradleProject.class).orElse(null);
+                    if (gradleProject != null) {
+                        if (acc.pluginIdToNewVersion.containsKey("org.springframework.boot") &&
+                                gradleProject.getPlugins().stream().anyMatch(plugin -> "io.spring.dependency-management".equals(plugin.getId())) &&
+                                gradleProject.getPlugins().stream().anyMatch(plugin -> "org.springframework.boot".equals(plugin.getId()))) {
```
shanman190, line 242: Not strictly necessary, but we could use the `GradlePlugin` trait which would allow you to test and get the version directly if it's discoverable.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return super.visit(tree, ctx);
+    }
+
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+        J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+
+        // Can't do anything without GradleProject
+        if (gp == null) {
```
shanman190, line 75: Would it make more sense to use a `stopAfterPreVisit()` instead to save us the traversals?

Jenson3210: This is not a preVisit method. You suggest that I try to move everything to a preVisit block?

Jenson3210:
Thanks @shanman190! 
I have updated the code to reflect 2 comments but still have 1 comment/question open.

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddExplicitDependencyVersion.java
shanman190: So more of what I was meaning is that we could do one of the following:
1. In the existing `visit(Tree, P)`, if we don't have a `GradleProject` marker, then we can just skip visiting the subtree entirely by not executing `super.visit(Tree, P)`.
2. If we change `visit(Tree, ExecutionContext)` to a `preVisit(Tree, ExecutionContext)`, then we could use the `stopAfterPreVisit()` helper method to also avoid visiting the subtree entirely.

The `visitMethodInvocation` logic would be left as is. In this way, we could avoid visiting the subtree when the `GradleProject` marker isn't present.

Jenson3210:
@shanman190 do you see anything else here? Or are we good to go?

rewrite-gradle/src/main/java/org/openrewrite/gradle/RemoveRedundantDependencyVersions.java
```diff
@@ -426,15 +431,27 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
-                            for (GradleDependencyConstraint constraint : testRuntimeConfiguration.getConstraints()) {
+                        GradleDependencyConfiguration gdc = gp.getConfiguration(configuration);
+                        if (gdc != null) {
+                            for (GradleDependencyConstraint constraint : gdc.getConstraints()) {
                                 if ("org.springframework.boot".equals(constraint.getGroupId()) && constraint.getStrictVersion() != null) {
                                     return constraint.getStrictVersion();
                                 }
                             }
-                            for (ResolvedDependency dependency : testRuntimeConfiguration.getDirectResolved()) {
+                            for (ResolvedDependency dependency : gdc.getDirectResolved()) {
```
shanman190: Do we have any concerns about Spring Boot being a transitive dependency of an internally managed wrapper suite?

PR #6554, opened 2026-01-16
Fix dependency management logic resolution order

shanman190:
## What's changed?
* Maven does not merge managed dependency attributes. It replaces them entirely.
* When a managed dependency doesn't have a version, it short circuits and effectively removes the managed dependency rules.
* Given we resolve managed versions starting from the closest, then going to the furthest entries, we should retain the first encountered entry. If that entry comes from a BOM, then the nearest parent managed dependency can override it.

## What's your motivation?
Encountered a diamond problem which illustrated an issue with managed dependency handling.

N/A

@timtebeek @sambsnyd 

N/A

## Any additional context
I did clean up one extra test to be a bit more idiomatic with OpenRewrite standards.

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
@@ -4941,11 +4941,11 @@ class MavenParserTest implements RewriteTest {
                   assertThat(maybeMrr).isPresent();
 
                   MavenResolutionResult mrr = maybeMrr.get();
-                  assertThat(mrr.getDependencies().get(Scope.Test)).haveExactly(1, new Condition<>(dep ->
-                    "org.bouncycastle".equals(dep.getGroupId()) &&
-                      "bcprov-jdk18on".equals(dep.getArtifactId()) &&
-                      "1.79".equals(dep.getVersion()),
-                    "expecting `org.bouncycastle:bcprov-jdk18on:1.79`"));
+                  assertThat(mrr.getDependencies().get(Scope.Test))
+                    .filteredOn(dep -> "org.bouncycastle".equals(dep.getGroupId()) && "bcprov-jdk18on".equals(dep.getArtifactId()))
+                    .singleElement()
+                    .extracting(ResolvedDependency::getVersion)
+                    .isEqualTo("1.79");
               })
             )
           )


rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -4831,4 +4832,118 @@ void bomsShouldNotAppearInRuntimeDependencies() {
...
+                  assertThat(pom).isNotNull();
+                  Optional<MavenResolutionResult> maybeMrr = pom.getMarkers().findFirst(MavenResolutionResult.class);
+                  assertThat(maybeMrr).isPresent();
+
+                  MavenResolutionResult mrr = maybeMrr.get();
+                  assertThat(mrr.getDependencies().get(Scope.Test)).haveExactly(1, new Condition<>(dep ->
+                    "org.bouncycastle".equals(dep.getGroupId()) &&
+                      "bcprov-jdk18on".equals(dep.getArtifactId()) &&
+                      "1.79".equals(dep.getVersion()),
+                    "expecting `org.bouncycastle:bcprov-jdk18on:1.79`"));
```
timtebeek: The way the assertion is set up currently we could still allow multiple different versions; with this pattern we make that  not possible either.
```suggestion
                  assertThat(mrr.getDependencies().get(Scope.Test))
                    // Should only contain bcprov-jdk18on once, with exact version from grandparent
                    .filteredOn(dep -> "org.bouncycastle".equals(dep.getGroupId()))
                    .filteredOn(dep -> "bcprov-jdk18on".equals(dep.getArtifactId()))
                    .singleElement()
                    .extracting(ResolvedDependency::getVersion)
                    .isEqualTo("1.79");
```

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
```diff
@@ -4831,4 +4832,118 @@ void bomsShouldNotAppearInRuntimeDependencies() {
...
+                    <packaging>pom</packaging>
+                    <properties>
+                        <bouncycastle.version>1.68</bouncycastle.version>
+                    </properties>
+                    <dependencyManagement>
+                        <dependencies>
+                            <dependency>
+                                <groupId>org.bouncycastle</groupId>
+                                <artifactId>bcprov-jdk18on</artifactId>
+                            </dependency>
```
timtebeek, line 4889: Quite a confusing setup here ; to be clear the property shouldn't do anything here right? Perhaps a small comment above the property and managed dependency could help explain what we're expecting this to (not) do.

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
shanman190: So that's why I'm using `haveExactly(1, ...)` in particular, so that there can only be one occurrence of the dependency in that list. When there are 0 or multiple, the assertion fails as expected.

timtebeek: I'd read that `haveExactly(1, new Condition` as there may only be one dependency of that particular `GAV`, but there may be other vrsions.

Whereas with what I'm proposing there may only be one dependency of a `GA`, and the version of that single one should be the `V` asserted separately.

Not sure if that part was clear from my earlier message, but figured make sure we're on the same page.

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
shanman190: I've added a couple of comments with respect to this since it is quite confusing. Let me know if it's not clear enough or if you were thinking something else.

rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
shanman190: I see what you're meaning now. I'll update accordingly.

timtebeek (approved):
Thanks a lot for the research, fix and adjustments!

PR #6557, opened 2026-01-16
Reduce scope of recipe resolvers

shanman190:
## What's changed?
* Make recipe resolvers method parameters for `RecipeListing#describe` and `RecipeListing#prepare`.
* `RecipeBundleResolver` and `RecipeBundleReader` now implement `Closeable`, so that they can release resources when no longer needed.

## What's your motivation?
* Based upon learnings, having recipe resolvers set as optional fields on the marketplace proved to be slightly awkward.
* `RecipeBundleResolver` instances need to be closable so that resources can be released

N/A

@jkschneider or delegate

N/A

## Any additional context
N/A

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeBundleResolver.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeBundleResolver.java
@@ -15,15 +15,12 @@
  */
 package org.openrewrite.marketplace;
 
-import java.io.Closeable;
-import java.io.IOException;
-
-public interface RecipeBundleResolver extends Closeable {
+public interface RecipeBundleResolver extends AutoCloseable {
     String getEcosystem();
 
     RecipeBundleReader resolve(RecipeBundle bundle);
 
-    default void close() throws IOException {
+    default void close() throws Exception {
         // no-op
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
@@ -66,7 +66,7 @@ public class RecipeListing implements Comparable<RecipeListing> {
     @With(AccessLevel.PACKAGE)
     private final RecipeBundle bundle;
 
-    private RecipeBundleReader resolve(List<RecipeBundleResolver> resolvers) {
+    private RecipeBundleReader resolve(Collection<RecipeBundleResolver> resolvers) {
         for (RecipeBundleResolver resolver : resolvers) {
             if (resolver.getEcosystem().equals(bundle.getPackageEcosystem())) {
                 return resolver.resolve(bundle);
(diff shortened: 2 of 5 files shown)


rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeBundleResolver.java
```diff
@@ -15,8 +15,15 @@
  */
 package org.openrewrite.marketplace;
 
-public interface RecipeBundleResolver {
+import java.io.Closeable;
+import java.io.IOException;
+
+public interface RecipeBundleResolver extends Closeable {
```
jkschneider: I don't believe there's any advantage anymore to using `Closeable` instead of `AutoCloseable`.

rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
```diff
@@ -66,23 +66,21 @@ public class RecipeListing implements Comparable<RecipeListing> {
...
+                return resolver.resolve(bundle);
             }
         }
-        throw new IllegalStateException("This listing has not been configured with a resolver.");
+        throw new IllegalStateException(String.format("No available resolver for '%s' ecosystem", bundle.getPackageEcosystem()));
     }
 
-    public RecipeDescriptor describe() {
-        return resolve().describe(this);
+    public RecipeDescriptor describe(List<RecipeBundleResolver> resolvers) {
```
jkschneider: Widen to Collection or Iterable I think

rewrite-core/src/test/java/org/openrewrite/rpc/RewriteRpcTest.java
```diff
@@ -64,16 +63,15 @@ void before() throws IOException {
...
+        marketplace = env.toMarketplace(runtimeClasspath());
 
         JsonMessageFormatter clientFormatter = new JsonMessageFormatter(new ParameterNamesModule());
         JsonMessageFormatter serverFormatter = new JsonMessageFormatter(new ParameterNamesModule());
 
         client = new RewriteRpc(new JsonRpc(new HeaderDelimitedMessageHandler(clientFormatter, clientIn, clientOut)), marketplace)
           .batchSize(1);
 
-        server = new RewriteRpc(new JsonRpc(new HeaderDelimitedMessageHandler(serverFormatter, serverIn, serverOut)), marketplace)
+        server = new RewriteRpc(new JsonRpc(new HeaderDelimitedMessageHandler(serverFormatter, serverIn, serverOut)), marketplace, singletonList(new TestRecipeBundleResolver()))
```
jkschneider: Can use List.of in tests.

rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
shanman190: I've widened this to `Collection`, so if we ever need to use some of the helpful methods at that level we can.

rewrite-core/src/test/java/org/openrewrite/rpc/RewriteRpcTest.java
shanman190: Force of habit. 😆

PR #6559, opened 2026-01-17
Use intersection types for RightPadded/LeftPadded in JavaScript AST

jkschneider:
## Summary

- Changes `RightPadded<T extends J>` and `LeftPadded<T extends J>` from wrapper objects with an `.element` property to intersection types where the padded value IS the element with padding mixed in
- Aligns the TypeScript model more closely with the Java model and simplifies property access
- Adds function overloads for `visitLeftPadded`/`visitRightPadded` with improved type safety (primitives never return undefined)
- Fixes various issues discovered during the migration:
  - Fixed `unwrap()` in comparator.ts for intersection types
  - Added guards for missing padding in formatting methods
  - Fixed `WhitespaceReconciler` to use ID comparison for `stopAfter` functionality
  - Added `hasSameKind` override in semantic comparator for void/undefined equivalence


knutwannheden:
I think there is still a problem in the JavaSender and JavaReceiver on the JS side. When sending a `J.LeftPadded` or `J.RightPadded` we basically need to "emulate" that by sending an element with the corresponding kind. Similarly, in the receiver we need to "unwrap" and "invert" the padding and its element.

PR #6560, opened 2026-01-17
De-duplicate declarative recipes

sambsnyd:
Let's stop spending time executing recipes that shouldn't do anything anyway - because a copy of them earlier in the recipe list already did the work

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
@@ -420,8 +420,8 @@ public class DeclarativeRecipe extends ScanningRecipe<DeclarativeRecipe.Accumula
             this.source,
             this.causesAnotherCycle,
             this.maintainers,
-            new ArrayList<>(deduplicatedChildren),
-            new ArrayList<>(this.preconditions),
+            deduplicatedChildren,
+            this.preconditions,
             this.validation,
             this.initValidation
         );


rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
```diff
@@ -348,12 +377,56 @@ public final List<Recipe> getRecipeList() {
...
+            this.name,
+            this.displayName,
+            this.description,
+            this.tags,
+            this.estimatedEffortPerOccurrence,
+            this.source,
+            this.causesAnotherCycle,
+            this.maintainers,
+            new ArrayList<>(deduplicatedChildren),
+            new ArrayList<>(this.preconditions),
```
timtebeek: Do we need the `new ArrayList<>` calls here? For `deduplicatedChildren` we already know it's a new instance coming from `ListUtils.map`, which itself also creates a new instance.

sambsnyd: Correct this was a leftover

jkschneider (changes requested):
Could we please wait until at least Monday. I'm not saying it's wrong but would like to carefully consider. Very much worry about this.

jkschneider:
One of my initial concerns: it feels pretty wrong to apply this uniquely to DeclarativeRecipe. Like using `buildRecipe` with a language that has less verbose syntax, I don't think YAML even has much attraction. And yet the same consideration should apply in those situations. Feels like the wrong place to solve this.

sambsnyd:
We will do this instead: https://github.com/openrewrite/rewrite/pull/6619

PR #6561, opened 2026-01-17
Add recipes for Kotlinx coroutines and serialization, based on `ReplaceWith`

timtebeek:
Follows the pattern we already have for `InlineMethodCalls`, but now for Kotlin.

- Fixes https://github.com/openrewrite/rewrite/issues/5929
- Closes https://github.com/openrewrite/rewrite-third-party/pull/57


rewrite-kotlin/src/test/java/org/openrewrite/kotlin/replace/ReplaceKotlinMethodTest.java
```diff
@@ -140,6 +141,30 @@ fun test(a: Char, b: Char): Int {
         );
     }
 
+    @Test
+    void replaceConstructorCall() {
+        rewriteRun(
+          spec -> spec.recipe(new ReplaceKotlinMethod(
+            "Holder <constructor>(kotlin.Int)",
```
knutwannheden: Nit: We now prefer `<init>` over `<constructor>` in `MethodMatcher`.
```suggestion
            "Holder <init>(kotlin.Int)",
```

PR #6566, opened 2026-01-19
Preventing the addition of a version tag on a direct dependency for scenarios where the local parent had the managed dependency that we were also changing. Also prevented property creation / modification that could cause breakages of other dependencies.

steve-aom-elliott:
## What's changed?
- When we're calling `ChangeDependencyGroupIdAndArtifactId`, where we'd also be updating the managed dependency, and the managed dependency lives in a parent POM (that is within the repo), it will now be prevented from also adding the version tag to the direct dependency because it should be getting inherited from the managed dependency in the parent.
- When we're calling `ChangeDependencyGroupIdAndArtifactId` or `ChangeManagedDependencyGroupIdAndArtifactId`, it will now prevent the creation or changing of a local POM property when changing a tag value for `groupId`, `artifactId` or `version`.

## What's your motivation?
- When you had the managed dependency outside of the current POM but were going to change it as well, because the operations ran using `doAfterVisit(..)`, the updated POM for the parent would not be visible at the time of deciding whether to add a `version` tag to your direct dependency.
- When changing the dependency's GAV, namely the `groupId` and `artifactId`, you're no longer truly representing the same dependency anymore, and so introducing a local property to override an inherited property or changing an existing local property could cause it to break other inherited managed dependencies or direct dependencies also relying on inherited properties.

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -175,19 +175,13 @@ public class ChangeDependencyGroupIdAndArtifactId extends Recipe {
                 if (isOldDependencyTag || isPluginDependency || isAnnotationProcessorPath) {
                     String groupId = newGroupId;
                     if (groupId != null) {
-                        Optional<Xml.Tag> groupIdTag = t.getChild("groupId");
-                        if (groupIdTag.isPresent()) {
-                            t = (Xml.Tag) new ChangeTagValueVisitor<>(groupIdTag.get(), groupId).visitNonNull(t, ctx);
-                        }
+                        t = (Xml.Tag) new ChangeTagValueVisitor<>(t.getChild("groupId").orElse(null), groupId).visitNonNull(t, ctx);
                     } else {
                         groupId = t.getChildValue("groupId").orElseThrow(NoSuchElementException::new);
                     }
                     String artifactId = newArtifactId;
                     if (artifactId != null) {
-                        Optional<Xml.Tag> artifactIdTag = t.getChild("artifactId");
-                        if (artifactIdTag.isPresent()) {
-                            t = (Xml.Tag) new ChangeTagValueVisitor<>(artifactIdTag.get(), artifactId).visitNonNull(t, ctx);
-                        }
+                        t = (Xml.Tag) new ChangeTagValueVisitor<>(t.getChild("artifactId").orElse(null), artifactId).visitNonNull(t, ctx);
                     } else {
                         artifactId = t.getChildValue("artifactId").orElseThrow(NoSuchElementException::new);
                     }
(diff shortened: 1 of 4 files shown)


steve-aom-elliott:
With regards to property changes for existing properties (which isn't covered currently), perhaps it makes sense to effectively say that all bets are off in terms of updating existing properties when you use `ChangeDependencyGroupIdAndArtifactId` or `ChangeManagedDependencyGroupIdAndArtifactId` specifically, as there are too many situations to contend with, and in those cases, opt for not modifying the properties but rather use the actual values on the changed dependency rather than placeholders.

shanman190:
I feel like trying to reuse the property definitely has risks for doing harm. Even trying to match all of the currently used GAs to the target GA doesn't always pan out for cases like `jackson-bom`.

steve-aom-elliott:
@shanman190 This last push is quick pass of making it so that if you ask for properties to not be changed (such as when we're running `ChangeDependencyGroupIdAndArtifactId` or `ChangeManagedDependencyGroupIdAndArtifactId`) that it also does not change the existing local POM properties.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -292,15 +292,23 @@ protected boolean isTag(String name) {
...
+
+    protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullable String newValue, P p, boolean doNotAlterAnyProperties) {
+        return changeChildTagValue(tag, childTagName, newValue, false, p, doNotAlterAnyProperties);
     }
 
     protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullable String newValue, @Nullable Boolean addPropertyIfMissing, P p) {
+        return changeChildTagValue(tag, childTagName, newValue, addPropertyIfMissing, p, false);
+    }
+
+    protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullable String newValue, @Nullable Boolean addPropertyIfMissing, P p, boolean doNotAlterAnyProperties) {
```
sambsnyd: I don't like this `doNotAlterAnyProperties` flag. Perhaps the recipe should use the change tag visitor directly rather than this helper method if the behavior of this helper method is not desired.

steve-aom-elliott: @sambsnyd In the latest push, I dropped the flag an opted for a separate helper `changeChildTagValueInlineOnly(..)` that takes will only do a check for whether the child tag (if present) is the same value as requested and if not, runs the `ChangeTagValueVisitor` to change the value instead.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -291,6 +291,17 @@ protected boolean isTag(String name) {
...
+    protected Xml.Tag changeChildTagValueInlineOnly(Xml.Tag tag, String childTagName, @Nullable String newValue, P p) {
+        Optional<Xml.Tag> childTag = tag.getChild(childTagName);
+        if (childTag.isPresent()) {
+            String oldValue = childTag.get().getValue().orElse(null);
+            if (newValue != null && !newValue.equals(oldValue)) {
+                tag = (Xml.Tag) new ChangeTagValueVisitor<>(childTag.get(), newValue).visitNonNull(tag, p);
+            }
+        }
+        return tag;
+    }
```
sambsnyd: I don't love adding API surface area to MavenVisitor for this. Almost always a maven recipe author will want the property update. Without the property updates it's just an invocation of `ChangeTagValueVisitor` which can be done in a single line.

steve-aom-elliott: @sambsnyd Latest push drops the method in `MavenVisitor`. I need to instead check if the child tag exists first in the places I was calling this (at least for `groupId` and `artifactId`, as they weren't guarded in all places externally).

Issue #2504, opened 2022-12-08
Do not reprocess dependencyManagement imports found in parents


nmck257:
test case:
```java
    @Test
    void resolveDependencyManagementImportsOnlyOnce() {
        rewriteRun(
          pomXml("""
                <?xml version="1.0" encoding="UTF-8"?>
                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
                  <modelVersion>4.0.0</modelVersion>
                
                  <parent>
                    <groupId>org.springframework.boot</groupId>
                    <artifactId>spring-boot-starter-parent</artifactId>
                    <version>2.7.5</version>
                  </parent>
                
                  <groupId>org.sample</groupId>
                  <artifactId>sample</artifactId>
                  
                  <properties>
                    <mockito.version>3.9.0</mockito.version> <!-- mockito-bom did not exist prior to 4.x -->
                  </properties>
                  
                  <dependencyManagement>
                    <dependencies>
                      <dependency>
                        <groupId>org.mockito</groupId>
                        <artifactId>mockito-bom</artifactId>
                        <version>4.6.1</version>
                      </dependency>
                    </dependencies>
                  </dependencyManagement>
                </project>
                """)
        );
    }
```

This test fails during parsing with:

> org.openrewrite.maven.UncheckedMavenDownloadingException: Failed to download dependencies for pom.xml:
> [...]
> Caused by: org.openrewrite.maven.MavenDownloadingExceptions
>	at org.openrewrite.maven.MavenDownloadingExceptions.append(MavenDownloadingExceptions.java:44)
>	at org.openrewrite.maven.UncheckedMavenDownloadingException.<init>(UncheckedMavenDownloadingException.java:32)
>	... 75 more
>	Suppressed: org.openrewrite.maven.MavenDownloadingException: Unable to download POM. Tried repositories:
> [...]

And the `MavenDownloadingException`'s `failedAt` is `org.mockito:mockito-bom:3.9.0`.

For context, `spring-boot-dependencies` 2.7.5 contains this:
```xml
<dependencyManagement>
  <dependencies>
  ...
      <dependency>
        <groupId>org.mockito</groupId>
        <artifactId>mockito-bom</artifactId>
        <version>${mockito.version}</version>
[...]

PR #6569, opened 2026-01-19
Resolve Maven's `dependencyManagement` imports only once

timtebeek:
- Fixes https://github.com/openrewrite/rewrite/issues/2504

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/MavenParserTest.java
@@ -4950,37 +4950,4 @@ class MavenParserTest implements RewriteTest {
           )
         );
     }
-
-    @Issue("https://github.com/openrewrite/rewrite/issues/2504")
-    @Test
-    void resolveDependencyManagementImportsOnlyOnce() {
-        rewriteRun(
-          pomXml(
-            """
-              <project>
-                <modelVersion>4.0.0</modelVersion>
-                <parent>
-                  <groupId>org.springframework.boot</groupId>
-                  <artifactId>spring-boot-starter-parent</artifactId>
-                  <version>2.7.5</version>
-                </parent>
-                <groupId>org.sample</groupId>
-                <artifactId>sample</artifactId>
-                <properties>
-                  <mockito.version>3.9.0</mockito.version> <!-- mockito-bom did not exist prior to 4.x -->
-                </properties>
-                <dependencyManagement>
-                  <dependencies>
-                    <dependency>
-                      <groupId>org.mockito</groupId>
-                      <artifactId>mockito-bom</artifactId>
-                      <version>4.6.1</version>
-                    </dependency>
-                  </dependencies>
-                </dependencyManagement>
-              </project>
-              """
-          )
-        );
-    }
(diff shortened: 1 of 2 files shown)


shanman190:
So I don't think that we can do this. The reason here is that parents have higher precedence than BOMs. So if you were to use `mockito.version=4.3.0` with Maven, as an example, the managed dependencies found in the effective POM are 4.3.0 versions. With this change here, we'd use 4.6.1 erroneously.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -904,14 +907,18 @@ private GroupArtifactClassifierType createDependencyManagementKey(Defined define
...
+                Set<GroupArtifact> importedBoms) throws MavenDownloadingException {
             Pom pom = pomAncestry.get(0);
             for (ManagedDependency d : incomingDependencyManagement) {
                 if (d instanceof Imported) {
                     GroupArtifactVersion groupArtifactVersion = getValues(((Imported) d).getGav());
-                    if (isAlreadyResolved(groupArtifactVersion, pomAncestry)) {
+                    GroupArtifact bomGa = new GroupArtifact(groupArtifactVersion.getGroupId(), groupArtifactVersion.getArtifactId());
+                    // Skip if this BOM (by groupId:artifactId) was already imported or declared - first definition wins
+                    if (importedBoms.contains(bomGa) || isAlreadyResolved(groupArtifactVersion, pomAncestry)) {
                         continue;
```
sambsnyd, line 919: Maven's configuration model tends to start with the parent pom, then layer on child poms from there. This naturally adds up to "last wins" for most things.
We did not fully understand all the nuances when we implemented our Maven emulation originally, so we don't always do things in the same order, leading to edge cases like this popping up.

From maven's perspective the dependency management in the current pom is applied after the dependency management from the parent and last wins. But since we do some stuff a bit backwards it looks like first wins.

I bet if you had multiple dependency management definitions at the same level the semantics would come out looking more last-wins.

Anyway...fixing the order we parse things to more closely match Maven is a bigger refactoring than this fix. So if you have validated that our output matches `mvn dependency:tree` for the provided test case that could be good enough for now.

sambsnyd (commented):
I'm a bit concerned that this might inadvertently take us further from maven's model rather than closer to it, see comment in the sources. But if you've verified that our model exactly matches `mvn dependency:tree` output for the test case maybe that's okay for now.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
nmck257: hmm -- I ran real-maven against the sample pom in this test case, and it aligns with what Shannon said earlier. Maven actually *does* similarly fail looking for the non-existent `org.mockito:mockito-bom:3.9.0`.

so, I think the test case I started with in #2504 must not have faithfully matched whatever I was seeing at the time. I think the best choice is to close the issue unless and until I encounter this mythical edge case again in the wild and can scrape out more specifics

thanks, all, for all the effort on this

PR #6556, opened 2026-01-16
Support varargs in JavaTemplate.Matcher

timtebeek:
- Fixes https://github.com/openrewrite/rewrite/issues/4397

Ran up against this limitation again when working on
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/889

Figured by adding support we can expand once again what we can cover with Refaster.

PR #6572, opened 2026-01-20
Fix JavaTemplate replacements passing a `J.NewArray` into a vararg `J.MethodInvocation`

timtebeek:
- Follows on from #6556

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/marketplace/RecipeMarketplaceReaderTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/marketplace/RecipeMarketplaceReaderTest.java
@@ -288,6 +288,82 @@ class RecipeMarketplaceReaderTest {
         assertThat(authorIndex).isLessThan(zebraIndex);
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateSubstitutionsTest.java
```diff
@@ -551,4 +551,50 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
...
+                      return Stream.of("a", "b", "c").toList();
+                  }
+              }
+              """,
+            """
+              import java.util.List;
+              
+              class Example {
+                  List<String> test() {
+                      return java.util.Arrays.asList("a", "b", "c");
```
timtebeek, line 594: Previously we were seeing `asList({"a", "b", "c"});` as a result from a choice in #6556 to reuse `J.NewArray` to represent the varargs elements. Still agree with that choice, but it does mean we need special handling in substitutions below.

PR #6576, opened 2026-01-20
Remove unnecessary dependencies based on declared BOM

bmuschko:
## What's changed?

Adds a new `RemoveBomManagedDirectDependencies` recipe for Maven and Gradle.

## What's your motivation?
Removes directly declared dependencies when they have an incompatible major version compared to the version managed by an imported BOM. This is useful during framework upgrades (e.g., Spring Boot) where transitive dependencies receive major version bumps and explicitly declared older versions should be removed to use the BOM-managed versions instead.

No

No

Manual removal of dependencies.


timtebeek:
Doesn't this duplicate the `RemoveRedundantDependencyVersions` recipes we already have? I thought those intend to do the same mostly/partially.

bmuschko:
> Doesn't this duplicate the RemoveRedundantDependencyVersions recipes we already have? I thought those intend to do the same mostly/partially.

The recipe functionality is slightly different. I'd say the two recipes are complimentary.

- `RemoveRedundantDependencyVersions`: Removes the version from a dependency declaration but keeps the dependency. Applies when versions are compatible/matching with the BOM.
- `RemoveBomManagedDirectDependencies`: Removes the entire dependency declaration. Applies when versions are incompatible (different major version) with the BOM. The dependency is deleted entirely, letting the BOM provide it transitively.

Maybe we can work that functionality into the existing `RemoveRedundantDependencyVersions` but for now I wanted to keep the concerns separated. Happy to discuss this more.

timtebeek:
Thanks for the explanation; I've exploring the problem space a bit still. For one I know we also have some functionality similar to `RemoveRedundantDependencyVersions` in `ChangeParentPom`, [as seen in this test](https://github.com/openrewrite/rewrite/blob/d735c2411e9c96aa51ff7cbfebee0fba1871aacf/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java#L1036-L1100). I do understand now that the scope for this recipe is wider: aiming to fully remove what I'd called "pinned" dependencies.

For the Maven case I do wonder if the use of `ChangeParentPom` as seen in larger composite Spring Boot upgrades would already lead us to [the test case tested here](https://github.com/openrewrite/rewrite/blob/86d235ac41480329485e668c4b35765bcf5007f7/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveBomManagedDirectDependenciesTest.java#L201-L232), where we keep dependencies that no longer carry an overridden version. In that case I don't see the inclusion of `RemoveBomManagedDirectDependencies` in a composite have a lot of effect.

I also worry a bit about the do-no-harm aspects of the recipe as currently implemented: it seems [we remove pinned dependencies](https://github.com/openrewrite/rewrite/blob/86d235ac41480329485e668c4b35765bcf5007f7/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveBomManagedDirectDependenciesTest.java#L29-L79) without checking whether the dependency is also coming in transitive. We merely check whether the version is managed, but could end up removing dependencies that still are used/needed, and might instead need to see their version number dropped through `RemoveRedundantDependencyVersions`.

What are your thoughts on the above? Note that I've focused on the Maven recipes for now; I haven't checked yet if the same applies for Gradle.

bmuschko:
Thanks Tim — picking up your other two points, plus an update on the do-no-harm concern:

**(1) Overlap with `ChangeParentPom`/`RemoveRedundantDependencyVersions`.** They operate at the *version* level: drop the explicit `<version>` when the parent or BOM manages the same coordinate. This recipe operates at the *declaration* level: remove the whole `<dependency>` when (a) the BOM manages it at an incompatible major and (b) it is reachable transitively through another direct dep (see (3) below). They are complementary — version-strip handles "you pinned the same major", this handles "you pinned an old major that another starter now brings in for you".

**(2) Effectiveness inside a Spring Boot composite.** Fair point: after `ChangeParentPom`/`RemoveRedundantDependencyVersions` run, a same-major pin loses its `<version>` and the linked test on L201–232 keeps it — and that's the right outcome. Where this recipe earns its place in the composite is the *different-major* case (e.g., `<version>9.0.50</version>` under a Spring Boot 3 BOM managing 10.x): the version-strip alone wouldn't fire there because the explicit version no longer matches anything safely removable, but a sibling starter does bring the dep in transitively. So in a composite this runs after the version-level cleanup and sweeps up the remaining incompatible-major pins.

**(3) Do-no-harm — addressed in 63afd1ed3.** The recipe now requires the dep to be reachable transitively through another direct dep before removing it. Maven's mediation collapses deeper occurrences of the target onto the depth-0 direct declaration, so the resolved tree alone isn't enough; instead we walk the raw POMs of the other resolved deps via `MavenPomDownloader` (deduped by GA, scoped to compile/runtime/provided) and look for the target in their declared `<dependencies>`. For Gradle the resolved tree retains the deeper occurrences, so a tree walk via `cfg.getDirectResolved()` + `ResolvedDependency.findDependencies(g, a)` suffices — run per-configuration, so a dep that's safe to remove from `implementation` but not from `testImplementation` is left alone. If nothing else brings the dep in, the recipe now leaves it untouched (new `keepDependencyNotReachableTransitively` test in both modules covers that case).

timtebeek:
@copilot resolve the merge conflicts in this pull request

greg-at-moderne (approved):

PR #6582, opened 2026-01-21
Fix gradle bom upgrade

kdvolder:
## What's changed?

Fixed two issues in `UpgradeDependencyVersion` that prevented dependency upgrades in certain scenarios:

1. **BOM imports with literal versions** - Previously only BOM imports using property references were upgraded. Now literal versions like `mavenBom 'group:artifact:7.1.0'` are properly detected and upgraded.

2. **ext.set() syntax support** - The recipe only handled the `ext { }` block syntax for property definitions. Now it also handles the `ext.set("propertyName", "value")` method syntax.

## What's your motivation?

This issue was discovered while working on Spring Boot 4.0 migration recipes. Customers using App Advisor reported that BOM version upgrades were falsely reported as successful but made no actual changes to their Gradle build files.

**Specific scenarios that were broken:**
```groovy
// Scenario 1: BOM with literal version (not upgraded)
dependencyManagement {
    imports {
        mavenBom 'com.google.cloud:spring-cloud-gcp-dependencies:7.1.0'
    }
}

// Scenario 2: ext.set() syntax (not upgraded)
ext.set("springCloudGcpVersion", "7.1.0")
dependencies {
    implementation "com.google.cloud:spring-cloud-gcp-dependencies:${springCloudGcpVersion}"
}
```

Both scenarios now work correctly.

1. **BOM literal version handling** - The fix adds a check in `UpdateGradle.visitMethodInvocation()` to detect BOM entries without version variables and calls the new `updateBomEntry()` method. This reuses the existing `SpringDependencyManagementPluginEntry.withGroupArtifactVersion()` API.

2. **ext.set() handling** - Added `visitMethodInvocation()` override in the `UpdateVariable` visitor to detect `ext.set()` calls. The implementation correctly identifies this as a method invocation (not an assignment) and updates the second argument.

3. **Test coverage** - Added comprehensive test suite (`UpgradeDependencyVersionBomTest`) with 6 test cases covering all combinations of direct dependencies and BOM imports with literal versions, property references, and ext.set() syntax.

**Alternative considered:** Reusing the existing `ChangeExtraProperty` recipe for property updates. However:
- It doesn't currently support `ext.set()` syntax either (same bug exists there)
- Would require multiple tree traversals (one per property)
- Our two-phase scanning approach (scan dependencies → resolve versions → update) doesn't fit well with recipe composition

**Potential follow-up:** Extract a shared utility visitor for "update ext property" logic that could be used by both `UpgradeDependencyVersion` and `ChangeExtraProperty`, eliminating code duplication and fixing the `ext.set()` bug in both places.

## Any additional context

The property-based BOM upgrade case was recently fixed in commit bf0a8abb3 (Jan 14, 2026) by @bmuschko. This PR complements that work by handling the literal version case that was still missing.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/LatestMinor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/LatestMinor.java
@@ -47,7 +47,7 @@ public class LatestMinor implements VersionComparator {
         }
 
         //noinspection ConstantConditions
-        return TildeRange.build("~" + Semver.majorVersion(currentVersion) + "." + Semver.minorVersion(currentVersion), metadataPattern)
+        return TildeRange.build("~" + Semver.majorVersion(currentVersion), metadataPattern)
                 .getValue()
                 .compare(currentVersion, v1, v2);
     }
--- a/rewrite-core/src/main/java/org/openrewrite/semver/LatestPatch.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/LatestPatch.java
@@ -18,6 +18,7 @@ package org.openrewrite.semver;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.Validated;
+import org.openrewrite.internal.StringUtils;
 
 @Value
 public class LatestPatch implements VersionComparator {
@@ -28,7 +29,7 @@ public class LatestPatch implements VersionComparator {
     public boolean isValid(@Nullable String currentVersion, String version) {
         Validated<? extends VersionComparator> validated = currentVersion == null ?
                 LatestRelease.buildLatestRelease("latest.release", metadataPattern) :
-                TildeRange.build("~" + Semver.majorVersion(currentVersion) + "." + Semver.minorVersion(currentVersion), metadataPattern, true);
+                TildeRange.build(buildTildeRange(currentVersion), metadataPattern, true);
 
         if (validated.isValid()) {
             VersionComparator comparator = validated.getValue();
@@ -47,9 +48,15 @@ public class LatestPatch implements VersionComparator {
         }
 
         //noinspection ConstantConditions
-        return TildeRange.build("~" + Semver.majorVersion(currentVersion) + "." + Semver.minorVersion(currentVersion), metadataPattern)
+        return TildeRange.build(buildTildeRange(currentVersion), metadataPattern)
                 .getValue()
                 .compare(currentVersion, v1, v2);
+    }
+
+    private static String buildTildeRange(String currentVersion) {
+        String major = Semver.majorVersion(currentVersion);
+        String minor = Semver.minorVersion(currentVersion);
+        return StringUtils.isNumeric(minor) ? "~" + major + "." + minor : "~" + major;
     }
 
     public static Validated<LatestPatch> build(String toVersion, @Nullable String metadataPattern) {
--- a/rewrite-core/src/test/java/org/openrewrite/semver/LatestMinorTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/semver/LatestMinorTest.java
@@ -36,6 +36,7 @@ public class LatestMinorTest {
         assertThat(latestMinor.isValid("1.0", "1.0.1")).isTrue();
         assertThat(latestMinor.isValid("1.0.0", "1.1.0")).isTrue();
         assertThat(latestMinor.isValid("1.0.0", "2.0.0")).isFalse();
+        assertThat(latestMinor.isValid("1.0.x", "1.0.1")).isTrue();
     }
 
     @Test
@@ -47,6 +48,7 @@ public class LatestMinorTest {
     void compare() {
         assertThat(latestMinor.compare(null, "1.0", "latest.minor")).isNegative();
         assertThat(latestMinor.compare(null, "latest.minor", "1.0")).isPositive();
+        assertThat(latestMinor.compare("1.x", "1.0", "1.1")).isNegative();
     }
 
     @Test
@@ -66,5 +68,8 @@ public class LatestMinorTest {
         assertThat(latestMinor.upgrade("2.10.0", List.of("3.0.0-SNAPSHOT"))).isNotPresent();
 
         assertThat(latestMinor.upgrade("2.10.0", List.of("3.0.0"))).isNotPresent();
+
+        assertThat(latestMinor.upgrade("2.x", List.of("2.9"))).isPresent();
+        assertThat(latestMinor.upgrade("2.x", List.of("3.0"))).isNotPresent();
     }
 }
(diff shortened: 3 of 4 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionBomTest.java
```diff
@@ -364,19 +346,19 @@ void upgradesBomImportWithSetSyntax() {
...
               ext.set("springCloudGcpVersion", "7.2.0")
-
+              
               dependencyManagement {
                   imports {
                       mavenBom "com.google.cloud:spring-cloud-gcp-dependencies:${springCloudGcpVersion}"
                   }
               }
-
+
```
timtebeek: Can we trim these lines again? The bot doesn't like them.

kdvolder: I used intellij 'format file' but I guess it doesn't trim those? What tool do you actually use to format things the way you like it?

kdvolder: Couldn't make intellij clean this up. But this maybe works: 
```
sed -i 's/[[:space:]]*$//' <filename>
```

kdvolder:
> Potential follow-up: Extract a shared utility visitor for "update ext property" logic that could be used by both UpgradeDependencyVersion and ChangeExtraProperty, eliminating code duplication and fixing the ext.set() bug in both places.

This now also implemented with an extra 'witness' test case added for the 'ext.set(...)' syntax in 'ChangeExtraProperty' which apparantly was not supported yet there either.

The code now shares the implementation of a singel visitor to update properties between the two recipes which now both support that syntax.

rewrite-gradle/src/main/java/org/openrewrite/gradle/internal/UpdateExtraPropertiesVisitor.java
```diff
@@ -0,0 +1,192 @@
...
+ * Visitor to update properties in Gradle build scripts (both Groovy and Kotlin DSL).
+ * Handles multiple syntax forms:
+ * <ul>
+ *   <li>Variable declarations: def propertyName = 'value' or val propertyName = "value"</li>
+ *   <li>ext { propertyName = 'value' }</li>
+ *   <li>ext.propertyName = 'value'</li>
+ *   <li>ext.set("propertyName", "value")</li>
+ * </ul>
+ */
+public class UpdateExtraPropertiesVisitor extends JavaIsoVisitor<ExecutionContext> {
```
shanman190: I wonder if it wouldn't make more sense to go one step further and write this as a `Trait` implementation? The idea being that properties on the `ExtraExtension` are a common pattern and there's a few different ways that those can be created. Traits can be used as visitors as well. Some traits even expose edit-based helper methods to simplify making common changes.

kdvolder:
I tried the trait approach as suggested, seems like a good fit. The PR has been updated and now use Trait for both recipes. And I deleted my custom visitor.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/ExtraProperty.java
```diff
@@ -0,0 +1,280 @@
...
+                    J.Literal newValueLiteral = updateLiteral(valueLiteral, newValue);
+                    return method.withArguments(ListUtils.mapLast(method.getArguments(), arg -> newValueLiteral));
+                }
+                break;
+        }
+
+        return tree;
+    }
+
+    private J.Literal updateLiteral(J.Literal literal, String newValue) {
```
shanman190: `ChangeStringLiteral` is a helper that can do this for you.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/ExtraProperty.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+        return literal.withValue(newValue)
+                .withValueSource(quote + newValue + quote);
+    }
+
+    public static class Matcher extends GradleTraitMatcher<ExtraProperty> {
+        @Nullable
+        private String propertyName;
+
+        private boolean matchVariableDeclarations = true;
```
shanman190, line 140: I can go back and forth on this one. A variable in a Gradle script isn't _technically_ an extra property.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/ExtraPropertyTest.java
```diff
@@ -0,0 +1,237 @@
...
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.gradle.Assertions.buildGradle;
+
+class ExtraPropertyTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new FindExtraPropertiesRecipe());
```
shanman190: `RewriteTest#toRecipe` is a helper methods that assists you in converting a visitor to a recipe for the purposes of easy testing. It's especially useful for Traits.

Example
https://github.com/openrewrite/rewrite/blob/d49d71ced019b6270c426223fa49230484ab6c04/rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java#L34

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/ExtraProperty.java
jkschneider: Addressed in commit 4d83c41db - now uses `ChangeStringLiteral.withStringValue()` helper.

rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/ExtraPropertyTest.java
jkschneider: Addressed in commit 4d83c41db - refactored tests to use `RewriteTest.toRecipe()` pattern matching GradleDependencyTest style.

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/ExtraProperty.java
jkschneider: Good point. The `matchVariableDeclarations` option (defaults to `true`) lets consumers choose whether to include variable declarations. The `ChangeExtraProperty` recipe sets this to `false` to only match actual ext properties, while `UpgradeDependencyVersion` sets it to `true` to handle the case where version properties are defined as simple variables.

PR #6583, opened 2026-01-21
Allowing properties to be updated for `ChangeDependencyGroupIdAndArtifactId` / `ChangeManagedDependencyGroupIdAndArtifactId` only under safe circumstances.

steve-aom-elliott:
## What's changed?
- When there are no additional usages of a property used for a `version` tag of a dependency (managed or direct) that we're modifying using the mentioned recipes, including in the current local POM, the local parents, the remote parent or the modules of the current local POM, then it will allow it to update the property.

## What's your motivation?
- Avoid quite as much detaching of properties from dependency setup

### Checklist

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -290,8 +290,7 @@ public class ChangeDependencyGroupIdAndArtifactId extends Recipe {
                         .collect(toSet());
                 relevantProperties = filterPropertiesWithOverlapInDependencies(relevantProperties, groupId, artifactId, requestedPom, resolvedPom, configuredToChangeManagedDependency);
                 relevantProperties = filterPropertiesWithOverlapInChildren(relevantProperties, groupId, artifactId, result, configuredToChangeManagedDependency);
-                relevantProperties = filterPropertiesWithOverlapInParents(relevantProperties, groupId, artifactId, result, configuredToChangeManagedDependency, ctx);
-                return relevantProperties;
+                return filterPropertiesWithOverlapInParents(relevantProperties, groupId, artifactId, result, configuredToChangeManagedDependency, ctx);
             }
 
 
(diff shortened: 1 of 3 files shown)


steve-aom-elliott:
Making one final change to improve the behaviour when the `artifactId` to `ChangeDependencyGroupIdAndArtifactId` is a glob pattern

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -145,12 +149,16 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+            private final boolean configuredToOverrideManagedVersion = overrideManagedVersion != null && overrideManagedVersion; // False by default
+            private final boolean configuredToChangeManagedDependency = changeManagedDependency == null || changeManagedDependency;  // True by default
 
             @Override
             public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
                 isNewDependencyPresent = checkIfNewDependencyPresent(newGroupId, newArtifactId, newVersion);
-                if (!oldGroupId.contains("*") && !oldArtifactId.contains("*") &&
-                        (changeManagedDependency == null || changeManagedDependency)) {
+                hasProblematicVersionPlaceholder = hasProblematicPropertyUsage(oldGroupId, oldArtifactId, ctx);
+                if (!oldGavPatternsUseGlob && configuredToChangeManagedDependency) {
```
sambsnyd: It should be possible to support glob patterns for this. Convert the globs into a list of matching dependencies and iterate over it providing the transformation for each.

steve-aom-elliott: @sambsnyd I've updated `ChangeManagedDependencyGroupIdAndArtifactId` to be able to handle globs now.

steve-aom-elliott:
New functionality added in last set of changes:
- `ChangeManagedDependencyGroupIdAndArtifactId` can now take glob expressions for `groupId` and `artifactId`
- Overlap detection will now figure out the ability to alter a property for each distinct property referenced in the versions tags when changing dependencies using glob expressions.
- Will now traverse multiple remote parents until it either runs out of candidate properties or runs out of parents

PR #6588, opened 2026-01-24
Make it easier to troubleshoot missing artifacts

timtebeek:
## What's changed?
Fix an issue with how artifacts are matched, and make it easier to troubleshoot any issues this might uncover.

## What's your motivation?
We could quite easily mismatch an artifact, as the full path was used, leading to hard to troubleshoot missing type information.

We could optionally only change the matcher, but then we'd make it hard to find any such issues downstream.

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/format/SpacesVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/format/SpacesVisitor.java
@@ -79,7 +79,7 @@ public class SpacesVisitor<P> extends org.openrewrite.java.format.SpacesVisitor<
         return super.visitSpace(space, loc, ctx);
     }
 
-    @ToBeRemoved(after = "30-01-2026", reason = "Replace me with org.openrewrite.style.StyleHelper.getStyle now available in parent runtime")
+    @ToBeRemoved(after = "2026-01-30", reason = "Replace me with org.openrewrite.style.StyleHelper.getStyle now available in parent runtime")
     private static <S extends Style> @Nullable S getStyle(Class<S> styleClass, List<NamedStyles> styles) {
         S style = NamedStyles.merge(styleClass, styles);
         if (style != null) {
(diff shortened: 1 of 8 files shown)


rewrite-java/src/main/java/org/openrewrite/java/internal/parser/RewriteClasspathJarClasspathLoader.java
```diff
@@ -59,9 +63,10 @@ public RewriteClasspathJarClasspathLoader(ExecutionContext ctx) {
 
     @Override
     public @Nullable Path load(String artifactName) {
-        Pattern jarPattern = Pattern.compile(artifactName + "-?.*\\.jar$");
+        Pattern jarPattern = Pattern.compile(artifactName + "-?.*\\.jar");
         for (Resource resource : resources) {
-            if (jarPattern.matcher(resource.getPath()).find()) {
+            String fileName = Paths.get(resource.getPath()).getFileName().toString();
+            if (jarPattern.matcher(fileName).matches()) {
```
timtebeek, line 69: This is where we mere matching against the full path, not just the filename, meaning ambiguous entries like `"log4j"` could mistakenly match others too.

rewrite-java/src/main/java/org/openrewrite/java/internal/parser/RewriteClasspathJarClasspathLoader.java
```diff
@@ -59,9 +63,10 @@ public RewriteClasspathJarClasspathLoader(ExecutionContext ctx) {
...
                 try {
                     Path artifact = getJarsFolder(ctx).resolve(Paths.get(resource.getPath()).getFileName());
```
sambsnyd, line 66: The dollar sign at the end meant "end of string" so removing it could potentially match `foo.jar.tar.gz` or something.  That said the danger of that actually happening is basically nonexistent, so it probably doesn't matter.

PR #6597, opened 2026-01-26
Add test for AddDependency upgrading version in parent POM

jkschneider:
## Summary
- Adds test coverage for the scenario where an in-repo parent has a dependency with a lower version than requested
- Verifies that AddDependency correctly upgrades the version in the parent POM

## Test plan
- [x] New test `addDependencyWhenInRepoParentHasDependencyWithLowerVersionAddsToParent` passes

- Related to: moderneinc/customer-requests#1482


timtebeek:
Closing as the issue can no longer be reproduced; likely fixed in
- https://github.com/openrewrite/rewrite/pull/6505

PR #6599, opened 2026-01-26
Handle invalid UTF-16 surrogate pairs in JavaScript/TypeScript parser

jkschneider:
## Summary
- Fix surrogate pair handling to prevent ingestion failures
- Extract surrogate unicode escapes into `J.LiteralUnicodeEscape` objects instead of throwing an error
- Handle both `\uXXXX` escape sequences and raw surrogate characters in source

## What's the problem?
When JavaScript/TypeScript files contain unmatched UTF-16 surrogate pairs (like `\uD800`), the parser was throwing `InvalidSurrogatesNotSupportedError`, causing ingestion to fail. This prevents parsing of valid JavaScript files like those in vscode, azuredatastudio, and other popular repositories.

## What's the solution?
Apply the same approach used in the Java parser:
1. Scan for surrogate pairs in the value source (both escape sequences and raw characters)
2. Extract them into `LiteralUnicodeEscape` objects with their position and code point
3. Remove them from the cleaned value source string

This allows Jackson to serialize the AST without errors, while preserving the surrogate information for when the source is printed back.

## Test plan
- [x] Existing tests in `LiteralTest.java` (`unmatchedSurrogatePair` and `unmatchedSurrogatePairInString`) should now pass
- [ ] Verify ingestion works for problematic repos:
  - https://github.com/microsoft/vscode
  - https://github.com/microsoft/azuredatastudio
  - https://github.com/ionic-team/stencil

- Fixes moderneinc/customer-requests#829
- Fixes moderneinc/customer-requests#830

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/parser.ts
@@ -43,7 +43,6 @@ import {
     getPreviousSibling,
     hasFlowAnnotation,
     isStatement,
-    isValidSurrogateRange,
     TextSpan
 } from "./parser-utils";
 import {JavaScriptTypeMapping} from "./type-mapping";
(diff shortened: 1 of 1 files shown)


knutwannheden:
FWIW: IIRC, this is something that we could fix by upgrading to Jackson 2.18.

PR #6605, opened 2026-01-26
Fix Jackson 2.21 compatibility for MavenSettings ServerConfiguration parsing

timtebeek:
## Summary

- Fix Jackson 2.21 compatibility issue when parsing `MavenSettings` with `ServerConfiguration` containing `httpHeaders`
- The issue was that Jackson 2.21 became stricter about matching constructor parameter names with XML property names

## Root Cause

The `@JacksonXmlProperty(localName = "property")` annotation on the `httpHeaders` field made Jackson think the property name was "property", but the Lombok-generated constructor had a parameter named "httpHeaders", causing a mismatch error:

```
InvalidDefinitionException: Could not find creator property with name 'httpHeaders' (known Creator properties: [property, timeout])
```

## Changes

- `ServerConfiguration`: Use a no-args constructor (marked with `@JsonCreator`) and setters for deserialization, with Jackson XML annotations moved to getters
- `HttpHeader`: Add explicit `@JsonCreator` constructor with `@JsonProperty` annotations

## Test plan

- [x] All `MavenSettingsTest` tests pass with both Jackson 2.17.x and 2.21.x
- [x] Full `rewrite-maven` test suite passes with both Jackson versions


timtebeek:
Verified locally with both 2.17.3 as well as the latest 2.21.0; we need this for the work done in
- https://github.com/openrewrite/rewrite-spring-to-quarkus/pull/63

Through
```kts
    implementation(platform("com.fasterxml.jackson:jackson-bom:2.21.0"))
```

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -446,22 +446,43 @@ public static class Server {
...
          */
         @Nullable
-        Long timeout;
+        private Long timeout;
+
+        @JsonCreator
+        public ServerConfiguration() {
+        }
+
+        public ServerConfiguration(@Nullable List<HttpHeader> httpHeaders, @Nullable Long timeout) {
```
shanman190, line 468: Any reason not to mark this as the `JsonCreator` given we have the javac `-parameters` option enabled?

It should remove the need for the setters as well.

timtebeek: Unfortunately yes, as that breaks parsing on 2.17.3 already:
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid definition for property 'httpHeaders' (of type `org.openrewrite.maven.MavenSettings$ServerConfiguration`): Could not find creator property with name 'httpHeaders' (known Creator properties: [property, timeout])
 at [Source: (org.openrewrite.internal.EncodingDetectingInputStream); line: 1, column: 1]
```

shanman190: I see.

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -446,22 +446,43 @@ public static class Server {
     }
 
     @SuppressWarnings("DefaultAnnotationParam")
-    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
-    @Data
```
shanman190, line 450: Are we concerned about losing `equals`/`hashCode`/`toString` as a side effect here?

PR #6607, opened 2026-01-26
Add tests documenting AddDependency behavior for existing dependencies

jkschneider:
## Summary
- Adds tests verifying AddDependency correctly avoids duplicating existing dependencies
- Documents that dependencies already declared in build.gradle are not added again
- Documents that transitively available dependencies (e.g., via Spring Boot starter) are not added when `acceptTransitive=true`

## Context
- Investigation of https://github.com/moderneinc/customer-requests/issues/760 confirmed that the AddDependency recipe correctly handles existing dependencies. These tests document and protect this behavior.

## Test plan
- [x] Run `./gradlew :rewrite-gradle:test --tests "org.openrewrite.gradle.AddDependencyTest"`


timtebeek:
Ok test passes; no need to merge as issue can no longer be replicated.

PR #6609, opened 2026-01-27
Select minimum compatible version when multiple artifacts share a Maven property

kmccarp:
## Summary

- When multiple dependencies share a Maven property but have different max versions, select the minimum version that works for all artifacts
- Track version constraints from all matching dependencies during the scanner phase
- Select the minimum version during the visitor phase to ensure all artifacts can be resolved

## Problem

When running `UpgradeDependencyVersion` with a wildcard artifactId pattern (e.g., `org.springframework.cloud:*`), if multiple dependencies use the same Maven property for their version (e.g., `${spring-cloud.version}`) but each artifact has different maximum versions available, the recipe could select a version that doesn't exist for all artifacts.

For example:
- `spring-cloud-commons` has version 4.2.4 available
- `spring-cloud-starter-consul-config` only has up to 4.2.3

The recipe would set the shared property to 4.2.4, causing resolution failures for consul-config.

## Solution

Modified the scanner and visitor phases in `UpgradeDependencyVersion`:

1. **Scanner phase**: Now tracks ALL version constraints for properties shared by multiple artifacts. When an artifact is already at its maximum compatible version, that version is recorded as a constraint to prevent other artifacts from pushing the shared property higher.

2. **Visitor phase**: When upgrading a dependency that uses a shared property, looks up the minimum version from all collected candidates and uses that instead of the artifact-specific version.

3. **Accumulator**: Changed from storing individual property values to storing sets of candidate versions per property, with a new `getResolvedPropertyVersion()` method that returns the minimum version.

## Test plan

- [x] Existing tests pass
- [x] New test `upgradeMultipleArtifactsWithSharedPropertySelectsMinimumVersion` added that verifies:
  - POM with `<spring-cloud.version>4.1.0</spring-cloud.version>`
  - Two dependencies using `${spring-cloud.version}` with different max versions
  - Recipe correctly selects 4.2.3 (the minimum) instead of 4.2.4

- Fixes moderneinc/customer-requests#1761


rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -2537,4 +2537,75 @@ void upgradeMultipleManagedDependenciesWithWildcards() {
...
+     * <p>
+     * This test simulates the Spring Cloud scenario where different artifacts have different max patch versions:
+     * - spring-cloud-commons has versions up to 4.2.4
+     * - spring-cloud-starter-consul-config only has versions up to 4.2.3
+     * <p>
+     * If both use the same property and we upgrade with `4.2.x`, the property should be set to 4.2.3
+     * (the minimum compatible version), not 4.2.4 (which would break consul-config).
+     *
+     * @see <a href="https://github.com/moderneinc/customer-requests/issues/1761">Customer issue</a>
+     */
```
kmccarp:
```suggestion

```

rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -2537,4 +2537,75 @@ void upgradeMultipleManagedDependenciesWithWildcards() {
...
+    @Test
+    @Issue("https://github.com/moderneinc/customer-requests/issues/1761")
```
kmccarp:
```suggestion
```

rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -2537,4 +2537,75 @@ void upgradeMultipleManagedDependenciesWithWildcards() {
...
+    void upgradeMultipleArtifactsWithSharedPropertySelectsMinimumVersion() {
+        // Using real Spring Cloud artifacts where:
+        // - spring-cloud-commons has 4.2.4 available
+        // - spring-cloud-starter-consul-config only has up to 4.2.3
```
kmccarp, line 2546:
```suggestion
        // - spring-cloud-starter-consul-config only has up to 4.2.3
        // we can only upgrade to the minimum (4.2.3)
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -2308,4 +2308,93 @@ void upgradesSpringDependencyManagementPluginMavenBomVersionPropertyWithKotlinDS
...
+
+    /**
+     * When multiple artifacts share a version variable and have different maximum available versions,
+     * the recipe should select the minimum compatible version that works for ALL artifacts.
+     * <p>
+     * spring-cloud-commons has versions up to 4.2.4
+     * spring-cloud-starter-consul-config only has versions up to 4.2.3
+     * <p>
+     * When upgrading to 4.2.x, the variable should be set to 4.2.3 (not 4.2.4) so both artifacts resolve.
+     */
```
kmccarp:
```suggestion
```

rewrite-gradle/src/test/java/org/openrewrite/gradle/UpgradeDependencyVersionTest.java
```diff
@@ -2308,4 +2308,93 @@ void upgradesSpringDependencyManagementPluginMavenBomVersionPropertyWithKotlinDS
...
+                  .doesNotContain("springCloudVersion = '4.2.4'");
+                return actual;
+            })
+          )
+        );
+    }
+
+    /**
+     * Same test as above but using gradle.properties instead of ext block.
+     */
```
kmccarp:
```suggestion
```

timtebeek:
Closing as suggested internally.

PR #6612, opened 2026-01-27
Keep trailing newlines in literal/folded YAML scalar values

knutwannheden:
Aligns the Java YAML parser behavior with the TypeScript parser for literal (`|`) and folded (`>`) scalars.

### Changes

- **YamlParser**: Remove newline stripping from literal/folded scalar values - the trailing newline is now part of the scalar value rather than being prepended to the next element's prefix
- **IndentsVisitor**: Update lastIndent tracking for mapping entries whose prefix has no leading newline (entries following literal/folded scalars)
- **DeleteKey**: Add prefix propagation when deleting sequence entries to preserve proper formatting

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -310,10 +310,6 @@ public class YamlParser implements org.openrewrite.Parser {
                             case FOLDED:
                             case LITERAL:
                                 scalarValue = reader.readStringFromBuffer(valueStart + 1, event.getEndMark().getIndex() - 1);
-                                if (scalarValue.endsWith("\n")) {
-                                    newLine = "\n";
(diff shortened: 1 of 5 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
```diff
@@ -310,10 +310,6 @@ private Yaml.Documents parseFromInput(Path sourceFile, EncodingDetectingInputStr
                             case FOLDED:
                             case LITERAL:
                                 scalarValue = reader.readStringFromBuffer(valueStart + 1, event.getEndMark().getIndex() - 1);
-                                if (scalarValue.endsWith("\n")) {
-                                    newLine = "\n";
-                                    scalarValue = scalarValue.substring(0, scalarValue.length() - 1);
-                                }
```
knutwannheden, line 316: Stripping the scalar's value like this from the terminating newline is semantically wrong, which this PR is all about.

PR #6619, opened 2026-01-27
Unique precondition

sambsnyd:
Add a "Highlander" precondition named "Unique" (since not everyone will get the "there can be only one" reference) which can be used in Declarative and non-declarative recipes to ensure that only one copy of the recipe gets to make changes within a recipe run.

Will follow up with a rewrite-rewrite recipe to add this to declarative recipes. 

I incidentally noticed and fixed some missing method delegations in DeclarativeRecipe.Bellwether classes. I'm not aware of the missing delegations being known to cause any bugs but it can only be more correct to include all of them.


jkschneider:
WDYT about "OnlyRunOnce" which may be more descriptive than "Unique"?

sambsnyd:
> WDYT about "OnlyRunOnce" which may be more descriptive than "Unique"?

Hmm that is more descriptive. But perhaps what it describes can be misinterpreted. One might assume that "Run only once" is talking about being single-cycle. A "highlander" recipe is still allowed to make changes across multiple cycles (so long as we continue to have that concept, anyway).

@jkschneider how do you feel about calling it `Deduplicate` or `NoDuplicates`? I think that gets at the heart of what is going on here. `RunOnce` or `SingleExecution` or similar sound like concepts that might affect a single instance of a recipe on its own.

PR #6624, opened 2026-01-28
Allow Gradle AddDependency to work with explicit configuration without Java sources

bmuschko:
## What's changed?

Modified `AddDependency` in rewrite-gradle to allow adding dependencies when an explicit configuration is provided, even without Java source files being scanned. Previously, the recipe required Java sources to determine which configuration to use. Now, when configuration is explicitly set (e.g., "implementation"), the recipe bypasses the scan phase requirement and adds the dependency directly.

## What's your motivation?

The behavior of the Maven and the Gradle `AddDependency` behavior differs. That's specifically problematic when you use the convenience `AddDependency` recipe from rewrite-java-dependencies.

The alternative is requiring callers to fully integrate with the scan phase by overriding `getInitialValue()` and `getScanner()`, but this adds significant complexity for recipes that already know which configuration they want to target.

## Any additional context

This aligns with how Maven's `AddDependency` works - it allows adding dependencies when `onlyIfUsing` is `null` without requiring scan phase data.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -181,9 +181,10 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
-        // Allow when configuration is explicitly provided OR when source files were scanned
+        // Allow when configuration is explicitly provided, when onlyIfUsing is not set (default to "implementation"),
+        // or when source files were scanned
         boolean hasExplicitConfiguration = !StringUtils.isBlank(configuration);
-        return Preconditions.check(hasExplicitConfiguration || !acc.configurationsByProject.isEmpty(),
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -181,7 +181,9 @@ private boolean usesType(SourceFile sourceFile, ExecutionContext ctx) {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
-        return Preconditions.check(!acc.configurationsByProject.isEmpty(),
+        // Allow when configuration is explicitly provided OR when source files were scanned
+        boolean hasExplicitConfiguration = !StringUtils.isBlank(configuration);
+        return Preconditions.check(hasExplicitConfiguration || !acc.configurationsByProject.isEmpty(),
```
sambsnyd: I think the behavior in maven is we assume "compile" when no explicit configuration, right? We could do the same thing here and default to "implementation" when no configuration is explicitly specified

bmuschko: done

PR #6630, opened 2026-01-29
`ChangeDependencyGroupIdAndArtifactId` updates version properties in parent POMs

MBoegers:
## Summary

- Convert `ChangeDependencyGroupIdAndArtifactId` from `Recipe` to `ScanningRecipe` to support updating version properties defined in parent POMs
- Add minimal scanner that only identifies parent POM version properties needing updates
- Add test for multi-module project with version property in parent POM

## Problem

When `ChangeDependencyGroupIdAndArtifactId` is used on a multi-module project where a child POM has a dependency with a version property (e.g., `${version.swagger}`) defined in the parent POM, the recipe changes the groupId in the child POM but fails to update the version property value in the parent POM.

- A [previous attempt](https://github.com/openrewrite/rewrite/pull/5815) to fix this was reverted ([#5903](https://github.com/openrewrite/rewrite/pull/5903)) because it moved all visitor logic into the scanner, causing recipe-chaining failures ([#5673](https://github.com/openrewrite/rewrite/issues/5673)).

## Solution

Take a minimal approach: keep all existing visitor logic unchanged and only add a scanner that does one thing — find version properties defined in parent POMs that need updating. The accumulator stores only immutable `(pomFilePath, propertyName, newValue)` tuples.

The visitor adds one piece of new logic: when visiting property tags, check the accumulator and apply `ChangeTagValueVisitor` if matched. This follows the same pattern used by `UpgradeDependencyVersion`.

### Why this avoids the previous regression

1. **No mutable per-dependency fields in the accumulator** — only immutable `PomProperty` records
2. **Existing visitor logic preserved** — `changeChildTagValue`, `maybeUpdateModel()`, managed dependency handling all stay as-is
3. **Recipe chaining safe** — the scanner only reads the Maven model (no modifications)

## Test plan

- [x] New test `changeVersionPropertyInParentPomSimple` passes
- [x] Full `ChangeDependencyGroupIdAndArtifactIdTest` suite passes
- [x] Full `:rewrite-maven:test` suite passes (no regressions)

- Fixes moderneinc/customer-requests#1374

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -297,6 +297,7 @@ public class ChangeDependencyGroupIdAndArtifactId extends ScanningRecipe<ChangeD
                         }
                         break;
                     }
+                    return t;
                 }
 
                 boolean isOldDependencyTag = isDependencyTag(oldGroupId, oldArtifactId);


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -142,7 +144,98 @@ public Validated<Object> validate() {
...
+        return new MavenIsoVisitor<ExecutionContext>() {
+            final @Nullable VersionComparator versionComparator = newVersion != null ? Semver.validate(newVersion, versionPattern).getValue() : null;
+            final boolean configuredToChangeManagedDependency = changeManagedDependency == null || changeManagedDependency;
+
+            @Override
+            public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                if (!isDependencyTag(oldGroupId, oldArtifactId) && !isPluginDependencyTag(oldGroupId, oldArtifactId) && !isAnnotationProcessorPathTag(oldGroupId, oldArtifactId)) {
+                    return super.visitTag(tag, ctx);
+                }
+                if (newVersion == null || versionComparator == null) {
```
steve-aom-elliott: Wouldn't these two conditions only ever both be true or both be false? Unsure if `Semver.validate(..).getValue()` can give back `null`

MBoegers: Yes it can, so both could be null, unlikely in the case of the. version comperator but could.

Jenson3210:
There's definitely less risk in this one compared to moving everything. Is it handled in the existing filtering that another submodule could use the same property for another non-matching dependency?  So perhaps just add a test for that?

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -176,6 +269,23 @@ public Xml visitDocument(Xml.Document document, ExecutionContext ctx) {
...
+                    Path pomSourcePath = getResolutionResult().getPom().getRequested().getSourcePath();
+                    for (PomProperty prop : acc.getPomProperties()) {
+                        if (!prop.getPomFilePath().equals(pomSourcePath) || !prop.getPropertyName().equals(tag.getName())) {
+                            continue;
+                        }
+                        Optional<String> value = tag.getValue();
+                        if (!value.isPresent() || !value.get().equals(prop.getNewValue())) {
+                            doAfterVisit(new ChangeTagValueVisitor<>(tag, prop.getNewValue()));
+                            maybeUpdateModel();
+                        }
```
steve-aom-elliott, line 297: In line with what Jente mentioned, this looks at though it will change the property's value regardless of whether the property was also used by a dependency that we aren't changing.

So if you had a parent with property `x` and a child that had dependency `y` and dependency `z` that both use property `x` for their version, and you're only modifying dependency `y` with this recipe, it can result in broken dependency resolution potentially.

MBoegers:
Thanks for reviewing @Jenson3210 and @steve-aom-elliott!

Both of your concerns are valid but are already handled by the existing `MavenDependencyPropertyUsageOverlap` filtering logic.

**The scenario you describe:**
- Parent POM has `${my.version}`
- Child POM has dependency A (matching) and dependency B (non-matching), both using `${my.version}`

**What the code does:**
1. The scanner finds the matching dependency with `${my.version}`
2. Before storing the property for update, it calls `getSafeVersionPlaceholdersToChange()` (line 174)
3. This method uses `filterPropertiesWithOverlapInDependencies()` which removes the property from the "safe" set if ANY non-matching dependency also uses it
4. Since the property is removed from the safe set, the scanner does NOT store it for update
5. The visitor then falls through to the existing inline-version logic

**Prove:**
I've added a new test `sharedPropertyInParentPomLeavesPropertyUnchangedAndInlinesVersion` to explicitly cover this scenario. The test confirms that when `io.swagger:swagger-annotations` and `io.swagger:swagger-models` both use `${version.swagger}` from the parent, changing only `swagger-annotations` will:
- Leave the parent property unchanged
- Inline the new version in the child dependency

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
MBoegers: Thanks for reviewing @Jenson3210 and @steve-aom-elliott!

Both of your concerns are valid but are already handled by the existing `MavenDependencyPropertyUsageOverlap` filtering logic.

**The scenario you describe:**
- Parent POM has `${my.version}`
- Child POM has dependency A (matching) and dependency B (non-matching), both using `${my.version}`

**What the code does:**
1. The scanner finds the matching dependency with `${my.version}`
2. Before storing the property for update, it calls `getSafeVersionPlaceholdersToChange()` (line 174)
3. This method uses `filterPropertiesWithOverlapInDependencies()` which removes the property from the "safe" set if ANY non-matching dependency also uses it
4. Since the property is removed from the safe set, the scanner does NOT store it for update
5. The visitor then falls through to the existing inline-version logic

**Prove:**
I've added a new test `sharedPropertyInParentPomLeavesPropertyUnchangedAndInlinesVersion` to explicitly cover this scenario. The test confirms that when `io.swagger:swagger-annotations` and `io.swagger:swagger-models` both use `${version.swagger}` from the parent, changing only `swagger-annotations` will:
- Leave the parent property unchanged
- Inline the new version in the child dependency

Jenson3210: I would just change the test a bit to also verify this works as expected when 2 different childs use the same property iso 1 child sharing the same property. (Currently the added test verifies the same property in parent from 1 child only) 

If that also passes (reviewing from phone), LGTM!

steve-aom-elliott: Ah, right, forgot what I had named the helper method in `MavenDependencyPropertyUsageOverlap`, so that's fair. Sounds good to me as well with adding the test showing this.

steve-aom-elliott:
@MBoegers Can you double check that your last changes also gets covered for `ChangeManagedDependencyGroupIdAndArtifactId` if applicable?

steve-aom-elliott:
@MBoegers Something to be aware of is that the child overlap check right now will also filter out properties defined in the child. I tried altering the logic so that it can allow changes in the scenario that a property is redefined, but due to the way that [this](https://github.com/openrewrite/rewrite/blob/d634c658674a88f1c85711ffb25cf7bf8a846b44/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java#L309) and [this](https://github.com/openrewrite/rewrite/blob/d634c658674a88f1c85711ffb25cf7bf8a846b44/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java#L325) operate (using `doAfterVisit(..)` without providing the focused context document in which to run), when it runs `ChangePropertyValue`, it will change _all_ occurrences of the property definition, not just the one you actually wanted to change.

A test that ideally would have passed, but won't is:
```java
@Issue("https://github.com/openrewrite/rewrite/pull/6630#discussion")
@Test
void sharedPropertyUsedNonConflictinglyInChildrenChangesProperty() {
    rewriteRun(
      spec -> spec.recipe(new ChangeDependencyGroupIdAndArtifactId(
        "io.swagger",
        "swagger-annotations",
        "io.swagger.core.v3",
        null,
        "2.2.x",
        null,
        null,
        null
      )),
      mavenProject("project",
        pomXml(
          //language=xml
          """
            <project>
              <artifactId>parent-project</artifactId>
              <version>1</version>
              <properties>
                <version.swagger>1.5.16</version.swagger>
              </properties>
              <modules>
                <module>sub-project</module>
              </modules>
              <dependencies>
                <dependency>
                  <groupId>io.swagger</groupId>
                  <artifactId>swagger-annotations</artifactId>
                  <version>${version.swagger}</version>
                </dependency>
              </dependencies>
            </project>
            """,
          //language=xml
          """
            <project>
              <artifactId>parent-project</artifactId>
              <version>1</version>
              <properties>
                <version.swagger>2.2.42</version.swagger>
              </properties>
              <modules>
                <module>sub-project</module>
              </modules>
              <dependencies>
                <dependency>
                  <groupId>io.swagger.core.v3</groupId>
                  <artifactId>swagger-annotations</artifactId>
                  <version>${version.swagger}</version>
                </dependency>
              </dependencies>
            </project>
            """,
          spec -> spec.path("pom.xml")
        ),
        mavenProject("sub-project",
          pomXml(
            //language=xml
            """
              <project>
                <artifactId>sub-project</artifactId>
                <version>1</version>
                <parent>
                  <artifactId>parent-project</artifactId>
                  <version>1</version>
                  <relativePath>../pom.xml</relativePath>
                </parent>
                <properties>
                  <version.swagger>1.5.17</version.swagger>
                </properties>
                <dependencies>
                  <dependency>
                    <groupId>io.swagger</groupId>
                    <artifactId>swagger-models</artifactId>
                    <version>${version.swagger}</version>
                  </dependency>
                </dependencies>
              </project>
              """,
            spec -> spec.path("sub-project/pom.xml")
          )
        )
      )
    );
}
```

MBoegers:
The overly conservative filter is actually a safety net against a deeper issue. `ChangePropertyValue` runs via `doAfterVisit` without scoping to the current document — meaning it changes the property in every POM that defines it, not just the one we intended. If we relaxed the filter to allow updating the parent property when a child redefines it, the `ChangePropertyValue` would also overwrite the child’s intentional override value, silently breaking it. The current behavior (inlining the version) is less elegant but correct. 

Fixing this properly requires scoping `doAfterVisit` to a single document, which is an architectural change outside the scope of this PR.

@steve-aom-elliott I added a test to document this behavior.

PR #6631, opened 2026-01-29
Fix JSON parser handling of special characters inside strings

timtebeek:
## Summary

- Fix `positionOfNext()` method to track string literal boundaries
- Add proper escape sequence handling for both single and double quoted strings
- Add 7 new test cases covering the reported issues

## Problem

The JSON parser's `positionOfNext()` method was not tracking whether it was inside a string literal. This caused special characters like `**`, `/*`, `//`, `:`, `,`, `{`, `}` inside strings to be incorrectly interpreted as structural delimiters or comment markers.

This led to idempotency failures for JSON files containing:
- Glob patterns like `**/*.java`
- URLs with protocols like `https://`
- Comment-like sequences in strings (`/* */`, `//`)
- FreeMarker templates (`<#if>`)
- Printf templates (`%s`)
- Delimiters inside string values

## Solution

Modified `positionOfNext()` to:
1. Track when inside double-quoted strings (`"..."`)
2. Track when inside single-quoted strings (`'...'` for JSON5)
3. Handle escape sequences properly (e.g., `\"` doesn't end the string)
4. Skip delimiter/comment detection when inside strings

## Test plan

- [x] New tests added covering all reported issue types
- [ ] Existing tests pass (requires CI with all Java toolchains)

- Fixes moderneinc/customer-requests#852

### DIFF
--- a/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
+++ b/rewrite-json/src/main/java/org/openrewrite/json/internal/JsonParserVisitor.java
@@ -325,39 +325,45 @@ public class JsonParserVisitor extends JSON5BaseVisitor<Json> {
     private int positionOfNext(String untilDelim, @Nullable Character stop) {
         boolean inMultiLineComment = false;
         boolean inSingleLineComment = false;
+        char stringQuote = 0; // 0 = not in string, '"' or '\'' = in that string type
 
         int delimIndex = cursor;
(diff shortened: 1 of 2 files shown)


timtebeek:
All the tests added up to now also work on the main branch; not sure this adds anything then for now.

timtebeek:
We can't seem to create any tests that don't already pass on main; closing this PR and the linked issue.

MBoegers:
see https://github.com/openrewrite/rewrite/pull/6635 this one has a failing test and a fix

PR #6635, opened 2026-01-29
Fix JSON parser idempotency issues with non-UTF-8 encoded files

MBoegers:
## Summary
- Fix encoding mismatch in `JsonParser` where source was read twice with potentially different encodings (ANTLR defaulting to UTF-8, visitor detecting Windows-1252). Now the source is read once and shared.
- Add null guard in `JsonParserVisitor.visitValue()` so invalid JSON produces `ParseError` instead of `NullPointerException`.
- Add test cases for accented characters, `**` in strings, no trailing newline, and invalid JSON handling.

## Test plan
- [x] New idempotency tests pass (accented characters, `**` in strings, no trailing newline)
- [x] Invalid JSON test produces `ParseError` instead of NPE
- [x] All existing `rewrite-json` tests pass

### DIFF
--- a/rewrite-json/src/test/java/org/openrewrite/json/JsonParserTest.java
+++ b/rewrite-json/src/test/java/org/openrewrite/json/JsonParserTest.java
@@ -16,12 +16,7 @@
 package org.openrewrite.json;
 
 import org.junit.jupiter.api.Test;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.InMemoryExecutionContext;
-import org.openrewrite.Issue;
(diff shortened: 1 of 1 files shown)


timtebeek:
Nice, do we have any way to get to a test that fails on main but succeeds with the changes here?

MBoegers:
Yes, your PR helped my instance to find an even better test to reproduce the bug

timtebeek (approved):
Great to see this replicated and fixed, finally. Thanks for working out that it was an encoding issue!

Issue #5490, opened 2025-05-24
Annotating method with `lombok.Generated` causes parser failure


protocol7:
## What version of OpenRewrite are you using?

org.openrewrite:rewrite-core:8.53.0

## What is the smallest, simplest way to reproduce the problem?

When a method is annotated with `lombok.Generated` the parser will fail. We had this with a source code file, likely originating from [delombok](https://projectlombok.org/features/delombok). Here's a test case:

```java
  @Test
  void lombokGeneratedParsing() {
    rewriteRun(
        java(
            """
            import lombok.Generated;

            public class Foo {

              @Generated
              public void foo() {}
            }
            """));
  }
```

This fails with:

```
java.lang.AssertionError: Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser.
import lombok.Generated;

public class Foo {~~(non-whitespace)~~>

  @Generated
  public void foo() {<~~}~~(non-whitespace)~~>
}<~~
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:324)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:130)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:125)
```

In our rewrites, this generated invalid code, causing compilation failures.

timtebeek:
Thanks! That's quite unexpected usage here. I suspect it's caused by our Lombok annotation handling, but would have to run your test to see what's going on. Is this a common pattern over there or did you already phase out what limited usage there was?

e5LA:
Indeed using `@lombok.Generated` manually like this is not typical / not expected usage. 
It's intended to be added by Lombok itself to generated methods (e.g., via `@Setter`, `@Getter`, etc.).

I've checked whether there's a reliable way to distinguish between `@lombok.Generated` annotations that are applied by Lombok vs ones written manually in source code, but unfortunately, I couldn't find solution..

I hoped that `synthesized` flag could help, but seems it is `false` in both cases.

onekzy:
@timtebeek Seems I faced a similar issue.
The run of **recipe org.openrewrite.java.format.TabsAndIndents** causes deleting the **public** modifier from class marked by **lombok.Generated** annotation.
Before:
`@Generated
public class Test`

After:
`@Generated
class Test`

[...]

PR #6638, opened 2026-01-29
Fix parsing of user-written `@lombok.Generated` annotations

knutwannheden:
- Fix `isLombokGenerated()` incorrectly skipping user-written methods annotated with `@Generated` from Lombok, which caused `IllegalStateException` when Javadoc contained `${...}` placeholders
- Refactor `isLombokGenerated()` in all parser visitors (Java 8, 11, 17, 21, 25) to incorporate the source-position check directly, replacing the separate `lombokGeneratedInSource()` method. The method now returns `true` only when a tree element is Lombok-generated AND not present in the original source file.
- Extract the pure symbol check into a `hasLombokGeneratedSymbol()` static helper used internally
- Add reproduction test in `LombokTest` for #5490


rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11ParserVisitor.java
```diff
@@ -468,7 +468,7 @@ public J visitClass(ClassTree node, Space fmt) {
         List<Tree> membersMultiVariablesSeparated = new ArrayList<>(node.getMembers().size());
         for (Tree m : node.getMembers()) {
             // skip lombok generated trees
-            if (isLombokGenerated(m)) {
+            if (isLombokGenerated(m) && !lombokGeneratedInSource((JCTree) m)) {
```
timtebeek: It looks like we always call these two together; should we just make the source check part of the `isLombokGenerated` method then? Otherwise it might make sense to rename that to `hasLombokGeneratedAnnotation` or similar if we really want to distinguish between the two.

knutwannheden: Good call!

PR #6646, opened 2026-01-29
Update projectPoms in UpdateMavenModel so child modules see parent changes

dsgrieve:
## Summary

- Update `projectPoms` in `UpdateMavenModel` after resolving a POM so child modules see changes made to parents
- Add test cases for multi-module scenarios where managed dependencies are changed

## Problem

When `ChangeManagedDependencyGroupIdAndArtifactId` or similar recipes modify a managed dependency in a parent POM (A), and then `UpdateMavenModel` tries to update child modules (B), it downloads the parent POM from the repository instead of using the already-resolved (modified) version. This causes child modules to not see the changes made by the recipe.

The issue occurs in this code path:
- `ChangeManagedDependencyGroupIdAndArtifactId` calls `maybeUpdateModel()`
- `UpdateMavenModel.updateResult()` resolves the current POM, then iterates over modules
- When resolving child module B, `MavenPomDownloader.download()` looks up the parent in `projectPoms`
- But `projectPoms` still contains the OLD parent POM, not the resolved one with changes

## Solution

After `resolutionResult.getPom().resolve()` completes, update the `projectPoms` map with the newly resolved `Pom`. This ensures that when child modules resolve their parent, they see the updated version.

```java
// Update projectPoms with the resolved Pom so child modules see changes made to parents
Path sourcePath = resolved.getRequested().getSourcePath();
if (sourcePath != null) {
    projectPoms.put(sourcePath, resolved.getRequested());
}
```

## Test plan

- [x] New test `childModuleInheritsVersionPropertyChangeFromParent` passes
- [x] New test `childModuleWithManagedDependencySeesParentChange` passes  
- [x] Full `ChangeManagedDependencyGroupIdAndArtifactIdTest` suite passes
- [x] Full `ChangeDependencyGroupIdAndArtifactIdTest` suite passes

As a spot check against regressions in usages of `maybeUpdateModel()`, I ran a sampling of unit tests for recipes external to `rewrite-maven` that use `maybeUpdateModel()` -- all of which passed. 

Fixes moderneinc/customer-requests#1522


rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java
```diff
@@ -171,6 +171,13 @@ private MavenResolutionResult updateResult(ExecutionContext ctx, MavenResolution
         AtomicReference<MavenDownloadingExceptions> exceptions = new AtomicReference<>();
         try {
             ResolvedPom resolved = resolutionResult.getPom().resolve(ctx, downloader);
+
+            // Update projectPoms with the resolved Pom so child modules see changes made to parents
+            Path sourcePath = resolved.getRequested().getSourcePath();
+            if (sourcePath != null) {
+                projectPoms.put(sourcePath, resolved.getRequested());
+            }
+
```
steve-aom-elliott, line 185: I get that this can pass downwards the information, but I don't believe it's actually updating the `projectPoms` on the `MavenResolutionResult` that we're trying to return from this method.

dsgrieve: Updating the projectPoms Map makes parent POM changes visible to child modules during the resolution process within the same call stack. This ensures child modules don't download outdated parent POMs from the repository.

timtebeek:
@dsgrieve can you resolve the conflicts here?

dsgrieve:
> @dsgrieve can you resolve the conflicts here?
@timtebeek After resolving the conflict, I have a unit test failure. Converted this to a draft PR pending resolution of the failure.

rewrite-maven/src/main/java/org/openrewrite/maven/UpdateMavenModel.java
```diff
@@ -171,6 +171,18 @@ private MavenResolutionResult updateResult(ExecutionContext ctx, MavenResolution
...
+            // Update projectPoms with the resolved Pom so child modules see changes made to parents.
+            // This updates the Map that was passed in, making the change visible to:
+            // 1. The MavenPomDownloader created above (which holds a reference to this Map)
+            // 2. Recursive updateResult calls for child modules (which receive this Map as a parameter)
+            // Note: The returned MavenResolutionResult doesn't store this Map directly; it will
+            // compute projectPoms dynamically from its updated tree structure when getProjectPoms() is called.
+            Path sourcePath = resolved.getRequested().getSourcePath();
+            if (sourcePath != null) {
+                projectPoms.put(sourcePath, resolved.getRequested());
```
timtebeek, line 183: The unit test below also passes when I comment out this line, making me doubt the solution. Let's close this effort for now until we can reproduce the issue in a unit test.

PR #6649, opened 2026-01-30
Add recipes to find end-of-life & normalize Docker base images

timtebeek:
## Summary

- Adds `FindEndOfLifeDockerImages` recipe that detects Docker base images that have reached end-of-life
- Reports findings in a data table with image details, EOL dates, and suggested replacements
- Includes recipe in `DockerSecurityBestPractices` composite recipe

## Problem

Using end-of-life Docker base images poses security risks as they no longer receive security patches. Users need a way to identify which Dockerfiles use EOL images so they can upgrade to supported versions.

## Solution

Created a search recipe that:
- Detects common EOL images: Debian (jessie, stretch, buster), Ubuntu (14.04-20.04), Alpine (3.14-3.17), Python (3.7, 3.8), Node.js (14, 16, 18)
- Uses glob patterns to match image variants (e.g., `buster-slim`, `3.7-alpine`)
- Reports findings in `EolDockerImages` data table
- Marks Dockerfile FROM instructions with SearchResult containing EOL date and suggested replacement
- Optional `includeApproaching` parameter to flag images approaching EOL within 6 months

## Test plan

- [x] Existing tests pass
- [x] 22 new tests covering all EOL image types and edge cases

### DIFF
--- a/rewrite-docker/src/test/java/org/openrewrite/docker/DockerBestPracticesTest.java
+++ b/rewrite-docker/src/test/java/org/openrewrite/docker/DockerBestPracticesTest.java
@@ -17,30 +17,16 @@ package org.openrewrite.docker;
 
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
-import org.openrewrite.config.Environment;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.docker.Assertions.docker;
 
 class DockerBestPracticesTest implements RewriteTest {
 
     @Override
     public void defaults(RecipeSpec spec) {
-        spec.recipe(Environment.builder()
-                .scanRuntimeClasspath("org.openrewrite.docker")
-                .build()
-                .activateRecipes("org.openrewrite.docker.DockerBestPractices"));
-    }
-
-    @Test
-    void recipeLoadsSuccessfully() {
-        assertThat(Environment.builder()
-                .scanRuntimeClasspath("org.openrewrite.docker")
-                .build()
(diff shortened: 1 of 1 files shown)


rewrite-docker/src/main/java/org/openrewrite/docker/search/EolImage.java
```diff
@@ -0,0 +1,121 @@
...
+    PYTHON_3_8("python", new String[]{"3.8", "3.8.*", "3.8-*"},
+            LocalDate.of(2024, 10, 31), "3.12 or 3.11"),
+
+    // Node.js releases - https://nodejs.org/en/about/releases/
+    NODE_14("node", new String[]{"14", "14.*", "14-*"},
+            LocalDate.of(2023, 4, 30), "22 or 20"),
+    NODE_16("node", new String[]{"16", "16.*", "16-*"},
+            LocalDate.of(2024, 4, 30), "22 or 20"),
+    NODE_18("node", new String[]{"18", "18.*", "18-*"},
+            LocalDate.of(2025, 4, 30), "22 or 20");
```
timtebeek: Move the data on EOL images into a classpath resource.

rewrite-docker/src/main/resources/eol-images.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+- imageName: node
+  tagPatterns: ["16", "16.*", "16-*"]
+  eolDate: "2024-04-30"
+  suggestedReplacement: 22 or 20
+
+- imageName: node
+  tagPatterns: ["18", "18.*", "18-*"]
+  eolDate: "2025-04-30"
+  suggestedReplacement: 22 or 20
```
shanman190: Maybe node 24 or 22 instead?

timtebeek: Yeah definitely, this part so far was Claude. :)

PR #6650, opened 2026-01-30
Fix AddDependency duplicating dependencies in applied scripts

### LINKED ISSUE #6610: org.openrewrite.gradle.AddDependency adds duplicate dependencies for multiple build scripts

I am using

- OpenRewrite v8.72.0
- Gradle plugin v7.25.0

I'm trying to add dependencies to a Gradle project (Gradle 9.0.0) using the _org.openrewrite.gradle.AddDependency_ recipe. If the main _build.gradle_ file includes another build script using e.g.

```
// included dependencies from file
apply from: "dependencies.gradle"
```

dependencies will be added to each build file (build.gradle and whatever file is included) resulting in duplicate dependency definitions.

I have attached a simple reproducer project. Run `gradlew rewriteDryRun` and inspect the *.gradle files. Each file has a dependency block with redundant dependencies added.

- Only one dependency should be added to the build. AFAIK, Gradle currently doesn't support mutating build scripts through its API, see https://github.com/gradle/gradle/issues/35283 but maybe OpenRewrite can only add the dependency to the first build script / dependency block it encounters. Alternatively, a parameter could be added to specify the build script to which dependencies should be added.

AnastasiiaDesiateryk:
## What's changed?
* `AddDependency` now mutates **only** `build.gradle` / `build.gradle.kts` and skips applied `*.gradle` scripts loaded via `apply from:` to prevent duplicate dependency declarations.

## What's your motivation?
- * Fixes #6610 — when a build uses `apply from: "dependencies.gradle"`, the recipe was adding the dependency to **both** files, creating duplicates.

* The early guard that restricts changes to primary build scripts only.
* The new regression test `doNotAddDependencyToAppliedScripts()`.

* Alternative would be a recipe option to specify the target build script, but simplest correct behavior is to only mutate the primary build script.

## Any additional context
* Test: `./gradlew :rewrite-gradle:test --tests "*AddDependencyTest*" --rerun-tasks`

### Checklist
* [x] I've added unit tests to cover both positive and negative cases
* [x] I've read and applied the recipe conventions and best practices
* [x] I've used the IntelliJ IDEA auto-formatter on affected files


rewrite-gradle/src/test/java/org/openrewrite/gradle/AddDependencyTest.java
```diff
@@ -1872,6 +1761,54 @@ fun githubPackages(): Pair<String, String> {
...
+                }
+                """
+            ),
+            buildGradle(
+              //language=groovy
+              """
+                dependencies {
+                }
+                """,
+              s -> s.path("dependencies.gradle")
```
timtebeek, line 1916: Thanks for the help here @AnastasiiaDesiateryk !

For @loetifuss I wonder if you had intended for the dependency to only be added to `dependencies.gradle` instead. Any quick context there?

AnastasiiaDesiateryk: @timtebeek @loetifuss Thanks for the review! The test is intentionally asserting the default behavior from #6610: only the primary `build.gradle` is mutated, and the applied script remains unchanged. Applied scripts like `dependencies.gradle` are commonly shared/reused, so mutating them by default caused duplicate dependencies and surprising side effects (the same dependency ended up in both files). In this setup, `dependencies.gradle` is provided as an additional source with only a “before”, which implies “after = unchanged”. If adding directly to an applied script is desired, I think that should be an explicit opt-in (e.g. a recipe option or separate recipe), rather than the default behavior.
>
> Feel free to reach me if you have any further questions or suggestions.

Jenson3210: I personally agree with @AnastasiiaDesiateryk here. 
If we can be 100% sure the `apply from:` is only present in 1 project, one could argue that we should add it in this dependencies file. However, getting that knowledge, I think that would introduce lots of complexity here (and make all impacted recipes ScanningRecipes) and OR it would be harder to "do no harm" as we might bump/add/remove dependencies to/from other projects also.

Jenson3210: Looking at our other tests throughout rewrite, I do see that we're modifying the `dependencies.gradle` script instead of the `build.gradle` when running any of:
- RemoveRedundantDependencyVersions
- DependencyUseMapNotation
- DependencyUseStringNotation
- ChangeDependencyArtifactId
- ChangeDependencyClassifier
- ChangeDependencyGroupId
- ChangeDependencyConfiguration
- ChangeDependencyExtension
- UpgradeDependencyVersion

All of these are modifying/removing existing code/dependencies
The ones where we do not have tests yet and at the moment add new code are indeed AddDependency and UpgradeTransitiveDependencyVersion.
Given that all our other recipes impact multiple projects, one could say we should also add the dependency to them. 
However, generating the code in 2 gradle files is not what we want to do. 

As both AddDependency and UpgradeTransitiveDependencyVersion might result in a new dependencies/constraints/resolutionStrategy block we can either: 
- just make the change only in the `build.gradle` (current implementation)
- since both are scanning recipes already, we could augment their behavior to become aware of `applied` scripts and only add the code if there is a dependencies/constraints/resolutionStrategy block present in the build.gradle file itself OR if none of the applied scripts contains a dependencies/constraints/resolutionStrategy block

@timtebeek any preference here?I think given that the others modify the applied script, we have to go that way also for consistency (though my personal preference would be to have had everything in the build.gradle files -> ofc, removing/changing dependencies is not really possible (unless resolutionstrategy)).

timtebeek: Agree it might be hard to update applied build scripts when adding dependencies; I think the approach we have here should suffice until we get reports that we'd need something more complicated. The problem with `apply` is that the same file might be reused, or not applied, or conditionally applied, ... all cases that are hard to account for.

loetifuss: > Thanks for the help here @AnastasiiaDesiateryk !
> 
> For @loetifuss I wonder if you had intended for the dependency to only be added to `dependencies.gradle` instead. Any quick context there?

Yes, in my case due to technical reasons the Gradle dependencies are located in a separate file "dependencies.gradle" located in each subproject next to the default "build.gradle". So for this special case the fix won't solve the problem unless it would be configurable which script file is mutated by the recipe.

timtebeek:
Thanks all; we'll merge this as is for now, as we don't see an immediate way to be sure to add the dependency correctly to any applied `dependencies.gradle`, so then this should at least avoid the duplicates for the most common cases.

PR #6652, opened 2026-01-30
Fix vulnerability reports

Jenson3210:
**rewrite**
Added cleanReports to drop the reports in all files. Not usi…ng clean here as the suppressions file need to be in the build folder to locally run.

**rewrite-java**
Removed assertj-core to get plugin provided version in correct scope

**rewrite-java-test**
rewrite(-java)-test: pinned implementation version to 3.+ for now

**rewrite-xml**
Bump commons-text as the commit was from the time 1.11 was the latest. latest.release should work as used in other places of the project and this drops the CVE. This quick win mitigates a Medium risk CVE.

### DIFF
--- a/rewrite-java-test/build.gradle.kts
+++ b/rewrite-java-test/build.gradle.kts
@@ -7,6 +7,7 @@ recipeDependencies {
 }
 
 dependencies {
+    implementation("org.assertj:assertj-core:3.+") // CVE-2026-24400 in 4.0.0-M1 and no higher versions available
     implementation(project(":rewrite-java"))
     implementation(project(":rewrite-kotlin"))
     implementation(project(":rewrite-groovy"))
--- a/rewrite-test/build.gradle.kts
+++ b/rewrite-test/build.gradle.kts
@@ -9,6 +9,7 @@ dependencies {
     api("org.junit.jupiter:junit-jupiter-api")
     api("org.junit.jupiter:junit-jupiter-params")
 
+    implementation("org.assertj:assertj-core:3.+") // CVE-2026-24400 in 4.0.0-M1 and no higher versions available
     implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-csv")
     implementation("org.slf4j:slf4j-api:1.7.36")
     implementation("org.slf4j:slf4j-nop:1.7.36")


rewrite-xml/build.gradle.kts
```diff
@@ -33,7 +33,7 @@ dependencies {
     implementation("org.antlr:antlr4-runtime:4.13.2")
     implementation("com.github.ben-manes.caffeine:caffeine:2.+")
     implementation("io.micrometer:micrometer-core:1.9.+")
-    implementation("org.apache.commons:commons-text:1.11.+")
```
Jenson3210, line 36: `1.11.0` was at the time of the commit the latest version.

build.gradle.kts
```diff
@@ -22,3 +22,11 @@ allprojects {
     group = "org.openrewrite"
     description = "Eliminate tech-debt. Automatically."
 }
+
```
Jenson3210, line 25: Facilitates/documents locally running the vulnerabilities with updated results every run. If projects still contain the report, a new one is not generated.

rewrite-java-test/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ recipeDependencies {
 }
 
 dependencies {
-    implementation("org.assertj:assertj-core:latest.release")
+    implementation("org.assertj:assertj-core:3.+") // CVE-2026-24400 in 4.0.0-M1 and no higher versions available
```
MBoegers, line 10: in the FFVD we do this with a constraint to prevent even transitive introduction. Is this something that makes sense here?

build.gradle.kts
MBoegers: Is this openrewrite/rewrite specific or could this be useful in other artifacts as well? If so we have built logic where we can do this.

rewrite-java-test/build.gradle.kts
Jenson3210: i would make sense if this was not typically a test dependency. 
In this case there is a version coming in from build-gradle-plugin (4.0.0-M1) for tests (so we are up to date while running tests), but include in compileClassPath a version that is not vulnerable. 

In other words, our public API is not impacted by the CVE but our tests receive the latest/greatest (though in this case this one is older). 

Screenshot 2026-01-30 at 10 56 18
Screenshot 2026-01-30 at 10 55 36

Jenson3210:
> Is this openrewrite/rewrite specific or could this be useful in other artifacts as well? If so we have built logic where we can do this.

It would be applicable on all multi-module projects. I did put it here as the command to run the checks is different for multi-module projects -> it is easier for the next person to figure out about this caching. Took me quite a while to notice that stuff was not being updated with latest scan results and my original fix was already working ^^

PR #6661, opened 2026-02-02
Add TraversalControl to DependencyTreeWalker for callback-controlled memoization

pstreef:
## Problem

`DependencyTreeWalker.containsDependency()` does an O(n) linear scan of the path for every node visited. Thread dumps from large monorepos show 150+ seconds CPU time stuck in this method with 25+ levels of recursive calls.

## Solution

- Replace O(n) `containsDependency()` linear scan with O(1) `HashSet.add()` check
- Add `TraversalControl` enum and `ControlledCallback` interface to let callbacks control traversal behavior

**New API (backward compatible):**

```java
// Existing Callback interface unchanged
DependencyTreeWalker.walk(root, matcher, (dep, path) -> {
    // process dependency
});

// New ControlledCallback for traversal control
DependencyTreeWalker.walk(root, matcher, (dep, path) -> {
    // process dependency
    return TraversalControl.SKIP_SAME_COORDINATES; // skip future visits to same coords
});
```

**TraversalControl options:**
- `CONTINUE` - proceed normally
- `HALT` - stop all traversal immediately
- `SKIP_SAME_COORDINATES` - skip future visits to same group:artifact (enables memoization)

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/cleanup/UnnecessaryParenthesesTest.java
@@ -932,6 +932,63 @@ class UnnecessaryParenthesesTest implements RewriteTest {
         );
     }
 
+    @Test
+    void unwrapUnaryInAssignment() {
+        rewriteRun(
+          unnecessaryParentheses(style -> style.withAssign(true)),
+          java(
+            """
+              class Test {
+                  void test() {
+                      int i = 1;
+                      int a = (++i);
+                      int b = (i++);
+                      int c = (--i);
+                      int d = (i--);
+                  }
+              }
+              """,
+            """
+              class Test {
+                  void test() {
+                      int i = 1;
+                      int a = ++i;
+                      int b = i++;
+                      int c = --i;
+                      int d = i--;
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void unwrapUnaryInMethodArgument() {
+        rewriteRun(
+          java(
+            """
+              class Test {
+                  void test() {
+                      int i = 1;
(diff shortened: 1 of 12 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/graph/DependencyTreeWalker.java
```diff
@@ -54,61 +55,107 @@ public interface Callback {
...
+        GroupArtifact ga = new GroupArtifact(dependency.getGroupId(), dependency.getArtifactId());
+
+        // Cycle detection - O(1) check if this dependency is already in the current path
+        if (!inPath.add(ga)) {
+            return;
+        }
+
+        // Global memoization - skip already-visited subtrees (only in walkFirstPath)
+        if (visited != null && !visited.add(ga)) {
+            inPath.remove(ga);
```
sambsnyd, line 179: Shannon and I tested this.
We found that the `inPath.add(ga)` check didn't really make a difference. 
The global `visited` map check _does_ make a difference, but we don't want to add a separate `walkFirstPath` override, or add the global memoization in all cases.

However I could accept the `Callback` interface returning an `enum` which could dictate whether to continue as normal, halt all traversal, or skip future encounters of the same coordinates. `DependencyInsight` could be configured to use the "skip" option to limit how much processing we do.

pstreef: updated!

sambsnyd (approved):

pstreef:
@sambsnyd @shanman190 I'm not sure if we really still need this change or not after your changes. If HasDependency no longer uses DependencyInsight perhaps this change doesnt add much?

shanman190:
Yeah, I don't think we need this anymore for now.

PR #6666, opened 2026-02-03
Break cycles when parsing Gradle projects.

sambsnyd:
Gradle doesn't allow cyclic dependencies... usually. But we've found at least one case where, with some creative usage of composite builds, it is possible.

This breaks cycles in the dependency representation, preserving the evidence in the GradleConfiguration's exception fields.

### DIFF
--- a/rewrite-gradle-tooling-model/model/src/test/java/org/openrewrite/gradle/marker/GradleProjectBuilderTest.java
+++ b/rewrite-gradle-tooling-model/model/src/test/java/org/openrewrite/gradle/marker/GradleProjectBuilderTest.java
@@ -24,6 +24,7 @@ import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
 
 import java.util.*;
 
+import static java.util.Collections.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
 class GradleProjectBuilderTest {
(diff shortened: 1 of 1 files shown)


rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/marker/GradleProjectBuilder.java
```diff
@@ -394,13 +412,16 @@ private static List<org.openrewrite.maven.tree.ResolvedDependency> resolved(
                     org.openrewrite.maven.tree.ResolvedDependency resolvedDependency = resolvedCache.get(resolvedGav);
                     if (resolvedDependency == null) {
                         org.openrewrite.maven.tree.Dependency requested = gaToRequested.getOrDefault(ga, dependency(resolved));
+                        // Track the traversal path to detect cycles at any depth
+                        List<GroupArtifactVersion> ancestorPath = new ArrayList<>();
```
shanman190, line 416: A `HashSet` would probably be a bit quicker and gives the same uniqueness constraint while also having the helpful `boolean add(T)` method.

I'm also not sure if we actually need to track version, but it won't hurt. Gradle's dependency tree puts the max version everywhere by the time we get to traverse it.

sambsnyd: `HashSet` loses order so we couldn't reproduce the A -> B -> C -> A as easily for meaningful error message. Could use `LinkedHashSet`, though

shanman190: I see, I see. It probably isn't worth it then in that case.

PR #6669, opened 2026-02-03
Support multi-component literal dependency declarations

MBoegers:
## Summary
- Add support for Gradle dependency declarations using separate string arguments like `implementation("group", "artifact", "version")` in the `GradleDependency` trait
- Previously only colon-separated (`"g:a:v"`), map notation, GString, and template formats were handled
- The fix updates parsing, accessors (`getDeclaredGroupId/ArtifactId/Version`), and mutators (`withDeclaredGroupId/ArtifactId/Version`, `removeVersion`)

- Fixes moderneinc/customer-requests#1791

## Test plan
- [x] Added failing test `multiComponentLiteralNotation` to `UpgradeDependencyVersionTest`
- [x] Test confirms version upgrade works with multi-component form
- [x] Full `UpgradeDependencyVersionTest` suite passes
- [x] Full `GradleDependencyTest` suite passes

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
@@ -16,27 +16,14 @@
 package org.openrewrite.gradle.trait;
 
 import org.junit.jupiter.api.Disabled;
-import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.DocumentExample;
-import org.openrewrite.java.tree.Expression;
-import org.openrewrite.java.tree.J;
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/trait/GradleDependencyTest.java
```diff
@@ -345,4 +359,194 @@ void enforcedPlatform() {
...
+              dependencies {
+                  /*~~(com.google.guava:guava:)~~>*/implementation("com.google.guava", "guava")
+              }
+              """
+          )
+        );
+    }
+
+    @Nested
+    class IsMultiComponentLiteralsTest {
```
shanman190: I'm not sure if it's super helpful to keep these tests and if not kept, then the two new methods should be switched to private as well.

MBoegers:
You mean checking to support for this pattern? Yes agree.

shanman190:
Yeah, updating the `DependencyUseStringNotation` recipe to detect this positional parameter pattern, then shifting it to a single String form.

shanman190 (approved):

PR #6670, opened 2026-02-03
Fix DeleteProperty losing newlines when deleting entries with block scalars

Jenson3210:
## Summary
- Fixed bug where deleting YAML entries with block scalar values (`>-` or `|`) caused loss of newlines between entries
- Block scalar content in YAML includes trailing newlines that serve as separators; these were being lost when entries were deleted
- Added proper newline restoration for both sibling entries and parent mapping entries

## Test plan
- [x] Added test `globPatternWithFoldedBlockScalar` that reproduces the issue
- [x] All existing DeletePropertyKeyTest tests continue to pass (22 tests)
- [x] Full rewrite-yaml test suite passes

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
@@ -112,12 +112,35 @@ public class DeleteProperty extends Recipe {
             @Override
             public Yaml.Sequence visitSequence(Yaml.Sequence sequence, ExecutionContext ctx) {
                 Yaml.Sequence s = super.visitSequence(sequence, ctx);
+                boolean childModified = (s != sequence);
                 List<Yaml.Sequence.Entry> entries = s.getEntries();
                 if (entries.isEmpty()) {
                     return s;
                 }
 
+                boolean changed = false;
                 entries = ListUtils.map(entries, entry -> ToBeRemoved.hasMarker(entry) ? null : entry);
+                if (entries.size() != s.getEntries().size()) {
+                    changed = true;
+                }
+
+                if ((changed || childModified) && s.getOpeningBracketPrefix() == null) {
+                    List<Yaml.Sequence.Entry> fixedEntries = null;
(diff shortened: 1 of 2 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
```diff
@@ -191,6 +216,32 @@ private static boolean containsOnlyWhitespace(@Nullable String str) {
         return true;
     }
 
+    private static boolean startsWithNewline(@Nullable String str) {
+        return str != null && !str.isEmpty() && str.charAt(0) == '\n';
```
steve-aom-elliott, line 242: Would this at all be a potential issue on Windows, or are they standardized already by the time they make it here?

Jenson3210: Should be standardized normally. I have always been using \n so far without issues. So should be okay!

steve-aom-elliott:
Probably worth checking on https://github.com/openrewrite/rewrite-quarkus/blob/de55225c1ab806bec823aebe02c28743ea565baf/src/test/java/org/openrewrite/quarkus/DeleteQuarkusPropertyTest.java#L184  As well to ensure those are still fine after these changes. I had dropped leading newlines from that test based on the behaviours being adjust in this PR

rewrite-yaml/src/test/java/org/openrewrite/yaml/DeletePropertyKeyTest.java
```diff
@@ -408,4 +408,32 @@ void globPatternWithExactMatch() {
...
+              """,
+            """
+              acme:
+                my-project:
+                  person:
+                    first-name: John
+                  employee:
+                    firstName: Jane
+              """
+          )
```
steve-aom-elliott, line 493: Might be worth an example where the leading key is the one removed instead, as that was closer to the situation encountered before wherein it was leaving empty space above.

PR #6678, opened 2026-02-04
C# Support

jkschneider:
## Summary

Add C# language support based on RewriteRpc.

### DIFF
--- a/rewrite-csharp/build.gradle.kts
+++ b/rewrite-csharp/build.gradle.kts
@@ -51,7 +51,7 @@ fun findDotnet(): String {
             // Command not found, try next
         }
     }
-    throw GradleException(".NET SDK not found. Please install .NET 8.0+ SDK and ensure 'dotnet' is on your PATH.")
+    throw GradleException(".NET SDK not found. Please install .NET 10.0+ SDK and ensure 'dotnet' is on your PATH.")
 }
 
 val csharpBuild by tasks.registering(Exec::class) {
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
@@ -279,7 +279,6 @@ public class CSharpRewriteRpc extends RewriteRpc {
                         "run",
                         "--project", entry.toAbsolutePath().normalize().toString(),
                         "--framework", "net10.0",
-                        "--no-build",
                         log == null ? null : "--log-file=" + log.toAbsolutePath().normalize(),
                         traceRpcMessages ? "--trace-rpc-messages" : null
                 );


macsux:
The modelling for preprocessor directives wouldn't work. Preprocessor directives do not follow syntax tree layout in any meaningful way - they exist at text block level even before tokenization happens. You can't assume that anything under their "branches" is a valid c# code/statement. They can chop c# source code at ANY point. Concrete example from production code:
https://github.com/macsux/cf-buildpack-template/blob/main/src/MyBuildpack/MyBuildpack.cs#L8-L13

Those directives could have just as easily sliced the code on just the base class which is just a single token (like a modifier) and not a valid statement on it's own. They could be placed around a keyword that's only to be included under specific scenario. 

This is totally valid
```csharp
public
#if (SOMETHING)
record
#else
class
#endif 
MyObject
{ }
```

Proprocessors are heavily used when targeting multiple versions of .NET in single source repo by providing polyfills for APIs that were introduced in later versions. It also has potential affect on type attestation resolution. Example:

```csharp
#nullable enable
using System;

namespace Polyfills
{
#if !NET6_0_OR_GREATER
    internal static class ArgumentNullException
    {
        // Polyfill: this symbol exists only on older targets.
        public static void ThrowIfNull(object? argument, string? paramName = null)
        {
            if (argument is null)
                throw new System.ArgumentNullException(paramName);
        }
    }
#endif
}

namespace Demo
{
    internal sealed class Widget
    {
        public Widget(string name)
        {
            // Binding changes:
            // - NET6_0_OR_GREATER: binds to System.ArgumentNullException.ThrowIfNull
            // - older targets: binds to Polyfills.ArgumentNullException.ThrowIfNull
            ArgumentNullException.ThrowIfNull(name, nameof(name));
        }
    }
}
```

macsux:
Using records for modelling has severe performance downsides. They do field equality instead of referential, meaning the entire tree will be walked and every property compared when doing a `.Equals` or `==`. Also calling a `with` on records ALWAYS clones the object, even if the values are the same. [Proposal to change that in c# was made by me last year and rejected.](https://github.com/dotnet/csharplang/discussions/8506)

macsux:
J model does not allow modelling for attributes applied to genetic arguments (and I believe something else) hence the reason for cs specific versions on both class and method. This was discussed ong ago and without altering J there was no way to account for this syntax.

jkschneider:
@macsux Thanks for the feedback on preprocessor directives. We've implemented a multi-parse approach that handles this:

```
Original source (with directives)
│
├─ ExtractSymbols() → {SOMETHING}
│
├─ GenerateUniquePermutations() → one clean source per symbol combination
│   │
│   ├─ Permutation 0 (SOMETHING=true):     Permutation 1 (SOMETHING=false):
│   │   "public\n"                          "public\n"
│   │   "\n"          ← #if removed         "\n"
│   │   "record\n"    ← active branch       "\n"          ← inactive → blank
│   │   "\n"          ← #else removed       "\n"
│   │   "\n"          ← inactive → blank    "class\n"     ← active branch
│   │   "\n"          ← #endif removed      "\n"
│   │   "MyObject\n"                        "MyObject\n"
│   │   "{ }\n"                             "{ }\n"
│   │
│   └─ Each is valid C# → parsed by Roslyn into a full Cs.CompilationUnit
│
└─ Assembled into:

    Cs.CompilationUnit (outer shell, Eof=Empty)
    └─ Members:
        └─ Cs.ConditionalDirective
            ├─ DirectiveLines:          ← metadata from original source
            │   ├─ Line 1: "#if (SOMETHING)"    Kind=If,   Group=0, ActiveBranch=0
            │   ├─ Line 3: "#else"              Kind=Else, Group=0, ActiveBranch=1
            │   └─ Line 5: "#endif"             Kind=Endif,Group=0, ActiveBranch=-1
            │
            └─ Branches: List<JRightPadded<Cs.CompilationUnit>>
                ├─ [0] CompilationUnit (SOMETHING=true)  ← has ConditionalBranchMarker
                └─ [1] CompilationUnit (SOMETHING=false) ← has ConditionalBranchMarker
```

The printer reconstructs the original by line-level interleaving: since each clean source replaces directive/excluded lines with blanks, all branches have the same line count. A stack tracks which branch is active at each line, emitting directive text at directive lines and branch content otherwise.

This handles all the cases you mentioned — keyword splitting, base class switching, polyfill classes — because each permutation is independently valid C# that Roslyn can fully parse and type-attribute.

jkschneider:
@macsux This is addressed by `Cs.ConstrainedTypeParameter`, which is stored in `J.TypeParameter.bounds[0]`. It carries `attributeLists` (for `[MaybeNull] T` etc.), `variance` (`in`/`out`), and `whereConstraint`/`constraints` for where clauses. So `J.ClassDeclaration` works directly for C# classes without needing a `CsClassDeclaration`.

`Cs.MethodDeclaration` still exists, but for other reasons (explicit interface specifiers like `void IFoo.Bar()`, C# attribute lists on the method itself) — its `typeParameters` field uses the same `ConstrainedTypeParameter` pattern for generic param attributes.

macsux:
This completely detaches directives from where where they are used, with line numbers being the joining criteria. How would this work for recipe writers though? Lines can and will shift after transformation, throwing all that into misalignment. There is no way for recipe author to track line numbers in any reasonably simple way. This feels like something that can be super easy for recipe author to mess up.

steve-aom-elliott:
~~@jkschneider @macsux I've pushed up~~
~~- https://github.com/openrewrite/rewrite/tree/steve/csharp-support~~

~~which is the rebased and patched version of the this branch. It drops about 20 commits that were already present on `main`~~

Issue #6644, opened 2026-01-29
Java25Parser compiled with Java 25 breaks Java 24 runtime compatibility


kdvolder:
# Java25Parser compiled with Java 25 breaks Java 24 runtime compatibility

## What version of OpenRewrite are you using?

- OpenRewrite core: **8.71.0** (also reproduced with 8.73.0-SNAPSHOT)
- Maven plugin: **6.27.1** 
- Gradle plugin: **7.19.0**
- Bug **NOT present** in OpenRewrite 8.65.0
- Bug introduced **sometime between 8.65.0 and 8.71.0** (intermediate versions not tested)

## How are you running OpenRewrite?

I am using both the Maven and Gradle plugins. The issue affects both single-module and multi-module projects.

**Maven configuration:**
```xml
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>6.27.1</version>
  <configuration>
    <activeRecipes>
      <recipe>com.vmware.tanzu.spring.recipes.boot40.UpgradeSpringBoot_4_0</recipe>
    </activeRecipes>
  </configuration>
  <dependencies>
    <dependency>
      <groupId>com.vmware.tanzu.spring.recipes</groupId>
      <artifactId>spring-boot-4-upgrade-recipes</artifactId>
      <version>1.5.5-SNAPSHOT</version>
    </dependency>
  </dependencies>
</plugin>
```

**Gradle configuration:**
```groovy
plugins {
    id("org.openrewrite.rewrite") version "7.19.0"
}

rewrite {
    activeRecipe("com.vmware.tanzu.spring.recipes.boot40.UpgradeSpringBoot_4_0")
}

dependencies {
    rewrite("com.vmware.tanzu.spring.recipes:spring-boot-4-upgrade-recipes:1.5.5-SNAPSHOT")
}
```

**Reproduction project:** https://github.com/vmware-tanzu/spring-upgrade-recipes/tree/main/bug-reports/java25parser-gradle

## What is the smallest, simplest way to reproduce the problem?

The issue occurs when running **any** OpenRewrite recipe on **Java 24 runtime**. Even a minimal project triggers the error:

**build.gradle:**
```groovy
plugins {
    id 'java'
    id 'org.openrewrite.rewrite' version '7.19.0'
}

rewrite {
    activeRecipe("org.openrewrite.java.format.AutoFormat")
}
```

**Any Java file:**
```java
[...]

PR #6682, opened 2026-02-05
Fix Java 24 runtime compatibility by falling back to Java 21 parser

kdvolder:
- Closes https://github.com/openrewrite/rewrite/issues/6644


kdvolder:
@timtebeek You asked me to tag you on this PR.

This is essentially a very simple one line change to make the condition for loading Java25parser reflect the fact that it is compiled on Java25, so should not be classloaded on anything lower than that.

timtebeek:
Hey thanks yes I'm abroad for now, but will check when I have access to my laptop again.

rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
```diff
@@ -441,21 +441,24 @@ class JdkParserBuilderCache {
...
-            // That means Java 23 and above need the Java 25 parser, whereas Java 22 and below need the Java 21 parser
-            if (version > 22) {
+            // Parsers are compiled only for their matching Java version, given their use of internal/unstable APIs.
+            // Language features introduced in non-LTS versions are only supported from the next LTS version onwards.
+            // We recommend running on LTS versions wherever possible; non-LTS versions may struggle given the above.
+            // e.g.: Java 23+ changed the return type of `DocCommentTable.getCommentTree` from `DCDocComment` to `DocCommentTree`
+
+            if (version >= 25) {
                 supplier = tryCreateBuilderSupplier("org.openrewrite.java.Java25Parser");
             }
```
timtebeek, line 451: Both [the Maven plugin](https://github.com/openrewrite/rewrite-maven-plugin/blob/fe36727ed80cd0980e53ed872d174f9aa0a6533a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L204-L206) and [the Gradle plugin](https://github.com/openrewrite/rewrite-gradle-plugin/blob/eafd7670d1ba75cce8bd1d0be3ce9b5790c0bba4/plugin/src/main/java/org/openrewrite/gradle/isolated/DefaultProjectParser.java#L957-L961) use `JavaParser.fromJavaVersion`, which underneath uses these lines just out of view:
https://github.com/openrewrite/rewrite/blob/f3784245d9a54114415302e4abfa2828800fe51d/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L435-L440

That to me then seems like we should only see a behavior change on non-LTS versions to there pick the lower parser version, where previously we had seen that `UnsupportedClassVersionError`.

timtebeek: One alternative I'd considered is weaving that `UnsupportedClassVersionError` into the catch blocks seen here:
https://github.com/openrewrite/rewrite/blob/f3784245d9a54114415302e4abfa2828800fe51d/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L479-L494

Not sure if that was intentionally left out of the `catch (ClassNotFoundException | NoSuchMethodException e)`, but it might fit in there as well.

PR #6687, opened 2026-02-05
fix an edge case in `RemoveUnusedProperties` when there is no parent in filtering

nmck257:
… throw if a singular/root pom (no parent in the in-memory relative path) included resource filtering

I saw this issue when running via OSS Maven plugin. The new unit test reproduced the error, although (minor caveat) there's a little bit of abstraction from reality here. The problem stems from the exact value of the `SourcePath` on the pom, and in the unit test we're explicitly setting the `SourcePath` value instead of relying on the real project parser (looking at a real filesystem) to decide what the value should be (ie relative vs absolute).

The source code change is pretty trivial, though.
(Actually, replacing `getParent().resolve(xyz)` with `resolveSibling(xyz)` might be a recipe-worthy strict improvement :) )

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveUnusedPropertiesTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/RemoveUnusedPropertiesTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.DocumentExample;
+import org.openrewrite.Issue;
 import org.openrewrite.test.RecipeSpec;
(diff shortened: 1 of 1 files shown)


timtebeek:
And indeed makes for a good potential recipe as well; here's the same as a Refaster style recipe:
```java
  static final class PathResolveSiblingPath {
    @BeforeTemplate
    Path before(Path path, Path other) {
      return path.getParent().resolve(other);
    }

    @AfterTemplate
    Path after(Path path, Path other) {
      return path.resolveSibling(other);
    }
  }

  static final class PathResolveSiblingString {
    @BeforeTemplate
    Path before(Path path, String other) {
      return path.getParent().resolve(other);
    }

    @AfterTemplate
    Path after(Path path, String other) {
      return path.resolveSibling(other);
    }
  }
```

PR #6691, opened 2026-02-06
Preserve implicitly defined version properties in Maven recipes

Jenson3210:
## Summary
- Fixes an issue where recipes like `DependencyVulnerabilityCheck`, `UpgradeDependencyVersion`, and `ChangeDependencyGroupIdAndArtifactId` would incorrectly replace implicitly defined version properties (like `${project.parent.version}`, `${project.version}`, `${revision}`) with hardcoded version values
- These properties represent intentional links to parent/project versions that should be preserved to maintain the relationship between child and parent module versions in multimodule Maven projects

## Changes
1. Added `isImplicitlyDefinedVersionProperty()` method to `MavenVisitor` to check if a value is an implicitly defined property
2. Updated `changeChildTagValue()` in `MavenVisitor` to skip changes when the old value is an implicitly defined version property
3. Added checks in `ChangeDependencyGroupIdAndArtifactId` and `ChangeManagedDependencyGroupIdAndArtifactId` where they directly use `ChangeTagValueVisitor`

## Test plan
- Added test `shouldPreserveProjectParentVersionPropertyInDependencyVersion` that verifies `${project.parent.version}` is preserved when upgrading dependency versions
- Added test `shouldPreserveProjectVersionPropertyInDependencyVersion` that verifies `${project.version}` is preserved
- Added test `shouldPreserveRevisionPropertyInDependencyVersion` that verifies `${revision}` is preserved
- All existing Maven tests pass

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
@@ -3130,10 +3130,6 @@ class ChangeDependencyGroupIdAndArtifactIdTest implements RewriteTest {
 
     @Test
     void shouldNotChangeDependencyWithImplicitlyDefinedVersionProperty() {
-        // When the dependency version uses an implicitly defined property like ${project.parent.version},
-        // and newVersion is specified, no changes should be made at all. This avoids partially
-        // updating the dependency (groupId/artifactId) while leaving the version property unchanged,
-        // which would break the intentional version relationship the user established.
         rewriteRun(
           spec -> spec.recipe(new ChangeDependencyGroupIdAndArtifactId(
             "org.junit.jupiter", "junit-jupiter-api",
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactIdTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactIdTest.java
@@ -633,10 +633,6 @@ class ChangeManagedDependencyGroupIdAndArtifactIdTest implements RewriteTest {
 
     @Test
     void shouldNotChangeManagedDependencyWithImplicitlyDefinedVersionProperty() {
-        // When the managed dependency version uses an implicitly defined property like ${project.version},
-        // and newVersion is specified, no changes should be made at all. This avoids partially
-        // updating the dependency (groupId/artifactId) while leaving the version property unchanged,
-        // which would break the intentional version relationship the user established.
         rewriteRun(
           spec -> spec.recipe(new ChangeManagedDependencyGroupIdAndArtifactId(
             "org.junit.jupiter", "junit-jupiter-api",


rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java
```diff
@@ -333,6 +337,11 @@ protected Xml.Tag changeChildTagValue(Xml.Tag tag, String childTagName, @Nullabl
         return tag;
     }
 
+    @Contract("null -> false")
+    protected boolean isImplicitlyDefinedVersionProperty(@Nullable String value) {
```
steve-aom-elliott, line 342: Should this be used by `isProperty(..)` directly below this?

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -3126,4 +3127,64 @@ void exclusionUpdatedWhenDependencyGroupIdChanges() {
...
+                    </parent>
+                    <artifactId>my-child</artifactId>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.junit.jupiter</groupId>
+                            <artifactId>junit-jupiter-engine</artifactId>
+                            <version>${project.parent.version}</version>
+                        </dependency>
+                    </dependencies>
+                </project>
```
steve-aom-elliott: Hmm, I'm unsure if we can say this is always going to be safe to only change the `groupId` / `artifactId` and not the version number at all. If we're changing it to a completely unrelated `groupId` / `artifactId` or at least one that doesn't share version numbering, this is likely to break, no?

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
```diff
@@ -154,7 +154,9 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                                 }
                                 String resolvedNewVersion = resolveSemverVersion(ctx, newGroupId, resolvedArtifactId, getResolutionResult().getPom().getValue(versionTag.get().getValue().orElse(null)));
                                 String versionTagValue = t.getChildValue("version").orElse(null);
-                                if (versionTagValue == null || !safeVersionPlaceholdersToChange.contains(versionTagValue)) {
+                                if (isImplicitlyDefinedVersionProperty(versionTagValue)) {
+                                    // Implicitly defined version properties like ${project.parent.version} should not be changed
```
steve-aom-elliott: Perhaps this might read better if we either nested the `if`s here (adding `!` to the front of this) or added a `continue` or something. It just feels a bit odd to have an `if` that solely has a comment inside of it.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -218,7 +218,9 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                                 } else {
                                     // Otherwise, change the version to the new value.
                                     String versionTagValue = t.getChildValue("version").orElse(null);
-                                    if (versionTagValue == null || !safeVersionPlaceholdersToChange.contains(versionTagValue)) {
+                                    if (isImplicitlyDefinedVersionProperty(versionTagValue)) {
+                                        // Implicitly defined version properties like ${project.parent.version} should not be changed
```
steve-aom-elliott: Perhaps this might read better if we either nested the ifs here (adding ! to the front of this) or added a continue or something. It just feels a bit odd to have an if that solely has a comment inside of it.

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -630,4 +630,50 @@ void doesNotMakeChangeWhenChangingBomImportManagedDependency() {
...
+                  <groupId>com.mycompany</groupId>
+                  <artifactId>my-app</artifactId>
+                  <version>5.7.2</version>
+                  <dependencyManagement>
+                      <dependencies>
+                          <dependency>
+                              <groupId>org.junit.jupiter</groupId>
+                              <artifactId>junit-jupiter-engine</artifactId>
+                              <version>${project.version}</version>
+                          </dependency>
```
steve-aom-elliott: Similar to the test for the non-managed version of the recipe, I'm not sure that continuing to change the `groupId` and `artifactId` in the event we won't change the version is particularly safe or will lead to the desired behaviour.

rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
```diff
@@ -2537,4 +2537,100 @@ void upgradeMultipleManagedDependenciesWithWildcards() {
...
+                        <groupId>com.mycompany</groupId>
+                        <artifactId>my-parent</artifactId>
+                        <version>13.0</version>
+                    </parent>
+                    <artifactId>my-child</artifactId>
+                    <dependencies>
+                        <dependency>
+                            <groupId>com.google.guava</groupId>
+                            <artifactId>guava</artifactId>
+                            <version>${project.parent.version}</version>
```
steve-aom-elliott, line 2578: In comparison to the others, this seems at least safe as it won't make a change that could break resolution.

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
Jenson3210: agreed, let's not change

timtebeek: I'm going to agree with Steve's concern here; do we need to address these two out of three recipes for the issue reported? Or can we only change `UpgradeDependencyVersion`?

Ideally of course it's already rare that folks reuse special tags from this list, but still I wonder if this is needed:
https://github.com/openrewrite/rewrite/blob/e0d7e9aecf6d679bb109329b2aac397f97a42707/rewrite-maven/src/main/java/org/openrewrite/maven/MavenVisitor.java#L52-L61

Jenson3210: I updated the recipes to not make changes to any dependency field if the version is an IMPLICITLY_DEFINED_VERSION_PROPERTIES. 

You mean to ask about this list of fixed properties? This list already goes back quite some time. Any suggestions instead of this list here?

timtebeek: Agree with the changes you've made; the list inline was just for context, no specific questions there. :)

PR #6693, opened 2026-02-06
Mark parent tag in DependencyInsight when dependency is inherited

Jenson3210:
## Summary

- Extended `DependencyInsight` to mark the `<parent>` tag when matching dependencies are inherited from parent POMs
- Added logic to correctly handle cases where a target dependency is reachable through multiple paths
- Only marks parent for targets that aren't covered by any declared dependency in the current POM

## Problem

When Spring dependencies are declared in an external parent POM (not part of the project being analyzed), `DependencyInsight` couldn't find them because it only marked `<dependency>` tags. When there's no `<dependency>` tag in the child POM for a dependency inherited from the parent, nothing got marked, and preconditions like `IsLikelySpringFramework` failed.

This caused the `SpringToSpringBoot` recipe to skip processing when Spring dependencies came solely from parent inheritance.

## Solution

Modified `DependencyInsight.MarkIndividualDependency` to also check and mark `<parent>` tags:

1. When visiting a `<parent>` tag, collect all target dependencies that are NOT reachable through any declared `<dependency>` in the current POM
2. Mark the parent tag with those inherited-only targets
3. Handle the case where a target is reachable through multiple paths (e.g., `spring-core` via both `spring-boot-starter-web` and `spring-web`) - if ANY path goes through a declared dependency, don't mark the parent for that target

## Test plan

- [x] Existing tests pass
- [x] Added `findDependencyDeclaredInProjectParent` - parent declares dependency, child inherits → child's `<parent>` tag is marked
- [x] Added `doNotMarkParentTagWhenDependencyDeclaredInChild` - child declares the target directly → only `<dependency>` tag is marked
- [x] Added `doNotMarkParentTagWhenTransitiveDependencyComesFromChildDeclaredDependency` - child declares dependency that transitively brings target → only `<dependency>` tag is marked

- Fixes moderneinc/customer-requests#1803

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/search/DependencyInsightTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/search/DependencyInsightTest.java
@@ -315,8 +315,6 @@ class DependencyInsightTest implements RewriteTest {
     @Issue("https://github.com/moderneinc/customer-requests/issues/1803")
     @Test
     void findDependencyDeclaredInProjectParent() {
-        // When the parent (in the same project) declares a dependency and the child
-        // inherits it without redeclaring, the <parent> tag in the child should be marked.
         rewriteRun(
           spec -> spec.recipe(new DependencyInsight("com.google.guava", "guava", null, null, null)),
           mavenProject("parent",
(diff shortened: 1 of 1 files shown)


Jenson3210:
verified on my reference project that this now marks the parent tag -> change will be detected.

PR #6694, opened 2026-02-09
Correct string block indent

BoykoAlex:
String blocks autoformat is based on spaces rather than whatever is in the style.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
@@ -498,7 +498,8 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
                 JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
                 SourcePositionService positionService = sourceFile.service(SourcePositionService.class);
                 String content = literal.getValueSource().substring(4);
-                int currentIndent = StringUtils.minCommonIndentLevel(content, style.getTabSize());
+                String finalContent = content; //TODO remove this assignment together with the next evaluate invocation as it is no longer needed.
+                int currentIndent = evaluate(() -> StringUtils.minCommonIndentLevel(finalContent, style.getTabSize()), StringUtils.minCommonIndentLevel(content));
                 content = StringUtils.trimIndent(content);
                 String[] lines = content.split("\n", -1);
                 int indent = getCursor().getNearestMessage("lastIndent", 0);


BoykoAlex:
@timtebeek this is to support autoformat for text blocks where idents are not spaces. Looks like the code currently is biased towards spaces based indents. The indent for multiline string is parent AST node indent + 2 indents - this seesm to be the same to JDT and IntelliJ.

timtebeek (approved):
Thanks for the immediate suggested fix! @Jenson3210 is our expert here, but this looks alright to me already. I'll merge before tomorrow's release if I don't hear from him before then. If he sees this and agrees he can immediately merge.

BoykoAlex:
@timtebeek i see there is a test failure for aligned text blocks. Trying to figure it out... I don't know why we're making a check `!literal.getPrefix().getLastWhitespace().contains("\n")` - I'd align the string even if `"""` on the new line... Something blows up with the cursor is the source position service if I remove this check. If you @Jenson3210 and @timtebeek have ideas/opinions/directions for me, please don't be shy :-)

BoykoAlex:
Okay, I think i came up with a good fix for the alignment things. Should be good to go now

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -498,22 +498,29 @@ public J.Literal visitLiteral(J.Literal literal, P p) {
                 JavaSourceFile sourceFile = getCursor().firstEnclosing(JavaSourceFile.class);
                 SourcePositionService positionService = sourceFile.service(SourcePositionService.class);
                 String content = literal.getValueSource().substring(4);
-                int currentIndent = StringUtils.minCommonIndentLevel(content);
+                int currentIndent = StringUtils.minCommonIndentLevel(content, style.getTabSize());
```
Jenson3210: Nice! great additional support here.
⚠️ @timtebeek, `internal` class being changed here -> expect to see some NoSuchMethod Exceptions. 

To avoid, we could do `evaluate(() -> StringUtils.minCommonIndentLevel(content, style.getTabSize()), StringUtils.minCommonIndentLevel(content))` for now

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -498,22 +498,29 @@ public J.Literal visitLiteral(J.Literal literal, P p) {
...
                 content = StringUtils.trimIndent(content);
                 String[] lines = content.split("\n", -1);
-                int indent = getCursor().getNearestMessage("lastIndent", 0);
+                // get tree parent cursor rather than itself top get the indent, i.e. from the variable declaration node.
+                int indent = getCursor().getParentTreeCursor().getNearestMessage("lastIndent", 0);
```
Jenson3210: I do not think this is needed. The original code should work also for the unit test provided, does it not?
```suggestion
                int indent = getCursor().getNearestMessage("lastIndent", 0);
```

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -498,22 +498,29 @@ public J.Literal visitLiteral(J.Literal literal, P p) {
...
                 if (evaluate(() -> wrappingStyle.getTextBlocks().getAlignWhenMultiline(), false)) {
                     if (!literal.getPrefix().getLastWhitespace().contains("\n")) {
                         ColSpan colSpan = positionService.columnsOf(getCursor(), literal);
                         indent += colSpan.getStartColumn() - colSpan.getIndent() - 1; // since position is index-1-based
+                    } else {
+                        // Take the indent of the whole literal to align the rest of it.
+                        // Terminate the indent with a char that would signal start of the content.
+                        indent = StringUtils.minCommonIndentLevel(literal.getPrefix().getIndent() + "a", style.getTabSize());
```
Jenson3210:
```suggestion
```

If the original literal has a whiteline in front of the `"""` beginning, we can expect that that block start will be indented to the correct indentation -> we just have to align the block content to the indent of the `"""` in that case so no changes should be needed here. The prefix's indent might be a single space, which will result in wrong indentations. 

eg.
```java
private final String foo = //this is a variable 
        """
        Foo
        """";
```

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -498,22 +498,29 @@ public J.Literal visitLiteral(J.Literal literal, P p) {
...
                     }
                 } else {
-                    indent += 2;
+                    indent += (style.getIndentSize() + style.getIndentSize());
```
Jenson3210:
```suggestion
                    indent += style.getContinuationIndent();
```

The `2` I had there was definitely a mistake. While testing, I had the test class written in rewrite-test package where we have a `.editorconfig` that sets continuation length to 2. But this indent you get from IntelliJ is actually continuation and not 2 times indent. It just happens to be the same by default.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
```diff
@@ -498,22 +498,29 @@ public J.Literal visitLiteral(J.Literal literal, P p) {
...
                 }
                 if (currentIndent != indent) {
                     StringBuilder builder = new StringBuilder().append("\"\"\"");
                     for (String line : lines) {
-                        builder.append("\n").append(StringUtils.repeat(" ", indent)).append(line);
+                        builder.append('\n');
+                        shift(builder, indent);
```
Jenson3210, line 518: Great re-use which makes this entire block-shifting tab-aware! 👍🏼

rewrite-java-test/src/test/java/org/openrewrite/java/format/TabsAndIndentsTest.java
```diff
@@ -2668,12 +2668,46 @@ class Test {
...
-                    YES
-                    ""\";
+                          YES
+                          ""\";
                   private final String bar =
                           ""\"
-                            NO
-                            ""\";
+                          NO
+                          ""\";
```
Jenson3210:
```suggestion
                                  NO
                                  ""\";
```

When setting of `align` is disabled, we should indent these with 1 continuation

rewrite-java-test/src/test/java/org/openrewrite/java/format/TabsAndIndentsTest.java
```diff
@@ -2668,12 +2668,46 @@ class Test {
...
+              """,
+            """
+              class Test { 
+                  private final String foo = ""\"
+              \t\t\tYES
+              \t\t\t""\";
+                  private final String bar =
+              \t\t\t""\"
+              \t\t\tNO
+              \t\t\t""\";
```
Jenson3210:
```suggestion
              \t\t\t\t\tNO
              \t\t\t\t\t""\";
```
When setting of align is disabled, we should indent these with 1 continuation

Jenson3210:
@BoykoAlex, thanks for identifying this gap!

I made some suggestions which I think will make the code behave even more accurate. Let me know if you have thoughts here.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
BoykoAlex: Aha, thank you for clarifying this! We should go with what you propose.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
BoykoAlex: Yes, tests are passing with this change. I'm somewhat surprised. I made the change because I saw the cursor message right on the literal during the second visit which already had the right ident hence it was indented further away every time visitor executes again.
Anyway, this works now and it was probably an erroneous state during debugging...

Jenson3210: I think at that moment you perhaps did either not shift the block with tabs or find the common indent for tab chars and the outcome of `withValueSource` was altered all the time to increase indentation. Anyways, this should be stable now with this one 👍🏼

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
BoykoAlex: Okay, yes, i had to add that because had a call to  `getCursor().getTreeParentCursor()`. Makes sense.

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
Jenson3210: @BoykoAlex feel free to ignore this comment. This is merely an indication for @timtebeek and myself!

rewrite-java-test/src/test/java/org/openrewrite/java/format/TabsAndIndentsTest.java
BoykoAlex: Okay, yes, I think i get it: no alignment then next line is a continuation.

rewrite-java-test/src/test/java/org/openrewrite/java/format/TabsAndIndentsTest.java
BoykoAlex: Okay, yes, I think i get it: no alignment then next line is a continuation.

BoykoAlex:
@Jenson3210 thank you for the review :-)

rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
timtebeek: Probably safest indeed; thanks for pointing that out! Are you ok to apply that before a merge? No rush, tomorrow is fine.

Jenson3210:
```suggestion
                String finalContent = content; //TODO remove this assignment together with the next evaluate invocation as it is no longer needed.
                int currentIndent = evaluate(() -> StringUtils.minCommonIndentLevel(finalContent, style.getTabSize()), StringUtils.minCommonIndentLevel(content));
```

Unfortunately, need to copy the string to a final one... Should be good like this, but can't push this back to a fork it seems 🤔

Jenson3210: Okay, it worked from a suggestion. Must have done something wrong in IntelliJ. This should be safe till I do my big round of removing `evaluate` invocations / methods. Thanks!

timtebeek:
Quick note that we're seeing [two failures downstream](https://github.com/openrewrite/rewrite-rewrite/blob/fc629f2143752b8bd16884684b1dd6448a48b4a1/src/test/java/org/openrewrite/java/recipes/SourceSpecTextBlockNewLineTest.java#L87-L137) that I think are related: https://ge.openrewrite.org/s/gtr3ruly7ncng
```
SourceSpecTextBlockNewLineTest > newlineBetweenMultipleTextBlocks() FAILED
    org.opentest4j.AssertionFailedError: [Unexpected result in "MyRecipeTest.java":
    diff --git a/MyRecipeTest.java b/MyRecipeTest.java
    index 54a59ee..bdfba39 100644
    --- a/MyRecipeTest.java
    +++ b/MyRecipeTest.java
    @@ -15,14 +15,14 @@ 
                  }
                  """,
                   """
    -                class Test {
    -                
    -                    
    -                    void test() {
    -                        System.out.println("Hello, world!");
    -                    }
    -                }
    -                """
    +                      class Test {
    +                      
    +                          
    +                          void test() {
    +                              System.out.println("Hello, world!");
    +                          }
    +                      }
    +                      """
              )
           );
         }
    ] 
```
Could either of you have a look? I want to prevent us having to field related questions after today's release.

PR #6699, opened 2026-02-09
Add ModuleContainsFile and ModuleUsesType scanning recipes

Jenson3210:
## Summary

- Add `ModuleContainsFile` scanning recipe that marks files at module level based on file patterns
- Add `ModuleUsesType` scanning recipe that marks files at module level based on type usage
- Both recipes include early-return optimization to skip already-identified modules

## Problem

Existing precondition recipes operate at either repository level (`RepositoryContainsFile`) or individual file level (`UsesType`, `FindSourceFiles`). For multi-module repositories, there's often a need to target recipes at the module level - for example, only applying changes to modules that contain certain configuration files or use certain types.

## Solution

Two new `ScanningRecipe` implementations:

**`ModuleContainsFile`**: Checks if a module contains files matching a glob pattern. Unlike `RepositoryContainsFile` (which marks ALL files if ANY match), this only marks files in modules where the match was found.

**`ModuleUsesType`**: Checks if a module uses a specified type. Unlike `UsesType` (which only marks files directly using the type), this marks ALL files in modules where the type is used.

Both recipes:
- Use `JavaProject` markers to track module membership
- Include early-return optimizations in scanners to skip files in already-identified modules
- Support being used as preconditions for other recipes in Scanning recipes

## Test plan

- [x] Tests for `ModuleContainsFile` covering single/multiple modules, multiple patterns, and no-match cases
- [x] Tests for `ModuleUsesType` covering exact types, wildcards, recursive patterns, and implicit usages
- [x] All existing tests continue to pass

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
@@ -95,14 +95,12 @@ public class ModuleContainsFile extends ScanningRecipe<ModuleContainsFile.Accumu
                     return tree;
                 }
                 SourceFile sourceFile = (SourceFile) tree;
-                if (sourceFile.getMarkers().findFirst(SearchResult.class).isPresent()) {
-                    return tree;
-                }
-                return sourceFile.getMarkers().findFirst(JavaProject.class)
+                if (sourceFile.getMarkers().findFirst(JavaProject.class)
                         .filter(acc.modulesWithMatchingFile::contains)
-                        .<SourceFile>map(jp -> sourceFile.withMarkers(sourceFile.getMarkers()
-                                .add(new SearchResult(Tree.randomId(), "Module contains file matching pattern: " + filePattern))))
-                        .orElse(sourceFile);
+                        .isPresent()) {
+                    return SearchResult.mergingFound(sourceFile, "Module contains file matching pattern: " + filePattern);
+                }
+                return sourceFile;
             }
         };
     }
--- a/rewrite-core/src/main/java/org/openrewrite/search/RepositoryContainsFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/search/RepositoryContainsFile.java
@@ -77,10 +77,7 @@ public class RepositoryContainsFile extends ScanningRecipe<AtomicBoolean> {
         return new TreeVisitor<Tree, ExecutionContext>() {
             @Override
             public Tree visit(Tree tree, ExecutionContext ctx) {
-                if(tree.getMarkers().findFirst(SearchResult.class).isPresent()) {
-                    return tree;
-                }
-                return tree.withMarkers(tree.getMarkers().add(new SearchResult(Tree.randomId(), "Repository contains file matching pattern: " + filePattern)));
+                return SearchResult.mergingFound(tree, "Repository contains file matching pattern: " + filePattern);
             }
         };
     }
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/ModuleUsesType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/ModuleUsesType.java
@@ -98,14 +98,12 @@ public class ModuleUsesType extends ScanningRecipe<ModuleUsesType.Accumulator> {
                     return tree;
                 }
                 SourceFile sourceFile = (SourceFile) tree;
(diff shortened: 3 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (!(tree instanceof SourceFile)) {
+                    return tree;
+                }
+                SourceFile sourceFile = (SourceFile) tree;
+                if (sourceFile.getMarkers().findFirst(SearchResult.class).isPresent()) {
+                    return tree;
```
timtebeek: I don't think this is sufficiently specific, especially when there might be two preconditions that should _both_ apply. Right now it seems like we wouldn't mark changes for this instance when an earlier precondition already marked the file.

Same below as well.

rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
```diff
@@ -0,0 +1,109 @@
...
+                }
+                return sourceFile.getMarkers().findFirst(JavaProject.class)
+                        .filter(acc.modulesWithMatchingFile::contains)
+                        .<SourceFile>map(jp -> sourceFile.withMarkers(sourceFile.getMarkers()
+                                .add(new SearchResult(Tree.randomId(), "Module contains file matching pattern: " + filePattern))))
```
timtebeek: Why not use one of the static methods on`SearchResult` here, like `SearchResult.found` or even the `merging` variants.

timtebeek (changes requested):
I fear the current implementation is insufficient when combined with other preconditions.

rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
Jenson3210: Implementation example was taken from RepositoryContainsFile. I will make updates to that one also!

rewrite-java/src/main/java/org/openrewrite/java/search/ModuleContainsFile.java
Jenson3210: Same as above. Will update all 3

timtebeek:
Locally I had added a test to verify that that dual preconditions case is now handled correctly, as this test now passes:
```java
    @Test
    void twoInstancesShouldOnlyMarkIfBothArePresent() {
        rewriteRun(
          spec -> spec.recipeFromYaml(
            """
              name: both-foo-and-bar
              displayName: "Replace all contents in modules containing both foo.txt and bar.txt"
              description: "Replace all contents in modules containing both foo.txt and bar.txt."
              preconditions:
                - org.openrewrite.java.search.ModuleContainsFile:
                    filePattern: "**/bar.txt"
                - org.openrewrite.java.search.ModuleContainsFile:
                    filePattern: "**/foo.txt"
              recipeList:
                - org.openrewrite.text.ChangeText:
                    toText: "Replaced"
              """,
            "both-foo-and-bar"),
          mavenProject("module-foo-and-bar",
            srcMainJava(
              java(
                """
                  class All {}
                  """,
                """
                  Replaced
                  """
              )
            ),
            srcMainResources(
              text(
                "bar",
                "Replaced",
                spec -> spec.path("bar.txt")
              ),
              text(
                "foo",
                "Replaced",
                spec -> spec.path("foo.txt")
              )
            )
          ),
          mavenProject("module-only-bar",
            srcMainJava(
              java(
                """
                  class Bar {}
                  """
              )
            ),
            srcMainResources(
              text(
                "bar",
                spec -> spec.path("bar.txt")
              )
            )
          ),
          mavenProject("module-only-foo",
            srcMainJava(
              java(
                """
                  class Foo {}
                  """
              )
            ),
            srcMainResources(
              text(
                "foo",
                spec -> spec.path("foo.txt")
              )
            )
          )
        );
    }
```
Not sure if it's worth adding, but figured share it here for your consideration.

timtebeek:
Thanks for the quick fix and neat addition; should make those downstream recipe conditions easier to work with.

PR #6704, opened 2026-02-09
Add Kotlin DSL variable downgrade tests for UpgradePluginVersion

MBoegers:
## Summary
- Add tests covering the scenario where `UpgradePluginVersion` with an exact version could downgrade plugins when versions are stored in a Kotlin DSL `val` variable shared across multiple plugin declarations
- Replace `assert` statements with proper null checks in the scanner for the GString and Identifier variable branches to avoid silent null storage when assertions are disabled

## Test plan
- [x] `dontDowngradeKotlinDslSharedVariable` — verifies no downgrade with shared `val` variable across multiple Kotlin plugins
- [x] `dontDowngradeKotlinDslSinglePluginVariable` — verifies no downgrade with single plugin using a `val` variable
- [x] `upgradeKotlinDslSharedVariable` — verifies upgrades work correctly with shared `val` variable
- [x] All existing `UpgradePluginVersionTest` tests pass


timtebeek:
This seems similar to another change I've seen come by; do we still need this PR?

MBoegers:
I don't think so, it was "just" a reproducer PR, can you linke the PR?

timtebeek:
- https://github.com/openrewrite/rewrite/pull/6746

MBoegers:
retested and resolved

PR #6712, opened 2026-02-10
Fix text block indentation

Jenson3210:
Followup after the failing PR. 

Verified this code against static analysis and rewrite-rewrite

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/format/TabsAndIndentsVisitor.java
@@ -538,10 +538,8 @@ public class TabsAndIndentsVisitor<P> extends JavaIsoVisitor<P> {
                     StringBuilder builder = new StringBuilder().append("\"\"\"");
                     for (String line : lines) {
                         builder.append('\n');
-                        if (!line.isEmpty()) {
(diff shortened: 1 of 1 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/format/TabsAndIndentsTest.java
```diff
@@ -2674,6 +2719,21 @@ class Test {
...
+                              ""\", ""\"
+                              indent me too!
+                              ""\");
+                      method(
+                              ""\"
+                                      indent me!
+                                      ""\",
+                              ""\"
+                                      indent me too!
+                                      ""\"
```
timtebeek, line 2734: This one surprises me a bit; do we want continuation indent after a text block start? I'd expect that to stay left aligned.

Jenson3210: Image

If align when multiline is disabled (default), IntelliJ adds a continuation indent. Personally, I do agree with you though...

timtebeek: Ok, I'll hold my nose and click approve then! 😀

PR #6713, opened 2026-02-10
Python: Fix `MatchSequence` parser incorrectly consuming child's delimiter

knutwannheden:
When parsing `case [c], _:`, the outer implicit tuple MatchSequence was greedily consuming the `[` that belongs to the inner child MatchSequence. This caused the pattern to be parsed as a single bracketed list `[c, _]` instead of an implicit tuple containing `[c]` and `_`.

The fix adds `__is_own_sequence_delimiter` which checks whether a `[` or `(` delimiter belongs to this MatchSequence or to its first child by combining AST inspection (is the first child also a MatchSequence?) with token peeking (are there consecutive delimiters?).

Also includes RPC server improvements: relativeTo path support for parse inputs, bare string PathInput handling, and better error logging.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/rpc/server.py
+++ b/rewrite-python/rewrite/src/rewrite/rpc/server.py
@@ -322,28 +322,23 @@ def handle_parse(params: dict) -> List[str]:
     """Handle a Parse RPC request."""
     inputs = params.get('inputs', [])
     relative_to = params.get('relativeTo')
-    logger.info(f"handle_parse: {len(inputs)} inputs, relativeTo={relative_to}")
     results = []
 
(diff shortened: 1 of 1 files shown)


rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -387,18 +405,18 @@ def handle_get_object(params: dict) -> List[dict]:
         before = remote_objects.get(obj_id)
 
         q = RpcSendQueue(source_file_type)
+        logger.info(f"handle_get_object: starting generate for {obj_id}")
```
jkschneider: logging statement leftovers

PR #6714, opened 2026-02-10
Support inline preconditions in declarative YAML recipeList

bmuschko:
## Summary
- Adds support for per-recipe `preconditions` in declarative YAML `recipeList` entries
- Parses the `preconditions` key from recipe arguments and wraps the recipe in a nested `DeclarativeRecipe`, reusing the existing bellwether pattern
- Only `YamlResourceLoader` is modified (no changes to `DeclarativeRecipe`)

### YAML syntax

```yaml
recipeList:
  - org.openrewrite.text.ChangeText:
      toText: changed
      preconditions:
        - org.openrewrite.FindSourceFiles:
            filePattern: "**/target.txt"
```

Inline preconditions use AND logic (all must be satisfied). They compose with top-level preconditions — both must pass for the recipe to run.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/YamlResourceLoader.java
@@ -58,6 +58,8 @@ public class YamlResourceLoader implements ResourceLoader {
     private static final PropertyPlaceholderHelper propertyPlaceholderHelper =
             new PropertyPlaceholderHelper("${", "}", ":");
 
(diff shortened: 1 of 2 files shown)


okundzich:
let's check with Jonathan on this

sambsnyd (commented):
There are implications for features like the saas recipe builder if we adopt this. Scheduling and capacity for downstream work like that might decide whether this gets adopted now, later, or never.

okundzich:
fyi @sjungling

Issue #6717, opened 2026-02-11
ChangeParentPom does not fully update local parent maven models in submodules


ryan-hudson:
As part of #6536, ChangeParentPom was refactored to propagate maven model parent version changes to submodules. This change now correctly sets the new parent GAV in submodules and results in overall correct models for these submodules, but the submodules understanding of other parent fields from its parent model (such as dependency management, plugin management, and properties) are still outdated.

In the below test case, I have a two module project containing a local parent and rest module. The local parent module has junit-bom as a parent and the rest module has the local parent as its parent. When running ChangeParentPom to change the version of this junit parent from 5.10.3 to 6.0.1, I would expect to see the new version dependency management when viewing the `resolutionResult.parent` model from the rest module. However, this model still contains the old Junit 5.10.3 dependency management, even though the `resolutionResult.parent.parent` (junit-bom) version is correctly set to the new 6.0.1 version.

```java
public class ChangeParentTest implements RewriteTest {
    @Override
    public void defaults(RecipeSpec spec) {
        spec.recipe(
            new ChangeParentPom(
                "org.junit",
                "org.junit",
                "junit-bom",
                "junit-bom",
                "6.0.1",
                null,
                null,
                null,
                false,
                null
            )
        );
    }
    @Test
    void junitBomParentResolution() {
        rewriteRun(
            mavenProject(
                "sample-parent",
                pomXml(
                    """
                    <?xml version=\"1.0\" encoding=\"UTF-8\"?>
                    <project xmlns=\"http://maven.apache.org/POM/4.0.0\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"
                             xsi:schemaLocation=\"http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd\">

                      <modelVersion>4.0.0</modelVersion>
                      <groupId>com.example</groupId>
                      <artifactId>sample-parent</artifactId>
                      <version>${revision}</version>
                      <packaging>pom</packaging>
                      <name>sample-parent</name>

                      <modules>
                        <module>sample-rest</module>
                      </modules>

                      <parent>
                        <groupId>org.junit</groupId>
                        <artifactId>junit-bom</artifactId>
                        <version>5.10.3</version>
                        <relativePath/>
                      </parent>
                    </project>
                    """.trim(),
                    """
                    <?xml version=\"1.0\" encoding=\"UTF-8\"?>
[...]

PR #6718, opened 2026-02-11
Fix ChangeParentPom not fully updating submodule parent models (#6717)

timtebeek:
## Summary
When ChangeParentPom changes a parent POM's version, submodules' parent markers had the GAV updated but retained stale dependency management, properties, and plugin management from the old parent version.

## Changes
- Re-resolve the updated root marker in the scanner phase so it has fully populated dependency management, properties, and plugin management from the new parent version
- Call maybeUpdateModel() for submodule documents so their own models are re-resolved against the updated parent hierarchy
- Add test case from the issue to verify the fix works for multi-module projects

## Test plan
- [x] New test `submoduleParentModelDependencyManagementUpdated` passes
- [x] All existing `ChangeParentPomTest` tests pass (40+ tests)

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
@@ -301,14 +301,10 @@ public class ChangeParentPom extends ScanningRecipe<ChangeParentPom.Accumulator>
                                     targetRelativePath
                             );
                             Pom updatedPom = mrr.getPom().getRequested().withParent(updatedParentRef);
-                            ResolvedPom updatedResolvedPom = mrr.getPom().withRequested(updatedPom);
-                            try {
-                                MavenPomDownloader downloader = new MavenPomDownloader(
-                                        mrr.getProjectPoms(), ctx,
-                                        mrr.getMavenSettings(), mrr.getActiveProfiles());
-                                updatedResolvedPom = updatedResolvedPom.resolve(ctx, downloader);
-                            } catch (MavenDownloadingException ignored) {
-                            }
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeParentPom.java
```diff
@@ -302,6 +302,13 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
                             );
                             Pom updatedPom = mrr.getPom().getRequested().withParent(updatedParentRef);
                             ResolvedPom updatedResolvedPom = mrr.getPom().withRequested(updatedPom);
+                            try {
+                                MavenPomDownloader downloader = new MavenPomDownloader(
+                                        mrr.getProjectPoms(), ctx,
+                                        mrr.getMavenSettings(), mrr.getActiveProfiles());
+                                updatedResolvedPom = updatedResolvedPom.resolve(ctx, downloader);
+                            } catch (MavenDownloadingException ignored) {
+                            }
```
sambsnyd: If we suppress the exception we will have no idea why it isn't working if there is a problem. I suggest saving the exception to the accumulator then attaching it as a `Warning` marker to the `<parent>` element during visiting.

timtebeek: Hmm; I've reworked it to now use the outer catch block that was already there, but similarly had discarded any such exception. Would you want to see the same change there? `git praise` said it was you ;)

sambsnyd: Lol I'll go tweak it to not suppress the error

PR #6719, opened 2026-02-12
AddOrUpdateAnnotationAttribute: Handle fully qualified enum values

Jenson3210:
## New functionalities

- **Import and shorten fully qualified enum values**
  ```java
  // Before
  @Foo
  // After (with attributeValue="org.example.Values.ONE")
  import org.example.Values;
  @Foo(Values.ONE)
  ```

- **Append class values to existing single-class annotations**
  ```java
  // Before
  @Foo(Integer.class)
  // After (with attributeValue="Long.class", appendArray=true)
  @Foo({Integer.class, Long.class})
  ```

## Bug fixes

- **Empty array handling with `appendArray=true`**
  ```java
  // Before
  @Foo(value = {})
  // Was producing: @Foo({, "hello"})
  // Now correctly: @Foo("hello")
  ```

- **Attribute name vs value check in array updates**
  ```java
  // Was checking attributeValue instead of attributeName when updating arrays
  ```

- **Comma-separated class values incorrectly matched as FQ class**
  ```java
  // "Integer.class,Long.class" was incorrectly treated as a single FQ class
  // Now correctly parsed as multiple class values
  ```

- **Missing type info when appending FQ class values**
  ```java
  // Before
  @Foo(Integer.class)
  // After (with attributeValue="com.example.MyClass.class", appendArray=true)
  import com.example.MyClass;
  @Foo({Integer.class, MyClass.class})  // MyClass.class now has proper type info
  ```

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
@@ -400,7 +400,7 @@ public class AddOrUpdateAnnotationAttribute extends Recipe {
                 newItemsList.add(new J.Literal(randomId(), SINGLE_SPACE, EMPTY, attribute, maybeQuoteStringArgument(annotation, attribute), null, JavaType.Primitive.String));
             }
             // Filter out empty elements (e.g., from an empty array initializer `{}`)
-            List<Expression> nonEmptyInitializer = ListUtils.map(initializerList, it -> it instanceof J.Empty ? null : it);
+            List<Expression> nonEmptyInitializer = ListUtils.filter(initializerList, it -> !(it instanceof J.Empty));
             return ListUtils.concatAll(nonEmptyInitializer, newItemsList);
         }
 


rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -276,18 +294,84 @@ private J.NewArray createNewArrayWithExistingAndNew(J.Annotation annotation, J.L
...
         String withoutClassSuffix = fqn.substring(0, fqn.length() - 6);
         return withoutClassSuffix.substring(withoutClassSuffix.lastIndexOf('.') + 1) + ".class";
     }
 
+    private boolean isFullyQualifiedEnumValue(J.Annotation annotation) {
+        if (attributeValue == null || attributeValue.endsWith(".class") || attributeValue.contains(",")) {
+            return false;
+        }
+        if (!attributeValue.matches("[\\w.$]+") ||
+            !findMethod(annotation, attributeName()).isPresent() ||
```
timtebeek: If we're looking at the method I'd half expected us to look whether the type of the attribute on the annotation is an enum. Would that help avoid confusion?

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -98,9 +98,12 @@ public J.Annotation visitAnnotation(J.Annotation original, ExecutionContext ctx)
                 }
 
                 String newAttributeValue;
-                if (attributeValue != null && attributeValue.endsWith(".class") && StringUtils.countOccurrences(attributeValue, ".") > 1) {
-                    maybeAddImport(attributeValue.substring(0, attributeValue.length() - 6));
+                if (isFullyQualifiedClass()) {
+                    maybeAddImport(attributeValue.substring(0, attributeValue.length() - 6), false);
                     newAttributeValue = attributeValue;
+                } else if (isFullyQualifiedEnumValue(a)) {
+                    maybeAddImport(getEnumClassName(attributeValue), false);
```
timtebeek: Why do we need the `onlyIfReferenced = false` here? Are those not detected? Or are they not set further down?

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
Jenson3210: so instead of attributeIsString we should check if attribute is a method for which the return type is an enum.

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -148,6 +152,9 @@ public J.Annotation visitAnnotation(J.Annotation original, ExecutionContext ctx)
 
                 if (original != a) {
                     doAfterVisit(new SimplifySingleElementAnnotation().getVisitor());
+                    if (isFullyQualifiedClass()) {
+                        doAfterVisit(new ShortenFullyQualifiedTypeReferences().getVisitor());
+                    }
```
timtebeek, line 157: I'm not seeing any difference from these lines being present or not; do we need this?

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
```diff
@@ -317,7 +399,9 @@ private List<Expression> updateInitializer(J.Annotation annotation, List<Express
                 }
                 newItemsList.add(new J.Literal(randomId(), SINGLE_SPACE, EMPTY, attribute, maybeQuoteStringArgument(annotation, attribute), null, JavaType.Primitive.String));
             }
-            return ListUtils.concatAll(initializerList, newItemsList);
+            // Filter out empty elements (e.g., from an empty array initializer `{}`)
+            List<Expression> nonEmptyInitializer = ListUtils.map(initializerList, it -> it instanceof J.Empty ? null : it);
```
timtebeek: Nit: we have a method to filter :)
```suggestion
            // Filter out empty elements (e.g., from an empty array initializer `{}`)
            List<Expression> nonEmptyInitializer = ListUtils.filter(initializerList, it -> !(it instanceof J.Empty));
```

rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java
Jenson3210: I am seeing  `@Import(OtherConfig.class)` ->   `@Import({OtherConfig.class, org.example.MyConfiguration.class})` if we do not shorten them. This also handles the import additions.

Jenson3210: Adding a test that mimics this

Jenson3210: we need this call as this visitor recognizes same-package classes and removes the FQ prefix

PR #6733, opened 2026-02-13
Remove all nulls in a single pass

timtebeek:
Avoids repeatedly looping over the collection for as many nulls as are present. Delegates to Collection.removeIf with
```java
    default boolean removeIf(Predicate<? super E> filter) {
        Objects.requireNonNull(filter);
        boolean removed = false;
        final Iterator<E> each = iterator();
        while (each.hasNext()) {
            if (filter.test(each.next())) {
                each.remove();
                removed = true;
            }
        }
        return removed;
    }
```

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/ListUtils.java
@@ -215,8 +215,7 @@ public final class ListUtils {
         }
 
         if (newLs != ls && nullEncountered) {
-            //noinspection StatementWithEmptyBody
-            while (newLs.remove(null)) ;
+            newLs.removeIf(Objects::isNull);
         }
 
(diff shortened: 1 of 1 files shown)


tkvangorder (commented):
Thanks Tim!

timtebeek:
@knutwannheden has been doing some benchmarking; looks like it's faster to keep (and document) the original, as we only sparsely have mulitple nulls that trigger repeated traversals, whereas there's a higher tax on using `removeIf`.

tkvangorder:
Interesting, so the overhead of creating the iterator + the fact that typically there are only one or two nulls means using `remove(object)` is quicker?

timtebeek:
Yep! I agree it looks odd, but that's what we're finding. I'll close this PR; we can separately add a comment to indicate so.

PR #6736, opened 2026-02-13
Fix JDK 25 LinkageError in JavaUnrestrictedClassLoader

timtebeek:
## Summary

Fixes a `LinkageError` that occurs on JDK 25 when parsing projects with Lombok annotations on comment-free main sources followed by test sources containing comments. JDK 25 changed `DocCommentTable` to reference public API types across module boundaries, causing the app classloader to eagerly load internal compiler classes. This change wraps the `defineClass()` call in a try-catch for `LinkageError` and falls back to parent delegation when the class was already loaded.

## Test Plan

- [x] `./gradlew :rewrite-java:test` passes
- [x] No LinkageError when parsing JDK 25 projects with the fix


timtebeek:
Verified downstream in rewrite-maven-plugin; as reported we still need to add `--add-exports`, but can now parse the reproducer provided separately.

timtebeek:
Pinging @shanman190 and @knutwannheden for async review, but I don't see any immediate concerns here.

PR #6742, opened 2026-02-13
Pass resolvers to JS/Python RPC so PrepareRecipe callbacks can resolve Maven recipes

kmccarp:
## Summary

- When a JavaScript or Python recipe's preconditions reference Java recipes (e.g. `UsesType`, `UsesMethod`), the RPC process calls back to Java via `PrepareRecipe`. These Java recipes are bundled under the `"maven"` ecosystem, but `JavaScriptRewriteRpc` and `PythonRewriteRpc` were calling the 2-arg `RewriteRpc` constructor which passes `emptyList()` for resolvers — causing `IllegalStateException("No available resolver for 'maven' ecosystem")`.
- Add `resolvers` field to both `JavaScriptRewriteRpc.Builder` and `PythonRewriteRpc.Builder` (defaulting to `emptyList()` for backward compatibility) and pass through to the 3-arg super constructor.
- Add defense-in-depth fallback in `RewriteRpc.PrepareRecipe` handler: if `listing.prepare()` throws `IllegalStateException` (resolver missing), fall through to `RecipeLoader` for recipes already on the classpath.

## Test plan

- [x] Existing `RewriteRpcTest` passes
- [x] New test `prepareRecipeFallsBackToRecipeLoaderWhenResolverMissing` verifies the fallback
- [ ] Companion CLI PR passes resolvers to the builders

### DIFF
--- a/rewrite-python/src/main/java/org/openrewrite/python/rpc/PythonRewriteRpc.java
+++ b/rewrite-python/src/main/java/org/openrewrite/python/rpc/PythonRewriteRpc.java
@@ -19,6 +19,7 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
+import org.openrewrite.marketplace.RecipeBundleResolver;
 import org.openrewrite.marketplace.RecipeMarketplace;
 import org.openrewrite.python.*;
 import org.openrewrite.python.marker.PythonResolutionResult;


rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -160,9 +160,15 @@ protected Object handle(Void noParams) {
         jsonRpc.rpc("PrepareRecipe", new PrepareRecipe.Handler(preparedRecipes, (id, opts) -> {
             RecipeListing listing = marketplace.findRecipe(id);
             if (listing != null) {
-                return listing.prepare(resolvers, opts);
+                try {
+                    return listing.prepare(resolvers, opts);
+                } catch (IllegalStateException e) {
+                    // Resolver not available for this ecosystem (e.g., Maven resolver not
+                    // provided to this RPC endpoint). Fall through to class-based loading.
+                }
```
shanman190: This part doesn't feel super necessary as long as we're providing the resolvers when needed. Class based loading is almost certainly going to fail anyway when a resolver isn't present as well.

PR #6746, opened 2026-02-16
Handle local version variables in UpgradePluginVersion for kotlin plugins

MBoegers:
## Summary

- Fix `UpgradePluginVersion` scanner to correctly resolve Kotlin plugin versions when the version is held in a local variable
- Collect local variable values during scanning so semver selectors like `latest.minor` work with variable-based versions
- Apply the same fix to both the `J.Identifier` and `G.GString` branches

## Problem

When a Kotlin Gradle build script defines a local version variable inside the `plugins` block:

```kotlin
plugins {
    val kotlinVersion = "1.9.25"
    kotlin("jvm") version kotlinVersion
    kotlin("plugin.spring") version kotlinVersion
}
```

`UpgradePluginVersion` fails to resolve the new version because the `J.Identifier` branch in the scanner constructs incorrect Maven coordinates. It uses `GroupArtifact("kotlin", "kotlin.gradle.plugin")` instead of `GroupArtifact("org.jetbrains.kotlin.jvm", "org.jetbrains.kotlin.jvm.gradle.plugin")`, which cannot resolve Maven metadata for semver selectors.

Additionally, without knowing the current version from the variable, selectors like `latest.minor` cannot determine the appropriate upgrade target.

## Solution

1. Added `visitVariable` to the scanner to collect local variable declaration values
2. When the version argument is a `J.Identifier`, look up the variable's current value and use `GroupArtifactVersion` (with current version) for proper semver resolution
3. Construct the full Kotlin plugin ID (`org.jetbrains.kotlin.xxx`) in the `J.Identifier` and `G.GString` branches, matching the existing `J.Literal` branch behavior
4. Removed incorrect `assert resolvedPluginVersion != null` that would fail during idempotency checks

## Test plan

- [x] Existing `UpgradePluginVersionTest` tests pass
- [x] Full `rewrite-gradle` test suite passes
- [x] New tests added:
  - `upgradeKotlinPluginLocalVariable` — exact version upgrade with local `val`
  - `upgradeKotlinPluginLocalVariableWithSemverSelector` — `latest.minor` with local `val`
  - `dontDowngradeKotlinPluginLocalVariable` — don't-downgrade guard

- Fixes moderneinc/customer-requests#1807


shanman190:
Should we consider a similar update to the `GradlePlugin` trait?

I do wonder if it couldn't end up also replacing much of the logic that we maintain here as well, but that can be for another time.

PR #6754, opened 2026-02-17
Fix empty dependencies block left after constraint-to-rule conversion

Jenson3210:
## Summary
- Fix `DependencyConstraintToRule.isEmptyDependenciesBlock` to handle truly empty `dependencies {}` blocks (0 statements)
- When `UpgradeTransitiveDependencyVersion` adds constraints to `build.gradle` and the Spring DM plugin converts them to `resolutionStrategy`, the leftover empty `dependencies {}` block was not being cleaned up
- Add test coverage for the `apply from:` pattern with transitive dependency constraints

## Changes
- **`DependencyConstraintToRule`**: Add check for `b.getStatements().isEmpty()` in `isEmptyDependenciesBlock` — previously only checked for single null-return statements
- **`UpgradeTransitiveDependencyVersionTest`**: Add `addConstraintWithApplyFrom()` and `useResolutionStrategyWithApplyFromWhenSpringDependencyManagementPluginIsPresent()` tests

## Test plan
- [x] `useResolutionStrategyWithApplyFromWhenSpringDependencyManagementPluginIsPresent` fails without the fix, passes with it
- [x] All existing `UpgradeTransitiveDependencyVersionTest` tests pass
- [x] All existing `DependencyConstraintToRuleTest` tests pass

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UpgradeTransitiveDependencyVersion.java
@@ -520,10 +520,7 @@ public class UpgradeTransitiveDependencyVersion extends ScanningRecipe<UpgradeTr
             if (tree instanceof JavaSourceFile) {
                 JavaSourceFile cu = (JavaSourceFile) tree;
                 GradleProject gradleProject = cu.getMarkers().findFirst(GradleProject.class).orElse(null);
-                // Only add constraints to primary build files, not to applied scripts like dependencies.gradle
-                String fileName = cu.getSourcePath().getFileName().toString();
-                boolean isPrimaryBuildFile = "build.gradle".equals(fileName) || "build.gradle.kts".equals(fileName);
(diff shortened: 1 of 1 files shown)


timtebeek:
There's a very similar effort in https://github.com/openrewrite/rewrite/pull/6650 that I'd love your eye on as well @Jenson3210

Jenson3210:
We might need this change in multiple recipes 🤔 

Will have a look at the other one tomorrow.

Jenson3210:
Good catch @timtebeek! I dug into why the tests pass without the `isPrimaryBuildFile` change and found that applied scripts like `dependencies.gradle` already don't get constraints because `DEPENDENCIES_DSL_MATCHER` doesn't match non-type-attributed method invocations in applied scripts. So that guard was redundant.

I've removed the `isPrimaryBuildFile` check in ec64ae3. The actual fix is just the `DependencyConstraintToRule.isEmptyDependenciesBlock` change, which properly handles truly empty `dependencies {}` blocks (0 statements) left after constraints are converted to `resolutionStrategy`. The `useResolutionStrategyWithApplyFromWhenSpringDependencyManagementPluginIsPresent` test now properly fails without this fix and passes with it.

Also noted the overlap with #6650 — the `AddDependency` recipe has a similar `apply from:` issue. Will take a look.

PR #6760, opened 2026-02-18
`ChangeImport`: rewrite qualified references when `old_name` is specified

knutwannheden:
## Summary
- When `old_name` is set, `ChangeImport` now also rewrites qualified references (`X.name()` → `Y.name()`) and manages direct imports (`import X` → `import Y`)
- Supports aliased imports (`import X as f` / `f.name()` → `import Y` / `Y.name()`)
- Updates type attribution on rewritten nodes (`Identifier._type`, `MethodInvocation._method_type`)
- Fixes pre-existing bugs in `AddImport` (insertion point) and `RemoveImport` (usage collection) for standalone `Import` nodes

## Test plan
- [x] 5 new tests covering qualified method calls, aliased imports, retained imports, field access, and different new names
- [x] All 10 `test_change_import` tests pass
- [x] All 23 recipe tests pass (no regressions)

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/python/recipes/change_import.py
+++ b/rewrite-python/rewrite/src/rewrite/python/recipes/change_import.py
@@ -14,7 +14,7 @@
 
 """Recipe to change Python imports from one module/name to another."""
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace as dc_replace
 from typing import Any, Optional
 
 from rewrite import ExecutionContext, Recipe, TreeVisitor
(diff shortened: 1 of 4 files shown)


rewrite-python/rewrite/src/rewrite/python/recipes/change_import.py
```diff
@@ -124,48 +135,80 @@ def __init__(self):
...
+                self._new_module_type: Optional[JavaType.Class] = None
 
             def visit_compilation_unit(self, cu: CompilationUnit, p: ExecutionContext) -> J:
-                # First pass: check if the old import exists
+                # Reset state for each compilation unit
                 self.has_old_import = False
                 self.old_alias = None
+                self.has_direct_module_import = False
+                self.module_alias = None
+                self.rewrote_qualified_refs = False
```
timtebeek, line 146: What's the significance of setting these fields both in `__init__` and in `visit_compilation_unit` ? I'd somewhat expect only one of those to be enough, or are you guarding against overrides that don't call super?

knutwannheden: Good observation! The `__init__` constructor is in Python code often used to declare fields. But I've now separated the declaration from the initialization.

PR #6761, opened 2026-02-18
Parse more checkstyle rules into OpenRewrite styles.

sambsnyd:
This ended up being a larger, more impactful change than I expected. 

I started off implementing support for checkstyle rules we did not previously support.
In the process I noticed that some of our existing checkstyle defaults did not match checkstyle's documented defaults, so I fixed those discrepancies.
Then I saw that an OSS contributor added a `CustomImportOrder` rule... but none of our formatting recipes knew it existed so it was completely pointless. So I converted that into the `ImportLayoutStyle`.
Then I wondered if the import layout style we loaded from checkstyle configuration would correctly take precedence over the autodetected import layout style and discovered it would not.

So I dug into how Styles are loaded and merged together and found that we had inadvertently implemented "first wins" rather than "last wins" semantics. Which meant that autodetected styles would supersede configured styles, oops. Perhaps it is a testament to  the usefulness of Autodetect that we didn't get more complaints about this before now.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
@@ -605,7 +605,6 @@ public class CheckstyleConfigLoader {
                     boolean separateLineBetweenGroups = module.prop("separateLineBetweenGroups", true);
                     String specialImportsRegExp = module.properties.get("specialImportsRegExp");
                     String standardPackageRegExp = module.properties.get("standardPackageRegExp");
-                    String thirdPartyPackageRegExp = module.properties.get("thirdPartyPackageRegExp");
                     String rulesInput = module.properties.get("customImportOrderRules");
 
                     if (specialImportsRegExp == null || specialImportsRegExp.isEmpty()) {
@@ -638,12 +637,12 @@ public class CheckstyleConfigLoader {
                             builder.importStaticAllOthers();
                             hasStaticCatchAll = true;
                         } else if ("STANDARD_JAVA_PACKAGE".equals(rule)) {
-                            addRegExpPackages(builder, standardPackageRegExp, false);
+                            addRegExpPackages(builder, standardPackageRegExp);
                         } else if ("THIRD_PARTY_PACKAGE".equals(rule)) {
                             builder.importAllOthers();
                             hasNonStaticCatchAll = true;
                         } else if ("SPECIAL_IMPORTS".equals(rule)) {
-                            addRegExpPackages(builder, specialImportsRegExp, false);
+                            addRegExpPackages(builder, specialImportsRegExp);
                         } else if (rule.startsWith("SAME_PACKAGE")) {
                             // SAME_PACKAGE(n) cannot be accurately represented in ImportLayoutStyle
                             builder.importAllOthers();
@@ -663,7 +662,7 @@ public class CheckstyleConfigLoader {
                 .collect(toSet());
     }
 
-    private static void addRegExpPackages(ImportLayoutStyle.Builder builder, String regExp, boolean isStatic) {
+    private static void addRegExpPackages(ImportLayoutStyle.Builder builder, String regExp) {
         // Convert common regex patterns to package wildcards
         // e.g., "^(java|javax)\\." → java.*, javax.*
         // e.g., "^org\\." → org.*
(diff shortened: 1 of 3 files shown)


rewrite-core/src/main/java/org/openrewrite/style/NamedStyles.java
pstreef, line 97: This says “Styles earlier in the list take precedence over styles later in the list” but the other merge() overload above (line 64) says the opposite: “gives precedence to non-null values from styles which appear later in the list.” Looking at the code, the new comment seems correct.

rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
```diff
@@ -584,27 +594,518 @@ public String toString() {
...
+
+    private static @Nullable Set<TabsAndIndentsStyle> tabCharacterStyle(Map<String, List<Module>> conf) {
+        List<Module> moduleList = conf.get("FileTabCharacter");
+        if (moduleList == null) {
+            return null;
+        }
+        // OK to return a style with nullable non-null fields, if Autodetections doesn't provide Style.applyDefaults() will
+        //noinspection DataFlowIssue
+        return moduleList.stream()
+                .map(module -> new TabsAndIndentsStyle(false, null, null, null, null))
```
pstreef, line 735: I was a little confused about this, but it seems `FileTabCharacter` means tabs are violations so this is correct.

Maybe a comment on the meaning of the rule would be useful though.

rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
```diff
@@ -584,27 +594,518 @@ public String toString() {
...
+    }
+
+    // https://checkstyle.sourceforge.io/checks/imports/avoidstarimport.html
+    private static @Nullable Set<ImportLayoutStyle> avoidStarImportStyle(Map<String, List<Module>> conf) {
+        List<Module> moduleList = conf.get("AvoidStarImport");
+        if (moduleList == null) {
+            return null;
+        }
+        return moduleList.stream()
+                .map(module -> new ImportLayoutStyle(
```
pstreef: Heads up: this creates an `ImportLayoutStyle` with an empty layout, which falls back to IntelliJ defaults in the constructor. When merged with another `ImportLayoutStyle` (e.g. from `ImportOrder`), `ImportLayoutStyle.merge()` just picks the higher-priority one, so whichever is processed last wins as a whole, including the layout.

For some configs this works out because `importOrderStyle()` runs after `avoidStarImportStyle()` in the stream, so `ImportOrder`’s layout wins. But it also means the `Integer.MAX_VALUE` star import thresholds from here get silently dropped. If someone has `AvoidStarImport` without `ImportOrder`, they will get IntelliJ default layout as an unexpected side effect.

Not blocking, but worth thinking about whether `AvoidStarImport` could set only the thresholds without emitting a full `ImportLayoutStyle`.

rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
```diff
@@ -269,10 +282,10 @@ public static Checkstyle loadCheckstyleConfig(InputStream checkstyleConf, Map<St
                         arrayInitializer = tokens.contains("ARRAY_INIT");
                         annotation = tokens.contains("AT");
                         indexOperation = tokens.contains("INDEX_OP");
-                        dot = tokens.contains("INDEX_OP");
+                        dot = tokens.contains("DOT");
                         inc = tokens.contains("INC");
                         dec = tokens.contains("DEC");
-                        bnoc = tokens.contains("BNOC");
+                        bnoc = tokens.contains("BNOT");
```
pstreef, line 288: nit: should this also be renamed to `bnot`?

rewrite-java/src/main/java/org/openrewrite/java/style/CheckstyleConfigLoader.java
```diff
@@ -78,7 +80,17 @@ public static Checkstyle loadCheckstyleConfig(InputStream checkstyleConf, Map<St
...
+                                indentationStyle(conf),
+                                rightCurlyStyle(conf),
+                                leftCurlyStyle(conf),
+                                whitespaceAroundStyle(conf),
+                                whitespaceAfterStyle(conf),
+                                importOrderStyle(conf),
+                                parenPadStyle(conf))
                         .filter(Objects::nonNull)
                         .flatMap(Set::stream)
                         .collect(toSet()));
```
pstreef, line 96: Nice catch converting `CustomImportOrderStyle` into something actually used!

`customImportOrderStyle()`, `avoidStarImportStyle()`, and `importOrderStyle()` can all produce `ImportLayoutStyle` instances. Since they're collected into a `HashSet` and `ImportLayoutStyle.merge()` does last-wins, the result depends on iteration order — which is non-deterministic for `HashSet`. Should this be a `LinkedHashSet` to make the stream ordering deterministic?
 
Longer term, it might be cleaner to not emit a full `ImportLayoutStyle` from `AvoidStarImport` and instead only influence the star import thresholds on whatever layout is already configured.

PR #6764, opened 2026-02-18
Restore recipe hierarchy rows in SourcesFileResults

timtebeek:
## Summary

- Restore recursive hierarchy recording in `RecipeRunCycle` that was removed in #6022
- Hierarchy rows now use `0` for `estimatedTimeSaving` so the column remains summable
- Add test for multi-level recipe hierarchy

## Problem

- The Sankey visualization shows recipe execution hierarchies by reading `(parentRecipe, recipe)` pairs from the `SourcesFileResults` data table. PR #6022 removed the recursive recording of hierarchy rows to fix non-summable `estimatedTimeSaving` values. This caused the Sankey to only show 1-2 levels instead of the full drill-down (5+ levels for migrations like Java 25).

## Solution

Restore the `recordSourceFileResult()` method but pass `0L` for `estimatedTimeSaving` on all non-leaf hierarchy rows. Only the leaf recipe that actually made the change carries its effort value. This gives the Sankey all the edges it needs while keeping the effort column summable.

## Test plan

- [x] Existing `SourceFileResultsTest.hierarchical` passes unchanged
- [x] Existing `RecipeEstimatedEffortTest.estimatedTimeSavingsForMultipleRecipes` passes unchanged (effort sums still correct)
- [x] New `SourceFileResultsTest.deepHierarchyRecordsAllLevels` verifies hierarchy rows are produced with zero effort

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -285,7 +285,7 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
                 afterPath,
                 parentName,
                 recipe.getName(),
-                0L,
+                0L, // Zero here, as we later sum only the recipes that themselves made changes
                 cycle));
         recordSourceFileResult(beforePath, afterPath, recipeStack.subList(0, recipeStack.size() - 1), ctx);
     }


timtebeek:
Any immediate thoughts here @pstreef given your involvement with
- https://github.com/openrewrite/rewrite/pull/6022 ?

pstreef:
yeah, perhaps we should just have a different data-table if this is just for 1 visualization? The reason for removal was more than just being able to sum. It was also confusing.

I would involve @jkschneider and/or @sambsnyd here for opinions

timtebeek:
Thanks both! I'll merge as we're one to one restoring what was removed in https://github.com/openrewrite/rewrite/pull/6022, with the exception of the effort which is explicitly set to zero and documented as such. I want to acknowledge that Peter suggested an alternative data table, but seeing how it would have to hook into RecipeRunCycle, and be attached for every run I think it'd be better not to introduce a near duplicate if the zero-time rows here work as well.

PR #6766, opened 2026-02-18
Upgrade rewrite-kotlin to Kotlin 2 compiler (K2)

jkschneider:
## Summary
- Upgrades `kotlin-compiler-embeddable` from 1.9.25 to 2.2.0 (Kotlin 1 is EOL and incompatible with Java 25)
- Migrates all K2 FIR API changes in-place: session creation via `JvmFrontendPipelinePhase`, type ref/expression renames, `FirResolvedArgumentList.mapping` for correct argument-to-parameter resolution
- Overrides `classpathFromResources` in `KotlinParser.Builder` to resolve full JARs instead of `.tt` stubs (K2's FIR compiler cannot read type-transformed class files)

- Supersedes #5575 and #5963.

Co-authored-by: Marius Barbulescu <marius.barbulescu@gmail.com>

### DIFF
--- a/rewrite-java-test/build.gradle.kts
+++ b/rewrite-java-test/build.gradle.kts
@@ -16,6 +16,7 @@ dependencies {
     testImplementation("io.github.classgraph:classgraph:latest.release")
     testImplementation("org.junit-pioneer:junit-pioneer:2.0.0")
     testRuntimeOnly(project(":rewrite-java-21"))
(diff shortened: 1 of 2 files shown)


timtebeek:
Looks like we've arrived at some of the same changes as seen in 
- https://github.com/openrewrite/rewrite/pull/6338

Noting that here to double check and close that PR as well once this is merged.

flex-gwanbin:
finally! I cannot wait to use it.

PR #6776, opened 2026-02-19
Use resolved version of managed dependency in `ResolvedPom`'s managed dependencies

steve-aom-elliott:
## What's changed?
- Alters `ResolvedManagedDependency` creation in `ResolvedPom` so that it will actually have the resolved version of the managed dependency used for the GAV (requested GAV remains untouched) rather than it being allowed to present as `x:y:latest` or `x:y:release`, which caused comparison issues during upgrade checks.

## What's your motivation?
If you had something like
```xml
<project>
    <groupId>com.mycompany</groupId>
    <artifactId>some-project</artifactId>
    <version>1</version>
    <dependencyManagement>
        <dependencies>
            <dependency>
                <version>latest</version>
            </dependency>
        </dependencies>
    </dependencyManagement>
    <dependencies>
        <dependency>
        </dependency>
    </dependencies>
</project>
```
and requested to upgrade `com.google.guava:guava` to a version that's not the latest (`latest` at time of writing is `33.5.0-jre`), such as `29.0-jre`, it was clobbering the `latest` version with the requested one, despite the requested being lower than `latest`. This was due to the `ResolvedManagedDependency` having a GAV with `latest` still as the version, which didn't get processed safely during trying to upgrade the version (it changed it to `0.0.0` which implied anything requested would be an upgrade, which was wrong)

Technically `MavenDependency.findNewerVersion(..)` could have been altered to attempt to account for `latest` / `release` versions, but their usage in comparisons differ by the `VersionComparator` used, and the existing setup differed in behaviour between direct dependencies (that acquired their version numbers from resolving `latest` / `release` from the managed dependency) and the managed dependency itself, which didn't feel correct.

### Checklist

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
@@ -2114,7 +2114,7 @@ class AddDependencyTest implements RewriteTest {
     }
 
     @Test
-    void blah2() {
+    void doesNotClobberLatestWithExplicitVersion() {
         rewriteRun(
           spec -> spec.recipe(addDependency("com.google.guava:guava:29.0-jre")),
           mavenProject("project",


rewrite-maven/src/test/java/org/openrewrite/maven/AddDependencyTest.java
```diff
@@ -2079,6 +2079,66 @@ void addDependencyUpgradesVersionOnlyWhenExistingDependencyHasEquivalentScopeAnd
...
+                    </dependencies>
+                </project>
+                """
+            )
+          )
+        );
+    }
+
+    @Test
+    void blah2() {
```
sambsnyd:
```suggestion
    void doesNotClobberLatestWithExplicitVersion() {
```

PR #6778, opened 2026-02-20
Filter duplicate generated files in ScanningRecipe.generate()

### LINKED ISSUE #6771: ScanningRecipe.generate() should guard against creating files that already exist
## Problem

`ScanningRecipe.generate()` can return source files whose paths collide with files already in the source set. The framework adds these to the accumulator without any collision check (`RecipeRunCycle.java:155`):

```java
List<SourceFile> generated = new ArrayList<>(scanningRecipe.generate(...));
acc.addAll(generated);  // no collision check against existing source paths
```

When this happens, the recipe run produces a "new file" diff for a path that already exists. The resulting patch cannot be applied because git rejects creating a file that already exists:

```
org.eclipse.jgit.api.errors.PatchApplyException: Cannot apply:
  [Error applying patch in src/main/resources/application-dev.yml:
   Cannot perform UNIFIED action which overrides an existing file]
```

### Concrete example

We hit this with `SeparateApplicationYamlByProfile` in rewrite-spring. A repository has both:
- `application.yml` containing a `spring.config.activate.on-profile: dev` section
- An existing `application-dev.yml`

The recipe scans `application.yml`, finds the dev profile section, and generates a new `application-dev.yml` via `generate()` — without checking that the file already exists. The result is a "new file" diff that collides with the existing file, and the patch fails to apply.

### Framework gap

The `ScanningRecipe` Javadoc acknowledges this is the recipe author's responsibility ("in almost every case we check that a file doesn't yet exist"), but in practice recipes miss this. The existence of `FindCollidingSourceFiles` as a diagnostic recipe further suggests this is a known footgun.

## Suggested direction

Add a framework-level guard in `RecipeRunCycle` that filters out generated files whose `sourcePath` matches an existing file in the source set. This would protect against any recipe that forgets the check, rather than relying on each recipe author to get it right.

Open question:
- Should collisions be silently skipped, logged as a warning, or cause an error?

pstreef:
## Problem

`ScanningRecipe.generate()` can return source files whose paths collide with files already in the source set. The framework adds these without any collision check, producing "new file" diffs that fail to apply because the file already exists.

We hit this with `SeparateApplicationYamlByProfile` in rewrite-spring: when a repo has both `application.yml` (with a dev profile section) and an existing `application-dev.yml`, the recipe generates a new `application-dev.yml` that collides with the existing one.

## Solution

Silently filter out generated files whose `sourcePath` collides with an existing file, a previously generated file, or a duplicate within the same `generate()` batch. A `LargeSourceSet.onGenerateCollision()` callback allows the test framework to raise an assertion failure, helping recipe authors catch this during development.

- Closes #6771

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -34,6 +34,7 @@ import org.openrewrite.table.SearchResults;
 import org.openrewrite.table.SourcesFileErrors;
 import org.openrewrite.table.SourcesFileResults;
 
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.*;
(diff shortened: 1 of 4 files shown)


rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -151,6 +152,25 @@ public LSS generateSources(LSS sourceSet) {
                     try {
                         List<SourceFile> generated = new ArrayList<>(scanningRecipe.generate(scanningRecipe.getAccumulator(rootCursor, ctx), unmodifiableList(acc), ctx));
                         generated.replaceAll(source -> addRecipesThatMadeChanges(recipeStack, source));
+                        Set<Path> seenInThisBatch = new HashSet<>();
```
sambsnyd, line 155: This prevents the same `generate()` invocation from producing the same file twice, but doesn't seem to prevent a file from being generated with a path matching an existing file. The scanning phase of the run could populate this map so we could also avoid overwriting an existing file.

pstreef: The existing file case is covered — the `removeIf` block checks `sourceSet.getBefore(sourcePath)` first (line 157), which catches collisions with files already in the source set before the `seenInThisBatch` check. The `generatesFileCollidingWithExistingFile` test verifies this path.

PR #6781, opened 2026-02-20
Fix Java 21 parser dropping `var` in record pattern binding variables

cuioss-oliver:
## Problem

The Java 21 parser drops the `var` keyword when parsing record pattern binding variables like `Point(var x, var y)`. The `var` ends up in the whitespace prefix of the variable name instead of being recognized as a type identifier.

This causes any recipe that round-trips the AST (e.g. `AutoFormat`) to silently strip `var`, producing code that won't compile under `--release 21` (unnamed patterns require Java 22+ / JEP 456).

## Cause

In `visitVariables()`, when processing a binding variable from a record deconstruction pattern, `vartype` is `null` and `declaredUsingVar()` returns `false` (because it checks `vartype != null`). The `var` keyword is never consumed as a token, so `sourceBefore(variableName)` absorbs it into the variable's whitespace prefix.

## Fix

Extend the existing `declaredUsingVar()` fallback to also detect `var` via a source-level check when the variable has the `MATCH_BINDING` flag (set by javac for pattern binding variables). Resolve the type from `node.sym.type` when `vartype` is null.

## Testing

- Added TCK test `shouldParseRecordPatternWithVarKeyword` that verifies parser round-trip for `Point(var x, var y)`
- Full `rewrite-java-21:compatibilityTest` suite passes with zero regressions (including all lambda parameter `var` tests)


timtebeek:
Thanks a lot for the immediate fix here @cuioss-oliver ! I've replicated the same to the Java 25 parser, unless you had intentionally left it out there?

cuioss-oliver:
@timtebeek : I left it out, because locally I did only reproduce it for the Java 21 Parser, but I did see it for Java 25 as well, but wasn't sure about adding it (At Java 25 this might be not a bug)

timtebeek:
Then let's keep the parser implementations in sync, because we want to be print idempotent, even if it might not be an issue to drop the `var` on Java 25. :) Thanks a lot for the help here!

PR #6783, opened 2026-02-20
Fix `ChangeManagedDependencyGroupIdAndArtifactId` overwriting `artifactId` with glob pattern

timtebeek:
## Summary

- Fix `ChangeManagedDependencyGroupIdAndArtifactId` incorrectly overwriting the `artifactId` XML tag with a literal glob string (e.g. `swagger-*`) when `ChangeDependencyGroupIdAndArtifactId` is invoked with a wildcard `oldArtifactId` and `newArtifactId=null`
- Only change `groupId`/`artifactId` tags when the new value actually differs from the old value; when they match (glob passthrough), preserve the original tag value
- Read actual `groupId`/`artifactId` from the XML tag for version resolution instead of using potentially-glob values

## Test plan

- [x] Added `globArtifactIdRetainedWhenNewMatchesOld` test in `ChangeManagedDependencyGroupIdAndArtifactIdTest`
- [x] All existing `ChangeManagedDependencyGroupIdAndArtifactIdTest` tests pass
- [x] All existing `ChangeDependencyGroupIdAndArtifactIdTest` tests pass
- [x] Full `rewrite-maven` test suite passes

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
@@ -54,13 +54,17 @@ public class ChangeManagedDependencyGroupIdAndArtifactId extends Recipe {
     String oldArtifactId;
 
     @Option(displayName = "New groupId",
-            description = "The new groupId to use.",
-            example = "corp.internal.openrewrite.recipe")
+            description = "The new groupId to use. Defaults to the existing group id.",
+            example = "corp.internal.openrewrite.recipe",
+            required = false)
+    @Nullable
     String newGroupId;
 
     @Option(displayName = "New artifactId",
-            description = "The new artifactId to use.",
-            example = "rewrite-testing-frameworks")
+            description = "The new artifactId to use. Defaults to the existing artifact id.",
+            example = "rewrite-testing-frameworks",
+            required = false)
+    @Nullable
     String newArtifactId;
 
     @Option(displayName = "New version",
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
steve-aom-elliott, line 207: May wish to check the logic here given the it no longer falls back to the `oldGroupId` / `oldArtifactId` when passing `newGroupId` / `newArtifactId` when passing these values to this recipe

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
```diff
@@ -147,19 +149,29 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
+                    }
                     if (newVersion != null) {
                         try {
                             Optional<Xml.Tag> versionTag = t.getChild("version");
                             if (versionTag.isPresent()) {
-                                String resolvedArtifactId = newArtifactId;
+                                String resolvedGroupId = t.getChildValue("groupId").orElse(newGroupId != null ? newGroupId : oldGroupId);
+                                String resolvedArtifactId = t.getChildValue("artifactId").orElse(newArtifactId != null ? newArtifactId : oldArtifactId);
                                 if (resolvedArtifactId.contains("${")) {
                                     ResolvedPom pom = getResolutionResult().getPom();
```
steve-aom-elliott: Could reduce some repetition if we extract this outside of the if and use it here, in the next if and for the `getValue(..)` call for setting the `resolvedNewVersion`

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -631,6 +631,61 @@ void doesNotMakeChangeWhenChangingBomImportManagedDependency() {
...
 
+    @Test
+    void globArtifactIdRetainedWhenNewArtifactIdIsNull() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeManagedDependencyGroupIdAndArtifactId(
+            "javax.activation",
+            "javax.activation-*",
+            "jakarta.activation",
+            null,
+            null
```
steve-aom-elliott, line 642: I think this may just be an unfortunate example, as the `jakarta.activation:javax.activation-api` isn't a real artifact, but `jakarta.activation:jakarta.activation-api` is, so perhaps reusing the `io.swagger:swagger-annotations` -> `io.swagger.core.v3:swagger-annotations` would be optimal

timtebeek: Thanks! Done.

PR #6785, opened 2026-02-20
Add ChangeRepository recipe for Gradle build scripts

Jenson3210:
## Summary

- Adds a new `ChangeRepository` recipe to replace or remove repository declarations in `build.gradle` and `build.gradle.kts` files
- Enables organizations to standardize repository usage across all projects (e.g. migrate from `jcenter` to `mavenCentral`, consolidate internal Nexus URLs)
- Handles deduplication: if the target repository already exists, removes the old one instead of creating a duplicate

## Solution

All four parameters are optional, supporting flexible matching and replacement:

| Parameter | Description |
|-----------|-------------|
| `oldType` | Match by type (`jcenter`, `maven`, etc.). Omit to match any type. |
| `oldUrl` | Match by URL. Omit to match any URL of the given type. |
| `newType` | Replacement type. Omit (with no `newUrl`) to remove the match. |
| `newUrl` | Replacement URL. |

Supported transformations:
- **Named → Named**: `jcenter()` → `mavenCentral()`
- **Named → Custom**: `jcenter()` → `maven { url = "https://..." }`
- **Custom → Named**: `maven { url = "..." }` → `mavenCentral()`
- **Custom → Custom**: URL change while keeping the type
- **Remove**: match and delete without replacement
- Both **Groovy** and **Kotlin DSL**

### Example usage for org-wide uniformity

```yaml
name: com.myorg.StandardizeRepositories
recipeList:
  - org.openrewrite.gradle.ChangeRepository:
      oldType: jcenter
      newType: mavenCentral
  - org.openrewrite.gradle.ChangeRepository:
      oldUrl: https://old-nexus.internal.com/releases
      newType: maven
      newUrl: https://artifactory.myorg.com/releases
```

## Test plan

- [x] 23 tests covering all transformation types, both DSLs, idempotency, deduplication, URL-only matching, and removal mode


rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeRepositoryTest.java
```diff
@@ -0,0 +1,539 @@
...
+              repositories {
+                  mavenLocal()
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void removeNamedRepository() {
```
shanman190: This feels a little weird for a "Change" recipe to be removing entries. We do also already have `RemoveRepository`, so maybe remove isn't necessary for this recipe?

I think changing to something that is already present is fine to trigger a removal for though, so as to not lead to a duplicate definition.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeRepository.java
```diff
@@ -0,0 +1,329 @@
...
+    @Override
+    public String getDisplayName() {
+        return "Change repository";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace or remove a repository in Gradle build scripts to standardize repository usage across an organization. " +
+               "When no new type or URL is specified, the matching repository will be removed.";
+    }
```
shanman190: nit: For these we should utilize the `=` syntax that was adopted in other recipes.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeRepository.java
```diff
@@ -0,0 +1,329 @@
...
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.toList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class ChangeRepository extends Recipe {
```
shanman190, line 41: callout: We should add a validation to check that old and new repositories are not the same.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeRepository.java
```diff
@@ -0,0 +1,329 @@
...
+            required = false,
+            example = "https://new-nexus.example.com/releases")
+    @Nullable
+    String newUrl;
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        MethodMatcher repoMatcher = new MethodMatcher("org.gradle.api.artifacts.dsl.RepositoryHandler " + (oldType != null ? oldType : "*") + "(..)", true);
+
+        return Preconditions.check(new IsBuildGradle<>(), new JavaIsoVisitor<ExecutionContext>() {
```
shanman190, line 94: optional: We have a couple of recipes now that interact with repositories. Should we consider wrapping the discovery portion of this up into a `Trait` instead?

Jenson3210:
Thanks for the early review @shanman190! Was a draft as I did not look at CC's output before heading into the weekend. Will take this into account when asking him to "do more"! 🙏

shanman190:
No worries at all; I figured that I'd just write up some comments to help out. 🙂

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeRepository.java
Jenson3210: I think that is definitely a possibility here. 
Just wondering if we should do this in this PR or in a next one. 
We have a customer usecase atm for this functionality where a trait might take a larger refactor/bigger PR/ longer throughput. 
So thinking we could potentially split the Trait in a different PR and already release the recipe functionality. Internal kitchen of it can still change.

Jenson3210:
@timtebeek This recipe has been shared with a customer earlier as they had this need. We did not have new efforts in this. We can close as I think we can re-create this one as it is quite simple... WDYT?

timtebeek:
Thanks! Then let's close indeed for now.

PR #6786, opened 2026-02-20
Annotation attribute non jvm class misses import

BoykoAlex:
Unit test demonstrates the issue


timtebeek (approved):
Nice! Thanks for pulling up that `JavaType.ShallowClass.build`; good to support custom classes in that way.

PR #6789, opened 2026-02-22
Add Java 26 early-access support via rewrite-java-next module

timtebeek:
## Summary

Adds a new `rewrite-java-next` module to support Java 26 early-access builds. This enables early validation of the OpenRewrite parser against pre-release JDK versions. The module is structured as a copy of `rewrite-java-25` with version references updated to 26.

## Changes

- New `rewrite-java-next` module with Java 26 toolchain and parser implementation
- Updated `JavaParser` version detection cascade to prefer `JavaNextParser` for JDK 26+
- Extended CI workflows to test against Java 26-ea
- All version-specific classes renamed from `Java25` to `JavaNext` (e.g., `JavaNextParser`, `ReloadableJavaNextParser`)

### DIFF
--- a/rewrite-java-next/src/main/java/org/openrewrite/java/isolated/ReloadableJavaNextParser.java
+++ b/rewrite-java-next/src/main/java/org/openrewrite/java/isolated/ReloadableJavaNextParser.java
@@ -111,6 +111,9 @@ public class ReloadableJavaNextParser implements JavaParser {
         Options.instance(context).put("-proc", "none");
         Options.instance(context).put("-parameters", "true");
 
+        // Enable preview features so the parser can handle upcoming language features
+        Options.instance(context).put("--enable-preview", "");
+
         // Ensure type attribution continues despite errors in individual files or nodes.
         // If an error occurs in a single file or node, type attribution should still proceed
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     uses: openrewrite/gh-automation/.github/workflows/ci-gradle.yml@main
     with:
       java_version: |
+        26
```
timtebeek: Guess we need to wait a bit longer before 26-ea becomes available.

```suggestion
        26-ea
```

.github/workflows/publish.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     uses: openrewrite/gh-automation/.github/workflows/publish-gradle.yml@main
     with:
       java_version: |
+        26
```
timtebeek:
```suggestion
        26-ea
```

timtebeek:
It looks like we should be good to support the non-preview features in Java 26 already, and even the `JEP 530: Primitive Types in Patterns, instanceof, and switch` does not require parser changes.

We'll defer this effort until later, such that we don't have to unnecessarily maintain yet another Java parser with any changes.

Issue #6770, opened 2026-02-19
Maven add missing version to a dependency


wesboe:
## What problem are you trying to solve?
Spring boot and other frameworks manage dependencies. It can happen the dependency is removed from the dependencies management part. After migration to the new parent version, your application will complain the version of the dependency is missing. This happens when you migrate to Spring Boot 4 for dependencies spring-retry and rest-assured.

## What precondition(s) should be checked before applying this recipe?
If the dependency has a version or not.

## Description recipe
The recipe should follow the input of other maven recipes like upgradeparentversion.
- groupId
- artifactId
- newVersion
- versionPattern
Optional is acceptTransitive like with recipe AddDependency. To not add the version if dependency is found in management dependency.

## Draft recipe
`
@Value
@EqualsAndHashCode(callSuper = false)
public class ExplicitAddDependencyVersion extends Recipe {

    @Getter
    final String displayName = "Add version to dependency";

    @Getter
    final String description =
            "Add version to a dependency that is missing it. This can be required when dependency is not managed anymore by a parent pom.";

    @Option(
            displayName = "Group",
            description = "The first part of a dependency coordinate `com.google.guava:guava:VERSION`.",
            example = "com.google.guava")
    String groupId;

    @Option(
            displayName = "Artifact",
            description = "The second part of a dependency coordinate `com.google.guava:guava:VERSION`.",
            example = "guava")
    String artifactId;

    @Option(
            displayName = "Version",
            description = "An exact version number or node-style semver selector used to select the version number.",
            example = "29.X")
    String newVersion;

    @Override
    public Validated<Object> validate() {
        Validated<Object> validated = super.validate();
        //noinspection ConstantConditions
        if (newVersion != null) {
            validated = validated.and(Semver.validate(newVersion, null));
        }

        return validated;
    }

    @Override
    public TreeVisitor<?, ExecutionContext> getVisitor() {
        return new MavenIsoVisitor<>() {
            @Override
            public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                Xml.Tag t = super.visitTag(tag, ctx);
                if (isDependencyTag(groupId, artifactId)) {
                    if (t.getChildValue("version").isEmpty()) {
                        Xml.Tag versionTag = Xml.Tag.build("<version>" + newVersion + "</version>");

                        t = (Xml.Tag)
                                new AddToTagVisitor<>(t, versionTag, new MavenTagInsertionComparator(t.getChildren()))
[...]

PR #6790, opened 2026-02-22
Add test for ChangeParentPom retaining managed version from grandparent

timtebeek:
## Context

- In https://github.com/openrewrite/rewrite/issues/6770 it was reported that `ChangeParentPom` would not add an explicit version for `spring-retry` when upgrading `spring-boot-starter-parent` from 3.5.9 to 4.0.2, since Spring Boot 4 no longer manages `spring-retry`.

## What this PR does

Adds a test `bringsDownRemovedManagedVersionFromGrandparent` that verifies `ChangeParentPom` correctly retains managed versions even when the dependency management comes through a grandparent POM (e.g. `spring-boot-starter-parent` → `spring-boot-dependencies`).

The test confirms that `ChangeParentPom` **does** correctly:
1. Detect that `spring-retry` is managed by `spring-boot-starter-parent:3.5.9` (via `spring-boot-dependencies:3.5.9`)
2. Detect that `spring-boot-starter-parent:4.0.2` no longer manages `spring-retry`
3. Add `spring-retry` with explicit version to `<dependencyManagement>` in the child POM

This is a regression test complementing the existing `bringsDownRemovedManagedVersion` test which only covers the single-level parent case (`spring-boot-dependencies` directly).

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
@@ -1564,6 +1564,69 @@ class ChangeParentPomTest implements RewriteTest {
               )
             );
         }
+
+        @Issue("https://github.com/openrewrite/rewrite/issues/6770")
+        @Test
+        void bringsDownRemovedManagedVersionFromGrandparent() {
+            rewriteRun(
+              spec -> spec.recipe(new ChangeParentPom(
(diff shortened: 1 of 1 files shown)


timtebeek:
The context here is that a community user had reported an issue, but we've not yet been able to replicate it; we can keep this open for a short while to make it easier to suggest updates to the test, or otherwise close the PR again.

timtebeek:
No further details were received; closing this effort at least for now.

PR #6792, opened 2026-02-22
Add Kotlin script template support for Gradle KTS

timtebeek:
## Summary

Leverage K2 FIR compiler extensions to configure implicit receivers and default imports for `build.gradle.kts` and `settings.gradle.kts` files. This improves type information for Gradle DSL methods, enabling better recipe matching and method resolution.

## Changes

- **ScriptCompilerPlugin.kt** — New K2 compiler plugin registering FIR extensions for script configuration
- **KotlinParser.java** — Added `scriptImplicitReceivers()` and `scriptDefaultImports()` builder methods
- **GradleParser.java** — Configured both KTS parsers with `Project`/`Settings` receivers and Gradle default imports
- **DefaultImportsCustomizer.java** — Made `DEFAULT_IMPORTS` public for reuse

## Test Plan

- ✅ All existing `rewrite-kotlin` tests pass
- ✅ All existing `rewrite-gradle` tests pass
- ✅ Gradle KTS tests specifically verified (MigrateDependenciesToVersionCatalogKotlinDSLTest, etc.)

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeListing.java
@@ -26,8 +26,6 @@ import org.openrewrite.config.RecipeDescriptor;
 import java.time.Duration;
 import java.util.*;
 
-import static java.util.Collections.singleton;
-
 @Getter
 @RequiredArgsConstructor
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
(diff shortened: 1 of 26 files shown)


rewrite-gradle/src/main/groovy/Plugin.groovy
```diff
@@ -13,11 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-interface PluginSpec {
+import org.gradle.plugin.use.PluginDependenciesSpec
+import org.gradle.plugin.use.PluginDependencySpec
+
+interface PluginSpec extends PluginDependenciesSpec {
```
timtebeek, line 19: Specifically for plugins we're not yet seeing the new improved types available, but I figured we can iterate on that in a subsequent PR and already get the improved types for dependencies merged early.

shanman190: That's likely because our local method is being preferred over the parent method as the only difference is the return type.

PR #6797, opened 2026-02-23
DeleteProperty: preserve inline comments when deleting entries

Jenson3210:
## Summary
- `DeleteProperty` corrupted inline YAML comments when deleting entries from a mapping
- The newline fix-up logic used `startsWithNewline(prefix)` which missed prefixes containing inline comments (e.g. ` # comment\n    ` starts with a space)
- Changed to `containsNewline(prefix)` so prefixes that already contain a line break are left intact

## What changed
Inline comments in YAML are stored in the prefix of the **next** element. For example:
```yaml
keep: yes # inline comment
other: data
```
The `other` entry's prefix is `" # inline comment\n    "`. When `DeleteProperty` removed a sibling entry, it incorrectly prepended `\n` to this prefix (because it doesn't start with `\n`), moving the comment onto its own line:
```yaml
keep: yes
 # inline comment
    other: data
```

The fix changes all three occurrences of `startsWithNewline` to `containsNewline` in `DeleteProperty`, covering `visitMapping` (both the inline deletion loop and the post-deletion fix-up) and `visitSequence`.

## Test plan
- [x] Added `deleteLastEntryPreservesInlineCommentOnPreviousEntry` test in `DeletePropertyKeyTest`
- [x] All existing `DeletePropertyKeyTest` tests pass (27 total)
- [x] Full `rewrite-yaml` test suite passes

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
@@ -193,15 +193,17 @@ public class DeleteProperty extends Recipe {
                 }
 
                 // Preserve inline comment from deleted trailing entries on the last kept entry
-                if (trailingInlineComment != null && !entries.isEmpty()) {
-                    int lastIndex = entries.size() - 1;
-                    Yaml.Mapping.Entry lastKept = entries.get(lastIndex);
-                    if (lastKept.getValue() instanceof Yaml.Scalar &&
-                        ((Yaml.Scalar) lastKept.getValue()).getStyle() == Yaml.Scalar.Style.PLAIN) {
-                        Yaml.Scalar scalar = (Yaml.Scalar) lastKept.getValue();
-                        entries.set(lastIndex, lastKept.withValue(
-                                scalar.withValue(scalar.getValue() + trailingInlineComment)));
-                    }
+                if (trailingInlineComment != null) {
+                    String comment = trailingInlineComment;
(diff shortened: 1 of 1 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/DeleteProperty.java
```diff
@@ -238,8 +256,18 @@ private static boolean containsOnlyWhitespace(@Nullable String str) {
...
+
+    /**
+     * Extract an inline comment from a prefix string. Inline comments appear before the
+     * first newline in the prefix (e.g. {@code " # comment\n  "} → {@code " # comment"}).
+     */
+    private static @Nullable String extractInlineComment(String prefix) {
+        int newlineIndex = prefix.indexOf('\n');
+        String beforeNewline = newlineIndex >= 0 ? prefix.substring(0, newlineIndex) : prefix;
+        return beforeNewline.contains("#") ? beforeNewline : null;
     }
```
timtebeek, line 273: @Jenson3210 this is new but I'd like your eyes on whether it's correct; the test we had before didn't correctly reproduce the issue as there as an other item between the last removed and the comment.

Jenson3210: Ah, I think there were 2 bugs then: 
- the starts with in these examples were evaluating to false as the comments did not start with a new line -> line moved to next line
- the removal also removed the comment
This indeed looks like it would be better as we will now also keep the comments. Thanks!

PR #6799, opened 2026-02-23
Use flexible version assertions in ChangeDependencyGroupIdAndArtifactId tests

Jenson3210:
## Summary

- Switch 3 tests in `ChangeDependencyGroupIdAndArtifactIdTest` from hardcoded version assertions to dynamic pattern matching using `spec.after()` with regex and AssertJ

## Problem

Three tests fail because they assert `<version>2.21.0</version>` for `jackson-jakarta-rs-json-provider`, but the `2.x` version pattern now resolves to `2.21.1` on Maven Central. These tests break every time a new Jackson 2.x version is published.

## Solution

Use the established `spec.after(after -> { ... })` callback pattern (already used in `UpgradeDependencyVersionTest` and others) to:
1. Extract the actual resolved version from the output XML using regex
2. Assert it matches `2.\d+.\d+` via AssertJ
3. Return the expected XML with the extracted version interpolated

Tests fixed:
- `versionNotDroppedWhenDependencyExcludedByOtherDependency`
- `versionNotDroppedWhenTransitiveDependencyHasManagedVersion`
- `exclusionUpdatedWhenDependencyGroupIdChanges`

## Test plan

- [x] All 55 tests in `ChangeDependencyGroupIdAndArtifactIdTest` pass
- [x] No other changes outside the test file

### DIFF
--- a/rewrite-python/rewrite/pyproject.toml
+++ b/rewrite-python/rewrite/pyproject.toml
@@ -24,7 +24,7 @@ requires-python = ">=3.10"
 dependencies = [
     "cbor2>=5.6.5",
     "more_itertools>=10.0.0",
-    "ty-types>=0.0.19.dev20260223102528",  # Type inference CLI for Python type attribution
+    "ty-types>=0.0.19.dev20260223122104",  # Type inference CLI for Python type attribution
     "parso>=0.7.1,<0.8",  # Python 2/3 parser with CST support (0.8+ dropped Python 2.7 grammar)
 ]
 
--- a/rewrite-python/rewrite/src/rewrite/python/remove_import.py
+++ b/rewrite-python/rewrite/src/rewrite/python/remove_import.py
@@ -20,7 +20,7 @@ from uuid import uuid4
 
 from rewrite.java import J
 from rewrite.java.support_types import JContainer, JRightPadded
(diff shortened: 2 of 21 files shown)


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -3329,33 +3333,39 @@ void versionNotDroppedWhenDependencyExcludedByOtherDependency() {
...
-                          <version>2.21.0</version>
-                      </dependency>
-                  </dependencies>
-              </project>
-              """
+            spec -> spec.after(after -> {
+                Matcher matcher = Pattern.compile("jackson-jakarta-rs-json-provider</artifactId>\\s*<version>(.+)</version>").matcher(after);
+                assertThat(matcher.find()).isTrue();
+                String version = matcher.group(1);
+                assertThat(version).matches("2\\.\\d+\\.\\d+");
```
timtebeek: Wherever possible I'd prefer that we move to `after -> assertThat(after).......actual()`

timtebeek (approved):
Thanks a lot! Nice not to have to maintain these anymore with new releases.

PR #5746, opened 2025-07-11
Process exclusion with ChangeDependencyGroupIdAndArtifactId

gliptak:
## What's changed?

Process exclusion with ChangeDependencyGroupIdAndArtifactId

## What's your motivation?

closes #2933 

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

PR #6800, opened 2026-02-23
Preserve old exclusions and add sibling for new coordinates on dependency change

Jenson3210:
## Summary

- `ChangeDependencyGroupIdAndArtifactId` now preserves existing exclusions for old coordinates and adds a sibling exclusion for the new coordinates
- Previously, old exclusions were replaced with new coordinates via `ChangeExclusion`, which broke transitive dependency blocking

## Background

[PR #5746](https://github.com/openrewrite/rewrite/pull/5746) introduced the `ChangeExclusion` recipe and wired it into `ChangeDependencyGroupIdAndArtifactId`, so that when renaming a dependency, any exclusions referencing the old coordinates would also be renamed to the new coordinates. [PR #6595](https://github.com/openrewrite/rewrite/pull/6595) further refined this behavior.

The problem is that this replacement removes the exclusion for the _old_ artifact, which allows the old (potentially vulnerable) dependency to leak back in through transitive dependencies.

## Problem

Consider a project that excludes `commons-lang:commons-lang` from `struts2-core` to block a vulnerable transitive dependency:

```xml
<dependency>
    <groupId>org.apache.struts</groupId>
    <artifactId>struts2-core</artifactId>
    <exclusions>
        <exclusion>
            <groupId>commons-lang</groupId>
            <artifactId>commons-lang</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

When `ChangeDependencyGroupIdAndArtifactId` renames `commons-lang:commons-lang` → `org.apache.commons:commons-lang3`, the old behavior **replaced** the exclusion:

```xml
<!-- Old behavior: vulnerable commons-lang is no longer blocked! -->
<exclusion>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-lang3</artifactId>
</exclusion>
```

Since `struts2-core` still transitively depends on `commons-lang:commons-lang`, the old vulnerable artifact now leaks through because nothing excludes it anymore. The exclusion was there specifically to prevent the old artifact from entering the dependency tree, and replacing it with the new coordinates defeats that purpose entirely.

## Solution

Instead of replacing exclusions, the recipe now:
1. **Preserves** the old exclusion (to keep blocking the old transitive dep)
2. **Adds a sibling exclusion** for the new coordinates alongside it

```xml
<!-- New behavior: both old and new artifacts are excluded -->
<exclusions>
    <exclusion>
        <groupId>commons-lang</groupId>
        <artifactId>commons-lang</artifactId>
    </exclusion>
    <exclusion>
        <groupId>org.apache.commons</groupId>
        <artifactId>commons-lang3</artifactId>
    </exclusion>
</exclusions>
```

## Test plan

- [x] New test `exclusionPreservedAndSiblingAddedForNewCoordinates` for the commons-lang scenario
- [x] Updated test `exclusionPreservedAndSiblingAddedWhenDependencyGroupIdChanges` for the Jackson/Jakarta scenario
- [x] Wildcard exclusion tests still pass (no sibling added for `<artifactId>*</artifactId>`)
- [x] All other `ChangeDependencyGroupIdAndArtifactIdTest` tests pass
- [x] `ChangeExclusionTest` and `ChangeManagedDependencyGroupIdAndArtifactIdTest` unaffected

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
@@ -284,15 +284,18 @@ public class ChangeDependencyGroupIdAndArtifactId extends ScanningRecipe<ChangeD
                             boolean hasOldExclusion = t.getChildren("exclusion").stream()
                                     .anyMatch(e -> matchesGlob(e.getChildValue("groupId").orElse(null), oldGroupId) &&
                                                    matchesGlob(e.getChildValue("artifactId").orElse(null), oldArtifactId));
+                            if (!hasOldExclusion) {
+                                return t;
+                            }
                             boolean hasNewExclusion = t.getChildren("exclusion").stream()
                                     .anyMatch(e -> effectiveNewGroupId.equals(e.getChildValue("groupId").orElse(null)) &&
                                                    effectiveNewArtifactId.equals(e.getChildValue("artifactId").orElse(null)));
-                            if (hasOldExclusion && !hasNewExclusion) {
+                            if (!hasNewExclusion) {
                                 t = (Xml.Tag) new AddToTagVisitor<>(t, Xml.Tag.build(
                                         "<exclusion>\n" +
-                                        "<groupId>" + effectiveNewGroupId + "</groupId>\n" +
-                                        "<artifactId>" + effectiveNewArtifactId + "</artifactId>\n" +
-                                        "</exclusion>"))
+                                                "<groupId>" + effectiveNewGroupId + "</groupId>\n" +
+                                                "<artifactId>" + effectiveNewArtifactId + "</artifactId>\n" +
+                                                "</exclusion>"))
                                         .visitNonNull(t, ctx, getCursor().getParentOrThrow());
                                 maybeUpdateModel();
                             }


steve-aom-elliott:
From the test failure, looks like we'd want to change `ChangeDependencyGroupIdAndArtifactIdTest.versionNotDroppedWhenDependencyExcludedByOtherDependency()` and `ChangeDependencyGroupIdAndArtifactIdTest.versionNotDroppedWhenTransitiveDependencyHasManagedVersion()` to use pattern matching for the version number if viable

Jenson3210:
> From the test failure, looks like we'd want to change `ChangeDependencyGroupIdAndArtifactIdTest.versionNotDroppedWhenDependencyExcludedByOtherDependency()` and `ChangeDependencyGroupIdAndArtifactIdTest.versionNotDroppedWhenTransitiveDependencyHasManagedVersion()` to use pattern matching for the version number if viable

Yeah, I pushed that in a separate PR as that one was blocking main. Now updates with main and tests are passing

rewrite-maven/src/test/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactIdTest.java
```diff
@@ -3530,13 +3531,75 @@ void exclusionUpdatedWhenDependencyGroupIdChanges() {
               </project>
               """,
             spec -> spec.after(actual -> assertThat(actual)
-              .containsPattern("<groupId>com\\.fasterxml\\.jackson\\.jakarta\\.rs</groupId>\\s*<artifactId>jackson-jakarta-rs-json-provider</artifactId>\\s*<version>2\\.\\d+\\.\\d+</version>")
+              // Old exclusion preserved
+              .contains("<groupId>com.fasterxml.jackson.jaxrs</groupId>")
+              .contains("<artifactId>jackson-jaxrs-json-provider</artifactId>")
+              // Sibling exclusion added for new coordinates
+              .contains("<groupId>com.fasterxml.jackson.jakarta.rs</groupId>")
+              .contains("<artifactId>jackson-jakarta-rs-json-provider</artifactId>")
```
steve-aom-elliott: Does this part actually check that it's an exclusion or might this just match on the direct dependency as well?

timtebeek:
Follow up to clean out unnecessary exclusions coming in
- https://github.com/openrewrite/rewrite/pull/6802

PR #6802, opened 2026-02-23
Add RemoveUnnecessaryExclusions recipe

Jenson3210:
## Summary
- Adds a declarative YAML recipe `org.openrewrite.maven.cleanup.RemoveUnnecessaryExclusions` that removes Maven dependency exclusions which are not necessary because the excluded dependency is not actually a transitive dependency of the project
- Wraps the existing `RemoveExclusion` recipe with `onlyIneffective: true` and wildcard matchers
- Adds the new recipe to `org.openrewrite.maven.BestPractices`

## Test plan
- [x] New `RemoveUnnecessaryExclusionsTest` with 3 test cases: removes ineffective exclusion, retains effective exclusion, mixed scenario
- [x] Existing `RemoveExclusionTest` passes (no regressions)


E14:
Maybe odd question... This was closed and marked done - but I can't find this recipe in current sources, did you add this to a different artifact?

PR #6804, opened 2026-02-23
Add type attribution for Gradle plugin DSL in KTS files

timtebeek:
## Summary
Adds Kotlin stubs to GradleParser to provide type information for the `plugins {}` block in .gradle.kts files, enabling recipes to use MethodMatcher consistently for both Groovy and KTS. Removed `isKotlin` branching patterns from RemovePluginVisitor and updated FindPlugins to use real Gradle API types instead of synthetic stubs.

## Changes
- **GradleParser.java**: Added KTS stub strings defining extension functions (`Project.plugins()`, `Settings.plugins()`, `PluginDependenciesSpec.kotlin()`) via `dependsOn()` mechanism to provide K2 compiler with type context for plugin DSL methods.
- **RemovePluginVisitor.java**: Removed `isKotlin` checks and Kotlin tree import. Now uses name-based check for `plugins` container (unambiguous in Gradle files) and MethodMatcher with `matchUnknownTypes=true` for inner calls, working consistently for both Groovy and KTS.
- **FindPlugins.java**: Updated matchers from synthetic stub types (`PluginSpec`, `Plugin`) to real Gradle API types (`PluginDependenciesSpec`, `PluginDependencySpec`) with `matchOverrides=true`.

## Testing
All `:rewrite-gradle:test` tests pass including existing Groovy and KTS plugin tests.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
@@ -37,7 +37,7 @@ import static java.util.Collections.singletonList;
 @RequiredArgsConstructor
 public class GradleParser implements Parser {
     @SuppressWarnings("LanguageMismatch")
-    private static final String KTS_BUILD_STUBS = "" +
+    private static final String KTS_BUILD_STUBS =
             "package org.gradle.api\n" +
             "import org.gradle.plugin.use.PluginDependenciesSpec\n" +
             "import org.gradle.plugin.use.PluginDependencySpec\n" +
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/GradleParser.java
```diff
@@ -36,6 +36,24 @@
...
+
+    @SuppressWarnings("LanguageMismatch")
+    private static final String KTS_SETTINGS_STUBS =
+            "package org.gradle.api.initialization\n" +
+            "import org.gradle.plugin.use.PluginDependenciesSpec\n" +
+            "import org.gradle.plugin.use.PluginDependencySpec\n" +
+            "import org.gradle.plugin.management.PluginManagementSpec\n" +
+            "fun Settings.plugins(block: PluginDependenciesSpec.() -> Unit) {}\n" +
+            "fun Settings.pluginManagement(block: PluginManagementSpec.() -> Unit) {}\n" +
+            "fun PluginDependenciesSpec.kotlin(module: String): PluginDependencySpec = id(module)\n";
```
shanman190, line 55: So there are a lot of extension functions that help with various levels of type attribution.
https://github.com/gradle/gradle/tree/master/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl

Are we thinking that we'll just add to this one by one or do we need to have something more sophisticated?

timtebeek: I didn't yet have a structured idea on adding more; figured the ones we have are most often used, also in our recipes. We can add others as needed, but figured keep it small for now.

PR #6805, opened 2026-02-24
Handle null class body in KotlinPrinter

Jenson3210:
## Summary

- Fix NPE in `KotlinPrinter.visitClassDeclaration0` when printing a class declaration with null body
- This crash occurs when a context-sensitive `JavaTemplate` is applied inside a Kotlin class: the `BlockStatementTemplateGenerator` creates a stripped-down class via `withBody(null)` and prints it to build the template context, but the Kotlin printer didn't handle null body
- Add regression test that applies a context-sensitive template inside a Kotlin class method

## Test plan

- [x] New test `KotlinTemplateTest.addStatementToMethodInClass` reproduces the NPE without the fix and passes with it
- [x] All existing `KotlinTemplateTest` tests pass


MBoegers (approved):
Thanks for the prompt reaction on this <3

PR #6818, opened 2026-02-25
Surface actual parse error in MergeYaml via Validated API

timtebeek:
## Summary
- When `MergeYaml` encountered a YAML parsing failure, the old `maybeParse` method silently discarded the `ParseError` result (via a `Yaml.Documents` instanceof filter), producing only a generic "Must be valid YAML" message
- Now extracts the actual error from the `ParseExceptionResult` marker and surfaces it through the `Validated` API, making failures diagnosable (e.g. SnakeYAML syntax errors, print-idempotence failures)
- Introduces `validateParseable()` returning `Validated<Yaml>` so that `validate()` uses the Validated API instead of exception-based control flow
- `parse()` delegates to `validateParseable()`, eliminating duplicated parsing logic

## Test plan
- [x] `invalidYaml()` test updated to match new validation message
- [x] `sourceNull()` test still passes with existing "Must be valid YAML" message
- [x] All existing `MergeYamlTest` tests pass

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
@@ -95,7 +95,7 @@ public class MergeYaml extends Recipe {
         this.createNewKeys = createNewKeys;
     }
 
-    public enum InsertMode { Before, After, Last }
+    public enum InsertMode {Before, After, Last}
 
     @Nullable
     @NonFinal
@@ -118,6 +118,7 @@ public class MergeYaml extends Recipe {
                 .and(Validated.test("insertProperty", "Insert property must be filed when `insert mode` is either `BeforeProperty` or `AfterProperty`.", insertProperty,
                         s -> insertMode == null || insertMode == Last || !isBlank(s)));
     }
+
     String displayName = "Merge YAML snippet";
 
     @Override


timtebeek:
We saw this somehow pass validation but fail during execution, so rewrote the logic to capture and forward any parse failure message.

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYaml.java
```diff
@@ -111,8 +109,12 @@ public Validated<Object> validate() {
                             if (yaml == null) {
                                 return false;
                             }
-                            MergeYaml.maybeParse(yaml).ifPresent(it -> incoming = it);
-                            return incoming != null;
+                            try {
+                                incoming = MergeYaml.parse(yaml);
+                                return true;
+                            } catch (IllegalArgumentException e) {
```
MBoegers: It feels strange to have an exception to signal failure. Failure is nothing special its expected in validation.
Also, creating an exception causes stack walking to fill the stacktrace and JIT deopt, maybe not very strong arguments in our case but still.

We could omit the exception and instead return the parsing error or erroneous and check with instaceof?

MBoegers: or turn the parse into a validateParseable method and return a validation result insteadt?

timtebeek:
@MBoegers any chance you could review? Or should I ask someone else?

PR #6823, opened 2026-02-25
Migrate from services.gradle.org to downloads.gradle.org

timtebeek:
## Summary
- `services.gradle.org` now returns 301 redirects to `downloads.gradle.org`. Updated all distribution and version API URLs to use the new canonical domain directly, avoiding unnecessary redirect hops.
- The old `services.gradle.org` domain is still accepted in existing wrapper properties files for backward compatibility via a `startsWith` check in `GradleWrapper.determineGradleVersion`.
- Updated production code, tests, examples, and the project's own `gradle-wrapper.properties`.

## Test plan
- [ ] Verify `./gradlew :rewrite-gradle:test` passes (integration tests hit live `downloads.gradle.org` endpoints)
- [ ] Verify existing projects with `services.gradle.org` wrapper URLs are still handled correctly (backward compat via dual `startsWith` check)


steve-aom-elliott (approved):

kevin-wise:
I'm not sure this was a good change. The official wrapper still sets it to services.gradle.org.

timtebeek:
Thanks! Looks to have been a bit of unnecessary churn indeed; I've minimized the changes now in
- https://github.com/openrewrite/rewrite/pull/8915

PR #6828, opened 2026-02-26
UpgradePluginVersion: skip adding version when managed by pluginManagement

timtebeek:
## Summary
- When `addVersionIfMissing` is true, `UpgradePluginVersion` now checks whether the plugin already has a version managed via `<pluginManagement>` (local or inherited from a parent POM) before adding an explicit `<version>` tag
- This prevents unwanted version additions in child modules that inherit their plugin version from a parent
- Uses the same `pluginManagement` lookup pattern as `RemoveRedundantDependencyVersions.getManagedPluginVersion()`

- Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/992
- Follows https://github.com/openrewrite/rewrite-migrate-java/pull/989#issuecomment-3965457005

## Test plan
- [x] Added `shouldNotAddVersionWhenManagedByParent` test verifying no `<version>` is added to a child module when the parent's `<pluginManagement>` manages the plugin version
- [x] Verified the parent's managed version is still upgraded
- [x] All existing `UpgradePluginVersionTest` tests pass
- [x] `Java8toJava11Test` in `rewrite-migrate-java` passes (uses `addVersionIfMissing: true` for `maven-compiler-plugin`)

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradePluginVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradePluginVersion.java
@@ -156,8 +156,8 @@ public class UpgradePluginVersion extends Recipe {
             private boolean hasManagedPluginVersion(ResolvedPom resolvedPom, String groupId, String artifactId) {
                 for (Plugin p : ListUtils.concatAll(resolvedPom.getPluginManagement(),
                         resolvedPom.getRequested().getPluginManagement())) {
-                    if (Objects.equals(p.getGroupId(), groupId) &&
-                            Objects.equals(p.getArtifactId(), artifactId) &&
+                    if (p.getGroupId().equals(groupId) &&
+                            p.getArtifactId().equals(artifactId) &&
                             p.getVersion() != null) {
                         return true;
                     }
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradePluginVersion.java
```diff
@@ -146,6 +153,18 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 return super.visitTag(tag, ctx);
             }
 
+            private boolean hasManagedPluginVersion(ResolvedPom resolvedPom, String groupId, String artifactId) {
+                for (Plugin p : ListUtils.concatAll(resolvedPom.getPluginManagement(),
+                        resolvedPom.getRequested().getPluginManagement())) {
+                    if (Objects.equals(p.getGroupId(), groupId) &&
```
MBoegers: nit picking, why `Object.equals` instead of `groupId.equald(p.getGroupId())` as groupId is not nullable?

steve-aom-elliott (approved):

PR #6830, opened 2026-02-26
Preserve GString/StringTemplate in ChangeDependency when version unchanged

timtebeek:
## Summary
- Add `overrideManagedVersion` guard to GString and Kotlin StringTemplate code paths in `ChangeDependency`, matching the existing guard on the literal path
- When only group/artifact changes (version unchanged), preserve the interpolated string structure instead of collapsing to a literal with a pinned version
- Update two existing tests (`worksWithGString`, `kotlinDslStringInterpolation`) to expect the new behavior

## Problem
When `ChangeDependency` renames a dependency that uses a GString or Kotlin StringTemplate (e.g. `"group:artifact:${version}"`), it unconditionally collapsed the interpolated string to a literal and pinned the resolved version. This caused the Spring Boot 4 migration recipe to turn `"org.springframework.boot:spring-boot-starter-web:${springBootVersion}"` into `"org.springframework.boot:spring-boot-starter-webmvc:4.0.3"`, destroying the property reference.

The literal code path already had the correct guard: it only pins a new version when the original dependency had an explicit version or `overrideManagedVersion=true`. The GString and Kotlin StringTemplate paths were missing this guard.

- Fixes https://github.com/moderneinc/customer-requests/issues/1920

## Test plan
- [x] Existing `worksWithGString` test updated to expect GString preservation
- [x] Existing `kotlinDslStringInterpolation` test updated to expect StringTemplate preservation
- [x] All 19 `ChangeDependencyTest` tests pass
- [x] Integration test in rewrite-spring confirms Spring Boot 4 migration no longer pins starter versions

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -36,10 +36,13 @@ import org.openrewrite.java.tree.JavaSourceFile;
 import org.openrewrite.kotlin.tree.K;
 import org.openrewrite.maven.MavenDownloadingException;
 import org.openrewrite.maven.table.MavenMetadataFailures;
+import org.openrewrite.properties.PropertiesVisitor;
+import org.openrewrite.properties.tree.Properties;
 import org.openrewrite.semver.DependencyMatcher;
 import org.openrewrite.semver.Semver;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
@@ -58,6 +61,9 @@ public class ChangeDependency extends Recipe {
 
     @EqualsAndHashCode.Exclude
     transient MavenMetadataFailures metadataFailures = new MavenMetadataFailures(this);
+
+    @EqualsAndHashCode.Exclude
+    transient Map<String, String> pendingPropertyUpdates = new ConcurrentHashMap<>();
 
     @Option(displayName = "Old groupId",
             description = "The old groupId to replace. The groupId is the first part of a dependency coordinate 'com.google.guava:guava:VERSION'. Supports glob expressions.",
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -264,7 +264,7 @@ implementation platform("commons-lang:commons-lang:${version}")
 
               def version = '2.6'
               dependencies {
-                  implementation platform("org.apache.commons:commons-lang3:3.11")
+                  implementation platform("org.apache.commons:commons-lang3:${version}")
```
timtebeek: Now we're effectively not making any change, other than the markers; We should add `gradle.properties` files here and assert those are updated.

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -238,6 +239,12 @@ implementation platform("org.apache.commons:commons-lang3:3.11")
     void worksWithGString() {
         rewriteRun(
           spec -> spec.recipe(new ChangeDependency("commons-lang", "commons-lang", "org.apache.commons", "commons-lang3", "3.11.x", null, null, true)),
+          properties(
+            """
+              commonsLangVersion=2.6
```
timtebeek, line 244: This version should be changed to a specific version in the 3.11 range. Without it projects would break.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
Jenson3210, line 52: Will this "expectation" work without transitioning to a Scanning recipe? 
Which will bring a whole new set of potential issues due to the timing of scan/edit.

timtebeek: Honestly, I'm a bit fried looking at this PR; if you see a good way forward feel free to take over, or close this if there's a better alternative. With travel ahead I don't think I'll get to this one this week.

Jenson3210: The reason this works is because the build gradle files are handled before the properties file and the transient map is used like an accumulator. 
This is not an approach we want to adopt as we do not have this certainty, I guess.
The only way forward I see to support this is to change this recipe like we've changed all the other ones -> make it scanning.

timtebeek: Thanks! Then let's explore that separately and close this effort. Good to have clarity!

PR #6841, opened 2026-02-27
Skip recipe validation for imperative recipes in assertRecipesConfigure()

### LINKED ISSUE #6840: RecipeLoadingTest fails for recipes with optional parameters and custom validate()
## Problem

`assertRecipesConfigure()` fails for imperative recipes that have multiple optional parameters (`@Option(required = false)`) and override `validate()` to enforce that at least one is present.

### Root cause

`assertRecipesConfigure()` discovers all recipes via `ClasspathScanningLoader`, which instantiates imperative recipes with no arguments — all `String` `@Option` fields become `null`. It then passes these unconfigured recipes to `rewriteRun()`, which calls `recipe.validateAll()`. A custom `validate()` like:

```java
@Override
    return super.validate().and(
            Validated.required("optionA", optionA)
                    .or(Validated.required("optionB", optionB)));
}
```

returns `Invalid` because both optional params are `null`, and the test assertion at `RewriteTest.java:231` fails.

### Impact

Any recipe repo that has a `RecipeLoadingTest` calling `assertRecipesConfigure()` will fail if it contains an imperative recipe with this pattern. Recipes in the `rewrite` monorepo using the same pattern (e.g., `ChangeDependency`, `EnableDevelocityBuildCache`, `MoveFile`, `AddOciLabels`, `ChangeFrom`) are unaffected only because the monorepo has no bulk `RecipeLoadingTest`.

### Expected behavior

`assertRecipesConfigure()` should:
- **Validate declarative recipes** (from YAML) — their options come from YAML configuration and should be checked
- **Skip validation for imperative recipes** loaded with no arguments — they are not configured for execution, so validation failures are expected and meaningless

### Suggested fix

Add a `recipeValidation` flag to `RecipeSpec` (following the existing `serializationValidation` pattern) and gate `recipe.validateAll()` on it in `rewriteRun()`. In `assertRecipesConfigure()`, disable validation for non-`DeclarativeRecipe` instances.

MBoegers:
## Summary

- `assertRecipesConfigure()` now skips `validateAll()` for imperative recipes loaded with no user-provided arguments
- Declarative recipes (from YAML) still get full validation to catch misconfigured options
- Adds `recipeValidation` flag to `RecipeSpec` following the existing `serializationValidation` pattern

- Fixes #6840

## Test plan

- [x] `./gradlew :rewrite-test:test` passes
- [ ] Verify `RecipeLoadingTest` in a recipe repo with optional-param + validate() recipes no longer fails
- [ ] Verify a deliberately misconfigured declarative YAML recipe still fails the loading test

### DIFF
--- a/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
+++ b/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-wrapper/versions.csv
@@ -1,4 +1,5 @@
 version,gradlew,gradlewBat
+9.4.0,e5083bb6,9af86e4b
 9.4.0-milestone-4,e5083bb6,9af86e4b
 9.4.0-milestone-3,e5083bb6,9af86e4b
 9.4.0-milestone-2,e5083bb6,9af86e4b


rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -90,10 +91,21 @@ default void assertRecipesConfigure(String packageName) {
...
                 softly.assertThatCode(() -> {
                     try {
                         rewriteRun(
-                                spec -> spec.recipe(recipe),
+                                spec -> {
+                                    spec.recipe(recipe);
+                                    if (skipValidation) {
+                                        spec.validateRecipe(false);
+                                    }
+                                },
```
timtebeek: I'd have expected something along the lines of:
```suggestion
                                spec -> spec.recipe(recipe).validateRecipe(recipe instanceof DeclarativeRecipe),
```
Would that not work here?

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -90,10 +91,16 @@ default void assertRecipesConfigure(String packageName) {
             // scanRuntimeClasspath picks up all recipes in META-INF/rewrite regardless of whether their
             // names start with the package we intend to filter on here
             if (recipe.getName().startsWith(packageName)) {
+                // Imperative recipes are loaded with no user-provided arguments, so all optional
+                // parameters are null. Skip recipe validation for these since custom validate()
+                // methods (e.g. requiring at least one of several optional parameters) would
+                // fail on an unconfigured instance. Declarative recipes have their options
+                // configured from YAML and should still be validated.
+                boolean skipValidation = !(recipe instanceof DeclarativeRecipe);
```
timtebeek: Unused variable I think; should we use this instead of the inlined copy further down?

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -90,10 +91,16 @@ default void assertRecipesConfigure(String packageName) {
...
                 softly.assertThatCode(() -> {
                     try {
                         rewriteRun(
-                                spec -> spec.recipe(recipe),
+                                spec -> spec.recipe(recipe).validateRecipe(recipe instanceof DeclarativeRecipe),
```
timtebeek: If I understand correctly we only really have issues with recipes with options right? Should we then use
```suggestion
                                spec -> spec.recipe(recipe).validateRecipe(recipe.getOptionDescriptors().isEmpty()),
```

MBoegers: I understand the test as "Are the declarative recipes configured properly" That's why I filter by `DeclarativeRecipe`. But technically, yes, the problem can only occur for imperative recipes with options. Your call.

timtebeek: Then let's adopt the above, as it'll be harder to spot this selective check and reason for it otherwise later on.

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
MBoegers: no need, left over.

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
timtebeek: To be clear: I meant to adopt `validateRecipe(recipe.getOptionDescriptors().isEmpty())`

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -226,11 +232,13 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
         for (SourceSpec<?> s : sourceSpecs) {
             s.customizeExecutionContext.accept(ctx);
         }
-        List<Validated<Object>> validations = new ArrayList<>();
-        recipe.validateAll(ctx, validations);
-        assertThat(validations.stream().filter(Validated::isInvalid))
-                .as("Recipe validation must have no failures")
-                .isEmpty();
+        if (testClassSpec.recipeValidation && testMethodSpec.recipeValidation) {
```
timtebeek: Do we even need the fields added to `RecipeSpec` if we can do this check here?
```suggestion
        if (recipe instanceof DeclarativeRecipe || recipe.getDescriptor().getOptions().isEmpty()) {
```
I don't really see any other use cases to disable validation.

MBoegers: this way is less obvious but true, ATM I don't see any other situation either.

MBoegers: We can even be more clear, we want declarative recipes and those without optional options beeing validated.

rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
```diff
@@ -226,11 +227,18 @@ default void rewriteRun(Consumer<RecipeSpec> spec, SourceSpec<?>... sourceSpecs)
...
-        recipe.validateAll(ctx, validations);
-        assertThat(validations.stream().filter(Validated::isInvalid))
-                .as("Recipe validation must have no failures")
-                .isEmpty();
+
+        // Imperative recipes are loaded with no user-provided arguments, so all optional parameters are null.
+        // Skip recipe validation for these since custom validate() methods (e.g. requiring at least one of several
+        // optional parameters) would fail on an unconfigured instance.
+        if (recipe instanceof DeclarativeRecipe
+                || recipe.getDescriptor().getOptions().stream().allMatch(OptionDescriptor::isRequired)) {
```
timtebeek, line 235: Shouldn't this be
```suggestion
                || recipe.getDescriptor().getOptions().stream().noneMatch(OptionDescriptor::isRequired)) {
```

timtebeek: My thinking here is we can only fall to validate not provided but required options.

MBoegers: No, required options are fine, the validation fills them. Problematic is a recipe that contains at least one non required. 
With`nonMatch(isRequired)` we would allow recipes with only optional options.

MBoegers: isOptional or Predicat::not would make it a little easier to read

MBoegers: Oh and the main problem here are custom validation methods, like or checking for two optional parameter.

timtebeek: So the problem is the either/or nature of optional arguments that are both not filled here, do I understand that correctly?

MBoegers: Yes, that's the problem. You can find these recipes even in the core, f.ex. ChangeDependency.
Maybe the `Validated#or` is the real problem, other recipes that use a similar structure work with if, f.e. `o.o.docker.ChangeFrom`
But these would also fail for `RewriteTest#assertRecipesConfigure`

MBoegers:
The behavior change affects a few tests. Here are the usages of asserts for exceptions in our codebases.

| File | Pattern | Status |
  |------|---------|--------|
  | `rewrite-yaml/.../MergeYamlTest.java` | `assertThrows(AssertionError` | Fixed |
  | `rewrite-maven/.../ChangeDependencyGroupIdAndArtifactIdTest.java` | `assertThatExceptionOfType(AssertionError` | Fixed |
  | `rewrite-maven/.../ChangeManagedDependencyGroupIdAndArtifactIdTest.java` | `assertThatExceptionOfType(AssertionError` | Fixed |
  | `rewrite-test/.../RewriteTestTest.java` | `assertThrows(AssertionError` | Fine (tests name/scanning validation, not `validate()`) |
  | `rewrite-java-test/.../RepeatTest.java` | `assertThrows(AssertionError` | Fine (tests cursor passing during execution) |
  | `rewrite-core/.../AppendToTextFileTest.java` | `assertThatExceptionOfType(AssertionError` | Fine (tests missing expected generated files) |

MBoegers:
Thanks @timtebeek ❤️

PR #6843, opened 2026-02-27
Fix `IllegalArgumentException` in `Input.getRelativePath` for relative paths

timtebeek:
## What's Changed

When a `Parser$Input` has a relative path (e.g. synthetic `dependsOn` stubs created by `KotlinParser.determinePath()` via `Paths.get(pkg, prefix + className + ".kt")`), calling `getRelativePath(relativeTo)` with an absolute `relativeTo` path causes `Path.relativize()` to throw `IllegalArgumentException: 'other' is different type of Path` on JDK 21+.

This happens because JDK 21+ strictly enforces that both paths must be of the same type (both absolute or both relative) in `Path.relativize()`.

The fix returns relative paths as-is since they're already relative and don't need relativization.

- Fixes https://github.com/openrewrite/rewrite-gradle-plugin/issues/424

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -678,7 +678,7 @@ public class KotlinParser implements Parser {
 
     static class SourcePathFromSourceTextResolver {
         private static final Pattern packagePattern = Pattern.compile("^package\\s+(\\S+)");
-        private static final Pattern classPattern = Pattern.compile("(class|interface|enum class)\\s*(<[^>]*>)?\\s+(\\w+)");
+        private static final Pattern classPattern = Pattern.compile("(class|interface|data class|enum class)\\s*(<[^>]*>)?\\s+(\\w+)");
         private static final Pattern publicClassPattern = Pattern.compile("public\\s+" + classPattern.pattern());
 
         private static Optional<String> matchClassPattern(Pattern pattern, String source) {


shanman190:
It seems like the `KotlinParser` doesn't follow the convention of prefixing `dependsOn` snippets like every other parser that I've checked.

https://github.com/openrewrite/rewrite/blob/ae70bc3f59323778b63ddcd9eedbae0af552b11b/rewrite-java%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fjava%2FJavaParser.java#L267

https://github.com/openrewrite/rewrite/blob/ae70bc3f59323778b63ddcd9eedbae0af552b11b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L369-L390

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -352,7 +356,7 @@ public Builder addClasspathEntry(Path entry) {
 
         public Builder dependsOn(@Language("kotlin") String... inputsAsStrings) {
             this.dependsOn = Arrays.stream(inputsAsStrings)
-                    .map(input -> Input.fromString(determinePath("dependsOn-", input), input))
+                    .map(input -> Input.fromString(determinePath("", input), input))
```
shanman190: I think normally the absolute path just gets pretended here?

timtebeek: I went by what I'd seen in the JavaParser:
https://github.com/openrewrite/rewrite/blob/ae70bc3f59323778b63ddcd9eedbae0af552b11b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L267

shanman190: That is doing a `Paths.get("")` which isn't represented here. That's the same thing that I was meaning.

timtebeek: Hmm; with how that `Paths.get("")` is being used I couldn't mentally compute any difference, but I'll replicate it here.
https://github.com/openrewrite/rewrite/blob/ae70bc3f59323778b63ddcd9eedbae0af552b11b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L369-L389

timtebeek: We could even remove the now always empty prefix for no real effect that I can see; am I missing anything?
image

PR #6845, opened 2026-02-27
Handle unresolved Maven credential placeholders gracefully

timtebeek:
## What's changed

When a Maven `settings.xml` uses environment variable placeholders for server credentials (e.g. `${env.MAVEN_USER}`) and those variables are not set, the `PropertyPlaceholderHelper` preserves the literal `"${env.MAVEN_USER}"` string. This was then sent as actual Basic Auth credentials, causing 403 errors when downloading artifact JARs.

POM downloads were unaffected because `MavenPomDownloader` has `requestAsAuthenticatedOrAnonymous()` fallback logic (added for #1859), but `MavenArtifactDownloader` lacked this fallback.

In `MavenArtifactDownloader`, when credentials are applied and the server responds with a 4xx client error, retry the download anonymously — matching the existing `MavenPomDownloader` behavior.

- Fixes moderneinc/customer-requests#1928

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/AddToGitignore.java
+++ b/rewrite-core/src/main/java/org/openrewrite/AddToGitignore.java
@@ -126,6 +126,7 @@ public class AddToGitignore extends ScanningRecipe<AtomicBoolean> {
 
             private String mergeGitignoreEntries(String existing, String newEntries) {
                 String separator = existing.contains("\r\n") ? "\r\n" : "\n";
+                boolean endsWithNewline = existing.isEmpty() || existing.endsWith("\n");
 
                 Set<String> existingRules = new LinkedHashSet<>();
                 Set<String> existingWildcardPatterns = new LinkedHashSet<>();
@@ -181,7 +182,8 @@ public class AddToGitignore extends ScanningRecipe<AtomicBoolean> {
 
                 result.addAll(linesToAdd);
 
-                return String.join(separator, result);
+                String joined = String.join(separator, result);
+                return endsWithNewline ? joined + separator : joined;
             }
 
             private String normalizeRule(String rule) {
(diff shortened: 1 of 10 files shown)


timtebeek:
Hi Nick, thanks for the helpful link to the earlier work in https://github.com/openrewrite/rewrite/issues/1859 ; guess we had missed to do the same for artifacts at the time, which is a separate downloader. Now made consistent here to hopefully resolve your immediate issue.

rewrite-maven/src/main/java/org/openrewrite/maven/utilities/MavenArtifactDownloader.java
```diff
@@ -113,8 +113,19 @@ public MavenArtifactDownloader(MavenArtifactCache mavenArtifactCache,
                 try (HttpSender.Response response = Failsafe.with(retryPolicy).get(() -> httpSender.send(request.build()));
                      InputStream body = response.getBody()) {
                     if (!response.isSuccessful() || body == null) {
+                        int code = response.getCode();
+                        // If credentials caused a client-side error, retry anonymously
+                        MavenRepository repository = dependency.getRepository();
```
jkschneider: Every artifact downloaded is going to cause a client-side error. That may trigger anomaly detection on that server.

timtebeek: An alternative I'd considered was to not make an authenticated request when the username or password starts with `${`, but that rules out some very special passwords.

The pattern to repeat with an unauthenticated request is taken from `MavenPomDownloader`, where we had added that four years ago
- https://github.com/openrewrite/rewrite/issues/1859

I'll restore skipping authenticated requests when there's still a likely placeholder left over, unless you'd wanted to nudge in a different direction still.

nmck257: I suppose we could also make this a bit stateful -- toggle between trying authenticated or anonymous requests first, based on success of previous attempts.
if we land on a really good pattern, maybe also worth refactoring so the Pom and Artifact downloaders both share the same impl

timtebeek: Adjusted to logic to now only make a single unauthenticated request when the username/password still contain a placeholder. I've not yet replicated the same to the pom downloader, but could if desired.

sambsnyd (approved):

PR #6852, opened 2026-03-02
Add RemoveRedundantDependencyVersions.Skip marker

Jenson3210:
## Summary

- Add a `Skip` marker nested class in `RemoveRedundantDependencyVersions` that prevents version removal when present on a dependency tag
- Check for the marker in both `isDependencyTag()` and `isManagedDependencyTag()` branches of `visitTag`
- Revert the previous `retainExplicitVersion` flag approach in favor of this marker-based design

## Problem

When `UpgradeDependencyVersion` upgrades a dependency to match the parent-managed version, `RemoveRedundantDependencyVersions` subsequently removes the explicit `<version>` tag. In some cases, users want to retain the explicit version even when it matches the managed version, but there was no way for an earlier recipe to signal this intent.

## Solution

Instead of adding a configuration flag to `UpgradeDependencyVersion`, introduce a `RemoveRedundantDependencyVersions.Skip` marker that any recipe can place on a `<dependency>` tag. When `RemoveRedundantDependencyVersions` encounters this marker, it skips version removal for that tag. This keeps the mechanism generic and avoids adding recipe-specific boolean flags.

## Test plan

- [x] Existing `RemoveRedundantDependencyVersionsTest` tests pass (56 tests)
- [x] New `SkipMarker` nested test class with two tests:
  - `dependencyWithSkipMarkerKeepsVersion` - version preserved when marker present
  - `dependencyWithoutSkipMarkerHasVersionRemoved` - default behavior unchanged
- [x] `UpgradeDependencyVersionTest` tests pass

Fixes moderneinc/customer-requests#1942


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -100,6 +101,33 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
...
 
+    @Option(displayName = "Retain explicit version",
+            description = "When set to `true`, the explicit version in the dependency tag will not be removed even if it " +
+                          "matches the version managed by the parent POM. This is useful when you want to keep the version " +
+                          "tag in the child POM for clarity or policy reasons. Default `false`.",
+            required = false)
+    @Nullable
+    Boolean retainExplicitVersion;
+
+    public UpgradeDependencyVersion(String groupId, String artifactId, String newVersion,
```
timtebeek:
```suggestion
    @Deprecated
    public UpgradeDependencyVersion(String groupId, String artifactId, String newVersion,
```

rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
steve-aom-elliott, line 281: Is there not the possibility that their explicit versions will be wiped out anyway at some point if something like this is reached (possibly even for a different invocation of `UpgradeDependencyVersion` that is unrelated)?

PR #6857, opened 2026-03-03
Fix classcastexception

Jenson3210:
Daily tests showed a ClassCastException. 

For some reason the literal in there is a String type, but the actual value is a boolean. 
This can easily be asserted before casting in the already existing if, mitigating the risk here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@
     </picture>
   </a>
 </p>
-test
 
 <div align="center">
 


steve-aom-elliott:
The change itself seems safe at least, but I'd be really curious what LST element ended up parsing with type of String but value of boolean

timtebeek:
Indeed I worry that we might be hiding an issue. Might this be an apply: false that's incorrectly picked up?

Jenson3210:
appears to have been resolved when rerunning -> perhaps the if added few days ago was just not deployed 🤔

PR #6858, opened 2026-03-03
Convert ChangeDependency to ScanningRecipe to preserve GString/StringTemplate versions

Jenson3210:
## Summary

- Convert `ChangeDependency` from `Recipe` to `ScanningRecipe` to reliably handle version variables in GString and Kotlin StringTemplate dependency declarations
- Simplify `ChangeDependency` by leveraging the `GradleDependency` trait's `withDeclaredGroupId`, `withDeclaredArtifactId`, and `withDeclaredVersion` methods
- Fix bugs in the `GradleDependency` trait where `withDeclaredGroupId`/`withDeclaredArtifactId` on `K.StringTemplate` collapsed the template to a literal (destroying version interpolation), and `withDeclaredVersion` on `G.GString` converted to `J.Literal` (causing the Groovy printer to add unwanted parentheses)

## Why ScanningRecipe?

When a dependency uses a version variable (e.g., `implementation "group:artifact:${springBootVersion}"`), we want to:
1. **Preserve the interpolated string structure** — only update group/artifact in the literal prefix, keep `${variable}` intact
2. **Resolve the new version once** and update the variable at its definition site (`gradle.properties` or local `def`/`val`)

This requires two passes: a scan phase to discover which variables need updating and resolve the target version, then a visit phase to apply changes. A single-pass `Recipe` cannot reliably do this because it processes files independently without shared context.

**For dependencies without version variables** (plain string literals, map notation, multi-component literals), the scanning phase simply collects no variable information, and the visitor applies changes directly — exactly as the old non-scanning `Recipe` did. The `ScanningRecipe` conversion has no behavioral impact on these cases.

This approach was recommended in PR #6830 review feedback as the reliable alternative to the `ConcurrentHashMap`-based approach that depended on file processing order.

### `ChangeDependency.java`
- Converted from `Recipe` to `ScanningRecipe` with an `Accumulator` that maps version variable names to resolved versions (or `MavenDownloadingException`)
- Replaced ~400 lines of manual notation-specific code (GString element manipulation, StringTemplate extraction, platform unwrapping) with calls to the `GradleDependency` trait: `withDeclaredGroupId()`, `withDeclaredArtifactId()`, `withDeclaredVersion()`, `getVersionVariable()`, `getDeclaredVersion()`, `isPlatform()`
- Scanner detects shared version variables (used by both matching and non-matching dependencies) and falls back to collapsing interpolation to a literal in those cases

### `GradleDependency.java` (trait)
- Fixed `withDeclaredGroupId` and `withDeclaredArtifactId` for `K.StringTemplate`: now preserves template structure (updates only the literal prefix) instead of collapsing to a plain `J.Literal`
- Fixed `withDeclaredVersion` for `G.GString`: collapses to a single-element `GString` instead of `J.Literal`, which preserves Groovy command expression formatting (`implementation "..."` stays without parentheses)

### Tests
- 3 new `ChangeDependencyTest` tests for shared version variable detection and collapse
- 6 new `GradleDependencyTest` tests covering `withDeclaredGroupId`/`withDeclaredArtifactId`/`withDeclaredVersion` on `GString` and `K.StringTemplate`
- Updated existing interpolation tests to expect preserved variable references

## Test plan

- [x] All 26 `ChangeDependencyTest` tests pass (23 existing + 3 new shared variable tests)
- [x] All 6 new `GradleDependencyTest` trait tests pass
- [x] All existing `GradleDependencyTest` tests pass

Fixes moderneinc/customer-requests#1920

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -188,6 +188,9 @@ public class ChangeDependency extends ScanningRecipe<ChangeDependency.Accumulato
             public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
                 if (tree instanceof JavaSourceFile) {
                     gradleProject = tree.getMarkers().findFirst(GradleProject.class).orElse(null);
+                    if (gradleProject == null) {
+                        return (J) tree;
+                    }
                 }
                 return super.visit(tree, ctx);
             }
@@ -195,9 +198,6 @@ public class ChangeDependency extends ScanningRecipe<ChangeDependency.Accumulato
             @Override
             public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation m = (J.MethodInvocation) super.visitMethodInvocation(method, ctx);
-                if (gradleProject == null) {
-                    return m;
-                }
 
                 new GradleDependency.Matcher().get(getCursor()).ifPresent(dep -> {
                     String varName = dep.getVersionVariable();


rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -159,9 +162,111 @@ public Validated<Object> validate() {
...
+                        String resolvedGroupId = !StringUtils.isBlank(newGroupId) ? newGroupId : oldGroupId;
+                        String resolvedArtifactId = !StringUtils.isBlank(newArtifactId) ? newArtifactId : oldArtifactId;
+                        try {
+                            String resolvedVersion = new DependencyVersionSelector(metadataFailures, gp, null)
+                                    .select(new GroupArtifact(resolvedGroupId, resolvedArtifactId), m.getSimpleName(), newVersion, versionPattern, ctx);
+                            if (resolvedVersion != null) {
+                                acc.versionVariableUpdates.put(varName, resolvedVersion);
+                            }
+                        } catch (MavenDownloadingException e) {
+                            // Version resolution failed; will be handled in the visitor
```
Jenson3210: I think in other places we tend to store the downloadingException to later add it as a marker. see also UpgradeDependencyVersion for reference

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -159,9 +165,131 @@ public Validated<Object> validate() {
...
+                    String varName = extractVersionVariableName(strings);
+                    if (varName != null && !StringUtils.isBlank(newVersion)) {
+                        resolveAndRecordVersion(varName, m, ctx);
+                    }
+                }
+            }
+
+            private void resolveAndRecordVersion(String varName, J.MethodInvocation m, ExecutionContext ctx) {
+                String resolvedGroupId = !StringUtils.isBlank(newGroupId) ? newGroupId : oldGroupId;
+                String resolvedArtifactId = !StringUtils.isBlank(newArtifactId) ? newArtifactId : oldArtifactId;
```
Jenson3210: will this support globs? Looks like this would do a glob lookup. We should use the dependency information of the Trait instead perhaps?

rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
```diff
@@ -1235,7 +1245,7 @@ public GradleDependency withDeclaredVersion(@Nullable String newVersion) {
                 }
             }
         } else if (firstArg instanceof G.GString) {
-            // For GString, we convert to a simple string literal with the new version
+            // For GString, collapse to a single-element GString to preserve Groovy command expression formatting
```
Jenson3210, line 1248: I do not think we need to do this and can keep the original intent here

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -238,386 +314,89 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
                         try {
                             resolvedVersion = new DependencyVersionSelector(metadataFailures, gradleProject, null)
-                                    .select(new GroupArtifact(updatedGroupId, updatedArtifactId), m.getSimpleName(), newVersion, versionPattern, ctx);
+                                    .select(new GroupArtifact(
+                                                    !StringUtils.isBlank(newGroupId) ? newGroupId : dep.getGroupId(),
+                                                    !StringUtils.isBlank(newArtifactId) ? newArtifactId : dep.getArtifactId()),
+                                            dep.getConfigurationName(), newVersion, versionPattern, ctx);
                         } catch (MavenDownloadingException e) {
                             return e.warn(m);
                         }
```
Jenson3210: why do we again need to resolve? the scan would have done this, right?

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -159,9 +162,85 @@ public Validated<Object> validate() {
...
+                return (sourceFile instanceof G.CompilationUnit && sourceFile.getSourcePath().toString().endsWith(".gradle")) ||
+                        (sourceFile instanceof K.CompilationUnit && sourceFile.getSourcePath().toString().endsWith(".gradle.kts"));
+            }
+
+            @Override
+            public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (tree instanceof JavaSourceFile) {
+                    gradleProject = tree.getMarkers().findFirst(GradleProject.class).orElse(null);
+                }
+                return super.visit(tree, ctx);
```
shanman190, line 195: We should probably just short circuit the tree traversal here to save us the time of visiting the entire source file when the gradleProject marker is absent.

We'd also remove the early return from the `visitMethodInvocation` because it wouldn't be needed anymore.

PR #6860, opened 2026-03-03
Fix Maven parser on ${project.version} in parent reference

MBoegers:
## Summary

- Try reactor lookup with raw (unresolved) GAV before property resolution in `resolveParentPom()`
- Default blank `<relativePath>` to `..` per Maven POM §4.0.0 in path-based parent lookup
- Relax version check in path-based match since version may contain unresolved placeholders

## Problem

Multi-module Maven projects that define `<project.version>` as an explicit property and use `${project.version}` in child parent references fail to parse. `ResolvedPom.resolveParentPom()` resolves `${project.version}` through `getValues()` before trying to find the parent in the reactor. The child has no resolved parent yet, so property resolution hits the built-in `project.version` handler which detects a circular reference and returns `"error.circular.project.version"`.

Customer setup:
- Root POM: `<version>${project.version}</version>` with `<project.version>X.Y</project.version>` in properties
- Child POMs: `<parent><version>${project.version}</version></parent>`
- CI overrides with `-Dproject.version=XYZ`

## Solution

**`ResolvedPom.resolveParentPom()`**: When the parent version contains `${`, attempt `downloader.download()` with the raw GAV first. The downloader's iteration strategy (lines 502-516) resolves placeholders via merged project properties and finds the parent in the reactor. Falls back to the original `getValues()` path if the reactor lookup fails.

**`MavenPomDownloader.download()`**: Default blank `relativePath` to `..` (per Maven POM §4.0.0, see `DefaultModelBuilder#readParentLocally`), and drop the version check in the relative-path match since the version may contain unresolved placeholders.

## Test plan

- [x] Two new multi-module test cases: inter-module dependency with `${project.version}`, builder property override
- [x] Existing circular reference tests still pass (`circularMavenProperty`, `dependencyWithCircularProjectVersionReference`)
- [x] Full `rewrite-maven` test suite passes

- Fixes moderneinc/customer-requests#1931

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -516,20 +516,26 @@ public class MavenPomDownloader {
             }
         }
 
-        if (containingPom != null && containingPom.getRequested().getSourcePath() != null &&
-            !StringUtils.isBlank(relativePath) && !relativePath.contains(":")) {
-            Path folderContainingPom = containingPom.getRequested().getSourcePath().getParent();
-            if (folderContainingPom != null) {
-                Pom maybeLocalPom = projectPoms.get(folderContainingPom.resolve(Paths.get(relativePath).resolve("pom.xml"))
-                        .normalize());
-                // Even poms published to remote repositories still contain relative paths to their parent poms
-                // So double check that the GAV coordinates match so that we don't get a relative path from a remote
-                // pom like ".." or "../.." which coincidentally _happens_ to have led to an unrelated pom on the local filesystem
-                if (maybeLocalPom != null &&
-                    gav.getGroupId().equals(maybeLocalPom.getGroupId()) &&
-                    gav.getArtifactId().equals(maybeLocalPom.getArtifactId()) &&
-                    gav.getVersion().equals(maybeLocalPom.getVersion())) {
-                    return maybeLocalPom;
+        if (containingPom != null && containingPom.getRequested().getSourcePath() != null) {
+            // Maven POM §4.0.0 specifies that <relativePath> defaults to ".." when omitted.
+            // See DefaultModelBuilder#readParentLocally in maven-model-builder.
+            String effectiveRelativePath = StringUtils.isBlank(relativePath) ? ".." : relativePath;
+            if (!effectiveRelativePath.contains(":")) {
+                Path folderContainingPom = containingPom.getRequested().getSourcePath().getParent();
+                if (folderContainingPom != null) {
+                    Pom maybeLocalPom = projectPoms.get(folderContainingPom.resolve(Paths.get(effectiveRelativePath).resolve("pom.xml"))
+                            .normalize());
+                    // Even poms published to remote repositories still contain relative paths to their parent poms.
+                    // Like Maven 3's DefaultModelBuilder#readParentLocally, check groupId, artifactId,
+                    // and version to guard against a relative path like ".." or "../.." coincidentally
+                    // resolving to an unrelated local pom. The version check is relaxed when it contains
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -516,20 +516,28 @@ public Pom download(GroupArtifactVersion gav,
...
+                    // Check groupId and artifactId to guard against a relative path like ".." or "../.."
+                    // coincidentally resolving to an unrelated local pom. Version is intentionally not
+                    // checked here: the requested version may still contain unresolved placeholders (e.g.
+                    // ${project.version}) that only the parent itself can resolve. Maven's own
+                    // DefaultModelBuilder#readParentLocally checks version but falls back to
+                    // readParentExternally on mismatch — we don't have that two-phase fallback, so we
+                    // accept the match here when groupId and artifactId agree.
+                    if (maybeLocalPom != null &&
+                            gav.getGroupId().equals(maybeLocalPom.getGroupId()) &&
+                            gav.getArtifactId().equals(maybeLocalPom.getArtifactId())) {
```
timtebeek: Is there any risk associated with not checking the version here anymore? It looks like we would match any GA, ignoring V even if that might be hardcoded and different.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -572,7 +572,28 @@ private void resolveParentPluginsRecursively(List<Pom> pomAncestry) throws Maven
         }
 
         private Pom resolveParentPom(Pom pom) throws MavenDownloadingException {
-            @SuppressWarnings("DataFlowIssue") GroupArtifactVersion gav = getValues(pom.getParent().getGav());
+            @SuppressWarnings("DataFlowIssue") GroupArtifactVersion rawGav = pom.getParent().getGav();
+            if (rawGav.getVersion() == null) {
+                throw new MavenParsingException("Parent version must always specify a version " + rawGav);
```
timtebeek, line 577: I thought in Maven 4 they support version less parent references:
https://maven.apache.org/whatsnewinmaven4.html#Parent_inference

MBoegers: Yes they do, Maven 3 rejects parent definitions without a version. So does our current implementation on main.
I can work on a fix for this situation, but Maven 4 is another topic I think.

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
MBoegers: True, if a parent is defined with a hardcoded GA and V and a relative path, but the parent under the relative path has another hardcoded V*. Than GAV != GAV* which would force Maven to resolve G:A:V from remote, which works if the parent with V* were published.

I restored and relaxed for properties:
- Hardcoded version mismatch => rejected (matches Maven 3’s readParentLocally)
- Version with `${` → accepted (can’t be resolved yet, reactor is the right source)

Using contains to allow situation like `1.0.${revision}`.

rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
sambsnyd: Agree with Tim it would be nice to go ahead and support parent inference, doesn't seem like it would be too difficult since we've got the list of project poms in the execution context.

MBoegers: see https://github.com/openrewrite/rewrite/issues/6869 I am on it.

PR #6875, opened 2026-03-04
Honor empty <relativePath/> in MavenPomDownloader

Jenson3210:
## Summary

- Distinguish between omitted `<relativePath>` (defaults to `../pom.xml`) and explicit empty `<relativePath/>` (skip local parent lookup) in `MavenPomDownloader`
- Fix affects both `getParentWithinProject()` and the relative-path resolution block in `download()`

## Problem

When a POM declares `<relativePath/>` (empty self-closing tag), Maven skips local parent lookup entirely and goes straight to remote repositories ([DefaultModelBuilder.java#L1048](https://github.com/apache/maven/blob/c94b33c84d4e4851c0fe780d73f0650fea445974/compat/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java#L1048)). OpenRewrite's `MavenPomDownloader` was treating empty `relativePath` the same as omitted because both `null` and `""` pass through `StringUtils.isBlank()`, causing it to default to `../pom.xml` in both cases.

## Solution

Check for empty string explicitly before the `isBlank` fallback:
- `null` (element absent) → default to `../pom.xml`
- `""` (explicit empty `<relativePath/>`) → skip local parent lookup
- non-blank string → use that path

## Test plan

- [x] Existing tests pass (`MavenPomDownloaderTest`, `MavenParserTest`, `ChangeParentPomTest`, `ResolvedPomTest`)
- [x] New test `emptyRelativePathSkipsLocalParentLookup` verifies the fix

- Fixes moderneinc/customer-requests#1950

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/internal/MavenPomDownloaderTest.java
@@ -1117,57 +1117,6 @@ class MavenPomDownloaderTest implements RewriteTest {
               () -> downloader.download(requestedGav, "", resolvedPom, singletonList(nonexistentRepo)));
         }
 
-        @Issue("https://github.com/moderneinc/customer-requests/issues/1950")
-        @Test
-        void emptyRelativePathSkipsPathBasedLookupWithPlaceholderVersion() {
-            // A local parent POM exists at ../pom.xml, but the child specifies <relativePath/>
-            // (empty string) and references the parent with a placeholder version ${my.version}.
-            // Without the fix, the path-based lookup (stage 3) would default relativePath to "..",
-            // find the parent, and accept it because "${" in the version relaxes the version check.
-            // With the fix, stage 3 is skipped entirely due to empty relativePath.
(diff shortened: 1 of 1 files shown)


Jenson3210:
@MBoegers I could use your recent investigations to get bit more insights on this one. 
Is this a valid change?

MBoegers:
**Added another test:**

The test exercises these steps through `download()`:
1. **Stage 1** exact GAV lookup in `projectPomsByGav`: no match because `${my.version} != 1.0.0`
2. **Stage 2** iterate all project POMs, find `test:parent`, try to resolve the placeholder: fails because the parent doesn’t define `my.version`, so `${my.version}` stays unresolved and doesn’t match `1.0.0`
3. **Stage 3** path-based lookup using `relativePath`:
  - **Without fix:** empty `relativePath` defaults to `..`, resolves `child/../pom.xml`, finds the parent, and the `${` in the version relaxes the version check → returns local POM (wrong)
  - **With fix:** empty relativePath means “skip local lookup” → block is skipped entirely
Remote download — falls through to fetch from the nonexistent repo → throws exception

The `assertThrows(Exception.class)` confirms stage 3 was skipped — if it wasn’t, the local POM would be returned and no exception would be thrown.

PR #6879, opened 2026-03-05
Fix getRelativePath throwing on relative paths (JDK 21+)

MBoegers:
## Summary

- `Parser.Input.getRelativePath()` now returns relative paths as-is instead of calling `Path.relativize()` with a mismatched absolute base
- Prevents `IllegalArgumentException` thrown by JDK 21+ when `relativize()` is called with an absolute path against a relative path

## Problem

When a Kotlin stub fails to parse, `KotlinParser` calls `ParseError.build()`, which calls `input.getRelativePath(relativeTo)`. The `relativeTo` from the Gradle plugin is absolute (e.g. `/home/user/project`). The `path` on the `Input` from `KotlinParser.determinePath()` is relative (e.g. `com/example/MyClass.kt`).

On JDK 21+, `Path.relativize()` throws `IllegalArgumentException: 'other' is different type of Path`. This causes parse failures to propagate as uncaught exceptions rather than clean `ParseError` nodes, breaking LST serialization for all Kotlin-containing projects.

- Note: [#6843](https://github.com/openrewrite/rewrite/pull/6843) addressed this issue by removing the `"dependsOn-"` prefix but did not fix the actual crash site in `getRelativePath()`. The stubs are still relative paths, and the `IllegalArgumentException` still occurs.

## Fix

Guard in `getRelativePath()`: if `path` is already relative, return it as-is. `relativize()` is only called when both paths are absolute.

- Closes moderneinc/customer-requests#1954

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Parser.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Parser.java
@@ -208,7 +208,10 @@ public interface Parser {
         }
 
         public Path getRelativePath(@Nullable Path relativeTo) {
-            return relativeTo == null ? path : relativeTo.relativize(path);
+            if (relativeTo == null || !path.isAbsolute()) {
+                return path;
+            }
+            return relativeTo.relativize(path);
         }
 
         public EncodingDetectingInputStream getSource(ExecutionContext ctx) {


timtebeek:
- https://github.com/openrewrite/rewrite/pull/6882

PR #6882, opened 2026-03-05
Fix IllegalArgumentException in KotlinParser when dependsOn paths are relative

jkschneider:
## Summary
- `dependsOn` inputs created via `determinePath` have relative paths, but source inputs may have absolute paths
- When `relativeTo` is absolute, `Path.relativize` throws `IllegalArgumentException` due to the absolute/relative mismatch
- Resolves dependsOn paths against `relativeTo` before relativizing, preventing the crash

## Test plan
- [ ] Verify existing Kotlin parser tests pass
- [ ] Test with Kotlin sources that have absolute paths and dependsOn with relative paths

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Parser.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Parser.java
@@ -208,7 +208,10 @@ public interface Parser {
         }
 
         public Path getRelativePath(@Nullable Path relativeTo) {
-            return relativeTo == null ? path : relativeTo.relativize(path);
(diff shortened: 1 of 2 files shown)


asbachb:
Could you trigger a rebuild of gradle plugin with this fix included?

timtebeek:
Hi @asbachb ; the fix should already be available through [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions); we'll resume our regular release schedule next week. Hope that helps!

PR #6892, opened 2026-03-09
Capture Spring dependency-management plugin managed versions in GradleProject marker

Jenson3210:
## Summary

- Introduces `SpringDependencyManagement` as a dedicated serializable model capturing the full state of the Spring `io.spring.dependency-management` plugin
- Separates implicit (BOM-imported) versions from explicit (`dependencyManagement { dependencies {} }`) versions
- Captures `overriddenByDependencies` flag and imported BOM coordinates
- Supports both global and configuration-specific dependency management
- Moved reflection logic from `GradleProjectBuilder` into `SpringDependencyManagement` itself

## Problem

When the Spring dependency-management plugin is applied, it overrides dependency versions via Gradle's resolution strategy. Recipes that try to bump a dependency version by editing the `dependencies {}` block fail silently because the plugin takes precedence. The GradleProject marker had no way to indicate which dependencies are managed by this plugin, so recipes couldn't make informed decisions about *how* to change a version.

## Solution

A new `SpringDependencyManagement` class models the plugin's internal state using deep reflection into the `DependencyManagementExtension`:

- **`DependencyManagement`** (inner class): Holds `implicitVersions` (from BOMs), `explicitVersions` (from `dependencyManagement { dependencies {} }`), and `importedBoms` (list of `GroupArtifactVersion`)
- **`overriddenByDependencies`**: When true (default), explicit versions in `dependencies {}` take precedence over BOM versions. When false, BOM versions always win.
- **`globalDependencyManagement`**: The global managed versions (from `dependencyManagement {}` without a configuration scope)
- **`configurationDependencyManagement`**: Per-configuration managed versions (from `dependencyManagement { implementation { ... } }`)
- **`getManagedVersion(configuration, group, name)`**: Resolves the effective managed version, respecting `overriddenByDependencies` ordering

The field lives on `GradleProject` as `@Nullable SpringDependencyManagement springDependencyManagement`, null when the plugin is not applied (backward-compatible with old serialized markers).

### Files changed

- **`SpringDependencyManagement.java`** (new) — Serializable model with inner `DependencyManagement` class. Contains the reflective extraction logic (`springDependencyManagement(Project)`), accessing private fields (`dependencyManagementSettings`, `dependencyManagementContainer`, `versions`, `explicitVersions`, `importedBoms`).
- **`GradleProject.java`** — New `@Nullable SpringDependencyManagement springDependencyManagement` field with `@With`. Updated constructor calls.
- **`GradleDependencyConfiguration.java`** — Removed `springDependencyManagementManagedVersions` field (moved to project-level `SpringDependencyManagement`). Reverted deprecated constructors and `clone()`.
- **`GradleProjectBuilder.java`** — Calls `SpringDependencyManagement.springDependencyManagement(project)` and passes result to `GradleProject` constructor. Removed old reflection method.
- **`GradleSettingsBuilder.java`** — Updated `dependencyConfigurations()` call signature.
- **`Assertions.java`** — Updated `GradleProject` constructor call.
- **`GradleProjectTest.java`** — Three test classes: `springDependencyManagement` (with `overriddenByDependencies=false`), `springDependencyManagementOverriddenByDependencies` (default `true`), and `noSpringPlugin` (null marker).

## Test plan

- [x] BOM-managed versions captured (Spring Boot BOM, explicit JUnit BOM import)
- [x] Explicit `dependencyManagement { dependencies {} }` entries captured
- [x] Unmanaged dependencies return null (both absent and present in project)
- [x] `overriddenByDependencies=false`: managed version returned even when dependency declares a newer version
- [x] `overriddenByDependencies=true` (default): BOM version still reported for managed deps
- [x] Spring Boot starters managed with and without explicit version
- [x] `overriddenByDependencies` flag correctly captured for both true and false
- [x] Serialization round-trip preserves the model
- [x] Null when Spring plugin is not applied
- [x] All existing `:rewrite-gradle-tooling-model:model:test` tests pass

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/marker/SpringDependencyManagement.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/marker/SpringDependencyManagement.java
@@ -44,21 +44,16 @@ public final class SpringDependencyManagement implements Serializable {
      * project's dependencies. The default is {@code true}.
      */
     @Builder.Default
-    private final
     boolean overriddenByDependencies = true;
 
-    @Nullable
-    @Builder.Default
-    private final
-    DependencyManagement globalDependencyManagement = null;
+    DependencyManagement globalDependencyManagement;
 
     @Builder.Default
     @With
-    private final
     Map<String, DependencyManagement> configurationDependencyManagement = emptyMap();
 
     @JsonCreator
-    public SpringDependencyManagement(boolean overriddenByDependencies, @Nullable DependencyManagement globalDependencyManagement, Map<String, DependencyManagement> configurationDependencyManagement) {
+    public SpringDependencyManagement(boolean overriddenByDependencies, DependencyManagement globalDependencyManagement, Map<String, DependencyManagement> configurationDependencyManagement) {
         this.overriddenByDependencies = overriddenByDependencies;
         this.globalDependencyManagement = globalDependencyManagement;
         this.configurationDependencyManagement = configurationDependencyManagement;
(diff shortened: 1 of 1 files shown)


Jenson3210:
@shanman190 Tim mentioned your early eyes on something like this would be nice. 

Reasoning of adding this: 
- when FFVD bumps a vulnerability (eg. jackson-core) it tries to identify a direct dependency it can bump:
```
    some-dependency
        spring-boot-starter-x
            jackson-core
```

This would bump `some-dependency` if there is a newer version that brings a higher version of `spring-boot-starter`. 
However, spring boot starter version is managed by the dependency management plugin -> jackson is not effectively bumped. The FFVD however sees that the file was changed so does not attempt a constraint addition. 

If we have this I could update the logic in FFVD to search for a direct one to bump but do not consider `some-dependency` as a direct candidate as it will not resolve the jackson-core vulnerability. 

WDYT?

shanman190:
So I'd have to test again, but I think constraints have lower precedence than resolution strategy rules, meaning that since the Spring Dependency Management plugin uses resolution strategy rules we would have to add specific property overrides on each of the target transitive dependencies in order for them to take effect.

Jenson3210:
@shanman190 you are right about that but the upgradeTransitiveDependency will handle that. Running it for a transitive one when the spring plugin is active results in a constraint being created as resulutionatrategy.

Jenson3210:
Was able to work with the existing Constraints here.

PR #6895, opened 2026-03-09
Fix NPE in Markup.markup() when tree node has null markers

bmuschko:
## Summary

Fixes a `NullPointerException` that occurs during recipe error recovery when a YAML tree node has `null` markers.

### What causes the NullPointerException

The `Tree` interface contract implies that `getMarkers()` should never return `null`, but YAML tree nodes can end up with `null` markers through **RPC deserialization**. Specifically, `YamlReceiver.preVisit()` calls:

```java
return y.withMarkers(q.receive(y.getMarkers()));
```

When `q.receive()` returns `null` (e.g., due to missing or malformed data in the RPC message), the resulting tree node has a `null` `markers` field. The YAML parser itself always initializes nodes with `Markers.EMPTY`, so this issue is specific to the RPC pathway.

When a recipe subsequently runs on such a tree and encounters an error, the framework's error recovery mechanism (`FindRecipeRunException`) attempts to annotate the failing node with an error marker by calling `Markup.error(tree, exception)`. This delegates to `Markup.markup()`:

```java
static <T extends Tree> T markup(T t, Markup markup) {
    return t.withMarkers(t.getMarkers().compute(markup, ...));
    //                    ^^^^^^^^^^^^^^ NPE when null
}
```

Since `getMarkers()` is `null`, this throws a `NullPointerException` **inside the error handler itself**, which:
1. Masks the original exception that triggered the error recovery
2. Makes the root cause of the actual recipe failure impossible to diagnose from the stack trace

### What we changed

**Defensive fix in `Markup.markup()` (`rewrite-core`):** Added a null check that falls back to `Markers.EMPTY` when `getMarkers()` returns `null`. This ensures error recovery never crashes regardless of which tree node type or deserialization pathway produced the null markers.

**Preventive fix in `YamlReceiver.preVisit()` (`rewrite-yaml`):** Added a null guard on the value returned by `q.receive(y.getMarkers())`, falling back to `Markers.EMPTY`. This prevents YAML tree nodes from being constructed with null markers during RPC deserialization in the first place.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
@@ -84,8 +84,7 @@ public interface Markup extends Marker {
     }
 
     static <T extends Tree> T markup(T t, Markup markup) {
-        Markers markers = t.getMarkers() != null ? t.getMarkers() : Markers.EMPTY;
-        return t.withMarkers(markers.compute(markup, (s1, s2) -> s1 == null ? s2 : s1));
+        return t.withMarkers(t.getMarkers().compute(markup, (s1, s2) -> s1 == null ? s2 : s1));
     }
 
     @Value
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/MarkupNullMarkersTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/MarkupNullMarkersTest.java
@@ -16,9 +16,10 @@
 package org.openrewrite.yaml;
 
 import org.junit.jupiter.api.Test;
-import org.openrewrite.*;
+import org.openrewrite.Cursor;
 import org.openrewrite.internal.FindRecipeRunException;
 import org.openrewrite.internal.RecipeRunException;
+import org.openrewrite.marker.Markers;
 import org.openrewrite.marker.Markup;
 import org.openrewrite.yaml.tree.Yaml;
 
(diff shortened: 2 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/marker/Markup.java
```diff
@@ -84,7 +84,8 @@ static <T extends Tree> T debug(T t, String message, @Nullable String detail) {
     }
 
     static <T extends Tree> T markup(T t, Markup markup) {
-        return t.withMarkers(t.getMarkers().compute(markup, (s1, s2) -> s1 == null ? s2 : s1));
+        Markers markers = t.getMarkers() != null ? t.getMarkers() : Markers.EMPTY;
+        return t.withMarkers(markers.compute(markup, (s1, s2) -> s1 == null ? s2 : s1));
```
timtebeek: I'm not too sure about the defensive fix here; it leads to warnings in the IDE, and signifies that the non-nullability can not be trusted: that could lead to further such needless checks then. I think I'd prefer that we only handle this in the Rpc deserialization, as you've already done below.

Image

timtebeek:
```suggestion
        return t.withMarkers(t.getMarkers().compute(markup, (s1, s2) -> s1 == null ? s2 : s1));
```

bmuschko: done

PR #6896, opened 2026-03-09
UpgradeTransitiveDependencyVersion should not add managed deps to child modules

steve-aom-elliott:
## Summary

In a multi-module Maven project, `UpgradeTransitiveDependencyVersion` was adding `<dependencyManagement>` entries to every module independently. When a parent pom has a dependency that transitively brings in a library, child modules inherit both the dependency and its transitives. The recipe was redundantly adding the same managed dependency entries to child poms that already inherit them from the parent.

The fix checks whether any ancestor pom in the source set also has the same transitive dependency. If so, the ancestor will receive the managed dependency entry and children will inherit it, so the child is skipped.

## What changed

- `UpgradeTransitiveDependencyVersion.getVisitor()` now walks the parent chain (`getResolutionResult().getParent()`) and removes matching transitive dependencies that an ancestor will also manage
- Added test `doesNotAddManagedDependencyToChildModuleWhenInheritedFromParent` demonstrating the fix with a parent+child multi-module project

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeTransitiveDependencyVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeTransitiveDependencyVersion.java
@@ -190,7 +190,6 @@ public class UpgradeTransitiveDependencyVersion extends ScanningRecipe<AddManage
                 MavenResolutionResult current = getResolutionResult();
                 while (current.parentPomIsProjectPom()) {
                     MavenResolutionResult parentResult = current.getParent();
-                    assert parentResult != null;
                     List<ResolvedDependency> parentTransitiveDeps = parentResult.findDependencies(groupId, artifactId, null);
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeTransitiveDependencyVersion.java
```diff
@@ -182,6 +184,24 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
                 if (matchingDependencies.isEmpty()) {
                     return document;
                 }
+
+                // Skip transitive dependencies that a parent in the source set also has,
+                // since the parent will get the managed dependency and children will inherit it
+                MavenResolutionResult parentResult = getResolutionResult().getParent();
```
Jenson3210: not 100% familiar with the MavenResolutionResult but I though parent could also live remotely? Will this loop only contain local parents?

PR #6909, opened 2026-03-10
Fix `AddProperty` depending on `<relativePath>` being present

Goshenkata:
## What's changed?
I think the previous way to detect parent project has a bug because <relativePath> can be ommited (i which case its implicitly "../pom.xml")

## Any additional context
I discovered this issue while using UpgradeDependencyVersion on a project that did have relativePath and this condition caused the recipe to not override properties in the pom.xml. I will make a seperate PR for UpgradeDependencyVersion's issue :)

### Checklist

### DIFF
--- a/rewrite-csharp/.gitignore
+++ b/rewrite-csharp/.gitignore
@@ -1,0 +1,1 @@
+src/main/resources/META-INF/rewrite-csharp-version.txt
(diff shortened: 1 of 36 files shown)


Goshenkata:
Breaks callers...

timtebeek:
Hi @Goshenkata ! Thanks for exploring an immediate fix; let us know if you need any help!

PR #6935, opened 2026-03-11
AddDependency only scans test sources when scope is test

Jenson3210:
## Context

- Fixes moderneinc/customer-requests#1972

During Spring Boot migration, test dependencies (e.g., `spring-boot-starter-restclient-test`) are added to projects that have no tests or no tests using the concepts those dependencies provide. This happens because `AddDependency` scans all source files for `onlyIfUsing` matches regardless of the target scope.

## Summary

- When `scope` is explicitly `"test"` (Maven) or `configuration` targets a test source set like `"testImplementation"` (Gradle), and `onlyIfUsing` is set, the scanner now only checks test source files
- If the type is only used in production code, the test dependency is no longer added
- Existing behavior for `scope=compile`, `scope=runtime`, `scope=provided`, and inferred scopes is unchanged

## Test plan

- [x] Maven: `scope=test` + type only in main sources → dependency NOT added
- [x] Maven: `scope=test` + type in test sources → dependency IS added with test scope
- [x] Gradle: `testImplementation` + type only in main sources → dependency NOT added
- [x] Gradle: `testImplementation` + type in test sources → dependency IS added
- [x] Existing `onlyIfUsingTestScope` and `onlyIfUsingCompileScope` tests still pass
- [x] `addScopedDependency` for `provided`/`runtime` scopes still works (unchanged behavior)

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -169,7 +169,8 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
                     return tree;
                 }
                 SourceFile sourceFile = (SourceFile) tree;
-                if (configuration != null && onlyIfUsing != null &&
+                if (sourceFile instanceof JavaSourceFile &&
+                        configuration != null && onlyIfUsing != null &&
                         configuration.startsWith("test") && configuration.length() > 4 &&
                         Character.isUpperCase(configuration.charAt(4)) &&
                         sourceFile == hasTestSourceSet.visit(sourceFile, ctx)) {


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -167,6 +169,13 @@ private boolean usesType(SourceFile sourceFile, ExecutionContext ctx) {
                     return tree;
                 }
                 SourceFile sourceFile = (SourceFile) tree;
+                if (configuration != null && onlyIfUsing != null &&
+                        configuration.startsWith("test") && configuration.length() > 4 &&
+                        Character.isUpperCase(configuration.charAt(4))) {
+                    if (sourceFile == isLikelyTest.visit(sourceFile, ctx)) {
+                        return tree;
+                    }
+                }
```
timtebeek: Would it not make sense to use a single `if` here? We're already using `&&` 4x anyway.

```suggestion
                if (configuration != null && onlyIfUsing != null &&
                        configuration.startsWith("test") && configuration.length() > 4 &&
                        Character.isUpperCase(configuration.charAt(4)) &&
                        sourceFile == isLikelyTest.visit(sourceFile, ctx)) {
                    return tree;
                }
```

rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java
```diff
@@ -171,10 +172,17 @@ public Scanned getInitialValue(ExecutionContext ctx) {
...
+
             @Override
             public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                 SourceFile sourceFile = (SourceFile) requireNonNull(tree);
                 if (tree instanceof JavaSourceFile) {
+                    if ("test".equals(scope) && onlyIfUsing != null) {
+                        if (sourceFile == isLikelyTest.visit(sourceFile, ctx)) {
+                            return sourceFile;
+                        }
+                    }
```
timtebeek:
```suggestion
                    if ("test".equals(scope) && onlyIfUsing != null && sourceFile == isLikelyTest.visit(sourceFile, ctx)) {
                        return sourceFile;
                    }
```

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -150,6 +151,7 @@ public TreeVisitor<?, ExecutionContext> getScanner(Scanned acc) {
 
             @Nullable
             UsesType<ExecutionContext> usesType = null;
+            final TreeVisitor<?, ExecutionContext> isLikelyTest = new IsLikelyTest().getVisitor();
```
timtebeek: I'm doubting slightly between `IsLikelyTest` vs `HasSourceSet` here:
https://github.com/openrewrite/rewrite/blob/419f3b9e5ccb89c96473e11671780e28ba8c86d9/rewrite-java/src/main/java/org/openrewrite/java/search/IsLikelyTest.java#L32-L39

https://github.com/openrewrite/rewrite/blob/419f3b9e5ccb89c96473e11671780e28ba8c86d9/rewrite-java/src/main/java/org/openrewrite/java/search/HasSourceSet.java#L32-L42

timtebeek: I feel the latter might be more closely tied to the scope we use for dependencies, as compared to test classes which don't cover all possible cases, and might even mismatch on something like our own [rewrite-java-tck/src/main/java/org/openrewrite/java/tree](https://github.com/openrewrite/rewrite/tree/main/rewrite-java-tck/src/main/java/org/openrewrite/java/tree). Any thoughts here?

Jenson3210: I did not know the latter one and agree that's a better choice here!

PR #6936, opened 2026-03-11
Replace DeclarativeRecipe ThreadLocal accumulator with cursor-based lookup

pstreef:
## Problem

- PR #6810 changed `DeclarativeRecipe.accumulator` from a plain field to `ThreadLocal<Accumulator>` to fix concurrent recipe runs overwriting each other's accumulator. However, when a recipe run yields and resumes on a different thread, the ThreadLocal value is lost.

The `orVisitors()` method reads the accumulator from the ThreadLocal to resolve scanning precondition visitors. After a thread switch, `accumulator.get()` returns `null`, causing scanning preconditions to receive null accumulators. This affects any `DeclarativeRecipe` that uses a `ScanningRecipe` as a precondition.

## Solution

The accumulator is already stored on the rootCursor by `ScanningRecipe.getAccumulator()`, which is per-run and survives thread switches. The ThreadLocal was redundant secondary storage.

- Remove `ThreadLocal<Accumulator>` entirely
- Defer precondition resolution to visit time when the rootCursor is available (via `setCursor()` set by `RecipeRunCycle`)
- `orVisitors()` reads the accumulator from the cursor via `getAccumulator(rootCursor, ctx)` instead of ThreadLocal
- Remove `onComplete()` ThreadLocal cleanup and simplify `clone()`

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/DeclarativeRecipe.java
@@ -230,26 +230,30 @@ public class DeclarativeRecipe extends ScanningRecipe<DeclarativeRecipe.Accumula
         @Override
         public TreeVisitor<?, ExecutionContext> getVisitor() {
             return new TreeVisitor<Tree, ExecutionContext>() {
+                // Safe to cache: each getVisitor() call creates a new instance, and each
+                // instance is only used by a single thread (RecipeRunCycle processes files
+                // sequentially). Resolved lazily because getCursor() requires the framework
+                // to set the rootCursor first (via setCursor) before isAcceptable/visit.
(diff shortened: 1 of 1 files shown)


pstreef:
Closing this — after deeper investigation of the worker's yield mechanics, the ThreadLocal approach from #6810 is correct.

The rootCursor is preserved across yield boundaries within the same cycle, and scan/edit happen within the same RecipeRunCycle batch on the same thread. The ThreadLocal is set during scan and read during edit of the same batch. On resume after yield, a new batch does its own scan phase which sets the ThreadLocal fresh.

The actual fix for the yield-related accumulator loss was #6913 (delegating getAccumulator in BellwetherDecoratedScanningRecipe to the wrapped recipe's stable UUID).

PR #6952, opened 2026-03-12
Use long in Semver.max() for large version segments

Jenson3210:
- Follow-up to #6944. Completes the int-to-long migration in the semver package.

`TildeRange`, `CaretRange`, and `LatestRelease.compare()` already use `long`; `Semver.max()` was the last method using `Integer.parseInt`, which would throw `NumberFormatException` for date-based version segments like `202302104298`.


Jenson3210:
@timtebeek this was not covered by (as it was not needed for that particular recipe failure)
- https://github.com/openrewrite/rewrite/pull/6944

PR #6955, opened 2026-03-12
Skip redundant property override in child when parent declares same dependency

### LINKED ISSUE #6945: UpgradeDependencyVersion with overrideManagedVersion does not add property if <relativePath> is missing
Adding this test case in UpgradeDependencyVersionTests fails - the recipe makes no changes
```java
    @Test
    void upgradeVersionDefinedViaImplicitPropertyInRemoteParentOfLocalParent_withoutRelativePath() {
        rewriteRun(
          spec -> spec.recipe(new UpgradeDependencyVersion("org.flywaydb", "flyway-core", "10.15.0", "", true, null)),
          pomXml(
            """
              <project>
                  <parent>
                      <artifactId>spring-boot-dependencies</artifactId>
                      <version>3.3.0</version>
                  </parent>
                  <groupId>com.mycompany</groupId>
                  <artifactId>my-parent</artifactId>
                  <version>1</version>
                  <dependencies>
                      <dependency>
                          <groupId>org.flywaydb</groupId>
                          <artifactId>flyway-core</artifactId>
                      </dependency>
                  </dependencies>
              </project>
              """,
            """
              <project>
                  <parent>
                      <artifactId>spring-boot-dependencies</artifactId>
                      <version>3.3.0</version>
                  </parent>
                  <groupId>com.mycompany</groupId>
                  <artifactId>my-parent</artifactId>
                  <version>1</version>
                  <properties>
                      <flyway.version>10.15.0</flyway.version>
                  </properties>
                  <dependencies>
                      <dependency>
                          <groupId>org.flywaydb</groupId>
                          <artifactId>flyway-core</artifactId>
                      </dependency>
                  </dependencies>
              </project>
              """
          ),
          mavenProject("my-child",
            pomXml(
              """
                <project>
                    <parent>
                        <groupId>com.mycompany</groupId>
                        <artifactId>my-parent</artifactId>
                        <version>1</version>
                    </parent>
                    <groupId>com.mycompany</groupId>
                    <artifactId>my-child</artifactId>
                    <version>1</version>
                    <dependencies>
                        <dependency>
                            <groupId>org.flywaydb</groupId>
                            <artifactId>flyway-core</artifactId>
                        </dependency>
                    </dependencies>
                </project>
                """
            )
          )
        );
    }
 ```

The root cause is `AddProperty` where the is this check
```java
// If there is a parent pom in the same project, update the property there instead
                if (document.getRoot().getChild("parent")
                        .flatMap(tag -> tag.getChild("relativePath"))
                        .flatMap(Xml.Tag::getValue)
                        .flatMap(v -> v.trim().isEmpty() ? Optional.empty() : Optional.of(v))
                        .isPresent()
```
However relativePath can be ommited and then it's implicitly "../pom.xml" a more proper check could be:
```java
if (getResolutionResult().parentPomIsProjectPom())
```
But this breaks some callers of the AddProperty.

MBoegers:
## Summary
- When `UpgradeDependencyVersion` processes a managed dependency whose version is a property (e.g. `${flyway.version}`), it now checks if a local parent POM also declares the same `<dependency>`. If so, the parent will be processed by the same recipe and will handle the property change — the child skips the redundant override.
- Adds `isDeclaredByLocalParent()` helper that walks up the local parent chain checking for declared dependencies (not just managed ones), complementing the existing `isManagedByLocalParent()` method.

- Fixes #6945

## Test plan
- [x] `UpgradeDependencyVersionTest.upgradeVersionDefinedViaImplicitPropertyInRemoteParentOfLocalParent_withoutRelativePath` — parent and child both declare flyway-core, property only added to parent
- [x] Full `UpgradeDependencyVersionTest` suite passes (no regressions)
- [x] `ChangePropertyValueTest` and `AddPropertyTest` suites pass

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
@@ -316,8 +316,9 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
 
             /**
              * Check if a local parent POM (in the same repository) also declares a dependency
-             * on the same groupId:artifactId. If so, the parent will also be processed by
-             * this recipe and will handle the property change — the child can skip it.
+             * on the same groupId:artifactId without an explicit version. If so, the parent
+             * will also go through the managed-dependency property path and handle the property
+             * change — the child can skip it.
              */
             private boolean isDeclaredByLocalParent(String groupId, String artifactId) {
                 MavenResolutionResult current = getResolutionResult().getParent();
@@ -327,7 +328,8 @@ public class UpgradeDependencyVersion extends ScanningRecipe<UpgradeDependencyVe
                         break; // Reached a non-local (remote) parent
                     }
                     for (Dependency dep : parentPom.getRequested().getDependencies()) {
-                        if (groupId.equals(dep.getGroupId()) && artifactId.equals(dep.getArtifactId())) {
+                        if (groupId.equals(dep.getGroupId()) && artifactId.equals(dep.getArtifactId())
+                                && dep.getVersion() == null) {
                             return true;
                         }
                     }


rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java
```diff
@@ -330,9 +351,12 @@ private Xml.Tag upgradeDependency(ExecutionContext ctx, Xml.Tag t) throws MavenD
                             // if a managed dependency is expressed as a property, change the property value
                             // do this only when a requested bom is absent, otherwise changing property has no effect
                             if (dm != null && isProperty(dm.getRequested().getVersion()) && dm.getRequestedBom() == null) {
-                                doAfterVisit(new ChangePropertyValue(dm.getRequested().getVersion().substring(2,
-                                        dm.getRequested().getVersion().length() - 1),
-                                        newerVersion, overrideManagedVersion, false).getVisitor());
+                                // if a local parent also declares this dependency, it will handle the property change
+                                if (!isDeclaredByLocalParent(d.getGroupId(), d.getArtifactId())) {
```
timtebeek, line 357: Should we maybe only skip if the version is `null` here? To avoid that we keep a version that then pins back a newer version from a parent? Or is that concern already handled separately?

MBoegers: yes good call applied

PR #6967, opened 2026-03-12
Place fenced afterSyntax markers before trailing newline in PlainText diffs

steve-aom-elliott:
## Summary

- Fix `PlainTextPrinter.visitMarkableText()` to emit `afterSyntax` markers (e.g. fenced `{{uuid}}` closers) before the trailing line ending (`\n` or `\r\n`), not after it
- Previously, the closing fence on a PlainText file with a trailing newline created a phantom unterminated line, causing jgit's `DiffFormatter` to produce a spurious `\ No newline at end of file` metadata line and inflate the hunk line count (e.g. `@@ -1,3 +1,4 @@` instead of `@@ -1,3 +1,3 @@`)
- Handles both Unix (`\n`) and Windows (`\r\n`) line endings
- Representation of what that previously looked like once it made it to the diff viewer in the tenant:
image

## Test plan

- [x] Added test: fenced search result on file with trailing newline does not produce "No newline at end of file"
- [x] Added test: fenced search result on file without trailing newline still works correctly
- [x] Added test: fenced search result on file with Windows line endings does not produce "No newline at end of file"
- [x] Added test: fenced search result preserves correct line count in hunk header
- [x] Existing `fencedMarkerPrinterIsApplied` test continues to pass
- [x] Full `InMemoryDiffEntryTest` suite passes

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/PlainTextPrinter.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/PlainTextPrinter.java
@@ -56,7 +56,7 @@ public class PlainTextPrinter<P> extends PlainTextVisitor<PrintOutputCapture<P>>
         // fenced markers (e.g. {{uuid}}) don't create a phantom unterminated
         // line that causes jgit to produce "\ No newline at end of file".
         String lineEnding = null;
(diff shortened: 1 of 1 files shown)


timtebeek:
Nice find, thanks!

PR #6980, opened 2026-03-13
Add favicon at project root

greg-at-moderne:
## What's changed?

Adding the project icon/favicon.

## What's your motivation?

In an attempt for it to show up in Conductor following https://docs.conductor.build/faq#where-does-conductor-get-the-repo-icon


timtebeek:
Out of curiosity: why at the root? I'd been using `.idea/icon.png` to more easily distinguish between OpenRewrite and Moderne projects locally, as that then shows up in the IntelliJ project selector.

greg-at-moderne:
> Out of curiosity: why at the root?

Added above.

greg-at-moderne:
After a restart and waiting a bit (some caches probably) it seems to have worked:
image

I am for keeping it this way.

PR #6982, opened 2026-03-13
Use source file charset when converting diff bytes to String

Jenson3210:
## Summary

- `InMemoryDiffEntry.getDiff()` used `ByteArrayOutputStream.toString()` which relies on the platform default charset
- On Java <18 containers with POSIX/C locale, the default charset may be US-ASCII, corrupting non-ASCII characters in the diff output
- When the corrupted diff is later applied by jgit's `ApplyCommand`, context line bytes don't match the file on disk, causing `PatchApplyException`
- Fix: store the source file's charset and use it explicitly via `new String(patch.toByteArray(), diffCharset)`

- Fixes moderneinc/customer-requests#1994

## Test plan

- [ ] `InMemoryDiffEntryTest.unicodeDiff()` — verifies getDiff() preserves unicode characters in context lines
- [ ] `InMemoryDiffEntryTest.unicodeDiffCanBeAppliedByJgit()` — full pipeline: generate diff → apply with jgit ApplyCommand → verify result
- [ ] All existing `InMemoryDiffEntryTest` tests continue to pass

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/internal/InMemoryDiffEntryTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/internal/InMemoryDiffEntryTest.java
@@ -34,6 +34,7 @@ import org.openrewrite.text.PlainTextParser;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileOutputStream;
+import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


Jenson3210:
@timtebeek / @kmccarp  This is out of my comfort zone (no idea what the impact on platform etc might be... So we should not merge without checking there.

PR #6984, opened 2026-03-14
Upgrade kotlin-reflect from 1.6.10 to 2.2.0

timtebeek:
## Summary
- `kotlin-compiler-embeddable` declares a transitive dependency on `kotlin-reflect:1.6.10`, which is flagged for CVE-2020-29582 (insecure temp file permissions in `createTempDir`/`createTempFile`)
- Adds an explicit `kotlin-reflect` dependency at `2.2.0` (matching the existing `kotlinVersion`) so Gradle's conflict resolution selects the newer version
- This single change resolves the CVE across all **47 downstream recipe repos** that transitively depend on `rewrite-kotlin`

## Verification
- `./gradlew :rewrite-kotlin:dependencies --configuration runtimeClasspath` confirms `kotlin-reflect:1.6.10 -> 2.2.0`
- `./gradlew :rewrite-kotlin:assemble` compiles cleanly

- Ref: moderneinc/dependency-vulnerability-reports#1010

### DIFF
--- a/rewrite-kotlin/build.gradle.kts
+++ b/rewrite-kotlin/build.gradle.kts
@@ -14,6 +14,7 @@ dependencies {
     implementation(project(":rewrite-java"))
 
     implementation(kotlin("compiler-embeddable", kotlinVersion))
+    implementation(kotlin("reflect", kotlinVersion))
     implementation(kotlin("stdlib", kotlinVersion))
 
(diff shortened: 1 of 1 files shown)


timtebeek:
We'd recently already bumped our Kotlin version to 2.2.0; it looks like kotlin-reflect was left behind, but best aligned now
- https://github.com/openrewrite/rewrite/pull/6766

PR #6989, opened 2026-03-16
HasSourceSet should match any SourceFile, not just Java files

Jenson3210:
## Summary
- `HasSourceSet` used `JavaIsoVisitor` and only checked `JavaSourceFile` instances, which meant resource files (YAML, properties, XML, etc.) in a source set were never matched
- Changed to a generic `TreeVisitor<Tree, ExecutionContext>` that checks the `JavaSourceSet` marker on any `SourceFile`
- This allows `HasSourceSet` to be used as a `Preconditions.check()` guard for recipes that operate on non-Java files like `SeparateApplicationYamlByProfile`

## Test plan
- [x] Existing `main()` and `test()` tests still pass (Java source files)
- [x] New `mainResources()` test: properties file in `srcMainResources` is matched by `HasSourceSet("main")`
- [x] New `testResourcesNotMatchedByMain()` test: properties file in `srcTestResources` is NOT matched by `HasSourceSet("main")`
- [x] New `testResources()` test: properties file in `srcTestResources` is matched by `HasSourceSet("test")`

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/search/HasSourceSetTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/search/HasSourceSetTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.*;
+import static org.openrewrite.test.SourceSpecs.text;
 
 class HasSourceSetTest implements RewriteTest {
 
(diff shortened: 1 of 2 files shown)


Jenson3210:
@timtebeek should we worry in this case about recipes that might use this precondition and also expect the result to now be a JavaSourceFile? I think that should ideally be a non-issue as these recipes would have JavaVisitor's isAcceptable, right?

timtebeek (approved):

timtebeek:
Quick scan of usage looks alright: indeed already using JavaVisitor in our Prethink reciped, and I don't think others would have adopted this already in problematic ways.

PR #6990, opened 2026-03-16
Wire up `dataTableOutputDir` for RPC data table CSV exchange

knutwannheden:
When a recipe runs on a remote RPC peer (Python, JS, or C#), any data table rows it inserts via `DataTable.insertRow()` are silently lost. The host process never tells the remote where to store them, and the `ExecutionContext` RPC codec (`rpcSend`/`rpcReceive`) has always been a no-op for data tables — it contains a TODO comment but no implementation. This means recipes like `Find` (which populates `TextMatches`) produce empty data tables when executed through RPC.

The earlier attempt to solve this (#5260, "DataTableStore so that RewriteRpc does not need to push ExecutionContext or DataTable over the wire") proposed a full architectural refactoring: extracting a `DataTableStore` interface, replacing the in-memory `DATA_TABLES` map entirely, and removing `RecipeRun.dataTables`. That PR was closed as stale because the breaking API changes required coordinated updates across multiple repositories (maven plugin, gradle plugin, SaaS).

This PR takes a less invasive approach. Rather than replacing the existing storage mechanism, it adds a file-based CSV channel alongside it. The host passes a directory path to the remote via `PrepareRecipe`. The remote's `Visit`/`BatchVisit` handlers attach a `CsvDataTableStore` to the `ExecutionContext`, which writes `{dataTableName}.csv` files as rows are inserted. After the recipe run, `RpcRecipe.onComplete()` reads the CSVs back and reconstructs typed row objects into the host's `DATA_TABLES` map via reflection. The existing in-memory path is unchanged — non-RPC recipe execution works exactly as before.

## Changes

- **New `CsvDataTableStore`** in `rewrite-core` — writes CSV files with RFC 4180 escaping using per-table `PrintWriter`s, plus a `readIntoContext()` method that reconstructs typed rows via `@Column` annotation metadata and Lombok-generated constructors
- **`PrepareRecipe`** now carries an optional `dataTableOutputDir` field, stored in `PreparedRecipeCache`
- **`Visit`/`BatchVisit` handlers** create a `CsvDataTableStore` on the `ExecutionContext` when `dataTableOutputDir` is configured (shared helper in `PreparedRecipeCache.maybeSetupDataTableStore`)
- **`DataTable.insertRow()`** delegates to `CsvDataTableStore` when present in the context, in addition to the existing in-memory storage
- **`RpcRecipe.onComplete()`** reads CSV files from the output directory and inserts rows into the host's `DATA_TABLES`
- **`RewriteRpc`** exposes `dataTableOutputDir(Path)` fluent setter; when not configured, a temp directory is created automatically and cleaned up on shutdown
- **Builder API** — `dataTableOutputDir(Path)` added to all three subclass builders (`PythonRewriteRpc.Builder`, `JavaScriptRewriteRpc.Builder`, `CSharpRewriteRpc.Builder`)
- **JS handlers** wired up via `RpcConfig` shared state between `PrepareRecipe`, `Visit`, and `BatchVisit`
- **C# handlers** wired up: `PrepareRecipeRequest.DataTableOutputDir`, `GetOrCreateExecutionContext` creates `CsvDataTableStore`
- **Python** was already wired — accepts `dataTableOutputDir` in `PrepareRecipe` and creates `CsvDataTableStore` in `Visit`/`BatchVisit`

## Test plan

- [x] Enabled previously-disabled `runRecipe()` test — `Find` recipe produces `TextMatches` rows through RPC
- [x] New `batchVisitWithDataTables()` test — two `Find` recipes in a `CompositeRecipe` batched via `BatchVisit`, data table rows collected
- [x] All existing `RewriteRpcTest` tests pass (no regressions)
- [x] Full `rewrite-core:test` suite passes
- [x] TypeScript typechecks cleanly
- [x] All three subclass builders (`rewrite-python`, `rewrite-javascript`, `rewrite-csharp`) compile
- [ ] End-to-end verification with Python/JS/C# subprocesses (requires full multi-process test setup)

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
@@ -208,6 +208,11 @@ public class RewriteRpc {
         return this;
     }
 
+    public RewriteRpc dataTableOutputDir(@Nullable Path dataTableOutputDir) {
+        this.dataTableOutputDir = dataTableOutputDir;
+        return this;
+    }
+
     public void shutdown() {
         PrintStream logOut = log.get();
         if (logOut != null) {
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
@@ -183,6 +183,7 @@ public class CSharpRewriteRpc extends RewriteRpc {
         private @Nullable Path csharpServerEntry;
         private @Nullable Path log;
         private Duration timeout = Duration.ofSeconds(60);
+        private @Nullable Path dataTableOutputDir;
         private boolean traceRpcMessages;
         private @Nullable Path workingDirectory;
         private @Nullable Path profileOutputPath;
@@ -226,6 +227,11 @@ public class CSharpRewriteRpc extends RewriteRpc {
 
         public Builder timeout(Duration timeout) {
             this.timeout = timeout;
+            return this;
+        }
+
+        public Builder dataTableOutputDir(@Nullable Path dataTableOutputDir) {
+            this.dataTableOutputDir = dataTableOutputDir;
             return this;
         }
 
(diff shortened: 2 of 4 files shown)


jkschneider:
Considering a more complete form of this.

PR #6992, opened 2026-03-16
Fix C-style array method declaration parsing

### LINKED ISSUE #6398: Java Parser might fail when array brackets `[]` come after method parameters

TIL this is valid Java syntax.

The following test when added to `JavaParserTest`:
```java
    @Test
    void cstyleArrayMethod() {
        rewriteRun(
          // language=java
          java(
            """
            class A {    
                public byte toByteArray()[] {
                  return new byte[]{};
                }
            }
            """
          )
        );
    }
```

fails with:
```
org.opentest4j.AssertionFailedError: [When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error. "A.java":
```

```diff
 class A {
-    public byte toByteArray()[] {
+    public byte toByteArray()[]toByteArray() {
       return new byte[]{};
     }
 }
```

## OSS repro

- https://github.com/hank/litecoinj/blob/b9d74f5d82314487fcdbd484b5776f60f922ecc5/core/src/main/java/com/google/litecoin/core/UnsafeByteArrayOutputStream.java#L111C1-L113C6

Jenson3210:
## Summary

- Fix parser to correctly handle C-style array method declarations (e.g., `byte toByteArray()[]`)
- Add `dimensionsAfterName` field to `J.MethodDeclaration`, consistent with how `NamedVariable` already handles C-style array variables
- Updated all parsers (Java 8/11/17/21/25, Groovy, Kotlin) and RPC layers (Python, JavaScript, C#)

## Problem

C-style array method declarations like `byte toByteArray()[]` are valid Java syntax, but the parser mangled them during printing — producing `byte toByteArray()[]toByteArray()` instead of preserving the original source. The parser's `arrayTypeTree()` method greedily searched forward for `[` brackets, consuming the method name and params as prefix whitespace.

## Solution

Adopted the same approach already used by `VariableDeclarations.NamedVariable` for C-style array variables:

1. Check javac's `endPosTable` to detect whether `[]` follows immediately after the element type in source
2. If not (C-style), only convert the base type as the return type expression
3. After parsing method parameters, consume the `[]` brackets using `arrayDimensions()`
4. Store dimensions in a new `dimensionsAfterName` field on `J.MethodDeclaration` (matching the existing field on `NamedVariable`)
5. The printer outputs the dimensions after the parameter list

## Test plan

- [x] Added test for basic C-style array method (`byte toByteArray()[]`)
- [x] Added test for C-style array method with parameters (`byte getData(int offset, int length)[]`)
- [x] Added test for multi-dimensional C-style arrays (`int get()[][]`)
- [x] Existing `JavaParserTest` and `rewrite-java` tests pass

- Fixes #6398


rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3968,6 +3968,14 @@ public MethodDeclaration withParameters(List<Statement> parameters) {
             return getPadding().withParameters(JContainer.withElements(this.parameters, parameters));
         }
 
+        /**
+         * C-style array dimensions that appear after the method parameters rather than
+         * after the return type. For example: {@code byte toByteArray()[]}.
```
timtebeek, line 4011: Thanks, I hate it. 🙃 

To be clear: the notation style, not you adding support for it!

rewrite-java/src/main/java/org/openrewrite/java/tree/J.java
```diff
@@ -3968,6 +3968,14 @@ public MethodDeclaration withParameters(List<Statement> parameters) {
...
+         */
+        @With
+        @Getter
+        List<JLeftPadded<Space>> dimensionsAfterName;
```
knutwannheden, line 4015: We need a strategy for how to deal with existing LST models that get deserialized and where this property will be `null`. So either we need an `@JsonCreator` constructor to set this to some non-null value or we need to declare it as nullable. In addition we should also add some comment so that we can see to removing this workaround after some time again.

Jenson3210:
Addressed @knutwannheden's feedback: added a `@JsonCreator` factory method on `MethodDeclaration` that accepts `@Nullable dimensionsAfterName` and defaults it to `emptyList()` for backwards-compatible deserialization of older LSTs. Comment added noting it can be removed once older LSTs are out of circulation.

PR #6993, opened 2026-03-16
Use File(URI).toPath() to fix Windows drive letter handling

pstreef:
## Problem

v8.75.6 changed `Paths.get(URI.create(...))` to `Paths.get(URI.create(...).getPath())` to fix percent-encoded non-ASCII characters in file URIs. However, `URI.getPath()` on `file:///C:/Users/...` returns `/C:/Users/...` — the leading `/` before the drive letter causes `Paths.get(String)` to fail on Windows.

## Solution

Replace `Paths.get(URI.create(...).getPath())` with `new File(URI.create(...)).toPath()` in all three locations. `new File(URI)` internally calls `URI.getPath()` for percent-decoding and then `fs.fromURIPath()` which strips the leading `/` on Windows drive letters — handling both non-ASCII characters and Windows paths correctly.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -273,7 +273,7 @@ public class MavenPomDownloader {
 
                     if ("file".equals(scheme)) {
                         // A maven repository can be expressed as a URI with a file scheme
-                        Path path = Paths.get(URI.create(baseUri + "maven-metadata-local.xml").getPath());
+                        Path path = new File(URI.create(baseUri + "maven-metadata-local.xml")).toPath();
                         if (Files.exists(path)) {
                             MavenMetadata parsed = MavenMetadata.parse(Files.readAllBytes(path));
                             if (parsed != null) {
@@ -392,7 +392,7 @@ public class MavenPomDownloader {
     }
 
     private MavenMetadata.@Nullable Versioning directoryToVersioning(String uri, GroupArtifactVersion gav) throws MavenDownloadingException {
-        Path dir = Paths.get(URI.create(uri).getPath());
+        Path dir = new File(URI.create(uri)).toPath();
         if (Files.exists(dir)) {
             try (DirectoryStream<Path> stream = Files.newDirectoryStream(dir)) {
                 List<String> versions = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -273,7 +273,7 @@ public MavenMetadata downloadMetadata(GroupArtifactVersion gav, @Nullable Resolv
 
                     if ("file".equals(scheme)) {
                         // A maven repository can be expressed as a URI with a file scheme
-                        Path path = Paths.get(URI.create(baseUri + "maven-metadata-local.xml").getPath());
+                        Path path = new File(URI.create(baseUri + "maven-metadata-local.xml")).toPath();
```
shanman190, line 276: If you drop the `getPath` on the URI that should fix this.

ie.
```
Paths.get(URI.create(baseUri + "maven-metadata-local.xml"))
```

pstreef: But that also just reverts the original change to unescape non ascii chars like ü.

pdelagrave: I saw that late sorry (CC was waiting for the workflow to go green before merging, and had no instruction to check for new comments before merging, updated CLAUDE.md now).

That's what we used to do until this PR last week: https://github.com/openrewrite/rewrite/pull/6960

shanman190: I'm observing that Java NIO Path does unescape the URI string correctly using the original code, so I think we're still missing something that's not being addressed by either of these two PRs.

pstreef:
```
@Test
void fileUriWithNonAsciiPath(@TempDir Path tempDir) throws Exception {
    Path repo = tempDir.resolve("müller/.m2/repository");
    String uri = "file://" + repo.toAbsolutePath() + "/";

    assertThatThrownBy(() -> Paths.get(URI.create(uri)))
      .isInstanceOf(IllegalArgumentException.class)
      .hasMessageContaining("Bad escape");
}

@Test
void fileUriWithNonAsciiPath(@TempDir Path tempDir) throws Exception {
    Path repo = tempDir.resolve("müller/.m2/repository");
    String uri = "file://" + repo.toAbsolutePath() + "/";

    assertDoesNotThrow(() -> new File(URI.create(uri)).toPath());
}
```

PR #7012, opened 2026-03-17
Propagate SSL/proxy environment to pip and uv subprocesses

pdelagrave:
## Problem

The environment map configured via `PythonRewriteRpc.builder().environment()` (containing `SSL_CERT_FILE`, proxy vars, etc.) was applied to the RPC server process but **not** to:
1. The pip bootstrap subprocess in `bootstrapOpenrewrite()`
2. uv subprocesses spawned via `UvExecutor` (used by `DependencyWorkspace`, `UvLockRegeneration`, etc.)

This caused `pip install openrewrite` and uv commands to fail behind corporate proxies with SSL inspection, because the subprocesses never received `SSL_CERT_FILE`, `PIP_CERT`, or proxy configuration.

```
WARNING: Retrying after connection broken by 'SSLError(SSLCertVerificationError(1,
  '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate'))': /simple/openrewrite/
```

fixes: https://github.com/moderneinc/customer-requests/issues/1891

## Solution

- Apply the `environment` map to the pip `ProcessBuilder` in `bootstrapOpenrewrite()`
- Add an `environment` parameter to `UvExecutor.run()` and thread it explicitly through `DependencyWorkspace`, `UvLockRegeneration`, `SetupCfgParser`, and `RequirementsTxtParser` via method/constructor parameters
- In `PythonRewriteRpc`, pass `commandEnv` to `parseManifest()` parsers and `createSetupPyMarker()` so all downstream subprocess calls inherit SSL/proxy configuration
- Backward-compatible overloads with empty maps for all public APIs

Companion CLI PR: https://github.com/moderneinc/moderne-cli/pull/3520


rewrite-python/src/main/java/org/openrewrite/python/internal/UvExecutor.java
```diff
@@ -55,12 +56,26 @@ public static class RunResult {
      * @return the run result
      */
     public static RunResult run(Path workDir, String uvPath, String... args) throws IOException, InterruptedException {
```
jkschneider: Is this still valid or is leaving it inviting a bug in a situation that fails to forward environment?

rewrite-python/src/main/java/org/openrewrite/python/internal/UvLockRegeneration.java
```diff
@@ -67,6 +69,21 @@ public static Result regenerate(String pyprojectContent) {
      * @return a result containing the new lock file content, or an error message
      */
     public static Result regenerate(String pyprojectContent, @Nullable String existingLockContent) {
```
jkschneider, line 71: same comment as above. AIs tends to want to leave things for backwards compatibility, but if leaving it opens us up for bugs, better to remove aggressively.

pdelagrave: The other one was unused so I removed it, this one is still being used.

PR #7023, opened 2026-03-18
Fix IDE.properties module exclusion

timtebeek:
The IDE.properties mechanism was broken because it always included all projects, making IDE.properties filtering ineffective. This change makes IDE.properties actually exclude projects from the Gradle build.

Changes:
- Only include projects listed in IDE.properties (or all if not present/IDE not running)
- Separate gradle-tooling-model:* into alwaysIncluded since colons can't be represented in .properties files
- Remove the unused dependency substitution block
- Add missing entries to IDE.properties.tmp (rewrite-csharp, rewrite-javascript, rewrite-python)

### DIFF
--- a/IDE.properties.tmp
+++ b/IDE.properties.tmp
@@ -1,9 +1,11 @@
 ########################################################
 # If you are only working on a subset of the modules in this project, you can optimize your
 # IDE to only load those modules. Copy `IDE.properties.tmp` to `IDE.properties` and comment out
-# any lines that correspond to modules you do not want to work on. This will cause Gradle to
-# swap those project dependencies for binary dependencies resolved from either Maven local or
-# the OSS snapshots repository, and speed up your IDE.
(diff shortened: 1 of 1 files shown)


IDE.properties.tmp
```diff
@@ -1,9 +1,11 @@
 ########################################################
```
greg-at-moderne, line 1: Should it be called `.tmp`?
It suggests it's some left-over temporary file.

Maybe `IDE.properties.template`?

timtebeek: hmm thanks, that would be a slightly larger change also including the docs and instructions for our users and customers. I'd think perhaps best not to make that change right now just yet:
https://docs.openrewrite.org/reference/building-openrewrite-from-source#optimizing-your-ide-for-only-modules-you-want-to-work-on

I'd mostly wanted to restore this behavior as we've added C#, JS & Python, which loaded unconditionally with no opt out, yet not everyone will have all four installed locally.

PR #7032, opened 2026-03-18
Fix ChangeType inner class import handling when outer class import is present

MBoegers:
## Summary

- When renaming an inner class (e.g. `foo.A$Builder` → `bar.B$Builder`) where the source file imports the outer class (`import foo.A`), `ChangeType` now correctly force-adds `import bar.B` and omits the redundant `import bar.B.Builder` - since `import bar.B` already makes `B.Builder` accessible.
- Fixes the case where the new outer class import was silently dropped, leaving the code with an unresolved reference.
- Adds an `owningClassSame` helper to detect when only the package changes (not the outer class name), enabling `AddImport` to decide based on actual type references in that case.

## Test plan

- [x] `changeTypeOfInnerClass` - inner class renamed within same outer class name, different package: only outer class import added
- [x] `changeTypeOfInnerCompletely` - inner class renamed to an entirely different outer class: only new outer class import added, no redundant inner import
- [x] `changeTypeOfInnerClassImplicitly` - outer class renamed with inner class following implicitly via import


rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -2519,6 +2520,135 @@ A test() {
...
+                  public class B {
+                    public static class Builder {}
+                  }
+                  """
+              )
+            ),
+          java(
+            """
+              import foo.A;
+              import foo.A.Builder;
```
timtebeek: Why do we have both imports here in the test? I'd think we only need the first one right?

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -2519,6 +2520,135 @@ A test() {
...
+                    public static class Builder {}
+                  }
+                  """,
+                """
+                  package bar;
+
+                  public class B {
+                    public static class Builder {}
+                  }
+                  """
```
timtebeek, line 2691: We shouldn't need this second depends on block right? As ChangeType synthesizes types for what it changes, and we only need to provide the correct types to compile the imports?

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -2519,6 +2520,135 @@ A test() {
         );
     }
 
+    @Test
+    void renameInnerClassWithinSameOuterClass() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("foo.A$Builder", "foo.A$Creator", null))
```
timtebeek, line 2530: Right now we're testing the fairly common case of A.B, but should we also test the variants for A.B.C ? I wonder if those are handled correctly for all the permutations of imported as A, A.B or A.B.C, and then changed as A, A$B or A$B$C.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -253,13 +257,21 @@ private void addImport(JavaType.FullyQualified owningClass) {
...
-                                addImport(owningClass);
-                            }
-                            if (!"java.lang".equals(fullyQualifiedTarget.getPackageName())) {
-                                addImport(fullyQualifiedTarget);
+                        if (!"java.lang".equals(fullyQualifiedTarget.getPackageName()) && hasNoConflictingImport(sf)) {
+                            if (owningClass != null) {
+                                // Force-add the outer class import when the original outer class import was
+                                // explicitly removed (meaning code used the outer class by name, e.g. "B" in "B.Builder").
+                                // When the outer class import was absent, the code uses only the simple inner class
+                                // name (e.g. just "Y"), so AddImport can decide based on actual references.
```
timtebeek: In our comments we first use B, then switch to Y here; that seems needlessly confusing, or was that intentional?

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
MBoegers: good call, revealed a small bug..

rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
MBoegers: Added a comment; it was one of initial observed problems where we don't handle the not needed, but also not unused second import. Some older import guidelines are allowing this.

timtebeek:
I'm going to defer further review on this, since it's such a core recipe to all we do, and I can't for the moment oversee the impact of this change across all languages that leverage this recipe. Please reach out internally for someone confident fitting in a thorough review.

PR #7041, opened 2026-03-19
Simplify csharpPublishLocal to use local feed

knutwannheden:
## Summary
- Replace the temp-project + `dotnet restore` approach in `csharpPublishLocal` with a simple copy of `.nupkg` files to `~/.nuget/local-feed/`
- NuGet populates its global packages cache automatically when consumers restore, so the manual cache population was unnecessary
- Still clears the stale cached version so consumers pick up fresh packages on rebuild

## Test plan
- [x] Ran `csharpPublishLocal`, confirmed packages appear in `~/.nuget/local-feed/`
- [x] Ran `dotnet restore` in recipes-csharp, confirmed it resolves `8.76.0-zlocal` from the local feed

### DIFF
--- a/rewrite-csharp/build.gradle.kts
+++ b/rewrite-csharp/build.gradle.kts
@@ -236,91 +236,36 @@ val csharpPublish by tasks.registering(Exec::class) {
 // Usage: ./gradlew :rewrite-csharp:csharpPublishLocal
 val csharpPublishLocal by tasks.registering {
     group = "csharp"
-    description = "Pack and install C# NuGet packages into local NuGet cache"
+    description = "Pack and publish C# NuGet packages to local feed"
 
     dependsOn(csharpPack)
 
     doLast {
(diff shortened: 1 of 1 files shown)


knutwannheden:
Side note on `recipes-csharp`'s `Directory.Build.props`: the local source detection uses a hardcoded relative path (`../../openrewrite/rewrite`) which only works when the two repos are siblings in a specific directory layout. It breaks when using git worktrees (e.g. `.worktrees/brisk-shrimp/`), since the relative path resolves to `.worktrees/openrewrite/rewrite` which doesn't exist. In that case it silently falls back to the NuGet `*-*` floating version.

This means local cross-repo development only works from the main worktree checkout, not from any `.worktrees/` branch. Worth considering whether to make this more robust (e.g. resolve symlinks, walk up to the real repo root, or use an env var override).

PR #7046, opened 2026-03-19
Gradle 9.4.1 entry in gradle wrapper versions CSV

greg-at-moderne:
## What's changed?

Adding an entry for Gradle 9.4.1 to gradle versions CSV manually.

## What's your motivation?

Tests fail:
```
UpdateGradleWrapperTest > defaultsToLatestRelease() FAILED
        Caused by:
        java.lang.NullPointerException: Cannot invoke "org.openrewrite.gradle.internal.GradleWrapperScriptLoader$Version.getGradlewChecksum()" because "this.resolved" is null
            at org.openrewrite.gradle.internal.GradleWrapperScriptLoader$Nearest.gradlew(GradleWrapperScriptLoader.java:69)
```


timtebeek:
I wonder if for any new version not yet mapped, we should just assume the hashset for the latest version. That tends to be the pattern they use, and avoids runtime issues until the next release of rewrite-gradle comes out.

PR #7055, opened 2026-03-19
Add NoSystemScopeDependencies cleanup recipe

timtebeek:
## Summary

Adds a new cleanup recipe `NoSystemScopeDependencies` that fixes Sonar rule S3422 by removing system-scoped Maven dependencies when they are available in configured repositories.

System scope is problematic because it requires a machine-specific `<systemPath>`, is non-portable, and cannot be packaged in artifacts. The recipe verifies artifact availability before removing the scope, leaving truly local-only dependencies unchanged.

## Changes

- New recipe class: `NoSystemScopeDependencies.java`
- Recipe only removes `<scope>system</scope>` and `<systemPath>` if the dependency version exists in configured repositories
- Works for both `<dependencies>` and `<dependencyManagement>` sections
- Comprehensive test suite with 6 test cases covering all scenarios

## Test Plan

- All 6 tests in `NoSystemScopeDependenciesTest` pass
- Tests cover: artifact available in repo, no systemPath element, dependencyManagement section, artifact not in any repo, non-system scopes, and dependencies without scope

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/NoSystemScopeDependencies.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/NoSystemScopeDependencies.java
@@ -49,10 +49,16 @@ public class NoSystemScopeDependencies extends Recipe {
                         return super.visitTag(tag, ctx);
                     }
 
+                    // Resolve property placeholders (e.g. ${some.version}) against the POM
+                    String resolvedVersion = getResolutionResult().getPom().getValue(version);
+                    if (resolvedVersion == null) {
+                        return super.visitTag(tag, ctx);
+                    }
+
                     try {
                         MavenMetadata metadata = downloadMetadata(groupId, artifactId, ctx);
                         if (metadata.getVersioning() == null ||
-                                !metadata.getVersioning().getVersions().contains(version)) {
+                                !metadata.getVersioning().getVersions().contains(resolvedVersion)) {
                             return super.visitTag(tag, ctx);
                         }
                     } catch (MavenDownloadingException e) {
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/NoSystemScopeDependencies.java
```diff
@@ -0,0 +1,72 @@
...
+import org.openrewrite.xml.tree.Xml;
+
+public class NoSystemScopeDependencies extends Recipe {
+
+    @Getter
+    final String displayName = "Dependencies should not have `system` scope";
+
+    @Getter
+    final String description = "Replaces `<scope>system</scope>` with the default compile scope and removes " +
+            "`<systemPath>` for dependencies that are available in configured repositories.";
```
timtebeek, line 35: So we know this isn't a complete solution for all cases, but it allows folks to push that jar somewhere and then from there see the system scope dropped here.

rewrite-maven/src/main/java/org/openrewrite/maven/cleanup/NoSystemScopeDependencies.java
```diff
@@ -0,0 +1,72 @@
...
+                    String artifactId = tag.getChildValue("artifactId").orElse(null);
+                    String version = tag.getChildValue("version").orElse(null);
+                    if (groupId == null || artifactId == null || version == null) {
+                        return super.visitTag(tag, ctx);
+                    }
+
+                    try {
+                        MavenMetadata metadata = downloadMetadata(groupId, artifactId, ctx);
+                        if (metadata.getVersioning() == null ||
+                                !metadata.getVersioning().getVersions().contains(version)) {
```
steve-aom-elliott: I think the `groupId` and `artifactId` get resolved as part of `downloadMetadata`, but we should probably resolve the `version` against the resolved POM prior to using it for this comparison, in case it's defined with a placeholder.

PR #7086, opened 2026-03-20
ChangeType: visit annotations on package declarations

timtebeek:
## Summary
- Call `super.visitPackage(pkg, ctx)` instead of returning `pkg` directly in `JavaChangeTypeVisitor.visitPackage()`, ensuring annotations on package declarations are properly visited during type changes
- Add test for `ChangeType` with a fully-qualified annotation on a package declaration in `package-info.java`

- Closes https://github.com/openrewrite/rewrite-migrate-java/issues/1018

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
@@ -413,6 +413,26 @@ class ChangeTypeTest implements RewriteTest {
     }
 
     @Test
+    void fullyQualifiedAnnotationOnPackageDeclaration() {
+        rewriteRun(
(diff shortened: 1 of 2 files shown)


rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
```diff
@@ -412,6 +412,26 @@ void annotation() {
...
+          java(
+            """
+              @a.b.c.A1 package foo;
+              """,
+            """
+              @A2 package foo;
+
+              import a.b.d.A2;
+              """,
+            spec -> spec.path("foo/package-info.java")
```
timtebeek, line 430: In all honesty we saw this test pass even when the main code change was rolled back, but the change makes sense, and I suspect it's to do with our testing framework.

jkschneider (approved):

PR #7092, opened 2026-03-23
Add Proxy/Proxies data model to MavenSettings

timtebeek:
## Summary
- Adds `Proxies` and `Proxy` inner classes to `MavenSettings`, following the existing `Servers`/`Server` pattern
- Supports parsing `<proxies>` from Maven's `settings.xml`, including merge, property interpolation, and encrypted password decryption
- Adds a test verifying proxy field deserialization

- Closes https://github.com/openrewrite/rewrite-maven-plugin/issues/543

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
@@ -487,11 +487,19 @@ public class MavenSettings {
 
         public Proxies merge(@Nullable Proxies proxies) {
             final Map<String, Proxy> merged = new LinkedHashMap<>();
+            int nullIndex = 0;
             for (Proxy proxy : this.proxies) {
-                merged.put(proxy.id, proxy);
+                String key = proxy.id != null ? proxy.id : "__null_" + nullIndex++;
+                merged.put(key, proxy);
             }
             if (proxies != null) {
-                proxies.getProxies().forEach(proxy -> merged.putIfAbsent(proxy.getId(), proxy));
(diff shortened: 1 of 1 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -446,6 +475,54 @@ public static class Server {
...
+
+        @Nullable
+        Boolean active;
+
+        @Nullable
+        String protocol;
+
+        String host;
+
+        int port;
```
steve-aom-elliott: FYI `port` has a default value in the documentation, so perhaps this should be a `@Nullable Integer`? The default is `8080`.
- https://maven.apache.org/ref/3.9.14/maven-settings/settings.html#proxy-1

rewrite-maven/src/main/java/org/openrewrite/maven/MavenSettings.java
```diff
@@ -446,6 +475,54 @@ public static class Server {
...
+            }
+            return new Proxies(new ArrayList<>(merged.values()));
+        }
+    }
+
+    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+    @Data
+    @With
+    public static class Proxy {
+        String id;
```
steve-aom-elliott, line 513: Similar to below for `port`, there is actually a default value for this, but given we're using it as a key, I'm not sure how you want to angle this. The default is `default`.
- https://maven.apache.org/ref/3.9.14/maven-settings/settings.html#proxy-1

steve-aom-elliott (approved):

PR #7096, opened 2026-03-23
Fix JavaTemplate expression replacement inside return statements

timtebeek:
## Summary

- Fix `JavaTemplate.apply()` with `mi.getCoordinates().replace()` throwing `IllegalArgumentException` when the target method invocation is inside a `J.Return` (or other non-block expression context)
- When the parent tree is not a `J.Block`, use `parseExpression` instead of `parseBlockStatements` — following the existing precedent in `visitFieldAccess`
- Apply the same fix to `visitNewClass` which has the identical issue

## Test plan

- [x] Added `replaceMethodInvocationInsideReturn` test in `JavaTemplateTest8Test`
- [x] All existing `JavaTemplate*` tests pass

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
@@ -65,6 +65,44 @@ class JavaTemplateTest8Test implements RewriteTest {
         );
     }
 
+    @Test
+    void replaceNewClassInsideReturn() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
+              final JavaTemplate t = JavaTemplate.builder("new StringBuilder(#{any(String)})")
+                .build();
+
+              @Override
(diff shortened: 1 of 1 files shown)


timtebeek:
We already use the same pattern for field access, so I'd think this is fairly safe and fixes similar issues elsewhere:
https://github.com/openrewrite/rewrite/blob/8a181c335554f949826c74ae08dcd03365cccdaf/rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateJavaExtension.java#L236-L243

timtebeek:
Since we already use the same pattern elsewhere I'm inclined to merge this with asynchronous review if needed.. It's specific enough and tested that I don't foresee any issues.

knutwannheden:
Nice! Surprised this never surfaced before...

timtebeek:
It needed a small tweak to fix an incorrect cursor in rewrite-testing-frameworks AssertThrowsOnLastStatement
- https://github.com/openrewrite/rewrite/pull/7123
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/938

PR #7102, opened 2026-03-23
fix(javascript): support member renaming for aliased imports in ChangeImport

danne-jm:
## What's changed?
Fixed a bug in ChangeImport where member renaming failed to update the exported name when used inside an alias (e.g., import { old as alias }). Also added 1 positive and 1 negative unit test in test/javascript/recipes/change-import.test.ts.

## What's your motivation?
Wanted to use the ChangeImport recipe on my own project, however, recipe would execute with no changes applied cause nothing matched the recipe. I have a package that exports logging(), but in my update it is now logV2(), under the same aliased import. This wasn't picked up by the recipe. 

Actual: import { logging as baseLogging } from '@danne-jm/fictional-logger';
Expected change: import { logV2 as baseLogging } from '@danne-jm/fictional-logger';

Reality: Recipe executed and completed, but no changes applied.

Was working my way up making a custom recipe for this until I realized this built-in recipe should support it. It also seemed like the existing code tried handling alias scenarios but missed this? So I figured this might be a blind spot.

### Checklist

### DIFF
--- a/rewrite-javascript/rewrite/test/javascript/recipes/change-import.test.ts
+++ b/rewrite-javascript/rewrite/test/javascript/recipes/change-import.test.ts
@@ -404,6 +404,35 @@ describe("change-import", () => {
                 );
             }, { unsafeCleanup: true });
         });
+
+        test("does not rename if member is local alias only", async () => {
+            const spec = new RecipeSpec();
+            spec.recipe = new ChangeImport({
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks a lot for the immediate fix @danne-jm ; curious to hear what else you're considering building over there.

PR #7105, opened 2026-03-24
Fix AddSettingsPluginRepository duplicate pluginManagement when not first statement

MBoegers:
## Summary

- Fix `addPluginManagementRepos` to scan all statements for an existing `pluginManagement` block instead of only checking `statements.get(0)`
- Add tests for `pluginManagement` not at first position (add-to-existing, skip-when-exists, both Groovy and KTS)

## Problem

- When `settings.gradle(.kts)` has statements before `pluginManagement` (e.g. `rootProject.name = "demo"`), the recipe fails to find the existing block and inserts a duplicate `pluginManagement` at index 0. PR #6959 fixed `J.Return` wrapping for Kotlin but missed this positional lookup bug.

Reported by customer — bug still present in rewrite 8.75.6.

## Solution

Changed the lookup from `statements.get(0)` to a loop that scans all statements. When found, uses the actual index for the in-place update. Also flipped guard clauses for readability (early return on non-Lambda instead of deep nesting).

## Test plan

- [x] Existing tests pass
- [x] New tests: `skipWhenExistsGradlePluginPortalAlone` (Groovy + KTS)
- [x] New tests: `addToExistingPluginManagementNotFirstStatement` (Groovy + KTS, 3-cycle idempotency)
- [x] New tests: `skipWhenExistsPluginManagementNotFirstStatement` (Groovy + KTS)

- Fixes moderneinc/customer-requests#2052

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
@@ -120,7 +120,7 @@ public class AddSettingsPluginRepository extends Recipe {
 
             private List<Statement> addPluginManagementRepos(List<Statement> statements, J pluginManagement) {
                 List<Statement> mapped = ListUtils.map(statements, statement -> {
-                    J.MethodInvocation existing = unwrapPluginManagement(statement);
+                    J.MethodInvocation existing = unwrapMethodCall(statement, "pluginManagement");
                     if (existing == null) {
                         return statement;
                     }
@@ -133,10 +133,7 @@ public class AddSettingsPluginRepository extends Recipe {
                         return lambda.withBody(block.withStatements(ListUtils.map(block.getStatements(), stmt ->
                                 addRepoToRepositoriesBlock(stmt, pluginManagement))));
                     }));
-                    if (statement instanceof J.Return) {
-                        return ((J.Return) statement).withExpression(m);
-                    }
-                    return m;
+                    return rewrap(statement, m);
                 });
                 if (mapped != statements) {
                     return mapped;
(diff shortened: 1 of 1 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
```diff
@@ -117,48 +117,56 @@ public K.CompilationUnit visitCompilationUnit(K.CompilationUnit k, ExecutionCont
...
+                            "pluginManagement".equals(((J.MethodInvocation) statement).getSimpleName())) {
+                        existingPluginManagement = (J.MethodInvocation) statement;
+                        pluginManagementIndex = i;
+                        break;
+                    } else if (statement instanceof J.Return && ((J.Return) statement).getExpression() instanceof J.MethodInvocation &&
+                            "pluginManagement".equals(((J.MethodInvocation) ((J.Return) statement).getExpression()).getSimpleName())) {
+                        existingPluginManagement = (J.MethodInvocation) ((J.Return) statement).getExpression();
+                        pluginManagementIndex = i;
+                        break;
+                    }
```
Jenson3210:
```suggestion
                    if (statement instanceof J.Return && ((J.Return) statement).getExpression() instanceof Statement) {
                        statement = (Statement) ((J.Return) statement).getExpression();
                    }
                    if (statement instanceof J.MethodInvocation &&
                            "pluginManagement".equals(((J.MethodInvocation) statement).getSimpleName())) {
                        existingPluginManagement = (J.MethodInvocation) statement;
                        pluginManagementIndex = i;
                        break;
                    }
```
I know it's not "your code" but we could do something like this here to avoid the duplicate block?

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
```diff
@@ -117,48 +117,56 @@ public K.CompilationUnit visitCompilationUnit(K.CompilationUnit k, ExecutionCont
             }
 
             private List<Statement> addPluginManagementRepos(List<Statement> statements, J pluginManagement) {
-                Statement statement = statements.get(0);
```
timtebeek, line 120: Odd how we had this to begin with.

rewrite-gradle/src/main/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepository.java
```diff
@@ -117,48 +117,56 @@ public K.CompilationUnit visitCompilationUnit(K.CompilationUnit k, ExecutionCont
...
+                int pluginManagementIndex = -1;
                 J.MethodInvocation existingPluginManagement = null;
-                if (statement instanceof J.MethodInvocation &&
-                        "pluginManagement".equals(((J.MethodInvocation) statement).getSimpleName())) {
-                    existingPluginManagement = (J.MethodInvocation) statement;
-                } else if (statement instanceof J.Return && ((J.Return) statement).getExpression() instanceof J.MethodInvocation &&
-                        "pluginManagement".equals(((J.MethodInvocation) ((J.Return) statement).getExpression()).getSimpleName())) {
-                    existingPluginManagement = (J.MethodInvocation) ((J.Return) statement).getExpression();
+                for (int i = 0; i < statements.size(); i++) {
```
timtebeek: Seeing this pattern I wonder if we should instead use `ListUtils.map` across statements, immediately make the change needed when found, and then do an identity check between the original `statements` vs what's returned from `ListUtils.map`. If changed return, and only if those are identical, then introduce a new pluginManagement block. I feel that would reduce a lot of the index wrangling and element extraction. What are your thoughts here.

PR #7113, opened 2026-03-24
Fix MergeYaml sequence entry indentation with objectIdentifyingProperty

timtebeek:
- Closes https://github.com/openrewrite/rewrite/discussions/7107

## Summary
- When merging new entries into a YAML sequence via nested `MergeYamlVisitor` calls (e.g. `key: $` with a nested sequence), `autoFormat` produced incorrect indentation, causing new entries to appear as children of the last existing entry's sub-sequence rather than as siblings.
- Fix by re-indenting the auto-formatted prefix to match existing sequence entries, while preserving comments and blank lines.
- Added a test reproducing the exact scenario from the discussion.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
@@ -20,6 +20,7 @@ import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.Cursor;
 import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.style.GeneralFormatStyle;
 import org.openrewrite.style.Style;
 import org.openrewrite.yaml.MergeYaml.InsertMode;
@@ -363,7 +364,8 @@ public class MergeYamlVisitor<P> extends YamlVisitor<P> {
      */
     private Yaml.Sequence.Entry reindentEntry(Yaml.Sequence.Entry entry, String sequenceEntryIndent) {
         String prefix = entry.getPrefix();
-        String targetIndent = sequenceEntryIndent.replaceFirst("^\n", "");
+        String targetIndent = sequenceEntryIndent.startsWith("\n") ?
+                sequenceEntryIndent.substring(1) : sequenceEntryIndent;
 
         // Compute how much autoFormat over- or under-indented relative to the target
         String actualIndent = prefix.substring(prefix.lastIndexOf('\n') + 1);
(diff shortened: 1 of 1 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
```diff
@@ -330,9 +330,13 @@ private Yaml.Sequence mergeSequence(Yaml.Sequence s1, Yaml.Sequence s2, P p, Cur
...
         });
 
+        // Determine the indentation from existing sequence entries to fix auto-format indentation
+        String existingEntryPrefix = s1.getEntries().get(0).getPrefix();
+        String sequenceEntryIndent = existingEntryPrefix.substring(existingEntryPrefix.lastIndexOf('\n'));
+
         List<Yaml.Sequence.Entry> entries = concatAll(
                 filter(s1.getEntries(), it -> !mutatedEntries.contains(it)),
-                map(mutatedEntries, it -> autoFormat(it, p, cursor)),
+                map(mutatedEntries, it -> reindentEntry(autoFormat(it, p, cursor), sequenceEntryIndent)),
```
Jenson3210, line 340: If we would pass the correct parent entry cursor here, would that make the formatting not do the correct things here?

rewrite-yaml/src/test/java/org/openrewrite/yaml/MergeYamlTest.java
```diff
@@ -3503,4 +3503,42 @@ void insertBeforeAfterPlainScalar() {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite/discussions/7107")
+    @Test
+    void mergeNewEntryIntoSequenceWithObjectIdentifyingProperty() {
+        rewriteRun(
+          spec -> spec.recipe(new MergeYaml(
```
Jenson3210, line 3555: Can we add a test that adds something more complex also? 
eg
```yaml
              fruit:
              - name: blueberry
                specs:
                - id: 2
```

Jenson3210: asking because I find this start-indent a bit weird... 
```yaml
fruit:
  - name: blueberry
    specs:
      - id: 2
```
would normally be a "starting" formatted yaml, no?

rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
timtebeek: Went through some iterations here that culminated in the Javadoc in `reindentEntry`. Welcome to have a go as well if you think the cursor approach is doable too; that had my preference, but didn't want any ripple effects from a small change here.

Jenson3210:
@timtebeek I wrote a unit test that verifies we simply do not support same-column sequences in any formatting. Questioning if we should not try to add support rather than only have some backport that works on this one?

Jenson3210:
- https://github.com/openrewrite/rewrite/pull/7135 is created to support this style in the detection -> we might not need this PR anymore and have support in all yaml recipes

PR #7123, opened 2026-03-24
Fix JavaTemplate statement replacement regression for non-J.Block parent cursors

timtebeek:
## Summary

- Fixes a regression introduced by #7096 where `JavaTemplate.apply()` with `coordinates.replace()` throws `IndexOutOfBoundsException` when the cursor's parent tree is not a `J.Block` and the template produces a statement (not an expression)
- The negative check `!(parent instanceof J.Block)` over-matched, routing statement templates to `parseExpression()` which returns an empty list
- Replaced with a positive check for the specific expression-wrapping contexts (`J.Return`, `J.Assignment`, `J.AssignmentOperation`) that #7096 intended to handle

## Test plan

- [x] Existing `JavaTemplateTest8Test` tests from #7096 still pass (expression replacement inside return statements)
- [x] All `*JavaTemplate*` tests pass

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateJavaExtension.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/JavaTemplateJavaExtension.java
@@ -435,8 +435,11 @@ public class JavaTemplateJavaExtension extends JavaTemplateLanguageExtension {
                     }
                     return m;
                 }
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks yes we added those for returns in https://github.com/openrewrite/rewrite/pull/7096
The issue in rewrite-testing-frameworks was in incorrect cursor, but that ought to be rare.

PR #7129, opened 2026-03-24
Add visitArrayAccess to JavaPrinter for marker rendering

steve-aom-elliott:
## Summary
- `JavaPrinter` was missing a `visitArrayAccess` override, which meant `SearchResult` markers on `J.ArrayAccess` nodes were never rendered during printing
- Every other J node type had an explicit override that called `beforeSyntax`/`afterSyntax` (which renders markers), but `ArrayAccess` fell through to the base `JavaVisitor.visitArrayAccess` which doesn't call `beforeSyntax`
- This prevented taint analysis recipes from marking array access expressions as sinks (e.g., `items[taintedIdx]`)


timtebeek:
Nice find!

PR #7131, opened 2026-03-25
Close Scanner in majorVersion

tejasae-afk:
Came across this in rewrite-core/src/main/java/org/openrewrite/semver/Semver.java while following the call chain — The resource opened there can leak if majorVersion exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Feel free to close if not relevant.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


timtebeek (approved):
Makes sense to me.. Thanks for the immediate fix!

timtebeek:
What happened here?

Issue #7125, opened 2026-03-24
IndentsVisitor does not support same-column sequence indentation style


Jenson3210:
## Problem

YAML supports two common indentation styles for sequences:

**Indented style** (what `IndentsVisitor` produces):
```yaml
fruit:
  - name: apple
    color: red
```

**Same-column style** (not supported):
```yaml
fruit:
- name: apple
  color: red
```

Both are valid YAML, but `IndentsVisitor` always normalizes to indented style by unconditionally applying `indent + indentSize` for sequence entries (line 81 of `IndentsVisitor.java`). There is no `IndentsStyle` option or cursor mechanism to preserve same-column sequences.

## Impact

- This affects `MergeYaml` when used with `objectIdentifyingProperty` on documents that use same-column style. The `autoFormat` call produces indented-style output, requiring a post-hoc `reindentEntry` workaround to correct the indentation back to match existing sibling entries (see PR #7113).

## Test case

Branch [`Jenson3210/same-column-seq-indent`](https://github.com/openrewrite/rewrite/tree/Jenson3210/same-column-seq-indent) adds a test in `IndentsTest` that documents the current behavior: same-column input is normalized to indented output.

## Possible solution

Add support for same-column sequence indentation in `IndentsStyle` and `IndentsVisitor`, potentially via autodetection similar to how `Autodetect.tabsAndIndents` already detects indent size. This would allow `autoFormat` to preserve the document's existing sequence style, eliminating the need for workarounds like `reindentEntry` in `MergeYamlVisitor`.

PR #7135, opened 2026-03-25
Support same-column sequence indentation in YAML

Jenson3210:
## Summary

- Adds `indentedSequences` boolean to `IndentsStyle` to distinguish between indented style (`  - item`) and same-column style (`- item` at parent key column)
- Extends `Autodetect` to detect which sequence indentation style a document uses by comparing sequence entry indent with parent mapping entry indent
- Updates `IndentsVisitor` to respect the detected style, placing dashes at the correct column and computing `lastIndent` so sibling mapping entries align properly

- Fixes https://github.com/openrewrite/rewrite/issues/7125

## Test plan

- [x] Existing `IndentsTest` tests pass (no regressions for indented style)
- [x] New `SameColumnSequenceIndent` nested test class: basic indent correction, preservation of correct indent, nested mappings, multiple entries, mixed-type sequences
- [x] New `AutodetectTest`: detects indented style, detects same-column style, defaults to indented when no sequences present
- [x] Full `rewrite-yaml` test suite passes

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/style/IndentsStyle.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/style/IndentsStyle.java
@@ -21,7 +21,6 @@ import lombok.Data;
 import lombok.With;
 import lombok.experimental.FieldDefaults;
 import org.jspecify.annotations.Nullable;
-import org.openrewrite.internal.ToBeRemoved;
 
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
 @Data
@@ -38,7 +37,6 @@ public class IndentsStyle implements YamlStyle {
     }
 
     @Deprecated
-    @ToBeRemoved(after = "2026-06-01", reason = "All parent runtimes have had few weeks to update")
     public IndentsStyle(int indentSize) {
         this(indentSize, null);
     }


timtebeek:
- Related? https://github.com/openrewrite/rewrite/pull/7174

Jenson3210:
no, AFAICS @greg-at-moderne changed the indentation after the hyphen. 
This PR changes the indentation + detection of the indentation before the hyphen. 

In a yaml: 
```
some-sequence:
-
  some-value
```

```
some-sequence:
<~~My detection + indentation fix~~>-<~~A new line~~>
<~~Greg PR indentation fix for the hyphen on previous line~~>some-value
```

Jenson3210:
Good to see you watching out. Will make some changes to be safer and only work when it really exists but not crash when not.

rewrite-yaml/src/main/java/org/openrewrite/yaml/style/IndentsStyle.java
```diff
@@ -24,5 +24,7 @@
 @Data
 @With
 public class IndentsStyle implements YamlStyle {
-    private int indentSize;
+    int indentSize;
+
+    boolean indentedSequences;
 }
```
timtebeek, line 43: Is there an implied former one arg constructor that we need to make explicit to ease the transition?

timtebeek (approved):
Approved with one more question, just to avoid getting NoSuchMethodError at runtime.

rewrite-yaml/src/main/java/org/openrewrite/yaml/style/IndentsStyle.java
Jenson3210: I do not know / fully understand when these are used and when not, but just added them (+ a null supporting one) for safety here

PR #7150, opened 2026-03-26
C#: Add `CSharpTypeUtils.isNullableType` utility method

knutwannheden:
## Motivation

In `recipes-csharp`, the ThrowIf* recipes need to check whether a parameter's type is `Nullable<T>` (e.g. `int?`) to avoid invalid transformations — `Nullable<T>` doesn't satisfy `IComparable<T>` and lifted comparison operators have different semantics. This check is generally useful for C# recipes, so it belongs in the SDK rather than as a local helper.

## Examples

```java
// Check if a type is Nullable<T>
CSharpTypeUtils.isNullableType(identifier.getType()) // true for int?, bool?, etc.
```

## Summary

- Add `CSharpTypeUtils` class in `rewrite-csharp` at `org.openrewrite.csharp.tree`
- `isNullableType(@Nullable JavaType)` returns `true` when the type is `System.Nullable<T>`
- Delegates to existing `TypeUtils.asParameterized()` and `TypeUtils.isOfClassType()` for consistency
- Takes `JavaType` (not `Identifier`) for broader applicability

## Test plan

- [x] Unit tests for `Nullable<Int32>` → true
- [x] Unit tests for non-nullable parameterized type (`List<String>`) → false
- [x] Unit tests for plain class type → false
- [x] Unit tests for primitive type → false
- [x] Unit tests for null input → false

### DIFF
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/tree/CSharpTypeUtils.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/tree/CSharpTypeUtils.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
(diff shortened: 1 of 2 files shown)


knutwannheden:
I think we can just do pattern matching instead.

```cs
id.Type is JavaType.Parameterized { Type: JavaType.Class { FullyQualifiedName: "System.Nullable" } }
```

PR #7177, opened 2026-03-27
Fix context-sensitive template crash for anonymous class args in assignments

timtebeek:
## Summary
- Fixes `IndexOutOfBoundsException` in `JavaTemplateParser.parseExpression()` when a context-sensitive template replaces a method invocation inside an assignment, where the containing anonymous class is an argument to another method call (e.g., `Preconditions.check()`)
- The stub generator produced invalid Java (`__M__.any(new Foo(){...};)` — semicolon inside method argument) causing the parser to drop `/*__TEMPLATE__*/` comment markers during error recovery
- Conditionally omit the semicolon after anonymous class body when the parent is a `MethodCall`, consistent with the existing pattern used elsewhere in the same method
- Also detect `__TEMPLATE__` comment in `JLeftPadded.before` padding for assignment-context templates where the comment lands in the padding space rather than on a tree node prefix

## Test plan
- [x] Added `replaceContextSensitiveMethodInvocationInsideAssignment` test reproducing the exact crash scenario
- [x] All existing `JavaTemplateTest*` tests pass
- [x] Full `rewrite-java-test` and `rewrite-java` test suites pass

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
@@ -180,6 +180,83 @@ class JavaTemplateTest8Test implements RewriteTest {
         );
     }
 
+    @Test
+    void replaceContextSensitiveMethodInvocationInsideAssignment() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
+              @Override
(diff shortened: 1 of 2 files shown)


timtebeek:
Discussed on Slack with @greg-at-moderne : https://moderneinc.slack.com/archives/C02MMFD9DKJ/p1774613169108859?thread_ts=1774588052.282249&cid=C02MMFD9DKJ

Consensus was looks fine and likely fixes the issue we had seen in today's run.

PR #7185, opened 2026-03-28
AutoFormat support removal of extraneous newlines in more circumstances

sambsnyd:
I noticed that autoformat wasn't fixing extraneous newlines in variable declarations today. 
Here's a proposed fix, @Jenson3210 let me know if you're aware of any reasons this would be problematic that aren't covered by the current tests.


sambsnyd:
Good point Jente. thanks for taking a look.

PR #7202, opened 2026-03-30
Expand star imports in ChangePackage and related recipes when they would create ambiguity

steve-aom-elliott:
## Summary

- When `ChangePackage` renames a star import (e.g., `import javax.validation.constraints.*` → `import jakarta.validation.constraints.*`), it now detects if the new package contains types whose simple names clash with types from other star imports (e.g., `org.hibernate.validator.constraints.*` also has `NotBlank`)
- If ambiguity is detected, only the changed star import is expanded into explicit imports for the types actually referenced in the source, leaving other star imports untouched
- Uses `JavaSourceSet` classpath to enumerate types in each package, then walks AST identifiers to determine which types are actually referenced
- `ChangeType` now also detects when its target type is covered by a star import that would become ambiguous with another star import, and inserts an explicit import to disambiguate
- `ChangeDependencyGroupIdAndArtifactId` (Maven) and `AddDependency` (Maven and Gradle) now update the `JavaSourceSet` marker when changing or adding dependencies, so downstream recipes like `ChangePackage` and `ChangeType` have accurate classpath data for ambiguity detection
- `JavaSourceSetUpdater.addDependency` now tries the project's actual Maven repositories instead of hardcoding Maven Central, fixing failures in air-gapped environments or projects using private repositories
- Shared `withSourceTypesOnClasspath` test helper extracted to `Assertions.java` for reuse across test classes

## Test plan

- [x] `changePackageExpandsStarImportWhenItWouldCreateAmbiguity` — uses real `javax.validation`, `jakarta.validation`, and `org.hibernate.validator` types; verifies star import is expanded to explicit imports when another star import contains a type with the same simple name
- [x] `changePackagePreservesStarImportWhenNoAmbiguity` — verifies star import is preserved when no ambiguity exists
- [x] `changeTypeAddsExplicitImportWhenStarImportsWouldBeAmbiguous` — verifies ChangeType inserts an explicit import when two star imports would both provide the target type's simple name
- [x] `composedWithChangePackageUpdatesImports` — end-to-end test composing `ChangeDependencyGroupIdAndArtifactId` + `ChangePackage` to verify import renaming works when the JavaSourceSet is updated by the dependency change
- [x] Existing `ChangePackageTest`, `ChangeTypeTest`, and `ChangeDependencyGroupIdAndArtifactIdTest` suites pass

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
@@ -363,19 +363,12 @@ public class AddDependency extends ScanningRecipe<AddDependency.Scanned> {
                 if (!maybeJp.isPresent() || !acc.configurationsByProject.containsKey(maybeJp.get())) {
                     return sf;
                 }
-                Optional<JavaSourceSet> maybeSourceSet = sf.getMarkers().findFirst(JavaSourceSet.class);
-                if (!maybeSourceSet.isPresent() || maybeSourceSet.get().getGavToTypes().isEmpty()) {
-                    return sf;
-                }
                 if (updater == null) {
                     updater = new JavaSourceSetUpdater(ctx);
                 }
-                JavaSourceSet updated = updater.addDependency(maybeSourceSet.get(),
-                        groupId, artifactId, acc.resolvedVersion, acc.repositories);
-                if (updated != maybeSourceSet.get()) {
-                    return sf.withMarkers(sf.getMarkers().setByType(updated));
-                }
-                return sf;
+                return JavaSourceSet.updateOnSourceFile(sf, sourceSet ->
+                        sourceSet.getGavToTypes().isEmpty() ? sourceSet :
+                                updater.addDependency(sourceSet, groupId, artifactId, acc.resolvedVersion, acc.repositories));
             }
         };
     }
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
@@ -562,32 +562,21 @@ public class ChangeDependency extends ScanningRecipe<ChangeDependency.Accumulato
                 if (oldDep == null) {
                     return sf;
                 }
-                Optional<JavaSourceSet> maybeSourceSet = sf.getMarkers().findFirst(JavaSourceSet.class);
-                if (!maybeSourceSet.isPresent()) {
-                    return sf;
-                }
-                String cacheKey = maybeJp.get().getId().toString() + ":" + maybeSourceSet.get().getName();
-                JavaSourceSet cached = updatedSourceSets.get(cacheKey);
-                if (cached != null) {
-                    return sf.withMarkers(sf.getMarkers().setByType(cached));
-                }
                 if (updater == null) {
                     updater = new JavaSourceSetUpdater(ctx);
                 }
-                String effectiveNewGroupId = newGroupId != null ? newGroupId : oldDep.getGroupId();
-                String effectiveNewArtifactId = newArtifactId != null ? newArtifactId : oldDep.getArtifactId();
-                ResolvedGroupArtifactVersion newGav = new ResolvedGroupArtifactVersion(
-                        oldDep.getGav().getRepository(),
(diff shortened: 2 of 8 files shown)


rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
```diff
@@ -185,6 +195,24 @@ private boolean usesType(SourceFile sourceFile, ExecutionContext ctx) {
...
+                            try {
+                                DependencyVersionSelector selector = new DependencyVersionSelector(metadataFailures, maybeGp.get(), null);
+                                acc.resolvedVersion = selector.select(
+                                        new GroupArtifact(groupId, artifactId), "implementation",
+                                        version, versionPattern, ctx);
+                                if (acc.resolvedVersion != null) {
+                                    acc.repositories = maybeGp.get().getMavenRepositories();
+                                }
+                            } catch (MavenDownloadingException e) {
+                                // Version resolution failed
```
sambsnyd: Attach a warning marker to the tree

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -500,8 +526,44 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
...
+                ResolvedGroupArtifactVersion newGav = new ResolvedGroupArtifactVersion(
+                        oldDep.getGav().getRepository(),
+                        effectiveNewGroupId, effectiveNewArtifactId, oldDep.getVersion(), null);
+                ResolvedDependency newDep = oldDep
+                        .withGav(newGav)
+                        .withRepository(MavenRepository.MAVEN_CENTRAL);
+                JavaSourceSet updated = updater.changeDependency(maybeSourceSet.get(), oldDep, newDep);
+                if (updated != maybeSourceSet.get()) {
+                    return sf.withMarkers(sf.getMarkers().setByType(updated));
+                }
```
sambsnyd: Every source file within the same source set will have an identical `JavaSourceSet` marker. Is this recomputing per-file? You should really only need to compute it once per source set

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
```diff
@@ -500,8 +526,44 @@ public Properties visitEntry(Properties.Entry entry, ExecutionContext ctx) {
                     }
                     return tree;
                 }
+                // For non-Gradle Java source files, update JavaSourceSet marker
+                if (hasModulesWithOldDep && tree instanceof JavaSourceFile &&
+                    !(tree instanceof G.CompilationUnit) && !(tree instanceof K.CompilationUnit)) {
+                    return updateJavaSourceSet((SourceFile) tree, ctx);
+                }
```
sambsnyd: The guard is probably counter-productive. Groovy and kotlin sources both might have a `JavaSourceSet` marker. Gradle build files wont have one.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
steve-aom-elliott: Changed this to compute once per source set instead.

rewrite-gradle/src/main/java/org/openrewrite/gradle/ChangeDependency.java
steve-aom-elliott: Dropped that portion of the guard related to the compilation units

rewrite-gradle/src/main/java/org/openrewrite/gradle/AddDependency.java
steve-aom-elliott: I believe the marker adding ended up having to be in the regular visiting rather than the scanning visitor (accrued in the scanning visitor, though), but should be added in the last commit.

steve-aom-elliott:
@sambsnyd Ended up squashing most of the previous commits in some form and rebasing the branch again. The changes that I made following your last comment are all part of: [a5abf62](https://github.com/openrewrite/rewrite/pull/7202/commits/a5abf6235363a5d4e86c9514dee022cc90630d43)

PR #7207, opened 2026-03-31
Fix NumberFormatException in version comparison for versions with 6+ numeric parts

Jenson3210:
## Summary

- Extend `RELEASE_PATTERN` from 5 to 9 numeric capturing groups
- Use a named `(?<qualifier>...)` group so qualifier access is independent of group count
- Remove the `Math.min(..., 5)` cap in `LatestRelease.compare()`

## Problem

`RELEASE_PATTERN` only had 5 numeric capturing groups. Versions with 6+ numeric parts (e.g., `1.2.3.4.5.6`) caused two issues:

1. **`compare()` threw `NumberFormatException`**: The loop was capped at 5, but when uncapped the 6th regex group returned the qualifier suffix (e.g., `.6`) instead of a digit, causing `Long.parseLong()` to fail.
2. **`isValid()` rejected valid versions**: The 6th numeric part was captured as the qualifier group (e.g., `.6`), which didn't match any release suffix, so `LatestRelease.isValid()` returned `false`.

## Solution

Extend `RELEASE_PATTERN` to support up to 9 numeric parts and use a named group for the qualifier suffix. This fixes both `compare()` and `isValid()` without changing the comparison logic.

## Test plan

- [x] New tests for versions with 6+ numeric parts (comparison and validation)
- [x] All existing `LatestReleaseTest` tests pass
- [x] All semver tests pass
- [x] `RemoveRedundantDependencyVersions` tests pass

Fixes moderneinc/customer-requests#2121

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/semver/VersionComparator.java
+++ b/rewrite-core/src/main/java/org/openrewrite/semver/VersionComparator.java
@@ -25,7 +25,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public interface VersionComparator extends Comparator<String> {
-    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?<qualifier>[-.+].*)?");
+    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?<qualifier>[-.+].*?$)?");
     String[] RELEASE_SUFFIXES = new String[]{".final", ".ga", ".release"};
     Pattern PRE_RELEASE_ENDING = Pattern.compile("[.-](alpha|a|beta|b|milestone|m|rc|cr|snapshot)[.-]?\\d*$", Pattern.CASE_INSENSITIVE);
 


timtebeek (commented):
Good to introduce a named group indeed.

rewrite-core/src/main/java/org/openrewrite/semver/VersionComparator.java
```diff
@@ -25,7 +25,7 @@
 import java.util.regex.Pattern;
 
 public interface VersionComparator extends Comparator<String> {
-    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?([-.+].*)?");
+    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?<qualifier>[-.+].*)?");
```
timtebeek: I wonder if we should make the qualifier wildcard non-eager with the addition of `?` after `.*`, to avoid potentially consuming a final numerical part as a qualifier instead. Has that case already been explored with a unit test? Not a requirement to make this change before we merge, but it popped up here so I figured ask for your thoughts on the matter, as you've looked closer.
```suggestion
    Pattern RELEASE_PATTERN = Pattern.compile("(\\d+)(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?:\\.(\\d+))?(?<qualifier>[-.+].*?)?");
```

Jenson3210: Actually, turns out non-greedy does break things! `LatestRelease.compare()` uses `.find()` (not `.matches()`), and with `.*?` the qualifier group captures only the minimum — e.g. just `-` instead of `-RC1`. This causes qualifier comparison failures. Keeping it greedy.

Jenson3210: Applied as `.*?$` — the `$` anchor forces capture to end of string, so it's safe with both `.matches()` and `.find()` callers while being non-greedy as a defensive measure.

Jenson3210: @timtebeek let me know what you think about these options

timtebeek: Seems fine indeed to include that end marker; thanks!

PR #7211, opened 2026-03-31
Fix context-sensitive template replacement for standalone expression statements

timtebeek:
## Summary
- Fixes `JavaTemplate` failing with "generated 0 statements" when replacing a standalone method invocation expression statement inside a non-void method using a context-sensitive template
- Root cause: the semicolon terminating the expression was placed after the `if(true)` block close brace and `return` statement in the generated stub, instead of directly after the template content
- Discovered via `UseVisitWithParentCursor` recipe failing on [AbstractLogEnabledToSlf4j.java#L110](https://github.com/openrewrite/rewrite-apache/blob/0e4d4c18c53bda294a215ab0d916855c02db10fd/src/main/java/org/openrewrite/codehaus/plexus/AbstractLogEnabledToSlf4j.java#L110)

## Test plan
- [x] Added `replaceContextSensitiveMethodInvocationStandaloneStatement` test in `JavaTemplateTest8Test`
- [x] All `JavaTemplateTest8Test` tests pass
- [x] All `rewrite-java` and `rewrite-java-test` module tests pass

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaTemplateTest8Test.java
@@ -257,6 +257,83 @@ class JavaTemplateTest8Test implements RewriteTest {
         );
     }
 
+    @Test
+    void replaceContextSensitiveMethodInvocationStandaloneStatement() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new JavaIsoVisitor<>() {
+              @Override
+              public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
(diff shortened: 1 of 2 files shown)


timtebeek:
Test failure unrelated; merging:
```
MavenPomDownloaderTest > WithNativeHttpURLConnectionAndTLS > useHttpWhenHttpsFails() FAILED
    expected: MavenRepository(id=id, uri=http://localhost:37747/maven/, releases=null, snapshots=null, knownToExist=false, username=null, *** timeout=null, deriveMetadataIfMissing=null)
     but was: null
        at app//org.openrewrite.maven.internal.MavenPomDownloaderTest$WithNativeHttpURLConnectionAndTLS.useHttpWhenHttpsFails(MavenPomDownloaderTest.java:357)
```

PR #7213, opened 2026-03-31
Rewrite encoding detection as inverted UTF-8 validator

pstreef:
## Problem

`EncodingDetectingInputStream.guessCharset()` enumerated known-bad UTF-8 byte patterns, which meant any pattern not explicitly listed would slip through undetected. This caused ISO-8859-1 files (e.g. containing German umlauts like `ü` = `0xFC`) to be misdetected as UTF-8, leading to `mod git apply` failures.

Specific gaps: bytes 0xF8-0xFF fell through entirely, and 0xC0/0xC1 (overlong encodings forbidden by RFC 3629) were accepted as valid two-byte sequence leads.

## Solution

Invert the detection logic: instead of listing invalid bytes, the new implementation tracks a `remainingContinuationBytes` counter and only accepts bytes that are explicitly valid in their position. Anything else triggers Windows-1252 fallback.

This is simpler (15 lines vs 30, one int field vs three booleans + two prev-byte trackers), and eliminates detection gaps by default — unknown bytes are rejected rather than silently accepted.

Also tightens the 4-byte lead range from 0xF0-0xF7 to 0xF0-0xF4 (0xF5-0xF7 would encode above U+10FFFF).

### DIFF
--- a/rewrite-csharp/build.gradle.kts
+++ b/rewrite-csharp/build.gradle.kts
@@ -68,9 +68,9 @@ val csharpBuild by tasks.registering(Exec::class) {
     workingDir = csharpDir
     commandLine(findDotnet(), "build")
 
-    inputs.files(fileTree(csharpDir.resolve("OpenRewrite")) { exclude("**/bin/**", "**/obj/**") })
+    inputs.files(fileTree(csharpDir.resolve("OpenRewrite")) { exclude("**/bin/**", "**/obj/**", "**/build/**") })
         .withPathSensitivity(PathSensitivity.RELATIVE)
-    inputs.files(fileTree(csharpDir.resolve("OpenRewrite.Tool")) { exclude("**/bin/**", "**/obj/**") })
+    inputs.files(fileTree(csharpDir.resolve("OpenRewrite.Tool")) { exclude("**/bin/**", "**/obj/**", "**/build/**") })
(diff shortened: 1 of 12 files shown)


rewrite-core/src/test/java/org/openrewrite/internal/EncodingDetectingInputStreamTest.java
```diff
@@ -74,34 +75,59 @@ void skipUTF8BomKnownEncoding() throws Exception {
...
+            "世,              UTF-8,        UTF-8",
+            "가,              UTF-8,        UTF-8",
+            "Hello 世界,      UTF-8,        UTF-8",
+            "café 世界 🌟,    UTF-8,        UTF-8",
+            "Café,           Windows-1252, Windows-1252",
+            "Lyðræðisríki,   Windows-1252, Windows-1252",
+    })
+    void detectsCharsetForEncodedStrings(String text, String sourceCharset, String expectedCharset) throws Exception {
+        try (EncodingDetectingInputStream is = read(text, Charset.forName(sourceCharset))) {
+            assertThat(is.getCharset()).isEqualTo(Charset.forName(expectedCharset));
```
timtebeek, line 92: Source and expected charset always match? Do we then even need both?

PR #7218, opened 2026-03-31
Add ParseError support to rewrite-go parser and RPC

jkschneider:
## Summary
- Capture Go parse failures as `ParseError` instead of failing the entire parse request, allowing partial results to flow through the RPC pipeline
- Add RPC send/receive codecs for `ParseError` and `ParseExceptionResult` marker, matching the Java-side `org.openrewrite.tree.ParseError` contract
- Harden the parser with defensive bounds checks and panic recovery, and register additional Java-side markers (LstProvenance, BuildMetadata, etc.) as `GenericMarker` with a fallback for unknown marker types

## Test plan
- [ ] Verify Go parse with a mix of valid and invalid source files returns `ParseError` for bad files and valid LSTs for good files
- [ ] Verify round-trip RPC serialization of `ParseError` and `ParseExceptionResult` marker
- [ ] Verify that unknown markers from newer rewrite-core versions don't panic the Go process

### DIFF
--- a/rewrite-go/rewrite/pkg/rpc/receive_queue.go
+++ b/rewrite-go/rewrite/pkg/rpc/receive_queue.go
@@ -18,6 +18,9 @@ package rpc
 
 import (
 	"fmt"
+	"strings"
+
+	"github.com/openrewrite/rewrite/pkg/tree"
 )
 
 // ReceiveQueue deserializes RpcObjectData messages from the RPC channel.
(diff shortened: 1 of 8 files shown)


rewrite-go/rewrite/pkg/rpc/receive_queue.go
```diff
@@ -178,9 +181,14 @@ func RegisterFactory(javaClassName string, factory func() any) {
...
 
 // newObj creates a new empty instance by Java class name.
+// Unknown marker types are treated as GenericMarker to avoid panics
+// from markers added in newer versions of rewrite-core.
 func newObj(javaClassName string) any {
 	if factory, ok := factories[javaClassName]; ok {
 		return factory()
 	}
+	if strings.HasPrefix(javaClassName, "org.openrewrite.marker.") {
+		return tree.GenericMarker{}
```
sambsnyd, line 191: Will this be a problem for marker types defined in recipe modules? Admittedly there aren't too many of them but there isn't anything which prevents recipes from using their own marker types.

PR #7232, opened 2026-04-01
Always include Gradle API stubs in GradleParser classpath

pstreef:
## Problem

When `settingsClasspath` or `buildscriptClasspath` is explicitly set on `GradleParser.Builder` (even to an empty list), the parser replaces the default Gradle API stubs entirely instead of merging them. This causes incorrect type attribution on DSL methods like `pluginManagement()`, `repositories()`, and `gradlePluginPortal()`.

For example, `pluginManagement()` ends up with a garbage declaring type like `org.gradle.api.initialization.org.gradle.api.initialization.13789550265458Kt` instead of the correct `org.gradle.api.initialization.Settings`. Recipes that use `MethodMatcher` (e.g. `FindRepository`) then fail to match these methods, causing downstream recipes like `AddSettingsPluginRepository` to not detect existing entries and add duplicates.

This happens in practice when the Gradle plugin extracts the settings buildscript classpath (typically empty when there are no custom settings plugins) and passes it to `GradleParser`.

## Solution

Introduce `mergeClasspath()` which always includes the default Gradle API stubs (gradle-core-api, gradle-base-services, etc.) alongside any externally-provided classpath, using a `LinkedHashSet` to avoid duplicates. All four parser initialization paths (groovy/kotlin x build/settings) now use this method.

Includes a reproducer test that sets `settingsClasspath` to an empty list and verifies `AddSettingsPluginRepository` correctly detects an existing `gradlePluginPortal()`.

- See also #7230 which adds a recipe-level workaround for the same symptom.

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepositoryTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepositoryTest.java
@@ -18,10 +18,19 @@ package org.openrewrite.gradle.plugins;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.SourceFile;
+import org.openrewrite.gradle.GradleParser;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.kotlin.KotlinParser;
+import org.openrewrite.kotlin.tree.K;
 import org.openrewrite.test.RewriteTest;
+import org.openrewrite.test.SourceSpec;
+import org.openrewrite.test.SourceSpecs;
+import org.openrewrite.test.TypeValidation;
+
+import java.nio.file.Paths;
+import java.util.Collections;
 
 import static org.openrewrite.gradle.Assertions.settingsGradle;
 import static org.openrewrite.gradle.Assertions.settingsGradleKts;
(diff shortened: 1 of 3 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/plugins/AddSettingsPluginRepositoryTest.java
```diff
@@ -682,4 +691,35 @@ void addToExistingPluginManagementWithPluginsBlockKts() {
...
+        SourceSpec<K.CompilationUnit> gradle = new SourceSpec<>(K.CompilationUnit.class, "gradle", parser, before, null);
+        gradle.path(Paths.get("settings.gradle.kts"));
+        return gradle;
+    }
+
+    @Test
+    void skipWhenExistsGradlePluginPortalKtsWithEmptyClasspath() {
+        rewriteRun(
+          spec -> spec.recipe(new AddSettingsPluginRepository("gradlePluginPortal", null))
+            .typeValidationOptions(TypeValidation.none()),
```
timtebeek, line 713: Hah, I'd half-hoped we could remove this suppression; can we make it more specific perhaps? As in not suppress all, but just what we need specifically.

shanman190:
When parsing Gradle scripts in process, the Gradle API will be available, so we wouldn't want to add the Gradle 6 implementation of classes back in that we use when out of process as a workaround.

The class value discovered looks like a dynamic proxy which likely extends the target class, so that makes me suspicious that our matcher isn't checking supertypes in `FindRepository`.

sambsnyd (approved):
Makes sense to me

sambsnyd:
> When parsing Gradle scripts in process, the Gradle API will be available, so we wouldn't want to add the Gradle 6 implementation of classes back in that we use when out of process as a workaround.

That's a good point. Of course we also use `GradleParser` in recipes to generate new LST snippets and we _do_ want the API available there as much as possible. 

> The class value discovered looks like a dynamic proxy which likely extends the target class, so that makes me suspicious that our matcher isn't checking supertypes in `FindRepository`.

That sounds plausible

shanman190:
> That's a good point. Of course we also use GradleParser in recipes to generate new LST snippets and we do want the API available there as much as possible.

Yep. In recipes, we don't configure a `classpath` (buildscript or settings), so that is what results in use default loading the Gradle 6 shim already.

pstreef:
Thanks for the detailed feedback. You raise a good point about the in-process case — `DefaultProjectParser` runs inside Gradle where the real API is on the classloader, so unconditionally merging the bundled Gradle API jars isn't right.

The supertype/dynamic proxy angle is interesting too. I'm going to leave this PR as draft — @shanman190, would you be open to picking up the root cause fix? You clearly have better context on the MethodMatcher/FindRepository interaction and the in-process parsing constraints.

In the meantime, #7230 (the recipe-level workaround) is merged and handles the immediate symptom. I have a follow-up ready for Shannon's review comment there too (https://github.com/openrewrite/rewrite/pull/7245 — gating the name-based fallback on `url == null` so it doesn't incorrectly block multiple `maven {}` entries).

MBoegers:
@pstreef @shanman190 Do we need this PR still or is it closed by the other PRs? If so I will close the linked issue.

pstreef:
There is a workaround in the recipe, but the root problem probably still exists. @shanman190 what do you think?

shanman190:
We know that there's an issue, but I don't think it's correct to be adding the Gradle 6 API when we're expecting to be in a Gradle process using its own version of the Gradle API at runtime. That's surely to conflict and result in a different class of issues.

PR #7250, opened 2026-04-02
Go: function expressions wrapped with `StatementExpression`

greg-at-moderne:
## What's changed?

Changing how functions as expression are parsed.
Since they are parsed as `J.MethodDeclaration`, they need to be wrapped into `StatementExpression`s.

## What's your motivation?

Otherwise it fails during RPC with messages like:
```
ClassCastException: J.MethodDeclaration cannot be cast to Expression
```


greg-at-moderne:
Superseded by #7261.

PR #7354, opened 2026-04-12
Auto-deduce valid values for enum @Option fields

timtebeek:
## Summary
- When an `@Option` annotation does not explicitly specify `valid` values and the annotated type is an enum, the valid values are now automatically populated from the enum constants via reflection.
- Adds a private `validValues()` helper in `Recipe.java` used by all three option discovery paths (field, getter method, constructor parameter).
- Explicit `valid` values on the annotation still take precedence over auto-deduction.

## Test plan
- [x] Added tests for enum options on fields, getter methods, and constructor parameters
- [x] Added test verifying explicit `valid` takes precedence over enum constants
- [x] All existing `RecipeBasicsTest` tests continue to pass

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Recipe.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Recipe.java
@@ -52,6 +52,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptySet;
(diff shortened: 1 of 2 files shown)


timtebeek:
I noticed we're not always remembering to set the valid values correctly; e.g.
https://github.com/openrewrite/rewrite/blob/317e4b9add842db1e99dc730417f944a71d4e62b/rewrite-maven/src/main/java/org/openrewrite/maven/IncrementProjectVersion.java#L53-L63

Which then results in an awkward text field:
image

This ensures we get all values exposed unless there's an explicit override.

sambsnyd (approved):

PR #7356, opened 2026-04-12
Make AddProfile idempotent

### LINKED ISSUE #7355: AddProfile is not idempotent: removes and re-adds profile every cycle
## Problem

`org.openrewrite.maven.AddProfile` unconditionally removes and re-adds a profile with the same `id` on every recipe cycle, even when the existing profile already matches the desired state. This makes it non-idempotent.

The relevant code in `AddProfile.java`:

```java
if (maybeProfile.isPresent()) {
    Xml.Tag profile = maybeProfile.get();
    t = (Xml.Tag) new RemoveContentVisitor(profile, false, false)
        .visitNonNull(t, ctx, getCursor().getParentOrThrow());
}
// Always adds, even if the existing profile was identical
Xml.Tag profileTag = Xml.Tag.build(...);
t = (Xml.Tag) new AddToTagVisitor<>(profiles, profileTag)
    .visitNonNull(t, ctx, getCursor().getParentOrThrow());
```

## Impact

Any recipe using `AddProfile` will always report changes, even on the second cycle when nothing should change. Tests that use `AddProfile` need `expectedCyclesThatMakeChanges(2)` as a workaround.

- Found while fixing https://github.com/openrewrite/rewrite-spring-to-quarkus/issues/68.

## Suggested Fix

Before removing and re-adding, compare the existing profile tag content to the one being built. If they're equivalent, return early without making changes.

timtebeek:
## Summary

- Fixes https://github.com/openrewrite/rewrite/issues/7355

`AddProfile` unconditionally removed and re-added a profile with the same `id` on every recipe cycle, even when the existing profile already matched. This uses `SemanticallyEqual` to compare the existing profile to the desired one and returns early when they already match, making the recipe idempotent.

## Changes
- Build the desired profile tag before checking for an existing one
- Compare existing profile using `SemanticallyEqual.areEqual()` and skip modification when content matches
- Remove `expectedCyclesThatMakeChanges(2)` workaround from all tests
- Add `preExistingIdenticalProfile` test verifying no changes when profile already matches

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddProfile.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddProfile.java
@@ -24,6 +24,7 @@ import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.xml.AddToTagVisitor;
 import org.openrewrite.xml.RemoveContentVisitor;
+import org.openrewrite.xml.SemanticallyEqual;
 import org.openrewrite.xml.tree.Xml;
 
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


timtebeek:
Intended for use in
- https://github.com/openrewrite/rewrite-spring-to-quarkus/pull/81
- https://github.com/openrewrite/rewrite-spring-to-quarkus/issues/68

PR #7358, opened 2026-04-13
Prevent marker-only changes from producing phantom diffs

### LINKED ISSUE #7349: ChangeDependencyGroupIdAndArtifactId generates empty diffs on Java files after #7202

steve-aom-elliott:
## Summary

- Dependency-modifying recipes now update `JavaSourceSet` markers (#7202), which can produce `Result` objects with identical before/after printed text. This caused "An empty diff was generated" test failures in downstream repos:

- moderneinc/rewrite-dropwizard — `addsSpringStarterJdbcWhenDropwizardDbIsUsed`
- openrewrite/rewrite-hibernate — `groupIdHibernateOrmRenamed`
- openrewrite/rewrite-migrate-java — `changeAndUpgradeDependencyIfAnnotationJsr250Present`
- openrewrite/rewrite-spring — `flywayStarterOmitsVersionWhenManagedByParent`

Three fixes, each addressing a different layer:

1. **`JavaSourceSet.addTypesForGav` idempotency** — returns `this` when the gavKey already exists with the same types, preventing unnecessary object allocation (especially for LSTs with empty `gavToTypes` maps where the existing guard in `changeDependency` doesn't fire).

2. **`JavaSourceSet.updateOnSourceFile` (cached overload)** — checks whether the source file's current marker is already the cached instance before calling `withMarkers`, preventing O(n) new tree references across all files in a source set when only the first file truly changed.

3. **`RewriteTest` safety net** — accepts marker-only changes (identical before/after text) instead of failing, since these represent legitimate internal state updates. Calls `afterRecipe` so marker assertions still work.

All four downstream test failures were verified to pass with these changes.

- Closes #7349

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/marker/JavaSourceSet.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/marker/JavaSourceSet.java
@@ -66,6 +66,11 @@ public class JavaSourceSet implements SourceSet {
      * @return a new JavaSourceSet with the types added
      */
     public JavaSourceSet addTypesForGav(String gavKey, List<JavaType.FullyQualified> types) {
+        List<JavaType.FullyQualified> existing = gavToTypes.get(gavKey);
+        if (existing != null && existing.equals(types)) {
+            return this;
+        }
+
         List<JavaType.FullyQualified> newClasspath = new ArrayList<>(classpath);
         newClasspath.addAll(types);
 
@@ -178,6 +183,9 @@ public class JavaSourceSet implements SourceSet {
         String cacheKey = maybeJp.get().getId().toString() + ":" + maybeSourceSet.get().getName();
         JavaSourceSet cached = cache.get(cacheKey);
         if (cached != null) {
+            if (cached == maybeSourceSet.get()) {
+                return sf;
+            }
             return sf.withMarkers(sf.getMarkers().setByType(cached));
         }
         JavaSourceSet updated = transform.apply(maybeSourceSet.get());
--- a/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
+++ b/rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java
@@ -562,7 +562,15 @@ public interface RewriteTest extends SourceSpecs {
                             boolean isRemote = result.getAfter() instanceof Remote;
                             if (!isRemote && Objects.equals(result.getBefore().getSourcePath(), result.getAfter().getSourcePath()) &&
                                 Objects.equals(before, actualAfter)) {
-                                fail("An empty diff was generated. The recipe incorrectly changed a reference without changing its contents.");
+                                // Marker-only change (e.g. updated JavaSourceSet classpath) — no visible diff.
+                                // Still call afterRecipe so marker assertions work, then skip.
+                                allResults.remove(result);
+                                try {
+                                    //noinspection unchecked
+                                    ((Consumer<SourceFile>) sourceSpec.afterRecipe).accept(result.getAfter());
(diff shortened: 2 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/marker/JavaSourceSet.java
```diff
@@ -66,6 +66,11 @@ public class JavaSourceSet implements SourceSet {
      * @return a new JavaSourceSet with the types added
      */
     public JavaSourceSet addTypesForGav(String gavKey, List<JavaType.FullyQualified> types) {
+        List<JavaType.FullyQualified> existing = gavToTypes.get(gavKey);
+        if (existing != null && existing.equals(types)) {
```
timtebeek, line 70: I suppose there's no way to prevent the deep equality comparison here? Are there any shortcuts like comparing the sizes? Or is that already built into equals?

timtebeek: Looks like the size check is [one of the first things it does](https://github.com/openjdk/jdk/blob/909d4e758c045a0d8cea52f6a1333839f7d6c43e/src/java.base/share/classes/java/util/ArrayList.java#L637), thankfully. Then this is less of a concern.

timtebeek (approved):

PR #7369, opened 2026-04-14
`Cursor.firstEnclosing()` micro optimization

timtebeek:
## What's changed?

Optimizes `Cursor.firstEnclosing()` by replacing the `CursorIterator` allocation with a direct parent-chain walk, and using `Class.isInstance()` instead of `Class.isAssignableFrom(value.getClass())`.

## Origin/Context

- This fix has been suggested in #7259 and in PR:
- https://github.com/HeshamHM28/rewrite/pull/15

- This is the last remaining "Tier 1" improvement from the [prioritized analysis](https://github.com/openrewrite/rewrite/issues/7259#issuecomment-4232506844).

Kudos to @HeshamHM28 🙇‍♂️

## What's your motivation?

`firstEnclosing()` has 174 call sites and is called thousands of times per recipe run in formatting visitors and trait matchers. Avoiding iterator allocation and using `isInstance()` provides a ~94% speedup on the hot path.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Cursor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Cursor.java
@@ -165,13 +165,14 @@ public class Cursor {
     }
 
     public <T> @Nullable T firstEnclosing(Class<T> tClass) {
-        CursorIterator iter = new CursorIterator(this);
-        while (iter.hasNext()) {
-            Object value = iter.next();
-            if (tClass.isAssignableFrom(value.getClass())) {
+        Cursor current = this;
+        while (current != null) {
+            Object currentValue = current.value;
+            if (tClass.isInstance(currentValue)) {
                 //noinspection unchecked
-                return (T) value;
+                return (T) currentValue;
             }
+            current = current.parent;
         }
         return null;
     }


greg-at-moderne (approved):

rewrite-core/src/main/java/org/openrewrite/Cursor.java
```diff
@@ -165,13 +165,14 @@ public Object next() {
...
 
     public <T> @Nullable T firstEnclosing(Class<T> tClass) {
-        CursorIterator iter = new CursorIterator(this);
-        while (iter.hasNext()) {
-            Object value = iter.next();
-            if (tClass.isAssignableFrom(value.getClass())) {
+        Cursor current = this;
+        while (current != null) {
+            Object currentValue = current.value;
+            if (tClass.isInstance(currentValue)) {
```
timtebeek, line 171: The one part here that gave me pause is the `isAssignableFrom` being changed to `isInstance`. Mostly because we do some creative classloader things. I wonder if that might impact things here. 🤔

timtebeek: These are semantically equivalent for non-null objects: `tClass.isAssignableFrom(obj.getClass())` and `tClass.isInstance(obj)` are defined to return the same result. `isInstance` is the dynamic equivalent of `instanceof`.

Regarding classloader concerns: both methods behave identically across classloader boundaries. If `tClass` is loaded by classloader A and the object's class is loaded by classloader B (loading the same class independently), **both** will return `false`. Neither has special cross-classloader behavior — they rely on the same runtime type identity checks.

The only behavioral difference is that `isInstance(null)` returns `false`, whereas the original would NPE on `null.getClass()`. Since cursor values shouldn't be null, that's a minor safety improvement rather than a risk.

So this is a strict micro-optimization (avoids the `.getClass()` indirection) with no semantic difference in any classloader scenario.

greg-at-moderne: ^--- my Claude gave a very similar comment, using very similar wording when I asked about it before :D

timtebeek: Haha, yes just looking to make sure there's no unexpected side effects from this. Looks to be ok from reading up into it as well.

jkschneider: This looks just too risky to be worthwhile

PR #7370, opened 2026-04-14
`LocalRemoteArtifactCache.hashUri()` micro optimization

timtebeek:
## What's changed?

Optimizes `LocalRemoteArtifactCache.hashUri()` by:
- Reusing `MessageDigest` via `ThreadLocal` instead of calling `MessageDigest.getInstance("SHA-256")` on every invocation
- Using a pre-computed hex lookup table instead of `Integer.toHexString()` + `StringBuilder`

## Origin/Context

- This fix has been suggested in #7259 and in PR:
- https://github.com/HeshamHM28/rewrite/pull/27

- This was the top "Tier 2" improvement from the [prioritized analysis](https://github.com/openrewrite/rewrite/issues/7259#issuecomment-4232506844).

Kudos to @HeshamHM28 🙇‍♂️

## What's your motivation?

~697% per-call speedup by avoiding repeated `MessageDigest` instantiation and using direct hex encoding. Relevant when downloading many dependencies.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/remote/LocalRemoteArtifactCache.java
+++ b/rewrite-core/src/main/java/org/openrewrite/remote/LocalRemoteArtifactCache.java
@@ -30,6 +30,15 @@ import java.util.UUID;
 import java.util.function.Consumer;
 
 public class LocalRemoteArtifactCache implements RemoteArtifactCache {
+    private static final char[] HEX_ARRAY = "0123456789abcdef".toCharArray();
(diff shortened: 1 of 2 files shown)


greg-at-moderne (commented):
For this one, to be honest, I am a little hesitant for two reasons:

- this complicates code, while I am not sure there's evidence it being any kind of hot spot
- we don't have a single (?) test case for the class in question. There's `org.openrewrite.remote.RemoteArchiveTest#gradleWrapperConcurrent`, not sure how well it covers the LRAC class

At the same time - the code looks fine. I don't see any issue.

greg-at-moderne (approved):

PR #7373, opened 2026-04-14
Fix ChangeDependency routing Gradle build files to updateJavaSourceSet instead of gradleVisitor

steve-aom-elliott:
## Summary

- Fix a bug where `ChangeDependency` made no changes to Gradle build files when Java source files were also present in the source set
- The outer `TreeVisitor` in `getVisitor()` routes `JavaSourceFile` instances to `updateJavaSourceSet` for marker updates, but `G.CompilationUnit` (Gradle build scripts) also implements `JavaSourceFile`, causing build files to be misrouted instead of being handled by `gradleVisitor`
- Guard `updateJavaSourceSet` with a check for the `JavaSourceSet` marker — build scripts don't carry one, so they naturally fall through to `gradleVisitor`

## Context

This was introduced in #7202 which added `JavaSourceSet` marker updating to `ChangeDependency`. The `isAcceptable` method correctly distinguishes Gradle files from Java files, but the `visit` method's `instanceof JavaSourceFile` check catches both.

This breaks `ChangeDependency` for any Gradle project where the scanner finds the old dependency (making `hasModulesWithOldDep` true), which is the common case. Downstream, this causes test failures in rewrite-spring for `RenameDeprecatedStartersManagedVersionsTest`, `MigrateToModularStartersTest`, and `Boot3UpgradeTest`.

## Approach evolution

1. **Initial fix**: Added explicit `G.CompilationUnit` / `K.CompilationUnit` type checks before the `JavaSourceFile` branch
2. **Refined**: Switched from type checks to source path extension checks, since `K.CompilationUnit` is used for both `.gradle.kts` build scripts and regular `.kt` source files — the type check would incorrectly skip `JavaSourceSet` updates for Kotlin source files
3. **Final**: Replaced path checks with a `JavaSourceSet` marker presence check — build scripts don't carry this marker, so they fall through to `gradleVisitor`. This is both simpler and more semantically correct.

## Test plan

- [x] New test `changeDependencyWhenJavaSourcesPresent` — fails before fix, passes after
- [x] All existing `ChangeDependency` tests pass

### DIFF
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
@@ -32,6 +32,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.gradle.Assertions.buildGradle;
 import static org.openrewrite.gradle.Assertions.buildGradleKts;
 import static org.openrewrite.gradle.toolingapi.Assertions.withToolingApi;
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.mavenProject;
 import static org.openrewrite.properties.Assertions.properties;
 
 class ChangeDependencyTest implements RewriteTest {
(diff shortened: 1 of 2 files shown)


rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
```diff
@@ -1143,4 +1145,50 @@ void doesNotChangeGroupIdWhenNewCoordinatesDontResolve() {
           )
         );
     }
+
+    @Test
+    void changeDependencyWhenJavaSourcesPresent() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeDependency("commons-lang", "commons-lang", "org.apache.commons", "commons-lang3", "3.11.x", null, null, true)),
```
Jenson3210, line 1152: Did you check if LST's have this marker in a real example? 
If they do, we should probably mock on in here also.

Jenson3210 (approved):

Jenson3210:
1 question, but looks okay to me

rewrite-gradle/src/test/java/org/openrewrite/gradle/ChangeDependencyTest.java
steve-aom-elliott: The JavaParser adds the JavaSourceSet marker during parsing, so the java file is going to get it when it gets parsed as part of `java()`. The Gradle file is not going to get it.

PR #7374, opened 2026-04-14
Fix NPE in ChangeDependencyGroupIdAndArtifactId for POMs without dependencyManagement

MBoegers:
## Summary
- `Pom` fields use `@Builder.Default` to initialize list fields to `emptyList()`, but deserialization can bypass the builder and leave them `null`
- `ChangeDependencyGroupIdAndArtifactId` crashes with NPE when run against a POM that has no `<dependencyManagement>` section
- Added null guards with early returns in three locations:
  - `MavenDependencyPropertyUsageOverlap.filterPropertiesWithOverlapInDependencies()` — guards both `getDependencyManagement()` and `getDependencies()`
  - Both `getSafeVersionPlaceholdersToChange()` methods in `ChangeDependencyGroupIdAndArtifactId` — guards `getDependencies()`

## Test plan
- [x] Added `noDependencyManagementSection()` test that runs `ChangeDependencyGroupIdAndArtifactId` on a POM with no `<dependencyManagement>` section
- [x] Existing tests pass: `./gradlew :rewrite-maven:test --tests "*ChangeDependencyGroupIdAndArtifactId*"`


MBoegers:
merging this after in person discussion with tim

PR #7386, opened 2026-04-15
Remove CLI specific NPE checks

MBoegers:
## What's changed?

To enable the usage of `ChangeDependencyArtifactIdAndGroupId` in the Moderne CLI I added the null checks. After we fix the original bug in the CLI we should roll back the changes.

## What's your motivation?

Don't pollute recipes with CLI specific fixes.

### Checklist


MBoegers:
The fix has been merged upstream; we can remove it again.

timtebeek (approved):

PR #7387, opened 2026-04-16
Harden handleError() against secondary exceptions

Jenson3210:
## Summary

- Wraps the `FindRecipeRunException` re-traversal in `RecipeRunCycle.handleError()` with a try-catch
- If the tree is too broken for node-level error marker insertion, falls back to `Markup.error()` on the entire file
- Ensures errors are always visible even when the tree has invariant violations (e.g. null collection fields after deserialization)

## Context

When a recipe throws on a tree with broken invariants, `handleError()` attempts to re-traverse the tree via `FindRecipeRunException` to attach an error marker to the specific node that failed. If that re-traversal itself throws (because the tree is too broken to traverse), the secondary exception could escape `handleError()`. This change catches that secondary exception and falls back to a file-level error marker instead.

## Test plan

- [x] `./gradlew :rewrite-core:test` passes

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -25,6 +25,7 @@ import org.openrewrite.*;
 import org.openrewrite.config.DeclarativeRecipe;
 import org.openrewrite.internal.ExceptionUtils;
 import org.openrewrite.internal.FindRecipeRunException;
+import org.openrewrite.marker.Markup;
 import org.openrewrite.internal.RecipeRunException;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.marker.*;
@@ -718,8 +719,13 @@ public class RecipeRunCycle<LSS extends LargeSourceSet> {
         ctx.getOnError().accept(t);
 
         if (t instanceof RecipeRunException && after != null) {
-            RecipeRunException vt = (RecipeRunException) t;
-            after = (SourceFile) new FindRecipeRunException(vt).visitNonNull(after, 0);
+            try {
+                RecipeRunException vt = (RecipeRunException) t;
+                after = (SourceFile) new FindRecipeRunException(vt).visitNonNull(after, 0);
+            } catch (Throwable ignored) {
+                // Tree is too broken for node-level marker — fall back to marking the whole file
+                after = Markup.error(after, t);
+            }
         }
 
         // Use the original source file to record the error, not the one that may have been modified by the visitor.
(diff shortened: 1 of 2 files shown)


rewrite-docker/src/main/java/org/openrewrite/docker/internal/DockerPrinter.java
```diff
@@ -157,13 +157,15 @@ public Docker visitArg(Docker.Arg arg, PrintOutputCapture<P> p) {
     public Docker visitEnv(Docker.Env env, PrintOutputCapture<P> p) {
         beforeSyntax(env, p);
         p.append(env.getKeyword());
-        for (Docker.Env.EnvPair pair : env.getPairs()) {
-            visitSpace(pair.getPrefix(), p);
-            visit(pair.getKey(), p);
-            if (pair.isHasEquals()) {
-                p.append("=");
+        if (env.getPairs() != null) {
```
timtebeek: I think Merlin would have fixed this in his AST write PR. Not sure if we'd need this still then, as I would not expect the field to be null ever, but haven't checked (on mobile on plane)

rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
```diff
@@ -718,8 +719,13 @@ private void recordSourceFileResult(@Nullable String beforePath, @Nullable Strin
         ctx.getOnError().accept(t);
 
         if (t instanceof RecipeRunException && after != null) {
-            RecipeRunException vt = (RecipeRunException) t;
-            after = (SourceFile) new FindRecipeRunException(vt).visitNonNull(after, 0);
+            try {
+                RecipeRunException vt = (RecipeRunException) t;
+                after = (SourceFile) new FindRecipeRunException(vt).visitNonNull(after, 0);
+            } catch (Throwable ignored) {
```
timtebeek, line 725: Good catch!

rewrite-docker/src/main/java/org/openrewrite/docker/internal/DockerPrinter.java
Jenson3210: https://github.com/openrewrite/rewrite/blob/main/rewrite-docker/src/main/java/org/openrewrite/docker/DockerVisitor.java#L133 is null safe, but this one is not. 

I will let you decide if we want to have this "defensive" fix in or not :)

timtebeek: Ah I now see neither field had a safe default, so yes then this makes sense, thanks. https://github.com/openrewrite/rewrite/blob/4eb011c83f35d78f9200c1e189b87a79551756b9/rewrite-docker/src/main/java/org/openrewrite/docker/tree/Docker.java#L368

Might even want to mark those as nullable then, unless I misunderstood. Still reading from a small screen.

timtebeek (approved):

timtebeek (approved):

PR #7392, opened 2026-04-16
Omit version when annotation processor is managed by parent POM

### LINKED ISSUE #7384: Lombok Plugin Version is added, even when managed by Spring Boot

I am using rewrite 6.36.0 ran via Maven CLI.

`mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_5`

There is no specific openrewrite configuration in the `pom.xml` file.

See the attached Zip, or https://github.com/scottbusche-usps/open-rewrite-7384, but essentially it's a Spring Boot maven project that doesn't have the `maven-compiler-plugin` block defined, and OpenRewrite will add it along with a Lombok version, instead of letting Boot manage the version. 
[openrewrite-lombok-reproduction.zip](https://github.com/user-attachments/files/26753755/openrewrite-lombok-reproduction.zip)

Note: I have not encountered this with every upgrade we've ran, but I found this way to reproduce it and am opening an issue. If I come across others, I will open a new ticket or add the details here.

This `maven-compiler-plugin` configured correctly, but without the version, since Spring Boot will manage it. https://docs.spring.io/spring-boot/3.5/appendix/dependency-versions/coordinates.html Boot 3.5 is on 1.18.44 as of the latest 3.5.X.

```xml
			<plugin>
				<artifactId>maven-compiler-plugin</artifactId>
				<configuration>
					<annotationProcessorPaths>
						<path>
							<groupId>org.projectlombok</groupId>
							<artifactId>lombok</artifactId>
						</path>
					</annotationProcessorPaths>
				</configuration>
			</plugin>
```

## What did you see instead?
```xml
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <configuration>
          <annotationProcessorPaths>
            <path>
              <groupId>org.projectlombok</groupId>
              <artifactId>lombok</artifactId>
              <version>1.18.40</version>
            </path>
          </annotationProcessorPaths>
        </configuration>
      </plugin>
```

timtebeek:
## Summary
- `AddAnnotationProcessor` now checks `ResolvedPom.getManagedVersion()` before adding a new `<path>`; when the coordinate is already managed via effective `<dependencyManagement>` (e.g. inherited from `spring-boot-starter-parent`), the `<version>` is omitted so modern `maven-compiler-plugin` (3.12+) resolves it from `dependencyManagement`.
- Added an idempotency guard so an existing versionless `<path>` entry is left untouched on subsequent runs.

## Test plan
- [x] New `omitVersionWhenManagedByParent` test covers a parent POM with `<dependencyManagement>` and a reactor child, asserting the added `<path>` has no `<version>`.
- [x] `./gradlew :rewrite-maven:test --tests "org.openrewrite.maven.AddAnnotationProcessorTest"` — 24/24 pass.

- Fixes #7384


rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
```diff
@@ -239,6 +239,12 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                                             continue;
                                         }
 
+                                        if (!child.getChildValue("version").isPresent()) {
+                                            // No explicit version: the path intentionally defers to
+                                            // the effective POM's dependencyManagement. Leave it alone.
+                                            return tg;
+                                        }
```
steve-aom-elliott, line 246: I suppose I wasn't aware that the annotation processor paths could defer to the POM's dependencyManagement to begin with, as I've. generally only seen explicit versions in the past.

rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
```diff
@@ -261,10 +267,17 @@ public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
+                                    // dependencyManagement already manages this coordinate — modern
+                                    // maven-compiler-plugin (3.12+) resolves annotation processor path
+                                    // versions from dependencyManagement automatically.
+                                    String managedVersion = currentMrr.getPom().getManagedVersion(groupId, artifactId, null, null);
+                                    String pathXml = managedVersion != null ?
+                                            String.format("<path>\n<groupId>%s</groupId>\n<artifactId>%s</artifactId>\n</path>",
+                                                    groupId, artifactId) :
+                                            String.format("<path>\n<groupId>%s</groupId>\n<artifactId>%s</artifactId>\n<version>%s</version>\n</path>",
+                                                    groupId, artifactId, version);
+                                    return tg.withContent(ListUtils.concat(tg.getChildren(), Xml.Tag.build(pathXml)));
```
steve-aom-elliott, line 280: Does any of the logic need to be guarded by actually using `maven-compiler-plugin` 3.12+? If this ended up being called and they were on a lower version of the plugin than `3.12`, might it not end up potentially adding the annotation processor path without a version tag, breaking compilation?

rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
```diff
@@ -105,6 +106,94 @@ void addAnnotationProcessor() {
...
+                              <version>0.2.0</version>
+                          </dependency>
+                      </dependencies>
+                  </dependencyManagement>
+                  <build>
+                      <pluginManagement>
+                          <plugins>
+                              <plugin>
+                                  <groupId>org.apache.maven.plugins</groupId>
+                                  <artifactId>maven-compiler-plugin</artifactId>
```
steve-aom-elliott: I suppose as a more general note, are we implied to have a version for this plugin implicitly from somewhere else that I'm just not familiar with?

rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
timtebeek: Yeah, I only recently learned about this too. Maven Compiler Plugin 3.12+ falls back to effective `<dependencyManagement>` when a `<path>` has no version — see [MCOMPILER-391](https://issues.apache.org/jira/browse/MCOMPILER-391).

rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
timtebeek: Great catch! I've just pushed a guard: `compilerPluginSupportsManagedVersions()` looks up the effective maven-compiler-plugin version from `ResolvedPom.getPlugins()` / `getPluginManagement()` and only omits `<version>` when it's `>= 3.12.0`. For older plugin versions (or when the version can't be determined), we still emit the explicit `<version>` as before. Added `keepVersionWhenCompilerPluginIsOlderThan312` to pin that behavior.

rewrite-maven/src/test/java/org/openrewrite/maven/AddAnnotationProcessorTest.java
timtebeek: Good question — the original test was relying on Maven's super-POM default for the maven-compiler-plugin, which hides what's actually being tested. I've updated both tests to declare the plugin version explicitly (`3.13.0` in the positive test, `3.11.0` in the new negative test) so the version-gating logic is visible from the test alone.

steve-aom-elliott:
Going to try adding tests for just a few more scenarios shortly to add a bit more coverage.

steve-aom-elliott (approved):

Issue #1804, opened 2022-05-16
Support changing from a constructor to a factory method in ChangeMethodTargetToStatic


rdifrango:
I would like to be able to convert code that is calling `new Object` to a Factory method perhaps in ChangeMethodTargetToStatic.  

Here's an example of one of the conversions I'd like to see:

```
MyCustomLogger logger = new MyCustomLogger(MyClass.class);
```

That I want to convert to proper SLF4J loggers so it would be:

```
Logger logger = LoggerFactory.getLogger(MyClass.class);
```

I tried the following but it didn't change anything:

```yaml
---
type: specs.openrewrite.org/v1beta/recipe
name: com.difrango.ChangeSimpleTrace
displayName: Change method target to static example
recipeList:
  - org.openrewrite.java.ChangeMethodTargetToStatic:
      methodPattern: com.difrango.MyCustomLogger <constructor>(Class)
      fullyQualifiedTargetTypeName: org.slf4j.LoggerFactory
      returnType: org.slf4j.Logger
```

PR #7399, opened 2026-04-16
Fix/change method target to static constructor wildcard

kdvolder:
## What's changed?

Added a guard clause to `ChangeMethodTargetToStatic.visitNewClass` so it only transforms constructors into static method calls when the method pattern explicitly targets a constructor (e.g., using `<constructor>` or `<init>`).

## What's your motivation?

This fixes a bug where `ChangeMethodTargetToStatic` was incorrectly matching and transforming all `new` object creations when provided with a broad wildcard method pattern like `*(..)`. This broke declarative recipes that used wildcards to match all methods on a class.

- The fix ensures that the "Replace Constructor with Factory Method" refactoring use case (Issue #1804) still works, while preventing unintended side effects on wildcard matches.

An alternative would be changing `MethodMatcher` itself so that the `*` wildcard does not match `<init>`, but that would be a much larger breaking change to core matching behavior that has been in place since 2022.

### Checklist

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodTargetToStaticTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodTargetToStaticTest.java
@@ -463,7 +463,7 @@ class ChangeMethodTargetToStaticTest implements RewriteTest {
             """
               import a.A;
               import b.B;
-
+              
               class C {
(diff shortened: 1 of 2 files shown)


timtebeek:
Yes makes sense; thanks for the immediate fix @kdvolder !

timtebeek (approved):

PR #7414, opened 2026-04-17
Kotlin: declare caller-scope type variables on synthetic Template class

### LINKED ISSUE #7407: BlockStatementTemplateGenerator leaks unbound type variables from caller scope into synthetic template class

rewrite 8.80.0-SNAPSHOT (rewrite-java / rewrite-kotlin)

Via `JavaTemplate.apply` / `KotlinTemplate.apply` with an `Expression` parameter whose declared type is `JavaType.Parameterized` or `JavaType.GenericTypeVariable` bound to a caller-scope type variable.

In a recipe, pass an Expression parameter whose type carries an unbound type variable. For example, when replacing `builder.default { ... }` where `builder` has type `PolymorphicModuleBuilder<T>` (caller-side `<T : Any>`), the parameter is serialized by `Substitutions`/`TypeUtils.toString()` into:

```
class Template {
/*__TEMPLATE__*/__P__./*__p0__*/p<kotlinx.serialization.modules.PolymorphicModuleBuilder<T>>().defaultDeserializer(...)
;
}
```

The caller's type variable `T` is undeclared inside the synthetic `class Template {}`, so the parser rejects the template with `Could not parse as Java: ...`.

Template parameters whose declared type references unbound type variables should be erased (raw type, or bound) during synthetic template generation.

## What did you see instead?

`IllegalArgumentException: Could not parse as Java: ...`

## Suggested fix

In `Substitutions.getTypeName()` / `TypeUtils.toString(JavaType)` as used by the template parameter rendering path, apply type erasure:
- `JavaType.Parameterized` → its raw `getType()`
- `JavaType.GenericTypeVariable` → first bound (or `java.lang.Object` if empty)
- Recurse for `JavaType.Array` element types

This mirrors the JVM bytecode erasure and keeps the synthetic template class self-contained.

## Workaround

Strip generics from each `Expression` parameter before calling `apply`: https://github.com/moderneinc/rewrite-migrate-kotlin/commit/8f2ec1b — see `withRawType` / `toRawType` in `ReplaceKotlinMethod.java`.

timtebeek:
## Summary
- The Kotlin `BlockStatementTemplateGenerator` previously ignored the type variables collected from parameter types, emitting a bare `class Template {` even when substituted parameters referenced caller-scope generics like `Container<T>`. The resulting synthetic class failed to parse because `T` was undeclared.
- This change declares those type variables on the synthetic class using Kotlin syntax (`T : Bound` for a single bound, `where` clause for multi-bound), mirroring the Java generator's `Template<T>` handling.

- Fixes #7407

## Test plan
- [x] Added `KotlinTemplateTest.parameterTypeWithCallerScopeTypeVariable` asserting the generated template contains `class Template<T`
- [x] `./gradlew :rewrite-kotlin:test`


rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinBlockStatementTemplateGenerator.java
```diff
@@ -61,4 +64,44 @@ protected void contextFreeTemplate(Cursor cursor, J j, Collection<JavaType.Gener
             before.insert(0, anImport);
         }
     }
+
+    private static String kotlinTypeParameters(Collection<JavaType.GenericTypeVariable> typeVariables) {
```
greg-at-moderne: Not fully sure, can we re-use the printer here?

greg-at-moderne (approved):

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/template/KotlinBlockStatementTemplateGenerator.java
timtebeek: The `KotlinPrinter` operates on LST nodes (e.g. `J.TypeParameter`), whereas here we only have `JavaType.GenericTypeVariable` metadata from `Substitutions.getTypeVariables()` — there's no direct conversion. ~To improve reuse I've extracted the string-building helper into `KotlinTypeUtils.toKotlinTypeParameters(...)` so it's available to other template/code-generation paths that also start from `JavaType` metadata.~

PR #7417, opened 2026-04-17
`AppendToTextFile`: support glob patterns for the relative file name

timtebeek:
## Summary
- `AppendToTextFile` now accepts a glob expression (e.g. `**/*.txt`) as `relativeFileName`, letting a single recipe invocation append to many existing files.
- Matching is delegated to `PathMatcher` with `glob:` syntax, aligning with the convention in `DeleteSourceFiles`, `RenameFile`, `MoveFile`, and `SetFilePermissions`. A literal path still works because it matches itself as a glob.
- When the pattern contains glob metacharacters no new file is generated (a glob does not specify a concrete path to create); literal paths preserve the existing create-if-missing behavior. The visitor also now guards its `PlainText` cast with `instanceof`, avoiding a `ClassCastException` when a glob matches a non-plaintext file.

## Test plan
- [x] `./gradlew :rewrite-core:test --tests "org.openrewrite.text.AppendToTextFileTest"` — new `Glob` nested class covers multi-file match, non-creation for globs, and skipping non-matching files.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/text/AppendToTextFile.java
+++ b/rewrite-core/src/main/java/org/openrewrite/text/AppendToTextFile.java
@@ -34,7 +34,8 @@ import static java.util.stream.Collectors.toSet;
 @EqualsAndHashCode(callSuper = false)
 public class AppendToTextFile extends ScanningRecipe<AtomicBoolean> {
     @Option(displayName = "Relative file name",
-            description = "File name, using a relative path. If a non-plaintext file already exists at this location, then this recipe will do nothing.",
+            description = "File name, using a relative path. May also be a glob expression (e.g. `**/*.txt`) to match one or more existing files; " +
+                          "when a glob is supplied no new file is created. If a non-plaintext file matches, that file is left unchanged.",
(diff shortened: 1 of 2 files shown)


rewrite-core/src/main/java/org/openrewrite/text/AppendToTextFile.java
```diff
@@ -103,6 +104,11 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
 
     @Override
     public Collection<PlainText> generate(AtomicBoolean fileExists, Collection<SourceFile> generatedInThisCycle, ExecutionContext ctx) {
+        // A glob can match multiple existing files but doesn't specify a concrete path to create.
+        if (relativeFileName.matches(".*[*?\\[{].*")) {
```
Jenson3210, line 108: this might not match on files like backstage nunjuck templates as these contain `{{varname}}`. 

Also not sure if the [] characters might not be misleading here. But probably overthinking this and will be fine in 99% of the cases.

Jenson3210 (approved):

PR #7422, opened 2026-04-17
Fix TypeTable round-trip for Kotlin top-level functions

steve-aom-elliott:
## Summary

Three interlocking bugs in `TypeTable` caused Kotlin top-level functions (top-level regular functions and extension functions) to lose method type attribution when `KotlinParser` received a classpath sourced from a TypeTable-synthesized directory (as `mod` does for downloaded LSTs) rather than the original jar. The symptom: `J.MethodInvocation.getMethodType()` returns `null` for calls like `jacksonObjectMapper()` (from `jackson-module-kotlin`), and `MethodMatcher`/`UsesMethod` therefore cannot match them.

### The three fixes

1. **`ACC_SYNTHETIC` methods were filtered out.** Kotlin extension functions compile to `public static final synthetic` on a `*Kt` facade class, so they were being dropped at write time. Now filter on `ACC_PRIVATE` only (and skip `<clinit>`), matching what's needed for type attribution.

2. **`InnerClasses` attribute was not captured.** Kotlin FIR silently fails to resolve methods on a class whose `@kotlin.Metadata.d1` protobuf references inner classes (lambdas, companion objects, etc.) unless those inner classes are declared in the class's `InnerClasses` attribute. The Writer's `ClassVisitor` didn't override `visitInnerClass`, so every round-tripped class came out with an empty `InnerClasses` attribute. Now serialize each entry as `name,outerName,innerName,access` into a new `innerClasses` TSV column, and replay them on the Reader side before any other class events so ASM writes them back out.

3. **`META-INF/*.kotlin_module` resources were dropped.** The Kotlin compiler uses these files to map package names to their facade classes (`com.fasterxml.jackson.module.kotlin` → `ExtensionsKt`). Without it, the resolver can't find the top-level function's owner. Store `.kotlin_module` files as resource rows using a `classAccess = -2` sentinel with base64-encoded content in the `constantValue` column, plumbed through a new `ResourceConsumer` callback on `Reader.read()` so callers can persist them alongside the synthesized `.class` files.

### Tests

- `MethodMatcherTest#libraryTopLevelFunctionPopulatesMethodType` — asserts method type is non-null for `jacksonObjectMapper()`.
- `MethodMatcherTest#usesMethodMatchesLibraryTopLevelFunction` — end-to-end `UsesMethod` match on the Kotlin top-level extension function.
- `TypeTableTypeAnnotationsTest#columnOrderCorrect` — updated for the added column.

Both Kotlin tests use `KotlinParser.builder().classpath(\"jackson-module-kotlin\")`, which resolves via the shipped TypeTable — exactly the code path that was broken.

## Test plan

- [x] `./gradlew :rewrite-java:test --tests "org.openrewrite.java.internal.parser.*"`
- [x] `./gradlew :rewrite-kotlin:test --tests "org.openrewrite.kotlin.MethodMatcherTest"`
- [ ] Full CI

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
@@ -753,8 +753,7 @@ public class TypeTable implements JavaParserClasspathLoader {
              */
             public void writeResource(String resourcePath, byte[] content) {
                 out.println(TsvRow.resourceRow(groupId, artifactId, version,
-                        resourcePath, Base64.getEncoder().encodeToString(content))
-                        .build());
+                        resourcePath, Base64.getEncoder().encodeToString(content)));
             }
 
             /**
@@ -1018,8 +1017,7 @@ public class TypeTable implements JavaParserClasspathLoader {
                             jar.groupId, jar.artifactId, jar.version,
                             classAccess, className,
                             classSignature, classSuperclassName, classSuperinterfaceSignatures,
-                            classAnnotations, classTypeAnnotations, innerClasses)
-                            .build());
+                            classAnnotations, classTypeAnnotations, innerClasses));
 
                     for (Writer.Member member : members) {
                         member.writeMember(jar, this);
@@ -1067,8 +1065,7 @@ public class TypeTable implements JavaParserClasspathLoader {
                             parameterNames, exceptions,
                             elementAnnotations,
                             serializeParameterAnnotations(parameterAnnotations, descriptor),
-                            typeAnnotations, constantValue)
-                            .build());
+                            typeAnnotations, constantValue));
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java
```diff
@@ -657,13 +714,47 @@ public void write(Path jar) {
...
+                // base64-encoded in the constantValue column. All other columns are empty.
+                out.printf(
+                        "%s\t%s\t%s\t%d\t%s\t%s\t%s\t%s\t%d\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s%n",
+                        groupId, artifactId, version,
+                        -2, resourcePath,
+                        "", "", "",
+                        -1, "", "", "", "", "",
+                        "", "", "",
+                        Base64.getEncoder().encodeToString(content),
+                        "");
```
sambsnyd: Even with the comment this feels a bit opaque, constants sprinkled amongst empty strings like a magical incantation.
I'd rather see a struct with a builder which has a toString, so this could be written more legibly.

Maybe something like:
```java
out.printf(ResourceRow.builder()
        .groupId(groupId)
        .artifactId(artifactId)
        .verison(version)
        .type(ResourceRow.Type.Resource)
        .path(resourcePath)
        .toString()
```

sambsnyd (commented):
The explanation makes sense and the fixes look reasonable.

The various places where loosely typed columns are accessed by magic indexes was already there before this PR. But code like:

```java
                            // Resource row — see Writer.Jar.writeResource
                            String resourcePath = fields[4];
                            String base64 = fields.length > 17 ? fields[17] : "";
```

or
```java
                        String artifactVersion = fields[1] + "-" + fields[2];
```

would be more readable and maintainable if we had some struct or view which gave each field a name and proper type rather than just an index and `String`.

steve-aom-elliott:
Going to add one last update to the actual tests that I forgot, but other than that, should be converted

sambsnyd (approved):

PR #7423, opened 2026-04-18
Minimize the eager resolution of transitive dependencies when upgrading dependency versions.

sambsnyd:
In environments with very slow artifact repositories any unnecessary request can dramatically slow recipe runs.
I observed that UpgradeDependencyVersion is unintentionally eagerly resolving the poms of transitive dependencies post-upgrade of a direct dependency. 
These alterations lead to further deferral of transitive resolution until something actually needs it.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
@@ -996,11 +996,27 @@ public class ResolvedPom {
     }
 
     public List<ResolvedDependency> resolveDependencies(Scope scope, MavenPomDownloader downloader, ExecutionContext ctx) throws MavenDownloadingExceptions {
-        return resolveDependencies(scope, new HashMap<>(), downloader, ctx);
+        return doResolveDependencies(scope, new HashMap<>(), true, downloader, ctx);
     }
 
     public List<ResolvedDependency> resolveDependencies(Scope scope, Map<GroupArtifact, VersionRequirement> requirements,
                                                         MavenPomDownloader downloader, ExecutionContext ctx) throws MavenDownloadingExceptions {
+        return doResolveDependencies(scope, requirements, true, downloader, ctx);
+    }
+
+    /**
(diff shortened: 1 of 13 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/ResolvedPom.java
```diff
@@ -996,11 +996,27 @@ private boolean isAlreadyResolved(GroupArtifactVersion groupArtifactVersion, Lis
...
+     * need the direct coordinates (for example to re-resolve after a version mutation) to avoid the cost of
+     * transitive POM downloads.
+     */
+    public List<ResolvedDependency> resolveDirectDependencies(Scope scope, MavenPomDownloader downloader, ExecutionContext ctx) throws MavenDownloadingExceptions {
+        return doResolveDependencies(scope, new HashMap<>(), false, downloader, ctx);
+    }
+
+    private List<ResolvedDependency> doResolveDependencies(Scope scope, Map<GroupArtifact, VersionRequirement> requirements,
+                                                           boolean resolveTransitives,
+                                                           MavenPomDownloader downloader, ExecutionContext ctx) throws MavenDownloadingExceptions {
```
Jenson3210, line 1019: Not all customers are on cli v4 yet. Won't this break for them also?
Questioning if this change is worth introducing a cli v3 release due to widespread impact.

kmccarp: I'm not sure I understand the concern @Jenson3210, this method is new code? The old code will continue to work?

Jenson3210: Oh, nevermind. I missed the `private` part here as I was going over it... 
Got confused with [the newly added public one](https://github.com/openrewrite/rewrite/pull/7423/changes#diff-e4b347bf30f2a500eab97a61f8416836ae300f0b420b5874322aff8d6bc1aea3R116) here and thought recipes were using this one instead of the other one which would throw `NoSuchMethod`.

PR #7428, opened 2026-04-19
Kotlin: remap builtins on Java-origin method/field types

### LINKED ISSUE #7427: KotlinTypeMapping: remapKotlinBuiltin not applied to method parameter types
## Summary

`MethodMatcher` patterns that reference Java built-in types (e.g. `String`, `Throwable`) do not match equivalent Kotlin constructor/method invocations, because `KotlinTypeMapping.methodInvocationType` records parameter types as `kotlin.String` / `kotlin.Throwable` instead of the remapped `java.lang.String` / `java.lang.Throwable`.

- PR #7364 (\"Align Kotlin type model with Java parser output\") added `remapKotlinBuiltin` in multiple places, but missed method parameter types.

## Reproduction

Given Kotlin source:

```kotlin
import com.fasterxml.jackson.core.JsonGenerationException
import com.fasterxml.jackson.core.JsonGenerator

class Test {
    fun example(gen: JsonGenerator, cause: Throwable) {
        throw JsonGenerationException(\"message\", cause, gen)
    }
}
```

- On rewrite-kotlin 8.80.0-SNAPSHOT (built 2026-04-19, after #7364 merged 2026-04-14), the `J.NewClass` node's `constructorType` parameter types are:

```
[kotlin.String, kotlin.Throwable, com.fasterxml.jackson.core.JsonGenerator]
```

So a `MethodMatcher` like:

```java
new MethodMatcher(
    \"com.fasterxml.jackson.core.JsonGenerationException <constructor>(String, Throwable, com.fasterxml.jackson.core.JsonGenerator)\"
).matches(newClass) // => false
```

...returns `false` on Kotlin, even though the equivalent Java source matches fine. This breaks any recipe that relies on `MethodMatcher` / `ReorderMethodArguments` / etc. for Kotlin sources when the signature references Java built-ins.

- Downstream impact example: https://github.com/moderneinc/customer-requests/issues/2235 — the Jackson 2→3 migration's argument-reordering for `JsonGenerationException` → `StreamWriteException` does not fire on Kotlin code and we had to add a downstream workaround in rewrite-jackson (`ReorderJsonGenerationExceptionArguments`) that accepts both `kotlin.String` and `java.lang.String` (and the `Throwable` equivalents).

## Root cause

In `rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinTypeMapping.kt`, the `methodInvocationType` function (around line 866–1010) builds parameter types like this (around line 992):

```kotlin
for (p in function.valueParameters) {
    val t = type(p.returnTypeRef)
    paramTypes.add(t)  // <-- no remapKotlinBuiltin applied
}
```

- `remapKotlinBuiltin` is already applied at #7364 in:

- Supertypes (line ~485)
- Interfaces
- Generic bounds (lines ~332, ~334, ~1205, ~1207)
- Annotations (lines ~1704, ~1722)

...but not to method parameter types in `methodInvocationType`, nor (likely) to the method return type or to field types. A comprehensive pass is probably needed.

## Proposed fix

Apply `remapKotlinBuiltin` to each parameter type before adding it to `paramTypes` in `methodInvocationType`. Audit the rest of that function (return type, thrown exceptions, receiver type) for the same gap. Also audit `variableType` and related functions for the same class of bug on field/property types.

## How to verify the fix

Add a test in rewrite-kotlin that parses a Kotlin constructor call taking a `String` / `Throwable` and asserts the resulting `JavaType.Method.parameterTypes` contain `java.lang.String` / `java.lang.Throwable` rather than `kotlin.String` / `kotlin.Throwable`.

End-to-end, once this lands and propagates:

1. In rewrite-jackson, revert the workaround recipe `ReorderJsonGenerationExceptionArguments` back to the three standard `ReorderMethodArguments` entries in `src/main/resources/META-INF/rewrite/jackson-2-3.yml`.
2. The two Kotlin tests (`reorderThreeArgsKotlin`, `reorderTwoArgsKotlin`) in `ReorderStreamWriteExceptionArgumentsTest` should still pass.

## Context for whoever picks this up

- Upstream PR aligning Kotlin & Java type model: https://github.com/openrewrite/rewrite/pull/7364
- The remap helper already exists: `KotlinTypeUtils.remapKotlinBuiltin` — just needs to be invoked in more places.
- Downstream workaround that demonstrates the gap (and should be removed once this lands):
  - `rewrite-jackson/src/main/java/org/openrewrite/java/jackson/ReorderJsonGenerationExceptionArguments.java`
  - Note the `isString` / `isThrowable` helpers there accept both `kotlin.*` and `java.lang.*` FQNs precisely to paper over this bug.

timtebeek:
## Summary

- Fixes #7427.
- Closes https://github.com/moderneinc/customer-requests/issues/2235

`KotlinTypeMapping` applied `remapKotlinBuiltin` to supertypes, interfaces, generic bounds, and annotations in #7364, but not to method parameter/return/receiver types or field types — so a Kotlin call to a Jackson `JsonGenerationException("msg", cause, gen)` constructor produced parameter types `kotlin.String` / `kotlin.Throwable`, and a `MethodMatcher` written against the JVM FQNs failed to match. Per review, the remap is scoped to methods/fields whose declaring class is `FirJavaClass` (third-party Java libraries on the classpath); Kotlin-declared signatures like `kotlin.io.ConsoleKt.println(kotlin.Any)` and user properties keep their author-intended Kotlin builtin names. JDK classes go through Kotlin's classfile loader as a different FIR type and are out of scope here — recipes targeting JDK signatures over Kotlin code need a Kotlin-aware `MethodMatcher`.

## Test plan

- [x] `./gradlew :rewrite-kotlin:test` passes
- [x] New `constructorParameterTypesRemapKotlinBuiltinsForJavaOrigin` test asserts Jackson's `JsonGenerationException(String, Throwable, JsonGenerator)` surfaces JVM FQNs and a Java-FQN `MethodMatcher` matches
- [x] Existing assertions for Kotlin-declared methods (`println`, `methodInvocationWithDefaults`) unchanged — they still expect `kotlin.*` parameter types

### DIFF
--- a/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
+++ b/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
@@ -535,7 +535,7 @@ class KotlinTypeMappingTest {
                             @Override
                             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicBoolean atomicBoolean) {
                                 if ("println".equals(method.getSimpleName())) {
-                                    assertThat(method.getMethodType().toString()).isEqualTo("kotlin.io.ConsoleKt{name=println,return=void,parameters=[kotlin.Any]}");
+                                    assertThat(method.getMethodType().toString()).isEqualTo("kotlin.io.ConsoleKt{name=println,return=void,parameters=[java.lang.Object]}");
                                     found.set(true);
                                 }
                                 return super.visitMethodInvocation(method, atomicBoolean);
@@ -543,6 +543,39 @@ class KotlinTypeMappingTest {
                         }.visit(cu, found);
                         assertThat(found.get()).isTrue();
                     })
+              )
+            );
+        }
+
+        @Issue("https://github.com/openrewrite/rewrite/issues/7427")
+        @Test
+        void constructorParameterTypesRemapKotlinBuiltins() {
+            // A Kotlin constructor taking `String` / `Throwable` must surface parameter types
+            // as the Java-parser's `java.lang.String` / `java.lang.Throwable`, not the Kotlin
+            // builtins, so `MethodMatcher` patterns written against the Java FQN match.
+            rewriteRun(
+              kotlin(
+                """
+                  fun example(cause: Throwable): IllegalArgumentException {
+                      return IllegalArgumentException("message", cause)
+                  }
+                  """,
+                spec -> spec.afterRecipe(cu -> {
+                    var matcher = new MethodMatcher("java.lang.IllegalArgumentException <constructor>(String, Throwable)");
+                    var found = new AtomicBoolean(false);
+                    new KotlinIsoVisitor<AtomicBoolean>() {
+                        @Override
+                        public J.NewClass visitNewClass(J.NewClass newClass, AtomicBoolean atomicBoolean) {
+                            var paramTypes = newClass.getConstructorType().getParameterTypes();
+                            assertThat(paramTypes.get(0).toString()).isEqualTo("java.lang.String");
+                            assertThat(paramTypes.get(1).toString()).isEqualTo("java.lang.Throwable");
+                            assertThat(matcher.matches(newClass)).isTrue();
+                            atomicBoolean.set(true);
+                            return super.visitNewClass(newClass, atomicBoolean);
+                        }
+                    }.visit(cu, found);
+                    assertThat(found.get()).isTrue();
+                })
               )
             );
         }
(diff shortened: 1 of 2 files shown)


rewrite-kotlin/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -547,6 +547,39 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Atomi
...
+            // builtins, so `MethodMatcher` patterns written against the Java FQN match.
+            rewriteRun(
+              kotlin(
+                """
+                  fun example(cause: Throwable): IllegalArgumentException {
+                      return IllegalArgumentException("message", cause)
+                  }
+                  """,
+                spec -> spec.afterRecipe(cu -> {
+                    var matcher = new MethodMatcher("java.lang.IllegalArgumentException <constructor>(String, Throwable)");
```
timtebeek, line 564: The problem we saw was specifically with MethodMatcher not being KotlinTypeUtils aware, and hence seeing Kotlin types here instead of their Java equivalents.

rewrite-kotlin/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -535,7 +535,7 @@ fun method() {
                             @Override
                             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicBoolean atomicBoolean) {
                                 if ("println".equals(method.getSimpleName())) {
-                                    assertThat(method.getMethodType().toString()).isEqualTo("kotlin.io.ConsoleKt{name=println,return=void,parameters=[kotlin.Any]}");
+                                    assertThat(method.getMethodType().toString()).isEqualTo("kotlin.io.ConsoleKt{name=println,return=void,parameters=[java.lang.Object]}");
```
timtebeek, line 538: This is where I'm a bit doubtful; do we really want to map all Kotlin types to their Java equivalents, even for packages like kotlin.io ?

greg-at-moderne: I suppose not all Kotlin types have a direct Java equivalent. E.g. `Nothing`?
So I guess the answer would be no? For consistency.

greg-at-moderne (commented):
.

rewrite-kotlin/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
timtebeek: Scoped to Java-origin classes (FirJavaClass) in 97295cd43. The motivating Jackson case still works; Kotlin-declared signatures like `kotlin.io.ConsoleKt.println(kotlin.Any)` are now preserved. JDK classes go through Kotlin's classfile loader as a different FIR type, so they're out of scope here — that'd need a Kotlin-aware `MethodMatcher`.

greg-at-moderne (approved):

PR #7437, opened 2026-04-21
Fix local YAML recipe scanning to be recursive

kdvolder:
## What's changed?
Fixes a bug introduced in #7338 where `ClasspathScanningLoader` failed to recursively scan subdirectories for YAML recipes when loading from a local path (e.g., `target/classes` during `mvn test`). 

Replaced the non-recursive `Files.newDirectoryStream` with `Files.walk` in `listYamlResourcesFromPath`, and added handling for `UncheckedIOException` which can be thrown by the stream traversal if it encounters unreadable files/directories.

## What's your motivation?
We have a large suite of YAML recipes organized into subdirectories (e.g., `META-INF/rewrite/framework60/`). After upgrading to 8.79.6, all of these recipes started failing to load during local Maven test runs with "Recipe(s) not found" errors.

The exception handling around the `Files.walk` stream. We catch both `IOException` (for the initial directory open) and `UncheckedIOException` (for any unreadable subdirectories encountered during traversal) to maintain the previous behavior of gracefully ignoring unreadable files.

@sam-snyder (since this relates to the changes in #7338)

The only workaround for users currently is to flatten their entire `META-INF/rewrite` directory structure, which is not ideal for large recipe collections.

## Any additional context
A witness test `scanYamlInSubdirectory` was added to `ClasspathScanningLoaderTest` to verify recursive YAML loading works correctly.

### Checklist

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/ClasspathScanningLoader.java
@@ -39,6 +39,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+import java.util.stream.Stream;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the immediate fix!

greg-at-moderne (approved):

PR #7455, opened 2026-04-22
Avoid triggering full transitive resolution from GradleDependencyConfiguration.findResolvedDependency

steve-aom-elliott:
## What

`GradleDependencyConfiguration.findResolvedDependency(groupId, artifactId)` iterates the configuration's direct dependencies looking for a match. It was calling [`getDirectResolved()`](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/marker/GradleDependencyConfiguration.java#L102-L108), which triggers a full transitive download if the configuration has been marked for re-resolution (e.g. after `AddDependency` / `UpgradeDependencyVersion` / `ChangeDependency` mutations).

Switch it to `getDirectResolvedShallow()`, which resolves only the direct dependencies and defers the transitive closure until a caller actually walks it.

## Why

The method is called from `ChangeDependency`, `RemoveRedundantDependencyVersions`, `RemoveRedundantSecurityResolutionRules`, `GradleConfigurationFilter`, and several downstream recipe repositories (`AddJupiterDependencies` in `rewrite-testing-frameworks`, `MergeBootstrapYamlWithApplicationYaml` in `rewrite-spring`, `AddJaxbRuntime` / `AddJaxwsRuntime` in `rewrite-migrate-java`, etc.).

- In a migration recipe chain that mutates many dependencies per `build.gradle`, each `findResolvedDependency` call was paying the full transitive-resolution cost via `LazyResolutionContext.resolve()` — which re-downloads POMs from the configured repositories. Building on Sam's [`getDirectResolvedShallow()`](https://github.com/openrewrite/rewrite/pull/7423) work, this collapses another common hot path onto the shallow resolution.

## How

- `findResolvedDependency(String, String)` now iterates `getDirectResolvedShallow()`.
- Javadoc clarified: the method matches direct dependencies of the configuration. To match a transitive, callers should iterate `getResolved()` (the flat resolved list) directly.
- Updated `GradleProjectTest#changeConstraint`, which used `findResolvedDependency` to look up `jackson-databind` (a transitive of `rewrite-core`). It now streams `getResolved()` and filters.

## Tests

- Full `rewrite-gradle:test` suite passes (20m 5s).
- `GradleProjectTest`, `ChangeDependencyTest`, `RemoveRedundantDependencyVersionsTest`, `RemoveRedundantSecurityResolutionRulesTest`, `UpgradeDependencyVersionTest` pass.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/marker/GradleDependencyConfiguration.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/marker/GradleDependencyConfiguration.java
@@ -439,8 +439,13 @@ public class GradleDependencyConfiguration implements Serializable, Attributed {
         return null;
     }
 
+    /**
+     * Find a direct dependency of this configuration matching the given coordinates. This does not walk into the
+     * transitive closure of each direct dependency; use {@link #getResolved()} and iterate there if you need to
+     * match a transitive. Uses {@link #getDirectResolvedShallow()} so it does not trigger a full transitive download.
+     */
     public @Nullable ResolvedDependency findResolvedDependency(@Nullable String groupId, String artifactId) {
-        for (ResolvedDependency d : getDirectResolved()) {
+        for (ResolvedDependency d : getDirectResolvedShallow()) {
             ResolvedDependency dependency = d.findDependency(groupId == null ? "" : groupId, artifactId);
             if (dependency != null) {
                 return dependency;
--- a/rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
+++ b/rewrite-gradle/src/test/java/org/openrewrite/gradle/marker/GradleProjectTest.java
@@ -554,10 +554,16 @@ class GradleProjectTest implements RewriteTest {
                   .extracting(GradleDependencyConstraint::getRequiredVersion)
                   .isEqualTo("2.19.2");
 
-                // Verify the runtimeClasspath has the updated resolved version
+                // Verify the runtimeClasspath has the updated resolved version. jackson-databind is a transitive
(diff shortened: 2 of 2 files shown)


Jenson3210 (approved):

Jenson3210 (commented):
Note: `ResolvedDependency.findDependency()` recursively walks `this.dependencies`. With `getDirectResolvedShallow()`, those dependency lists are empty (transitive POMs were never downloaded), so `findDependency()` effectively degrades to matching on the direct dependency's own coordinates only — it won't walk into transitives and won't trigger any downloads.

This means callers like `AddJupiterDependencies` (rewrite-testing-frameworks) that search for `junit-jupiter-api` via `findResolvedDependency("org.junit.jupiter", "junit-jupiter-api")` will no longer find it if it's only present as a transitive (e.g., pulled in by `spring-boot-starter-test` → `junit-jupiter`). The recipe would then redundantly add `junit-jupiter` — harmless but slightly noisy.

Worth noting in the Javadoc that this is now a direct-only match, since the method name `findResolvedDependency` doesn't make that obvious and callers in other repos (rewrite-testing-frameworks, rewrite-spring, rewrite-migrate-java) may rely on transitive matching.

Jenson3210 (changes requested):
See comment above — findDependency() still walks this.dependencies recursively, but with shallow resolution those lists are empty. This silently changes findResolvedDependency from "search direct + transitive" to "search direct only" without updating callers in other repos that rely on transitive matching (AddJupiterDependencies, MergeBootstrapYaml, AddJaxbRuntime, etc.).

Jenson3210:
Perhaps we can build a smarter loop that stops the moment it is found and does not resolve the later/deeper dependencies anymore? 
Not sure how we can overcome this one

PR #7460, opened 2026-04-22
Introduce a JavaProject dirtiness registry for dependency-mutation safety

steve-aom-elliott:
## Summary

When dependency-mutating recipes (`AddDependency`, `ChangeDependency`, `ChangeDependencyGroupIdAndArtifactId`, `UpgradeDependencyVersion`, `RemoveDependency`, etc.) run, `JavaSourceSetUpdater` previously downloaded the new dependency's JAR and scanned it for types so that downstream ambiguity-sensitive recipes (`ChangePackage`, `ChangeType`, `OrderImports`, `AddImport`) could see the post-change classpath. On large multi-module projects driven by recipes like `UpgradeSpringBoot_4_0`, that JAR retrieval can dominate recipe wall time — in one reference run ~97s (~50% of recipe time) was spent there, even with PRs #7454 and #7455 in place.

This PR replaces the per-source JAR download with a **`DirtyJavaSourceSets` registry**: an ExecutionContext-scoped set of project names that have had their dependency graph mutated during the run. Ambiguity-sensitive recipes consult the registry and take the safe path when their project is marked dirty.

## Design

- **Registry** (`rewrite-java/.../marker/DirtyJavaSourceSets`): a `Set<String>` of project names, stored on the `ExecutionContext` under a single key. `mark(ctx, JavaProject|String)` adds; `isDirty(jp, ctx)` reads. A conservative null-project fallback returns `true` whenever the registry is non-empty, so source files without a `JavaProject` marker still take the safe path when *any* mutation has happened in the run.
- **Why name-keyed, not `JavaProject`-keyed**: production parsers (Maven CLI, Gradle plugin, Moderne CLI) populate `JavaProject.Publication` with groupId/artifactId/version; producer-side synthesized markers cannot reconstruct that identically, so `Set<JavaProject>` would silently miss in production. Name-level granularity is sufficient for the ambiguity-detection use case and conservatively over-marks only on the rare case of name collisions across unrelated builds.
- **Key prefix** is `org.openrewrite.maven.dirtyJavaProjects` so the registry passes `CursorValidatingExecutionContextView`'s allowlist (which permits cross-recipe ExecutionContext mutation only for keys under `org.openrewrite.maven.*` or `io.moderne.*`). The registry is not Maven-specific; the prefix is purely a permission-list concession, documented inline. Renaming to a properly-scoped key is a follow-up that depends on the allowlist being broadened.

## Producer wiring

- **Maven**: centralized in `UpdateMavenModel.visitDocument` — fires once per real mutation (gated on the recipe having called `maybeUpdateModel()`). Also walks `MavenResolutionResult.getModules()` recursively, so a parent-pom mutation **propagates to every transitively aggregated child module's project**.
- **Gradle** (`AddDependency`, `ChangeDependency`, `ChangeDependencyConfiguration`, `ChangeManagedDependency`, `RemoveDependency`, `UpgradeDependencyVersion`): per-recipe `DirtyGradleSourceSets.markIfChanged(ctx, before, after)` calls at each visitor's \"build file actually changed\" gate. The helper resolves the project name from either a `JavaProject` marker (if attached) or the `GradleProject` marker that the Gradle tooling model attaches to `build.gradle` and `build.gradle.kts`.

## Consumer wiring

- `ImportLayoutStyle.isPackageFoldable` returns `false` when the project is dirty → `OrderImports` never folds a group of imports into a star it cannot prove is unambiguous. `AddImport` passes the same flag through.
- `ChangePackage.hasAmbiguity` returns `true` when dirty → existing star imports are expanded before package rewriting.
- `ChangeType.maybeAddExplicitImportForAmbiguity` adds the explicit import unconditionally when dirty.

## Safety semantics

Two simple rules replace the old classpath inspection:

1. **Dirty → expand stars.** If a classpath is potentially stale, we cannot prove a star is unambiguous, so we assume it may be and expand.
2. **Dirty → never fold into a star.** Likewise, we cannot prove folding is safe, so we keep imports explicit.

Both decisions are strictly conservative relative to the old classpath-based check — a dirty signal never produces a result the old check would have forbidden.

## Known gaps

These are deliberately out of scope for this PR and documented for follow-up:

- **Gradle `gradle.properties` and `gradle/libs.versions.toml`**: the Gradle tooling model does not attach `JavaProject` markers to these files, and `GradleSettings` does not enumerate subprojects. A mutation to a shared properties file or version catalog cannot fan out marks correctly without either parser-side enrichment or a `ScanningRecipe` accumulator that collects every project name in the scan pass. For now, the conservative null-project fallback in the consumer still provides degraded safety when any sibling marker exists.
- **Maven `dependencyManagement` upgrades for unused entries**: upgrading a managed version of an artifact the project doesn't actually consume still fires `maybeUpdateModel()` and marks the project dirty. Conservatively over-marks; acceptable cost for safety.
- **Plugin block / settings.gradle / gradle.lockfile**: not currently in the producer set.
- **Allowlist key prefix**: the `org.openrewrite.maven.*` prefix is a permission-list workaround. Cleanest follow-up is to either broaden `CursorValidatingExecutionContextView`'s allowlist for a properly-scoped key, or migrate the cross-recipe channel to a `DataTable`-style sanctioned mechanism.

## Test plan

- [x] `rewrite-java:test` — passes (incl. new `DirtyJavaSourceSetsTest`)
- [x] `rewrite-java-test:test` — passes (new coverage: `OrderImportsTest.doNotFoldIntoStarWhenSourceSetIsDirty`, `ChangePackageTest.changePackageExpandsStarImportWhenSourceSetIsDirty`, `ChangeTypeTest.changeTypeAddsExplicitImportWhenSourceSetIsDirty`, all wrapping sources via `mavenProject(name, ...)` so the precise registry lookup is exercised)
- [x] `rewrite-maven:test` — passes for affected recipes (`DirtyJavaSourceSetsProducerTest`, `RemoveDependencyTest`, `ChangeDependencyGroupIdAndArtifactIdTest.marksProjectDirtyAfterChangingDependencyCoordinates`, `ChangeDependencyGroupIdAndArtifactIdTest.composedWithChangePackageUpdatesImports`). Remaining suite failures are HTTP 429 from Maven Central rate limiting, unrelated to this change.
- [x] `rewrite-gradle:compileJava` and producer-touching recipes — passes
- [x] `rewrite-kotlin:test` — passes


sambsnyd (approved):

steve-aom-elliott:
@sambsnyd Just wanted to check whether we still feel alright about this going in.

steve-aom-elliott:
@sambsnyd I've reworked this again based on registering information about dirty projects in the ExecutionContext instead of trying to do it through markers, which would have needed to propagate changes to many more files each time. Was wondering if you might be able to take a look again when you have an opportunity.

rewrite-java/src/main/java/org/openrewrite/java/marker/DirtyJavaSourceSets.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>
+ * The registry keys on {@link JavaProject#getProjectName()} rather than the marker itself
+ * because {@link JavaProject#equals} also compares the optional {@link JavaProject.Publication},
+ * which producer call sites typically cannot reconstruct identically to what the parser attached.
+ * Name-level granularity is sufficient for the ambiguity-detection use case and conservatively
+ * over-marks in the rare case of name collisions across unrelated builds.
+ * <p>
+ * State is held only on the {@link ExecutionContext} and disappears when the recipe run ends.
+ */
+public final class DirtyJavaSourceSets {
```
sambsnyd: Do we need a class for for this? Feels like `JavaSourceSet` itself could have an `isDirty(ExecutionContext)` method

steve-aom-elliott: I think we still need more than just the `ExecutionContext` in order to not just indicate that every source set is dirtied, but I'm working on sketching out what moving things over might look like.

steve-aom-elliott:
@sambsnyd Okay, for this latest pass, I ended up shifting all the methods into the `JavaSourceSet` instead. The dirty marking still can take a `SourceFile` (from which we can obtain the project name via markers) or a `String` (the project name directly). One key change was the introduction of an intermediate interface extending `Marker` called `ProjectIdentity` that describes a `getProjectName()` method to be implemented implicitly by `JavaProject` via Lombok and explicitly as a pass-through to an alternatively named field in `GradleProject`. Doing this allows for both the `GradleDirtySourceSets` and `JavaDirtySourceSets` to be collapsed into `JavaSourceSet` in the same way and reduces a lot of the boilerplate wrappers.

Hopefully this makes a bit more sense / is a bit cleaner, but definitely let me know if you have further thoughts on it.

PR #7472, opened 2026-04-24
Scanning recipe for upgrading parent blows up

BoykoAlex:
An example of a scanning recipe for upgrading the parent version that just blows up. Unclear how to deal with this. If an an attempt to update the jackson BOM version before the parent version is updated doesn't help. In fact the scanning part of the `UpdateParentVersion` blows up even if this is the only recipe being executed. Doesn't seem like changing the parent and resolving everything in the scanning phase is a good idea...
@timtebeek @kdvolder


kdvolder:
Just checking in on this PR and the underlying scanning phase limitation. We are currently blocked on migrating some customer projects because we cannot rename properties (like `jackson-bom.version`) in the same cycle as upgrading the parent POM. Do you have any thoughts on how we might bypass this scanning phase trap or handle these interconnected upgrades?

PR #7477, opened 2026-04-27
Anchor nestmates to their NestHost's defining loader

bmuschko:
**This is a fix with a potentially big blast radius.**

## Summary

Fix a latent classloading bug in `RecipeClassLoader` where recipes shaped as `Recipe`/`ScanningRecipe` + anonymous `TreeVisitor` returned from `getVisitor()` + private nested visitor instantiated from inside the anonymous class can fail at runtime with `IllegalAccessError` when nest members end up on different defining loaders.

The fix peeks the `NestHost` attribute via ASM before delegation runs and routes every nest member to whichever loader defined the host. When the host is on this loader, no parent fallback is attempted — a missing member surfaces as `ClassNotFoundException` rather than silently splitting the nest.

## The concrete failure

`org.openrewrite.java.ChangePackage` is the recipe that triggered this:

```
IllegalAccessError: failed to access class org.openrewrite.java.ChangePackage$JavaChangePackageVisitor
  from class org.openrewrite.java.ChangePackage$2
  (...$JavaChangePackageVisitor is in unnamed module of loader java.net.URLClassLoader @3cd1a2f1;
   ...$2          is in unnamed module of loader io.moderne.cli.recipe.CliRecipeClassLoader @731f4b3e)
```

`ChangePackage$2` is the anonymous `TreeVisitor` returned by `getVisitor()`; it instantiates `new JavaChangePackageVisitor()` (`ChangePackage.java:141`). `JavaChangePackageVisitor` is `private` (`ChangePackage.java:158`). When the two end up on different defining loaders the JVM denies the access.

## How we usually patch this in `moderne-cli`

The tactical fix downstream is to add the offending class to [`CliRecipeClassLoader.getAdditionalParentDelegatedPackages()`](https://github.com/moderneinc/moderne-cli/blob/main/core/recipes/src/main/java/io/moderne/cli/recipe/CliRecipeClassLoader.java#L15-L35) (in `moderneinc/moderne-cli`) so the same loader resolves all of the nestmates. It works, but it is reactive: every new affected recipe needs another entry, and third-party recipes with the same shape stay unprotected until someone hits the failure in production.

## Why centralize the fix

This isn't a one-off. ~35 classes in `openrewrite/rewrite` share the exact `Recipe` + anonymous visitor + private nested visitor shape:

| Module | Count | Examples |
|---|---|---|
| `rewrite-java` | 15 | `AddImport`, `AddLiteralMethodArgument`, `AddMethodParameter`, `AddNullMethodArgument`, `ChangeMethodTargetToStatic`, `ChangeMethodTargetToVariable`, `ChangePackage`, `ChangeType`, `DeleteMethodArgument`, `RemoveUnusedImports`, `ReorderMethodArguments`, `ReplaceConstantWithAnotherConstant`, `UseStaticImport`, `BlockStatementTemplateGenerator`, `FindTypes` |
| `rewrite-gradle` | 11 | `MigrateDependenciesToVersionCatalog`, `UpdateGradleWrapper`, `UpgradeDependencyVersion`, `UpgradeTransitiveDependencyVersion` |
| `rewrite-maven` | 7 | `ChangeDependencyGroupIdAndArtifactId`, `ManageDependencies`, `RemoveRedundantDependencyVersions`, `RemoveUnusedProperties`, `UpdateMavenWrapper`, `UpgradeDependencyVersion` |
| `rewrite-yaml` | 2 | `UnfoldProperties` |
| `rewrite-docker` | 1 | `AddOrUpdateLabel` |

Plus the broader recipe ecosystem (`rewrite-spring`, `rewrite-static-analysis`, `rewrite-migrate-java`, …) and any third-party recipe written in this idiomatic shape. None are protected today; the bug is latent across all of them.

Putting the fix in `RecipeClassLoader` itself means every consumer (`moderne-cli`, SaaS, the IntelliJ plugin, anything extending `RecipeClassLoader`) is structurally protected from this entire class of issue, including recipes that don't exist yet.

## What changed

- `rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeClassLoader.java` — overrides `loadClass(name, resolve)` to peek the `NestHost` attribute via ASM (already a `rewrite-core` dependency) before the existing delegation logic runs. Three branches:
  - No `NestHost` attribute, or the class is its own host → fall through to existing logic. No behavior change.
  - `NestHost` declares another class → load the host first; the loader that defines the host owns the entire nest. Members route to that loader.
  - When the host is on this loader, member loading uses `findClass(name)` directly with no parent fallback, so a missing member fails loudly as `ClassNotFoundException` instead of silently splitting the nest.
- `rewrite-core/src/test/java/org/openrewrite/marketplace/RecipeClassLoaderTest.java` — adds three tests that build a synthetic nest (`nest.Outer` / `Outer$1` / private `Outer$Inner`) at runtime via `JavaCompiler` and exercise: host-on-child happy path, host-on-parent reverse direction, and the regression case (child missing the inner) failing loudly.

## Follow-up after release

Several entries in `RecipeClassLoader.PARENT_DELEGATED_PREFIXES` and in the downstream `CliRecipeClassLoader.getAdditionalParentDelegatedPackages` list (e.g. `Cobol*Visitor`, `DevCenter*`) appear to be ad-hoc workarounds for this same nest-split issue. Worth auditing each one once the structural fix lands — entries that exist purely as nestmate workarounds can be retired.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeClassLoader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeClassLoader.java
@@ -16,8 +16,12 @@
 package org.openrewrite.marketplace;
 
 import org.jspecify.annotations.Nullable;
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.Opcodes;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -103,6 +107,14 @@ public class RecipeClassLoader extends URLClassLoader {
                 return foundClass;
             }
 
+            foundClass = loadNestMember(name);
+            if (foundClass != null) {
+                if (resolve) {
+                    resolveClass(foundClass);
+                }
+                return foundClass;
+            }
+
             // Determine delegation strategy
             try {
                 if (shouldDelegateToParent(name)) {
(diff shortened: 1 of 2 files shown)


sambsnyd (commented):
I'm not wholly sure what the consequences of this are. I'm not familiar with the concept or implications of classes coming from the same "nest". I would be more comfortable if I knew this had been tested with CLI built with this change and recipe runs / data tables / etc. can be installed and run successfully.

bmuschko:
Thanks for your input, @sambsnyd. I am going to close this PR and let someone else from the product team take this one over. I believe we will want to provide a general purpose solution to the problem instead of managing a list of classes that should be handled by a different classloader. As explained in the description of the PR, this is not an isolated problem to a handful of recipes but a somewhat more widespread issue.

PR #7478, opened 2026-04-27
UpgradeDependencyVersion: warn on install failure instead of throwing

Jenson3210:
## Summary

When the package-manager install run by `UpgradeDependencyVersion` fails (typically an `ERESOLVE` peer-dep conflict), the recipe currently `throw`s, aborting the entire recipe chain. That prevents *later* recipes — which might be the very ones that resolve the conflict, e.g. by bumping a peer-pinning library — from running at all.

The sibling `AddDependency` recipe already handles install failure by returning `markupWarn(doc, ...)` and letting the chain continue. This PR aligns the two so partial failures degrade gracefully.

## Motivation

Real-world trigger: an Angular 17 project with `@ng-bootstrap/ng-bootstrap@^16.0.0` (strict peer `@angular/core@^17.0.0`) being upgraded via a chain of `UpgradeDependencyVersion` calls (one for `@angular/*`, one for `@ng-bootstrap/ng-bootstrap`). Today the first failing install kills the run, and the dep graph stays in its initial inconsistent state forever. With this change, the chain continues, the marker records the failure, and a subsequent recipe can converge the graph (or a second pass of the same chain reaches fixed-point).

## Diff

```diff
- import {replaceMarkerByKind} from "../../markers";
+ import {markupWarn, replaceMarkerByKind} from "../../markers";
…
  if (failureMessage) {
      const names = updateInfo.matchedDependencies.map(d => d.packageName).join(', ');
-     throw new Error(`Failed to upgrade ${names} to ${recipe.newVersion}: ${failureMessage}`);
+     return markupWarn(doc, `Failed to upgrade ${names} to ${recipe.newVersion}`, failureMessage);
  }
```

## Tests

The two existing tests that asserted on the throw — `throws when install fails for non-existent version` and `throws when install fails due to engine version mismatch` — are reworked to assert on the `MarkupWarn` marker, mirroring the pattern in `add-dependency.test.ts::adds warning marker when package does not exist`. Renamed to `adds warning marker when install fails …` to match the new behaviour.

All 25 tests in `upgrade-dependency-version.test.ts` pass locally.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npx vitest run test/javascript/recipes/upgrade-dependency-version.test.ts` — 25/25 pass
- [ ] Run full suite in CI
- [ ] Verify the unblocked-recipe-chain behaviour against a known-failing project (Angular 17 + ng-bootstrap@16 → 18 + ng-bootstrap@17)

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/recipes/upgrade-dependency-version.ts
+++ b/rewrite-javascript/rewrite/src/javascript/recipes/upgrade-dependency-version.ts
@@ -31,7 +31,7 @@ import {
 import * as path from "path";
 import * as semver from "semver";
 import * as picomatch from "picomatch";
-import {replaceMarkerByKind} from "../../markers";
+import {markupWarn, replaceMarkerByKind} from "../../markers";
 import {TreePrinters} from "../../print";
 import {
     createDependencyRecipeAccumulator,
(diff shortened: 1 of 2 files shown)


Jenson3210:
Reopening from origin branch instead of fork (project convention).

PR #7480, opened 2026-04-27
Throw MavenParsingException when a POM is missing groupId or version

steve-aom-elliott:
## What

`RawPom.toPom` calls `Objects.requireNonNull` on the project's `groupId` and `version` ([RawPom.java:437,439](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java#L437-L439)). When either is absent, the resulting `NullPointerException` propagates out of `mod build` / `MavenParser.parseInputs` with no artifact context, only this stack:

```
Caused by: java.lang.NullPointerException
    at java.util.Objects.requireNonNull (Objects.java:203)
    at org.openrewrite.maven.internal.RawPom.toPom (RawPom.java:437)
    at org.openrewrite.maven.internal.MavenPomDownloader.download (MavenPomDownloader.java:595)
    at org.openrewrite.maven.tree.ResolvedPom.resolveDependencies (ResolvedPom.java:1050)
    ...
```

There is no signal as to which transitive dependency's POM was malformed.

## Why

A POM can legitimately omit a `<groupId>` or `<version>` element (relying on `<parent>` inheritance), but it can also be invalid — for example:

- a stub POM uploaded to an internal Artifactory/Nexus alongside a vendored JAR via `mvn install:install-file` or REST upload (which bypasses `ModelValidator`)
- a POM emitted by non-Maven tooling that wrote `<groupId/>` or `<groupId></groupId>` instead of omitting the element (Jackson-XML's `EMPTY_ELEMENT_AS_NULL` deserializes those to `null`)
- a POM with whitespace-only content like `<version>  </version>` (which `MavenXmlMapper`'s `StringTrimModule` reduces to `""`, then sailed past every downstream null-check)

In each of these cases the resolver cannot construct a usable `Pom`, but the user has no way of telling which artifact is the source.

## How

Two contained changes inside `RawPom`:

1. `getGroupId()` and `getVersion()` now treat blank (null or whitespace-only) values as missing, so an empty `<groupId/>` or `<version>  </version>` is treated the same as an omitted element. Scope is limited to these two methods; nothing outside `RawPom` is affected.
2. `toPom()` checks the resolved coordinates and throws `MavenParsingException` (already used at sibling validation points in `ResolvedPom`) naming the artifact and source repository:

   ```
   POM is missing a required coordinate: groupId for {null:my-app:1} (from https://repo.example.com/...)
   ```

`MavenParsingException` is unchecked, so no signature changes ripple out to `MavenPomDownloader`, `MavenParser`, `Assertions`, or tests.

## Tests

Added five tests to `RawPomTest`:
- `missingGroupIdThrowsParsingException`
- `missingVersionThrowsParsingException`
- `emptyGroupIdElementThrowsParsingException`
- `whitespaceVersionElementThrowsParsingException`
- `groupIdInheritedFromParentStillResolves` (regression check that legitimate parent inheritance still works)

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/RawPom.java
@@ -36,7 +36,6 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to capture these with more details, thanks!

PR #7483, opened 2026-04-28
Build fast "has type" lookup into TypesInUse cache, speeding up repeated type queries

sambsnyd:
LSTs already have a TypesInUse field which is a WeakReference'd cache of type information present in a given LST. But UsesType and UsesMethod (another very popular precondition / implementation detail) still iterate over every single type in use looking for matches. So in a large composite recipe:
thousands of recipe instance are, via UsesType / UsesMethod, iterating over lists of hundreds or thousands of types, for hundreds or thousands of source files. It explains why AddDependency dominates the scanning phase in terms of time spent in the spring boot migration, followed closely by ChangeType.

So build a lazily produced prefix tree into TypesInUse. Big savings in the asociated benchmark. In the benchmark which simulates running 100 HasType in succession this is 100x faster since the trie only gets built once.


rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -42,6 +49,44 @@ public class TypesInUse {
...
+     *       check these.</li>
+     *   <li>{@code alias*} — the canonical form of an inner-class FQN ({@code $} replaced with
+     *       {@code .}) ends here. Lets {@link #hasType} match either form against
+     *       {@link TypeUtils#fullyQualifiedNamesAreEqual} semantics without confusing package-shape
+     *       queries.</li>
+     * </ul>
+     * Visibility splits each pair into explicit (reachable from types-in-use or imports — visible
+     * without {@code includeImplicit}) and implicit (reachable only through used-method types).
+     */
+    @Nullable
```
Jenson3210, line 68: ⚠️ This is an internal class being changed which is parent loaded

sambsnyd: Indeed it is. Good callout that this change requires redeploy.

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -42,6 +49,44 @@ public class TypesInUse {
...
+     * Per-{@link MethodMatcher} memo for {@link #hasMethod}. Lazily allocated on first use, keyed
+     * by {@code matchOverrides ⊕ matcher.toString()} so equivalent matchers from different
+     * recipe instances share cache entries.
+     */
+    @Nullable
+    private volatile Map<String, Boolean> usedMethodCache;
+
+    /** Same shape as {@link #usedMethodCache} but over {@link #declaredMethods}. */
+    @Nullable
+    private volatile Map<String, Boolean> declaredMethodCache;
```
Jenson3210: will we interrogate the same TypesInUse instance for the same methodPattern usages/declarations? 
Not sure if we have a lot of invocations with **exactly the same** method pattern in our codebase

rewrite-benchmarks/src/jmh/java/org/openrewrite/benchmarks/java/UsesTypeBenchmark.java
```diff
@@ -0,0 +1,270 @@
...
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+public class UsesTypeBenchmark {
+
+    /**
+     * 100 FQNs to query: a mix of hits (reachable through types-in-use, imports, or supertype
+     * walks) and misses (types not used by the sample file). The proportions are roughly
+     * representative of recipe preconditions in a mixed migration composite — most queries miss.
+     */
+    static final List<String> QUERIES = Arrays.asList(
```
Jenson3210, line 58: Should we also benchmark different wildcard patterns used throughout our recipes? 
```
*..*
some.packagename..*
some.packagename..*Identifier*
some.*.packagename.Identifier
some..packagename.Identifier
...
```

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -42,6 +49,44 @@ public class TypesInUse {
...
+     *   <li>{@code alias*} — the canonical form of an inner-class FQN ({@code $} replaced with
+     *       {@code .}) ends here. Lets {@link #hasType} match either form against
+     *       {@link TypeUtils#fullyQualifiedNamesAreEqual} semantics without confusing package-shape
+     *       queries.</li>
+     * </ul>
+     * Visibility splits each pair into explicit (reachable from types-in-use or imports — visible
+     * without {@code includeImplicit}) and implicit (reachable only through used-method types).
+     */
+    @Nullable
+    private volatile FqnTrie trie;
```
Jenson3210, line 69: Would we need to add tests here for all the patterns? 
```
some.packagename.Identifier
*..*
some.packagename..*
some.packagename..*Identifier*
some.*.packagename.Identifier
some..packagename.Identifier
...
```

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
sambsnyd: Good call

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
sambsnyd: Done

rewrite-benchmarks/src/jmh/java/org/openrewrite/benchmarks/java/UsesTypeBenchmark.java
sambsnyd: Nothing runs the benchmarks on the regular, so after the initial validation they're more showing-your-work/documentation than anything. Not strictly necessary to check in.

rewrite-java/src/main/java/org/openrewrite/java/search/DeclaresMethod.java
```diff
@@ -21,7 +21,6 @@
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaSourceFile;
-import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.marker.SearchResult;
 
 public class DeclaresMethod<P> extends JavaIsoVisitor<P> {
```
Jenson3210, line 28: As these are not parent loaded, I think we should wrap the codechanges in `try/catch NoSuchMethod` blocks keeping the old code alive also for the time being?

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -52,6 +85,391 @@ public static TypesInUse build(JavaSourceFile cu) {
...
+            String[] childSegments = EMPTY_SEGMENTS;
+            Node[] childNodes = EMPTY_NODES;
+            int childCount;
+
+            boolean leafExplicit;
+            boolean leafImplicit;
+            boolean aliasExplicit;
+            boolean aliasImplicit;
+            boolean descendantExplicit;
+            boolean descendantImplicit;
```
knutwannheden: We could pack these 6 flags into a single `byte` field. When running with Java 25 and `-XX:+UseCompactObjectHeaders` this would with the padding result in 24-byte objects rather than 32-byte objects.

rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
```diff
@@ -52,6 +85,391 @@ public static TypesInUse build(JavaSourceFile cu) {
...
+                        if (explicit) {
+                            n.descendantExplicit = true;
+                        } else {
+                            n.descendantImplicit = true;
+                        }
+                    }
+                    n = n.findOrAddChild(path, start, i);
+                    start = i + 1;
+                }
+            }
```
knutwannheden: By leveraging the JVM intrinsic and SIMD-friendly `indexOf()` we can gain another 15%. Here is the patch for that:

```patch
diff --git a/rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java b/rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
index 723084f0a4..6b085292a7 100644
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java
@@ -296,20 +296,23 @@ private void insertPath(String path, boolean explicit, boolean alias) {
             Node n = root;
             int start = 0;
             int len = path.length();
-            for (int i = 0; i <= len; i++) {
-                if (i == len || path.charAt(i) == '.') {
-                    // Descendant rollup propagates only for raw-leaf insertions; alias paths do
-                    // not contribute to "this package has any class beneath it" queries.
-                    if (!alias) {
-                        if (explicit) {
-                            n.descendantExplicit = true;
-                        } else {
-                            n.descendantImplicit = true;
-                        }
+            while (true) {
+                int dot = start < len ? path.indexOf('.', start) : -1;
+                int end = dot < 0 ? len : dot;
+                // Descendant rollup propagates only for raw-leaf insertions; alias paths do
+                // not contribute to "this package has any class beneath it" queries.
+                if (!alias) {
+                    if (explicit) {
+                        n.descendantExplicit = true;
+                    } else {
+                        n.descendantImplicit = true;
                     }
-                    n = n.findOrAddChild(path, start, i);
-                    start = i + 1;
                 }
+                n = n.findOrAddChild(path, start, end);
+                if (dot < 0) {
+                    break;
+                }
+                start = end + 1;
             }
             if (alias) {
                 if (explicit) {
@@ -401,14 +404,14 @@ private static boolean anyLeafFqn(Node node, StringBuilder fqn, TypeNameMatcher
             Node n = root;
             int start = 0;
             int len = pkg.length();
-            for (int i = 0; i <= len; i++) {
-                if (i == len || pkg.charAt(i) == '.') {
-                    n = n.findChild(pkg, start, i);
-                    if (n == null) {
-                        return null;
-                    }
-                    start = i + 1;
+            while (start < len) {
+                int dot = pkg.indexOf('.', start);
+                int end = dot < 0 ? len : dot;
+                n = n.findChild(pkg, start, end);
+                if (n == null) {
                 }
+                start = end + 1;
             }
             return n;
         }
```

sambsnyd: Excellent thank you for the tip

PR #7486, opened 2026-04-28
FindStyles: honor custom @JsonSerialize so ImportLayoutStyle.layout isn't dropped

Jenson3210:
## Summary

- `FindStyles` was dumping each `Style` via SnakeYAML with `BeanAccess.FIELD`, which bypassed any class-level `@JsonSerialize`. For `ImportLayoutStyle` that meant the `Block` instances (sealed interface, `private final` fields, no setters) could not be introspected and the recipe reported empty `layout: []` / `packagesToFold: []` even when the merged style was fully populated — exactly the field people most often want to verify.
- Each `Style` is now routed through `ObjectMappers.propertyBasedMapper(...).convertValue(style, Map.class)` first, so any custom serializer (e.g. `ImportLayoutStyle.Serializer`, which renders Blocks as readable strings like `import java.*`, `<blank line>`, `import all other imports`) is invoked before SnakeYAML sees the value. The `BeanAccess.FIELD` setup and the class-tag-suppressing `Representer` are no longer needed and have been removed.
- Added an `importLayoutStyleSerializesBlocksAsStrings` regression test that builds an `ImportLayoutStyle` via `Builder` with a real layout + `packagesToFold` and asserts the output contains the readable Block strings and not `layout: []` / `packagesToFold: []`. The previous tests only exercised trivial POJO styles and would never have caught this.

### Before

```yaml
- org.openrewrite.java.style.ImportLayoutStyle:
    classCountToUseStarImport: 9999
    layout: [
      ]
    nameCountToUseStarImport: 9999
    packagesToFold: [
      ]
```

### After

```yaml
- org.openrewrite.java.style.ImportLayoutStyle:
    classCountToUseStarImport: 9999
    nameCountToUseStarImport: 9999
    layout:
    - import java.*
    - <blank line>
    - import all other imports
    - <blank line>
    - import static all other imports
    packagesToFold:
    - import java.util.*
```

The fix is general: any `Style` with a custom `@JsonSerialize` benefits, not just `ImportLayoutStyle`.

## Test plan

- [x] `./gradlew :rewrite-test:test --tests "org.openrewrite.FindStylesTest"` (incl. new regression test)
- [x] `./gradlew :rewrite-core:test :rewrite-test:test`
- [x] `./gradlew :rewrite-java:test --tests "*ImportLayoutStyleTest*"` to confirm `Serializer`/`Deserializer` round-trip is untouched
- [x] `./gradlew :rewrite-core:licenseFormat :rewrite-test:licenseFormat`

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/FindStyles.java
+++ b/rewrite-core/src/main/java/org/openrewrite/FindStyles.java
@@ -15,20 +15,18 @@
  */
 package org.openrewrite;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.internal.ObjectMappers;
 import org.openrewrite.marker.SearchResult;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.style.Style;
 import org.openrewrite.table.StylesInUse;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
-import org.yaml.snakeyaml.introspector.BeanAccess;
-import org.yaml.snakeyaml.introspector.Property;
-import org.yaml.snakeyaml.nodes.MappingNode;
-import org.yaml.snakeyaml.nodes.Tag;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

Jenson3210:
@steve-aom-elliott @MBoegers tagged you as you both have people interested in styles

PR #7510, opened 2026-04-28
Restore JavaScript/Python/C# entries for Java recipes in recipes.csv

timtebeek:
## Summary

- Re-adds the 21 cross-ecosystem rows in `rewrite-java/src/main/resources/META-INF/rewrite/recipes.csv` that were inadvertently dropped by #6991 when `MavenRecipeMarketplaceGenerator` regenerated the file. The generator does a clean classpath scan and overwrites the file, so the manually-added rows from #6485 (JavaScript), #6593 (Python), and #6678 (C#) were silently removed.
- Restores `ChangeMethodName`, `ChangePackage`, `ChangeType`, `DeleteMethodArgument`, `ReorderMethodArguments`, `search.FindMethods`, and `search.FindTypes` for each of those three ecosystems. The new rows are derived from today's Java rows so the option and data-table metadata is up to date (including the `instanceName` field added in 66497c6).
- Note: the same regression will recur next time someone runs `:rewrite-maven:generateRecipeMarketplace` and copies `build/recipes.csv` over the source file. A follow-up to make the regenerator merge cross-ecosystem rows from the existing CSV (or move them to a separate file) would prevent this.

## Test plan

- [ ] CI green
- [ ] Spot-check that `MavenRecipeBundleReader.read()` (the runtime path that reads `recipes.csv` straight out of the published JAR) sees the JavaScript/Python/C# categories for these 7 recipes again


timtebeek:
image

mike-solomon (approved):
Thank you!

PR #7511, opened 2026-04-28
RecipeMarketplace.Category.merge: copy unmatched source children

natedanner:
When `Category.merge` finds no name match for a source child, it adopts that child by reference (`categories.add(subCategory)`). Later merges that name-match into the adopted child then mutate the original source's `recipes` and `categories`. Most upstream usage hides this — `RecipeMarketplace.install(BundleReader)` uses a throwaway source — but consumers that cache and reuse a `RecipeMarketplace` as a merge source see concurrent `ConcurrentModificationException` / `ArrayIndexOutOfBoundsException` from the shared `ArrayList`s, and silent state drift across sequential merges.

This PR allocates a fresh `Category` on the receiver and recurses to copy contents instead. The recipes loop was already safe — `recipe.withMarketplace(this)` returns a new instance.

`RecipeMarketplaceMergeTest` adds two tests pinning the invariant; both fail on parent, pass with the fix. The existing merge tests in `RecipeMarketplaceReaderTest` only exercise the name-match path and pass unchanged.

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/marketplace/RecipeMarketplaceMergeTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/marketplace/RecipeMarketplaceMergeTest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


natedanner:
Superseded by #7597, which landed the same fix on 2026-05-07. Closing.

PR #7527, opened 2026-04-30
Fix AddAnnotationProcessor when parent matches by GAV alone

MBoegers:
## Summary

- Fix orphan modules being silently skipped when their external parent's GAV happens to also exist as a separately ingested LST artifact (`MavenResolutionResult.parentPomIsProjectPom()` is GAV-only).
- Verify each parent→child link by walking aggregator `<modules>` chains; treat the child as an orphan when no aggregator actually links them.
- Fill in missing `<configuration>` / `<annotationProcessorPaths>` when the matched plugin is pre-declared without it (previously `AddPluginVisitor` left such plugins untouched).

## Problem

`AddAnnotationProcessor` produced zero changes on Maven modules that:

1. Have no in-reactor parent in their actual build, but coincidentally co-exist in the same LST as a POM that shares their parent's GAV (e.g. an externally-ingested corporate parent), **or**
2. Pre-declare `maven-compiler-plugin` without `annotationProcessorPaths` (e.g. only to set `<source>`/`<target>` or for inherited config).

In case 1, `parentPomIsProjectPom()` returns true purely because GAVs match, so the child is routed to the in-reactor-parent branch and never added to `candidateOrphanPaths`. In the visitor it is neither parent nor orphan and is skipped, so its own `build/plugins` is never updated.

In case 2, `AddPluginVisitor`'s "plugin already exists" branch leaves the plugin untouched, and the path-adding inner visitor only operates on existing `<annotationProcessorPaths>` tags — so nothing is ever added.

## Solution

1. **Reactor-aware parent classification.** During scan, record `tentativeChildToParent` links and aggregator `<modules>` membership (read off the requested Pom, since `ResolvedPom.subprojects` is unreliable in some LST fixtures and `mrr.getModules()` lists POMs claiming this one as `<parent>` rather than ones aggregated by it). Before the visitor runs, walk aggregator `<modules>` chains via BFS; uphold a tentative link only when the child is reactor-reachable. Otherwise reclassify as orphan and add the processor to its own `build/plugins`. Skip dangling `packaging=pom` POMs that are neither claimed parents nor aggregators.

2. **Bypass for GAV-coincident orphans.** Pass the source path as `AddPluginVisitor`'s `filePattern` only for the GAV-coincident orphan case so its own `parentPomIsProjectPom()` short-circuit doesn't refuse to add the plugin. Genuine in-reactor parents and true single-module orphans hit the existing acceptance path with `filePattern=null`.

3. **Fill in missing `<configuration>` structure.** Before the path-adding visitor runs, ensure the matched plugin has `<configuration><annotationProcessorPaths/></configuration>`. The helper is conservative: only fills in whichever level is missing.

## Test plan

- [x] New `orphanModuleSharedLstWithExternalParent` reproducer fails on `main` and passes after the fix.
- [x] New `addConfigurationToExistingPluginWithNoConfig` and `addAnnotationProcessorPathsToExistingConfiguration` reproducers fail on `main` and pass after the fix.
- [x] All 30 existing `AddAnnotationProcessorTest` cases continue to pass.
- [x] Full `:rewrite-maven:test` passes.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
@@ -33,8 +33,13 @@ import org.openrewrite.xml.XmlIsoVisitor;
 import org.openrewrite.xml.tree.Xml;
 
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/AddAnnotationProcessor.java
```diff
@@ -197,21 +257,23 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
...
                     return tree;
                 }
 
-                // First, ensure the plugin exists - use the source path as file pattern
+                // GAV-coincident orphan: AddPluginVisitor.isAcceptable would
+                // otherwise short-circuit via its parentPomIsProjectPom()
+                // check and refuse to add the plugin. Targeting the visitor
+                // at this exact source path bypasses that guard.
+                boolean isGavCoincidentOrphan = !isParent && mrr.parentPomIsProjectPom();
+                String pluginFilePattern = isGavCoincidentOrphan ? sourcePath.toString() : null;
```
steve-aom-elliott: `sourcePath.toString()` will use the host OS's separator. The codebase convention for stringifying a `sourcePath` for downstream pattern/path comparisons is `PathUtils.separatorsToUnix(sourcePath.toString())` — see e.g. `AddDevelocityMavenExtension.java:167`, `EffectiveMavenRepositories.java:75/80/95`, `AddRuntimeConfig.java:116`.

In practice `PathUtils.matchesGlob` is separator-tolerant (its `tokenize` accepts both `/` and `\`), so this isn't a correctness bug today, but applying the convention here keeps it from drifting.

Probably moot if the `filePattern` overload is replaced by an explicit `boolean force` (or similarly-named) constructor flag on `AddPluginVisitor` — see the separate top-level comment about promoting this into a more truthful `MavenResolutionResult` API.

steve-aom-elliott:
A couple of edge cases that aren't regressions versus `main` but are worth either narrowing the scope claim or expanding coverage for:

**`ensureAnnotationProcessorPathsTag` only handles top-level `<configuration>`.**
Plugins where `annotationProcessorPaths` lives inside `<executions><execution><configuration>` (the per-execution form, common when a project binds the compiler plugin to a non-default phase or splits main/test compilation) won't be touched by the new helper. The existing `addConfigurationToExistingPluginWithNoConfig` / `addAnnotationProcessorPathsToExistingConfiguration` tests only cover the plugin-level case. Either widen the helper to also walk per-execution configs, or rename the tests to make the scope explicit ("...AtPluginLevel") so a future reader doesn't assume coverage that isn't there.

**Stale `<modules>` in externally-ingested aggregators could falsely link unrelated POMs.**
The new BFS seeds from every POM that declared `<modules>`, including a corp-parent that legitimately aggregates modules in its own (out-of-LST) reactor. Those module strings get resolved relative to the corp-parent's `sourcePath` *as it sits in the LST*. Almost always those resolved paths point nowhere — but if a co-ingested unrelated repo happens to have a matching directory layout, two genuinely independent POMs could end up reactor-linked by coincidence. Edge case, probably fine in practice, but worth a code comment near `aggregatorSubmodulePaths` acknowledging the assumption ("we trust that aggregator `<modules>` entries refer only to POMs from the same reactor; collisions across co-ingested repos are theoretically possible but rare in practice").

MBoegers:
Thanks for the review, Steve. Pushed 2a660fd57 addressing all three points:

1. **`PathUtils.separatorsToUnix` convention** — applied at the `pluginFilePattern` assignment.

2. **`ensureAnnotationProcessorPathsTag` per-execution coverage** — agreed the helper only handles the plugin-level `<configuration>`. Renamed both reproducers to `...AtPluginLevel` to scope-fence the claim.

3. **Stale `<modules>` BFS** — added a comment near `aggregatorSubmodulePaths` documenting the assumption. Triple-coincidence (corp-parent declares `<modules>`, paths resolve into the LST, and a co-ingested unrelated repo occupies the colliding path) makes it a non-issue in practice, but the assumption is now explicit for future readers.

steve-aom-elliott (approved):

PR #7528, opened 2026-04-30
Introduce `JavaTypeFactory` and extend `JavaSourceSet` with classpath fast paths

knutwannheden:
## Motivation

Parser-time type construction is independently re-implemented in every TypeMapping (Java 8/11/17/21/25, Groovy, Kotlin, KotlinIr, Scala, JavaReflection). Each one caches differently, uses a different idiom for breaking recursive resolution, and accesses `JavaTypeCache` directly. Recipes that want to influence parser type identity — most importantly, sharing types across a `JavaTemplate` splice and the surrounding LST — have no abstraction to hook into.

`JavaSourceSet` exposes a flat `List<JavaType.FullyQualified>` for the classpath. Recipes that look types up by FQN or by package have been doing linear scans, which doesn't scale on classpaths with thousands of types. There is no SPI for marker producers to plug in a sorted/indexed backing.

`TypeTable` writes per-artifact classes-directories. When those land on `JavaTemplate`'s parser classpath, javac uses `DirectoryContainer`, which calls `openat` per `list()`. On parsing-heavy recipe runs the syscall cost can dominate wall time.

## Summary

**`JavaTypeFactory`**

- New `org.openrewrite.java.internal.JavaTypeFactory` interface with a `compute*` family covering the seven parser-constructed `JavaType` kinds: `computeClass`, `computeMethod`, `computeVariable`, `computeArray`, `computeIntersection`, `computeGenericTypeVariable`, `computeParameterized`.
- Each `compute*` returns the cached instance keyed by signature, or allocates a stub, registers it in the cache before invoking the supplied initializer, then runs the initializer to populate the stub. Registering up front is what makes recursive resolution safe — a recursive lookup for the same signature finds the stub instead of looping.
- `DefaultJavaTypeFactory` implements `compute*` directly. Initializers populate the stub by calling `unsafeSet` on it.
- All TypeMappings migrated: `ReloadableJava{8,11,17,21,25}TypeMapping`, `GroovyTypeMapping`, `KotlinTypeMapping`, `KotlinIrTypeMapping`, `ScalaTypeMapping`, `JavaReflectionTypeMapping`.

**`JavaTypeFactory.Provider`**

- Parser builders (`JavaParser`, `Java*Parser`, `GroovyParser`, `KotlinParser`, `ScalaParser`, `GradleParser`) accept a `Provider`. The legacy `typeCache(JavaTypeCache)` builder is `@Deprecated` in favor of `typeFactory(JavaTypeFactory)`.
- `JavaTemplateParser` consumes the same `Provider` from the recipe-scoped root cursor (via `TYPE_FACTORY_PROVIDER_KEY`). `RecipeScheduler` exposes `rootCursorProvider(Supplier<Cursor>)` so a recipe-scoped `JavaTypeFactory` can flow into `JavaTemplate` parses, keeping splice-time type identity consistent with the surrounding LST.
- `RecipeClassLoader` parent-loads `JavaTypeFactory` so recipe-side and parser-side factory instances unify.

**`JavaSourceSet` classpath fast paths**

- New `JavaSourceSet.ClasspathIndex` SPI: marker producers can plug in a sorted/indexed backing for the classpath list.
- `findClasspathType(fqn)` — sorted-bsearch via `ClasspathIndex.findFullyQualified`, falls back to linear scan for legacy markers.
- `classpathTypesInPackage(pkg)` — sorted-prefix range scan via `ClasspathIndex.typesInPackage`, same fallback.
- `removeTypesMatching` / `removeTypesForGav` route through `ClasspathIndex.withGavsRemoved` when the marker backing supports it, so an indexed classpath view is preserved across recipe edits instead of being collapsed into a flat list.
- `ImportLayoutStyle` uses the per-package fast path.
- `JavaSourceSetCompat` retains the legacy materialization path under `@ToBeRemoved(after = "2026-06-30")` for recipes still calling `JavaSourceSet.build(...)` directly.

**`TypeTable` writes per-artifact JARs**

- Output is now one JAR per `(groupId, artifactId, version)`, written via temp file + atomic move.
- Javac uses `ArchiveContainer` for these (central directory cached once) instead of `DirectoryContainer`. On a parsing-heavy recipe run (AssertJ migration on `spring-data-commons`), worker `__open` samples drop 921 → 157 (-83%) on async-profiler wall mode; recipe wall time drops ~9%.
- `AnnotationDeserializer`, `TypeSignature`, and `TypeTableSink` are extracted as public types so writers can produce structured annotation data without string round-tripping.

## Test plan

- [x] `gw check` clean across all modules (53m, 162 tasks, 0 failures)
- [x] Existing TypeMapping tests pass under each Java/Groovy/Kotlin/Scala module
- [x] `JavaTemplateParserProviderTest` covers `Provider` plumbing through `JavaTemplate`
- [x] `TypeTableTest` updated to assert JAR entries instead of directory contents
- [x] Legacy `JavaSourceSet.build(...)` path covered by `JavaSourceSetCompat` and remaining recipe tests


sambsnyd:
Nice :+1:

PR #7531, opened 2026-04-30
Java: realign Javadoc cursor when javac strips continuation whitespace

timtebeek:
## Summary

- Java 23+ javac's `DCText`/`DCComment` `getBody()` strips per-line leading whitespace that older versions preserved. When an older parser (e.g. `rewrite-java-21`) runs on a Java 25 JVM — as `rewrite-static-analysis` does in CI — `visitText` falls one character behind on every continuation line, dropping the `*`-margin trailing space and emitting a duplicated trailing char from the leftover-source branch in `visitDocComment`.
- After consuming a `\n`, advance the cursor over any source whitespace that the body lacks at the corresponding position. Applied to `rewrite-java-{8,11,17,21}` visitors; the Java 25 visitor already does this inline.
- Adds three round-trip regression tests in `JavadocTest` (plain ASCII, non-ASCII, and HTML comment + plain text).

## Test plan
- [x] `:rewrite-java-{17,21,25}:compatibilityTest` green
- [x] `:rewrite-java-test:test` green
- [x] Full `rewrite-static-analysis` test suite green against local 8.82.0-SNAPSHOT (including the previously failing `MavenJavadocNonAsciiRecipeTest`)

### DIFF
--- a/rewrite-benchmarks/build.gradle.kts
+++ b/rewrite-benchmarks/build.gradle.kts
@@ -14,7 +14,7 @@ dependencies {
 
     jmh(project(":rewrite-core"))
     jmh(project(":rewrite-java"))
-    jmh(project(":rewrite-java-25"))
+    jmh(project(":rewrite-java-21"))
     jmh(project(":rewrite-javascript"))
     jmh(project(":rewrite-maven"))
     jmh("org.antlr:antlr4-runtime:4.13.2")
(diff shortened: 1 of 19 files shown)


timtebeek:
We've since fixed the issue downstream by upgrading to `rewrite-java-25` there when running our tests on Java 25:
- https://github.com/openrewrite/rewrite-static-analysis/pull/869

greg-at-moderne (approved):

PR #7534, opened 2026-04-30
Pick Gradle distribution based on client JVM version

timtebeek:
## Summary
- `OpenRewriteModelBuilder` previously hardcoded a fallback Gradle version of `8.12`, which can't host its daemon on Java 24+ or Java 25+. Projects without a `gradle-wrapper.properties` therefore fail to build their tooling model on newer JDKs.
- Replace the constant with `defaultGradleVersion(int javaFeatureVersion)` that maps the running Java version to the lowest compatible distribution per the [Gradle compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html):
  - Java <= 23 -> `8.12` (unchanged)
  - Java 24 -> `8.14.3`
  - Java 25+ -> `9.1.0`
- Java version is read from `java.specification.version` (Java 8 compatible); the existing `org.openrewrite.test.gradleVersion` system-property override and wrapper-detection paths are untouched.
- Added `OpenRewriteModelBuilderTest` to lock in the version mapping.

## Notes / open questions
- Gradle 9.x daemons require Java 17+. That's not a regression here since clients on Java <17 still resolve to `8.12`, but worth flagging.
- Tooling API client jar version isn't bumped in this PR; the existing client supports driving the newer daemons we now request.
- Consider whether `8.14.3` / `9.1.0` should be parameterized (e.g. read from a properties file) so we can advance them without code changes — happy to follow up if preferred.

## Test plan
- [x] `./gradlew :rewrite-gradle-tooling-model:model:test --tests "org.openrewrite.gradle.toolingapi.OpenRewriteModelBuilderTest"` passes locally
- [ ] Manually exercise `OpenRewriteModelBuilder.forProjectDirectory` from a project without a wrapper, on Java 25, and confirm the daemon starts
- [ ] CI green

### DIFF
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModelBuilder.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModelBuilder.java
@@ -145,10 +145,7 @@ public class OpenRewriteModelBuilder {
         if (javaFeatureVersion >= 25) {
             return "9.1.0";
         }
-        if (javaFeatureVersion >= 24) {
-            return "8.14.3";
-        }
-        return "8.12";
+        return "8.14.3";
     }
 
     private static int javaSpecificationVersion() {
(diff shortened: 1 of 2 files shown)


rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModelBuilder.java
```diff
@@ -132,6 +132,37 @@ public static OpenRewriteModel forProjectDirectory(File projectDir, @Nullable Fi
...
+    }
+
+    static String defaultGradleVersion(int javaFeatureVersion) {
+        if (javaFeatureVersion >= 25) {
+            return "9.1.0";
+        }
+        if (javaFeatureVersion >= 24) {
+            return "8.14.3";
+        }
+        return "8.12";
```
timtebeek:
```suggestion
        return "8.14.3";
```

timtebeek:
Making this change now, as we're seeing downstream test failures when running on Java 25
 - https://github.com/openrewrite/rewrite-feature-flags/pull/65

shanman190 (approved):

sullis:
nice work @timtebeek

PR #7537, opened 2026-05-01
Pin Tooling-API client default to Gradle 8.14.3

timtebeek:
## Summary

- Reverts #7534, which dispatched the Tooling API client's Gradle distribution by JVM feature version (Gradle 9.1.0 for Java 25+, otherwise 8.14.3).
- Pins the default to a single hardcoded `8.14.3` (was briefly `8.14.4` on this branch — see below).
- Drops the now-unused `defaultGradleVersion()` / `javaSpecificationVersion()` helpers and the test class that only exercised them.
- Preserves a `wrapperGradleVersion()` / `isGradle9OrLater()` codepath so that callers that supply a `gradle-wrapper.properties` pointing at Gradle 9 still get the right CLI shape (see Gradle 9 note below). With the 8.14.3 default this code is inert until a wrapper opts in.

### 8.14.4: Kotlin DSL regression
Gradle 8.14.4 introduced a Kotlin DSL compilation regression that breaks `plugins {}` blocks containing local `val` declarations. The Kotlin script compiler fails to resolve the `PluginDependenciesSpecScope` supertype:

```
Supertypes of the following classes cannot be resolved.
class org.gradle.kotlin.dsl.PluginDependenciesSpecScope, unresolved supertypes: org.gradle.plugin.use.PluginDependenciesSpec
```

This caused 16 `:rewrite-gradle:test` failures on this branch (e.g. `UpgradePluginVersionTest.upgradeKotlinPluginLocalVariable`, `UpgradeTransitiveDependencyVersionTest.kotlinDslAddConstraint`). The same suite passes against 8.14.3.

### 9.x: CLI shape changed
Gradle 9 dropped the `-b <buildFile>` argument used by `OpenRewriteModelBuilder` to point at non-conventional build files. The replacement is to declare `rootProject.buildFileName` in `settings.gradle`, which forces the Tooling API client to synthesize a settings file for tests that don't ship one. Going to 9.x by default would also raise the minimum daemon JVM and force every consumer onto a major upgrade, which is more change than this PR wants to land. The new helpers in `OpenRewriteModelBuilder` (`isGradle9OrLater`, settings.gradle writeback) are kept so consumers who *do* opt into Gradle 9 via their own wrapper still work.

### 8.14.3: no known regressions
8.14.3 is the same minor the picker chose for Java &lt; 25 before #7534 was reverted, so this PR is a strict simplification of behavior that was already shipping on `main`. CI runs on Java 21 and Java 25; the `rewrite-gradle` test JVM is pinned to a Java 21 toolchain so 8.14.3 is JVM-compatible in both matrix legs.

## Test plan

- [ ] CI build passes on the PR.
- [x] Local `:rewrite-gradle:test` passes against 8.14.3 (the same suite exhibited the 8.14.4 Kotlin DSL failures).


timtebeek:
It turns out Gradle 8.x saw a backport of Java 25 support in 8.14.4, which was missed earlier. Proven to work downstream, and simpler than working around the dropped `-b` flag in 9.x here in a subsequent PR.

timtebeek:
Merging already after tests with downstream projects to avoid longer unavailability of builds & snapshots downstream.

PR #7539, opened 2026-05-01
Java: fix Javadoc roundtrip for multi-line HTML comments inside block tags

knutwannheden:
## Motivation

- Reported via [moderneinc/customer-requests#2277](https://github.com/moderneinc/customer-requests/issues/2277): a Javadoc with a multi-line HTML comment inside a `@version` block tag fails to parse-and-print back to the original source. The leading `*` margins on the inner lines of the HTML comment are dropped, and stray `*` lines appear before the closing `**/`.

The root cause is in `ReloadableJava{8,11,17,21,25}JavadocVisitor.convertMultiline`. Block-tag bodies (`@version`, `@author`, `@deprecated`, `@since`, `@hidden`, `@return`, `@param`, `@throws`, `@serial`, …) flow through `convertMultiline`, which dispatches non-text children via `scan(dt, emptyList())`. For an HTML comment that lands in `visitComment`, which only handles single-line bodies — it does `cursor += node.getBody().length()` and emits a single `Javadoc.Text` with the entire body inline. For a multi-line body this:

- never visits the body's newlines, so the margin `LineBreak` markers in the visitor's `lineBreaks` map are never consumed and leak out at the end of `visitDocComment` as the extra `*` lines before `**/`;
- flattens the body into a single `Text` with embedded `\n`s, so the printer never re-emits the per-line `*` margin.

`visitDocComment` already special-cases `DCComment` at the top level and routes it through `visitText(body)`, which handles newlines correctly. The fix extends that routing to `convertMultiline` so block-tag bodies behave the same.

## Summary

- `convertMultiline` now dispatches `DCTree.DCComment` to `visitText(body)`, mirroring the existing handling for `DCText` (and `DCRawText` on Java 25).
- Identical change applied to all five versioned Javadoc visitors (`rewrite-java-8/11/17/21/25`).
- Added a regression test `JavadocTest.htmlCommentNestedAfterVersionTag` covering the customer's input.

## Test plan

- [x] `./gradlew :rewrite-java-8:compatibilityTest :rewrite-java-11:compatibilityTest :rewrite-java-17:compatibilityTest :rewrite-java-21:compatibilityTest :rewrite-java-25:compatibilityTest --tests "org.openrewrite.java.tree.JavadocTest"` — all pass on every version, including the new regression test.


knutwannheden:
Looks like this was fixed already. Just found this old branch on my computer...

timtebeek:
Linking the earlier related work:
- https://github.com/openrewrite/rewrite/pull/7500
- https://github.com/openrewrite/rewrite/pull/7531

PR #7542, opened 2026-05-01
HCL: allow `if` and `in` as attribute and object key names

### LINKED ISSUE #6446: HclParser might fail if you use `if` as property name

scuba10steve:
## Summary
- The HCL lexer reserves `IF` and `IN` as dedicated tokens for use in for-expressions, but the HCL spec does not reserve these as keywords at the body level. Terraform itself accepts `if` and `in` as attribute names and object keys.
- Extend `attribute` and `qualifiedIdentifier` rules in `HCLParser.g4` to accept `IF` and `IN` alongside `Identifier`. The visitor walks `qualifiedIdentifier` children in source order so dotted-name positions stay correct when alternatives are interleaved.
- For-expression parsing (`for x in xs : x if cond`) is unaffected — those tokens are still consumed at the for-rule level. Verified by the existing `HclForTest` suite.

- Fixes #6446

## Test plan
- [x] `rewrite-hcl:test` — 9 new + existing tests pass:
  - `usingIfAsAttributeName` / `usingInAsAttributeName` — issue's reproducer
  - `keywordsAsTopLevelAttributes` — body-level form, no enclosing block
  - `keywordAttributeAlongsideForExpression` — coexistence with `for ... in ... if ...`
  - `keywordAttributeWithCommentAndUnusualSpacing` — formatting variations
  - `keywordsAsObjectKeys` / `keywordObjectKeyMixedWithRegularKeys` — object literal keys
- [x] `rewrite-hcl:check` — license, lint, all green
- [x] Full repo `./gradlew build` — `BUILD SUCCESSFUL in 10m 13s`, 442 tasks, no regressions in any downstream consumer
- [x] `HclForTest` — for-expression regression coverage still passes


timtebeek:
Thanks! We'd had a parallel effort open as well with a slightly different approach:
- https://github.com/openrewrite/rewrite/pull/6473

For both I'm wondering whether we should have these keywords called out in the grammar and generated parser, or whether we can better scope when to interpret and `if` as a conditional. No immediate suggestions here, but thought to call it out. Greg or Knut might have some ideas as well.

scuba10steve:
> Thanks! We'd had a parallel effort open as well with a slightly different approach:
> 
> * [Allow 'if' as HCL attribute name #6473](https://github.com/openrewrite/rewrite/pull/6473)
> 
> For both I'm wondering whether we should have these keywords called out in the grammar and generated parser, or whether we can better scope when to interpret and `if` as a conditional. No immediate suggestions here, but thought to call it out. Greg or Knut might have some ideas as well.

ha!

If you already had something going let's just take yours.  I haven't really tested this at all so it was just something I threw together.

PR #7545, opened 2026-05-01
Normalize null collections in RecipeDescriptor deserialization

jkschneider:
## What

`Recipe.createRecipeDescriptor()` always passes non-null lists for `tags`, `options`, `preconditions`, `recipeList`, `dataTables`, `maintainers`, `contributors`, and `examples` — and callers across the ecosystem rely on that, iterating these getters without null checks.

When a `RecipeDescriptor` is deserialized from JSON returned by a polyglot RPC peer (rewrite-javascript-remote, rewrite-csharp-remote, rewrite-python-remote), those peers omit empty collection-valued fields. Jackson's default `@AllArgsConstructor(@JsonCreator)` binding then leaves those fields `null`, and downstream code blows up:

```
Cannot invoke "java.util.List.iterator()" because the return value of
"org.openrewrite.config.RecipeDescriptor.getPreconditions()" is null
```

## How

Replace the Lombok `@Value` + `@AllArgsConstructor(onConstructor = @__(@JsonCreator))` with an explicit `@JsonCreator` constructor that normalizes null → empty at the deserialization boundary. Field-level immutability, getters, `equals`/`hashCode`, `toString`, and `@With` are preserved via explicit `private final` + `@Getter` / `@ToString` / `@EqualsAndHashCode`.

Locally constructed `RecipeDescriptor`s are unaffected (they were already passing non-null collections).

## Test plan

- [x] New `RecipeDescriptorTest.deserializeOmittedCollectionsAsEmpty` locks the contract: a JSON document missing `tags`, `options`, `preconditions`, `recipeList`, `dataTables`, `maintainers`, `contributors`, and `examples` deserializes to a descriptor whose getters return empty collections, never null.
- [x] Full `:rewrite-core:test` passes.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/config/RecipeDescriptor.java
@@ -16,9 +16,10 @@
 package org.openrewrite.config;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import lombok.AllArgsConstructor;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.EqualsAndHashCode;
-import lombok.Value;
+import lombok.Getter;
+import lombok.ToString;
 import lombok.With;
 import org.intellij.lang.annotations.Language;
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


jkschneider:
Withdrawing — taking a different approach. Will normalize at the per-language RPC bundle reader (NpmRecipeBundleReader / NuGetRecipeBundleReader / PipRecipeBundleReader) instead of restructuring RecipeDescriptor.

PR #7552, opened 2026-05-03
JS: skip `npm publish` when snapshot version already exists

timtebeek:
## Summary
- The scheduled `main` CI job failed at `:rewrite-javascript:npmPublish` with `npm error 403 ... You cannot publish over the previously published versions: 8.82.0-20260502-150813.` The snapshot version is derived from the latest commit timestamp, so when no new commit lands between scheduled runs, npm rejects the duplicate publish.
- npm has no `--skip-duplicate` equivalent, so this adds an `onlyIf` to `npmPublish` that runs `npm view @openrewrite/rewrite@<version>` and skips the task when the version is already in the registry. Mirrors the intent of #7544 (which added `--skip-duplicate` for NuGet).

## Test plan
- [ ] Next scheduled `main` run succeeds (or logs `Skipping npmPublish: ... already published` if the timestamp hasn't moved).

### DIFF
--- a/rewrite-javascript/build.gradle.kts
+++ b/rewrite-javascript/build.gradle.kts
@@ -253,6 +253,24 @@ val npmPublish = tasks.register<NpmTask>("npmPublish") {
     }
 
     workingDir.set(file("rewrite"))
+
+    // npm has no `--skip-duplicate`; the daily scheduled publish would fail with a 403
+    // whenever no new commit had landed since the previous run (the snapshot version is
+    // derived from the latest commit timestamp). Skip the task if the version already exists.
+    onlyIf {
+        val versionToCheck = extractVersionFromJar() ?: datedSnapshotVersion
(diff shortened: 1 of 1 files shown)


greg-at-moderne (approved):
Glad to see this minor thing fixed.

timtebeek:
Heads up: this introduced a regression I'm tracking in #7570. The `onlyIf` block uses `redirectErrorStream(true)` and then does `output.contains(versionToCheck)`, but npm's 404 error message itself contains the version we asked for (`"The requested resource '@openrewrite/rewrite@<version>' could not be found"`), so the contains-check matches on every miss and the publish has been silently skipped since 2026-05-02.

Net effect: Maven snapshots keep advancing, npm prerelease tag is frozen at `8.82.0-20260502-150813`, and downstream `npx --package=@openrewrite/rewrite@<jar-version> rewrite-rpc` calls fail with `ETARGET`. That's what's been breaking rewrite-static-analysis scheduled CI since 2026-05-04. #7570 has the fix.

PR #7560, opened 2026-05-04
Go: fix parsing of condition-only for loops using semicolons

greg-at-moderne:
## What's changed?

Fixing the case of `for` loops which don't use the 3-arg (`int i=0; i<5; i++`), but rather the 1-arg version. If they happen to contain `;` even in a literal, they shouldn't be treated as the 3-arg, e.g.
```
for tok != ';' { ... }
```

## What's your motivation?

The parsing fails with idempotency issues.


greg-at-moderne:
The CI failures are unrelated (they pertain to Gradle).

PR #7581, opened 2026-05-06
Gradle 9: move top-level sourceCompatibility/targetCompatibility into `java { }`

Jenson3210:
## Summary
Gradle 9 removed `JavaPluginConvention` (deprecated in 8.2), so top-level convention assignments in Groovy build scripts no longer compile:

```groovy
sourceCompatibility = '17'
```

Adds a new recipe `org.openrewrite.gradle.gradle9.UseJavaExtensionBlock` that:

- Detects top-level `sourceCompatibility` / `targetCompatibility` assignments in `build.gradle` (Groovy DSL).
- Removes them and re-emits them inside the `java { }` extension block (creating the block if missing, merging into an existing block if present).
- Normalizes the value to `JavaVersion.VERSION_<n>` form.
- Mirrors the missing counterpart so both properties are set explicitly when only one was declared at the top level.
- When both an existing `java { }` entry and a top-level convention assignment declare the same property, the existing block's entry is kept.

Wired into `org.openrewrite.gradle.MigrateToGradle9`.

- Resolves [moderneinc/customer-requests#2332](https://github.com/moderneinc/customer-requests/issues/2332).

### Example
Input:
```groovy
sourceCompatibility = '17'
```

Output:
```groovy
java {
    sourceCompatibility = JavaVersion.VERSION_17
    targetCompatibility = JavaVersion.VERSION_17
}
```

### Out of scope
- Kotlin DSL (`build.gradle.kts`): top-level convention assignments are far rarer there since Kotlin DSL is typed; can be a follow-up if needed.
- Sibling convention removals (`BasePluginConvention`, `ApplicationPluginConvention`, `WarPluginConvention`, `EarPluginConvention`).

## Test plan
- [x] `UseJavaExtensionBlockTest` covers: source-only, target-only (mirroring), both, merge into existing block, keep existing values when also set at top level, enum/numeric/double normalization, idempotency, and Kotlin DSL no-op.
- [x] `MigrateToGradle9Test` exercises the new recipe end-to-end via the composite recipe.
- [x] `./gradlew :rewrite-gradle:test --tests "*UseJavaExtensionBlock*" --tests "*MigrateToGradle9Test*"` passes.


rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseJavaExtensionBlock.java
```diff
@@ -0,0 +1,273 @@
...
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsBuildGradle<>(), new JavaVisitor<ExecutionContext>() {
+
+            @Override
+            public @Nullable J visit(@Nullable Tree tree, ExecutionContext ctx) {
+                J visited = super.visit(tree, ctx);
+                Map<String, Integer> versionsToMove = getCursor().pollMessage(VERSIONS_KEY);
+                if (!(visited instanceof G.CompilationUnit) || versionsToMove == null || versionsToMove.isEmpty()) {
+                    return visited;
+                }
```
steve-aom-elliott, line 77: Do we also need to handle `K.CompilationUnit` for `build.gradle.kts` files?

Edit: Ah, I see you have it listed as a follow-up instead.

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseJavaExtensionBlock.java
```diff
@@ -0,0 +1,273 @@
...
+        if (s instanceof J.Return) {
+            Expression e = ((J.Return) s).getExpression();
+            if (e instanceof J.Assignment) {
+                return (J.Assignment) e;
+            }
+        }
+        return null;
+    }
+
+    private static @Nullable Integer extractVersion(Expression e) {
```
steve-aom-elliott, line 233: Is it possible the value for the `sourceCompatibility` /  `targetCompatibility` at the top level might come from a variable rather than a literal, field access or method invocation?

Jenson3210: while certainly possible, I think we can add such iterations when we encounter them like we did for the gradle dependency trait/recipes

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseJavaExtensionBlock.java
Jenson3210: Claude:
yes in principle, but it’s much rarer in practice and worth doing as a follow-up.

Why Kotlin DSL is mostly unaffected: Kotlin DSL is statically typed, so writing sourceCompatibility = "17" at the top level of build.gradle.kts already required jumping through hoops in Gradle 8 (typically `(project as JavaPluginConvention).sourceCompatibility = ...)`. Most Kotlin DSL files have always used `java { sourceCompatibility = ... }` directly, so they’re not affected by the convention removal.

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseJavaExtensionBlock.java
Jenson3210: It might even come from a gradle.properties file. 
But not in my current reported usecase. 
I would leave such optimizations to further reports (if ever)

Jenson3210: I added something that will still move, but might not compile in their machine if they are using variables. but they will be supported.

rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseJavaExtensionBlock.java
```diff
@@ -0,0 +1,294 @@
...
+
+    private static void appendEntry(StringBuilder snippet, String name, Map<String, Expression> entries, Set<String> needsReplacement) {
+        if (!entries.containsKey(name)) {
+            return;
+        }
+        Integer version = extractVersion(entries.get(name));
+        if (version != null) {
+            snippet.append("    ").append(name).append(" = ").append(enumForm(version)).append("\n");
+        } else {
+            snippet.append("    ").append(name).append(" = JavaVersion.VERSION_17\n");
```
steve-aom-elliott, line 193: So to be clear, hardcoded as `VERSION_17` here because it's replaced later?

Jenson3210: yes indeed

steve-aom-elliott (approved):
I believe it makes sense. I'm a not as confident in my understanding of this one compared to others only because there's so many passes of parsing / manipulation, but it seems to be getting the correct results.

Jenson3210:
The recipe: 

visitCU
  super.visit -> makes sure that visitMI is called / cursor will have message
  Makes sure that both compatibilities are in the map (if one is not mentioned, copy the other)
  ListUtils.map the statements and find the java block and merge our statements in (keeping existing ones in favor)
  If the previous did not update, then no java block exists, so add one with both compatibilities

visitMI
  puts message of the detected assignments
  
hope that clears it up a bit?

sullis:
nice work @Jenson3210

PR #7585, opened 2026-05-06
Gradle 9: rewrite spread-`all*` in `configurations { }` to `configurations.all { }`

Jenson3210:
## Summary
- Adds `org.openrewrite.gradle.gradle9.RewriteSpreadAllInConfigurationsBlock` to the `MigrateToGradle9` composite. Resolves [moderneinc/customer-requests#2335](https://github.com/moderneinc/customer-requests/issues/2335).

Gradle 9 throws

> `Cannot mutate the dependencies of configuration ':all' after the configuration was resolved.`

when a `configurations { }` closure uses Groovy's spread-dot form `all*.<method>(args)`. Customer reproduction confirmed the breaking pattern is specifically the spread-dot form (lifting the block out of `dependencies { }` does **not** help, and `.all → .configureEach` is out of scope — deprecation only). Replacing the spread with the closure form keeps eager-`all` semantics and is accepted by Gradle 9.

### Example
Before:
```groovy
configurations {
    all*.exclude group: "log4j", module: "log4j"
}
```

After:
```groovy
configurations.all {
    exclude group: "log4j", module: "log4j"
}
```

### Implementation notes
- Detects spread-dot via the `org.openrewrite.groovy.marker.StarDot` marker on the inner method name; handles `J.Return`-wrapped tail expressions in Groovy closures.
- Builds the new `configurations.all { }` block via `GradleParser.builder().build().parse(ctx, …)` (mirroring the pattern from #7581) and grafts the original (stripped) statements into its lambda body.
- Only rewrites when **every** statement inside the `configurations { }` block uses the spread form; mixed blocks are left untouched for manual review.

### Out of scope
- Mixed `configurations { … }` blocks containing both spread and non-spread statements (split into siblings) — punting until we see this in customer code.
- `.all → .configureEach` flip (deprecation only, not a build break).

## Test plan
- [x] `RewriteSpreadAllInConfigurationsBlockTest`: top-level, nested in `dependencies { }`, multiple spread statements collapsed, plus no-change cases for already-closure form, no-spread blocks, spread on non-`all` identifiers, mixed blocks, and `settings.gradle` files.
- [x] `MigrateToGradle9Test` (composite) still passes.
- [x] `./gradlew :rewrite-gradle:test --tests "*RewriteSpreadAllInConfigurationsBlockTest*" --tests "*MigrateToGradle9Test*"` passes.
- [x] `recipes.csv` updated: bumped `MigrateToGradle9` (40 → 41) and `GradleBestPractices` (50 → 51) recipe counts; added the new recipe row under `Gradle9`.


steve-aom-elliott:
Is a `JavaIsoVisitor` capable of visiting both a `build.gradle` and `build.gradle.kts`? I'm just wondering the if the part where you're producing a `G.CompilationUnit` needs to be safety checked further at all.

Jenson3210:
@steve-aom-elliott Good catch — `IsBuildGradle` matches both `.gradle` and `.gradle.kts`, so technically `JavaIsoVisitor` could end up here on a `K.CompilationUnit`. In practice the spread-dot form is Groovy-only syntax (Kotlin has no `*.`), so `isSpreadAll` would never match on a Kotlin file and `parseConfigurationsAllTemplate` wouldn't be reached. But you're right that it's better to be explicit than to rely on that.

Added an early-out in the visitor that bails when the enclosing source file isn't a `G.CompilationUnit`, plus a `kotlinDslLeftAlone` no-change test. Already pushed (force-push after rebasing onto main, since #7581 merged in the meantime).

Jenson3210:
```
                if (statements.isEmpty() || !statements.stream().allMatch(RewriteSpreadAllInConfigurationsBlock::isSpreadAll)) {
                    return m;
                }
``` 
would fail for kotlin as SpreadAll is Groovy only, but add as check

steve-aom-elliott (approved):

PR #7589, opened 2026-05-06
python: replace _Collector visitor with iterative tree walker in _collect_search_result_ids

jkschneider:
## Summary

`_collect_search_result_ids` previously built a `TreeVisitor` subclass inside the function and called `.visit(tree, None)` to traverse the tree and inspect each node's markers:

```python
class _Collector(TreeVisitor):
    def pre_visit(self, tree, p):
        _walk(tree)
        return tree
_Collector().visit(tree, None)
```

The collector extends `TreeVisitor` itself (not a language-specific subclass), so when it visited a Python `CompilationUnit` the LST's `accept_python` would call `v.visit_compilation_unit(self, p)` — a method `TreeVisitor` doesn't have. That raised `AttributeError` on every Python compilation unit it encountered (and similar errors for other language-specific node types).

The result was a per-file exception storm — **~9k exceptions per medium-baseline recipe run**. Each exception is wrapped in a `RecipeRunException` by `TreeVisitor.visit`'s outer try/except, which short-circuits the visit, so even when the call appeared to "work" it was returning a partial set of marker IDs.

This PR replaces the visitor approach with a hand-rolled iterative walker that traverses dataclass `__dict__` values directly. It recurses only into LST tree nodes (those with `_id`) and the structural padding wrappers (`JLeftPadded`/`JRightPadded`/`JContainer`) that link them, with an identity-based `seen` set for cycle protection. No visitor framework, no language-specific dispatch, no exceptions.

Side benefits:

- Removes per-call visitor dispatch overhead on top of the exception-handling cost
- Self-contained: doesn't depend on the language-specific visitor hierarchy, so adding a new language doesn't require updating this collector

## Test plan

- [x] Worktree run of `mod run small --recipe=org.openrewrite.python.migrate.UpgradeToPython314` (19 Python repos) succeeds; the search-result-id diff in `BatchVisit` continues to populate correctly
- [x] Manually traced LST node graphs to verify the walker reaches all reachable tree nodes via `_id`-bearing nodes plus the three padding wrapper types
- [x] `seen` set keyed by `id(node)` handles the cyclic-graph case

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/rpc/server.py
+++ b/rewrite-python/rewrite/src/rewrite/rpc/server.py
@@ -1350,25 +1350,59 @@ def handle_batch_visit(params: dict) -> dict:
 
 
 def _collect_search_result_ids(tree) -> set:
-    """Collect all SearchResult marker UUIDs from a tree."""
+    """Collect all SearchResult marker UUIDs reachable from a tree.
+
+    Hand-rolled iterative walker. The previous implementation built a
+    ``TreeVisitor`` subclass and called ``.visit(tree, None)``, which threw
+    ``AttributeError: '_Collector' object has no attribute 'visit_compilation_unit'``
+    on every Python ``CompilationUnit`` it encountered (and similar errors for
+    other language-specific node types) because ``TreeVisitor`` is the
+    language-agnostic base class. That produced a per-file exception storm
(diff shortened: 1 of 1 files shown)


jkschneider:
Rejected this hack around a fundamental bug in using TreeVisitor to visit SourceFile instances.

PR #7594, opened 2026-05-07
Ability to opt-out from Maven Settings being injected in selected test cases

greg-at-moderne:
## What's changed?

A follow-up to #7559.
A change to some of the Maven resolution tests. Adding a mechanism to explicitly get rid of the implied `maven-settings.xml` injected into all test cases. For these selected test cases, it won't make sense.

## What's your motivation?

Otherwise the tests would fail.


timtebeek:
I think we've gotten it all mostly working, except for maybe on forks; can we close this effort?

greg-at-moderne:
OK.
There are still some disabled tests though. But we can just disable them conditionally on CI and that's it.

PR #7596, opened 2026-05-07
Use application mainClass in MigrateToGradle9

sullis:
## Summary

Use application mainClass in MigrateToGradle9 recipe

# Test cases
- handle build.gradle 
- handle build.gradle.kts


sullis:
Github Actions

image

Jenson3210 (approved):
Wondering if we should also account for the use case where there already is an existing `application` present. I guess current implementation will create a new block next to the existing one. Similar to how `UseJavaExtensionBlock` would merge the new block with the existing if present.

sullis:
CI Build is green.

sullis:
@Jenson3210  can we merge this?

Jenson3210:
4 failures are due to fork being used it seems.

PR #7644, opened 2026-05-11
Fix NPE in ChangeManagedDependencyGroupIdAndArtifactId for deserialized Pom

MBoegers:
## Summary
- `getSafeVersionPlaceholdersToChange` threw an NPE at line 221 when `requestedPom.getDependencyManagement()` returned null.
- Although the field is declared `@Builder.Default = emptyList()`, deserialization can bypass the builder and leave it null (same diagnosis as the comment in `MavenDependencyPropertyUsageOverlap.java:43`).
- Added a null guard returning `emptySet()`, matching the existing precedent. `@SuppressWarnings("ConstantConditions")` is added because IntelliJ's static analysis cannot see the deserialization path — same approach already used on `resolveSemverVersion` in this file.

## Test plan
- [x] `./gradlew :rewrite-maven:test --tests "*ChangeManagedDependencyGroupIdAndArtifactIdTest"` passes
- [ ] Consider adding a regression test exercising a deserialized Pom with null `dependencyManagement`

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/ChangeManagedDependencyGroupIdAndArtifactId.java
@@ -30,6 +30,7 @@ import org.openrewrite.xml.tree.Xml;
 
 import java.util.*;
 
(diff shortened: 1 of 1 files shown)


timtebeek:
do we still need this?

PR #7649, opened 2026-05-11
Update server configuration in CI workflow

Jenson3210:
Enable fork repos to also run using the artifactory-cache

We still use the mirror, but do not add the servers as we do not have the secrets available on forks. This will use no-auth access on our cache, which should also work.


Jenson3210:
Not using if for now as this approach keeps using the mirror

PR #7651, opened 2026-05-11
Fix charset misdetection causing UnmappableCharacterException for CJK…

### LINKED ISSUE #7636: EncodingDetectingInputStream misdetects UTF-8 files with CJK content as Windows-1252, causing UnmappableCharacterException on write
### Problem

`EncodingDetectingInputStream.guessCharset()` uses a "first violation fails" strategy — if any single byte violates UTF-8 rules, the entire file is immediately classified as Windows-1252. This causes `UnmappableCharacterException` when writing back UTF-8 files that contain CJK characters alongside any non-UTF-8 byte (even a single stray byte on an empty line).

More critically, **even without any corrupted bytes, a UTF-8 file with CJK content misdetected as Windows-1252 will almost certainly fail to write back**. This is because UTF-8 multi-byte sequences for CJK characters routinely contain bytes in the `0x80-0x9F` range where Windows-1252 has 5 undefined positions (`0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D`). When decoded as Windows-1252, these bytes become `U+FFFD` (replacement character), which cannot be re-encoded as Windows-1252, triggering the exception.

### Root Cause

The current algorithm in `guessCharset()` sets `charset = WINDOWS_1252` on the first invalid byte encountered, with no possibility of recovery. This is overly aggressive for real-world UTF-8 files that may have minor corruption.

Additionally, Windows-1252 is a poor fallback charset because it has 5 undefined byte positions. When UTF-8 bytes are misinterpreted as Windows-1252 and hit these positions, the resulting `U+FFFD` characters make the content unencodable in Windows-1252, causing a guaranteed write-back failure.

### Reproduction

A `.properties` file with valid UTF-8 Chinese comments and a single stray `0xAA` byte on an otherwise empty line:

```
server.port=8080
app.name=my-service

<0xAA byte here>
db.url=jdbc:mysql://localhost:3306/mydb

# 关闭内置tomcat
spring.main.web-environment=false

# 配置，详见 http://example.com/docs
```

The `0xAA` byte is a bare continuation byte (invalid in UTF-8), triggering immediate Windows-1252 classification. When the file is later written back using the detected Windows-1252 charset, the Chinese characters (which were decoded as Windows-1252 codepoints) cannot be re-encoded because some UTF-8 bytes land on Windows-1252 undefined positions (producing `U+FFFD`), resulting in:

```
java.nio.charset.UnmappableCharacterException: Input length = 1
    at java.nio.charset.CoderResult.throwException
    at sun.nio.cs.StreamEncoder.implWrite
    at java.io.BufferedWriter.close
    at org.openrewrite.maven.AbstractRewriteRunMojo.writeAfter
```

In the test case, the following Chinese comment lines all contain bytes that map to undefined Windows-1252 positions:

| Chinese text | Undefined bytes hit |
|---|---|
| `配置，详见` | `0x8D`, `0x81` |
| `配置` | `0x8D` |
| `必填。订阅的` | `0x9D` |
| `注意：配置项需要配置到该属性下` | `0x8F`, `0x8D`, `0x81` |
| `应用的` | `0x90`, `0x8D` |
| `选填。不配置默认为` | `0x8D` |

### Proposed Fix

Add a verification step in `readFully()`: after decoding with the detected charset, check if the result can be re-encoded with that charset. If not, the detection was likely wrong — re-decode with UTF-8. This catches any misdetection regardless of the `guessCharset` implementation, and requires changing only one method.

**`EncodingDetectingInputStream.java`** — Modify `readFully()`:

```java
// Before
public String readFully() {
    try (InputStream is = this) {
        ByteArrayOutputStream bos = new ByteArrayOutputStream() {
            @Override
            public synchronized String toString() {
                return new String(buf, 0, count, getCharset());
            }
        };
        byte[] buffer = new byte[4096];
        int n;
        while ((n = is.read(buffer, 0, buffer.length)) != -1) {
            bos.write(buffer, 0, n);
        }
        return bos.toString();
    } catch (IOException e) {
        throw new UncheckedIOException(e);
    }
}

// After
public String readFully() {
    try (InputStream is = this) {
        ByteArrayOutputStream bos = new ByteArrayOutputStream() {
            @Override
            public synchronized String toString() {
                return new String(buf, 0, count, getCharset());
            }
        };
        byte[] buffer = new byte[4096];
        int n;
        while ((n = is.read(buffer, 0, buffer.length)) != -1) {
            bos.write(buffer, 0, n);
        }

        Charset detectedCharset = getCharset();
        String result = bos.toString();

        // If the detected charset can't re-encode the content, the detection
        // was likely wrong — fall back to UTF-8. This catches cases where
        // UTF-8 files with CJK content are misdetected as Windows-1252,
        // since Windows-1252 has 5 undefined byte positions that produce
        // U+FFFD (which can't be re-encoded as Windows-1252).
        if (detectedCharset != StandardCharsets.UTF_8 &&
                !detectedCharset.newEncoder().canEncode(result)) {
            charset = StandardCharsets.UTF_8;
            return new String(bos.toByteArray(), StandardCharsets.UTF_8);
        }

        return result;
    } catch (IOException e) {
        throw new UncheckedIOException(e);
    }
}
```

**Advantages:**
- Minimal code change (~5 lines added)
- No change to `guessCharset()` or `read()` — no performance impact on normal files
- Fixes the problem at the decoding stage where the raw bytes are still available for re-decoding
- All parsers benefit automatically since `readFully()` is the single entry point used by all parsers (XmlParser, YamlParser, PropertiesParser, JsonParser, JavaParser, KotlinParser, etc.)
- Genuine Windows-1252 files are unaffected (`canEncode` returns true for them)

### Rationale

- This fix leverages the fact that `U+FFFD` cannot be encoded in Windows-1252 as a reliable signal that the charset detection was wrong — a genuine Windows-1252 file would never produce `U+FFFD` during decoding.
- Windows-1252 is a particularly poor fallback charset because it has 5 undefined byte positions (`0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D`) that produce `U+FFFD` when decoding UTF-8 bytes, making write-back always fail for CJK content.
- When re-decoded as UTF-8, corrupted bytes become `U+FFFD` (replacement character), which is both safe and semantically correct.
- The fix is applied in `readFully()`, which is the single entry point used by all parsers (XmlParser, YamlParser, PropertiesParser, JsonParser, JavaParser, KotlinParser, etc.), so all parsers benefit automatically.
- ISO-8859-1 (with its identity byte-to-codepoint mapping) would be a safer fallback charset than Windows-1252, but fixing the detection verification is the preferred solution.

### Environment

- rewrite-maven-plugin 6.36.0
- JDK 8

XiaoSK:
- Fix #7636

When EncodingDetectingInputStream detects a file as Windows-1252 due to a stray invalid UTF-8 byte, but the file actually contains CJK content in UTF-8, the decoded content will contain U+FFFD characters (from Windows-1252 undefined byte positions) that cannot be re-encoded as Windows-1252, causing UnmappableCharacterException on write-back.

Add a canEncode verification in readFully(): after decoding with the detected charset, if the result cannot be re-encoded with that charset, re-decode with UTF-8 instead. This catches misdetections at the decoding stage where raw bytes are still available, and all parsers benefit automatically since readFully() is the shared entry point.


timtebeek (approved):
Thanks again for the suggestion and fix here @XiaoSK ! Took a little longer to get to than I would have liked, but I'd wanted to see if we could avoid any performance penalties here, which I think we've tackled with the changes I've pushed to your branch. With that I think we should be ok to merge this improvement, and I hope it helps you there!

greg-at-moderne:
Thanks for the contribution, @XiaoSK!

PR #7670, opened 2026-05-12
Stop closing recipe classloaders on CachingMavenRecipeBundleResolver close

jkschneider:
## Summary

`CachingMavenRecipeBundleResolver` hands out `RecipeBundleReader` instances whose `RecipeClassLoader`s are retained externally. Once a `Recipe` has been prepared, the JVM's class cache contains classes defined by that classloader, and any later `getResourceAsStream` / inner-class load against those classes requires the classloader's `URLClassPath` jar handles to remain open.

Two close paths in the prior implementation closed those handles while consumers still held the classloader:

- 1. **`close()` cascade.** Added by #7277 to fix Windows CI file-handle leaks, this iterated every cached `MavenRecipeBundleResolver` and called `close()` on it, transitively closing the reader and its `URLClassLoader`.
2. **`resolverFor()` version eviction.** Compared the cached entry's `bundle.getVersion()` against the incoming bundle's and called `markEvicted()` on mismatch, also closing the inner resolver/reader/classloader. Because `MavenRecipeBundleResolver.resolve()` mutates the input bundle's `version` to the dated snapshot after resolution, any fresh `RecipeBundle` still on the requested `-SNAPSHOT` version mismatches the cached entry and triggers eviction. The RPC `PrepareRecipe` handler routinely passes such fresh bundles back through the resolvers.

The visible symptom is inner-class `NoClassDefFoundError` on a classloader whose `==` identity is unchanged but whose `URLClassPath` has been closed: `findResource` returns null silently; `loadClass` falls back to parent which doesn't have the recipe class; calling code sees `NCDFE` on `new InnerClass()`. Reproduced against `rewrite-prethink` (jtokkit `cl100k_base.tiktoken` null → `ComprehendCodeTokenCounter\$CharCountTokenizer` NCDFE) and `rewrite-static-analysis` (`FinalClassVisitor\$FinalizingVisitor` NCDFE).

- See moderneinc/customer-requests#2346.

## Changes

- Drop the lease/proxy/eviction machinery from `CachingMavenRecipeBundleResolver`. The cache is now a plain package-coordinate `computeIfAbsent`.
- `close()` is a no-op with a class-level javadoc explaining the ownership-transfer semantics. Consumers that need deterministic cleanup must close the `RecipeBundleReader`s they retain (or rely on JVM exit for one-shot processes like `mod`).
- Effectively reverts the destructive part of #7277. The CI file-handle leak it addressed is the responsibility of test code that retained `RecipeBundleReader` instances without closing them.

## Reproducer

New `CachingMavenRecipeBundleResolverTest.closeDoesNotCloseHandedOutClassLoaders` instruments a `TrackingRecipeClassLoader` and asserts \`closeCount == 0\` after \`resolver.close()\`. Verified the test fails against the prior implementation (\`closeCount == 1\`) and passes after this change.

## Companion fix

A defense-in-depth no-op override of `CliRecipeClassLoader.close()` will land in moderne-cli alongside the bom bump, so any other close path we have not yet identified cannot regress the symptom. `mod` is one-shot; the OS reclaims jar handles at JVM exit.

## Test plan

- [x] \`./gradlew :rewrite-maven:test --tests "org.openrewrite.maven.marketplace.*"\` passes
- [x] New reproducer test fails on \`origin/main\` (closeCount was 1) and passes after this change

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/marketplace/CachingMavenRecipeBundleResolver.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/marketplace/CachingMavenRecipeBundleResolver.java
@@ -25,15 +25,36 @@ import org.openrewrite.maven.utilities.MavenArtifactDownloader;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
-import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ * Deduplicates {@link MavenRecipeBundleResolver} instances per package coordinate so that
+ * repeated {@link #resolve(RecipeBundle)} calls for the same package share one reader
+ * (and therefore one {@link org.openrewrite.marketplace.RecipeClassLoader}).
+ * <p>
+ * <strong>Lifetime of handed-out readers.</strong> The readers (and the
+ * {@link org.openrewrite.marketplace.RecipeClassLoader}s they own) returned by
+ * {@link #resolve(RecipeBundle)} are intended to outlive this resolver. Once a {@link Recipe}
+ * has been prepared from a reader, the JVM's class cache contains classes defined by that
+ * classloader, and any later {@code getResourceAsStream}/inner-class load against those
+ * classes requires the classloader's {@code URLClassPath} jar handles to remain open.
+ * <p>
+ * For that reason {@link #close()} only clears the deduplication cache. It does <em>not</em>
+ * close the cached inner {@link MavenRecipeBundleResolver}s -- doing so would close their
+ * readers, which would close the underlying {@code URLClassLoader}s while consumers still
+ * hold references. The visible failure mode is inner-class {@code NoClassDefFoundError} and
+ * silently-null {@code getResourceAsStream} returns on a classloader that {@code ==} identity
+ * confirms is still alive (see
+ * <a href="https://github.com/moderneinc/customer-requests/issues/2346">customer-requests#2346</a>).
+ * <p>
+ * Consumers that need deterministic cleanup must close the {@link RecipeBundleReader}s they
+ * retain (or rely on JVM exit for one-shot processes).
+ */
 @RequiredArgsConstructor
 public class CachingMavenRecipeBundleResolver implements RecipeBundleResolver {
     private final ExecutionContext ctx;
     private final MavenArtifactDownloader downloader;
     private final RecipeClassLoaderFactory classLoaderFactory;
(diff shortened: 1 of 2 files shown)


jkschneider:
> That explains why it always works on one repo, but I'm confused why it sometimes works on multiple repos

Two factors compound to make multi-repo data-dependent:

**1. URLClassPath closure doesn't invalidate the JVM's already-loaded class cache.** After `URLClassPath.closeLoaders()` clears the loaders list and sets `closed=true`, `findResource` returns null silently — but `URLClassLoader.loadClass(name)` checks `findLoadedClass(name)` first. Classes the loader already defined stay valid for the loader's lifetime; only *new* lazy loads (and any `getResourceAsStream`) fail. So whether NCDFE manifests depends on whether the recipe needs to load an inner class it hasn't loaded yet. For prethink, `ComprehendCodeTokenCounter$CharCountTokenizer` only loads when `new OpenAiTokenizer("gpt-4")` throws — i.e. only when `cl100k_base.tiktoken` already returned null, which is the same closure state. For static-analysis, `FinalClassVisitor$FinalizingVisitor` only loads when a source actually needs finalization. Inputs that never take those branches stay green.

**2. There are multiple close triggers, not just end-of-partition.** Your customer probe showed the URLClassPath transitioning from open to closed *inside* a single `getInitialValue` call — that's not the end-of-partition `resolver.close()` cascade. It's the second path I cover here: `resolverFor`'s version-mismatch `markEvicted()`. `MavenRecipeBundleResolver.resolve()` mutates `bundle.version` to the dated snapshot after first resolution. Any later call with a fresh `RecipeBundle` for the same package still on the requested `-SNAPSHOT` version mismatches and closes the cached resolver's classloader. The RPC `PrepareRecipe` handler at `RewriteRpc.java:160` does exactly this — it calls `marketplace.findRecipe(id).prepare(resolvers, opts)` with the global config marketplace's bundle (un-mutated), and any RPC peer callback during an umbrella recipe is a candidate eviction event. Whether one fires depends on the recipe + RPC peer activity.

So: single repo is deterministic because nothing closes the classloader before the run ends. Multi-repo is variable because (a) parallel partitions race against each other's `resolver.close()`, and (b) RPC peer PrepareRecipe callbacks fire opportunistically and can evict mid-recipe — but only manifest as NCDFE if the recipe subsequently needs an inner class the JVM hasn't already cached.

This PR fixes both close paths. The companion `CliRecipeClassLoader.close()` no-op in moderne-cli is the safety net for any other close path we haven't found.

PR #7671, opened 2026-05-12
Track resolved repository URLs per recipe bundle

pdelagrave:
## Problem

The recipe-marketplace platform installs recipe bundles from a configured list of artifact repositories, but has no record of which repository URLs the resolver actually fetched from. When a repo is removed from config, the platform can't tell which previously-installed bundles are now broken. To fix it on the platform side, we need that information from the rewrite resolvers.

## Solution

Adds `RecipeBundleReader.getResolvedFromRepositories()` with a default-empty implementation, plus per-ecosystem overrides:

* **Maven**: walks `MavenResolutionResult.getDependencies()` and reads each `ResolvedDependency.getRepository().getUri()`.
* **npm** (TS worker): parses `package-lock.json` after install, strips each `resolved` tarball URL down to the registry base.
* **pip** (Python worker): adds `--report <tempfile>` to `pip install`, reduces each `download_info.url` to its `scheme://host` origin.
* **NuGet** (C# worker): runs an explicit `dotnet restore --packages <fresh-scratch> --force --no-cache --verbosity normal` and parses the `Installed ... from <url>` log lines.

URLs are returned as the underlying tool reported them; normalization is a platform-side concern.

The CLI-side consumer is moderneinc/moderne-cli#3849.

## Test plan

- [x] Unit tests for each reader and each worker-side parser.
- [x] End-to-end install of a real recipe package through the patched mod CLI for all four ecosystems (Maven on host; npm/pip/NuGet each in a dedicated docker image). Sidecar JSON written by the CLI contained the expected source URL for every install.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeBundleReader.java
+++ b/rewrite-core/src/main/java/org/openrewrite/marketplace/RecipeBundleReader.java
@@ -18,7 +18,9 @@ package org.openrewrite.marketplace;
 import org.openrewrite.Recipe;
 import org.openrewrite.config.RecipeDescriptor;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 
 public interface RecipeBundleReader extends AutoCloseable {
     RecipeBundle getBundle();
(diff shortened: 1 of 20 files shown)


pdelagrave:
Closing, won't implement: tools for ecosystem other than maven aren't providing a reliable enough trace of which repository was used to resolve/fetch which artifact.

PR #7678, opened 2026-05-13
Support parsing Python 2.7 source through the existing parser pipeline

bmuschko:
Adds full Python 2.7 parsing support, plumbed end-to-end from Java through the RPC pipeline to the in-process printer's round-trip check.

## Public API

- `PythonParser.builder().languageLevel(PythonLanguageLevel.PYTHON_2_7)` —  per-parse version selection. Enum follows the `KotlinParser` /  `KotlinLanguageLevel` house style.
- `Parse` RPC request gains a nullable `Map<String, String> options` field (backwards-compatible; null on older clients).

## Auto-detection

When `languageLevel` isn't set, the Python RPC server picks Py2 vs Py3 per parse using (in order): explicit option, in-source signals (PEP-263 magic comment / shebang), project metadata (`pyproject.toml` classifiers and `requires-python`, `setup.cfg` classifiers), `REWRITE_PYTHON_VERSION` env default. The Moderne CLI's existing `parseProject(...)` path picks up Py2 repos with no CLI change required.

## Parser

The pre-existing `_py2_parser_visitor.py` was a stub: it handled only leaves, `print`/`exec`, simple assignment, and backticks; everything else returned placeholder identifiers like `<funcdef>`. This PR fills in the grammar:

- Expressions: binary fold (incl. `not in` / `is not`), unary, ternary, parens, lambda, calls (positional + keyword + `*`/`**`), attribute chains, subscripts, slices.
- Statements: `if`/`elif`/`else`, `while`, `for` (with tuple destructure and `else:` clause), `try`/`except`/`else`/`finally`, `with` (multi-context), all import shapes (incl. relative + star + parenthesized), `raise`/`assert`/`del`/`global`/`yield`, chained and augmented assignment, decorators (single, call-form, dotted, multi), classes (incl. old-style no-parens), functions (all param shapes).
- Atoms: list/dict/tuple/set/empty literals, all comprehension kinds, generator-as-call-arg.
- Literals: long suffix `100L`, Py2 octal `0777`, all string prefixes (`u`, `r`, `ur`, `b`), implicit string concatenation.
- Whitespace: newline-between-statement threading, trailing-comma preservation via `TrailingComma` marker, trailing-comment preservation, `OmitParentheses` on bare testlists.

## Py2-only markers + matching printer branches

- `TupleExceptClause` → emits `except E, e:` (Py2 comma form)
- `LegacyNotEqual` → emits `<>` (note: parso < 0.8 doesn't actually parse `<>`, so the marker is reachable only via future grammar work)
- `RaiseTuple` → emits `raise E, v, tb`

## Known gaps (documented in code)

- `<>` operator support requires source pre-processing because parso doesn't parse it. Marker, fold logic, and printer hook are in place; the parse rewrite is left for a follow-up.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/python/markers.py
+++ b/rewrite-python/rewrite/src/rewrite/python/markers.py
@@ -89,6 +89,58 @@ class SuppressNewline(Marker):
 
 
 @dataclass(frozen=True, eq=False)
+class LegacyNotEqual(Marker):
+    """Marker for the Python 2 ``<>`` not-equal operator on a :class:`j.Binary`.
+
+    The default printer renders ``Binary.Type.NotEqual`` as ``!=``; when
+    this marker is present it emits the legacy ``<>`` spelling instead.
+    """
+    _id: UUID
+
+    @property
+    def id(self) -> UUID:
+        return self._id
+
+    def with_id(self, id_: UUID) -> 'LegacyNotEqual':
+        return self if id_ is self._id else replace(self, _id=id_)
+
+
+@dataclass(frozen=True, eq=False)
+class RaiseTuple(Marker):
+    """Marker for the Python 2 three-argument ``raise E, v, tb`` form on a :class:`j.Throw`.
+
+    The exception slot of the Throw is a tuple literal of the three operands;
+    when this marker is present the printer drops the parentheses and renders
+    the legacy comma-separated form instead.
+    """
+    _id: UUID
+
+    @property
+    def id(self) -> UUID:
+        return self._id
+
+    def with_id(self, id_: UUID) -> 'RaiseTuple':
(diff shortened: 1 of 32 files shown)


rewrite-python/rewrite/src/rewrite/python/markers.py
```diff
@@ -88,6 +88,58 @@ def with_id(self, id_: UUID) -> 'SuppressNewline':
         return self if id_ is self._id else replace(self, _id=id_)
 
 
+@dataclass(frozen=True, eq=False)
+class LegacyNotEqual(Marker):
```
knutwannheden, line 92: We should probably define these new marker types in Java code as well.

knutwannheden (approved):

PR #7685, opened 2026-05-14
Pom only fallback non 404

Jenson3210:
# Fall back to POM-only resolution when optional `.module` fetch returns non-404

## Problem

When a POM contains the `published-with-gradle-metadata` marker,
`MavenPomDownloader` fetches the corresponding `.module` file to enrich
resolution. Today only an exact `404` is tolerated — any other status
aborts resolution of an otherwise-valid POM.

This breaks against common setups where a missing artifact isn't a 404:

- **Artifactory virtual repos accessed unauthenticated** → `401`

Several widely-used POMs (e.g. `spring-integration-bom`) carry the
Gradle-metadata marker without actually publishing a `.module`, so the
missing fetch is the rule, not the exception.

## Fix

Swallow the error whenever the server returned any response; only
propagate true connection-level failures and timeouts.

```java
if (!e.isServerReached() && !e.isClientSideException()) {
    throw e;
}
```

The `.module` fetch is best-effort enrichment — POM-only resolution
remains correct in its absence.

## Test

Added a unit test covering the `401` case, asserting that resolution
still succeeds.

- [si-3570]: https://github.com/spring-projects/spring-integration/issues/3570

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -657,7 +657,6 @@ public class MavenPomDownloader {
                                     new ByteArrayInputStream(pomResponseBody),
                                     Objects.equals(versionMaybeDatedSnapshot, gav.getVersion()) ? null : versionMaybeDatedSnapshot
                             );
-                            boolean cacheEmptyResult = true;
                             try (BufferedReader reader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(pomResponseBody)))) {
                                 String line;
                                 while ((line = reader.readLine()) != null) {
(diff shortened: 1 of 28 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -658,6 +657,7 @@ public Pom download(GroupArtifactVersion gav,
                                     new ByteArrayInputStream(pomResponseBody),
                                     Objects.equals(versionMaybeDatedSnapshot, gav.getVersion()) ? null : versionMaybeDatedSnapshot
                             );
+                            boolean cacheEmptyResult = true;
```
timtebeek:
```suggestion
```

timtebeek (approved):

PR #7697, opened 2026-05-15
C#: fix InstallRecipes echoing wildcard versions in response.Version

natedanner:
## Summary

- **Fixed:** the `InstallRecipes` RPC was returning the literal version-constraint string (`*`, `*-*`, etc.) as `InstallRecipesResponse.Version` instead of the resolved concrete SemVer. Downstream consumers (moderne-cli `recipes-v5.csv`, moderne-saas GraphQL) surfaced these as the "installed version" and had to work around it.
- **Root cause:** `dotnet add package <pkg> --version *` preserves the wildcard verbatim in the csproj's `<PackageReference>`. The old `ResolveVersionFromCsproj` read from that attribute, so wildcards flowed through unchanged. Verified against real `dotnet` output (`Version="*"` is written literally).
- **Fix:** new public helper `MSBuildProjectHelper.GetResolvedPackageVersion(projectDir, packageName)` reads `obj/project.assets.json` (NuGet's per-TFM resolution map written by restore) and returns the resolved version of a direct dependency. Reads from `targets` to match the convention already used by `ReadResolvedPackages` in the same file. Throws `InvalidOperationException` with specific diagnostics for: missing `project.assets.json`, package not a direct dep, or no matching resolved entry.
- **`InstallRecipes` call site** (`RewriteRpcServer.cs`) now delegates to the helper; the old private `ResolveVersionFromCsproj` is removed along with the unused `System.Xml.Linq` import.
- **Regression coverage** in `OpenRewrite.Tests/Rpc/InstallVersionResolverTests.cs`: theory rows for `*` and `*-*` against concrete-resolved fixtures, plus facts for case-insensitive package match, unknown-package throw, and missing-assets-json throw.

## Test plan

- [x] `dotnet test --filter InstallVersionResolverTests` — 5/5 pass in ~10 ms.
- [x] `./gradlew :rewrite-csharp:csharpTest` — 1936/1936 pass.
- [x] Mutation test: replacing the `targets` lookup with `project.frameworks.<tfm>.dependencies.<pkg>.version` (the original bug shape) causes the two wildcard rows + case-insensitive case to fail (3/5 failures). Confirms the tests actually cover the regression.
- [x] Verified against real `dotnet add package Newtonsoft.Json --version "*"` output: csproj receives `Version="*"`; `obj/project.assets.json` contains `targets.<tfm>."Newtonsoft.Json/13.0.4"` and `libraries."Newtonsoft.Json/13.0.4"`; helper returns `13.0.4`.

## Out of scope

- Pre-existing `_recipesProjectDir` caching across `InstallRecipes` calls (no per-call cleanup, no concurrency control). Not introduced by this PR.
- moderne-cli and moderne-saas workarounds that scan `~/.nuget/packages/` or post-process the version string; those become dead code once this fix ships.

### DIFF
--- a/rewrite-csharp/csharp/OpenRewrite.Tests/Rpc/InstallVersionResolverTests.cs
+++ b/rewrite-csharp/csharp/OpenRewrite.Tests/Rpc/InstallVersionResolverTests.cs
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using OpenRewrite.CSharp;
+
+namespace OpenRewrite.Tests.Rpc;
+
+/// <summary>
+/// Regression coverage for the InstallRecipes RPC, which used to echo wildcard versions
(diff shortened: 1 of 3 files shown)


macsux (commented):
/lgtm

PR #7700, opened 2026-05-16
Scala: handle varargs splat expressions

greg-at-moderne:
## What's changed?

Teach Scala parser to handle (varargs) splat expressions.

## What's your motivation?

They are not supported at all.


greg-at-moderne:
7700 in-flight emergency

Issue #7554, opened 2026-05-03
JDK26 behaves differently than JDK25 - "There were problems parsing"


koppor:
## What problem are you trying to solve?

Use OpenRewrite on JDK 26

~~Easy bug reporting to OpenRewrite and frictionless tool use.~~

~~Since years, dozens of files are not parseable by OpenRewrite **on Windows**.~~

## Describe the solution you'd like

OpenRewrite can parse all Java files.

## Have you considered any alternatives or workarounds?

Discuss java parsers with OpenRewrite team. [java parser](https://github.com/javaparser/javaparser) is maintained, but the current parsing logic is too heavy woven into a parser developed in parallel to javaparser.

## Additional context

```
There were problems parsing jabgui\src\main\java\org\jabref\gui\autocompleter\AutoCompletionTextInputBinding.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\cleanup\CleanupJournalRelatedPanel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\collab\DatabaseChangeMonitor.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\commonfxcontrols\CitationKeyPatternsPanel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\commonfxcontrols\CitationKeyPatternSuggestionCell.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\commonfxcontrols\FieldFormatterCleanupsPanel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\commonfxcontrols\SaveOrderConfigPanel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\consistency\ConsistencyCheckAction.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\consistency\ConsistencyCheckDialog.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\consistency\ConsistencyCheckDialogViewModel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\documentviewer\DocumentViewerView.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\documentviewer\DocumentViewerViewModel.java
There were problems parsing jabgui\src\main\java\org\jabref\gui\edit\automaticfiededitor\copyormovecontent\CopyOrMoveFieldContentTabView.java
[...]

PR #7716, opened 2026-05-18
Java parser: handle JDK 26 implicit lambda parameter type

timtebeek:
## Summary

- On JDK 26 the synthesized `vartype` for an implicit lambda parameter (e.g. `field` in `.filter(field -> ...)`) is a zero-width `JCErroneous` sitting at the parameter name, not a `NOPOS` node as on JDK ≤ 25 — so the existing `endPos(vartype) < 0` check in `ReloadableJava25ParserVisitor#visitVariables` no longer fires, and `convert(vartype)` produces a `J.Erroneous` that fails the implicit cast to `TypeTree` (surfacing downstream as the `StringIndexOutOfBoundsException` reported in #7554).

Fix detects both shapes via `endPos(vartype) <= vartype.getStartPosition()` gated on `Flags.PARAMETER`. Only `rewrite-java-25` is touched: `JavaParser` picks `Java25Parser` for any `version >= 25` so the older modules are never selected on JDK 26.

- Refs https://github.com/openrewrite/rewrite/issues/7554

## Test plan

- [x] New `Java26ImplicitLambdaParameterTest` parses a stream-chain `.filter(field -> ...)` — fails before the fix on JDK 26, passes after on both JDK 25 and JDK 26.
- [x] `:rewrite-java-25:compatibilityTest` still green on JDK 25.

### DIFF
--- a/rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
+++ b/rewrite-java-25/src/main/java/org/openrewrite/java/isolated/ReloadableJava25ParserVisitor.java
@@ -1750,6 +1750,16 @@ public class ReloadableJava25ParserVisitor extends TreePathScanner<J, Space> {
                 visitVariables(singletonList(node), fmt); // method arguments cannot be multi-declarations
     }
 
+    private boolean isImplicitLambdaParameterType(JCVariableDecl node, JCExpression vartype) {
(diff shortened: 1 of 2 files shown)


timtebeek:
Considering the new rewrite-java-next approach now that we're finding Java 26 specific parser fixes are needed here and in
- #7717

PR #7718, opened 2026-05-18
Skip incompatible registered mixins in TreeVisitorAdapter

timtebeek:
## Summary

Filter registered mixins by language compatibility in `TreeVisitorAdapter.discoverRegisteredMixin` so that adapting a base visitor to one language's `adaptTo` doesn't pick up another language's mixin and crash.

## Background

`RemoveAnnotationKotlinMixin` is registered for `RemoveAnnotationVisitor` via `META-INF/rewrite/mixins/org.openrewrite.java.RemoveAnnotationVisitor`. When `RemoveAnnotationVisitor` runs against a Groovy file, `G.accept` adapts it to `GroovyVisitor` — but `discoverRegisteredMixin` returned the Kotlin mixin unconditionally, and the assignability check at the top of `adapt` then threw:

```
java.lang.IllegalArgumentException: Mixin org.openrewrite.kotlin.RemoveAnnotationKotlinMixin
  is not assignable to org.openrewrite.groovy.GroovyVisitor; the mixin must extend
  (or be a subtype of) the adaptTo class so that proxy generation can extend it.
```

This affects any recipe that uses `RemoveAnnotation` (or any other base visitor with a language-specific mixin registered) when it touches a Groovy / non-Kotlin source file. Surfaced via `SpringBootVersionUpgradeTest.upgradeSpringCloudAwsVersion` in `rewrite-spring`, which exercises this on a Gradle Groovy build script.

## Fix

In `discoverRegisteredMixin`, skip entries whose class isn't a subtype of `adaptTo` rather than returning them. A registry file may list multiple language-specific mixins for the same base — the discovery loop now passes over incompatible ones and finds the matching one (or returns `null` so dispatch falls through to the base).

## Test plan

- [x] New `crossLanguageRegisteredMixinSkipped` in `TreeVisitorAdapterTest` — fails without the fix with the exact `IllegalArgumentException` above, passes with it.
- [x] Existing `TreeVisitorAdapterTest` cases still pass.
- [x] End-to-end: published locally and re-ran the originally failing `SpringBootVersionUpgradeTest.upgradeSpringCloudAwsVersion` in `rewrite-spring` — now green.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/TreeVisitorAdapter.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/TreeVisitorAdapter.java
@@ -368,6 +368,15 @@ public class TreeVisitorAdapter {
                                 throw new IllegalStateException(
                                         "Registered mixin " + line + " is not a TreeVisitor (registered at " + url + ").");
                             }
+                            // A single base visitor (e.g. RemoveAnnotationVisitor) may have
+                            // mixins registered from multiple language modules — Kotlin's
+                            // mixin extends KotlinIsoVisitor, a (hypothetical) Groovy mixin
+                            // would extend GroovyVisitor, etc. Only one can compose with
+                            // the current adaptTo target; skip mixins that aren't a subtype
+                            // so dispatch for other languages falls through to the base.
+                            if (!adaptTo.isAssignableFrom(mixinClass)) {
+                                continue;
+                            }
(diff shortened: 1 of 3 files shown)


Jenson3210:
Looks like we've both been working on the mixin downstream. 
- https://github.com/openrewrite/rewrite/pull/7720

I'll leave it up to you to determine which is the best solution

timtebeek:
Thanks! Yours is more succinct; let's go for that one.

PR #7719, opened 2026-05-18
Add rewrite-java-next module for non-LTS JDK support

timtebeek:
## Summary

- Adds a new `rewrite-java-next` module copied from `rewrite-java-25`, intended as a moving target for the latest non-LTS JDK releases — initially targeting JDK 26, will advance to 27/28 as those ship, then be renamed to `rewrite-java-29` once 29 LTS lands.
- Bundles the JDK 26 parser fixes from #7716 and #7717 into this new module instead of `rewrite-java-25`: implicit lambda parameter type (zero-width `JCErroneous` vartype), single-element annotation `value=` shorthand (zero-width `JCAssign`), and enum-constant synthesized class identifier.
- `JavaParser` now dispatches to `JavaNextParser` for JDK 26+, falling through to `Java25Parser` for 25 etc.
- Three new test classes (`Java26ImplicitLambdaParameterTest`, `Java26AnnotationValueTest`, `Java26EnumConstructorTest`) cover the JDK 26 regressions referenced in #7554.

- Closes #7716
- Closes #7717

## Test plan

- [ ] `./gradlew :rewrite-java-next:test :rewrite-java-next:compatibilityTest` on a JDK 26 toolchain
- [ ] `./gradlew :rewrite-java-25:test :rewrite-java-25:compatibilityTest` still green on JDK 25 (no regression in the LTS module)
- [ ] `./gradlew :rewrite-java:test` validates the dispatcher change in `JavaParser`


timtebeek:
This finally adopts the style discussed with Shannon when we started rewrite-java-25 and dropped the backports that made the parsers hard to reason about and maintain. rewrite-java-next would be used for projects running on Java 26 or newer, and continuously updated as new versions come out. We still need the matching changes downstream in the Maven and Gradle plugins and the CLI.

timtebeek:
## Parser visitor diff: `ReloadableJavaNextParserVisitor` vs `ReloadableJava25ParserVisitor`

_Updated after diagnosing the final failures locally on JDK 26._

Of the 10 source files in `rewrite-java-next/src/main`, 9 are byte-identical to their `rewrite-java-25` counterparts after the `Java25 → JavaNext` rename. Only `ReloadableJavaNextParserVisitor.java` diverges. All divergences exist to handle a single JDK 26 javac behavior change: synthesized AST nodes that previously carried `NOPOS` (-1) source positions now carry **zero-width** positions instead.

### Four call-site changes

**`visitAnnotation`** — single-element annotation shorthand `@Ann("foo")`
```diff
- if (arg instanceof JCAssign) {
-     if (endPos(arg) < 0) {
-         expressions = singletonList(convert(((JCAssign) arg).rhs, t -> sourceBefore(")")));
-     } else {
-         expressions = singletonList(convert(arg, t -> sourceBefore(")")));
-     }
+ if (arg instanceof JCAssign && isSyntheticValueAssignment((JCAssign) arg)) {
+     expressions = singletonList(convert(((JCAssign) arg).rhs, t -> sourceBefore(")")));
  } else {
      expressions = singletonList(convert(arg, t -> sourceBefore(")")));
  }
```

**`visitLiteral`** — synthesized literal values (e.g. propagated annotation arguments on record components)
```diff
  int endPos = endPos(node);
  Object value = node.getValue();
- if (endPos == Position.NOPOS) {
+ int startPos = ((JCLiteral) node).getStartPosition();
+ if (endPos == Position.NOPOS || endPos <= startPos) {
      if (typeMapping.primitive(((JCLiteral) node).typetag) == JavaType.Primitive.String) {
          int quote = source.startsWith("\"\"\"", cursor) ? 3 : 1;
          int elementLength = quote == 3 ? source.indexOf("\"\"\"", cursor + quote) - cursor - quote : value.toString().length();
          endPos = cursor + quote + elementLength + quote;
      } else {
          endPos = indexOf(source, cursor,
                  ch -> Character.isWhitespace(ch) || ",;)]}+-*/%=!<>&|^?:.".indexOf(ch) != -1);
      }
+     startPos = cursor;
  }
  cursor(endPos);
- String valueSource = source.substring(((JCLiteral) node).getStartPosition(), endPos);
+ String valueSource = source.substring(startPos, endPos);
```

**`visitNewClass`** — enum-constant synthesized class identifier
```diff
- // for enum definitions with anonymous class initializers, endPos of node identifier will be -1
- TypeTree clazz = endPos(node.getIdentifier()) >= 0 ? convert(node.getIdentifier()) : null;
+ // For enum constants, the identifier is synthesized: on JDK ≤ 25 with endPos < 0,
+ // on JDK 26+ as a zero-width node at the enum-constant's own name position.
+ JCTree ident = (JCTree) node.getIdentifier();
+ TypeTree clazz = endPos(ident) > ident.getStartPosition() ? convert(node.getIdentifier()) : null;
```

**`visitVariables`** — `var`/`val` local declarations and lambda parameters
```diff
- if (vartype == null) {
+ if (vartype == null || isImplicitLambdaParameterType(node, vartype)) {
      typeExpr = null;
- } else if (endPos(vartype) < 0) {
-     if ((node.sym.flags() & Flags.PARAMETER) > 0) {
-         typeExpr = null;
-     } else {
-         Space space = whitespace();
-         boolean lombokVal = source.startsWith("val", cursor);
-         cursor += 3;
-         typeExpr = new J.Identifier(..., lombokVal ? "val" : "var", ...);
-     }
+ } else if (isSynthesizedVarType(vartype)) {
+     // `var` / `val` local declaration (and any non-parameter context where javac
+     // synthesizes the vartype, e.g. record patterns, try-with-resources, unnamed variables).
+     Space space = whitespace();
+     boolean lombokVal = source.startsWith("val", cursor);
+     cursor += 3;
+     typeExpr = new J.Identifier(..., lombokVal ? "val" : "var", ...);
  } else if (vartype instanceof JCArrayTypeTree) { ... }
```

### Three new helper methods

```java
private boolean isSyntheticValueAssignment(JCAssign arg) {
    // Detect via the source itself: a real `value = "x"` shows `=` between
    // the annotation's `(` and the rhs's start; shorthand `("x")` does not.
    // Position-based detection isn't reliable — JDK 26's encoding of the
    // synthesized JCAssign varies by context (zero-width on classes/methods,
    // overlapping with the rhs on record-component annotations).
    int rhsStart = ((JCTree) arg.rhs).getStartPosition();
    if (rhsStart < cursor || rhsStart > source.length()) {
        return endPos(arg) <= arg.getStartPosition();
    }
    return source.substring(cursor, rhsStart).indexOf('=') < 0;
}

private boolean isImplicitLambdaParameterType(JCVariableDecl node, JCExpression vartype) {
    if ((node.sym.flags() & Flags.PARAMETER) == 0) {
        return false;
    }
    return isSynthesizedVarType(vartype);
}

private boolean isSynthesizedVarType(JCExpression vartype) {
    // JDK ≤ 25: synthesized vartype has NOPOS, so endPos == -1.
    // JDK 26+: synthesized vartype is a zero-width node at the variable name's position.
    return endPos(vartype) <= vartype.getStartPosition();
}
```

### Why each fix is needed

| Affected TCK tests on JDK 26 (before fixes) | Cause | Fix |
|---|---|---|
| `AnnotationTest.annotationWithDefaultArgument` | Synthesized `value=` JCAssign no longer has `NOPOS`; old `endPos(arg) < 0` guard missed it, so the parser rendered both the synthesized `value` identifier and the rhs (producing `valueecked"`) | `isSyntheticValueAssignment` — source-text based |
| `RecordTest.{annotationsAndRecords, typeParameterAnnotation, recordWithMultipleAnnotations, differentLiteralTypesAnnotation}` | Propagated JCLiteral values inside record-component annotations carry zero-width ranges at startPos on JDK 26 (not NOPOS, not their real range); `cursor(endPos)` set cursor to its own position, leaving literal text unconsumed in source, and `convertAll` wrapped the otherwise-correct `J.Literal` in `J.Erroneous` whose printed text was the literal itself (hence the misleading `"native"`, `"A"`, `"a"`, `"b"` in failure output) | extended synthesized-literal detection in `visitLiteral`: `endPos == NOPOS \|\| endPos <= startPos` |
| `EnumTest.enumWithParameters` (plus the new `enumConstantWith{MethodCall,StringConcat}Argument` cases added to the TCK) | Synthesized class identifier on enum constants is now zero-width instead of `NOPOS`; old `endPos >= 0` guard let it through and `visitIdentifier` advanced the cursor past the args | `endPos > startPos` instead of `endPos >= 0` |
| `LambdaTest.untypedLambdaParameter` and any other implicit lambda parameter | Synthesized vartype for an implicit lambda parameter is now a zero-width `JCErroneous` at the parameter name; old `endPos(vartype) < 0` didn't fire and `convert(vartype)` failed the cast to `TypeTree` | `isImplicitLambdaParameterType` via `isSynthesizedVarType` |
| `VariableDeclarationsTest.{typeOnVarKeyword, string, finalVar, unknownVar, implicitlyDeclaredLocalVariable}`, `RecordPatternMatchingTest.shouldParseRecordPatternWithVarKeyword`, `TypeCastTest.intersectionCastAssignedToVar`, `TypeUtilsTest.{typeToString2, intersectionTypes}`, `LombokTest.{val, var}`, `UnnamedVariableTest.{unnamedVariableInTryWithResources, unnamedVariableInLocalDeclaration, unnamedVariableInForLoop, unnamedVariableWithSideEffects}` | Same dual-shape issue as the lambda parameter case, but for `var`/`val` local declarations (the original PR fix only covered the `PARAMETER`-flagged branch) | extract `isSynthesizedVarType` and apply it to both branches |

### Build config

`rewrite-java-next/build.gradle.kts` differs from `rewrite-java-25/build.gradle.kts` only in the toolchain bump (JDK 25 → 26) and a `testLogging { events("failed") }` block that surfaces TCK failure detail in CI logs — added because this module is the only one exercising the TCK on the latest non-LTS JDK and the HTML report isn't preserved as a CI artifact.

shanman190 (approved):

timtebeek:
It was decided to remove this again due to friction reported downstream in adopting Java 26
- https://github.com/openrewrite/rewrite/pull/7764

If anyone wants to restore the fixes we did towards Java 26, and run those in the plugins, look at these PRs:
- https://github.com/openrewrite/rewrite-maven-plugin/pull/1155
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/446

PR #7769, opened 2026-05-22
Default whitespace for keyword-prefixed JContainers in withX(...) mutators

steve-aom-elliott:
## Summary

When a recipe calls `withThrows(...)`, `withImplements(...)`, `withPermits(...)`, `TypeParameter.withBounds(...)`, or `Try.withResources(...)` on a node that previously had no such clause, the resulting LST today jams the keyword against the surrounding tokens because:

1. `JContainer.withElementsNullable(null, list)` builds the new container with `Space.EMPTY` as its leading space, so the space between the preceding token and the keyword is lost.
2. The first element keeps whatever prefix the caller provided — and callers reasonably expect the API to handle the space between the keyword and the first element.

Recipes work around both today by setting `Space.SINGLE_SPACE` on the first element manually and then running `maybeAutoFormat` to repair the leading space. That works, but it leaks LST-printer mechanics into every caller.

## Change

Adds a four-argument overload:

```java
JContainer.withElementsNullable(before, elements, defaultBefore, defaultFirstElementPrefix)
```

When the container is newly built (`before == null` and `elements` non-empty), `defaultBefore` becomes the container's leading space and `defaultFirstElementPrefix` becomes the first element's prefix **only when that prefix is currently empty**. Intentional non-empty prefixes (e.g. newline indentation for wrapping) are preserved. Existing containers are passed through untouched.

The existing two-argument `withElementsNullable(before, elements)` now delegates to the four-argument form with `(EMPTY, EMPTY)` — strict no-op for every other caller.

Five keyword-prefixed `withX` mutators in `J.java` now declare their own conventions:

| `withX` method | `defaultBefore` | `defaultFirstElementPrefix` |
|---|---|---|
| `MethodDeclaration.withThrows` | `SINGLE_SPACE` | `SINGLE_SPACE` |
| `ClassDeclaration.withImplements` | `SINGLE_SPACE` | `SINGLE_SPACE` |
| `ClassDeclaration.withPermits` | `SINGLE_SPACE` | `SINGLE_SPACE` |
| `TypeParameter.withBounds` | `SINGLE_SPACE` | `SINGLE_SPACE` |
| `Try.withResources` | `SINGLE_SPACE` | `EMPTY` |

`try (...)` is the only case where the two slots differ: `try` needs a space after it, but the `(` sits flush against the first resource. Empirically verified against parser output for all five containers.

## Test plan

- [x] New `AddKeywordContainerSpacingTest` — five end-to-end tests, one per `withX` mutator, each creating elements with `Space.EMPTY` prefixes and verifying the API produces correctly-spaced source (`throws IOException`, `implements Serializable`, `permits Circle`, `try (AutoCloseable r = null)`, `<T extends Comparable>`).
- [x] Extended `JContainerTest` with seven new unit tests covering the new overload: defaults, normalization on empty prefix, preservation of intentional non-empty prefix, `(SINGLE_SPACE, EMPTY)` shape for try-resources, and pass-through when an existing container is provided.
- [x] `./gradlew :rewrite-java:test` passes.


rewrite-java/src/main/java/org/openrewrite/java/tree/JContainer.java
```diff
@@ -161,11 +161,28 @@ public JContainer<T> withElements(List<JRightPadded<T>> elements) {
...
+
+    /**
+     * Replace or build the elements of a {@link JContainer}. When the container is newly built
+     * ({@code before} is {@code null} and {@code elements} is non-empty), {@code defaultBefore}
+     * becomes its leading {@link Space} (printed before the delimiter or keyword), and
+     * {@code defaultFirstElementPrefix} becomes the first element's prefix (printed between the
+     * delimiter/keyword and the first element) if that prefix is currently empty. Existing
+     * containers are untouched; only the elements are replaced.
+     */
+    public static <J2 extends J> @Nullable JContainer<J2> withElementsNullable(@Nullable JContainer<J2> before, @Nullable List<J2> elements, Space defaultBefore, Space defaultFirstElementPrefix) {
```
timtebeek: Quick note here that while additive, it does mean that the moment any recipe uses this method from having been compiled against a newer rewrite-java snapshot, that it's then likely no longer able to be run from an older more slowly updated CLI version.

Could that be reason enough to potentially handle the addition in each of the wither methods above instead?

steve-aom-elliott: You mean to patch the result of the `JContainer.withElementsNullable` in the wither methods?

steve-aom-elliott: Ah, I think you're just meaning because the method signature itself wouldn't exist. Hmm, fair I suppose. I can pull them upwards I guess. Was just kind of hoping to reduce duplication more so originally.

timtebeek: Yeah under normal circumstanced I'd applaud that, but since we're working with decoupled adoption rates here I'd start with inlined, and perhaps already expose this added method (but don't use it yet), such that we can later switch over. That latter part is optional.

steve-aom-elliott: I've pulled the new logic up into the withers now and dropped the additional signture `withElementsNullable(..)`. Did what I could to establish a clear flow for mutations rather than pulling the lower level `JContainer.build(..)` logic up as well.

rewrite-java/src/test/java/org/openrewrite/java/tree/AddKeywordContainerSpacingTest.java
```diff
@@ -0,0 +1,241 @@
...
+ * previously did not have one. Each container's leading {@link Space} must default to
+ * {@link Space#SINGLE_SPACE} so the keyword is separated from the preceding token.
+ */
+class AddKeywordContainerSpacingTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        // Reused across tests; each test sets its own recipe.
+    }
+
```
timtebeek:
```suggestion
```

timtebeek (approved):

PR #7779, opened 2026-05-26
Fix AddCommentToXmlTag for self-closing and empty tags

### LINKED ISSUE #7777: AddCommentToXmlTag doesn't handle self-closing or empty tags
## What

`org.openrewrite.xml.AddCommentToXmlTag` (in `rewrite-xml`) has two bugs that make it unusable for any element typically written in self-closing form:

1. **Self-closing tags are silently skipped.** The visitor short-circuits when `tag.getContent() == null`, so `<foo bar="baz"/>` is never modified — no comment, no error.
2. **Empty tags throw.** For `<foo></foo>` (or `<foo>\n</foo>` with whitespace-only content), `tag.getContent()` returns an empty list and the recipe crashes with `IndexOutOfBoundsException: Index 0 out of bounds for length 0` at the `contents.get(insertPos).getPrefix()` call.

Both forms are extremely common in real-world XML — e.g. Spring Integration configurations (`<int-jdbc:message-store id="x" data-source="ds"/>`) and any element that has no required child content.

## Where

[`org/openrewrite/xml/AddCommentToXmlTag.java`](https://github.com/openrewrite/rewrite/blob/main/rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java) — the visitor inside `getVisitor()`:

```java
@Override
    Xml.Tag t = (Xml.Tag) super.visitTag(tag, ctx);
    if (matcher.matches(getCursor())) {
        if (tag.getContent() != null) {                         // (1) skips self-closing
            List<Content> contents = new ArrayList<>(tag.getContent());
            boolean containsComment = contents.stream()
                    .anyMatch(c -> c instanceof Xml.Comment &&
                            commentText.equals(((Xml.Comment) c).getText()));
            if (!containsComment) {
                int insertPos = 0;
                Xml.Comment customComment = new Xml.Comment(randomId(),
                        contents.get(insertPos).getPrefix(),    // (2) crashes when empty
                        Markers.EMPTY,
                        commentText);
                contents.add(insertPos, customComment);
                t = t.withContent(contents);
            }
        }
    }
    return t;
}
```

## Repro

```java
class Repro implements RewriteTest {
    @Test
    void selfClosingTagIsSkipped() {
        rewriteRun(
          spec -> spec.recipe(new AddCommentToXmlTag("//foo", " hello ")),
          xml(
            "<root><foo bar=\"baz\"/></root>",
            "<root><foo bar=\"baz\"><!-- hello -->\n</foo></root>"  // expected
          )
        );
        // Actual: no change is made.
    }

    @Test
    void emptyTagThrows() {
        rewriteRun(
          spec -> spec.recipe(new AddCommentToXmlTag("//foo", " hello ")),
          xml("<root><foo></foo></root>")
        );
        // Actual: IndexOutOfBoundsException: Index 0 out of bounds for length 0
    }
}
```

## Proposed fix

`Xml.Tag#withContent(List)` already materializes a `Closing` tag when transitioning a self-closing tag to one that has content, so the fix is small: stop bailing on `null`/empty content, derive the comment prefix from the existing first child (or `""` when there is none), and always go through `withContent`. The visitor body becomes:

```java
@Override
    Xml.Tag t = (Xml.Tag) super.visitTag(tag, ctx);
    if (!matcher.matches(getCursor())) {
        return t;
    }
    List<? extends Content> existing = t.getContent();
    if (existing != null) {
        for (Content c : existing) {
            if (c instanceof Xml.Comment && commentText.equals(((Xml.Comment) c).getText())) {
                return t;                       // idempotency preserved
            }
        }
    }
    String prefix = (existing == null || existing.isEmpty()) ? "" : existing.get(0).getPrefix();
    Xml.Comment comment = new Xml.Comment(randomId(), prefix, Markers.EMPTY, commentText);
    List<Content> contents;
    if (existing == null || existing.isEmpty()) {
        contents = new ArrayList<>(Collections.singletonList(comment));
    } else {
        contents = new ArrayList<>(existing.size() + 1);
        contents.add(comment);
        contents.addAll(existing);
    }
    return t.withContent(contents);
}
```

Behavior summary after the fix:

| Tag form | Before | After |
|---|---|---|
| `<foo>existing content</foo>` | comment prepended ✓ | unchanged behavior ✓ |
| `<foo/>` (self-closing) | silently skipped ✗ | comment added, closing tag materialized ✓ |
| `<foo></foo>` (empty) | `IndexOutOfBoundsException` ✗ | comment added ✓ |
| same comment text twice | not re-added ✓ | unchanged behavior ✓ |

Happy to follow up with a PR that includes this change plus the three new test cases.

Jenson3210:
## Summary
- Resolves #7777: `AddCommentToXmlTag` silently skipped self-closing tags (`<foo/>`) and threw `IndexOutOfBoundsException` on empty tags (`<foo></foo>`).
- Visitor now derives the comment prefix from the existing first child (or `""` when content is null/empty) and always routes through `Xml.Tag#withContent`, which already materializes a `Closing` tag when transitioning a self-closing tag to one with content.
- Also fixes a latent bug where the previous implementation mixed `tag.getContent()` (pre-recursion) with `t.withContent(...)`, dropping nested transformations.

## Test plan
- [x] New test: comment is added to a self-closing tag (`<foo bar="baz"/>` → `<foo bar="baz"><!-- hello --></foo>`).
- [x] New test: comment is added to an empty tag (`<foo></foo>` → `<foo><!-- hello --></foo>`) with no exception.
- [x] New test: idempotency — recipe does not duplicate an existing matching comment.
- [x] Existing `addCommentToDependencyBlock` test still passes.
- [x] `./gradlew :rewrite-xml:test --tests "org.openrewrite.xml.AddCommentToXmlTagTest"` → BUILD SUCCESSFUL.

### DIFF
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java
@@ -54,24 +54,29 @@ public class AddCommentToXmlTag extends Recipe {
             @Override
             public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 Xml.Tag t = (Xml.Tag) super.visitTag(tag, ctx);
-                if (matcher.matches(getCursor())) {
-                    if (tag.getContent() != null) {
-                        List<Content> contents = new ArrayList<>(tag.getContent());
-                        boolean containsComment = contents.stream()
-                                .anyMatch(c -> c instanceof Xml.Comment &&
-                                        commentText.equals(((Xml.Comment) c).getText()));
-                        if (!containsComment) {
-                            int insertPos = 0;
-                            Xml.Comment customComment = new Xml.Comment(randomId(),
-                                    contents.get(insertPos).getPrefix(),
-                                    Markers.EMPTY,
-                                    commentText);
-                            contents.add(insertPos, customComment);
-                            t = t.withContent(contents);
+                if (!matcher.matches(getCursor())) {
+                    return t;
+                }
+                List<? extends Content> existing = t.getContent();
+                if (existing != null) {
+                    for (Content c : existing) {
+                        if (c instanceof Xml.Comment && commentText.equals(((Xml.Comment) c).getText())) {
+                            return t;
                         }
                     }
                 }
-                return t;
+                String prefix = (existing == null || existing.isEmpty()) ? "" : existing.get(0).getPrefix();
+                Xml.Comment comment = new Xml.Comment(randomId(), prefix, Markers.EMPTY, commentText);
+                List<Content> contents;
+                if (existing == null || existing.isEmpty()) {
+                    contents = new ArrayList<>();
+                    contents.add(comment);
+                } else {
+                    contents = new ArrayList<>(existing.size() + 1);
+                    contents.add(comment);
+                    contents.addAll(existing);
+                }
+                return t.withContent(contents);
             }
         };
     }
--- a/rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java
+++ b/rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java
@@ -18,13 +18,15 @@ package org.openrewrite.xml;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.marker.Markers;
 import org.openrewrite.xml.tree.Content;
 import org.openrewrite.xml.tree.Xml;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
+import static java.util.Collections.singletonList;
 import static org.openrewrite.Tree.randomId;
 
 @Incubating(since = "7.24.0")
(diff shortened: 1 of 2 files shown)


greg-at-moderne (approved):

rewrite-xml/src/test/java/org/openrewrite/xml/AddCommentToXmlTagTest.java
```diff
@@ -81,4 +81,36 @@ void addCommentToDependencyBlock() {
...
         );
     }
+
+    @Test
+    void addCommentToSelfClosingTag() {
+        rewriteRun(
+          spec -> spec.recipe(new AddCommentToXmlTag("//foo", " hello ")),
+          xml(
+            "<root><foo bar=\"baz\"/></root>",
+            "<root><foo bar=\"baz\"><!-- hello --></foo></root>"
```
timtebeek: I wonder if we should prepend the comment to the self closing tag instead of adding a new close tag to add the comment within the tag. What are your thoughts on that?

My motivation would be a more minimal diff, with a clear line introduced before the existing line in a well formatted XML document.

Jenson3210: I also preferred that, however, that is a bigger change as there is already expectations of comments being added inside the tags. 
Would you want to have different behavior when tag is empty vs non-empty? 
https://github.com/openrewrite/rewrite/blob/20470430d67047057681878f6899977c62885f25/rewrite-xml/src/test/java/org/openrewrite/xml/AddCommentToXmlTagTest.java#L29

timtebeek: I'd be fine with slightly different behavior if that leads to less surprises for the self closing case.

rewrite-xml/src/main/java/org/openrewrite/xml/AddCommentToXmlTag.java
```diff
@@ -54,25 +54,72 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                    String prefix = (existing == null || existing.isEmpty()) ? "" : existing.get(0).getPrefix();
+                    Xml.Comment comment = new Xml.Comment(randomId(), prefix, Markers.EMPTY, commentText);
+                    List<Content> contents;
+                    if (existing == null || existing.isEmpty()) {
+                        contents = new ArrayList<>();
+                        contents.add(comment);
+                    } else {
+                        contents = new ArrayList<>(existing.size() + 1);
+                        contents.add(comment);
+                        contents.addAll(existing);
```
Jenson3210: Replace with ListUtils?

rewrite-xml/src/test/java/org/openrewrite/xml/AddCommentToXmlTagTest.java
Jenson3210: implemented this change here.

timtebeek (approved):

PR #7810, opened 2026-05-28
Gradle: sync `version { strictly(...) }` constraints when upgrading dependencies

Jenson3210:
## Summary

- `UpgradeDependencyVersion` (and any other caller of `GradleDependency.withDeclaredVersion`) updated only the outer dependency notation. Any rich version constraint inside the trailing configuration closure — `version { strictly('…') }`, `require('…')`, or `prefer('…')` — kept pointing at the old version, producing Gradle resolution conflicts:

  ```gradle
  implementation('org.springframework:spring-context:6.2.18') {
      version { strictly('5.3.18') }  // ← left at old version
  }
  ```

- Extend `GradleDependency.withDeclaredVersion` to walk the resulting `J.MethodInvocation` and rewrite the literal argument of any `strictly` / `require` / `prefer` call whose cursor is enclosed by a `version { }` block. The check reuses the existing `withinBlock(cursor, "version")` helper and `ChangeStringLiteral.withStringValue` to stay consistent with the rest of the trait.

- `reject(...)` is intentionally left untouched — it lists versions to *exclude*, so auto-rewriting to the new target is not safe.

- Fixes [moderneinc/customer-requests#2454](https://github.com/moderneinc/customer-requests/issues/2454) (TIAA).

## Test plan

- [x] New tests in `UpgradeDependencyVersionTest`:
  - [x] `strictly` constraint in a Groovy DSL closure
  - [x] `require` + `prefer` constraints in a Groovy DSL closure
  - [x] multi-component `group:/name:/version:` notation with `strictly`
  - [x] Kotlin DSL with `strictly`
  - [x] negative: unrelated dependency's `strictly` block left alone
- [x] `./gradlew :rewrite-gradle:test` — BUILD SUCCESSFUL, no regressions.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependency.java
@@ -1375,6 +1375,28 @@ public class GradleDependency implements Trait<J.MethodInvocation> {
             }
         }
 
+        // Sync rich version constraints inside any trailing `version { ... }` closure
+        // so that calls like `strictly(...)`, `require(...)`, `prefer(...)` reference the new version.
+        J.MethodInvocation withSyncedConstraints = (J.MethodInvocation) new JavaVisitor<Integer>() {
+            @Override
+            public J visitMethodInvocation(J.MethodInvocation method, Integer i) {
+                J.MethodInvocation mi = (J.MethodInvocation) super.visitMethodInvocation(method, i);
+                String name = mi.getSimpleName();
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Makes sense to me; not sure if I've missed any nuance, so insights from Shannon would still be welcome.

shanman190 (approved):
This works nicely.

One subtlety that we don't need to worry about here right now I don't think is that a dependency declaration with just a prefer is redundant. It's almost as if that should get cleaned up by some form of "remove redundant dependency delcarations"-type of task. But again, not part of this particular PR or issue.

PR #7819, opened 2026-05-29
Add recipeInstallDir to CSharpRewriteRpc so callers control the recipe project location

shanman190:
## Problem

The C# RPC server always creates its recipe project (the `Recipes.csproj` against which `dotnet add package` resolves recipe NuGet packages) under a private temp directory. Because dotnet discovers `NuGet.config` by walking up from the **project directory**, a caller (e.g. the Moderne CLI) had no way to inject a package feed via a generated `NuGet.config` — there was nowhere to put it that dotnet would find.

## Solution

Bring the C# RPC up to parity with the JavaScript, Python, and Go RPC servers, which already expose `recipeInstallDir` / `--recipe-install-dir`. The C# server now creates `Recipes.csproj` in a caller-supplied directory when one is provided (otherwise the previous temp-dir behavior is unchanged), so a `NuGet.config` co-located in that directory is honored by dotnet's project-directory config discovery.

- `CSharpRewriteRpc.Builder.recipeInstallDir(Path)` appends `--recipe-install-dir=` to the server command (tool-path, csproj, and explicit-entry paths)
- `Program.cs` parses `--recipe-install-dir=` and threads it through `RunAsync`
- `RewriteRpcServer` creates the recipe project under that directory when set

The change is additive and backward-compatible: the new constructor/builder/RunAsync parameters default to the prior behavior.

## Context

Unblocks the Moderne CLI `mod config recipes nuget install` feature so it can resolve recipes from a configured artifact repository (the CLI writes a `NuGet.config` into the install dir). npm/pip/go deliver their configured feed via environment variables and need no server change; nuget is file-based, hence this hook.

## Testing

- C# tool/server and Java `:rewrite-csharp:compileJava` build clean.
- C# unit tests pass. The `CsprojParseTests` RPC integration tests require `RPC_TEST_SERVER_CLASSPATH` (or `./gradlew :rewrite-csharp:rpcTestClasspath`) and fail identically on `main` in a bare checkout — unrelated to this change.

### DIFF
--- a/rewrite-csharp/csharp/OpenRewrite.Tests/Rpc/RecipesNuGetConfigTests.cs
+++ b/rewrite-csharp/csharp/OpenRewrite.Tests/Rpc/RecipesNuGetConfigTests.cs
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using OpenRewrite.CSharp.Rpc;
+
+namespace OpenRewrite.Tests.Rpc;
+
+/// <summary>
+/// When a local NuGet feed (<c>~/.nuget/local-feed</c>) exists for cross-repo
+/// development, the server makes it additive to the recipe project's
+/// <c>nuget.config</c>: it creates one (public + local feed) when none exists,
+/// and otherwise only appends the local feed — preserving a caller-written
+/// config (e.g. an exclusive configured feed) untouched.
+/// </summary>
+public class RecipesNuGetConfigTests
+{
+    private const string LocalFeed = "/home/dev/.nuget/local-feed";
+
+    [Fact]
+    public void CreatesPublicPlusLocalFeedWhenNoConfigExists()
+    {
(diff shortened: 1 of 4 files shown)


okundzich:
what was the user visible effect prior to this change? What was broken and fixed with this? We have MS build demos going on next week, so would like to inform the team going there if something has materially changed.

shanman190:
This is needed to help support the CLI when installing recipes from private sources only. We need to be able to apply some changes to the temporary install project that is created for each recipe install, but unlike all of the other RPC implementations, C# was missing this mechanism.

macsux:
i see that we added the extra argument to the rpc class on java side to take in this optional path, but how is it getting in there? i don't see anything that would populate it neither from mod cli argument nor env var, so i don't think it's actually possible to realistically for user to inject this value.

shanman190:
I've got a local CLI change that makes use of it, but I haven't yet pushed that one as I was still testing with it. The premise of the change though is to support configuring private repository sources for the C# ecosystem.

macsux:
I think my other PR is probably better implementation (and simpler). Nuget sources are generally configured via nuget.config files. The existing code was already problematic because it stuffed nuget.org in there. We already saw this fail in places like BOFA where they don't allow direct connections to nuget.org, and instead rely on internal artifactory proxy - access to which is globally configured on machine via `~/.nuget/NuGet/nuget.config` and automatically picked up by all dotnet operations. We simply need to "augment it" with additional folder based source for when they can't put things in artifactory for whatever reason and instead place it on local drive. Realistically this can be accomplished by just editing the above mentioned global nuget.config. This is just a helper override that's light weight, which can be set once and not having to remember to pass it to every mod invocation.

It also may contain packages that are not just recipes, like the OpenRewrite.Recipes.Tool which is the harness used to run the actual recipes (but is distributed as nuget package itself), so the name `recipe-install-dir` is a bit misleading.

Finally I don't think this PR addresses a scenario where current workdir (or ancestor) provides a nuget.config file that as a `<clear/>` in it, which cancels out all other sources (including machine configured ones) and causes failure. I flagged this on a few failing repos. My PR ensures that workdir as altered to something else before nuget operations are attempted so any repo level nuget.config files don't interfere with loading our packages.

sambsnyd (approved):

jkschneider (approved):

PR #7833, opened 2026-05-29
Add optional `ecosystem` parameter to `ChangeType`

Jenson3210:
- > **Draft for discussion** — opening this to gather maintainer feedback on the approach before polishing. Resolves moderneinc/customer-requests#2463.

## Problem

`org.openrewrite.java.ChangeType` applies to every `JavaSourceFile`:

```java
public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
    return sourceFile instanceof JavaSourceFile || sourceFile instanceof SourceFileWithReferences;
}
```

Python and JavaScript reuse the Java LST model — `Py.CompilationUnit` and `JS.CompilationUnit` both implement `JavaSourceFile`. So a JVM-targeted `ChangeType` accepts every Python/JS file and runs its `UsesType` precondition, walking the entire AST looking for a JVM fully-qualified type that can never appear there.

In telemetry from a single recipe run (see issue for the full table), `ChangeType` spent **423 s** on `finos/symphony-bdk-python` — which has **zero** Java files — and changed nothing. On Python-heavy repos this is enough to push recipe runs past the timeout.

## Proposal

Add an optional `ecosystem` option (`jvm`, `python`, `javascript`) that scopes the change to a single ecosystem:

- **Default (`null`)** — unchanged behavior: the recipe runs on all source files.
- When set, files whose compilation unit belongs to a different ecosystem are skipped before the `UsesType` scan.
- Reference-based files (XML, properties, …) are always processed, since they may carry type references.

The new constructor is annotated `@JsonCreator`; the prior 3-arg constructor is preserved for backward compatibility.

## Open questions for discussion

- **Detection mechanism** — the current draft infers a file's ecosystem from its implementing class package (`org.openrewrite.python.*`, `org.openrewrite.javascript.*`, else `jvm`). Is there a cleaner / more robust signal we'd prefer (e.g. a marker or a `SourceFile`-level capability)?
- **Scope** — the issue notes this affects *other* JVM type/usage recipes too (`ChangeMethodName`, `UsesType`-based recipes, …). Should this live on `ChangeType` alone, or be lifted into a shared precondition/utility so every affected recipe benefits?
- **Option shape** — single `ecosystem` string vs. a more general include/exclude list? Naming (`jvm` vs `java`)?

## Changes

- `ChangeType`: new optional `ecosystem` option + ecosystem-skip logic, backward-compatible constructor.
- Tests in `ChangeTypeTest` (jvm applies / non-jvm skips Java files) and `ChangeTypeIntegTest` (jvm scope skips Python files).

## Possible follow-up: roll this out broadly via a best-practices recipe

The per-recipe option above is only worthwhile if it actually gets *set* across the large meta-recipes that drive the cost (Spring Boot upgrades, `migrate-java`, etc.), where hundreds of `ChangeType` steps each scan every Python/JS file.

Rather than hand-editing each declarative recipe, we could add a recipe to **rewrite best practices** that infers the right `ecosystem` for an existing `ChangeType` step from indicators already present in the declarative YAML — e.g. the package of the changed class (`org.springframework.*`, `javax.*`, … → `jvm`) — and fills in the `ecosystem` option automatically.

Run across `rewrite-migrate-java`, `rewrite-spring`, and the other recipe libraries, that would scope every `ChangeType` (and ideally the other `UsesType`-based steps) to `jvm` in one sweep — turning this from a single-recipe opt-in into a broad win, where the biggest savings land precisely on the Python/JS-heavy repos that pay the most today.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
@@ -56,6 +57,18 @@ public class ChangeType extends Recipe {
     @Nullable
     Boolean ignoreDefinition;
 
+    @Option(displayName = "Ecosystem",
+            description = "Restrict the type change to source files of a given ecosystem: `jvm` (Java, Kotlin, " +
+                    "Groovy), `python`, or `javascript`. Python and JavaScript reuse the Java LST model (their " +
+                    "compilation units implement `JavaSourceFile`), so without this filter a change targeting one " +
+                    "ecosystem still scans the others' source files unnecessarily. Reference-based files (such as " +
+                    "XML and properties) are always processed because they may carry type references. When `null` " +
+                    "(the default), the change is applied to all source files.",
+            valid = {"jvm", "python", "javascript"},
+            required = false)
+    @Nullable
+    String ecosystem;
+
     String displayName = "Change type";
 
     @Override
@@ -82,6 +95,46 @@ public class ChangeType extends Recipe {
 
     String description = "Change a given type to another.";
 
+    public ChangeType(String oldFullyQualifiedTypeName, String newFullyQualifiedTypeName, @Nullable Boolean ignoreDefinition) {
+        this(oldFullyQualifiedTypeName, newFullyQualifiedTypeName, ignoreDefinition, null);
+    }
+
+    @JsonCreator
+    public ChangeType(String oldFullyQualifiedTypeName, String newFullyQualifiedTypeName, @Nullable Boolean ignoreDefinition, @Nullable String ecosystem) {
+        this.oldFullyQualifiedTypeName = oldFullyQualifiedTypeName;
+        this.newFullyQualifiedTypeName = newFullyQualifiedTypeName;
+        this.ignoreDefinition = ignoreDefinition;
+        this.ecosystem = ecosystem;
+    }
+
+    private boolean skipForEcosystem(JavaSourceFile cu) {
+        if (ecosystem == null || ecosystem.isEmpty()) {
+            return false;
+        }
+        String impl = cu.getClass().getName();
+        String fileEcosystem;
+        if (impl.startsWith("org.openrewrite.python.")) {
+            fileEcosystem = "python";
+        } else if (impl.startsWith("org.openrewrite.javascript.")) {
+            fileEcosystem = "javascript";
+        } else {
+            // java, kotlin, groovy, scala, ...
+            fileEcosystem = "jvm";
+        }
+        return !normalizeEcosystem(ecosystem).equals(fileEcosystem);
+    }
+
+    private static String normalizeEcosystem(String ecosystem) {
+        switch (ecosystem.toLowerCase()) {
+            case "js":
+                return "javascript";
+            case "py":
+                return "python";
+            default:
+                return ecosystem.toLowerCase();
+        }
+    }
+
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         TreeVisitor<?, ExecutionContext> condition = new TreeVisitor<Tree, ExecutionContext>() {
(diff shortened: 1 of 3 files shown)


rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -82,6 +95,46 @@ public String getInstanceNameSuffix() {
...
+    private boolean skipForEcosystem(JavaSourceFile cu) {
+        if (ecosystem == null || ecosystem.isEmpty()) {
+            return false;
+        }
+        String impl = cu.getClass().getName();
+        String fileEcosystem;
+        if (impl.startsWith("org.openrewrite.python.")) {
+            fileEcosystem = "python";
+        } else if (impl.startsWith("org.openrewrite.javascript.")) {
+            fileEcosystem = "javascript";
```
steve-aom-elliott, line 119: Would we not be able to use `cu instanceof Py.CompilationUnit` or `cu instanceof JS.CompilationUnit` instead here?

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -82,6 +95,46 @@ public String getInstanceNameSuffix() {
...
+        }
+        return !normalizeEcosystem(ecosystem).equals(fileEcosystem);
+    }
+
+    private static String normalizeEcosystem(String ecosystem) {
+        switch (ecosystem.toLowerCase()) {
+            case "js":
+                return "javascript";
+            case "py":
+                return "python";
```
steve-aom-elliott, line 132: Shouldn't the alternatives also be in the valid set at that point?

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
Jenson3210: We do not have a dependency here in rewrite-python / js. And that would introduce circular dependency.

steve-aom-elliott:
I think really a core of the problem is that yes, this really would affect all the `UsesType` scenarios as well, as any of those are still going to traverse all those files. When we have a change on a recipe like this, yeah, in order to actually optimize, there'd be a lot of other changes.

One suggestion I saw proposed was related to having a `Language` property on SourceFile (could be provided via overridden method on the various source file types) that would allow you to detect which language the file represents without needing to do package matching. That of course only would help so much and wouldn't be backwards compatible either.

Additionally the posibility of having an applicable languages property on a recipe generically might allow you to scope things through that rather than having to pass it as an option: declarative recipes could be scoped in yaml and imperative could be scoped via overriding a method. That too changes the exposed api for a recipe, which gets messy in different ways.

Not sure how I feel about all these yet, as it's a not-insubstantial change for any path here.

sambsnyd (commented):
With the new V3 LST format it's going to be really fast to check whether or not a given source file has a given type - that's raised to an index that can be read without even reading the full LST into memory. So I don't think we'll need anything in particular to make `ChangeType` performant.

Jenson3210:
@sambsnyd I do not see how this will fix the issue.
ChangeType will still be part of tons of Java only recipe chains. 
The python files / markers will still pass these checks causing the lsts to be loaded for every run. 
The goal with ecosystem would be to be able to in the `isAcceptable` check that fast marker for the recipes that should act on python, js, java and for these ChangeType invocations that know they only need to do java changes skipp these files altogether. 

ChangeType does change Python types also, so we need to find a way to declare those changetype invocations inside rewrite-migrate-java to tell the python files to be skipped or they will pass this check you mention.

Jenson3210:
@timtebeek I think this draft is best kept open for discussion / showing the issue.

timtebeek:
This PR might do quite a bit to improve performance already without the need for an ecosystem filter
- https://github.com/openrewrite/rewrite/pull/8052

Note that it only affects the traversal we used to do to find declarations, but since that was executed anywhere the precomputed used types didn't match it had quite an impact when considering other language files.

I'd hope that the change in #8052 removes the need for the changes here, as there's quite a bit more impact in having to add an additional argument to every ChangeType.

I'd be inclined to close this PR already, also given the further discussion above, but I'll hold of until at least one of you agrees with me by closing this PR.

steve-aom-elliott:
> This PR might do quite a bit to improve performance already without the need for an ecosystem filter
> 
>     * [Speed up `ChangeType` precondition by gating the definition scan #8052](https://github.com/openrewrite/rewrite/pull/8052)
> 
> 
> Note that it only affects the traversal we used to do to find declarations, but since that was executed anywhere the precomputed used types didn't match it had quite an impact when considering other language files.
> 
> I'd hope that the change in #8052 removes the need for the changes here, as there's quite a bit more impact in having to add an additional argument to every ChangeType.
> 
> I'd be inclined to close this PR already, also given the further discussion above, but I'll hold of until at least one of you agrees with me by closing this PR.

Took a look over Knut's and Jente's PRs and I think I would agree that with Knut's change, we can probably close this PR for now.

PR #7834, opened 2026-05-29
xml: extend `RemoveEmptyXmlTags` with XPath whitelist, file matcher, and file deletion

Jenson3210:
## Summary

- Extends `org.openrewrite.xml.RemoveEmptyXmlTags` with three optional, JSON-deserializable options so it can serve as the generic cleanup recipe described in [moderneinc/customer-requests#2460 §11](https://github.com/moderneinc/customer-requests/issues/2460):

- **`xPaths`** — whitelist of XPath expressions identifying which empty tags are eligible for removal. Attribute predicates are reinterpreted as an *attribute allowlist*: a candidate tag matches only when its attribute set is a subset of the union of attribute names appearing in the matched predicates. `@*` is the wildcard, `@name` enumerates a single allowed name, and `@a or @b` allows either. So `/server` only matches attribute-free `<server>` tags, while `/server[@*]` matches any, `/server[@description]` matches no-attribute or description-only servers, and `/server[@description or @other]` matches when attributes are a subset of `{description, other}`.
- **`fileMatcher`** — glob restricting which files the recipe touches (wired via `Preconditions.check(new FindSourceFiles(...), …)`).
- **`deleteFileIfEmpty`** — when the root tag itself would be a removal candidate, delete the entire source file. Defaults to `true`.

The visitor is wrapped in `Repeat.repeatUntilStable(..., 50)` so nested empties cascade and trigger file deletion in a single recipe run. The original no-arg constructor is preserved via Lombok `@NoArgsConstructor(force = true)`, and the explicit all-args constructor is `@JsonCreator`-annotated. Existing callers and tests are unaffected — the defaults reproduce the prior behavior on documents whose root retains content.

Motivating example (Liberty `mp-telemetry.xml` cleanup once another recipe strips `<feature>mpTelemetry-2.0</feature>`):

```java
new RemoveEmptyXmlTags(
    List.of("/server/featureManager", "/server[@*]"),
    "**/mp-telemetry.xml",
    true)
```

collapses the now-empty `<featureManager>`, then `<server>` (the `[@*]` opts in to its `description` attribute), then deletes the file.

## Test plan

- [x] `./gradlew :rewrite-xml:test --tests "org.openrewrite.xml.RemoveEmptyXmlTagsTest"` — 14 tests pass (3 original + 11 new covering file deletion, attribute-allowlist semantics, `fileMatcher` scoping, and multi-level nested collapse).
- [x] `./gradlew :rewrite-xml:recipeCsvGenerate` regenerates `rewrite-xml/.../recipes.csv`; `recipeCsvValidateContent` passes.
- [x] `./gradlew :rewrite-xml:licenseMain :rewrite-xml:licenseTest` pass.


rewrite-xml/src/main/java/org/openrewrite/xml/RemoveEmptyXmlTags.java
```diff
@@ -15,31 +15,184 @@
...
+    public String getDisplayName() {
+        return "Remove empty XML tags";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Repeatedly removes empty XML tags (optionally scoped by an XPath whitelist) until the tree is stable, " +
+               "and optionally deletes the file when its root tag becomes empty. " +
+               "Useful as a follow-up to recipes that strip individual tags and leave empty containers behind.";
+    }
```
timtebeek: Surprised to see these turned into getters.

Jenson3210: That's my delomboking I guess... 🤔  Reverting

Jenson3210:
@timtebeek this one would definitely still be valuable / need review

rewrite-xml/src/main/java/org/openrewrite/xml/RemoveEmptyXmlTags.java
```diff
@@ -15,31 +15,178 @@
...
+            example = "**/server.xml")
+    @Nullable
+    String fileMatcher;
+
+    @Option(displayName = "Delete file if empty",
+            description = "Delete the source file when the root tag has no remaining content after collapsing empty tags. Defaults to true.",
+            required = false,
+            example = "false")
+    @Nullable
+    Boolean deleteFileIfEmpty;
```
timtebeek, line 68: High over I'm going to say the xpath argument is likely fine, but the fileMatcher is best used as a declarative precondition, and the empty file removal as a separate recipe. No rush on any of this, and also ok if we let this PR lapse.

PR #7843, opened 2026-05-29
Go: model `for ... range` as a rich Go.RangeLoop instead of cramming it into J.ForEachLoop

knutwannheden:
## Motivation

Running a recipe over Go source mangled every `for ... range` loop:

```diff
-	for _, v := range items {
+	for_:=range items {
```

Go's range loop binds a **key and an optional value** joined by `:=` or `=` (`for k, v := range x`, `for k := range x`, `for range x`). That doesn't fit Java's single-variable `J.ForEachLoop.Control` (one `variable` + `iterable`). The Go side worked around it by defining its *own* fatter `java.ForEachControl` (`Key`/`Value`/`Operator`/`Iterable`) **masquerading under the Java type name**, then bridging it to the real `J.ForEachLoop.Control` over RPC with a hand-written codec.

- That bridge is the bug: the real `J.ForEachLoop.Control` has nowhere to put a second variable or the `:=`/`=`, so the codec synthesized a single variable from the key and **dropped the value and the operator**. Like the operator fixes in #7842, it round-trips fine Go↔Go (both sides use the fat struct) but truncates on the Java leg — invisible until a recipe forces full deserialization.

## Examples

New node (mirrors the per-language loop-node pattern — JS has `ForOfLoop`/`ForInLoop`):

```
Go.RangeLoop   for k, v := range x { … }
  key       JRightPadded<Expression>   // nullable (for range x)
  value     JRightPadded<Expression>   // nullable (for k := range x)
  operator  JLeftPadded<Type>          // Type {Equals, Define}; before = space before `range`
  iterable  Expression
  body      JRightPadded<Statement>
```

## Summary

- Added `Go.RangeLoop` end-to-end (Java `Go.java`/`GolangVisitor`/`GolangSender`/`GolangReceiver`; Go tree + parser + printer + visitor + RPC sender/receiver + factory & `RegisterValueType`).
- **Removed the cheat entirely**: the Go-customized `java.ForEachLoop`/`java.ForEachControl` structs and methods, their RPC send/receive, the lossy `visitForEachControl`/`visitForEachLoop` bridge overrides, and the `J$ForEachLoop`/`J$ForEachLoop$Control` wire registrations. Template matching (`comparator`/`substitution`/`go_template`) now matches `Go.RangeLoop`.
- `AssignOp` (`Equals`/`Define`) is kept and reused as the operator enum, round-tripping as the `Go.RangeLoop.Type` constant name.

Note for reviewers: a Go LST node needs **both** `RegisterFactory` (receive) and `RegisterValueType` (send) — registering only one silently desyncs the wire into a `ParseError`.

- This is one of a series removing "cheats" where a Go-side `java.*` struct carries more than its real `J.*` counterpart and truncates over RPC (follow-up to #7842). Known remaining, each its own follow-up PR: multi-value `return`, `if init; cond`, `switch init; tag`, grouped `var(...)`/`const(...)` blocks.

## Test plan

- [x] New range cases in `GoRpcPrintRoundTripIntegTest` (key+value, key-only, no-variable, `=` assign) — all green
- [x] `./gradlew :rewrite-go:test :rewrite-go:integTest` passes (no regressions)
- [x] `go test ./...` passes


greg-at-moderne:
Superseded by #7944?

PR #7856, opened 2026-06-01
Fix implement interface double space

benjaminknauer:
## What's changed?
Add a regression test for `ImplementInterface` when adding the first parameterized interface to a class.

Update `ImplementInterface` so the generated type uses an empty prefix when creating a new `implements` clause. When appending to an existing list, it keeps the existing single-space prefix needed after the comma.

## What's your motivation?
This is a regression introduced between `v8.83.1` and `v8.83.2`.

- The behavior still worked correctly in [`v8.83.1`](https://github.com/openrewrite/rewrite/releases/tag/v8.83.1). It changed in [`v8.83.2`](https://github.com/openrewrite/rewrite/releases/tag/v8.83.2) after [#7769](https://github.com/openrewrite/rewrite/pull/7769), commit [`59c7f99a6`](https://github.com/openrewrite/rewrite/commit/59c7f99a698e7c9f4fbb21a390342e5854db7d5e), added default whitespace for keyword-prefixed `JContainer`s in `withX(...)` mutators.

`ClassDeclaration.withImplements(..)` now supplies spacing when it creates an `implements` container. `ImplementInterface` still added a leading prefix to the generated type.

For a parameterized interface, whitespace was then present at two AST levels:

```java
class A implements  B<String> { // two whitespaces after implements keyword
}
```
The outer `J.ParameterizedType` received the default prefix from `withImplements(..)`, while its nested class type still had the prefix added by `ImplementInterface`.

### Checklist

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/ImplementInterface.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ImplementInterface.java
@@ -25,7 +25,6 @@ import java.util.Objects;
 
 import static java.util.stream.Collectors.toList;
 import static org.openrewrite.Tree.randomId;
-import static org.openrewrite.java.tree.Space.format;
 
 public class ImplementInterface<P> extends JavaIsoVisitor<P> {
     private final J.ClassDeclaration scope;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot for the immediate fix, and referring back to the original change here.

rewrite-java/src/main/java/org/openrewrite/java/ImplementInterface.java
steve-aom-elliott: I'd like to propose a slight alteration to how that has been done.

- When constructing `impl` around line 70, do not set the `prefix` at all.
- When constructing the `J.ParameterizedType typedImpl` (around line 85), set the `space` as `Space.SINGLE_SPACE` instead of `Space.EMPTY`
- When passing `impl` to `ListUtils.concat(..)` (around line 94), use `impl.withPrefix(Space.SINGLE_SPACE)` instead of just `impl`

steve-aom-elliott (approved):

PR #7858, opened 2026-06-01
Stream CsvDataTableStore.getRows from disk lazily instead of buffering the whole table

krlittle:
## What's changed?

`CsvDataTableStore.getRows(...)` read every matching CSV file fully into a `List<Object>` and returned that list's stream, so reading back a large data table held all of its rows in memory at once. It now selects the matching file(s) by their `@name`/`@group` comment header and parses each file **lazily, one row at a time** via the Univocity parser, behind a closeable `Stream`. Rows are produced on demand, so a whole table is never materialized in memory.

The change is intended to be behavior-preserving:
- same row order, prefix/suffix column stripping, group / `@name` matching, and typed-row (vs raw `String[]`) results;
- the writer is still closed eagerly at the `getRows` call, so a mid-run read still finalizes the file — only row parsing is deferred;
- error handling is unchanged: open/parse failures (which are unchecked) propagate, and only a stream-close `IOException` is swallowed.

## What's your motivation?

Recipes that read their own data tables back — e.g. to export or aggregate them — can produce very large tables (one row per method/class across a large repository). Buffering the entire table into a `List` before the consumer sees a single row makes peak memory scale with table size; streaming bounds the store side to one row at a time.

### Checklist


rewrite-core/src/main/java/org/openrewrite/CsvDataTableStore.java
```diff
@@ -237,54 +237,103 @@ private <T> Stream<T> readRows(String dataTableName, @Nullable String group, @Nu
...
+                .flatMap(path -> (Stream<T>) streamRows(path, meta, prefixCount, suffixCount));
+    }
+
+    /**
+     * Lazily parse a single CSV file row-by-row. The returned stream owns its input
+     * stream and parser; fully consuming the stream (or closing it) releases them.
+     * Rows are produced one at a time, so a whole table's rows are never held in
+     * memory at once.
+     */
+    private Stream<Object> streamRows(Path path, @Nullable RowMetadata meta, int prefixCount, int suffixCount) {
```
sambsnyd, line 273: Since CsvDataTableStore is used directly in the CLI you'll have to do a release which includes this before this will work at runtime. Recipe modules built with this running on un-updated CLI may encounter `NoSuchMethodException` here... unless you've verified that isn't true, you might want to guard invocation of this method with reflection until CLI with this method have been released for a little while

sambsnyd (approved):

natedanner (commented):
Lazy streaming looks right and the happy path is fine (both `RecipeRun` callers `collect()`, fully draining). Two things worth addressing:

1. The returned stream now holds an open file handle until it's drained/closed. `flatMap` closes an inner stream only once drained, and the outer stream's `close()` doesn't reach a still-open inner stream — so a short-circuiting caller (`findFirst`/`limit`) leaks a handle even with try-with-resources. `InMemoryDataTableStore` returns a detached stream, so callers can't assume either way. Worth documenting on `DataTableStore.getRows` that the stream must be fully consumed/closed.
2. No tests for the new semantics; existing ones all fully drain, so a short-circuit leak wouldn't be caught.

krlittle:
> Lazy streaming looks right and the happy path is fine (both `RecipeRun` callers `collect()`, fully draining). Two things worth addressing:
> 
> 1. The returned stream now holds an open file handle until it's drained/closed. `flatMap` closes an inner stream only once drained, and the outer stream's `close()` doesn't reach a still-open inner stream — so a short-circuiting caller (`findFirst`/`limit`) leaks a handle even with try-with-resources. `InMemoryDataTableStore` returns a detached stream, so callers can't assume either way. Worth documenting on `DataTableStore.getRows` that the stream must be fully consumed/closed.
> 2. No tests for the new semantics; existing ones all fully drain, so a short-circuit leak wouldn't be caught.

Thanks @natedanner!  I added the notes and the requested test with [1a907ff](https://github.com/openrewrite/rewrite/pull/7858/commits/1a907ff3faa51b84628c6afa68ea664e4ad63fa4).  

As for the reported leak, I was unable to reproduce it.  Two things release the handle on their own: since JDK 10, `flatMap` closes the inner stream when you short-circuit, and Univocity (the delimited text file parsing library we're using) closes the reader once it hits the end of the file. Any file small enough to read in one pass hits both, so the handle always gets closed.  I left the `onClose` in as a safety net for a larger file case that wouldn't make sense as a unit test.

rewrite-core/src/main/java/org/openrewrite/CsvDataTableStore.java
krlittle: Appreciate the callout, @sambsnyd.  It is a good concern to have so I took a lot of care to prove things out for us.  I'm pretty sure I've verified all angles here.  Let me know if this settles the dust or not for you:

- `DataTableStore` interface is unchanged - I only updated a javadoc there.
- The new method `CsvDataTableStore.streamRows` is private, so it's only reachable internally to `CsvDataTableStore`. No separately-compiled artifact can see it to reference it, so there's no cross-artifact linkage to fail.

To further prove out the case of a recipe built against this future version of rewrite running on a CLI that hasn't been updated with it, I ran a quick A/B test to root out any gotchas:

- I built a small test caller that uses CsvDataTableStore, compiled it against this branch, then ran it against the old `rewrite-core` build. I used `javap` to confirm the caller's bytecode never references `streamRows` but instead only calls `getRows` as we'd hope to expect.  At runtime the loaded old core didn't even declare `streamRows`, yet `getRows` returned every row and did not throw a `NoSuchMethodError`.

I feel pretty good about moving forward without any additional reflection guards in CLI.  I'm going to move this out of draft. 

As a "for your awareness" of the whole picture, Jonathan needed https://github.com/openrewrite/rewrite-prethink/pull/20 and so that was merged recently.

rewrite-core/src/main/java/org/openrewrite/DataTableStore.java
```diff
@@ -73,6 +79,8 @@ static DataTableStore noop() {
      * @param group          the group identifying the bucket, or null for ungrouped
      * @param <Row>          the row type
      * @return a stream of typed rows, or an empty stream if no rows exist
+     * @apiNote The returned stream may hold an open file handle; consume it fully or close it
+     * (try-with-resources). Closing is idempotent and safe after short-circuiting.
```
jkschneider, line 83: What is a caller supposed to interpret as an AutoCloseable resource?

PR #7878, opened 2026-06-02
Make `AdaptiveRadixTree` stack-safe (iterative search, insert, and copy)

steve-aom-elliott:
## Motivation

`AdaptiveRadixTree` traversed and duplicated its nodes recursively: `InternalNode.search`, `InternalNode.insert`, and `copy()` each consumed one JVM stack frame per tree level. For normal keys that depth is bounded by the key length, but a chain of `keyLength == 0` internal nodes — which `LeafNode.insert` builds when you insert a strict-extension family like `"a"`, `"aa"`, `"aaa"`, … (one link per extension) — makes the depth unbounded by the data. Searching, inserting, or copying the longest such key then walks one frame per byte and throws `StackOverflowError` once the path exceeds the JVM's default stack (~10k frames).

This PR makes all three traversals iterative (heap-bounded) and removes a per-insert allocation regression that the first iterative `insert` draft had introduced.

## Summary

- **`search`** — rewritten as a `while`-loop. Each iteration strictly advances `depth` (consuming the node's partial key plus one slot byte, or just a slot byte for `keyLength == 0` nodes), and the descent delegates to the already-non-recursive `LeafNode.search` on reaching a leaf. Semantics are unchanged: same partial-key matching, same termination on `depth == key.length`, same null-on-missing-child behavior.
- **`insert`** — rewritten iteratively. The terminal action (split / value set / new leaf / leaf delegation) produces the new subtree root; because every slot descended through already exists in its node, `addChild` always replaces in place and never grows an ancestor, so only the **immediate parent** needs its slot repointed and the original root is returned unchanged. This is tracked with a single `parent`+`slot` pair, so there is **no per-insert heap allocation** — an earlier draft kept the full descent path in two `ArrayList`s (one boxing every slot byte), which this removes.
- **`copy()`** — converted from a recursive deep-copy to an iterative one. Each node now exposes a non-recursive `shallowCopy()` (clones its own fields and child references) plus `cloneChildrenInto()`, and a driver walks an explicit work stack bounded by heap rather than the call stack.
- Added `AdaptiveRadixTreeBenchmark` (JMH) covering insert/search for realistic shared-prefix type signatures and for deep chains, gc-profiled.

## Performance

Iterative + single-parent `insert`/`search` vs. the recursive baseline (local JMH, 2 forks × 10 iterations, `gc` profiler):

| Benchmark | Baseline (recursive) | This PR |
|---|---|---|
| insert (type signatures) | 2.02 ops/ms · 843,266 B/op | 2.09 ops/ms · 843,266 B/op |
| insert (deep chain) | 0.060 ops/ms · 190,410 B/op | 0.125 ops/ms · 190,380 B/op |
| search (type signatures) | 1.77 ops/ms | 2.18 ops/ms |
| search (deep chain) | 0.062 ops/ms | 0.116 ops/ms |

Insert allocation is byte-for-byte identical to the recursive baseline (the draft's regression is eliminated), deep-chain insert/search are ~2× faster, and the realistic workloads are at parity or better.

## Test plan

- `surviveDeepChainOnSmallStack` inserts and searches a 20,000-deep chain on a 128 KB-stack thread; it `StackOverflowError`s on the recursive code and passes here.
- `copySurvivesDeepChainOnSmallStack` copies a 20,000-deep chain on a 128 KB-stack thread and asserts the copy is complete and independent of the original.
- Existing `AdaptiveRadixTreeTest` cases (including `copy`) pass unchanged.
- `./gradlew :rewrite-core:test` green.

### DIFF
--- a/rewrite-benchmarks/src/jmh/java/org/openrewrite/benchmarks/core/AdaptiveRadixTreeBenchmark.java
+++ b/rewrite-benchmarks/src/jmh/java/org/openrewrite/benchmarks/core/AdaptiveRadixTreeBenchmark.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.benchmarks.core;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openrewrite.internal.AdaptiveRadixTree;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


knutwannheden (approved):

knutwannheden:
Unsure where we'd actually get any signatures which are as long as 20K characters, but making this iterative rather than recursive is a good thing.

PR #7880, opened 2026-06-03
Resolve RPC wire type names through the codec (SourceFile subclasses + JavaType)

knutwannheden:
## Motivation

- [#7873](https://github.com/openrewrite/rewrite/pull/7873) made the wire-level `sourceFileType` canonical for `SourceFile` subclasses, but `RpcSendQueue` still reported each value's *runtime* class name as its `valueType`. When a tree (or its attribution) is represented at runtime by a synthetic subclass generated outside the `org.openrewrite` packages — e.g. a moderne-cli V3 lazy-loading proxy — that runtime name is one the remote has no codec/factory for, so the remote (e.g. the Python receiver) fails with *"No RPC codec registered … for '\<runtime class name>'"* and the RPC queues desynchronize.

- #7873 worked around this with a package-prefix heuristic in `getValueType` (report the superclass name when a value's package is not `org.openrewrite` but its superclass's is). That heuristic had no codec/type guard, only stripped a single subclass level, and — together with the `JavaType`-specific reflective lookup — meant `rewrite-core` hardcoded knowledge of `org.openrewrite.java.tree.JavaType` by string.

This PR makes **the codec decide the wire type name**, for every decomposable value, so `RpcSendQueue` no longer special-cases any language type.

## Examples

`RpcCodec.valueType(Object)` is the new hook (defaulting to the runtime class name, correct for self-encoding values). Codecs override it to canonicalize:

```java
// DynamicDispatchRpcCodec (tree nodes): walk up while the superclass is still assignable to
// the language marker — strips synthetic proxy layers, lands on the registered tree type.
//   Lazy_j_literal extends J.Literal           -> "org.openrewrite.java.tree.J$Literal"

// JavaTypeRpcCodec: JavaType is a *real* class hierarchy, so a marker-walk would over-collapse
// (JavaType.Class -> FullyQualified). It strips only synthetic (foreign-package) layers instead.
new JavaTypeRpcCodec().valueType(JavaType.ShallowClass.build("java.lang.String"));
//   -> "org.openrewrite.java.tree.JavaType$ShallowClass"   (real subtype preserved)
//   a foreign-package proxy of it                          -> "...JavaType$ShallowClass"
```

## Summary

- **`RpcCodec.valueType(Object)`** — new default method returning the runtime class name; the codec owns the wire type name.
- **`DynamicDispatchRpcCodec`** — overrides `valueType` with the marker-assignability walk for tree nodes. `getCodec` now resolves cross-cutting codecs by assignability even when `sourceFileType` is null, and caches that resolution per value class via `ClassValue` (replacing the per-node cross-bucket scan — also takes self-encoding `Markers`/`Checksum` off the scan path).
- **`JavaTypeRpcCodec`** (new, `rewrite-java`) — `DynamicDispatchRpcCodec<JavaType>` registered via `ServiceLoader`. Standalone (not a self-codec) to keep the `JavaType` model RPC-free and avoid a tree↔`internal.rpc` cycle. Overrides `valueType` with a **package-gated** walk (stripping only synthetic foreign-package layers) because `JavaType` is a real class hierarchy; `rpcSend`/`rpcReceive` delegate to the existing `JavaTypeSender`/`JavaTypeReceiver`.
- **`RpcSendQueue`** — deletes the `JavaType` special-case: the `JAVA_TYPE_PACKAGE`/`JAVA_TYPE_NAME` constants, the reflective `Class.forName` lookup, the single-level superclass branch, and the `JavaType.Primitive` enum-guard. `valueType` flows through the codec; `core` no longer references `rewrite-java` by string.

`JavaType` decomposition is unchanged — it is still driven by `JavaTypeSender` via the existing `onChange` callbacks; the codec only supplies the wire type name (its `rpcSend`/`rpcReceive` are an off-hot-path fallback, since `JavaType` is always reached as a child with an `onChange`).

## Test plan

- [x] New `JavaTypeRpcCodecTest`: codec resolves for `JavaType` values; `valueType` does **not** collapse real subtypes (`ShallowClass` stays `ShallowClass`, guarding the over-collapse hazard); strips a synthetic foreign-package subclass layer. (TDD — watched the marker-walk fail before adding the package-gated override.)
- [x] `DynamicDispatchRpcCodecTest` (from the SourceFile commit) — green.
- [x] `JavaTypeAnnotationRpcTest` (full `JavaType` round-trip incl. `Primitive`), `JavaReceiverTest`, `SubclassCodecDispatchTest`, and all `rewrite-core` `rpc.*` tests — 0 failures.
- [ ] The pre-existing `@Disabled` `JavaSendReceiveTest.sendReceiveIdempotence` ("until we've cleaned up the enum serialization") still times out — a real tree has other raw enums (`J.ClassDeclaration.Kind`, modifiers, …) beyond `JavaType.Primitive`. This PR is a step toward that, not the completion; left disabled.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/DynamicDispatchRpcCodec.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/DynamicDispatchRpcCodec.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.WeakHashMap;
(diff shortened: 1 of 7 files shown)


jkschneider:
I don't love the inheritance use here:

image

It shows here:

image

There is no meaning to this constructed instance other than to serve what essentially amounts to a utility method. And that convinces me that this isn't really carrying its weight.

PR #7887, opened 2026-06-03
8.83.5 regression: RocksdbMavenPomCache fails to deserialize stale POM cache entries

benjaminknauer:
### Problem

After upgrading from OpenRewrite `8.83.4` to `8.83.5`, Maven resolution failed while reading from `RocksdbMavenPomCache`:

```text
No Object Id found for an instance of `org.openrewrite.maven.tree.GroupArtifactVersion` to assign to property '@ref'
```

Deleting ~/.rewrite-cache fixed the issue locally.

### Version
Regression observed after upgrading from 8.83.4 to 8.83.5.

Via the build plugin with Maven recipes enabled.

### Expected behavior
Existing RocksDB Maven POM cache entries written by 8.83.4 should either deserialize correctly or be invalidated automatically after upgrading.

### Actual behavior
8.83.5 attempts to deserialize stale cached POM data and fails with the Jackson @ref object-id error.

### Possible cause
8.83.5 introduced Jackson `@JsonIdentityInfo(... property = "@ref")` on Maven model types such as `GroupArtifactVersion` and `ResolvedGroupArtifactVersion`, changing the serialized form of cached Pom objects.

- The relevant change appears to have been introduced in openrewrite/rewrite@8dfd30b60 as part of PR #7845:

However, `Pom.getModelVersion()` was not incremented, so old RocksDB cache entries are still treated as compatible.

### Workaround
Delete ~/.rewrite-cache.

### Suggested fix
Increment `Pom.getModelVersion()` so stale RocksDB POM cache entries are invalidated after the Maven model serialization change.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
@@ -60,7 +60,7 @@ public class Pom {
      */
     public static int getModelVersion() {
         //NOTE: This value should be incremented if there are any model changes to Pom (or one of its referenced types)
-        return 2;
+        return 3;
     }
 
     @Nullable
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/cache/RocksdbMavenCacheTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/cache/RocksdbMavenCacheTest.java
@@ -15,12 +15,19 @@
  */
 package org.openrewrite.maven.cache;
 
+import com.fasterxml.jackson.annotation.JsonIdentityInfo;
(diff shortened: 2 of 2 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/tree/Pom.java
```diff
@@ -60,7 +60,7 @@ public class Pom {
      */
     public static int getModelVersion() {
         //NOTE: This value should be incremented if there are any model changes to Pom (or one of its referenced types)
-        return 2;
+        return 3;
```
jkschneider, line 63: I think this refers to Maven's model version, not OpenRewrite's serialized representation.

PR #7894, opened 2026-06-03
C#: decompose CsDocComment.DocComment over RPC via a proper visitor

macsux:
## Summary
- A `CSharpVisitor` pass parses a raw `CsDocCommentRawComment` into a structured `CsDocComment.DocComment` tree. Previously the RPC sender flattened that tree back to raw text; this PR instead adds a proper doc-comment tree visitor pair (`CsDocCommentSender` / `CsDocCommentReceiver`) that decomposes and reconstructs the tree over RPC, mirroring how the Java side keys `Javadoc.DocComment` by id. The comment is now keyed by its id in `Space`'s comment list and each node transmits its id, markers, and fields (including embedded `J` cref/name references).
- The C# side has no structured doc-comment model (it stores raw `XmlDocComment` text), so `CsDocCommentReceiver.cs` consumes the decomposed tree and re-flattens it to text, mirroring `CsDocCommentPrinter`. Without this the C# receiver threw `Cannot map Java type name: CsDocComment$DocComment` when a recipe-modified tree was serialized back for printing.
- Also bumps `Microsoft.CodeAnalysis.*` packages from 5.0.0 to 5.3.0.

## Test plan
- New Java sender→receiver round-trip tests in `CSharpSenderDocCommentTest`, including a multiline/nested-element case.
- New end-to-end regression test in `CSharpRecipeTest` that forces structuring via a `CSharpVisitor` pass then `print()`s through the real C# RPC process (the path that previously crashed).
- Full `rewrite-csharp` Java suite passes; all 1945 C# xUnit tests pass.

### DIFF
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpSender.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpSender.java
@@ -16,11 +16,16 @@
 package org.openrewrite.csharp.rpc;
 
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.PrintOutputCapture;
 import org.openrewrite.Tree;
 import org.openrewrite.csharp.CSharpVisitor;
+import org.openrewrite.csharp.CsDocCommentPrinter;
 import org.openrewrite.csharp.tree.Cs;
+import org.openrewrite.csharp.tree.CsDocComment;
 import org.openrewrite.csharp.tree.CsDocCommentRawComment;
 import org.openrewrite.csharp.tree.Linq;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.internal.rpc.JavaSender;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.rpc.RpcSendQueue;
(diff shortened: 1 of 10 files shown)


rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpSender.java
```diff
@@ -873,12 +878,24 @@ public CSharpSenderDelegate(CSharpSender delegate) {
 
         @Override
         public void visitSpace(Space space, RpcSendQueue q) {
-            q.getAndSendList(space, Space::getComments,
+            // Structured doc-comment trees are not transmitted over the wire; flatten them
```
jkschneider: We do have a JavadocVisitor on the java side, and I think we should probably implement a similar doc-comment tree visitor here on CS and decompose it over RPC as well.

jkschneider (changes requested):

rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpSender.java
macsux: Done in f77accaa7. Added a proper doc-comment tree visitor pair (`CsDocCommentSender`/`CsDocCommentReceiver`) that decomposes and reconstructs the `CsDocComment.DocComment` tree over RPC, mirroring how the Java side keys `Javadoc.DocComment` by id. No more flattening on the Java side.

Since the C# side has no structured doc-comment model (it stores raw `XmlDocComment` text), I also taught the C# receiver to consume the decomposed tree and re-flatten it to text (mirroring `CsDocCommentPrinter`) — otherwise it threw when a recipe-modified tree was serialized back for printing. Verified with a Java↔Java sender/receiver round trip and an end-to-end `print()` through the real C# RPC process; full `rewrite-csharp` Java suite and all 1945 C# xUnit tests pass.

macsux: @jkschneider lmk please if this is good to merge at this point. should fix some errors we're seeing in saas

PR #7919, opened 2026-06-06
rewrite-csharp: add LOCAL_NUGET_FEED env var and skip repo nuget.config on tool install

macsux:
Installs the C# tool from a fresh working directory with no `nuget.config` on any ancestor path, so NuGet's own config discovery applies the user/machine-level global configuration (preserving nuget.org / private-proxy feeds for transitive dependencies) while skipping repo-level overrides that could `<clear/>` sources or enable package source mapping. The tool package's feed now comes from `LOCAL_NUGET_FEED` when set, otherwise the NuGet global packages cache (`~/.nuget/packages/...`) as before. This also fixes RPC edit batching to resolve the canonical (codec-registered) source file type rather than the raw runtime class name, since under lazy-deserialization LSTs the runtime class is a generated proxy absent from the remote's declared languages, which would otherwise skip every batched edit on that file.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
+++ b/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java
@@ -36,6 +36,7 @@ import org.openrewrite.table.SourcesFileErrors;
 import org.openrewrite.table.SourcesFileResults;
 
 import org.openrewrite.marker.SearchResult;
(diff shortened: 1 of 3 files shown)


shanman190:
I've got a similar change within this PR.
https://github.com/openrewrite/rewrite/pull/7819

jkschneider:
Closed as superseded by #7819

PR #7931, opened 2026-06-07
JavaScript: Add optional `files` subset to RPC `parseProject`

knutwannheden:
## Motivation

Tools that do in-session incremental re-parsing of edited source files have no good way to refresh a single changed `.ts`/`.js` file today: a changed file can only be refreshed by re-parsing the entire JS project (slow) or by falling back to a Quark placeholder (no AST at all).

This adds an optional `files` subset parameter to the JavaScript RPC `parseProject` API so a caller can re-parse only a specific set of changed files within an already-established project context. The strategy is **"load the whole project for typing, return only the subset"**: the parser still discovers and type-checks the entire project (so types in the returned subset resolve exactly as in a full parse — including ambient/global declarations that a subset-rooted program would miss), but only serializes and returns the requested files, avoiding the cost of shipping every unchanged file across the RPC boundary.

The change is fully additive and backward-compatible: when `files == null`, behavior is identical to today. The same shape will later be applied to Python/Go/C#.

## Examples

```java
// Full project parse (unchanged):
client.parseProject(projectPath, ctx);

// Re-parse only the files that changed, with full project type context:
client.parseProject(projectPath, ParseProjectOptions.builder()
        .files(List.of("src/app.ts"))
        .build(), ctx);
```

`ParseProjectOptions` follows the builder pattern introduced for `rewrite-python`'s `parseProject`; the new `files` input is builder-only (mirroring how `dependencyPath` is builder-only there). Existing positional overloads are unchanged and delegate through the builder.

## Summary

- **`ParseProjectOptions`** (new): `@Value @Builder` bundling the optional parsing inputs (`exclusions`, `relativeTo`, `files`), mirroring the rewrite-python class.
- **`ParseProject`** (wire request): added `@Nullable List<String> files` and a frozen-but-unused `@Nullable Map<String, Object> options` forward-compat carrier (so a future "true incremental" optimization needs no second wire break), plus backward-compatible constructors.
- **`JavaScriptRewriteRpc`**: new canonical `parseProject(Path, ParseProjectOptions, ExecutionContext)` overload; the existing positional overloads delegate to it.
- **`parse-project.ts`** (server): when `files` is provided, non-JS categories are restricted to the subset up front; JS/TS files are still all parsed (one TypeScript program over the whole project) and filtered only at serialization time. Subset matching compares `path.relative(relativeTo, absolutePath)` normalized to `/` against the `files` set. A `files` entry that discovery wouldn't have parsed simply yields nothing. The `options` field is ignored for now.
- **`project-parser.ts`**: documented that `buildAutodetectMarker` yields detector defaults (rather than throwing) on an empty sample.

### Path contract

`files` entries and response `sourcePath`s share the same base — relative to `relativeTo` when set, else `projectPath` — normalized to `/`.

## Test plan

- [x] New `parseProjectSubset` integration test: full parse vs `parseProject(..., files=["app.ts"], ...)` — asserts (a) only `app.ts` is returned, (b) its `sourcePath` matches the full-parse path exactly, and (c) a type that resolves in a full parse (a cross-file `add(...)` call into an unreturned `math.ts`) still resolves in the subset parse, proving the whole project was loaded for typing.
- [x] New `parseProjectSubsetWithUnknownFileYieldsNothing` test: a subset entry discovery wouldn't have parsed yields an empty result, not an error.
- [x] Existing `parseProject`, `parseProjectWithExclusions`, `parseProjectWithVariousYamlStructures`, and `javaTypeAcrossRpcBoundary` integration tests still pass (full mode, `files == null`, unregressed).
- [x] `npm run typecheck` passes; existing `test/rpc/parse-project.test.ts` Node tests pass.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/project-parser.ts
+++ b/rewrite-javascript/rewrite/src/javascript/project-parser.ts
@@ -195,6 +195,10 @@ export class ProjectParser {
      * Builds an Autodetect marker from the given source files.
      * Samples all files to detect common formatting styles.
      * Uses dynamic import to avoid circular dependencies.
+     *
+     * When {@code sourceFiles} is empty (e.g. a subset parse that doesn't include any
+     * JS/TS files to sample) the detector yields its built-in defaults rather than
+     * throwing — a less-rich marker, but a safe one.
      */
     async buildAutodetectMarker(sourceFiles: SourceFile[]): Promise<Marker> {
         // Dynamic import to break circular dependency at module load time:
--- a/rewrite-javascript/rewrite/src/rpc/request/parse-project.ts
+++ b/rewrite-javascript/rewrite/src/rpc/request/parse-project.ts
@@ -54,7 +54,19 @@ export class ParseProject {
          * If not specified, paths are relative to projectPath.
          * Use this when parsing a subdirectory but wanting paths relative to the repository root.
          */
-        private readonly relativeTo?: string
+        private readonly relativeTo?: string,
+        /**
+         * Optional subset of files to serialize and return, identified by their source path relative to
+         * `relativeTo` (or `projectPath` when not set), normalized to forward slashes. When omitted the
+         * whole project is returned. When present the whole project is still discovered and type-checked,
+         * but only these files are serialized and returned.
+         */
+        private readonly files?: string[],
+        /**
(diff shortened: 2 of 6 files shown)


jkschneider:
Would like to consider if there is a more comprehensive change we can make to allow JavaScript prebuild.

PR #7942, opened 2026-06-09
Upgrade bom to the latest patch upgrades to the latest release

BoykoAlex:
Upgrading BOM to latest patch release pick the latest release version instead.


timtebeek (approved):
Thanks a lot for spotting and fixing the discrepancy here; I've added a `@Nullable` annotation here to make that even more explicit. Hope you agree!

PR #7954, opened 2026-06-09
rewrite-kotlin: stop mapped-type fallback shadowing real stdlib extensions

timtebeek:
`RecipeFirMappedTypeFallbackExtension` generated a synthetic top-level extension for every public instance method of each Java↔Kotlin mapped class, including names that already have a real Kotlin stdlib extension (`toUpperCase`, `trim`, `substring`, `isEmpty`, …), which shadowed the real `kotlin.text.StringsKt` extensions. As a result the recipe DSL's `MethodMatcher` spec recorded the synthetic `__GENERATED__CALLABLES__Kt` facade while parsing (where the plugin is inactive) resolves the canonical owner, so authored recipes matched nothing at runtime — regressing 9 `moderneinc/recipes-kotlin` tests. The fix only fills genuine gaps: it skips generating a stand-in whenever the method name already resolves as a real stdlib extension (queried via the non-generated `dependenciesSymbolProvider` to avoid recursion), while JDK additions the mapped-type customizer hides with no Kotlin equivalent (`stripIndent`, `formatted`, `translateEscapes`, …) are still generated so recipe authoring is unaffected. Adds `MappedTypeFallbackShadowingTest` (asserts the generated matcher targets `kotlin.text.StringsKt`, not the synthetic facade) and `RecipeFirMappedTypeFallbackParsingTest` (parser-side invariant). Verified: full `:rewrite-kotlin:test` green, and against a locally-published patched snapshot the previously-failing `recipes-kotlin` tests go from 9 failures to 0 with no test changes.

### DIFF
--- a/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeFirMappedTypeFallbackExtension.kt
+++ b/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeFirMappedTypeFallbackExtension.kt
@@ -25,6 +25,7 @@ import org.jetbrains.kotlin.fir.extensions.DeclarationGenerationContext
 import org.jetbrains.kotlin.fir.extensions.ExperimentalTopLevelDeclarationsGenerationApi
 import org.jetbrains.kotlin.fir.extensions.FirDeclarationGenerationExtension
 import org.jetbrains.kotlin.fir.plugin.createTopLevelFunction
+import org.jetbrains.kotlin.fir.resolve.providers.dependenciesSymbolProvider
 import org.jetbrains.kotlin.fir.symbols.impl.FirNamedFunctionSymbol
 import org.jetbrains.kotlin.fir.types.ConeKotlinType
 import org.jetbrains.kotlin.fir.types.ConeKotlinTypeProjectionOut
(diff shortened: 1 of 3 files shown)


timtebeek:
## Downstream impact: `moderneinc/recipes-kotlin`

I reproduced the regression against `recipes-kotlin` by building it (no test changes) with a locally-published `rewrite-kotlin:8.85.0-SNAPSHOT`, toggling only this PR's catalog filter.

### Failures without the fix (9 total)

All are migration recipes whose `rewrite { } to { }` before-pattern calls a `String` method that exists **both** as a `java.lang.String` instance method **and** as a real `kotlin.text` extension — exactly the methods the fallback extension was wrongly generating stand-ins for.

**`org.openrewrite.kotlin.migrate.Kotlin1To2Test` (5)**
- `toUpperCase becomes uppercase()` — `Recipe was expected to make a change but made no changes.`
- `toLowerCase becomes lowercase()` — `Recipe was expected to make a change but made no changes.`
- `toUpperCase with Locale becomes uppercase with Locale()` — `Recipe was expected to make a change but made no changes.`
- `toLowerCase with Locale becomes lowercase with Locale()` — `Recipe was expected to make a change but made no changes.`
- `Kotlin1To2 composite fires across categories()` — `Unexpected result` (the composite leaves the `toUpperCase`/`toLowerCase` call sites unchanged)

**`org.openrewrite.kotlin.stdlib.StdlibTest` (4)**
- `string trim isEmpty becomes isBlank()` — `Recipe was expected to make a change but made no changes.`
- `string trim isNotEmpty becomes isNotBlank()` — `Recipe was expected to make a change but made no changes.`
- `string substring from zero becomes take()` — `Recipe was expected to make a change but made no changes.`
- `string substring from n becomes drop()` — `Recipe was expected to make a change but made no changes.`

### Why they failed

`recipes-kotlin` compiles its recipe sources with `rewrite-kotlin` on its `kotlinCompilerPluginClasspath`, so the recipe-DSL FIR plugin is active at **authoring** time. For each before-pattern, e.g. `rewrite { s: String -> s.toUpperCase() }`, the synthetic stand-in shadowed the real `kotlin.text.StringsKt.toUpperCase`, so the IR matcher-spec computation recorded the synthetic facade:

```
__GENERATED__CALLABLES__Kt toUpperCase(*)     ❌ (before this PR)
kotlin.text.StringsKt toUpperCase(*)          ✅ (after this PR)
```

At recipe **runtime**, `KotlinParser` parses the target code **without** the plugin, so the same call resolves to the canonical `kotlin.text.StringsKt`. Authoring and parsing disagreed on the owner, the `UsesMethod` precondition / `MethodMatcher` never matched, and the recipe made no changes.

(Note `lowercase()`/`uppercase()` etc. were unaffected: `java.lang.String` has no `lowercase()` method, so no stand-in was ever generated for those names — only the `toUpperCase`/`toLowerCase`/`trim`/`substring` family, which Java does declare, were shadowed.)

### Effect of this change

By skipping stand-in generation for any name that already resolves as a real stdlib extension, `s.toUpperCase()`/`trim()`/`substring()` resolve to the genuine `kotlin.text.StringsKt` extensions at authoring time, so the generated matcher names the canonical owner and matches at runtime. Result against the patched snapshot, with **no test changes**:

- `Kotlin1To2Test`: 190 run, **0 failures** (was 5)
- `StdlibTest`: 15 run, **0 failures** (was 4)

Genuine-gap JDK methods (`stripIndent`, `formatted`, `translateEscapes`, …) are still generated, so recipe authoring that relies on them is unaffected (covered by the existing `RecipeMappedTypeFallbackTest`).

greg-at-moderne (approved):
I admit I don't fully understand this. At the same time I don't see any reasons not to merge it.

PR #7969, opened 2026-06-10
JavaScript: share active RewriteRpc connection across module copies (#7968)

### LINKED ISSUE #7968: JS recipe composing Java (RPC) recipes fails to install: "Ensure the constructor can be called without any arguments" (InstallRecipes path; follow-up to #7892)
A TypeScript recipe whose `recipeList()` includes Java recipes (composed via `prepareJavaRecipe`, e.g. `addDependency`/`removeDependency`) fails to install through the Moderne CLI with the generic wrapper error:

```
Request InstallRecipes failed with message: Failed to install recipe 'ReplaceNodeSassWithSass'. Ensure the constructor can be called without any arguments.
```

- This is the same symptom that #7892 ("JavaScript: don't try to install RPC recipes", merged 2026-06-03) set out to fix — but #7892 only guarded the **`PrepareRecipe`** sub-recipe-registration path (`src/rpc/request/prepare-recipe.ts`, skipping `instanceof RpcRecipe`). The **`InstallRecipes` → `RecipeMarketplace` `descriptor()`** path still fails:

`marketplace.js` `install()` → `new recipe({})` → `await recipeInst.descriptor()` → `Recipe.descriptor()` calls `await this.recipeList()` and then maps `r.descriptor()` over each sub-recipe. For a JS recipe whose `recipeList()` returns RPC/Java recipes, resolving those sub-recipes throws, and the marketplace catch re-wraps it as the misleading "Ensure the constructor can be called without any arguments" message (the real `err.cause` is not propagated across the JSON-RPC boundary, so it never reaches the CLI logs).

### Reproduction

`moderneinc/rewrite-angular` nightly `cli-smoke-test` job — recipe `org.openrewrite.angular.migration.replace-node-sass-with-sass`, whose `recipeList()` is:

```ts
async recipeList(): Promise<Recipe[]> {
    return [
        await removeDependency({packageName: "node-sass"}),
        await addDependency({packageName: "sass", version: "^1.35.0", scope: "devDependencies"}),
    ];
}
```

Failing run: https://github.com/moderneinc/rewrite-angular/actions/runs/27295741909

- Reproduced locally end-to-end with **post-#7892 artifacts**:

- Moderne CLI `4.2.13`
- `org.openrewrite:rewrite-javascript` jar `8.85.0-20260610.170124-54` (pins npm `8.85.0-20260610-134219`)
- npm `@openrewrite/rewrite@8.85.0-20260610-134219` — whose `dist/rpc/request/prepare-recipe.js` **does** contain the #7892 `instanceof RpcRecipe` guard

```
mod config recipes jar install org.openrewrite:rewrite-javascript
mod config recipes npm install <built recipes-angular/dist>
# => Failed to install recipe 'ReplaceNodeSassWithSass'. Ensure the constructor can be called without any arguments.
```

Because both the jar and npm package are post-fix, this is a genuine remaining gap rather than a stale version.

### Expected

- RPC/Java sub-recipes contained in a JS recipe's `recipeList()` should be skipped (or resolved remotely) during `InstallRecipes`/marketplace descriptor resolution, the same way #7892 handles them in the `PrepareRecipe` path — not instantiated/registered locally.

### Suggestions

- Apply the `instanceof RpcRecipe` skip (as in #7892) to the `RecipeMarketplace` install/`descriptor()` path as well.
- Separately, the marketplace catch should preserve and surface `err.cause` (it is currently dropped at the JSON-RPC serialization boundary, so the "constructor" message is misleading and hides the real failure).

timtebeek:
## What's changed?

- Fixes #7968. A TypeScript recipe whose `recipeList()` composes Java recipes via `prepareJavaRecipe` (e.g. `addDependency` / `removeDependency`) failed to install through the Moderne CLI with the misleading wrapper error:

```
Failed to install recipe 'ReplaceNodeSassWithSass'. Ensure the constructor can be called without any arguments.
```

- This is the same symptom #7892 set out to fix, but on the `InstallRecipes` → `RecipeMarketplace.descriptor()` path rather than the `PrepareRecipe` path.

## Root cause

The active `RewriteRpc` connection was held in a `static` field on the class. A recipe package that **bundles** `@openrewrite/rewrite` (or resolves it from its own `node_modules`) loads a *separate copy* of the module, with its own class object and therefore its own statics. The host sets `_global` on its copy; the recipe package's copy never sees it, so:

- `recipeList()` → `prepareJavaRecipe()` → `RewriteRpc.get()` returns `undefined`
- `prepareJavaRecipe` throws `"no active RewriteRpc connection"`
- during `InstallRecipes`, that real cause is dropped at the JSON-RPC boundary (only `message` crosses) and the marketplace re-wraps it as the generic "constructor" hint.

Verified the static is per-copy: loading two physical copies of the compiled module and calling `set()` on one leaves `get()` on the other returning `undefined`. Moving the slot to `globalThis` makes both copies agree.

## The fix

- **`rewrite-rpc.ts`** — store the active connection on `globalThis` under a `Symbol.for("org.openrewrite.rpc.RewriteRpc.global")` key. `Symbol.for` resolves to the same symbol in every module copy, so all copies (host, bundled recipe package, separately-installed `node_modules`) share one connection.
- **`marketplace.ts`** — fold the underlying cause into the install error message (`… Cause: <message>`) so it survives the JSON-RPC boundary and reaches the caller's logs instead of being hidden behind the constructor hint.

## Tests

- `test/rpc/global-connection.test.ts` — the active connection is shared via the `globalThis` `Symbol.for` slot (the cross-copy contract).
- `test/recipe.test.ts` — `marketplace.install` surfaces the underlying cause when descriptor computation fails.

All `test/rpc/` and `test/recipe.test.ts` suites pass (Java RPC integration tests included); `npm run typecheck` is clean.

### DIFF
--- a/rewrite-javascript/rewrite/src/marketplace.ts
+++ b/rewrite-javascript/rewrite/src/marketplace.ts
@@ -87,7 +87,14 @@ export namespace RecipeMarketplace {
                         const recipeInst = new recipe({});
                         this.recipes.set(await recipeInst.descriptor(), recipe);
                     } catch (e) {
-                        const err = new Error(`Failed to install recipe '${recipe.name}'. Ensure the constructor can be called without any arguments.`);
+                        // Surface the underlying cause inline: it is dropped at the
+                        // JSON-RPC serialization boundary (only `message` survives), so
+                        // folding it into the message is the only way it reaches the
+                        // caller's logs. Without this, a failure deeper in the recipe
+                        // (e.g. a sub-recipe needing an RPC connection) is hidden behind
+                        // the generic "constructor" hint. See gh-7968.
+                        const cause = e instanceof Error ? e.message : String(e);
+                        const err = new Error(`Failed to install recipe '${recipe.name}'. Ensure the constructor can be called without any arguments. Cause: ${cause}`);
                         (err as any).cause = e;
                         throw err;
                     }
--- a/rewrite-javascript/rewrite/src/rpc/rewrite-rpc.ts
+++ b/rewrite-javascript/rewrite/src/rpc/rewrite-rpc.ts
@@ -46,7 +46,21 @@ import {ReferenceMap} from "../reference";
 import {GetLanguages} from "./request/get-languages";
 
 export class RewriteRpc {
-    private static _global?: RewriteRpc;
+    /**
+     * Key for the active {@link RewriteRpc} connection on {@link globalThis}.
+     *
+     * It deliberately lives on `globalThis` rather than as a `static` field:
+     * a recipe package that bundles `@openrewrite/rewrite` (or resolves it from
+     * its own `node_modules`) loads a *separate copy* of this module, with its
+     * own class object and therefore its own statics. A `static` field set by
+     * the host would be invisible to such a copy, so `RewriteRpc.get()` (e.g.
+     * via `prepareJavaRecipe`) would return `undefined` and throw "no active
+     * RewriteRpc connection" — surfacing during `InstallRecipes` as the
+     * misleading "Ensure the constructor can be called without any arguments".
+     * {@link Symbol.for} resolves to the same symbol across every module copy,
+     * so all copies share the one active connection. See gh-7968.
+     */
+    private static readonly GLOBAL_KEY: symbol = Symbol.for("org.openrewrite.rpc.RewriteRpc.global");
 
     private readonly snowflake = SnowflakeId();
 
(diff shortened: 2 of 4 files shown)


timtebeek:
@greg-at-moderne could use your help to see if this makes any sense in light of the the issue and downstream failure 🙏🏻

greg-at-moderne (approved):
I didn't run the code, but have analyzed it locally and it makes sense.
There's only one instance of RewriteRpc process. It might have multiple modules loaded (potentially with different versions of dependencies, incl. different versions of the openrewrite/rewrite dependency), but they all share the single "connection" to the mother JVM process. The connection is via stdin/stdout, so there's no talk of multiple parallel connections.

After realizing this, the "very global" state of the `RewriteRpc` is the way to go.

Thanks for fixing this.

Issue #7986, opened 2026-06-12
Gradle 9.x: recipes for new 9.1–9.6 deprecations


timtebeek:
## Background

Gradle 9.x has introduced a number of new deprecations across minor releases (9.1–9.6), documented in the [Gradle 9.6 upgrade guide](https://docs.gradle.org/9.6.0-rc-2/userguide/upgrading_version_9.html#v9_deprecations_9_6). Several of these are mechanical DSL/API rewrites that are good candidates for OpenRewrite recipes and could be added to `org.openrewrite.gradle.MigrateToGradle9`.

This issue tracks candidate recipes for the new 9.2–9.6 deprecations.

## Already covered in `MigrateToGradle9`

- [x] 9.1 multi-string / map dependency notation → `DependencyUseStringNotation`
- [x] 9.1 one declaration per statement → `gradle9.OneDependencyDeclarationPerStatement`
- [x] `JavaPluginConvention` removal (top-level `sourceCompatibility`/`targetCompatibility`) → `gradle9.UseJavaExtensionBlock`
- [x] `mainClassName` → `mainClass` → `gradle9.UseMainClassProperty(ForApplication)`
- [x] `version = { }` → closure form → `gradle9.UseVersionClosure`
- [x] spread `*configurations.all` → `gradle9.RewriteSpreadAllInConfigurationsBlock`

## New recipe candidates (mechanical, good fit)

High value:

- [x] **Current-project dependency: module coordinates → `project(":x")`** (9.3) — replace `group:name:version` referring to the current project with `project()` notation — `gradle9.UseProjectDependencyInsteadOfModuleCoordinates` (#7989)
- [ ] **`Project` object as dependency notation → `project(":x")`** (9.6)
- [x] **`DomainObjectCollection.findAll(Closure)` → `matching(Spec)`** (9.4) — `gradle9.UseMatchingInsteadOfFindAll` (#7989)
- [ ] **`RepositoryHandler.flatDir(Map)` / `mavenCentral(Map)` → action form** (9.6)

Simple removals / renames:

- [ ] **`Dependencies.getProject()` → `project()` (no-arg)** (9.4)
- [ ] **`apply false` in precompiled script plugins → remove** (9.4)
- [ ] **`.version(...)` in precompiled *settings* script plugins → remove** (9.4)
- [ ] **PMD `targetJdk` property → remove** (9.6)
[...]

PR #7989, opened 2026-06-12
Gradle 9: recipes for findAll→matching (9.4) and current-project project() notation (9.3)

timtebeek:
Adds two new recipes to `MigrateToGradle9` covering Gradle 9.x deprecations that are mechanical and well-defined.

### `UseMatchingInsteadOfFindAll` (Gradle 9.4)
Gradle 9.4 deprecates the Groovy `DomainObjectCollection.findAll(Closure)` overload in favor of the lazy `matching(Spec)`. This recipe rewrites `findAll { ... }` to `matching { ... }` when the receiver is a known Gradle container collection (`tasks`, `configurations`, `sourceSets`, including chains through `withType`/`matching`/`named` and `project.tasks` field access).

```groovy
def checkTasks = tasks.findAll { it.name.startsWith("check") }
// becomes
def checkTasks = tasks.matching { it.name.startsWith("check") }
```

It deliberately leaves `subprojects`/`allprojects` (a plain `Set<Project>`, not a `DomainObjectCollection`) and arbitrary collections alone to avoid false positives.

### `UseProjectDependencyInsteadOfModuleCoordinates` (Gradle 9.3)
Gradle 9.3 deprecates depending on the current project by its own `group:name:version` module coordinates (in Gradle 10 this would attempt repository resolution instead of resolving the local project). This recipe replaces such a declaration with the equivalent `project("<path>")` notation, using the `GradleProject` marker to identify the current project's coordinates and path.

```groovy
dependencies {
    implementation 'com.example:my-project:1.0.0'
}
// becomes (when this build script IS com.example:my-project)
dependencies {
    implementation project(':my-project')
}
```

Both the no-parentheses (Groovy command syntax) and parenthesized forms are handled, preserving the `OmitParentheses` marker. Constraints blocks and non-self coordinates are left untouched.

### Notes
- Tracking issue: #7986
- Both recipes are Groovy-DSL focused (the Kotlin DSL uses `filter` rather than the deprecated `findAll(Closure)` overload, and Kotlin `project()` rewriting is left as follow-up).
- Reference: [Gradle 9.6 upgrade guide — deprecations](https://docs.gradle.org/9.6.0-rc-2/userguide/upgrading_version_9.html#v9_deprecations_9_6)


rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/UseMatchingInsteadOfFindAll.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class UseMatchingInsteadOfFindAll extends Recipe {
+
+    @Getter
+    final String displayName = "Use `matching(Spec)` instead of `findAll(Closure)` on Gradle container collections";
+
+    @Getter
+    final String description = "Gradle 9.4 deprecates the Groovy `DomainObjectCollection.findAll(Closure)` overload on " +
+            "containers such as `tasks`, `configurations`, and `sourceSets`. It is replaced by the lazy " +
+            "`DomainObjectCollection.matching(Spec)`, which returns a live collection that only filters " +
```
steve-aom-elliott: I'm looking at
- https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api/-domain-object-collection/index.html#388720099%2FFunctions%2F-1793262594
(perhaps the wrong place)

Would it be more accurate to say you're replacing it with `DomainObjectCollection.matching(Closure)` given I don't believe we're actually altering the shape of the argument that it is passed, just which method is used?

steve-aom-elliott (approved):

PR #8008, opened 2026-06-14
Fix Java 25 parser treating named classes as unnamed when a bare `;` precedes the class declaration

barbulescu:
## What's changed?
The fix is to use Flags.IMPLICIT_CLASS — the compiler's definitive flag for unnamed classes — rather than the source-text heuristic.

## What's your motivation?
migrate AssertJ javadoc to markdown

@timtebeek 

deleting manually the ;

### Checklist


timtebeek:
Thanks again for the help here @barbulescu !

PR #8009, opened 2026-06-14
Fix shortening java.lang annotations

### LINKED ISSUE #6066: ImportService: `shortenFullyQualifiedTypeReferencesIn` does not work for `java.lang`
**Rewrite-Maven-Plugin:** 6.18.0
**Rewrite-Recipe-Bom:** 3.14.1

I am writing my own recipe and use the provided testing framework to make sure it behaves as expected.

It is not easy to trigger the method in question without a lot of code, so I have created a small example with the dropwizard recipe where [it is called as part of the cleanup](https://github.com/openrewrite/rewrite-dropwizard/blob/main/src/main/java/org/openrewrite/java/dropwizard/annotation/AddClassAnnotationVisitor.java#L73). I have seen the same behavior in my own recipes, however, and have no doubt it will affect other recipes as well.

```java
package org.openrewrite.maven;

import static org.openrewrite.java.Assertions.java;

import org.openrewrite.java.dropwizard.annotation.AddClassAnnotationIfAnnotationExists;

class ProofOfConcept implements RewriteTest {

  @Test
  void addJavaLang() {
    rewriteRun(
        spec ->
            spec.recipe(
                new AddClassAnnotationIfAnnotationExists("java.lang.Deprecated", "lombok.Value", null)),
        java(
            """
                        import lombok.Value;

                        @Value
                        public class Foo {
                        }
                        """,
            """
                        import lombok.Value;

                        @Deprecated
                        @Value
                        public class Foo {
                        }
                        """));
  }

    @Test
    void addLombok() {
        rewriteRun(
                spec ->
                        spec.recipe(
                                new AddClassAnnotationIfAnnotationExists("lombok.Value","java.lang.Deprecated", null)),
                java(
                        """
                                    @Deprecated
                                    public class Foo {
                                    }
                                    """,
                        """
                                    import lombok.Value;
            
                                    @Deprecated
                                    @Value
                                    public class Foo {
                                    }
                                    """));
    }
}
```

Both tests should succeed. The recipe should:

- See the first annotation and add the second
- Proceed to clean up and remove the fully qualified types where not necessary

## What did you see instead?

- The second test adds `@lombok.Value` to the class, but shortens it to `@Value` because the corresponding import makes the fully qualified name unneccessary. **This is the expected behavior as I understand it.**
- The first test adds `@java.lang.Deprecated` to the class, but fails to shorten it to `@Deprecated`. My assumption is that this happens with all annotations and classes that do not require an explicit import, not just with annotations.

As I have seen the same behavior in my far simpler recipes, I am reasonably sure that "shortenFullyQualifiedTypeReferencesIn" is misbehaving here.

I wouldn't know where to start on this one, honestly. Understanding visitors and LSTs still makes my brain hurt.

sakshichitnis27:
- Fixes #6066

Added regression coverage for ImportService.shortenFullyQualifiedTypeReferencesIn shortening java.lang annotation references such as @java.lang.Deprecated to @Deprecated.

Also added a same-package conflict case to ensure java.lang annotations remain fully qualified when shortening would change resolution.

Tested with:
./gradlew :rewrite-java-test:test --tests "org.openrewrite.java.ShortenFullyQualifiedTypeReferencesTest" --rerun-tasks

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ShortenFullyQualifiedTypeReferencesTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ShortenFullyQualifiedTypeReferencesTest.java
@@ -26,6 +26,7 @@ import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.withSourceTypesOnClasspath;
 import static org.openrewrite.test.RewriteTest.toRecipe;
 
 class ShortenFullyQualifiedTypeReferencesTest implements RewriteTest {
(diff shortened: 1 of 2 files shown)


greg-at-moderne:
Thanks for the contribution!
I appreciate the tests.

PR #8011, opened 2026-06-15
Python: only force the [profiling] pip extra when profiling is enabled

timtebeek:
The Moderne CLI bootstrapped the pinned `openrewrite` package with the `[profiling]` extra unconditionally, which pulls in `pyroscope-io` — a package that ships no wheels for some platforms (e.g. Windows/CPython 3.14) and so forces an unbuildable native (Rust) source build, aborting the whole install. This changes `PythonRewriteRpc$Builder.bootstrapOpenrewrite` to install the plain `openrewrite==<version>` package by default and only request the `[profiling]` extra when profiling is actually active (a non-blank `PYROSCOPE_SERVER_ADDRESS` in the subprocess environment), with a fallback to the plain package if the extra-install fails. The pip-running logic was extracted into small testable helpers (`openrewriteInstallSpecs`, `pipInstallCommand`, `profilingEnabled`, `installPackage`) while preserving all existing behavior (`--upgrade`, `--target`, env passthrough, log redirect, 2-minute timeout, exit-code check, `.openrewrite-version` marker). A new `PythonRewriteRpcBuilderTest` asserts the gating and fallback ordering. The pinned version and public Builder API are unchanged.

### DIFF
--- a/rewrite-python/src/main/java/org/openrewrite/python/rpc/PythonRewriteRpc.java
+++ b/rewrite-python/src/main/java/org/openrewrite/python/rpc/PythonRewriteRpc.java
@@ -767,48 +767,18 @@ public class PythonRewriteRpc extends RewriteRpc {
             try {
                 Files.createDirectories(pipPackagesPath);
 
-                // Install with the [profiling] extra so pyroscope-io is available when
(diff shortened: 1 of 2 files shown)


timtebeek:
Replaced with
- https://github.com/openrewrite/rewrite/pull/8064

PR #8016, opened 2026-06-15
Build `J.Assignment` directly in `UsePropertyAssignmentSyntax`

steve-aom-elliott:
## What's wrong

`UsePropertyAssignmentSyntax` converts Gradle property method-call syntax (e.g. `group 'x'`) to assignment syntax (`group = 'x'`) through `GroovyTemplate`, which builds the replacement by re-parsing a stub via `GroovyBlockStatementTemplateGenerator` → `JavaTemplateParser`.

The generator picks its wrapping from the matched `J.MethodInvocation`'s method type:

- **void** return type → instance-initializer block `class Template {{ group = "x" }}` (parses fine)
- **null or non-void** → field initializer `class Template { Object o = group = "x"; }` — the Groovy parser rejects a bare identifier as the LHS of a class-body field initializer

When method types are resolved against a real Gradle classpath, a top-level `group '...'` can take the field-initializer branch and fail with `IllegalArgumentException: Could not parse as Java`. (Unit tests don't catch it because without a resolved classpath the method type is often null and the stub happens to parse for `version`/`description`.)

## The fix

The transformation is purely structural (`name arg` → `name = arg`, or `select.name arg` → `select.name = arg`), so synthesize the `J.Assignment` directly from the matched invocation's name, select, and argument instead of round-tripping through the template parser. This:

- removes the failure mode entirely — no parser, no void-vs-non-void wrap branch, no classpath sensitivity;
- preserves the original argument LST (types, markers, formatting), which matters for non-string arguments such as the existing `version computeVersion()` test case.

## Tests

Adds regression tests for a top-level `group '...'` and a `project.group '...'` selector exercising the `getSelect() != null` branch. `UsePropertyAssignmentSyntaxTest` passes (15/15).

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/UsePropertyAssignmentSyntax.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/UsePropertyAssignmentSyntax.java
@@ -18,9 +18,12 @@ package org.openrewrite.gradle;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.*;
-import org.openrewrite.groovy.GroovyTemplate;
 import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JLeftPadded;
+import org.openrewrite.java.tree.Space;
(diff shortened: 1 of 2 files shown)


sambsnyd (approved):
I don't have a problem with the recipe being implemented in this way, but it would be nice if separately groovy template could handle this case successfully

PR #8018, opened 2026-06-16
rewrite-csharp: add LOCAL_NUGET_FEED env var and skip repo nuget.config on tool install

macsux:
Installs the C# tool from a fresh working directory with no `nuget.config` on any ancestor path, so NuGet's own config discovery applies the user/machine-level global configuration (preserving nuget.org / private-proxy feeds for transitive dependencies) while skipping repo-level overrides that could `<clear/>` sources or enable package source mapping. The tool package's feed now comes from `LOCAL_NUGET_FEED` when set, otherwise the NuGet global packages cache (`~/.nuget/packages/...`) as before. This also fixes RPC edit batching to resolve the canonical (codec-registered) source file type rather than the raw runtime class name, since under lazy-deserialization LSTs the runtime class is a generated proxy absent from the remote's declared languages, which would otherwise skip every batched edit on that file.

### DIFF
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
@@ -480,6 +480,7 @@ public class CSharpRewriteRpc extends RewriteRpc {
         }
 
         private void installTool(Path dotnetPath, String version, Path toolPath) {
+            Path installCwd = null;
             try {
                 Files.createDirectories(toolPath);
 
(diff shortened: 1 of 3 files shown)


rewrite-csharp/src/main/java/org/openrewrite/csharp/rpc/CSharpRewriteRpc.java
shanman190, line 506: Should this have instead been the `recipeInstallDir` rather than `workingDirectory`?

macsux: Recipe install dir is the source where nuget packages live. This usually where nuget.config is pointing to. We're solving for an issue where we're executing tool install command unrelated to repo being built, which may supply its own nuget.config and screw up this unrelated operation. We're resetting current workdir for this one command to one that is unlikely to have any nuget.config file that will interfere with it.

PR #8034, opened 2026-06-16
Make Maven coordinate matching case-sensitive

knutwannheden:
## Motivation

Maven `groupId` and `artifactId` coordinates are case-sensitive, but `ResolvedDependency` (and its siblings `ResolvedPom` and `MavenResolutionResult`) matched them with `StringUtils.matchesGlob(...)`, which is case-insensitive: it compares characters via `Character.toUpperCase(ch) != Character.toUpperCase(other)`. This is both semantically wrong — case-insensitive matching can produce false coordinate matches — and a measurable hot spot. In a JFR profile of a large recipe run (`UpgradeSpringBoot_4_0` over ~11,700 sources), `CharacterDataLatin1.toUpperCase` reached ~3.5% of on-CPU time, dominated by `StringUtils.different` reached through `matchesGlob` from `ResolvedDependency`.

Because the second argument to `matchesGlob` is a glob pattern (callers may pass `*`/`?`), this can't simply become `equals()` — it needs case-sensitive matching that still honors glob wildcards.

## Examples

```java
// New case-sensitive overload (existing 2-arg form is unchanged and remains case-insensitive):
StringUtils.matchesGlob("com.example", "com.example", true);   // true
StringUtils.matchesGlob("com.example", "Com.Example", true);   // false — case matters
StringUtils.matchesGlob("com.fasterxml.jackson", "com.fasterxml.*", true); // true — wildcards still work
StringUtils.matchesGlob("com.fasterxml.jackson", "Com.*", true);           // false

// Maven coordinate lookups are now case-sensitive:
resolvedDependency.findDependency("com.google.guava", "guava"); // matches
resolvedDependency.findDependency("Com.Google.Guava", "guava"); // no longer matches
```

## Summary

- Added `StringUtils.matchesGlob(String, String, boolean caseSensitive)`. The existing 2-arg `matchesGlob` now delegates to it with `caseSensitive = false`, so its behavior is unchanged for all current callers.
- Threaded the flag down to the per-character comparison; in the case-sensitive path the compare is a plain `ch != other` (no `toUpperCase`).
- `different(ch, other, caseSensitive)` now short-circuits with `if (ch == other) return false;` before the `toUpperCase` pair. Equal characters are the common case (matching coordinate prefixes), so this eliminates most `toUpperCase` calls for **all** `matchesGlob` callers, with zero semantic change — a strictly-safe win independent of the case-sensitivity change.
- Switched the coordinate-matching call sites to the case-sensitive variant: `ResolvedDependency.findDependency0` / `findDependencies0` (groupId, artifactId, and exclusion matching), `ResolvedPom` effective-exclusion matching, and `MavenResolutionResult.findDependencies`.

Real-world `groupId`/`artifactId` are lowercase by convention, so the functional impact of the case-sensitivity change should be nil in practice; it only changes behavior for coordinates that differ only in case.

## Test plan

- Added `StringUtilsTest#globMatchingCaseSensitive` asserting case-sensitive literal and wildcard matching, and that the 2-arg overload stays case-insensitive.
- Added `MavenParserTest#coordinateMatchingIsCaseSensitive` asserting that `findDependencies`/`findDependency` match exact-case and case-sensitive globs but not mismatched case.
- `./gradlew :rewrite-core:test :rewrite-maven:test` — both green; no existing test relied on case-insensitive coordinate matching.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/StringUtils.java
@@ -447,8 +447,27 @@ public class StringUtils {
      * @param pattern the glob pattern to evaluate. null and "*" both match everything.
      * @return true if the input string matches the glob pattern, false otherwise
      * @see org.openrewrite.PathUtils#matchesGlob(Path, String)
+     * @see #matchesGlob(String, String, boolean)
      */
     public static boolean matchesGlob(@Nullable String str, @Nullable String pattern) {
+        return matchesGlob(str, pattern, false);
+    }
+
+    /**
+     * Checks if a given string matches a specified glob pattern, optionally honoring case. A glob
+     * pattern may include special characters such as '*' to represent any sequence of characters
+     * and '?' to represent any single character.
+     * <p>
+     * For file path matching, use {@link org.openrewrite.PathUtils#matchesGlob(Path, String)},
+     * which properly interprets '*' and '**' wildcards for file paths.
+     *
+     * @param str the input string to match against the pattern, can be null
+     * @param pattern the glob pattern to evaluate. null and "*" both match everything.
+     * @param caseSensitive whether literal characters must match exactly with respect to case
+     * @return true if the input string matches the glob pattern, false otherwise
+     * @see org.openrewrite.PathUtils#matchesGlob(Path, String)
+     */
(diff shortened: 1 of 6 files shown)


timtebeek (approved):

knutwannheden:
Probably won't be necessary with #8030 anymore.

Issue #1536, opened 2022-03-30
org.openrewrite.java.cleanup.RemoveUnusedPrivateMethods removed a method that is used


yeikel:
Example to reproduce the problem : 

```java 


import java.util.Map;

public class TestClass  {

	public static void main(String[] args){
		final String test = addParameter("path1", Map.of("Test","value"), "",1);

		System.out.println(test);

		final String test2 = addParameter("path2", Map.of("",1), "",1);

		System.out.println(test2);

	}

	private static <T> String addParameter(String path, final Map<String, T> placeholderValues, final T value, final Integer counter) {
		final String parameterPath = counter == null ? path : path + "_"+counter;
		final String placeholder = 	String.format("${%s}",parameterPath);
		placeholderValues.put(placeholder, value);
		placeholderValues.put(parameterPath, value);
		return placeholder;
	}
}


```

Output

```
[WARNING] Changes have been made to src/main/java/TestClass.java by:
[WARNING]     org.openrewrite.java.cleanup.RemoveUnusedPrivateMethods

```

Diff : 

![image](https://user-images.githubusercontent.com/8935151/160911342-d8c0a5cb-3396-4cfb-9fd7-c3fdfe3c7439.png)

```diff 
+++ b/src/main/java/TestClass.java
@@ -13,12 +13,4 @@ public class TestClass  {
                System.out.println(test2);
 
        }
-
-       private static <T> String addParameter(String path, final Map<String, T> placeholderValues, final T value, final Integer counter) {
-               final String parameterPath = counter == null ? path : path + "_"+counter;
-               final String placeholder =      String.format("${%s}",parameterPath);
-               placeholderValues.put(placeholder, value);
-               placeholderValues.put(parameterPath, value);
-               return placeholder;
-       }
 }
```

PR #8035, opened 2026-06-16
Add TypeUtils.isOfTypeIgnoringGenerics for matching invocations to declarations

timtebeek:
`JavaSourceFile#getTypesInUse().getUsedMethods()` is the standard way recipes find usages of a method declaration, but the `JavaType.Method` recorded for an invocation can diverge from the declaration's instance: an unchecked invocation (raw arguments) erases the result type to its raw form (e.g. `Set<String>` becomes raw `Set`), and a call binding a generic parameter records the concrete type — so `declaration.equals(used)` and `TypeUtils.isOfType` return false even though there's no ambiguity about which declaration the invocation resolves to. This adds `TypeUtils.isOfTypeIgnoringGenerics(JavaType.Method, JavaType.Method)`, a comparison that matches two method types insensitive to generic parameterization: a raw type matches its parameterized form and a generic type variable matches any type. The LST attribution is left untouched, so the call-site generic information on invocations is preserved. Recipes such as `RemoveUnusedPrivateMethods` can use this instead of hand-rolled bidirectional matchers, which fixes both the raw-collection case (openrewrite/rewrite-static-analysis#877) and generic methods that look unused when invoked (#1536).

### DIFF
--- a/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
+++ b/rewrite-java-tck/src/main/java/org/openrewrite/java/tree/TypeUtilsTest.java
@@ -185,6 +185,154 @@ class TypeUtilsTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite-static-analysis/issues/877")
+    @Test
+    void isOfTypeIgnoringGenericsMatchesRawCollectionInvocation() {
+        rewriteRun(
+          java(
+            """
+              import java.util.Collection;
+              import java.util.Map;
+              import java.util.Set;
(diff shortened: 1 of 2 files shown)


sambsnyd:
So we would be throwing away true information about the generic types of method invocations?  I don't love that. How about a comparison method in `TypeUtils` or on `JavaType.Method` itself which can be used for comparisons insensitive to generic parameterization?

timtebeek:
Thanks for the suggestion! Would you be ok with the current approach then? I'll work it in downstream if OK-ed and merged.

sambsnyd (approved):

PR #8036, opened 2026-06-16
Groovy: Added visitor for do-while loops

gtiwari333:
## What's changed?
  GroovyParserVisitor.java — Added a visitDoWhileLoop() method that:

  1. Consumes the do keyword whitespace via sourceBefore("do")
  2. Visits the loop body via doVisit(loop.getLoopBlock())
  3. Consumes the while keyword whitespace via sourceBefore("while")
  4. Wraps the condition expression in a J.ControlParentheses with proper padding around ( and )
  5. Constructs and enqueues a J.DoWhileLoop LST node with the body right-padded by the while keyword, and the left-padded condition

## What's your motivation?
Working on migrating old groovy based project.

N/A

N/A

No workaround

## Any additional context
  This follows the same pattern as the existing visitWhileLoop() method just below it, which handles while loops.

### Checklist


greg-at-moderne:
Thanks so much for this contribution too.

PR #8038, opened 2026-06-16
Groovy: fix parsing of groovy code that has default method with 'this' keyword

gtiwari333:
## What's changed?
In visitMethodCallExpression(), added a check after the normal select resolution: when no select was produced but the source has `this.` at the cursor position, it manually constructs a `this` identifier as the select, consumes `this` from source, and captures the `.` whitespace.  

Added test to showcase the bug and fix.

 

### Checklist


greg-at-moderne:
Thanks for the contributions!

PR #8043, opened 2026-06-17
Java: support parameterized types in `ChangeMethodInvocationReturnType`

### LINKED ISSUE #7502: ChangeMethodInvocationReturnType is not handling parameterized types
rewrite 8.75.2

Add this test case in `ChangeMethodInvocationReturnTypeTest`

```java
    @Test
    void replaceVariableAssignmentWithGenericReturnType() {
        rewriteRun(
          spec -> spec.recipe(new ChangeMethodInvocationReturnType("bar.Bar bar()", "java.util.Set<java.lang.String>"))
            .parser(JavaParser.fromJavaVersion()
              //language=java
              .dependsOn(
                """
                  package bar;
                  import java.util.List;
                  public class Bar {
                      public static List<String> bar() {
                          return null;
                      }
                  }
                  """
              )
            ),
          //language=java
          java(
            """
              import bar.Bar;
              import java.util.List;
              class Foo {
                  void foo() {
                      List<String> one = Bar.bar();
                  }
              }
              """,
            """
              import bar.Bar;

              import java.util.Set;

              class Foo {
                  void foo() {
                      Set<String> one = Bar.bar();
                  }
              }
              """
          )
        );
    }
```

Result:

Image

MBoegers:
- Closes #7502. Supersedes #7509 — thanks @timtebeek for the original approach and reproducers.

### Problem
`ChangeMethodInvocationReturnType` breaks on generic return types. `JavaType.buildType("java.util.Set<java.lang.String>")` doesn't understand generics, so it folds the `<...>` suffix into a malformed `ShallowClass` and the recipe emits a bare `J.Identifier` whose simple name still contains `<...>` — non-compilable output with missing imports for the type parameters.

### Change
- Add `TypeUtils.buildTypeTree(String)` — a single `String` → `TypeTree` builder supporting parameterized types with nested type arguments (`Map<String, List<Integer>>`), arrays, wildcards (`?`, `? extends X`, `? super X`), primitives and `java.lang` shortening. Its `JavaType` is resolved with the same parser, so `buildTypeTree(name).getType()` yields the matching type.
- Consolidate the near-duplicate hand-rolled `AddMethodParameter.createTypeTree` into `buildTypeTree`, fixing latent gaps: depth-aware splitting of multi-argument nested generics, a space after each comma, and `? super` bounds.
- `ChangeMethodInvocationReturnType` now builds a proper parameterized type expression, derives the new `JavaType` via `getType()`, shortens the fully-qualified tree via `ImportService` (importing the raw type and every type parameter), and recursively removes imports for the replaced type. The existing `initializedByMatch` guard is preserved.

The core `JavaType`/`TypeTree` interfaces are intentionally left untouched; the logic lives in `TypeUtils`. Promoting it into `JavaType.buildType` later would be a clean follow-up.

### Test plan
- [x] Added the repro (`List<String>` → `Set<String>`), nested generics (`Map<String, List<Integer>>`), a depth-stressing case (`Map<String, Map<Integer, Long>>`) that the old naive split breaks, and parameterized → raw (`List<String>` → `Object`).
- [x] `./gradlew :rewrite-java:test` passes; `AddMethodParameterTest`, TCK `TypeTreeTest`, `ChangeTypeTest` green; `licenseFormat` clean.

closing https://github.com/moderneinc/customer-requests/issues/2434


rewrite-java/src/main/java/org/openrewrite/java/tree/TypeUtils.java
```diff
@@ -97,6 +99,190 @@ public static boolean isObject(@Nullable JavaType type) {
...
+            return new J.ArrayType(
+                    randomId(),
+                    Space.EMPTY,
+                    Markers.EMPTY,
+                    elementType,
+                    null,
+                    JLeftPadded.build(Space.EMPTY),
+                    buildType(typeName)
+            );
+        }
```
sambsnyd: The challenge I have with this mechanism is that it is not classpath-aware. If you use this to build a type tree for anything not in the standard library you get a stubbed / shallow type.
Is there a reason a type tree produced by a context-insensitive `JavaTemplate` will not serve?

MBoegers: You were right, and I had another deep thinking here. It should now do the trick.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
```diff
@@ -80,48 +80,38 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations m
...
-                        JavaType.FullyQualified varType = TypeUtils.asFullyQualified(var.getType());
-                        if (varType != null && !varType.equals(newType)) {
-                            return var.withType(newType).withName(var.getName().withType(newType));
-                        }
-                        return var;
-                    }));
+                // `ShortenFullyQualifiedTypeReferences` deliberately leaves `java.lang.*` fully qualified, so strip it here.
+                String templateType = newReturnType.replaceAll("\\bjava\\.lang\\.([A-Z][A-Za-z0-9_]*)(?![.A-Za-z0-9_])", "$1");
+                J.VariableDeclarations resolved = JavaTemplate.builder(templateType + " __cmirt__")
+                        .contextSensitive()
```
timtebeek: I have a slight bit of worry that we don't have tests for this part: all the types swapped in are just from the JDK. That might not hold up in practice when folks write their own custom migrations.

If you're confident it works I'm ok to merge.

timtebeek (approved):

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
```diff
@@ -80,48 +80,38 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations m
...
-                    mv = mv.withVariables(ListUtils.map(mv.getVariables(), var -> {
-                        JavaType.FullyQualified varType = TypeUtils.asFullyQualified(var.getType());
-                        if (varType != null && !varType.equals(newType)) {
-                            return var.withType(newType).withName(var.getName().withType(newType));
-                        }
-                        return var;
-                    }));
+                // `ShortenFullyQualifiedTypeReferences` deliberately leaves `java.lang.*` fully qualified, so strip it here.
+                String templateType = newReturnType.replaceAll("\\bjava\\.lang\\.([A-Z][A-Za-z0-9_]*)(?![.A-Za-z0-9_])", "$1");
+                J.VariableDeclarations resolved = JavaTemplate.builder(templateType + " __cmirt__")
```
sambsnyd: What is the meaning of `"__cmirt__"` ?

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
```diff
@@ -80,48 +80,38 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations m
...
-
-                    mv = mv.withVariables(ListUtils.map(mv.getVariables(), var -> {
-                        JavaType.FullyQualified varType = TypeUtils.asFullyQualified(var.getType());
-                        if (varType != null && !varType.equals(newType)) {
-                            return var.withType(newType).withName(var.getName().withType(newType));
-                        }
-                        return var;
-                    }));
+                // `ShortenFullyQualifiedTypeReferences` deliberately leaves `java.lang.*` fully qualified, so strip it here.
+                String templateType = newReturnType.replaceAll("\\bjava\\.lang\\.([A-Z][A-Za-z0-9_]*)(?![.A-Za-z0-9_])", "$1");
```
sambsnyd: `ShortenFullyQualifiedTypeReferences` only deliberately leaves `java.lang` when there is a name conflict. For example it would leave `java.lang.Deprecated` if a user-defined type `com.foo.Deprecated` existed on the classpath. 

So this comment is either indicative of a bug in `ShortenFullyQualifiedTypeReferences` not shortening something it should, or this recipe shortening something it shouldn't.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
```diff
@@ -129,4 +141,75 @@ private boolean isInitializedByMatch(@Nullable Expression expression) {
             }
         });
     }
+
+    /**
+     * Synthesize a minimal source stub (with the correct type-parameter arity) for every non-JDK type
+     * referenced in {@code type}, so a {@link JavaTemplate} parser can attribute it to a real class without
+     * the declaring artifact on the classpath. JDK types (prefixed {@code java.}) are left to the default parser.
+     */
+    private static List<String> typeStubs(String type) {
```
timtebeek: Let's minimized these comments and instead use a more descriptive method name, like `synthesizeStubsForTypeAttribution`.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
MBoegers: tested it to be sure on a real project, turns out there is a better way.

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
MBoegers: Its just a placeholder, we are interested in t he parsed types. i'll change it to `__typePlaceholder__`

rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodInvocationReturnType.java
MBoegers: There is even a test for this that requires this. it was added with https://github.com/openrewrite/rewrite/commit/2ea58e616 as one precise commit, so this is on purpose. Let me change the test and not shorten `java.lang.*` as it is the expected OR behavior.

https://github.com/openrewrite/rewrite/blob/7988fc0392597885da9078c02bcddcfa384a3a3b/rewrite-java-test/src/test/java/org/openrewrite/java/ShortenFullyQualifiedTypeReferencesTest.java#L373-L386

PR #8046, opened 2026-06-17
Experiment: opt-in Cursor pooling (-Dorg.openrewrite.cursorPool)

jkschneider:
> **Status: experiment — opening to preserve the work, then closing immediately.**
> It is not ready to merge and may never be in this form (see *Why this is parked*).
> Default-off, so it changes nothing unless `-Dorg.openrewrite.cursorPool=true` is set.

## Motivation

Allocation profiling of large, visitor-heavy recipe runs (a Spring Boot migration over
the `spring-framework` codebase) shows `new Cursor(...)` is **~42% of all allocation**.
It is pure churn — one cursor per visited node per pass — and a cursor is logically dead
the instant the visitor ascends above its node. Trees are parsed once, so this is
allocation *rate*, not footprint.

## What this does

A per-thread free list that recycles the cursor the visitor creates on each descent
instead of allocating a new one:

- `Cursor.parent/value` become non-final + a package-private `reset(parent, value)`.
- `CursorPool` — `ThreadLocal<ArrayDeque<Cursor>>`, `acquire()/release()`. Roots are
  never recycled (held for the lifetime of a run). Per-thread because a single node's
  visit (acquire..release) always runs on one thread.
- `TreeVisitor.pushCursor/popCursor` encode acquire-on-descend / release-on-ascend,
  wired into `TreeVisitor.visit` and `JavaVisitor.visitRightPadded/visitLeftPadded/visitContainer`
  (together ~95% of cursor allocation by call site: visit 60%, right-padded 22%,
  left/container 7%/7%; the remaining ~4% are root cursors, not poolable).

Off by default; when off the helpers fold to the original `new Cursor` /
`setCursor(getParent())` (the flags are `static final`).

## Results (spring-framework, identical jar, flag the only variable)

| metric | pool off | pool on |
|---|---|---|
| Cursor share of allocation | 41.6% | **2.9%** (residual is ~all non-poolable root cursors) |
| GC pause count | 5620 | **3762 (-33%)** |
| wall-clock | 627.5s | **632.3s (+0.8%, flat)** |
| max RSS | 1455 MB | 1501 MB |

The allocation and GC-pause win is large and real, but **does not convert to wall-clock**:
G1 collection is largely concurrent, GC was only ~22% of CPU to begin with, and the
run's actual cost is elsewhere (type attribution, version/dependency resolution).

## Why this is parked

Comparing the set of files each run changed, **the pool-on run silently skipped 32
transformations** that pool-off applied (a strict subset; reproducible across runs).
All 32 are the instanceof-pattern rewrite, and the cause is a contract violation that
pooling makes fatal:

Recycling is only sound when a cursor that must outlive its visit is copied via
`Cursor.detach()`. A recipe that instead stashes a **live** `Cursor` in a long-lived
collection gets miscompiled under the pool — the recycled cursor's `equals/hashCode`
(parent + value) mutate, so the later lookup misses and the transformation is skipped.
`InstanceOfPatternMatch` (in `rewrite-static-analysis`) does exactly this:

```java
private final Map<J.InstanceOf, Set<Cursor>> contextScopes; // live cursors as set members
```

There is no reason to believe this is the only such site across the recipe ecosystem.
**A diagnostic `CursorEscapeDetector` exists** (committed earlier on this branch, opt-in
via `-Dorg.openrewrite.cursorEscapeCheck=true`) and confirms leaf cursors are never
retained, but it samples via weak refs and could not pin the coarse `CompilationUnit`/
`Block` retainers; it only instruments `TreeVisitor.visit`, not the padding cursors.

Net: an allocation-only optimization with no wall-clock benefit on the workload that
motivated it, and a real correctness hazard until cursor retention is universally
converted to `detach()`. Not worth shipping right now.

## If revived later — follow-up study

1. **Make output-safety provable, not hoped-for.** Improve the detector to *pin*
   retainers (heap-dump/MAT or a strong-ref breadcrumb), extend it to the padding
   cursors, and run it as a gate over the recipe ecosystem to enumerate every live-cursor
   retention. Then `detach()` each (same fix already applied to `VariableNameUtils` /
   `RemoveUnusedImports` on this branch; `InstanceOfPatternMatch.contextScopes`,
   `VariableNameStrategy.contextScopes`, and the `java/trait/Literal` cursor field are
   known next targets).
2. **Find a workload where the GC win matters.** This one isn't GC-throughput-bound;
   pooling may pay off on allocation-bound / memory-constrained runs. Measure before
   investing further.
3. **The unit suite is not a sufficient gate.** `rewrite-core` + `rewrite-java` were
   green under the flag (incl. the once-hung `SourcePositionServiceTest`); only the
   at-scale changed-file diff exposed the regression. Any future attempt needs a
   large-corpus output-equivalence gate.

### Engineering note (the one genuinely tricky bug)

`popCursor` recycles the cursor the frame *acquired at push*, never the current `cursor`
field, because a visit may legitimately re-point its own cursor: `SourcePositionService`'s
printer adopts the caller's cursor via `setCursor`, and `updateCursor` swaps in a fresh
value. Recycling the field instead returned a still-live foreign cursor to the pool and,
via the LIFO free list, made it its own child's parent — a 2-cycle that span forever in
`getParent()`-walking code. The fix is general against any `setCursor`/`updateCursor`
re-pointing.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/Cursor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Cursor.java
@@ -50,6 +50,16 @@ public class Cursor {
         this.value = value;
     }
 
+    /**
+     * Create a standalone copy of this cursor and its entire parent chain, detached from the traversal
+     * that produced it. Use this whenever a cursor (or its path) must outlive the visit of its node —
+     * e.g. when storing it as a long-lived map key or in a field — so that it stays valid even if the
+     * originating cursors are later reused/recycled by the visitor. Messages are not copied.
+     */
+    public Cursor detach() {
+        return new Cursor(parent == null ? null : parent.detach(), value);
+    }
+
     public Cursor getRoot() {
         Cursor c = this;
         while (c.parent != null) {
--- a/rewrite-core/src/main/java/org/openrewrite/Cursor.java
+++ b/rewrite-core/src/main/java/org/openrewrite/Cursor.java
@@ -37,10 +37,10 @@ public class Cursor {
     public static final String ROOT_VALUE = "root";
 
     @Nullable
-    private final Cursor parent;
+    private Cursor parent;
 
     @With
-    private final Object value;
+    private Object value;
 
     @Nullable
     private Map<String, Object> messages;
@@ -48,6 +48,22 @@ public class Cursor {
     public Cursor(@Nullable Cursor parent, Object value) {
         this.parent = parent;
         this.value = value;
+    }
+
+    /**
+     * Re-point this cursor at a new position so a {@link CursorPool} can recycle the object
+     * instead of allocating a fresh {@code new Cursor(...)} for every visited node. Messages
+     * are dropped, matching a newly constructed cursor.
+     * <p>
+     * Package-private and intended <strong>only</strong> for the pool driven by the visitor's
+     * own push/pop. Any cursor that must outlive the visit of its node must instead be copied
+     * with {@link #detach()} (which allocates a separate, never-pooled object), so that the
+     * cursor being reset here is provably dead.
+     */
+    void reset(@Nullable Cursor parent, Object value) {
+        this.parent = parent;
+        this.value = value;
+        this.messages = null;
     }
 
     /**
--- a/rewrite-core/src/main/java/org/openrewrite/CursorPool.java
+++ b/rewrite-core/src/main/java/org/openrewrite/CursorPool.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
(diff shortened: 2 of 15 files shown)


jkschneider:
Closing immediately — parking this as a documented launch point, not merging. The allocation win is real (Cursor 41.6%→2.9%, GC pauses −33%) but it does not move wall-clock on the motivating workload (+0.8%, flat), and the pool silently skips 32 transformations under it because recipes retain live cursors without `detach()` (e.g. `InstanceOfPatternMatch.contextScopes`). Not worth the correctness risk for no latency benefit right now. The branch + this PR preserve the design, the cycle-hang fix, and the follow-up study list for a future attempt.

PR #8048, opened 2026-06-18
Groovy: fix IndexOutOfBoundsException when recipe adds @Annotation to a method in anonymous class

gtiwari333:
## What's changed?

It fixes  IndexOutOfBoundsException that was being thrown when a recipe adds @Annotation to a method in anonymous class.  

## What's your motivation?

One of the custom migration script is failing. Unit test shows the example of such recipe.
 

### Checklist


timtebeek (approved):
Really appreciate the many small and targeted fixes lately!

PR #8067, opened 2026-06-19
Gradle: avoid allocating a new GradleDependencies trait when nothing changes

Jenson3210:
## What's changed

`GradleDependencies.mapStatements` always allocated a brand-new `GradleDependencies` trait, even when the mapping produced no change to the underlying `J.MethodInvocation`. Because `ListUtils.mapFirst` and `J.MethodInvocation.withArguments` already return the identical instance when nothing changes, the wrapper allocation was wasteful and broke reference-identity expectations for callers (e.g. recipes that check whether a trait actually changed).

This change returns `this` when `withArguments` yields the same tree:

```java
J.MethodInvocation newTree = getTree().withArguments(arguments);
if (getTree() == newTree) {
    return this;
}
```

## Test

`GradleDependenciesTest#noNewTraitInstanceWhenNothingIsRemoved` calls `removeDependency` for a dependency that is not present and asserts the returned trait is the **same instance** (`isSameAs`).

- With the fix: the test passes.
- Reverting the fix (always constructing a new `GradleDependencies`): the test fails because a new instance is allocated.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/GradleDependencies.java
@@ -82,7 +82,11 @@ public class GradleDependencies implements Trait<J.MethodInvocation> {
         if (arguments.isEmpty()) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
❤️

PR #8082, opened 2026-06-19
Restore dependency management for child modules of a local parent

timtebeek:
## Summary

- `ChangeParentPom` / `UpgradeParentVersion` now restore dropped dependency management in a local parent on behalf of the child modules that declare those dependencies, not just the parent's own dependencies.

## Problem

When upgrading a parent (e.g. Spring Boot 2.7 → 3.0), the recipe restores dependency management for dependencies that the new parent no longer manages. It only looked at dependencies declared by the pom whose `<parent>` was being changed.

In a multi-module project where a **local parent** declares the upgraded parent (e.g. `spring-boot-starter-parent`) and the **child modules** declare the affected dependencies without an explicit version, nothing was restored — the local parent declares none of those dependencies itself, and the child modules' direct parent (the local parent) isn't what's being changed. The result was unresolvable dependencies and module resolution errors after the upgrade.

- Reported in moderneinc/customer-requests#2589.

## Solution

In addition to the pom being changed, also inspect descendant modules that inherit from it (gathered from the scanning accumulator's resolution markers). Dependency management that the new parent drops is restored in the local parent on behalf of any descendant module that declares the dependency without an explicit version. Restored entries are de-duplicated by `groupId:artifactId`. Single-module behavior is unchanged.

### Note for reviewers

The added test asserts the management is restored in the local parent (the fix). The child module additionally renders a transient `No version provided for direct dependency ...` warning marker: within a single run the child's in-memory model isn't re-resolved against the local parent's newly-added management (cross-module model propagation here is order-dependent, and only the parent version is propagated through the resolution markers today). This marker is not written to the pom file and is present on `main` for this scenario as well; it clears on the next parse, so the resulting build resolves correctly. Making that propagation order-independent would be a larger, separate change to the multi-module marker machinery.

## Test plan

- [x] Existing `ChangeParentPomTest` / `UpgradeParentVersionTest` pass
- [x] New `ChangeParentPomTest.RetainVersions#bringsDownRemovedManagedVersionDeclaredInChildModuleOfLocalParent` fails on `main`, passes with the fix

- Fixes moderneinc/customer-requests#2589


rewrite-maven/src/test/java/org/openrewrite/maven/ChangeParentPomTest.java
```diff
@@ -1564,6 +1564,112 @@ void bringsDownRemovedManagedVersion() {
...
+        @Issue("https://github.com/moderneinc/customer-requests/issues/2589")
+        @Test
+        void bringsDownRemovedManagedVersionDeclaredInChildModuleOfLocalParent() {
+            // A child module declares a dependency without an explicit version, relying on management
+            // inherited from Spring Boot through the local parent. When the local parent is upgraded to a
+            // Spring Boot version that no longer manages that dependency, the management must be restored
+            // in the local parent so the child still resolves; otherwise the build fails with resolution
+            // errors. (The transient warning marker rendered on the child below reflects the child's
+            // in-memory model not yet being re-resolved against the parent's newly added management; it
+            // is not written to the file and is cleared on the next parse. It is present on `main` too.)
```
MBoegers:
```suggestion
```

MBoegers (approved):

PR #8086, opened 2026-06-20
Groovy: unbounded wildcard in method is appending '?extends'

gtiwari333:
## What's changed?

Without the fix the `Map<String, ?> r = b(a1)` is transformed to `Map<String, ?extends> r = b(a1)` making it invalid code.

```
   // these two method together fails
    String a(def a1) {
        Map<String, ?extends> r = b(a1)
        return r.key
    }
    Map<String, ?> b(def b) {
    }
}
```

 

### Checklist


timtebeek (approved):
Thanks again! Good to see all that you've uncovered addressed.

greg-at-moderne:
Thanks for the working fix!

PR #8089, opened 2026-06-20
Groovy: fixing range with complex bound parsing `s[0..(n - 1)]`

gtiwari333:
## What's changed?
Fixing parsing of range with complex bound eg `s[0..(n - 1)]` is fixed

## What's your motivation?
The test would fail with following error without the fix:
```
Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser. 
def slice(String s, int n) {
    s[0..(n - 1)]
}
def slice(String s, int n) {
    s[0..(n - 1)]
}
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:388)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:159)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:154)
	at org.openrewrite.groovy.tree.RangeTest.rangeWithComplexBoundExpression(RangeTest.java:107)

```
 
### Checklist


timtebeek (approved):
Thanks for yet another fix! It seemed counter intuitive at first, but the change seems correct for this and other synthetic elements, so thanks for finding and fixing that.

timtebeek:
## How this fixes the issue

`G.Range` (e.g. `0..(n - 1)`) is a synthetic node with **no `type` field of its own**:
- `getType()` delegates: `return from.getType();`
- `withType(t)` distributes that type onto **both** bounds: `withFrom(from.withType(t)).withTo(to.withType(t))`

The old `visitRange` ended with `return r.withType(visitType(r.getType(), p));`, which reads the **left** bound's type and writes it onto **both** bounds. That broke the no-op visit invariant.

### Why only a *complex* bound triggered it

The failure surfaces through `WhitespaceValidationService`, which walks the tree via the inherited `visitRange` and asserts the result is reference-identical to the input (a no-op visit must not rebuild). Lombok `@With` short-circuits on `==`:

- **Simple bounds** `0..10`: both literals carry the same `JavaType.Primitive.Int` **singleton**, so `to.withType(sameRef)` returns `this` → no rebuild → passes. (That's why `allRangeKinds` never caught it.)
- **Complex bound** `(n - 1)`: `to` is a `J.Parentheses`/`J.Binary` whose type reference differs from the `from` literal's, so `to.withType(fromType)` rebuilds the subtree → new `Range` → tree identity changes → assertion fails.

The generic "non-whitespace in whitespace" message is misleading; the real defect is that `visitRange` wasn't idempotent (and, semantically, it was clobbering the right bound's type with the left bound's type).

### Risk of removing the `withType` line

Negligible — it's a net correctness improvement:
1. **Redundant** — `from` and `to` are already visited as children via `visitAndCast`, so type-remapping recipes update their types through the normal `visitLiteral`/`visitBinary` paths. Since `G.Range` has no type field of its own, nothing on the Range itself remains to attribute.
2. **Actively wrong** — distributing one incoming type to both bounds overwrote `to`'s type with `from`'s whenever they differed.

This aligns `visitRange` with the convention for synthetic nodes lacking their own type field (contrast `JavaVisitor.visitBinary`, where `J.Binary` *does* have a real `type` field). Both `allRangeKinds` and the new `rangeWithComplexBoundExpression` test pass.

PR #8094, opened 2026-06-22
rewrite-python: fix empty output when a Java recipe edits a Python LST (#8093)

### LINKED ISSUE #8093: rewrite-python: editing a Python LST from a Java recipe prints an empty file (regression in 8.84.8, #7958)

rewrite `8.84.8` and later (current `8.85.4`). Last good version: `8.84.7`.

A Java recipe that runs over a Python/Starlark LST (via `rewrite-python`'s `PythonParser`/`PythonVisitor`). Surfaced by `moderneinc/rewrite-bazel`, whose `BazelParser` simply delegates to `PythonParser`.

### What is the smallest example that reproduces the problem?

Any Java recipe that **modifies** a Python LST (e.g. adds a marker) prints an **empty** file as the result. Pure parse→print round-trips (no modification) are unaffected.

Minimal reproducer (only `rewrite-python` + `rewrite-test`, no Bazel):

```java
import org.openrewrite.TreeVisitor;
import org.openrewrite.python.PythonVisitor;
import static org.openrewrite.python.Assertions.python;

class PythonMarkerReproTest implements RewriteTest {
    static class MarkAnyCall extends Recipe {
        @Override public String getDisplayName() { return "Mark any call"; }
        @Override public String getDescription() { return "Adds a SearchResult marker to every method invocation."; }
        @Override public TreeVisitor<?, ExecutionContext> getVisitor() {
            return new PythonVisitor<ExecutionContext>() {
                @Override public J visitMethodInvocation(J.MethodInvocation m, ExecutionContext ctx) {
                    return SearchResult.found(super.visitMethodInvocation(m, ctx), "found");
                }
            };
        }
    }

    @Test
    void markerDoesNotEraseContent() {
        rewriteRun(
          spec -> spec.recipe(new MarkAnyCall()),
          python(
            """
            def f():
                print("hello")
            """,
            """
            def f():
                /*~~(found)~~>*/print("hello")
            """
          )
        );
    }
}
```

The marker is added and the file content is preserved:

```
def f():
    /*~~(found)~~>*/print("hello")
```

### What did you see instead?

The printed result is the **empty string** — the entire file content is dropped:

```
diff --git a/file.py b/file.py
index e621639..e69de29 100644
--- a/file.py
+++ b/file.py
@@ -1,2 +0,0 @@
-def f():
-    /*~~(found)~~>*/print("hello")
\ No newline at end of file
expected: "def f():\n    /*~~(found)~~>*/print(\"hello\")"
 but was: ""
```

### Diagnosis (bisection + isolation)

- **Regression introduced in `8.84.8`.** Pinning `rewrite-bom` + the `openrewrite` PyPI wheel:
  - `8.84.7` → PASS
  - `8.84.8` → FAIL (and every release since, through `8.85.4`)
- **Isolated to the Python (`openrewrite` PyPI) side, not the Java printer.** Mixing versions:
  - Java `8.84.8` + Python wheel `8.84.7` → **PASS**
  - Java `8.84.7` + Python wheel `8.84.8` → **FAIL**
- **Attributed to #7958** ("Python: reduce LST memory footprint (`__slots__`, shared empty lists, compact ids)"), merged 2026-06-10, the only Python change in `v8.84.7...v8.84.8`. It is a **Python-only** PR (0 `.java` files), and it reworked LST id representation (UUID → 128-bit int with lazy `.id`) and the RPC sender/receiver (`rpc/python_sender.py`, `rpc/python_receiver.py`).
- **The bug is specific to the Java↔Python RPC round-trip of a *modified* tree:**
  - The pure-Python harness (`RecipeSpec().rewrite_run(...)` with the equivalent marker-adding `PythonVisitor`) **passes** — Python-side parse, edit, and print are individually fine.
  - An **unchanged** parse→print round-trip through Java also passes.
  - Only a Java recipe that **edits** the Python LST and then prints produces empty output.

This points at the RPC serialization of an edited Python LST after the id/`__slots__`/shared-empty-list rework — the Java receiver/printer ends up with an empty (or detached) statement list once an edit is applied.

### Impact

- `moderneinc/rewrite-bazel` CI has been red on every run since 2026-06-11 (28 tests, all `Find*` recipes), because every search recipe that adds a marker prints empty output.
- More broadly, this breaks **any** Java recipe that mutates a Python/Starlark LST through `rewrite-python`.

Happy to pair on a fix or test a candidate patch against the reproducer above.

timtebeek:
## What's changed

- Fixes #8093 — a regression introduced in `8.84.8` (#7958) where **any** Java recipe that *modifies* a Python/Starlark LST (e.g. a `Find*` recipe adding a `SearchResult` marker) prints an **empty** file.

## Root cause

- #7958 stored LST/marker ids as a 128-bit `int` and turned the public `.id` into a lazily-reconstructed property:

```python
@property
def id(self):
    return UUID(int=self._id)
```

The marker *receive* codecs (`_receive_search_result`, `_receive_parse_exception_result`, `_receive_markup_marker`, `_receive_parse_error`) still computed the "before" id via `marker.id`. On the **ADD** path the marker is first created by a `make_dataclass_factory` with every field `None`, so `marker._id is None` and `marker.id` raised:

```
TypeError: one of the hex, bytes, bytes_le, fields, or int arguments must be given
```

from `UUID(int=None)`. That exception desynchronised the RPC receive stream, so the Java→Python round-trip of an *edited* tree rebuilt an empty/detached statement list and the printer produced `""`.

This only manifests on the **edited-tree RPC round-trip** — pure parse→print round-trips never hit the ADD path, which is why unmodified trees and the pure-Python harness were unaffected.

## Fix

Compute the before-id from the internal `_id` int via `id_to_str`, guarding `None`, in all four marker receive codecs.

## Testing

- New `tests/rpc/test_marker_roundtrip.py` drives the `RpcSendQueue` → `RpcReceiveQueue` diff round-trip with `SearchResult`/`ParseExceptionResult` markers (Java-free, fast). Fails on baseline with the exact `TypeError`; passes with the fix.
- `FindMethodsIntegTest`, `FindTypesIntegTest`, `ChangeMethodNameIntegTest` (the real Java↔Python RPC reproducer) all green; `FindMethodsIntegTest.findPrint` fails on baseline (`--rerun-tasks`) and passes with the fix.
- Full Python tree/recipe suite (634 tests) green.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/rpc/python_receiver.py
+++ b/rewrite-python/rewrite/src/rewrite/rpc/python_receiver.py
@@ -1219,7 +1219,7 @@ def _receive_search_result(marker, q: RpcReceiveQueue):
     from rewrite.markers import SearchResult
 
     # SearchResult sends: id, description
-    before_id = str(marker.id) if marker and marker.id else None
+    before_id = id_to_str(marker._id) if marker is not None and marker._id is not None else None
     id_str = q.receive(before_id)
     description = q.receive(marker.description if marker else None)
 
@@ -1236,7 +1236,7 @@ def _receive_parse_exception_result(marker, q: RpcReceiveQueue):
     from rewrite.markers import ParseExceptionResult
 
     # Receive in Java's send order: id, parserType, exceptionType, message, treeType
-    id_str = q.receive(str(marker.id) if marker else None)
+    id_str = q.receive(id_to_str(marker._id) if marker is not None and marker._id is not None else None)
     parser_type = q.receive(marker.parser_type if marker else None)
     exception_type = q.receive(marker.exception_type if marker else None)
     message = q.receive(marker.message if marker else None)
@@ -1282,7 +1282,7 @@ def _receive_markup_marker(marker, q: RpcReceiveQueue, cls):
     Matches Java's Markup.{Warn,Error,Info,Debug}.rpcReceive().
     """
 
-    id_str = q.receive(str(marker.id) if marker else None)
+    id_str = q.receive(id_to_str(marker._id) if marker is not None and marker._id is not None else None)
     message = q.receive(marker.message if marker else None)
     detail = q.receive(marker.detail if marker else None)
 
@@ -1799,7 +1799,7 @@ def _receive_parse_error(parse_error, q: RpcReceiveQueue):
     from pathlib import Path
 
     # Receive all fields in order (matching Java's ParseError.rpcSend)
-    id_str = q.receive(str(parse_error.id) if parse_error else None)
+    id_str = q.receive(id_to_str(parse_error._id) if parse_error is not None and parse_error._id is not None else None)
     markers = q.receive_markers(parse_error.markers if parse_error else None)
     source_path = q.receive(str(parse_error.source_path) if parse_error else None)
     charset_name = q.receive(parse_error.charset_name if parse_error else None)
--- a/rewrite-python/rewrite/tests/rpc/test_marker_roundtrip.py
+++ b/rewrite-python/rewrite/tests/rpc/test_marker_roundtrip.py
@@ -1,0 +1,131 @@
+# Copyright 2025 the original author or authors.
+#
+# Licensed under the Moderne Source Available License (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://docs.moderne.io/licensing/moderne-source-available-license
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Regression test for the RPC diff round-trip of a *modified* Python LST.
+
+A Java recipe that edits a Python LST (e.g. adds a ``SearchResult`` marker)
+sends the edited tree back to Python as a delta against the original. Python's
+receiver then applies that delta and prints the result. Issue #8093: after the
+LST memory-footprint rework (#7958) ids became a lazily-reconstructed UUID, so
+``marker.id`` raised ``TypeError`` for a factory-created marker (``_id is None``)
+on the ADD path. That broke RPC stream alignment and printed an empty file.
+
+These tests exercise the same ``RpcSendQueue`` -> ``RpcReceiveQueue`` path
+directly (no Java needed) so they run fast and never hang.
+"""
+import ast
+from dataclasses import fields, is_dataclass
+
+from rewrite import random_id
+from rewrite.java.tree import MethodInvocation
+from rewrite.markers import Markers, ParseExceptionResult, SearchResult
+from rewrite.python._parser_visitor import ParserVisitor
+from rewrite.python.printer import PythonPrinter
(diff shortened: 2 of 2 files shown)


greg-at-moderne (approved):
I think it's fine. I admit I fail to appreciate the nuance found here.

timtebeek:
## Why this fixes it

### The mechanism that broke

#7958 changed how LST/marker ids are stored: instead of a `uuid.UUID` field, the id is now a 128-bit `int` (`_id`), and the public `.id` became a lazily-reconstructed property:

```python
@property
def id(self) -> UUID:
    return UUID(int=self._id)
```

`UUID(int=None)` raises `TypeError: one of the hex, bytes, bytes_le, fields, or int arguments must be given`. Before #7958, `.id` was a plain stored field, so reading it when it was `None` simply returned `None` — harmless.

### Where `_id` is `None`

The marker **receive** codecs ran `marker.id` to compute the "before" id:

```python
before_id = str(marker.id) if marker and marker.id else None
```

On the RPC **ADD** path (a marker the remote side just added — e.g. a `SearchResult` from a Java `Find*` recipe), the receiver first builds a blank instance via `make_dataclass_factory(...)`, which sets **every field to `None`**, including `_id`. The codec then runs against that blank `marker`, so `marker.id` evaluates `UUID(int=None)` and throws.

### Why an exception there empties the whole file

The receive codecs are driven by a single ordered `RpcReceiveQueue`. Each `q.receive(...)` consumes exactly one message, and sender/receiver must stay in lock-step. When `marker.id` throws mid-codec, the queue is left **desynchronised** — the field messages for that marker are never consumed, and every subsequent `take()` reads the wrong envelope. The `CompilationUnit` it rebuilds ends up with a detached/empty statement list, so the printer emits an empty string (or, as seen end-to-end in the CLI, the print RPC fails outright: `Failed to write patch ... No changes`).

This is why the bug only showed up on a **modified** tree: a plain parse→print round-trip never hits the ADD path, so `.id` is always reconstructed from a real `_id` and never sees `None`.

### The fix

Compute the before-id from the internal `_id` int directly, guarding `None`, instead of going through the throwing `.id` property:

```python
before_id = id_to_str(marker._id) if marker is not None and marker._id is not None else None
```

`id_to_str` formats the 128-bit int to the canonical UUID wire string without ever constructing a `UUID`, so a blank (`_id is None`) marker yields `None` cleanly — exactly the pre-#7958 behaviour. The queue stays in lock-step, the marker deserialises normally, and the edited tree prints intact. Applied to all four codecs that used the `.id`-on-a-possibly-blank-marker pattern: `SearchResult`, `ParseExceptionResult`, `Markup`, and `ParseError`. (The other marker codecs use `random_id()` factories, so their `_id` is never `None` — unaffected.)

### Verification

- New `tests/rpc/test_marker_roundtrip.py` drives the `RpcSendQueue` → `RpcReceiveQueue` diff round-trip directly: fails on baseline with the exact `TypeError`, passes with the fix.
- End-to-end through a locally-built Moderne CLI (`:mod:devFatJar`) running `FindMethods` over a Python file: **before** → `TypeError` + "No changes" + failed patch write; **after** → `SearchResults` result `/*~~>*/print("hello")` rendered correctly.

timtebeek:
thanks for the review! If there's no new objections by tomorrow then I'll merge as is. This aims to fix the persistent failures we've seen on rewrite-bazel

knutwannheden:
Thanks for the fix! That change to the `_id` field was a lot tricker than what it first looked like...

PR #8102, opened 2026-06-23
Implement "go mod tidy" equivalent recipe

sambsnyd:
This branch implements `org.openrewrite.golang.GoModTidy`, an OpenRewrite recipe that reproduces go mod tidy's effect on go.mod, validated end-to-end through the moderne-cli. 

There was a cross-language RPC bug where editing .go sources dropped whitespace from unchanged subtrees. 
Our visit / with methods were instantiating new structs even when no changes were made. Fixed that.

Implemented downloading of dependencies so that we can have the complete dependency graph required for tidy (and other dependency manipulations in the future). Added an `HttpSender` to the RPC mechanism so that we don't have to shell out to go to do this and get proxy/credentials/etc. from the execution context per OpenRewrite idiom. 

Validated on 27 real Go repositories via the CLI.


rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -233,9 +243,54 @@ protected Boolean handle(Void noParams) {
             }
         });
 
+        // Http: lets the RPC peer (e.g. the Go module-graph resolver fetching
+        // dependency go.mod files from a GOPROXY) perform an HTTP GET through
+        // the configured HttpSender, so proxy/auth/TLS are honored. Returns the
+        // status code and base64-encoded body.
+        jsonRpc.rpc("Http", new JsonRpcMethod<HttpRequest>() {
```
jkschneider: Want to take a close look at this, and any other modification to the Rewrite RPC spec.

sambsnyd: I'm totally open to other ideas about how to enable HTTP communication in the ecosystems we support. Figured we would want to reuse the http sender we have on the java side rather than wiring through the individual proxy settings / credentials / etc.

sambsnyd: @jkschneider I've moved the go dependency resolver to java. There are no longer any changes to RewriteRpc itself

greg-at-moderne:
BTW, shouldn't this recipe land in https://github.com/moderneinc/recipes-go ?

PR #8105, opened 2026-06-23
Gradle 9: attach each subproject's own GradleProject in withToolingApi

### LINKED ISSUE #8080: withToolingApi: subproject build.gradle gets the root project's GradleProject model on Gradle 9+ (per-subproject dependencies not resolved)

`rewrite-gradle-tooling-model` from `main` (reproduced with Gradle 9.x distributions; the relevant code path is gated on Gradle 9+).

In a recipe test that uses `org.openrewrite.gradle.toolingapi.Assertions.withToolingApi(...)`, declare a dependency in a **subproject** `build.gradle` of a multi-module build (i.e. a build that has a root `settings.gradle`), then run a recipe that reads the resolved Gradle dependency model for that subproject (e.g. `DependencyInsight`, or anything that inspects `GradleProject.getConfigurations()`).

```java
class ReproTest implements RewriteTest {
    @Override
        spec.beforeRecipe(withToolingApi())
            .recipe(new org.openrewrite.java.dependencies.DependencyInsight(
                    "com.google.guava", "guava", null, null));
    }

    @Test
    void findsDependencyDeclaredInSubproject() {
        rewriteRun(
          settingsGradle(
            """
              rootProject.name = 'root'
              include 'sub'
              """
          ),
          buildGradle(
            """
              plugins { id 'java' }
              dependencies {
                  implementation("com.google.guava:guava:33.0.0-jre")
              }
              """,
            // EXPECTED: this subproject build.gradle is marked because guava resolves here.
            spec -> spec.path("sub/build.gradle")
          )
        );
    }
}
```

The `GradleProject` marker attached to `sub/build.gradle` should contain that subproject's own resolved configurations (so `guava` is found and the recipe makes a change), mirroring how a real multi-module Moderne/OpenRewrite LST carries a per-subproject `GradleProject`.

### What did you see instead?

On Gradle 9+, the subproject `build.gradle` receives the **root** project's `GradleProject` model instead of its own. The subproject's declared dependency is therefore never present in `GradleProject.getConfigurations()` — `getConfigurations()` reflects only the root project (and is empty when the root has no `build.gradle`). As a result, dependency-model recipes silently find nothing for any dependency declared only in a subproject, and `withToolingApi`-based multi-module tests cannot exercise per-subproject dependency resolution.

This is silent: no error is thrown; the recipe simply makes no change.

### Root cause

`Assertions.withToolingApi(...)` builds one model per Gradle build file by calling, for each `build.gradle`/`build.gradle.kts` source:

```java
OpenRewriteModelBuilder.forProjectDirectory(
        projectDir.toFile(),                                  // build root
        tempDirectory.resolve(sourceFile.getSourcePath()).toFile(), // this subproject's build file
        initScriptContents);
```

Inside `OpenRewriteModelBuilder.forProjectDirectory(projectDir, buildFile, initScript)`, the connection is opened against `projectDir` (the root) and the specific `buildFile` was historically targeted with the `-b`/`--build-file` argument. Gradle 9 removed `-b`, and the replacement logic only kicks in when there is **no** existing `settings.gradle`:

```java
if (isGradle9OrLater(gradleVersion)) {
    File abs = buildFile.getAbsoluteFile();
    boolean atConventionalLocation =
            abs.equals(new File(projectDir, "build.gradle").getAbsoluteFile()) ||
            abs.equals(new File(projectDir, "build.gradle.kts").getAbsoluteFile());
    if (!atConventionalLocation) {
        Path projectPath = projectDir.toPath();
        if (!Files.exists(projectPath.resolve("settings.gradle")) &&
                !Files.exists(projectPath.resolve("settings.gradle.kts"))) {
            // write temp settings.gradle with rootProject.buildFileName = ...
        }
    }
    // else: nothing targets `buildFile`
}
```

In a multi-module build the root `settings.gradle` always exists, so:

- `atConventionalLocation` is `false` for a subproject build file, **and**
- the `rootProject.buildFileName` fallback is skipped because a `settings.gradle` is present.

The net effect is that `buildFile` is ignored entirely on Gradle 9+, and `model.getGradleProject()` returns the **root** project's model for every subproject build file. (Confirmed identical on Gradle 9.3.1 and 9.5.0; single-module builds work because the build file is at the conventional root location.)

### Suggested direction

The per-build-file targeting needs a Gradle-9-compatible replacement for `-b` that also works when a real `settings.gradle` exists. Options:

- Resolve the model **once** for the whole build and return a `GradleProject` per Gradle project, then attach each to the matching subproject source by project path/directory (rather than re-targeting a build file per source file). This also matches how production LSTs are built.
- Or, map the requested `buildFile` to its Gradle project path and select that project's model from a multi-project tooling model, instead of relying on build-file substitution.

Happy to help test a fix.

Jenson3210:
# Disclaimer: 

I did not have time to test the verification with cli/plugin against real repos. 
Perhaps good if @shanman190 or @sambsnyd also take a look here.

- Fixes #8080

## Problem

Gradle 9 removed the `-b`/`--build-file` argument that `Assertions.withToolingApi(...)` relied on to resolve a per-subproject `GradleProject`. On Gradle 9+, every subproject `build.gradle` in a multi-module build received the **root** project's `GradleProject` marker, so dependency-model recipes silently found nothing for a dependency declared only in a subproject (the subproject's configurations were never present). No error was thrown — recipes just made no change.

## Fix

- **Tooling plugin** (`ToolingApiOpenRewriteModelPlugin`): serialize the `GradleProject` for *every* project in the build, keyed by Gradle path (`":"`, `":sub"`). The already-built model for the requested project is reused rather than rebuilt, so its configuration-resolution state (and therefore its resolved transitive dependencies) is unchanged.
- **Model** (`OpenRewriteModelProxy` / `OpenRewriteModelImpl` / `OpenRewriteModel`): carry the per-path map. Models produced by older tooling plugins are handled gracefully via `UnsupportedMethodException`.
- **`Assertions.withToolingApi`**: prefer the project whose Gradle path matches each build file, falling back to the previous behavior when no match exists (e.g. `buildSrc`, included builds, or build files whose directory doesn't map to a configured project).

The per-build-file model building is intentionally retained, so Gradle 8 behavior is unchanged; the per-path map only *refines* the attached project when it contains a match — which is the Gradle 9 case.

## Test

`AssertionsTest#multimoduleSubprojectGetsItsOwnGradleProjectOnGradle9` declares a dependency in a subproject of a multi-module build on Gradle 9.5.1 and asserts the subproject's `build.gradle` carries its own `GradleProject` (`path == ":sub"`, with resolved configurations) rather than the root's.

- Before the fix: the subproject's marker had `path == ":"` (the root) — test fails.
- After the fix: `path == ":sub"` — test passes.

## Verification

Ran the new test plus the regression-sensitive consumer suites:
`AssertionsTest`, `DependencyInsightTest`, `RemoveDependencyTest`, `UpgradeDependencyVersionTest`, `UpgradeTransitiveDependencyVersionTest`, `AddDependencyTest`, `ChangeDependencyTest`, `AddSettingsPluginTest`, `FindGradleProjectTest`, `GradleProjectTest` — all green (default Gradle 8.14.5 and the Gradle 9 cases).

### DIFF
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/Assertions.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/Assertions.java
@@ -156,10 +156,14 @@ public class Assertions {
                         } else if (sourceFile.getSourcePath().endsWith("build.gradle") || sourceFile.getSourcePath().endsWith("build.gradle.kts")) {
                             OpenRewriteModel model = OpenRewriteModelBuilder.forProjectDirectory(projectDir.toFile(), tempDirectory.resolve(sourceFile.getSourcePath()).toFile(), initScriptContents);
                             GradleProject gradleProject = model.getGradleProject();
+                            GradleProject projectForPath = model.getGradleProjectsByPath().get(gradlePathFor(tempDirectory, projectDir, sourceFile));
+                            if (projectForPath != null) {
+                                gradleProject = projectForPath;
+                            }
                             allRepositories.addAll(gradleProject.getMavenRepositories());
                             allBuildscriptRepositories.addAll(gradleProject.getBuildscript().getMavenRepositories());
                             sourceFiles.set(i, sourceFile.withMarkers(sourceFile.getMarkers().setByType(gradleProject)));
-                            gradleProjects.put(getDirectory(sourceFile), model.getGradleProject());
+                            gradleProjects.put(getDirectory(sourceFile), gradleProject);
                         } else if (sourceFile.getSourcePath().toString().endsWith(".gradle") || sourceFile.getSourcePath().toString().endsWith(".gradle.kts")) {
                             freestandingScriptFound = true;
                         }
@@ -231,5 +235,17 @@ public class Assertions {
             return parent.toString();
         }
         return "";
+    }
+
+    private static String gradlePathFor(Path tempDirectory, Path projectDir, SourceFile sourceFile) {
+        Path buildFileDir = tempDirectory.resolve(sourceFile.getSourcePath()).getParent();
+        if (buildFileDir == null) {
+            return ":";
+        }
+        String relative = projectDir.relativize(buildFileDir).toString();
+        if (relative.isEmpty()) {
+            return ":";
+        }
+        return ":" + relative.replace(File.separatorChar, ':').replace('/', ':');
     }
 }
--- a/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModel.java
+++ b/rewrite-gradle-tooling-model/model/src/main/java/org/openrewrite/gradle/toolingapi/OpenRewriteModel.java
@@ -18,6 +18,7 @@ package org.openrewrite.gradle.toolingapi;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.Value;
+import org.gradle.tooling.model.UnsupportedMethodException;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.RecipeSerializer;
 import org.openrewrite.gradle.marker.GradleDependencyConfiguration;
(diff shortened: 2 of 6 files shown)


sambsnyd (approved):
This looks good to me. Thanks @Jenson3210 !

PR #8108, opened 2026-06-24
Revert "Delete OWASP suppressions file when no suppressions remain"

greg-at-moderne:
- Reverts openrewrite/rewrite#7995
as it causes trouble in real-life.
At least the dependencyCheck Gradle plugin fails to operate with no `suppressions.xml` file:
```
[WARNING] Unable to read suppression file 'suppressions.xml'
[ERROR]   Exception occurred initializing CPE Analyzer.
...
InitializationException: Warn initializing the suppression analyzer:
  Failed to load suppressions.xml, caused by Unable to read suppression file.
    caused by SuppressionParseException: Failed to load suppressions.xml,
      caused by Unable to read suppression file.
```


timtebeek (approved):
That's surprising. Can we not make the file argument conditional on it's experience in our pipelines?

PR #8110, opened 2026-06-24
Add oldDigest="" to ChangeFrom to skip digest-pinned FROMs

MBoegers:
## What's changed
`ChangeFrom` configured to upgrade by tag also matches and rewrites `FROM` lines pinned with both a tag and a digest, e.g.:

```dockerfile
FROM registry.example.com/runtime:3.9.16-jdk21@sha256:<digest>
```

A deliberate digest pin should usually be left alone. This adds a **per-`FROM`** opt-in via a third meaning for the existing `oldDigest` option — no new option, public API unchanged:

- `oldDigest = null` → matches any digest or none (unchanged)
- `oldDigest = "<glob>"` → matches only `FROM`s whose digest matches the glob (unchanged)
- `oldDigest = ""` → **new**: matches only `FROM`s that have **no** digest, so digest-pinned ones are left untouched

Also adds `DockerFrom.isDigestPinned()` (distinct from the tag-based `isUnpinned()`).

## Why this shape
The empty-string sentinel mirrors the existing `""`=absent convention on the `new*` side and avoids any constructor/API change. The check is per-`FROM`, so a multi-stage Dockerfile mixing pinned and unpinned stages only upgrades the unpinned ones — a file-wide precondition can't do that.

## Tests
`ChangeFromTest.DigestPinned` covers the full three-way `oldDigest` matrix:

| `oldDigest` | digest-pinned FROM | unpinned FROM |
|---|---|---|
| `null` | change | change |
| `""` | skip | change |
| `"sha256:*"` | change | skip |

plus per-`FROM` behavior in a multi-stage file. `DockerFromTest` adds direct coverage of `isDigestPinned()`.

Customer request: moderneinc/customer-requests#2594

### DIFF
--- a/rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
+++ b/rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
@@ -145,8 +145,7 @@ public class ChangeFrom extends Recipe {
             matcher.tag(oldTag);
         }
         // Empty oldDigest is a sentinel: match only FROMs without a digest
-        boolean requireNoDigest = oldDigest != null && oldDigest.isEmpty();
-        if (oldDigest != null && !requireNoDigest) {
+        if (oldDigest != null && !oldDigest.isEmpty()) {
             matcher.digest(oldDigest);
         }
         if (oldPlatform != null) {
@@ -157,7 +156,7 @@ public class ChangeFrom extends Recipe {
             Docker.From f = image.getTree();
 
             // oldDigest="" matches only FROMs without a digest; leave digest-pinned ones untouched
-            if (requireNoDigest && image.isDigestPinned()) {
+            if (oldDigest != null && oldDigest.isEmpty() && image.isDigestPinned()) {
                 return f;
             }
 
--- a/rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
+++ b/rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
@@ -91,6 +91,14 @@ public class ChangeFrom extends Recipe {
     @Nullable
     String newPlatform;
 
+    @Option(displayName = "Keep digest-pinned images",
+            description = "When `true`, any `FROM` that carries a digest (`@sha256:...`) is left untouched, " +
+                    "preserving deliberate pins. The check is per-`FROM`, so other stages in a multi-stage " +
+                    "build are still changed. Defaults to `false` (current behavior).",
+            required = false)
+    @Nullable
+    Boolean keepDigestPinned;
+
     @Override
     public String getDisplayName() {
         return "Change Docker FROM";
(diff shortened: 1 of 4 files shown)


MBoegers:
@timtebeek would value your take on a design choice here.

This adds `keepDigestPinned` as a 9th `@Option` on `ChangeFrom`, which changes the Lombok all-args constructor signature and breaks any positional Java callers (`new ChangeFrom(...)`). The alternative is to leave `ChangeFrom` untouched and ship a separate recipe.

My lean is to just break the constructor: `ChangeFrom` is fairly new and not widely used yet, so the blast radius is small, and one recipe with an extra optional flag is cleaner than maintaining two near-identical recipes (a second one would duplicate all 8 existing options). Declarative/YAML usage is unaffected either way.

Is that acceptable, or would you prefer a second recipe? Happy to go either way.

MBoegers:
@timtebeek update: went with overloading `oldDigest=""` to mean "match only `FROM`s without a digest" instead of adding a new `keepDigestPinned` option. That sidesteps the constructor break entirely — no new `@Option`, public API unchanged — so the question above is moot. The trade-off is discoverability (behavior lives in an empty-string sentinel + the option doc) vs. a named boolean flag. Happy to switch back to a flag if you'd prefer the explicitness.

rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
```diff
@@ -142,7 +144,8 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
         if (oldTag != null) {
             matcher.tag(oldTag);
         }
-        if (oldDigest != null) {
+        // Empty oldDigest is a sentinel: match only FROMs without a digest
+        if (oldDigest != null && !oldDigest.isEmpty()) {
```
timtebeek, line 149: Isn't this more explicit?
```suggestion
        if ("".equals(oldDigest)) {
```

rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
```diff
@@ -152,6 +155,11 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
         return matcher.asVisitor((image, ctx) -> {
             Docker.From f = image.getTree();
 
+            // oldDigest="" matches only FROMs without a digest; leave digest-pinned ones untouched
+            if (oldDigest != null && oldDigest.isEmpty() && image.isDigestPinned()) {
```
timtebeek:
```suggestion
            if ("".equals(oldDigest) && image.isDigestPinned()) {
```

rewrite-docker/src/main/java/org/openrewrite/docker/ChangeFrom.java
timtebeek: ah missed the negation here. thanks!

timtebeek (approved):

PR #8114, opened 2026-06-24
Keep Javadoc reference parameter names out of whitespace

timtebeek:
## Problem
Scheduled CI in [openrewrite/rewrite-analysis](https://github.com/openrewrite/rewrite-analysis/actions/runs/28124014547) (and any repo parsing similar Javadoc) fails with:

```
java.lang.AssertionError: Source file was parsed into an LST that contains non-whitespace
characters in its whitespace. This is indicative of a bug in the parser.
     * @see #bar(String~~(non-whitespace)~~> str<~~)
```

## Root cause
A Javadoc method reference may include a non-standard **parameter name** after the parameter type, e.g. `@see #bar(String str)`. The compiler's `DCReference.paramTypes` only records the parameter *types*, so the trailing name (`str`) was consumed by `sourceBeforeAsString(")")`/`sourceBeforeAsString(",")` straight into the right-padding `Space` of the parameter type expression — i.e. non-whitespace ended up inside a `Space`.

- This has been latent for a long time; #8107 made `JavaWhitespaceValidationService` descend into Javadoc comments, which (correctly) now surfaces it.

## Fix
- Record the parameter name in a new `JavadocParameterName` marker attached to the parameter type expression, keeping the `Space` whitespace-only, and re-emit it verbatim in `JavadocPrinter` so the source still round-trips. The marker mirrors the `Varargs` marker added in #8107 (incl. `RpcCodec`). Applied to the Java 8/11/17/21/25 Javadoc visitors.

## Tests
Added to the TCK `JavadocTest` (run across all `rewrite-java-N` modules):
- `seeMethodReferenceWithNamedParameter` — `@see #bar(String str)` (the exact rewrite-analysis case)
- `seeMethodReferenceWithNamedParameters` — multiple named params
- `linkMethodReferenceWithNamedVarargsParameter` — `{@link #bar(String... strs)}` (named + varargs)

Verified green on `compatibilityTest` for java-8, -11, -17, -21, -25, plus the existing `JavadocTest`/`JavadocPrinterTest` suites.

### DIFF
--- a/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
+++ b/rewrite-java-11/src/main/java/org/openrewrite/java/isolated/ReloadableJava11JavadocVisitor.java
@@ -33,6 +33,7 @@ import com.sun.tools.javac.util.Context;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.Tree;
 import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.marker.JavadocParameterName;
 import org.openrewrite.java.marker.LeadingBrace;
 import org.openrewrite.java.marker.Varargs;
 import org.openrewrite.java.tree.*;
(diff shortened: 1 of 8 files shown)


sambsnyd (approved):

timtebeek:
@sambsnyd I don't necessarily love adding markers for Javadoc elements, but it's what we've started earlier today and discovered downstream in rewrite-analysis to reach this further addition. Are we ok to merge?

PR #8115, opened 2026-06-24
Fix print idempotency when Lombok deletes a field annotation (@Delegate)

timtebeek:
When a class-level Lombok annotation (e.g. `@Getter`) triggers a full processing round, Lombok erases a field's `@Delegate` annotation from the javac AST while leaving it in the source, after which `sortedModifiersAndAnnotations` treated the leftover `@Delegate` token as an unknown modifier, desynced the cursor, and corrupted the following type — printing `@Delegate` as `Listegate` and producing a non-idempotent `ParseError`.

All five Reloadable Java parser visitors (8/11/17/21/25) now detect a source `@`-token with no corresponding AST node and reconstruct it via a new `sourceAnnotation()` helper (handling arguments and a preceding real annotation) instead of bailing.

Added `LombokTest` coverage for the bare, with-arguments, and preceding-annotation cases; full TCK passes on every Java version.

### DIFF
--- a/rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8ParserVisitor.java
+++ b/rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8ParserVisitor.java
@@ -2213,15 +2213,36 @@ public class ReloadableJava8ParserVisitor extends TreePathScanner<J, Space> {
                         noSpace = false;
                     }
                     if (!word.get().isEmpty()) {
+                        String keyword = word.get();
+                        // `@interface` is an annotation type declaration, not an annotation usage
+                        boolean isAnnotation = keyword.charAt(0) == '@' && !"@interface".equals(keyword);
                         Modifier matching = null;
-                        for (Modifier modifier : modifiers.getFlags()) {
-                            if (modifier.name().toLowerCase().equals(word.get())) {
-                                matching = modifier;
-                                break;
+                        if (!isAnnotation) {
+                            for (Modifier modifier : modifiers.getFlags()) {
+                                if (modifier.name().toLowerCase().equals(keyword)) {
+                                    matching = modifier;
+                                    break;
+                                }
                             }
                         }
 
-                        if (matching == null) {
+                        if (isAnnotation) {
+                            // An annotation that is present in the source but was removed from the AST by Lombok
+                            // (e.g. `@Delegate`, which Lombok deletes after generating the delegating methods).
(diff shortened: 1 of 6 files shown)


rewrite-java-8/src/main/java/org/openrewrite/java/ReloadableJava8ParserVisitor.java
```diff
@@ -2247,6 +2268,52 @@ private Java8ModifierResults sortedModifiersAndAnnotations(ModifiersTree modifie
...
+        Space prefix = sourceBefore("@");
+        int nameStart = cursor;
+        while (cursor < source.length() &&
+                (Character.isJavaIdentifierPart(source.charAt(cursor)) || source.charAt(cursor) == '.')) {
+            cursor++;
+        }
+        J.Identifier name = new J.Identifier(randomId(), EMPTY, Markers.EMPTY, emptyList(),
+                source.substring(nameStart, cursor), null, null);
+
+        JContainer<Expression> args = null;
```
sambsnyd, line 2293: Are there any annotations other than `@Delegate` itself that this is applicable to? Seems to me we can provide the true type attribution of `@Delegate` just fine without having to insert a stub/null value

timtebeek: Good call — done in a0ea52e. `resolveErasedAnnotationType` now looks the annotation up in the compilation unit's single-type imports (whose symbols stay attributed even though Lombok erased the field annotation) and attributes the reconstructed `J.Identifier` with the real `JavaType`. Added a `LombokTest` assertion that `@Delegate` resolves to `lombok.experimental.Delegate`.

On scope: in practice `@Delegate` is the only one. Lombok only deletes the source annotation for handlers that rewrite-java-lombok does *not* replace, and of those `@Delegate` is the one that sits on a member declaration followed by a type (so the cursor desync corrupts the type). I checked `@With` (another non-replaced, field-level handler) and it isn't erased. The code stays general (any erased annotation it can resolve gets its type; otherwise null) rather than hardcoding `@Delegate`. The annotation *arguments* remain best-effort source text since Lombok erases those too.

timtebeek:
Seeing how this only affects the experimental Lombok `@Delegate` I'm going to propose we abandon this attempt:
> Current status: negative - Currently we feel this feature will not move out of experimental status anytime soon, and support for this feature may be dropped if future versions of javac or ecj make it difficult to continue to maintain the feature.
-- https://projectlombok.org/features/experimental/Delegate

PR #8129, opened 2026-06-26
Avoid duplicate dependencies in ChangeDependency

timtebeek:
## What's changed

`ChangeDependencyGroupIdAndArtifactId` (the Maven engine behind `org.openrewrite.java.dependencies.ChangeDependency`) could produce a duplicate dependency: when the new coordinates were already declared directly **at a lower version** than the requested one, the recipe renamed the old declaration in place, leaving two declarations of the same dependency.

The recipe already removed the old declaration when the existing copy was at a version `>=` the requested one, but the version guard meant the lower-version case fell through to a rename and produced a duplicate.

This change:
- Removes the old declaration whenever an existing **direct** dependency already declares the new coordinates, matched on full identity (`groupId` + `artifactId` + `classifier` + `type` + `scope`) — independent of version.
- **Upgrades the surviving declaration in place** to the requested version pattern, reusing the recipe's existing version-resolution and managed-dependency handling.
- Leaves declarations that differ on classifier, type, or scope untouched, since those are distinct dependencies.

## Why

Concretely, migrating a POM that declares both `jakarta.jws:jakarta.jws-api:3.0.0` and `jakarta.xml.ws:jakarta.xml.ws-api:3.0.1` with `ChangeDependency` (jws-api → xml.ws-api, version `4.0.x`) previously yielded two `jakarta.xml.ws-api` entries. It now drops `jakarta.jws-api` and upgrades the existing `jakarta.xml.ws-api` to `4.0.x` — a single, correct entry.

- This is the upstream fix for the duplicate that [openrewrite/rewrite-migrate-java#1153](https://github.com/openrewrite/rewrite-migrate-java/pull/1153) compensates for downstream with `RemoveDuplicateDependencies`.
- Closes openrewrite/rewrite-migrate-java#1153

## Tests

- Reworked `shouldAddNewIfDependencyAlreadyExistsInOlderVersion` (which asserted the duplicate) into `shouldUpgradeExistingInPlaceIfDependencyAlreadyExistsInOlderVersion`, asserting a single upgraded entry.
- Added `shouldNotLeaveDuplicateWhenExistingNewDependencyIsAtLowerVersion` covering the jws/ws scenario.
- Added `shouldNotDeduplicateWhenClassifierDiffers` confirming distinct coordinates are preserved.
- Verified no regressions in `ChangeManagedDependencyGroupIdAndArtifactIdTest` and `UpgradeDependencyVersionTest`.

The Gradle `ChangeDependency` path is out of scope here, matching the Maven-only scope of the downstream compensation.


rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -259,14 +259,19 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
             final VersionComparator versionComparator = newVersion != null ? Semver.validate(newVersion, versionPattern).getValue() : null;
             @Nullable
             private Collection<String> availableVersions;
-            private boolean isNewDependencyPresent;
             private Set<String> safeVersionPlaceholdersToChange = new HashSet<>();
+            private final boolean dedupeEnabled = newGroupId != null && newArtifactId != null;
```
steve-aom-elliott: Why only when there's a new group id _and_ a new artifact id? Wouldn't just one of the two be sufficient on its own? I suppose you'd need to change the args you pass in some places to account for that as well, but I assume changing just the group id for example, we'd still want to detect duplicates fully.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -333,15 +338,22 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 }
 
                 boolean isOldDependencyTag = isDependencyTag(oldGroupId, oldArtifactId);
-                if (isOldDependencyTag && isNewDependencyPresent) {
+                boolean isNewDependencyTag = dedupeEnabled && isDependencyTag(newGroupId, newArtifactId);
+                // The new coordinates are already declared directly with the same classifier, type, and scope;
+                // drop the old declaration rather than renaming it into a duplicate of the existing one.
+                if (isOldDependencyTag && !isNewDependencyTag && matchesAnyDirectDependency(t, existingNewDirectDependencies)) {
```
steve-aom-elliott: Don't know that I necessarily understand this line. It's saying that it's:
- the old dependency coordinates (implying you'd want to change them normally, sure)
- that either you didn't provide new coordinates (which you kind of need to, unless there's something specifically different about the desired behaviour for providing new group id and artifact id vs just one of those that I'm missing) or you're not the new coordinates (which you shouldn't be if you're the old coordinates currently, so I'm not sure that part is needed per se)
- that your tag matches the new coordinates collected? (Wouldn't it not because we're the old dependency tag?)

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -333,15 +338,22 @@ public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
...
                     doAfterVisit(new RemoveContentVisitor<>(tag, true, true));
                     maybeUpdateModel();
                     return t;
                 }
+                // When the old declaration is being dropped as a duplicate, upgrade the surviving new
+                // declaration in place to the requested version instead of leaving it untouched.
+                boolean isSurvivingNewDependencyTag = isNewDependencyTag && newVersion != null &&
+                        matchesAnyDirectDependency(t, existingOldDirectDependencies);
```
steve-aom-elliott: If you're the new dependency tag, wouldn't you necessarily not have the old coordinates that would return true on `matchesAnyDirectDependency(t, existingOldDirectDependencies)`?

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
timtebeek: Good catch — you're right, that was too restrictive. I dropped the requirement to `newGroupId != null || newArtifactId != null` and now compute *effective* new coordinates (a null option keeps the existing value), so a group-id-only or artifact-id-only rename detects duplicates too. Added `shouldDeduplicateWhenOnlyArtifactIdChanges` to cover it.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
timtebeek: You were right to be confused — the old `matchesAnyDirectDependency` deliberately ignored groupId/artifactId (relying on exact pre-filtering) and only compared classifier/type/scope, which reads as nonsense here. I reworked it into `renamedOldTagWouldDuplicate(oldTag)`, which computes the coordinates this old declaration *becomes* after the rename and compares the full coordinates (groupId, artifactId, classifier, type, scope) against the already-declared new dependencies. The `!isNewDependencyTag` guard is only there for the case where old and new coordinates overlap (e.g. via globs), so we never remove a tag that is itself the intended new dependency.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
timtebeek: Same root cause as above: the helper ignored groupId/artifactId, so passing the new tag against the old dependencies looked wrong. It's now `isSurvivorOfRemovedOldDependency(newTag)`: for each old direct dependency being removed, it computes that dependency's *renamed* coordinates and checks whether they match this new tag's full coordinates (groupId, artifactId, classifier, type, scope). If so, this tag is the survivor that should take the requested new version — so the g/a comparison is between the renamed old dep and the new tag, not old-vs-new directly.

rewrite-maven/src/main/java/org/openrewrite/maven/ChangeDependencyGroupIdAndArtifactId.java
```diff
@@ -259,14 +259,21 @@ public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
             final VersionComparator versionComparator = newVersion != null ? Semver.validate(newVersion, versionPattern).getValue() : null;
             @Nullable
             private Collection<String> availableVersions;
-            private boolean isNewDependencyPresent;
             private Set<String> safeVersionPlaceholdersToChange = new HashSet<>();
+            private final String effectiveNewGroupId = newGroupId != null ? newGroupId : oldGroupId;
+            private final String effectiveNewArtifactId = newArtifactId != null ? newArtifactId : oldArtifactId;
```
steve-aom-elliott: Could probably use your new helpers from lower down.

steve-aom-elliott (approved):
I think it ended up correct, it's still just denser logic than I think ideally we'd be able to describe it as. Not sure how to cleanly simplify further, though.

PR #8132, opened 2026-06-26
Bound the buffer in BackwardCompatibleObjectIdModule

natedanner:
## What's changed

`BackwardCompatibleObjectIdModule` reads the resolved-dependency graphs (`ResolvedPom`, `ResolvedDependency`, …) nested in `MavenResolutionResult`/`GradleProject` markers, bridging the pre/post-`@JsonIdentityInfo` wire formats.

Its deserializer buffered each object's **entire subtree** into a `TokenBuffer`, scanned it for the `@ref` id, then replayed it. Since each object is read from its parent's replay buffer, every subtree is re-buffered at each ancestor — **O(depth²)** for deep graphs, and a major source of LST-deserialization allocation churn.

This makes the buffering **bounded**: it scans only the field-name prefix and, in the common case where `@ref` is written first, streams the remainder straight from the live parser via `JsonParserSequence` (Jackson's own `AsPropertyTypeDeserializer` idiom). Only the rare legacy case (no `@ref`) still fully buffers.

Delegate selection, parser end position, and wire format are all unchanged — purely a read-path optimization:

| Encoding | Delegate | Change |
|---|---|---|
| bare integer | id-aware (back-ref) | unchanged |
| object with `@ref` | id-aware | streamed, not buffered |
| object without `@ref` | id-free | unchanged |

## Why

A composite dependency-migration recipe over 300+ repos OOM-killed the CLI at `-Xmx12g`; smaller runs were fine. Allocation profiling put **~39%** of the churn in this module — every hot stack ending in `copyCurrentStructure → findObjectId → IntSequenceGenerator.key`, with `Arrays.copyOf`/`TokenBuffer$Segment` as the top self-allocators. The graphs cost far more to read than to hold.

## Performance

Synthetic A/B (same JVM, `getThreadAllocatedBytes`), `@ref`-first graphs:

| Shape | alloc before | alloc after | time before | time after |
|---|---|---|---|---|
| deep (depth 3000) | 1,150,758,584 B | 3,494,352 B | 425 ms | 1.7 ms |
| flat (width 20000) | 25,595,124 B | 21,150,946 B | 15.6 ms | 10.9 ms |

The win scales with nesting depth; shallow/wide payloads improve less.

## Testing

New `BackwardCompatibleObjectIdModuleTest` (none existed): all three encodings in JSON, the streamed path + back-reference through binary Smile, shared-instance identity, id-not-first detection, and a bounded-buffering test that fails against the old copy. Existing `GradleProject` round-trip and the rewrite-core suite stay green.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/internal/BackwardCompatibleObjectIdModule.java
+++ b/rewrite-core/src/main/java/org/openrewrite/internal/BackwardCompatibleObjectIdModule.java
@@ -17,6 +17,7 @@ package org.openrewrite.internal;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.util.JsonParserSequence;
 import com.fasterxml.jackson.databind.BeanDescription;
 import com.fasterxml.jackson.databind.DeserializationConfig;
 import com.fasterxml.jackson.databind.DeserializationContext;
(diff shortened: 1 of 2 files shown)


sambsnyd (approved):
Someone should double check, but I'm pretty sure the @ref property is always put at the beginning of the serialized representation. So realistically we shouldn't ever have to read very far to find out if it is there or not.

natedanner:
> Someone should double check, but I'm pretty sure the @ref property is always put at the beginning of the serialized representation. So realistically we shouldn't ever have to read very far to find out if it is there or not.

Confirmed `@ref` is always first, and added `idPropertyIsAlwaysSerializedFirst` to pin it (JSON + Smile). I'd still keep the scan rather than short-circuit on it. We read LSTs at runtime, so a writer the test can't cover would throw `UnresolvedForwardReference` and kill the whole recipe run, whereas the scan degrades gracefully. And since `@ref` is first, the new-format path is already O(1), so the scan only costs anything on the rare legacy read. Worth short-circuiting anyway, or keep the scan?

shanman190 (approved):

sambsnyd:
You're right not much point in optimizing the increasingly rare legacy objects lacking it

PR #8134, opened 2026-06-28
Fix C# recipe over-attribution when multiple packages share a dependency

shanman190:
## Problem

Installing several .NET recipe packages that share a transitive dependency —
e.g. `Core`, `CodeQuality`, and `Migration.DotNet` (which depends on `Core`) —
caused `Core`'s recipes to disappear from the deployment list. A dependent
package was effectively "stealing" the recipes its dependency contributed.

Two independent root causes:

1. **Cross-ALC name reflection during composition.** C# composite recipes
   resolved their children by *name* across sibling `AssemblyLoadContext`
   instances, which pulled dependency recipes into the shared marketplace and
   mis-attributed them.
2. **Origin-blind marketplace rows.** `GetMarketplace` rows carried no origin,
   so the host's `toMarketplace(bundle)` tagged *every* recipe with the single
   requested bundle — over-attributing recipes contributed by other packages.

### Composition isolation (whole-tree `PrepareRecipe`)
- `PrepareRecipe` returns the entire prepared child tree in one response, so a
  composite resolves its children by binary integration **within its own ALC**
  instead of reflecting across sibling contexts. `RpcRecipe` builds children
  locally from that tree.
- A guarded by-name fallback remains for RPC peers (Python/JS/Go) that don't yet
  return a prepared tree — they keep working unchanged until they adopt it.
- Each NuGet bundle installs into its own per-`(package, version)` directory and
  activates only its own `lib/<tfm>` assemblies; transitive dependency DLLs load
  (for composition) but are not activated into the marketplace.

### Origin attribution (marketplace)
- The C# server records each recipe's origin package at activation and emits it
  as `packageName` on each `GetMarketplace` row.
- `toMarketplace` attributes each row to its own bundle and **filters out
  foreign-origin rows**, so each bundle reader contributes only its own recipes.
  A `null` origin falls back to the requested bundle, so ecosystems not yet
  emitting origin (npm/Go/Python) behave exactly as before.

## Testing

- New end-to-end regression test `CSharpRecipeArtifactIsolationTest` installs the
  three real NuGet packages and asserts `Core` keeps `ChangeDotNetTargetFramework`,
  `Migration.DotNet` doesn't steal it, and `Migration.DotNet`'s composite still
  resolves `Core`'s recipe as a child via `PrepareRecipe`.
- Unit coverage for the per-row `toMarketplace` filter, server-side origin
  attribution, own-assembly scoping, whole-tree prepare, and per-bundle version
  isolation.

### DIFF
--- a/rewrite-csharp/csharp/OpenRewrite/CSharp/Rpc/RewriteRpcServer.cs
+++ b/rewrite-csharp/csharp/OpenRewrite/CSharp/Rpc/RewriteRpcServer.cs
@@ -48,6 +48,11 @@ public class RewriteRpcServer
     public static void SetCurrent(RewriteRpcServer? server) => _current = server;
 
     private readonly RecipeMarketplace _marketplace;
+
+    // Recipe name -> the package that contributed it, recorded at install time and persisted for the
+    // process lifetime so GetMarketplace can attribute each row even when a later install is a no-op.
+    private readonly ConcurrentDictionary<string, string> _recipeOrigin = new();
+
     private readonly ConcurrentDictionary<string, Recipe> _preparedRecipes = new();
     private readonly ConcurrentDictionary<string, object?> _recipeAccumulators = new();
     private readonly ConcurrentDictionary<string, ExecutionContext> _executionContexts = new();
@@ -517,7 +522,7 @@ public class RewriteRpcServer
         return Task.FromResult(rowByRecipeId.Values.ToList());
     }
 
-    private static void CollectRecipes(
+    private void CollectRecipes(
         Dictionary<string, GetMarketplaceResponseRow> rowByRecipeId,
         RecipeMarketplace.Category category,
         List<CategoryDescriptorDto> parentPath)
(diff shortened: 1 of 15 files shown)


rewrite-csharp/csharp/OpenRewrite/CSharp/Rpc/RewriteRpcServer.cs
```diff
@@ -48,6 +48,11 @@ public class RewriteRpcServer
     public static void SetCurrent(RewriteRpcServer? server) => _current = server;
 
     private readonly RecipeMarketplace _marketplace;
+
+    // Recipe name -> the package that contributed it, recorded at install time and persisted for the
+    // process lifetime so GetMarketplace can attribute each row even when a later install is a no-op.
+    private readonly ConcurrentDictionary<string, string> _recipeOrigin = new();
```
shanman190, line 54: This is a shortcut for now. I'd like to follow up with more complete origin attribution in the RPC-level marketplace across all current RPC implementations.

While each RPC may only manage a single ecosystem, origin identity is still important in order to correctly attribute a recipe to the bundle that provides it.

PR #8137, opened 2026-06-29
python: per-bundle recipe attribution and whole-tree PrepareRecipe

shanman190:
## What

Brings the Python (pip) RPC recipe server to parity with the C#, JS, and Go servers across three features.

### Per-bundle recipe attribution (GetMarketplace)

The Python server emitted origin-less `GetMarketplace` rows, so the host worked around the resulting over-attribution by reading the bundle's recipes from the `InstallRecipes` response instead (a bolt-on). `GetMarketplace` rows now carry each recipe's origin `packageName` — recorded per-distribution during discovery/install and looked up via `RecipeAttribution.package_for` — so the host's `toMarketplace` attributes each recipe to its own bundle. The reader (`PipRecipeBundleReader`) now reads from `getMarketplace(bundle)`, and the install-response `recipes` bolt-on is retired (dropped from the Python response and the `InstallRecipesResponse` DTO; the response is back to `{recipesInstalled, version}`, matching the other ecosystems). Local-path installs with no package identity stay unattributed and fall back to the requested bundle.

### Whole-tree (one-shot) PrepareRecipe + recursive validation

`PrepareRecipe` now prepares the whole recipe tree in one call and returns it as `recipeList` (the child prepared responses), mirroring the other servers, so the host builds the tree locally instead of a `PrepareRecipe` round trip per child. Required-option validation lives in that recursion (`_prepare_instance`), so it covers the whole tree — each node is validated against the values its parent set, not just the root. A child hosted on the peer (an `RpcRecipe`) is emitted as `delegatesTo` for the host to resolve; the rest are prepared and validated recursively.

## Testing

- New attribution tests (per-row `packageName` emit, unattributed → `None`) and prepare tests (missing required option rejected at the root and recursively for a composite's child; whole-tree `recipeList`).
- Migrated the two install-response attribution tests to the GetMarketplace path.
- Python suite green; `rewrite-python` Java compiles.

### DIFF
--- a/rewrite-python/src/main/java/org/openrewrite/python/marketplace/PipRecipeBundleReader.java
+++ b/rewrite-python/src/main/java/org/openrewrite/python/marketplace/PipRecipeBundleReader.java
@@ -34,9 +34,6 @@ public class PipRecipeBundleReader implements RecipeBundleReader {
 
     @Override
     public RecipeMarketplace read() {
-        // GetMarketplace rows now carry each recipe's origin packageName, so toMarketplace keeps
-        // only this bundle's recipes — the round trip that used to over-attribute the whole
-        // singleton marketplace to this bundle is now safe.
         return rpc.getMarketplace(bundle);
     }
 
--- a/rewrite-python/src/main/java/org/openrewrite/python/marketplace/PipRecipeBundleReader.java
+++ b/rewrite-python/src/main/java/org/openrewrite/python/marketplace/PipRecipeBundleReader.java
@@ -19,13 +19,11 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.openrewrite.Recipe;
 import org.openrewrite.config.RecipeDescriptor;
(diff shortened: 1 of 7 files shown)


rewrite-python/src/main/java/org/openrewrite/python/marketplace/PipRecipeBundleReader.java
```diff
@@ -19,33 +19,25 @@
...
-    private final InstallRecipesResponse installResponse;
 
     @Override
     public RecipeMarketplace read() {
-        GetMarketplaceResponse response = new GetMarketplaceResponse();
-        response.addAll(installResponse.recipesOrEmpty());
-        return response.toMarketplace(bundle);
+        // GetMarketplace rows now carry each recipe's origin packageName, so toMarketplace keeps
+        // only this bundle's recipes — the round trip that used to over-attribute the whole
+        // singleton marketplace to this bundle is now safe.
```
shanman190:
```suggestion
```

PR #8154, opened 2026-06-30
YAML: prevent block-scalar value mutations from corrupting siblings

steve-aom-elliott:
## Summary

Several built-in `rewrite-yaml` recipes silently corrupt YAML files containing FOLDED (`>`, `>-`, `>+`) or LITERAL (`|`, `|-`, `|+`) block scalars by clobbering the block envelope when the mapped property's value is replaced.

This PR fixes the value-replacement path through a new internal `BlockScalarUtils` helper, refactors the affected built-ins to use it, and documents the semantic asymmetry on `Yaml.Scalar.value`.

## Root cause

`Yaml.Scalar.value` carries different content depending on `style`:

- For PLAIN and quoted scalars: just the body content.
- For FOLDED/LITERAL: everything after the `>` or `|` indicator — the chomp indicator (if any), the newline terminating the block-scalar header, the indented body, **and the trailing whitespace that bounds the block from the next sibling mapping entry**.

For block scalars, the boundary newline that visually separates the scalar from the next sibling lives inside the previous scalar's `value`, not in the next entry's `prefix`. A naïve `scalar.withValue(...)` replacement clobbers the block envelope; the printer then concatenates `[> or |] + newPlainString + [next entry's empty prefix] + [next key]`, producing one-line corruption like `key: >replacedafter: tail`.

Round-trip without edits is byte-clean — this is purely a value-mutation problem in recipes, not a parser or printer bug.

## What's fixed

| Recipe | Before | After |
|---|---|---|
| `ChangePropertyValue` | corrupted all 6 chomp variants | preserves block envelope, re-indents body |
| `ChangeValue` | corrupted all 6 chomp variants | preserves block style when existing is FOLDED/LITERAL |
| `DeleteProperty` (delete entry after block scalar) | left an extra blank line | strips redundant prefix newline when block-scalar predecessor already supplies the separator |
| `trait/YamlValue.withValue(String)` | clobbered envelope | uses `BlockScalarUtils.withBody` |
| `trait/YamlReference.rename(...)` | clobbered envelope | uses `BlockScalarUtils.withBody` |

14 new parameterized tests across `ChangePropertyValueTest`, `ChangeValueTest`, `DeletePropertyKeyTest`.

## API surface

The helpers (`getBody`, `withBody`) are intentionally placed in `org.openrewrite.yaml.internal.BlockScalarUtils` rather than exposed as instance methods on `Yaml.Scalar`. The Javadoc on `BlockScalarUtils` records the future intent to promote them onto `Yaml.Scalar` once enough downstream environments have rolled forward — exposing them now risks downstream recipes depending on methods that older bundled `rewrite-yaml` versions don't provide.

`Yaml.Scalar` gets a Javadoc-only update on the `value` field documenting the per-style semantics and pointing to `BlockScalarUtils`. No new fields, no new public methods, no constructor changes.

## What's deliberately not fixed

- `MergeYamlVisitor` — the value-replacement path in `mergeScalar` has the same shape, but fixing it in isolation regresses `addLiteralStyleBlockWhichDoesAlreadyExist` because the surrounding merge-insertion logic computes new-entry prefixes assuming the broken value semantics. Needs a coordinated fix to both `mergeScalar` and the insertion prefix calculation. Follow-up PR.
- `CoalescePropertiesVisitor` — different bug entirely: the recipe doesn't reach steady state when block scalars are present; restructured mapping indentation is wrong. Not a `withValue` clobber. Follow-up PR.

## Test plan

- [x] All 642 existing `rewrite-yaml` tests pass
- [x] 14 new tests pass (6 chomp variants for `ChangePropertyValue`, FOLDED/LITERAL/KEEP for `ChangeValue`, delete-after/-before/literal-keep cases for `DeleteProperty`)
- [x] Multi-line replacement re-indents correctly
- [x] Regex replacement mode operates on body only (no longer matches against chomp indicator or trailing whitespace)
- [x] No regressions in `MergeYamlTest`, `CoalescePropertiesTest`, or other yaml tests


rewrite-yaml/src/main/java/org/openrewrite/yaml/internal/BlockScalarUtils.java
```diff
@@ -0,0 +1,142 @@
...
+ * field carries the block envelope (chomp indicator, indented body, trailing whitespace
+ * bounding the next sibling), so a naïve Lombok-generated {@code withValue} replacement
+ * corrupts the surrounding structure.
+ *
+ * <p><b>TODO:</b> promote these onto {@link Yaml.Scalar} as instance {@code getBody} /
+ * {@code withBody} methods once enough downstream CLI bundles ship a {@code rewrite-yaml}
+ * including them — promoting now would {@code NoSuchMethodError} customer recipes that
+ * adopted the new API but load against an older bundled {@code Yaml.Scalar}.
+ */
+public final class BlockScalarUtils {
```
jkschneider: Not super comfortable with Utils classes. Can you search for where else we would put such convience classes often?

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
```diff
@@ -307,6 +307,15 @@ class Scalar implements Block, YamlKey {
         @Nullable
         Tag tag;
 
+        /**
```
jkschneider, line 310: Overly verbose comment.

rewrite-yaml/src/main/java/org/openrewrite/yaml/internal/BlockScalarUtils.java
steve-aom-elliott: There's already a `StringUtils` in the same package. In the ideal, these would have just been put directly into `Yaml.Scalar` and not had a utility class at all, but given that changes the available method list on a core tree type, it would mean that newer recipes using those methods but running on an older CLI would more than likely end up throwing `NoSuchMethodError`, unless I've misunderstood the whole parent-loading thing.

rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java
steve-aom-elliott: Shortened it further.

rewrite-yaml/src/main/java/org/openrewrite/yaml/internal/BlockScalarUtils.java
sambsnyd: These should probably just be methods on `Yaml.Scalar` itself

steve-aom-elliott: @sambsnyd @jkschneider To confirm, is my concern about the `NoSuchMethodError` potential well founded here? Yes, I can add the methods to `Yaml.Scalar`. I just want to acknowledge that we will likely get reports of core YAML recipes not working if they _only_ upgrade their recipe artifact and not their CLI version as well.

sambsnyd: I don't think Yaml.Scalar is parent loaded so it should be fine. If you merge and publish the recipe jar and get runtime errors we need only publish a new version of moderne-worker to resolve

sambsnyd (approved):

steve-aom-elliott:
TODO: look into making this a Trait instead, as it's been shown locally that it will blow up with `NoSuchMethodError` if only in `Yaml.Scalar` and our CLI hasn't been updated.

Edit: done

PR #8157, opened 2026-06-30
YAML: fix MergeYaml + fix CoalesceProperties formatting on block scalars

steve-aom-elliott:
Follow-up to #8154 covering two more places where block-scalar handling was broken.

## `MergeYaml.mergeScalar`

Scalar overwrite now adopts the *incoming* snippet's format rather than preserving the existing scalar's envelope (per @sambsnyd's preference in the earlier discussion on this PR):

- **block → block, same style**: keep the existing envelope (chomp indicator, indent, trailing whitespace bounding the next sibling), swap only the body via `BlockScalar.withBody`.
- **anything → block**: return a block scalar built from the incoming snippet's value.
- **anything → plain**: return a plain scalar with the incoming value.

Because a block scalar's sibling-separator linebreak lives inside its trailing content (rather than the next entry's prefix), a block↔plain swap requires cross-sibling coordination. `mergeScalar` records the direction of the swap on the enclosing mapping cursor via a `SIBLING_BOUNDARY_REPAIR` message; `visitMapping` consumes it and, for `BLOCK_TO_PLAIN`, prepends a linebreak onto the next entry's prefix so the plain value doesn't glue onto the following sibling. `PLAIN_TO_BLOCK` is a no-op — `YamlParser` normalises the incoming block scalar's chomp semantics on the way in, and the sibling boundary falls out of either the next entry's existing prefix or the document tail.

## `ShiftFormatLeftVisitor`

Two fixes so `CoalesceProperties` (and `UnfoldProperties`, which shares this visitor) can dedent through block scalars correctly:

1. **New `visitScalar` override** — for FOLDED / LITERAL scalars in scope, shifts each body line and any trailing whitespace stored inside `Yaml.Scalar.value` by the same amount as the surrounding mapping-entry prefixes. Splits on `\n` and preserves embedded `\r` on interior lines so CRLF-authored files round-trip.

2. **Drops the `if (e.getPrefix().contains("\n"))` guard on `visitMappingEntry`.** The guard was legacy defensive code from before `shiftPrefix()` existed and incidentally skipped exactly the entries we need to dedent — those whose bare-whitespace prefix carries only the indent, because the boundary newline lives in the preceding block scalar's `value`. `shiftPrefix()` already handles the guard's original edge cases (empty prefix, same-line comment) via its per-split-element logic.

Kept out of `ShiftIndentVisitor` intentionally — its consumer (`MergeYaml` new-entry alignment) shifts entry prefixes at the top of the subtree only and doesn't currently exercise a block-scalar-body path, so this stays scoped to the coalesce case.

## `CoalescePropertiesVisitor`

With the `ShiftFormatLeftVisitor` fix in place, mappings containing block scalars can be coalesced without corrupting layout. New tests cover:

- Block-scalar sibling of a plain entry (`java_opts: >-` next to `after: tail`) — previously the naive coalesce absorbed `after: tail` into the block-scalar body.
- Lone block scalar (`inner: |`) — coalesces to `outer.inner: |` with the body properly re-indented.
- Deeply nested block scalar — multi-level coalesce compounds correctly.
- CRLF round-trip through the whole coalesce path.

## Test plan

- [x] `./gradlew :rewrite-yaml:test` green.
- New `MergeYamlTest` cases: `overwritePlainOverBlockFoldedAdoptsPlainFormat`, `overwritePlainOverBlockLiteralAdoptsPlainFormat`, `overwriteBlockOverPlainAdoptsIncomingBlockFormat`, `overwriteBlockOverBlockSameStyleKeepsExistingEnvelope`, plus three regression tests confirming block-scalar snippets without a trailing `\n` still produce valid output: `overwriteBlockStripOverPlainNoTrailingNewline`, `overwriteBlockClipOverPlainNoTrailingNewline`, `overwriteBlockOverPlainAsLastEntryNoTrailingNewline`.
- New `CoalescePropertiesTest` cases: `coalescesMappingSiblingOfBlockScalar`, `coalescesLoneBlockScalar`, `coalescesDeeplyNestedBlockScalar`, `coalescesBlockScalarPreservingCrlfBodyLineEndings`.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/MergeYamlVisitor.java
@@ -23,6 +23,7 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.style.GeneralFormatStyle;
 import org.openrewrite.style.Style;
 import org.openrewrite.yaml.MergeYaml.InsertMode;
+import org.openrewrite.yaml.trait.BlockScalar;
 import org.openrewrite.yaml.tree.Yaml;
 
 import java.util.ArrayList;
@@ -388,9 +389,14 @@ public class MergeYamlVisitor<P> extends YamlVisitor<P> {
     }
 
     private Yaml.Scalar mergeScalar(Yaml.Scalar y1, Yaml.Scalar y2) {
-        String s1 = y1.getValue();
-        String s2 = y2.getValue();
-        return !s1.equals(s2) && !acceptTheirs ? y1.withValue(s2) : y1;
+        BlockScalar.Matcher matcher = new BlockScalar.Matcher();
+        BlockScalar existing = matcher.get(new Cursor(null, y1)).orElse(null);
+        String s1 = existing != null ? existing.getBody() : y1.getValue();
+        String s2 = matcher.get(new Cursor(null, y2)).map(BlockScalar::getBody).orElseGet(y2::getValue);
+        if (s1.equals(s2) || acceptTheirs) {
+            return y1;
+        }
+        return existing != null ? existing.withBody(s2) : y1.withValue(s2);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


steve-aom-elliott:
Currently trying to determine the preferred outcome of starting with
```yaml
blah: >+
  some
  text
  here
xyz: 123
```
and wanting to merge `blah: other`. i.e. would it be
```yaml
blah: >+
  other
xyz: 123
```
or would it be
```yaml
blah: other
xyz: 123
```

In a similar vein, if we wanted to go from a non-block scalar, would we want to allow for trying to merge in a block scalar?

sambsnyd (approved):

PR #8158, opened 2026-06-30
Fall back to anonymous JAR download on 4xx when Maven credentials are rejected

Jammy-Louie:
## Summary
- `MavenArtifactDownloader` now retries the JAR download anonymously when `settings.xml` credentials are rejected by the remote repository (4xx), mirroring `MavenPomDownloader.requestAsAuthenticatedOrAnonymous()`. Credentials are applied first, so existing authenticated downloads against private repositories are unchanged and the anonymous attempt is only a fallback.
- Fix `LocalMavenArtifactCache` cache filename missing the hyphen before the classifier (`foo-1.0.0recipes.jar` → `foo-1.0.0-recipes.jar`).
- Include the classifier in the `MavenDownloadingException` message so artifact coordinates are complete during troubleshooting.

- Split from the first commit of #7447, to land this with minimal risk to existing behavior. A follow-up PR will mirror Apache Maven's anonymous-first behavior.

## Test plan
- [x] `fallsBackToAnonymousWhenServerReturns401`
- [x] `fallsBackToAnonymousWhenCredentialsRejected`
- [x] `downloadDependencies` / `downloadDependenciesWithClassifier`

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/cache/LocalMavenArtifactCache.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/cache/LocalMavenArtifactCache.java
@@ -85,7 +85,7 @@ public class LocalMavenArtifactCache implements MavenArtifactCache {
 
         return resolvedPath.resolve(dependency.getArtifactId() + "-" +
                                     (dependency.getDatedSnapshotVersion() == null ? dependency.getVersion() : dependency.getDatedSnapshotVersion()) +
-                                    (dependency.getRequested().getClassifier() == null ? "" : dependency.getRequested().getClassifier()) +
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks for diving in! I like the idea of splitting up the complexity to land this smaller part first. I've gone ahead and closed my earlier effort already.

Looking over your changes I've applied a few small fixes:
- do not retry for 408 timeouts, 425 or 429 rejected requests
- also retry anonymously when only header auth is present
- swap the evaluation order to check the `int` response code before the more complex authentication

Let me know if you agree with these changes, and whether you had any further plans before marking this as ready for review or merging.

Approved to merge from my side.

Jammy-Louie:
Changes LGTM

PR #8166, opened 2026-07-03
Add AttributeValue trait for non-literal annotation attribute values

### LINKED ISSUE #8160: `Annotated` trait cannot read non-literal attribute values (class literals, enum constants, constant references)
### What's missing

`org.openrewrite.java.trait.Annotated` exposes attribute values only via:

```java
Optional<Literal> getAttribute(String attribute)
Optional<Literal> getDefaultAttribute(@Nullable String defaultAlias)
```

Both resolve the value through `new Literal.Matcher().get(...)`. The `Literal` trait wraps only `J.Literal` (or a `J.NewArray` whose elements are *all* `J.Literal`). So these methods return `Optional.empty()` for any attribute whose value isn't a compile-time literal — including very common annotation shapes:

- **Class literals** — `@WebMvcTest(excludeAutoConfiguration = SecurityAutoConfiguration.class)` → value is a `J.FieldAccess`
- **Enum constants** — `@Retention(RetentionPolicy.RUNTIME)` → `J.FieldAccess`
- **Constant references** — `@Foo(name = Constants.NAME)` (`J.FieldAccess`) or `@Foo(name = NAME)` for a `static final` in scope (`J.Identifier`)
- **Arrays of any of the above** — `@Foo(exclude = {A.class, B.class})` → a `J.NewArray` of `J.FieldAccess`, which `Literal.Matcher` rejects

### Example

```java
new Annotated.Matcher("@org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest")
    .asVisitor((annotated, ctx) -> {
        // Optional.empty(): value is SecurityAutoConfiguration.class (J.FieldAccess), not a J.Literal
        Optional<Literal> attr = annotated.getAttribute("excludeAutoConfiguration");
        ...
    });
```

There is no way to reach the underlying `J.FieldAccess`/`J.Identifier` through the trait, so class/enum/constant attribute values are unreachable.

### Proposed solution

Rather than adding a new accessor, **expand the existing `getAttribute(...)` / `getDefaultAttribute(...)`** so they no longer return `Optional.empty()` for non-literal values. Concretely, broaden the underlying `Literal` trait and its `Literal.Matcher` to also match:

- `J.FieldAccess` — class literals (`X.class`), enum constants, and `static`-constant references
- `J.Identifier` — simple-name constant references
- `J.NewArray` whose elements include the above (today it only accepts arrays of `J.Literal`)

`Literal` already implements `Trait<Expression>`, so `getTree()` can already carry these nodes; `getValue(...)` can resolve them where a value is knowable (e.g. the referenced type for a class literal, the enum/constant for a field access) or otherwise expose the reference for the caller to inspect. This keeps the existing method signatures and simply widens what they can surface.

(If widening `Literal` isn't desirable semantically, the equivalent "expand, don't add" alternative is to generalize the existing accessors' return type from `Optional<Literal>` to `Optional<Expression>`.)

### Workaround

Read `J.Annotation.getArguments()` directly, find the `J.Assignment` for the attribute, and handle `J.NewArray` / `J.FieldAccess` / `J.Identifier` by hand — which is exactly what the trait is meant to abstract.

### Context

Surfaced writing a Spring Boot 4.0 migration recipe that appends `UserDetailsServiceAutoConfiguration` to any `exclude`/`excludeAutoConfiguration` attribute already excluding `SecurityAutoConfiguration`; the exclusion values are class literals, so `Annotated` couldn't be used.

MBoegers:
## What's changed?

A new cursor-bearing `AttributeValue` trait in `org.openrewrite.java.trait` that classifies every shape a JLS §9.7.1 annotation element value can take — literals, class literals, enum constants, constant references, nested annotations, and arrays — plus new `Annotated#getAttributeValue(String)` / `#getDefaultAttributeValue(@Nullable String)` accessors returning it.

- `getElements()` iterates braced arrays and the brace-less single-element form (`exclude = A.class`) uniformly, one cursor-bearing value per element for surgical edits
- `getClassValue()` / `isClassLiteral(fqn)` resolve class literals; `getReferencedField()` / `isEnumConstant(fqn, name)` cover enum constants and constant references in qualified, fully-qualified, and statically imported spellings
- `getConstantValue()` resolves constant references and constant expressions (`Constants.NAME`, `"a" + "b"`) to the compiler's fold recorded in `JavaType.Annotation.ElementValue` on the annotated declaration — including constants from binary dependencies
- `asLiteral()` / `asAnnotated()` bridge to the existing traits; accessor naming mirrors `SingleElementValue`'s constant/reference duality
- `Optional.empty()` now means only "attribute not explicitly written"; a present value whose typed accessors return null means "present but not resolvable that way" — the two conditions the `Optional<Literal>` accessors conflate
- Bug fix: `Literal` now matches empty array initializers — `{}` parses with a phantom `J.Empty` element, so `@Foo(x = {})` was indistinguishable from an absent attribute
- Groovy/Kotlin behavior is pinned by tests in both modules (list literals degrade to `EXPRESSION`, no constant fold, enum constants classify as `CONSTANT_REFERENCE` since neither type mapping sets `Flag.Enum`; Kotlin `A::class` / `A::class.java` classify as class literals)

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite/issues/8160 — class literals, enum constants, and constant references were unreachable through `Annotated`, forcing recipes to hand-roll `J.Annotation#getArguments()` walking (e.g. `MinimumJreConditions` in rewrite-testing-frameworks). Motivating case: a Spring Boot 4.0 migration recipe appending `UserDetailsServiceAutoConfiguration.class` to `exclude` attributes that are arrays of class literals.

## Any additional context

The existing `Optional<Literal>` accessors are untouched and not yet deprecated; all current consumers keep their exact behavior. The one deliberate behavior change is the `{}` fix: an explicitly empty array attribute now yields a present `Literal` with zero values, matching runtime reflection semantics (an annotation attribute is never "absent" at runtime; `{}` is an empty array). Per an org-wide `FindMethods` run over the recipe artifacts, only two artifacts currently consume the affected accessors and may be affected: https://app.moderne.io/results/C7GrmeXMY — follow-up PRs will add the one-line guards there and migrate `SpringRequestMapping` to the new API as its first adoption.

Planned follow-ups (separate PRs): docs (`traits.md`), rewrite-recipe-starter, a migration recipe in rewrite-rewrite, and only then `@Deprecated` on the old accessors.

### DIFF
--- a/rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
+++ b/rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
@@ -16,6 +16,7 @@
 package org.openrewrite.groovy;
 
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
 import org.openrewrite.java.trait.Annotated;
 import org.openrewrite.marker.SearchResult;
 import org.openrewrite.test.RewriteTest;
@@ -23,9 +24,13 @@ import org.openrewrite.test.RewriteTest;
 import static org.openrewrite.groovy.Assertions.groovy;
 
 /**
- * Pins the {@link org.openrewrite.java.trait.AttributeValue} degradations documented
- * for Groovy sources: no constant folding, no enum discrimination, list literals as
- * opaque expressions.
+ * The {@link org.openrewrite.java.trait.AttributeValue} behavior on Groovy sources,
+ * asserting the same semantics the trait has on javac-attributed Java sources.
+ * Tests annotated {@link ExpectedToFail} document known Groovy parser/type-mapping
+ * gaps: property-access references carry no {@code fieldType}
+ * ({@code GroovyParserVisitor#visitPropertyExpression}), no
+ * {@code JavaType.Annotation} element values are built (no constant folding), and
+ * list literals are {@code G.ListLiteral}, not {@code J.NewArray}.
  */
 class AttributeValueTraitTest implements RewriteTest {
 
@@ -91,8 +96,9 @@ class AttributeValueTraitTest implements RewriteTest {
         );
     }
 
+    @ExpectedToFail("Groovy list literals are G.ListLiteral, not J.NewArray; getElements() cannot normalize them from rewrite-java")
     @Test
-    void listLiteralIsAnOpaqueExpression() {
+    void listLiteralArray() {
         rewriteRun(
           spec -> spec.recipe(RewriteTest.toRecipe(() -> new Annotated.Matcher("@Example")
             .asVisitor(a -> SearchResult.found(a.getTree(),
@@ -114,7 +120,7 @@ class AttributeValueTraitTest implements RewriteTest {
                   String[] tags()
               }
 
-              /*~~(EXPRESSION:elements=1)~~>*/@Example(tags = ["a", "b"])
+              /*~~(ARRAY:elements=2)~~>*/@Example(tags = ["a", "b"])
               class Test {
               }
               """
@@ -122,8 +128,9 @@ class AttributeValueTraitTest implements RewriteTest {
         );
     }
 
+    @ExpectedToFail("GroovyParserVisitor#visitPropertyExpression attaches no type to the idiomatic bare class reference")
     @Test
-    void bareClassReferenceDegradesToConstantReference() {
+    void bareClassReference() {
         rewriteRun(
           spec -> spec.recipe(RewriteTest.toRecipe(() -> new Annotated.Matcher("@Example")
             .asVisitor(a -> SearchResult.found(a.getTree(),
@@ -145,7 +152,7 @@ class AttributeValueTraitTest implements RewriteTest {
                   Class type()
               }
 
-              /*~~(CONSTANT_REFERENCE:field=false:class=false)~~>*/@Example(type = String)
+              /*~~(CLASS_LITERAL:field=false:class=true)~~>*/@Example(type = String)
               class Test {
               }
               """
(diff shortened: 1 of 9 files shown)


rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
```diff
@@ -0,0 +1,233 @@
...
+              @Example(tags = ["a", "b"])
+              class Test {
+              }
+              """,
+            """
+              @interface Example {
+                  String[] tags()
+              }
+
+              /*~~(EXPRESSION:elements=1)~~>*/@Example(tags = ["a", "b"])
```
timtebeek: `elements` having a size of `1` is tripping me up a bit here; would it not be 2 for a & b?

rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
```diff
@@ -0,0 +1,233 @@
...
+            """
+              enum E {
+                  ONE, TWO
+              }
+
+              @interface Example {
+                  E e()
+              }
+
+              /*~~(CONSTANT_REFERENCE:isEnum=false)~~>*/@Example(e = E.ONE)
```
timtebeek: Surprised to see this `false`; do we have any cases where we do match `isEnumConstant`?

rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
MBoegers: This test shows pins down that for groovy the trait does not extract the`G.ListLiteral` elements; it handles it as a list. therefore `getElements("tags") = [ singletonList("a", "b") ]` which has size 1.
The behavior differs for `J.NewArray` where we unpack the list. This was meant as a simplification for implementation, but it is an unpleasant asymetry.

rewrite-groovy/src/test/java/org/openrewrite/groovy/AttributeValueTraitTest.java
MBoegers: groovy (and infact kotlin as well) thing. 
Both `TypeMapping`s do not set the Enum flag as Java does. That leads to a degradation to ConstantRef (`static final`) because `fieldType.hasFlags(Flag.Enum)`.
To overcome `KotlinTypeMapping` and `GroovyParserVisitor` are to fix.

timtebeek:
Given that it's such a small change I propose that we lump in the Kotlin enum parser fix here before we merge
- https://github.com/openrewrite/rewrite/pull/8178/

timtebeek (approved):

PR #8167, opened 2026-07-03
Add NpmDependency and NugetDependency traits

MBoegers:
## What's changed?
Two new traits:
- `org.openrewrite.javascript.trait.NpmDependency` — matches a dependency declaration in any of package.json's top-level dependency sections (`dependencies`, `devDependencies`, `peerDependencies`, `optionalDependencies`, `bundledDependencies`) and joins it with its version constraint and resolved version from the `NodeResolutionResult` marker. Falls back to the constraint's version floor (e.g. `^4.18.0` → `4.18.0`) when no resolution is available; protocol constraints (`file:`, `github:`) and open ranges yield no version.
- `org.openrewrite.csharp.trait.NugetDependency` — matches `<PackageReference>` tags in MSBuild project files and collects candidate versions per target framework from the `MSBuildProject` marker, preferring resolved over requested versions.

## What's your motivation?
Recipes over package.json / MSBuild project files currently have to hand-roll this matching and marker lookup. These traits are the analog of `org.openrewrite.maven.trait.MavenDependency` for the npm and NuGet ecosystems.

## Checklist

### DIFF
--- a/rewrite-csharp/src/main/java/org/openrewrite/csharp/trait/NugetDependency.java
+++ b/rewrite-csharp/src/main/java/org/openrewrite/csharp/trait/NugetDependency.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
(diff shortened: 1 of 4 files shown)


timtebeek:
I'd expect the majority of recipes for NPM and NuGet to be written in TS / C#, which can't use these Java traits. Do we really expect enough Java recipes to warrant a trait and expose that as a public API here? Or can we move these traits closer to where they are used, and make them package private there?

MBoegers:
good point, let me incubate them where I created them in the first place and let's see if needed to move to core later.

PR #8168, opened 2026-07-03
Fix `KotlinTemplate` stub generation for annotations on methods and variables

MBoegers:
## What's changed?

`AnnotationTemplateGenerator` hardcoded Java syntax for the dummy declarations that annotation template stubs attach to (` void $method() {}` and ` int $variable;`). This change extracts them into protected hooks — `addDummyMethod(StringBuilder)`, `addDummyVariable(StringBuilder)`, and the `variable()` helper — alongside the existing `addDummyClass`/`addDummyAnnotationType` hooks from #6522, and overrides them in `KotlinAnnotationTemplateGenerator` with Kotlin syntax.

## What's your motivation?

Any `KotlinTemplate` applied with annotation coordinates (e.g. `annotation.getCoordinates().replaceArguments()`) on a method or property failed, because the generated stub mixed Java member syntax into a Kotlin source:

```
Could not parse as Java:
class Test{/*__TEMPLATE_cfcc2025-6662__*/@Example("RedundantSuppression")
 void $method() {}}
annotation class `$Placeholder` {}
```

Two subtleties beyond the plain syntax swap:

- Stub fragments are concatenated on a single line, and a `return ...;` is appended when the enclosing function declares a return type — so the Kotlin stubs need explicit `;` terminators (`` val `$variable`: Int = 0; ``), otherwise a local `val` annotation inside `fun foo(): Int` produced the unparseable `` val `$variable`: Int = 0return null; ``.
- `variable()` (used to stub out locals that precede the templated annotation) emitted Java declaration order (`Int y;`); the Kotlin override emits `val y: Int`.

Base implementations keep byte-identical output, so Java and Groovy templates are unaffected.

- `ScalaAnnotationTemplateGenerator` has the same latent gap (it still inherits the Java method/variable stubs), left for a follow-up since rewrite-scala has no annotation-template test coverage yet.
- Test coverage spans the distinct stub-assembly paths: method in class, top-level function, property, class, local `val` (with enclosing return type, and preceded by a typed local), parameter substitution (`#{any()}`), and cursor-at-declaration via `addAnnotation`.

## Checklist

- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/recipes/recipes/recipeconventionsandbestpractices)
- [x] I've used the [IntelliJ IDEA auto-formatter](https://docs.openrewrite.org/reference/building-openrewrite-from-source#code-style) on affected files

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/internal/template/AnnotationTemplateGenerator.java
@@ -45,9 +45,9 @@ public class AnnotationTemplateGenerator {
 
         J j = cursor.getValue();
         if (j instanceof J.MethodDeclaration) {
-            after.insert(0, " void $method() {}");
+            addDummyMethod(after);
         } else if (j instanceof J.VariableDeclarations) {
-            after.insert(0, " int $variable;");
+            addDummyVariable(after);
         } else if (j instanceof J.ClassDeclaration) {
             addDummyClass(cursor, after);
         }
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Good to fix this generically, thanks!

PR #8169, opened 2026-07-03
Match Apache Maven's anonymous-first authentication for POM and JAR downloads

Jammy-Louie:
## Summary

- Follow-up to #8158. This brings both download paths onto Apache Maven Resolver's model: request anonymously first, send credentials only when the server challenges with a 4xx, and cache per host which repositories required auth so later requests authenticate preemptively without repeating the anonymous probe.

1. **Anonymous-first requests.** `MavenPomDownloader` (reachability probes, jar-existence HEAD, and body/metadata GETs) and `MavenArtifactDownloader` now request without credentials first and retry with credentials on a credential-rejecting 4xx. This also stops sending credentials to repositories that serve anonymously.
2. **Per-host auth cache.** A session-scoped `host:port` set on the `ExecutionContext` records which endpoints required authentication, mirroring Maven Resolver's `BasicAuthCache` (and the existing `getUnreachableEndpoints` pattern). Once a host is known to require auth, both downloaders authenticate preemptively. `MavenArtifactDownloader` gains additive `ExecutionContext` constructors, and `JavaRewriteRpc` passes its context through so the jar phase shares the cache warmed during POM resolution.

The first request to a private host costs one extra anonymous 401 probe before authenticating; the cache limits this to once per host per session, the same one-time cost Maven pays.

## Test plan
- [x] `MavenPomDownloaderTest#doesNotSendCredentialsWhenRepositoryServesAnonymously`
- [x] `MavenPomDownloaderTest#authenticatesPreemptivelyAfterCredentialsRequired`
- [x] `MavenArtifactDownloaderTest#retriesWithCredentialsWhenAnonymousReturns401`, `#retriesWithHttpHeaderAuthWhenAnonymousReturns401`
- [x] `MavenArtifactDownloaderTest#doesNotFallBackToAnonymousOnTransientClientError`
- [x] `MavenArtifactDownloaderTest#authenticatesPreemptivelyAfterFirstChallengeForSameHost`
- [x] Full `rewrite-maven` module test suite


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -1085,8 +1085,8 @@ private boolean jarExistsForPomUri(MavenRepository repo, String pomUrl) {
         try {
             try {
                 return Failsafe.with(retryPolicy).get(() -> {
-                    HttpSender.Request authenticated = applyAuthenticationAndTimeoutToRequest(repo, httpSender.head(jarUrl)).build();
-                    try (HttpSender.Response response = httpSender.send(authenticated)) {
+                    HttpSender.Request unauthenticated = applyTimeoutToRequest(repo, httpSender.head(jarUrl)).build();
+                    try (HttpSender.Response response = httpSender.send(unauthenticated)) {
```
timtebeek: Shouldn't we consult the cache here too to see if a header would be required?

rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenAuthenticationCache.java
```diff
@@ -0,0 +1,73 @@
...
+ * required credentials. Once an endpoint is known to require authentication, subsequent requests can send
+ * credentials preemptively instead of paying another anonymous round-trip.
+ * <p>
+ * This mirrors the per-session {@code BasicAuthCache} that Apache Maven Resolver keeps on its HTTP client:
+ * anonymous-first on the first contact with a host, then preemptive once the host has challenged. The cache
+ * lives on the {@link ExecutionContext}, so it is shared across POM/metadata resolution and artifact download
+ * whenever they run against the same context, and it never crosses process or serialization boundaries.
+ */
+public final class MavenAuthenticationCache {
+    private static final String MESSAGE_KEY = "org.openrewrite.maven.internal.authenticationRequiredEndpoints";
```
timtebeek: Whilst there's some elegance to having this as a named separate class, it does move away from an established pattern we already have of MavenExecutionContextView maintaining a list of endpoints that can't be reached:

https://github.com/openrewrite/rewrite/blob/82f7d5b73fe9558103bbee817ff02757b021d7e9/rewrite-maven/src/main/java/org/openrewrite/maven/MavenExecutionContextView.java#L81-L93

What's the motivation to branch this off into a separate class as opposed to having this live right beside the existing pattern, given that they are two sides of the same problem?

I fear having this one separate makes it harder to see the parallels between the two.

timtebeek (commented):
Couple guiding questions; nothing that can't be fixed I think. Let me know your thoughts and happy to work this in if you agree.

PR #8171, opened 2026-07-03
Support Gradle wrapper version detection for custom distribution URLs

bmuschko:
## What's changed

`UpdateGradleWrapper` (invoked with a wildcard version, e.g. via `MigrateToGradle7`'s `version: 7.x`) threw `IllegalStateException: Unsupported distribution url for Gradle wrapper version detection` when the existing `distributionUrl` pointed at a custom repository whose path did **not** start with `/artifactory` — e.g. `https://nexus.company.com/gradle-distributions/gradle-6.9.1-bin.zip`, where the name is only in the host.

Version detection was keyed off the literal `/artifactory` path segment, but a Gradle distribution can live in any Maven repository (Artifactory served at the host root, Nexus, an S3 bucket, a CDN, etc.), so it shouldn't depend on a particular host or path name.

`GradleWrapper.determineGradleVersion` now handles custom URLs in three clear steps:

1. gradle.org host — unchanged.
2. `selectPrivateArtifactoryVersion(...)` — best-effort JFrog storage-API listing (only when served under `/artifactory`); returns `Optional`, and on no-match/failure falls through instead of throwing.
3. `substituteCustomDistributionVersion(...)` — repo-agnostic fallback: resolve the concrete version from the public Gradle version list (for wildcards) and substitute it into the existing URL, preserving the custom host and path.

This matches the recipe's documented contract (queries `downloads.gradle.org` for available releases while preferring the existing artifact-repository URL) and removes the hard failure.

### Notes / limitations
- Exact versions on a custom URL resolve fully offline (pure substitution, no gradle.org call) — unchanged behavior.
- Wildcard versions on a custom URL still need `services.gradle.org` to resolve the concrete version. If that host is blocked, the update fails with a network error rather than the previous `IllegalStateException` (outcome unchanged; previously it also failed). Private-repo-first resolution for fully air-gapped shops is left as a possible follow-up.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/util/GradleWrapper.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/util/GradleWrapper.java
@@ -43,6 +43,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


shanman190:
I don't think we should be able to reach this point based on the configuration that was shared.

https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle%2Fsrc%2Fmain%2Fjava%2Forg%2Fopenrewrite%2Fgradle%2FUpdateGradleWrapper.java#L137-L146

bmuschko:
The failure comes from the `UpdateGradleWrapper` that `MigrateToGradle7` declares internally, which has a wildcard version and **no** `wrapperUri` ([`gradle-7.yml`](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/resources/META-INF/rewrite/gradle-7.yml)):

```yaml
- org.openrewrite.gradle.UpdateGradleWrapper:
    version: 7.x
    addIfMissing: false
```

Because `wrapperUri == null` there, `getGradleWrapper` takes the **L143** branch — `GradleWrapper.create(distributionUrl, distribution, version, ctx)` — not L140. That calls `determineGradleVersion`, where the existing `https://repo.example.com/gradle-distributions/gradle-6.9.1-bin.zip` matches none of the branches (not a gradle.org host, path doesn't start with `/artifactory`, and `7.x` isn't an `ExactVersion`), so it hits the `IllegalStateException` at [`GradleWrapper.java:147`](https://github.com/openrewrite/rewrite/blob/main/rewrite-gradle/src/main/java/org/openrewrite/gradle/util/GradleWrapper.java#L147).

The stack trace confirms this — the frame is `getGradleWrapper(UpdateGradleWrapper.java:143)`, the non-`wrapperUri` path. If it had come from the explicit recipe, it would read L140.

And it reproduces without the explicit recipe at all: running `mod run --recipe org.openrewrite.gradle.MigrateToGradle7` on its own throws the same `IllegalStateException`, since the only `UpdateGradleWrapper` involved is the `version: 7.x` one nested in `MigrateToGradle7`.

shanman190:
I see. We've had a few iterations on this, but the most recent one was that dynamic versions and a non-standard URL is an unsupported path.

In enterprises, often services.gradle.org is unavailable, so by falling back onto it for a non-standard URL we are very well trading one exception for a different one.

Maybe we just need to avoid the exception bubbling up out of the recipe and make it be a no-op instead. That way it maintains the custom urls get no dynamic versions and deeply chained dynamic recipes don't interrupt a recipe run.

bmuschko:
Agreed — for a non-standard URL that population is exactly the one that can't reach `services.gradle.org`, so resolving against it just trades one exception for another. I've reworked the PR to your suggestion: dynamic version + non-standard URL is now a **no-op** (the wrapper is left unchanged) rather than an exception, so a deeply chained run like `MigrateToGradle7` completes instead of interrupting. Exact versions on a non-standard URL still substitute offline, since that needs no external lookup. Added tests assert the no-op leaves the wrapper untouched *without* any call to gradle.org.

shanman190 (approved):

bmuschko:
@shanman190 Are you good with merging this PR or do you want someone else to look over it?

PR #8175, opened 2026-07-03
Add Reassignable trait

steve-aom-elliott:
## What

A new incubating `Reassignable` trait under `org.openrewrite.java.trait` that answers a syntactic question recipes tend to re-invent locally: *could this expression be the target of a plain assignment `target = ...` without introducing a compile error?*

## Why

- Recipes that mechanically rewrite `x.setFoo(...)` shapes into `x = x.rebuild()....build()` — or any other transformation that needs to reassign a receiver — currently either duplicate the walk-up-the-cursor-and-check-flags logic each time, or silently skip cases where the receiver is a field access. Concrete example: the `Reassignable`-shaped guard that landed in [`rewrite-jackson#153`](https://github.com/openrewrite/rewrite-jackson/pull/153) (`MigrateMapperSettersToBuilder#isReassignableReceiver`).

Extracting the check as a trait gives future recipes a one-liner and gives us one place to sharpen the semantics.

## Scope for v1

Deliberately narrow, matching the jackson recipe's semantics:

- **Matches** a `J.Identifier` referring to a non-final local declared in an enclosing block, and a `J.FieldAccess` whose target is a bare `J.Identifier` and whose field is non-final.
- **Rejects** method parameters (reassigning a parameter silently drops the caller's expected mutation), final locals and final fields, multi-hop qualified field access (`a.b.c`), array-element writes, parenthesized targets, and method-chain receivers.
- **Does not model** effectively-final semantics, definite assignment, or any control-flow analysis.

Follow-up: migrate the jackson recipe to use the trait once this lands.

## Test plan

- [x] `./gradlew :rewrite-java:test --tests 'org.openrewrite.java.trait.ReassignableTest'` — 5 cases (non-final local ✓, final local ✗, method parameter ✗, non-final `this.field` ✓, final `this.field` ✗).
- [x] `./gradlew :rewrite-java:test --tests 'org.openrewrite.java.trait.*'` — no regression in other trait tests.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/trait/Reassignable.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/trait/Reassignable.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


sambsnyd (commented):
Can you scan the codebase for places this could be used? I know there's the one in your recipe PR, but before we finalize the API it would be good to make sure this trait is applicable/handles variations

PR #8177, opened 2026-07-06
use gradle property assignment for distributionUrl

sullis:
## What's changed?

Add `org.openrewrite.gradle.UsePropertyAssignmentSyntax` (with `propertyName: distributionUrl`) to the `GradleBestPractices` recipe list so that the Gradle wrapper's `distributionUrl(...)` method call is rewritten to property assignment syntax:

```groovy
wrapper {
    distributionUrl("https://example.com/files/example.zip")
}
```

becomes:

```groovy
wrapper {
    distributionUrl = "https://example.com/files/example.zip"
}
```

## Testing

Added a test (`usePropertyAssignmentSyntaxForDistributionUrl`) to `GradleBestPracticesTest` covering the transformation.


timtebeek (approved):
Nice! Good to automate what you've seen in practice with recipes.

Issue #8170, opened 2026-07-03
Kotlin and Groovy type attribution gaps behind AttributeValue's degraded classification


MBoegers:
### What's the issue?

- #8166 introduced the `AttributeValue` trait (#8160), which classifies annotation attribute values syntax-first and discriminates enum constants via `Flag.Enum` (JVMS `ACC_ENUM`) on the referenced `JavaType.Variable` — the contract javac-attributed Java fulfills. On Kotlin and Groovy sources the same code degrades (enum constants classify as `CONSTANT_REFERENCE`, constant references don't fold), not because of the trait but because of parser/type-mapping gaps. The divergences are captured as `@ExpectedToFail` tests in `rewrite-groovy/.../AttributeValueTraitTest` and `rewrite-kotlin/.../AttributeValueTraitTest`, asserting the Java-aligned semantics — each flips red the moment its gap is fixed.
- https://github.com/openrewrite/rewrite/pull/8178 fixed the Kotlin part of this issue.

### The gaps

~**1. Kotlin: `Flag.Enum` is never set on enum entries** *(one-line fix)*~

~`KotlinTypeMapping.kt:1171` — `variableType(FirVariable, …)` computes flags via `mapToFlagsBitmap(visibility, modality, isStatic)` with no enum-entry branch, even though enum entries flow through this very method (`:519-520`) and the enum *class* does get the bit (`:446-447`, `flags or (1L shl 14)`). Fix: add the bit when `variable is FirEnumEntry`. Unblocks `AttributeValue#isEnumConstant(..)` on Kotlin.~

**2. Groovy: property-access references carry no `fieldType` at all**

`GroovyParserVisitor#visitPropertyExpression` converts the property name to a `J.Identifier` with `fieldType` hard-coded `null` — the resolved `FieldNode` is never looked up, although `GroovyTypeMapping.variableType(FieldNode)` exists and *declared* enum constants get proper attribution (`GroovyParserVisitor:850`). Groovy's enum-constant `FieldNode`s carry `ACC_ENUM` at `Phases.CANONICALIZATION` (set by `EnumHelper`), so attaching `typeMapping.variableType(classNode.getField(name))` when the receiver's static type resolves would fix enum discrimination *and* give `getReferencedField()` to all Groovy constant references, including idiomatic bare class refs (`@Foo(type = String)`). Needs scoping to statically unambiguous receivers (dynamic dispatch).

**3. Both: no `JavaType.Annotation` element values → no constant fold**

[...]

PR #8178, opened 2026-07-06
Kotlin: set Flag.Enum on enum entries in KotlinTypeMapping

timtebeek:
- Fixes gap 1 of #8170.

### What

`KotlinTypeMapping#variableType(FirVariable, …)` computed its flags via `mapToFlagsBitmap(visibility, modality, isStatic)` with no enum-entry branch — even though enum entries flow through this exact method and the enum *class* already gets the bit (`flags or (1L shl 14)`). As a result `Flag.Enum` was never set on enum entries, so `AttributeValue#isEnumConstant(..)` degraded to `CONSTANT_REFERENCE` on Kotlin sources instead of `ENUM_CONSTANT`.

This sets `1L shl 14` when the variable is a `FirEnumEntry`, aligning Kotlin with the javac-attributed Java contract the `AttributeValue` trait relies on.

### Test

Drops `@ExpectedToFail` from `AttributeValueTraitTest.enumConstant`, which now asserts `ENUM_CONSTANT:isEnum=true` and passes. The class javadoc is updated to no longer list the enum-entry gap.

- The remaining `@ExpectedToFail` cases (no `JavaType.Annotation` element values → constant fold, and `K.ListLiteral` array normalization — gaps 3 & 4 of #8170) are untouched.

### Stacking

- Stacked on #8160 (the `AttributeValue` trait + its `@ExpectedToFail` Kotlin/Groovy tests). Should merge after or with that PR.

### DIFF
--- a/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -1168,7 +1168,10 @@ class KotlinTypeMapping(
 
     @OptIn(SymbolInternals::class)
     fun variableType(variable: FirVariable, parent: Any?, signature: String): Variable {
-        val variableFlags = mapToFlagsBitmap(variable.visibility, variable.modality, variable.isStatic)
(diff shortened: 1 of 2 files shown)


MBoegers (approved):
Great to see this fixed!

PR #8182, opened 2026-07-06
C#: Preserve raw XML doc comments in CSharpVisitor.visitSpace

MBoegers:
## What's wrong?

`CSharpVisitor.visitSpace` eagerly replaced every `CsDocCommentRawComment` (as delivered by the RPC parser) with a freshly parsed structured `CsDocComment.DocComment` on *every* traversal. This made any read-only recipe return a modified tree for any C# file containing XML doc comments — the file registered as changed and reprinted lossily over RPC, dropping the XML attribute *values*:

```
-/// <param name="left">The left operand.</param>   -/// <see cref="System.Math"/>
+/// <param name>The left operand.</param>           +/// <see cref/>
```

Because `Cs.CompilationUnit implements JavaSourceFile`, plain `JavaIsoVisitor`-based recipes hit this too — which is how it surfaced (a Prethink run over C# code produced spurious diffs).

## What's the fix?

Keep the original raw comment when visiting the parsed tree leaves it unchanged (`visited == parsed`), only substituting the structured tree when a visitor actually modified it. `CsDocCommentVisitor` is identity-preserving (`ListUtils.map`/`withX`), so the guard holds for every read-only visitor — the same identity contract as the `JRightPadded` check already in this file.

## Anything else?

- Unit coverage (`CSharpVisitorDocCommentIdentityTest`, no .NET SDK needed): no-op visit returns the identical `Space` for both raw and structured doc comments; a modifying visit still swaps in the structured tree.
- RPC/integ coverage (`CSharpRecipeTest`): a read-only `JavaIsoVisitor` over the issue's exact sample leaves the source unchanged; tree identity is preserved; the structured-tree RPC print path is still exercised by an explicit conversion.

- Fixes moderneinc/customer-requests#2696

### DIFF
--- a/rewrite-csharp/src/integTest/java/org/openrewrite/csharp/CSharpRecipeTest.java
+++ b/rewrite-csharp/src/integTest/java/org/openrewrite/csharp/CSharpRecipeTest.java
@@ -15,20 +15,25 @@
  */
 package org.openrewrite.csharp;
 
+import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.Issue;
 import org.openrewrite.SourceFile;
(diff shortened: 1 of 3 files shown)


macsux:
Further investigation shows that this issue is caused by an obsolete code that used java side for parsing of XML. We have native xml parsing capabilities on c# side, so this "raw" concept is no longer relevant and will be removed. I'm closing this issue as it will be addressed as part of a proper long term fix.

PR #8183, opened 2026-07-06
Add the source file's imports when parsing context-sensitive method-parameter templates

MBoegers:
### What's changed?

`JavaTemplateParser.parseParameters` — the path behind `JavaCoordinates#replaceParameters()` — builds its stub with only the template's own `.imports(...)`. Unlike the block-statement path (`BlockStatementTemplateGenerator.contextTemplate`), it never embedded the enclosing compilation unit's imports, even for `contextSensitive()` templates.

As a result, a context-sensitive template that refers to a type by the simple name it's imported under could not resolve that type when the stub was compiled. The generated parameter came back with an unknown type, and because `JavaTemplateJavaExtension` rebuilds the method's `JavaType.Method` from the parsed parameters, the whole method declaration was flagged as having missing/malformed type information.

This teaches `parseParameters` to prepend the enclosing `JavaSourceFile`'s imports for context-sensitive templates, mirroring what `contextTemplate` already does for block statements.

### Testing

Added `JavaTemplateTest6Test#replaceParametersResolvesSimpleNamesThroughSourceFileImports`: it replaces a method's parameters with `List<String> names` via a context-sensitive template and asserts the resulting parameter type resolves through the file's `java.util.List` import. Fails before the change (unknown parameter type), passes after.

A companion change in rewrite-spring (`AutowiredFieldIntoConstructorParameterVisitor`) relies on this and will be opened separately.


MBoegers:
Note for reviewers — scope of this change

This fix touches a single path on purpose. `parseParameters` (behind `replaceParameters()`) was the **only** context-sensitive template path that embedded just the template's own imports into its stub and not the enclosing compilation unit's imports.

Every other context-sensitive path already prepends `cu.getImports()` — see `BlockStatementTemplateGenerator.contextTemplate`, which inserts each `cu.getImports()` (plus the package declaration) before compiling the stub. That is the established contract for "context-sensitive": the template is compiled as if it lived inside the original source file, so it can resolve types by their imported simple names.

Because `parseParameters` skipped that, a context-sensitive parameter template referencing a type by its imported simple name failed to resolve → the generated parameter, and the `JavaType.Method` rebuilt from it, came out without type attribution. This change just brings the parameter path in line with the existing contract; it's not new behavior, it's closing an inconsistency.

(Caveat unchanged by this PR: the import line resolving is necessary but not sufficient — the referenced type still has to be on the classpath for attribution to attach.)

timtebeek (approved):
Nicely found & fixed, thanks!

PR #8192, opened 2026-07-07
Fix MergeYaml indenting new entries to a column that mismatches siblings

bmuschko:
## What's changed?

When `MergeYaml` merges a snippet into an existing mapping, new entries were positioned by `autoFormat`, which derives indentation from the file's auto-detected indent size (`parentIndent + detectedIndentSize`). When a block's actual indentation deviates from that detected step, the new entry landed at the wrong column, producing invalid YAML.

For example, a `settings` block indented with a 4-space step in a file whose detected step is 2 would get a new entry at 2 spaces instead of 4:

```yaml
settings:
    access_days_threshold: 30
    ignore_dependencies: false
  retry_on_error: true          # <-- wrong: 2 spaces, breaks the file
```

New entries are now re-indented to match the existing sibling entries' actual indentation, preserving the relative indentation `autoFormat` produced for any nested content.

A new `ShiftFormatVisitor` is added as a bidirectional companion to the existing `ShiftFormatLeftVisitor`, shifting a subtree left or right by a fixed amount.

## What's your motivation?

New entries were merged at an indentation that did not line up with their siblings when the target block's indentation differed from the rest of the file, breaking the resulting YAML even though the recipe promises the indentation "will be adjusted to match."

The alignment is only applied when `shouldAutoFormat` is enabled and the target mapping has existing sibling entries; it is a no-op (delta 0) for well-behaved files, so existing behavior is unchanged.

## Checklist

- [x] I've added unit tests covering both shift directions, deeper nesting, and nested incoming snippets.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/ShiftFormatVisitor.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/ShiftFormatVisitor.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2026 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.yaml;
+
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.StringUtils;
+import org.openrewrite.yaml.tree.Yaml;
+
+import java.util.Arrays;
+
+/**
+ * Shifts the indentation of every element within {@code scope} by a fixed amount.
+ * A positive {@code shift} adds whitespace (moves right); a negative {@code shift}
+ * removes whitespace (moves left). The relative indentation between nested elements
+ * is preserved, so the whole subtree simply moves as a block.
+ */
+public class ShiftFormatVisitor<P> extends YamlIsoVisitor<P> {
+    private final Yaml scope;
+    private final int shift;
+
+    public ShiftFormatVisitor(Yaml scope, int shift) {
+        this.scope = scope;
+        this.shift = shift;
+    }
+
+    @Override
+    public Yaml.Sequence.Entry visitSequenceEntry(Yaml.Sequence.Entry entry, P p) {
+        Yaml.Sequence.Entry e = super.visitSequenceEntry(entry, p);
+        if (getCursor().isScopeInPath(scope) && e.isDash() && e.getPrefix().contains("\n")) {
+            e = e.withPrefix(shiftPrefix(e.getPrefix()));
+        }
+        return e;
+    }
+
+    @Override
+    public Yaml.Mapping.Entry visitMappingEntry(Yaml.Mapping.Entry entry, P p) {
+        Yaml.Mapping.Entry e = super.visitMappingEntry(entry, p);
+        if (getCursor().isScopeInPath(scope) && e.getPrefix().contains("\n")) {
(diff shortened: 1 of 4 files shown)


rewrite-yaml/src/main/java/org/openrewrite/yaml/ShiftFormatVisitor.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Shifts the indentation of every element within {@code scope} by a fixed amount.
+ * A positive {@code shift} adds whitespace (moves right); a negative {@code shift}
+ * removes whitespace (moves left). The relative indentation between nested elements
+ * is preserved, so the whole subtree simply moves as a block.
+ */
+public class ShiftFormatVisitor<P> extends YamlIsoVisitor<P> {
```
greg-at-moderne: Two suggestions worth considering:

1. I think it'd be better if the input was not the shift/offset to be moved, but the desired absolute indent. 
2. And if we do that, then the visitor should probably be renamed.

greg-at-moderne: Also. Judging by the length of it, I suspect this implementation might not be complete in terms of all possible cases.
Have you considered some code re-use with `org.openrewrite.yaml.format.IndentsVisitor`?

bmuschko: We do use `IndentsVisitor` (via `autoFormat`), but it can't do the alignment because it recomputes each entry's column absolutely as `parentIndent + detectedStep` — which is exactly the wrong number when a block's real step deviates, and rerunning it over the block would reformat untouched siblings. We need a *relative* translation of the finished subtree to the siblings' column, which is a different operation from `IndentsVisitor`'s absolute re-derivation — only the low-level helpers (`findIndent`/`shift`/`indentTo`) genuinely overlap.

bmuschko: Agree it reads better here — `alignToIndent` already knows the target column, so passing it and deriving the delta in the visitor removes that arithmetic.

Catch: this visitor is shared with `CoalesceProperties`/`UnfoldProperties`, whose op is inherently relative ("dedent this subtree by one level"). A subtree spans many columns, so there's no single absolute indent for it — only for its root — and an absolute API would force those callers into `rootIndent - level`.

Compromise: keep the relative core, add a thin `static toIndent(scope, targetColumn)` factory. MergeYaml uses `toIndent(...)`; the collapse callers stay relative. On the rename, I'd lean `ShiftIndentVisitor` if we keep it relative. WDYT?

dezsiszabi:
I'm the person who reported this at my company.

Thank you for starting this PR, here's hoping that a fix will be merged and released quickly,

I have a recipe which uses `MergeYaml` that I intend to run on all the repositories at work and I'm eagerly awaiting this fix before I do so.

rewrite-yaml/src/main/java/org/openrewrite/yaml/ShiftFormatVisitor.java
greg-at-moderne: Sounds good to me

bmuschko: Done in 2fcf30efc: renamed `ShiftFormatVisitor` → `ShiftIndentVisitor`, kept the relative-offset core, and added a thin `static toIndent(scope, targetColumn)` factory that derives the shift from the scope's current indent — so `MergeYaml` now passes the desired absolute column instead of computing a delta.

Consolidating the existing `ShiftFormatLeftVisitor` into this visitor (so the collapse callers reuse the relative core) is deferred until #8157 lands, since it edits that class and adds the block-scalar handling we'd want to inherit rather than duplicate. Follow-up checklist:

- [ ] Rebase on main once #8157 is merged
- [ ] Fold `ShiftFormatLeftVisitor` (block-scalar-aware) into `ShiftIndentVisitor`
- [ ] Repoint `CoalescePropertiesVisitor` / `UnfoldProperties` to `new ShiftIndentVisitor<>(scope, -n)`
- [ ] Delete `ShiftFormatLeftVisitor`
- [ ] Confirm Coalesce/Unfold tests (incl. #8157's block-scalar cases) stay green

PR #8193, opened 2026-07-08
Store DependencyInsight's dependency graph as a DAG to prevent OOM on ubiquitous transitive deps

### LINKED ISSUE #8181: Dependency Insight OOMs on ubiquitous transitive deps (jackson): DependencyTreeWalker enumerates all root→match paths with no DAG dedup
## Summary

`DependencyInsight` (Gradle/Maven) can OOM on a **single repository** when the target is a ubiquitous, deeply-transitive dependency such as `com.fasterxml.jackson*`. The underlying `DependencyTreeWalker.walkRecursive` enumerates **every root→match path** through the transitive dependency graph — it prunes cycles within the current path but has **no DAG-level dedup** — so a diamond-heavy graph (which jackson always produces) expands combinatorially, and the per-match callback appends each path into a non-deduplicated `DependencyGraph`. One repo materialized **~37 million `DependencyGraph$DependencyNode`** and blew a ~16 GB heap.

Observed while running "Dependency Insight for Gradle and Maven" for `com.fasterxml.jackson*` across a large repo set; the `mod` CLI heap-dumped three times at ~15–16 GB.

## Evidence (Eclipse MAT on the 16 GB heap dump)

**71.5% of the heap (10.1 GB) is retained by a single `ForkJoinPool` worker thread** — i.e. one repository's in-flight recipe run — split across two structures it is building:

- `org.openrewrite.maven.graph.DependencyGraph` → **5.04 GB**, accumulated in `DependencyGraph.append` (`DependencyGraph.java:59`)
- a `HashMap` in `DependencyInsight$1.collectMatchingDependencies` → **5.05 GB**

Object histogram (shallow):

| Instances | Class |
|---|---|
| 37,262,813 | `org.openrewrite.maven.graph.DependencyGraph$DependencyNode` |
| 37,264,087 | `java.util.TreeMap` (≈ one per node) |
| 37,262,853 | `java.util.TreeSet` (≈ one per node) |
| 44,936,020 | `java.util.TreeMap$Entry` |
| 7,672,348 | `DependencyGraph$ConfigurationNode` |

Retaining thread stack (dozens of `walkRecursive` frames):

```
DependencyGraph.formatDependency (DependencyGraph.java:116)
DependencyGraph.append           (DependencyGraph.java:59)
DependencyInsight$1.lambda$collectMatchingDependencies$5 (DependencyInsight.java:209)
DependencyTreeWalker$Matches.lambda$collect$2            (DependencyTreeWalker.java:198)
DependencyTreeWalker.walkRecursive (DependencyTreeWalker.java:95)
DependencyTreeWalker.walkRecursive (DependencyTreeWalker.java:99)   ← recursion, many frames deep
DependencyTreeWalker.walkRecursive (DependencyTreeWalker.java:99)
...
```

## Root cause

`rewrite-maven` `DependencyTreeWalker.walkRecursive`:

```java
private static void walkRecursive(ResolvedDependency dependency, ... Deque<ResolvedDependency> path, Callback callback) {
    // Cycle detection - check if we've already visited this dependency in the current path
    if (containsDependency(path, dependency)) return;      // prunes cycles WITHIN a path only
    path.addFirst(dependency);
    if (matcher == null || matcher.matches(...)) {
        callback.accept(dependency, new ArrayDeque<>(path)); // fires (and copies the path) per PATH
    }
    for (ResolvedDependency child : dependency.getDependencies()) {
        walkRecursive(child, matcher, path, callback);       // re-walks shared subtrees per path
    }
    path.removeFirst();
}
```

`containsDependency` only detects cycles *within the current path* — there is **no visited-set spanning the DAG**. So a dependency reachable through _k_ distinct paths (diamonds) has its entire subtree walked _k_ times, and the count of root→match paths grows combinatorially with graph depth/branching. jackson (`jackson-databind` → `jackson-core` + `jackson-annotations`, pulled in transitively by dozens of libraries) is a worst case.

`DependencyTreeWalker.Matches` itself is fine — `scopeToDirectDependency` / `directDependencyToTargetDependency` are `Set<GroupArtifactVersion>` and de-duplicate. The blow-up is entirely in the **optional per-match `Callback`**: `DependencyInsight.collectMatchingDependencies` (`DependencyInsight.java:206/209`) passes a callback that calls `DependencyGraph.append(...)` for every path, and `DependencyGraph` is a **tree with a node (each carrying a `TreeSet`/`TreeMap`) per path segment**, never sharing nodes across paths. Millions of near-duplicate paths → tens of millions of nodes → OOM for one repo.

## Reproduction

Run `DependencyInsight` for `com.fasterxml.jackson*` (groupId `com.fasterxml.jackson*`, artifactId `*`) against a project with a large, diamond-heavy dependency graph that transitively includes jackson via many routes (most Spring Boot / big Gradle apps qualify). Heap grows into the many-GB range for a single module.

## Suggested direction

The information `DependencyInsight` needs is bounded — each `(directDependency → matchedDependency)` relationship (already captured, deduped, in `Matches`), plus (optionally) *one representative* path to explain each match. The fix is to stop materializing *all* paths:

- Give `walkRecursive` an optional **global visited-set** (dedup shared subtrees) for callers that don't need every path; or
- Have the `DependencyGraph` share nodes (be a **DAG**, keyed by GAV) instead of expanding into a per-path tree; or
- In `DependencyInsight`, emit **one shortest/representative path per `(direct, matched)` pair** rather than invoking the append callback for every enumerated path.

Any of these turns the work from `O(paths)` (exponential) into `O(nodes)` and eliminates the OOM. Happy to open a PR if you agree on the approach.

jkschneider:
## What's the problem?

`DependencyInsight` (Gradle/Maven) can OOM on a single repository when the target is a ubiquitous, deeply-transitive dependency such as `com.fasterxml.jackson*`. `DependencyTreeWalker.walkRecursive` enumerates **every** root→match path — it prunes cycles within the current path but has **no DAG-level dedup** — so a diamond-heavy graph expands combinatorially, and the per-match callback appends each path into a non-deduplicated `DependencyGraph`. One repo materialized ~37M `DependencyGraph$DependencyNode` and blew a ~16 GB heap.

- Full analysis in #8181.

## The fix

Give the walk a single **visited-set spanning the whole traversal** (not just the current path). It both prevents cycles and dedups subtrees reachable through multiple paths, so a match reachable via _k_ diamonds fires the callback **once** with a representative path instead of _k_ times. Traversal goes from `O(paths)` (exponential) to `O(nodes)` (linear).

This also makes good on the walker's existing Javadoc — *"Uses caching to avoid re-traversing duplicate dependencies"* — which the implementation never actually honored.

The visited-set is **per direct-dependency root** (each `walk(root, …)` gets its own), so attribution is preserved: the `SearchResult` markers on the POM/build.gradle are **unchanged** (DFS reachability is identical). Only the `DependenciesInUse` / `ExplainDependenciesInUse` data tables change — `count` becomes the number of direct dependencies a match is pulled in through, and the dependency tree shows one representative path per root instead of the combinatorial expansion.

## Benchmark

Reproducing the exact OOM structure — one matched dependency reachable from the root via `2^depth` diamond paths — and running the pre-fix algorithm against the post-fix one on the identical graph:

**Total traversal work — `O(2ᵈ)` → `O(d)`:**

| depth | graph nodes | OLD node-visits | NEW node-visits |
|------:|------------:|----------------:|----------------:|
| 10 | 31 | 4,093 | 31 |
| 18 | 55 | 1,048,573 | 55 |
| 22 | 67 | 16,777,213 | 67 |
| 24 | 73 | 67,108,861 | **73** |

**`DependencyGraph` built the way `DependencyInsight` does (depth 18 = 262,144 paths to one match):**

```
OLD: size=262,144 appends, time=676 ms, retained ≈ 224 MB
NEW: size=1 append,        time=0 ms,   retained ≈ 0 MB
```

Extrapolated to the issue's ~depth-25 scale (~33M paths): OLD ≈ 28 GB, NEW stays flat — matching the reported 37M nodes / 16 GB heap in order of magnitude. For non-pathological graphs the walk visits each node once, so there's no slowdown.

## Tests

- New `DependencyTreeWalkerTest` (network-free): a 20-deep diamond chain (2²⁰ ≈ 1M paths, 61 nodes) asserts the callback fires 61 times, with a fail-fast cap so a regression trips instead of OOMing; plus cycle-termination and path-order coverage.
- Gradle `DependencyInsightTest` expectations updated to the deduped values (`spring-boot` 10→3, `jackson-core` 11→2, one representative-path tree). Markers unchanged.

- Fixes #8181

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/graph/DependencyGraph.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/graph/DependencyGraph.java
@@ -20,57 +20,106 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.maven.tree.GroupArtifactVersion;
 import org.openrewrite.maven.tree.ResolvedDependency;
 import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
+import org.openrewrite.semver.DependencyMatcher;
 
 import java.util.*;
 
 import static java.util.Collections.emptySet;
 
+/**
+ * An inverse dependency graph rooted at a matched dependency: it shows, for one matched
+ * coordinate, every direct dependency it is reachable from and the paths in between.
+ * <p>
+ * Nodes are shared by coordinate, so a dependency reached through many paths (a diamond) is
+ * stored once — the graph is a DAG, not a per-path tree. This keeps construction and storage
+ * {@code O(nodes + edges)} even when the number of distinct root→match paths is exponential,
+ * while {@code getSize()} still reports that (potentially huge) path count via dynamic
+ * programming rather than by enumerating the paths.
+ */
 @RequiredArgsConstructor
 public class DependencyGraph {
-    @Nullable Node root;
+    private final @Nullable Node root;
 
     @Getter
-    int size = 0;
+    private final int size;
 
     /**
-     * Append a leaf-to-root dependency path to this dependency graph.
+     * Build one inverse dependency graph per matching dependency reachable from {@code roots}.
+     *
+     * @param configOrScope the scope (Maven) or configuration name (Gradle) the roots belong to
+     * @param roots         the direct dependencies to search
+     * @param matcher       the matcher identifying the dependencies of interest
+     * @return matched coordinate → its inverse dependency graph, in the order matches are first seen
      */
-    public void append(String configOrScope, Collection<ResolvedDependency> path) {
-        if (path.isEmpty()) {
+    public static Map<ResolvedGroupArtifactVersion, DependencyGraph> build(
+            String configOrScope, List<ResolvedDependency> roots, DependencyMatcher matcher) {
+        Map<String, Node> nodes = new HashMap<>();
+        Map<String, ResolvedGroupArtifactVersion> matched = new LinkedHashMap<>();
+        Set<GroupArtifactVersion> explored = new HashSet<>();
+        for (ResolvedDependency root : roots) {
+            Node rootNode = intern(nodes, root.getGav());
+            rootNode.getChildren().add(new ConfigurationNode(configOrScope));
+            collect(root, rootNode, nodes, matched, explored, matcher);
+        }
+
+        Map<ResolvedGroupArtifactVersion, DependencyGraph> graphs = new LinkedHashMap<>();
+        for (Map.Entry<String, ResolvedGroupArtifactVersion> match : matched.entrySet()) {
+            Node node = nodes.get(match.getKey());
+            long paths = countPaths(node, new HashMap<>(), new HashSet<>());
+            graphs.put(match.getValue(), new DependencyGraph(node, (int) Math.min(paths, Integer.MAX_VALUE)));
+        }
+        return graphs;
+    }
+
+    private static void collect(ResolvedDependency dependency, Node node,
+                                Map<String, Node> nodes, Map<String, ResolvedGroupArtifactVersion> matched,
+                                Set<GroupArtifactVersion> explored, DependencyMatcher matcher) {
+        if (matcher.matches(dependency.getGroupId(), dependency.getArtifactId(), dependency.getVersion())) {
+            matched.putIfAbsent(node.getId(), dependency.getGav());
+        }
+        // A single visited-set spanning the whole traversal: each subtree is walked once, and each
+        // parent→child edge is recorded once, regardless of how many paths reach the child.
+        if (!explored.add(dependency.getGav().asGroupArtifactVersion())) {
             return;
         }
-
-        Iterator<ResolvedDependency> iterator = path.iterator();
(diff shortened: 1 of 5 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/graph/DependencyGraph.java
```diff
@@ -20,57 +20,110 @@
...
+ */
 public class DependencyGraph {
-    @Nullable Node root;
+    private final @Nullable Node root;
 
     @Getter
-    int size = 0;
+    private final int size;
+
+    private DependencyGraph(@Nullable Node root, int size) {
```
jkschneider: Use `@RequiredArgsConstructor`

jkschneider: Done in 99513ada03 — both fields are `final` so it generates the same `(root, size)` constructor.

PR #8197, opened 2026-07-08
Scala: fix parsing of backticked special names as method invocations

greg-at-moderne:
## What's changed?

In Scala fixing parsing of backticked (escaped) identifiers in method invocations.
Using the `Quoted` existing marker.

## What's your motivation?

They backticks would be dropped without the fix.

### DIFF
--- a/rewrite-scala/src/main/scala/org/openrewrite/scala/internal/ScalaTreeVisitor.scala
+++ b/rewrite-scala/src/main/scala/org/openrewrite/scala/internal/ScalaTreeVisitor.scala
@@ -762,7 +762,7 @@ class ScalaTreeVisitor(
           case _ => return visitUnknown(postfixOp)
         }
 
-        val opName = postfixOp.op.name.toString
+        val opName = backtickAwareName(postfixOp.op.span, postfixOp.op.name.toString)
         val odEnd = Math.max(0, postfixOp.od.span.end - offsetAdjustment)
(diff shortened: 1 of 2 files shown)


rewrite-scala/src/main/scala/org/openrewrite/scala/internal/ScalaTreeVisitor.scala
```diff
@@ -8749,6 +8749,31 @@ class ScalaTreeVisitor(
...
+      val adjStart = Math.max(0, nameSpan.start - offsetAdjustment)
+      val adjEnd = Math.max(0, nameSpan.end - offsetAdjustment)
+      if (adjEnd > adjStart && adjEnd <= source.length) {
+        val text = source.substring(adjStart, adjEnd)
+        if (text.length >= 2 && text.charAt(0) == '`' && text.charAt(text.length - 1) == '`') {
+          return text
+        }
+        if (adjStart > 0 && adjEnd < source.length &&
+            source.charAt(adjStart - 1) == '`' && source.charAt(adjEnd) == '`') {
+          return "`" + text + "`"
```
jkschneider: I wonder if this might be better as a `Backtick` marker?

greg-at-moderne: We already have a marked called Quoted for the exact reason. Fixed the code to use it in: 5e19289

PR #8210, opened 2026-07-09
Keep `latest.patch` within the patch range for versions without a numeric minor

timtebeek:
`latest.patch` was escaping the patch range and picking higher minor versions whenever the current version had no numeric minor segment: `LatestPatch.buildTildeRange` relied on `Semver.minorVersion`, which returns the whole version string as a fallback, so `1.Final` built a `~1` range (any minor in major 1) and a bare `1` built a nonsensical `~1.1`. This fix pins the minor to `0` (`~major.0`) when the minor is absent or a non-numeric qualifier, so `latest.patch` stays patch-bounded, while numeric minors (`~major.minor`) and X-range wildcards like `2.x`/`2.+` (`~major`) keep their existing behavior. Added regression tests covering `1` and `1.Final`; the full semver suite and the Maven `UpgradeDependencyVersion` `latest.patch` tests remain green.

### DIFF
--- a/rewrite-core/src/test/java/org/openrewrite/semver/LatestPatchTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/semver/LatestPatchTest.java
@@ -63,6 +63,18 @@ class LatestPatchTest {
     }
 
     @Test
+    void xRangeWildcardMinorAllowsAnyMinorWithinTheMajor() {
+        // A wildcard in the minor position ("2.x"/"2.+") deliberately leaves the minor unspecified,
+        // so any minor within the major is eligible, but a higher major is not.
+        for (String current : List.of("2.x", "2.X", "2.+", "2.*")) {
+            assertThat(latestPatch.isValid(current, "2.1.0")).as(current).isTrue();
+            assertThat(latestPatch.isValid(current, "2.9.9")).as(current).isTrue();
+            assertThat(latestPatch.isValid(current, "3.0.0")).as(current).isFalse();
+            assertThat(latestPatch.upgrade(current, List.of("2.1.0", "2.9.9", "3.0.0"))).contains("2.9.9");
+        }
+    }
+
+    @Test
     void upgrade() {
         var upgrade = latestPatch.upgrade("2.10.10.3.24", List.of("2.10.0"));
         assertThat(upgrade).isEmpty();
--- a/rewrite-core/src/test/java/org/openrewrite/semver/LatestPatchTest.java
+++ b/rewrite-core/src/test/java/org/openrewrite/semver/LatestPatchTest.java
@@ -48,6 +48,25 @@ class LatestPatchTest {
     }
 
     @Test
+    void staysWithinPatchRangeWhenMinorIsAbsentOrNonNumeric() {
+        // Current versions with no numeric minor segment must not range across minor (or major)
+        // versions. Previously "1" built a "~1.1" range and "1.Final" built a "~1" range, both of
+        // which let "latest.patch" pick a higher minor version.
+
+        // Bare major: pin minor to 0, so only 1.0.x patches are eligible.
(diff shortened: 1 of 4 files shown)


timtebeek:
## Regression archaeology: when did `latest.patch` start escaping the patch range?

I traced the history of `LatestPatch.buildTildeRange` and the `Semver.minorVersion` fallback. The two symptoms this PR fixes have **different origins** — one is a genuine regression, the other a long-standing latent bug:

### `1.Final` (non-numeric minor) → over-ranges: regressed in #6587
The `~1` "any minor in major 1" behavior was introduced by **51306f1e70 — "LatestMinor should not include minor version in tilde range" (#6587), 2026-01-25**.

- **Before #6587**, the range was built unconditionally as `"~" + majorVersion + "." + minorVersion`. For `1.Final`, `minorVersion` returns the whole string (`"1.Final"`), producing `~1.1.Final`. The `TildeRange` pattern `~(\d+)(?:\.(\d+))?…` only accepts numeric segments, so `~1.1.Final` **fails to match and is rejected** — `isValid` returns `false`, no version is selected. Wrong, but *safe* (no silent minor bump).
- **#6587** added the `StringUtils.isNumeric(minor)` guard and, for the non-numeric case, fell back to `"~" + major`. That was needed to stop the X-range case (`2.+`) throwing a `ValidationException`, but as a side effect `1.Final` now builds `~1` — which matches and **allows any minor within major 1**. That's the escape-the-patch-range regression.

So the clear moment is **#6587**: the fix for `2.+` unintentionally converted `1.Final` from "rejected" to "silently over-ranges."

### bare `1` → `~1.1`: not a regression, present since #1095 (2021)
The nonsensical `~1.1` for a bare `1` predates #6587. It has existed since the feature was introduced in **b98a74594b — "Support latest.patch …" (#1095), 2021-10-10**, which always built `"~" + majorVersion + "." + minorVersion` and relied on `minorVersion("1")` returning `"1"` as its fallback. #6587 didn't touch this path (`"1"` is numeric, so it kept `~1.1`).

### Summary
| Input | Before #1095 | #1095 (2021) | #6587 (2026-01-25) | This PR |
|---|---|---|---|---|
| `1.Final` | n/a | `~1.1.Final` → rejected | **`~1` → any minor (regression)** | `~1.0` |
| `1` | n/a | `~1.1` (latent bug) | `~1.1` (unchanged) | `~1.0` |

This PR is the right fix for both: pinning the minor to `0` when it's absent or a non-numeric qualifier keeps `latest.patch` patch-bounded, while preserving numeric minors and the X-range wildcard (`2.x`/`2.+` → `~major`) behavior that #6587 was originally protecting.

rewrite-core/src/test/java/org/openrewrite/semver/LatestPatchTest.java
```diff
@@ -47,6 +47,31 @@ void noSnapshots() {
...
+        assertThat(latestPatch.isValid("1.Final", "2.0.0")).isFalse();
+        assertThat(latestPatch.upgrade("1.Final", List.of("1.0.5", "1.9.9", "2.0.0"))).contains("1.0.5");
+    }
+
+    @Test
+    void xRangeWildcardMinorAllowsAnyMinorWithinTheMajor() {
+        for (String current : List.of("2.x", "2.X", "2.+", "2.*")) {
+            assertThat(latestPatch.isValid(current, "2.1.0")).as(current).isTrue();
+            assertThat(latestPatch.isValid(current, "2.9.9")).as(current).isTrue();
+            assertThat(latestPatch.isValid(current, "3.0.0")).as(current).isFalse();
```
timtebeek, line 70: Not sure I'm thrilled about the `true` cases here, but I suppose when a recipe isn't specific it should match, even for `latest.patch`.

sambsnyd (approved):

PR #8220, opened 2026-07-09
Kotlin: shorten fully qualified references in `AddImport` (#8214)

timtebeek:
Fixes #8214. Kotlin's `AddImport` did not shorten fully qualified references to a newly imported type, so a recipe that replaced a reference via a fully qualified `JavaTemplate` (e.g. Jackson's `UseFormatAlignedObjectMappers`) left the reference fully qualified even though it added the import — whereas Java emits the simple name. This mirrors `org.openrewrite.java.AddImport`'s `ShortenFullyQualifiedTypeReferences` step so the reference now uses the simple name backed by the import. Adds a regression test in `AddImportTest`.

Issue #8214, opened 2026-07-09
Kotlin - Jackson recipe misbehavior - fully qualified type reference


nmck257:
See unit tests, with commentary from LLM:
```kotlin
import org.junit.jupiter.api.Test
import org.junitpioneer.jupiter.ExpectedToFail
import org.openrewrite.java.Assertions.java
import org.openrewrite.java.JavaParser
import org.openrewrite.kotlin.Assertions.kotlin
import org.openrewrite.kotlin.KotlinParser
import org.openrewrite.test.RecipeSpec
import org.openrewrite.test.RewriteTest

/*
Description: When UpgradeJackson_2_3's
  UseFormatAlignedObjectMappers converts a `new ObjectMapper()` into the format-aligned JsonMapper, the
  KOTLIN output emits the reference FULLY QUALIFIED - `tools.jackson.databind.json.JsonMapper()` (with a
  stray leading-space whitespace artifact) - even though it ALSO adds `import tools.jackson.databind.json.JsonMapper`.
  Java emits the correct simple name `new JsonMapper()`. Both languages also leave the now-unused
  `import tools.jackson.databind.ObjectMapper` behind.
  (The separate duplicate-import defect is covered in JacksonMapperConversionDuplicatesImportTest.)

Expected behavior: the converted mapper reference should use the simple name `JsonMapper()` backed by the
  added import (as Java does), not a redundant fully-qualified `tools.jackson.databind.json.JsonMapper()`.

Actual behavior: Kotlin emits the reference fully qualified while also adding the import (see
  conversionFullyQualifiesReference_kotlin); Java uses the simple name (see _java). The conversion of
  ObjectMapper -> JsonMapper itself is intended (JsonMapper is the format-aligned mapper in Jackson 3);
  only the Kotlin full-qualification (and the leftover unused ObjectMapper import) are defects.

Version first seen: moderne-recipe-bom 0.38.0 (rewrite-jackson 1.24.x/1.25.x range resolved locally),
- External references: Related upstream https://github.com/openrewrite/rewrite-jackson/issues/89
  ("ObjectMapper() -> tools.jackson.databind.json.JsonMapper()" full-qualification / wrong-type), partially
  - fixed in PR #109 (rewrite-jackson v1.20.0) for fluent chains but the bare-constructor case still reproduces
  on the version under test.
*/
class JacksonMapperConversionFullyQualifiesTypeTest : RewriteTest {

[...]

PR #8221, opened 2026-07-09
Kotlin: drop stray leading space when grafting a Java `new X()` template

timtebeek:
- Follow-up to #8214 (stacked on #8220).

### Problem

When a language-agnostic Java recipe uses `JavaTemplate.builder("new X()")` and applies it to a **Kotlin** source, the output has a stray leading space where Java's `new` keyword would go — e.g. `= Foo()` rendered as `=  Foo()` (two spaces after `=`):

```kotlin
val x =  Random()   // <- stray leading space
```

Kotlin has no `new` keyword, but the grafted `J.NewClass` still carries the space that followed `new` on the class prefix, and `KotlinPrinter.visitNewClass` drops the `new` text while still emitting that space.

- This reproduces independently of imports — in particular when the constructed type is **already imported**, so `AddImport` never runs. PR #8220 deliberately excluded this because its `AddImport`-scoped shortening only touched the case where a fully qualified reference was also being shortened.

### Fix

Normalize the `getNew()`/class-prefix space in the Kotlin `SpacesVisitor` (`visitNewClass`). This visitor runs during the auto-format step of template application — and because `JavaVisitor#autoFormat` resolves the `AutoFormatService` off the *source file*, applying a plain Java `JavaTemplate` to a Kotlin source runs Kotlin's auto-format, so this is the correct Kotlin-aware layer at the graft boundary (rather than the printer, which merely renders what's in the tree).

A natively parsed Kotlin constructor call has `getNew() == Space.EMPTY` and an empty class prefix; the fix collapses the grafted Java tree to match. The `object : Super()` (`KObject`) and enclosing-type (`TypeReferencePrefix`) forms legitimately rely on that space and are left untouched, as are prefixes carrying comments or line breaks.

- ### Relationship to #8220

- This composes with #8220's `AddImport` shortening: the space is collapsed at graft time, so when #8220 later shortens a freshly imported fully qualified reference down to its simple name, the result is clean (`Random()`, not `tools.jackson...JsonMapper()` and not `=  JsonMapper()`).

### Tests

`NewClassTemplateTest` covers:
- already-imported type (`AddImport` not involved)
- freshly-imported type (composes with #8220's shortening)
- constructor with type arguments (`new ArrayList<String>()` → `J.ParameterizedType` clazz)
- constructor in argument position (`println(new Random())`)

Full `:rewrite-kotlin:test` suite is green.

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
@@ -1097,6 +1097,17 @@ public class SpacesVisitor<P> extends KotlinIsoVisitor<P> {
     @Override
     public J.NewClass visitNewClass(J.NewClass newClass, P p) {
         J.NewClass nc = super.visitNewClass(newClass, p);
+        // Kotlin has no `new` keyword; drop the space a grafted Java `new X()` leaves before the type.
+        if (!nc.getMarkers().findFirst(KObject.class).isPresent() &&
+            !nc.getMarkers().findFirst(TypeReferencePrefix.class).isPresent() &&
+            nc.getPadding().getEnclosing() == null) {
+            if (isCollapsibleSpace(nc.getNew())) {
+                nc = nc.withNew(Space.EMPTY);
+            }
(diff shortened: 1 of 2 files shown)


MBoegers (approved):
LGTM

PR #8228, opened 2026-07-10
Document HasMinimumJavaVersion behavior across version selector forms

steve-aom-elliott:
## Summary
- Adds tests to `HasMinimumJavaVersionTest` documenting how each `version` selector form (bare integer, `X` range, set range) actually maps to `JavaVersion` marker matches.
- The bare-integer and `X`-range forms both silently fail to match when the repo declares a higher Java version than the argument, which is surprising given the recipe's name. Downstream recipes using this as a precondition are silently skipped in that case.
- These tests only document current behavior; a follow-up can decide whether to change `validate()` to treat a bare integer as `[N,)`.

## Test plan
- [x] `./gradlew :rewrite-java-test:test --tests HasMinimumJavaVersionTest`

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/search/HasMinimumJavaVersionTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/search/HasMinimumJavaVersionTest.java
@@ -70,4 +70,125 @@ class HasMinimumJavaVersionTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void bareIntegerMatchesExactVersionOnly() {
+        rewriteRun(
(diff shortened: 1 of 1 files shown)


steve-aom-elliott:
@timtebeek Just adding you for visibility, as this was hit as part of JUnit 5 -> 6 migration recipe. Only adding more documenting tests.

timtebeek:
Thanks for the tag! I'm a bit surprised we don't actually match on versions higher than the given minimum already.

I'd expect `HasMinimumJavaVersion` with value `17` match anything `17+`.

It looks like what we actually have at the moment would be `HasExactJavaVersion`, which I highly doubt was the intent.

PR #8236, opened 2026-07-10
Do not apply Java typecast spacing to Kotlin `as` casts

vlsi:
### What's changed

`SpacesVisitor#visitTypeCast` now skips type casts that are not inside a `J.CompilationUnit`, so the Java typecast spacing rules no longer run on Kotlin `as` casts. Adds a `rewrite-kotlin` test covering `as` and `as?`.

### Why

Kotlin reuses `J.TypeCast` to model `expr as Type`, which has no parentheses. `visitTypeCast` assumes the Java `(Type) expr` layout — it sets the expression's prefix (the space after `)`) and applies the within-parentheses spacing to the type. On a Kotlin cast the expression is printed *before* `as`, so those edits move and drop whitespace:

```
val b = list.get(0) as List<*>   ->   val b =  list.get(0) asList<*>
```

The mangled output no longer compiles. Any Java spacing recipe run over a mixed Java/Kotlin project triggers it — this surfaced with `TypecastParenPad` pulled in by a `CodeCleanup` composite on Apache JMeter's mixed sources. The fix mirrors `AutoFormatVisitor`, which already restricts itself to `J.CompilationUnit`.

### How to verify

- `./gradlew :rewrite-kotlin:test --tests "*.TypecastParenPadKotlinTest"` — fails before the change, passes after.
- `./gradlew :rewrite-java-test:test --tests "*.SpacesTest" --tests "*.TypecastParenPadTest"` — Java behaviour is unchanged (204 tests still pass).


timtebeek:
hi! Thanks for the various reports and PRs in progress! Good to see what you're running up against, and see what we can do to get these resolved. In this specific case the fix might be a little more involved; we actually have language specific SpacesVisitor formatters that ought to be used. But `TypecastParenPad` likely predates that pattern. Let's digg a little deeper and not change the Java formatter to fix Kotlin.

timtebeek (approved):
Thanks afain for kicking this off! I've changed your approach a bit such that I'd like to get a second review before we merge, but nice to be consistent going forward even where the same construct looks quite different.

greg-at-moderne (approved):
LGTM. Thanks for the contribution!

Issue #8237, opened 2026-07-10
ShortenFullyQualifiedTypeReferences drops the qualifier and import for nested types on Kotlin


vlsi:
### What version of OpenRewrite are you using?

rewrite 8.86 / rewrite-gradle-plugin 7.35, reproduced against `main`.

### How are you running OpenRewrite?

`ShortenFullyQualifiedTypeReferences` (via a `CodeCleanup` composite) applied to a Kotlin source in a mixed Java/Kotlin module.

### What is the smallest, simplest way to reproduce the problem?

```kotlin
import org.apiguardian.api.API

@API(status = API.Status.EXPERIMENTAL, since = "5.6")
class Foo
```

`ShortenFullyQualifiedTypeReferences` rewrites the nested enum access `API.Status.EXPERIMENTAL` to `Status.EXPERIMENTAL` without adding an import for `Status`:

```kotlin
import org.apiguardian.api.API

@API(status = Status.EXPERIMENTAL, since = "5.6")
class Foo
```

The result no longer compiles (`unresolved reference: Status`).

A test that reproduces it is in the accompanying PR.

### What did you expect to see?

The reference is left unchanged — it already uses the simple, imported outer name `API`. This is what the recipe does on the equivalent Java source.

### What happened instead?

The qualifier `API.` is dropped and no import is added, producing uncompilable Kotlin.

### Root cause

The recipe guard is correct. `visitFieldAccess` only shortens a reference whose type is top-level: `type instanceof JavaType.Class && ((JavaType.Class) type).getOwningClass() == null`. For the nested enum `org.apiguardian.api.API.Status` the owning class is `API`, so a correctly attributed reference is skipped, as it is on Java.

The bug is in Kotlin type attribution: the nested reference `API.Status` was given the *outer* type `org.apiguardian.api.API` instead of the nested type `org.apiguardian.api.API$Status`. That outer type is genuinely top-level, so its `owningClass` is null, the guard passes, and the recipe shortens `API.Status` to `Status`. The follow-up `maybeAddImport` call then receives the outer name `org.apiguardian.api.API` and does nothing, because `API` is already imported, so no new import appears. (The earlier "`maybeAddImport` is a no-op for nested types" reading was the same wrong type surfacing again, not a second gap.)

[...]

PR #8238, opened 2026-07-10
Kotlin: fix type of nested class reached via imported outer type

vlsi:
### Why

On Kotlin sources, `ShortenFullyQualifiedTypeReferences` corrupts a nested type reached through an imported outer type. Given:

```kotlin
import org.apiguardian.api.API

@API(status = API.Status.EXPERIMENTAL, since = "5.6")
class Foo
```

it rewrites `API.Status.EXPERIMENTAL` to `Status.EXPERIMENTAL` without adding an import, so `Status` no longer resolves and the file fails to compile. The equivalent Java source is left untouched. This surfaced running `CodeCleanup` over Apache JMeter's mixed Java/Kotlin sources — https://github.com/openrewrite/rewrite/issues/8237.

### What

The root cause is in Kotlin type attribution, not the recipe. `PsiElementAssociations.matchClassId0` matched a reference's source text only against the fully-qualified class name (`org.apiguardian.api.API.Status`). The import-shortened form `API.Status` never matched, so the lookup walked up the `outerClassId` chain and returned the outer class `org.apiguardian.api.API`. The nested reference therefore carried the outer type, whose `owningClass` is null — which is exactly the guard `ShortenFullyQualifiedTypeReferences` uses to decide a reference is top-level and safe to shorten.

`matchClassId0` now resolves every way a nested type can be written to its own class id:

- fully qualified (`foo.bar.A.B`) and relative to an imported outer (`A.B`) match by text;
- an import alias (`import java.util.Locale as Loc`, then `Loc.Category`) matches by leaf name plus nesting depth, since the aliased root matches neither the fully-qualified nor the package-relative name;
- an imported nested type (`import foo.bar.A.B`, then `B.A.C`) matches a trailing run of the nested names.

The two multi-segment fallbacks require at least two segments, so a single receiver name whose leaf repeats deeper (the outer `A` of `A.B.A.C`) is not matched too deep.

- Broaden `matchClassId0` to the alias and imported-nested forms.
- Add type-mapping tests next to the existing `nestedFieldAccessType` / `nestedFieldAccessWithPackage` cases: import-shortened at one and multiple levels, same simple name across different outers, and an imported nested class.
- Add recipe-level reproduction tests from #8237 for the imported-outer and aliased-outer forms.
- Reproduce with `java.util.Locale` / `Locale.Category` rather than the reported `org.apiguardian.api.API` / `API.Status`, so the tests need no new dependency. The package matters: `ShortenFullyQualifiedTypeReferences` handles `java.lang.*` in a separate branch, so a nested type from there would survive for a reason unrelated to this fix.
- Closes http://github.com/openrewrite/rewrite/issues/8237

### How to verify

`./gradlew :rewrite-kotlin:test` — the full module passes, including the new `ShortenFullyQualifiedTypeReferencesKotlinTest` cases and the `KotlinTypeMappingTest` `nestedFieldAccess*` / `nestedTypesWithSameSimpleNameFromDifferentOuters` cases.

### DIFF
--- a/rewrite-kotlin/build.gradle.kts
+++ b/rewrite-kotlin/build.gradle.kts
@@ -54,6 +54,7 @@ dependencies {
     implementation(kotlin("stdlib", kotlinVersion))
 
     testImplementation("org.junit-pioneer:junit-pioneer:latest.release")
+    testImplementation("org.apiguardian:apiguardian-api:1.1.2")
     testImplementation(project(":rewrite-test"))
     testRuntimeOnly(project(":rewrite-java-21"))
     testRuntimeOnly("org.antlr:antlr4-runtime:4.13.2")
--- a/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/ShortenFullyQualifiedTypeReferencesKotlinTest.java
+++ b/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/ShortenFullyQualifiedTypeReferencesKotlinTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2026 the original author or authors.
  * <p>
  * Licensed under the Moderne Source Available License (the "License");
  * you may not use this file except in compliance with the License.
--- a/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/ShortenFullyQualifiedTypeReferencesKotlinTest.java
+++ b/rewrite-kotlin/src/test/java/org/openrewrite/kotlin/ShortenFullyQualifiedTypeReferencesKotlinTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.kotlin;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.Issue;
+import org.openrewrite.java.ShortenFullyQualifiedTypeReferences;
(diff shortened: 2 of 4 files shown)


vlsi:
I just saw the build produced 27635 lines. That is huge. Have you considered reducing the output? I see you have `--info` which sounds strange to me.
At this scale I do not think build log is helpful. How do you identify the reasons for the build failure?

rewrite-kotlin/build.gradle.kts
```diff
@@ -54,6 +54,7 @@ dependencies {
     implementation(kotlin("stdlib", kotlinVersion))
 
     testImplementation("org.junit-pioneer:junit-pioneer:latest.release")
+    testImplementation("org.apiguardian:apiguardian-api:1.1.2")
```
greg-at-moderne: I suggest we rewrite the test cases to be generic, i.e. possibly use some standard library references or existing dependencies. This way we can refrain from adding this (test) dependency.

vlsi: Done, the dependency is gone. The tests now use `java.util.Locale` / `Locale.Category`, which has the same shape as `API.Status`: a nested type reached through an imported outer type. The `classpath("apiguardian-api")` calls went away too, since JDK types are already visible to the parser.

One detail worth flagging if you reuse the trick elsewhere: it has to be `java.util` rather than `java.lang`. `ShortenFullyQualifiedTypeReferences` handles `java.lang.*` in a [separate branch](https://github.com/openrewrite/rewrite/blob/04a1ec1c510222dc9faae4a4dc465249f43305fa/rewrite-java/src/main/java/org/openrewrite/java/ShortenFullyQualifiedTypeReferences.java#L178), so `Thread.State` or `Character.UnicodeBlock` would be left alone for a reason unrelated to this fix and the test would pass with or without it.

Checked both directions. With the fix `:rewrite-kotlin:test` is green for the touched classes; reverting `matchClassId0` fails all six new tests, the two rewritten ones included, with exactly the reported corruption:

```
expected: "val category = Locale.Category.DISPLAY"
 but was: "val category = Category.DISPLAY"
```

rewrite-kotlin/src/test/java/org/openrewrite/kotlin/ShortenFullyQualifiedTypeReferencesKotlinTest.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2024 the original author or authors.
```
greg-at-moderne:
```suggestion
 * Copyright 2026 the original author or authors.
```

greg-at-moderne (approved):

greg-at-moderne:
Thanks for another valuable contribution!

PR #8242, opened 2026-07-12
Quiet polyglot test-runner build output by default

timtebeek:
The `rewrite-go`, `rewrite-python`, `rewrite-csharp`, and `rewrite-javascript` modules ran their native test runners in verbose mode, printing one or more lines per passing test and accounting for ~7,000 of ~7,500 content lines in a full build's stdout. This switches each runner to a concise reporter (gotestsum `pkgname`, pytest non-verbose, `dotnet --verbosity minimal`, vitest `dot`) and gates the JVM-side per-test event logging in `rewrite-python`/`rewrite-csharp` behind a new `-PverboseTests` flag. Failures and the JUnit XML reports are unchanged, so nothing is lost for diagnosis; pass `-PverboseTests` (or the `VERBOSE_TESTS` env var for vitest) to restore the full per-test output. Verified with `--rerun-tasks` across all four modules: BUILD SUCCESSFUL with the per-test noise eliminated.

### DIFF
--- a/rewrite-csharp/build.gradle.kts
+++ b/rewrite-csharp/build.gradle.kts
@@ -145,9 +145,11 @@ val csharpTest by tasks.registering(Exec::class) {
         // JunitXml.TestLogger resolves a relative LogFilePath against the test project's
         // TargetDir, so pin the absolute path and pass --results-directory to match.
         junitXmlFile.parentFile.mkdirs()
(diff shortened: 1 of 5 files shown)


greg-at-moderne (approved):
Great idea.
Let's see in practice if we lose anything. I bet we don't.

PR #8250, opened 2026-07-14
Add DockerCopyFrom trait for `COPY --from` / `ADD --from` image references

timtebeek:
Extends OpenRewrite's image-reference trait coverage beyond `FROM` to the `--from` flag of `COPY` and `ADD`, which `DockerFrom.Matcher` previously ignored. The shared `name:tag@digest` splitting logic is extracted from `DockerParserVisitor` into a new `ImageReferences.split` helper (a behavior-preserving refactor) and reused by the new `DockerCopyFrom` trait, which decomposes external image references and distinguishes them from build-stage references (by name or numeric index) via the enclosing `Docker.File` stages. The trait adds `withImageReference`/`withTag` edit helpers so recipes can upgrade or pin versions, and `DockerCopyFrom.Matcher` mirrors `DockerFrom.Matcher` (`imageName`/`tag`/`digest`) plus `excludeStageReferences()`. Tests cover matching (external image, stage-by-name, stage-by-index, bare non-stage name, digest, env-var value) and version-upgrade rewrites across both `COPY` and `ADD`; the full `:rewrite-docker:test` suite stays green.


rewrite-docker/src/main/java/org/openrewrite/docker/internal/ImageReferences.java
```diff
@@ -0,0 +1,139 @@
...
+                new Docker.Argument(randomId(), Space.EMPTY, Markers.EMPTY, digestContents);
+
+        return new Docker.@Nullable Argument[]{imageName, tag, digest};
+    }
+
+    /**
+     * Returns the index of the {@code :} that separates the tag, or -1 if there is none. Only a
+     * colon after the last {@code /} can be a tag separator; an earlier colon is a registry port
+     * (e.g. {@code registry.example.com:5000/app}), which is part of the image name.
+     */
```
MBoegers:
```suggestion
```

rewrite-docker/src/main/java/org/openrewrite/docker/trait/DockerCopyFrom.java
```diff
@@ -0,0 +1,414 @@
...
+            return null;
+        }
+        return ImageReferences.split(arg.getContents(), arg.getPrefix());
+    }
+
+    /**
+     * Returns the raw {@code --from} value with environment variable references preserved,
+     * or null if there is no {@code --from} flag.
+     */
+    public @Nullable String getFromValue() {
```
MBoegers: Shouln't this be an `Optional`? I think we use them in the Trait API.

MBoegers: same for `getImageName`, `getTag`, `getUnpinnedReason` and `getDigest`

MBoegers: It looks like I used `null` in the initial implementation. Still open for discussion here.

MBoegers (commented):
Besided the `null`vs. `Optional` discussion LGTM

PR #8256, opened 2026-07-14
Python: rewrite qualified attribute-access call sites in ChangeImport

timtebeek:
Extends the Python `ChangeImport` recipe so a single invocation also rewrites qualified references reached through a module import (e.g. `avro.schema.Parse(x)` → `avro.schema.parse(x)`), where previously only the `from`-import form and its bare-name usages were handled. It now covers the dotted-select, `import ... as` alias, and `from pkg import submod` forms by matching the module reference in `visit_method_invocation`/`visit_field_access` and rewriting only the leaf name when the module path is unchanged (preserving the select, alias, and whitespace, and leaving imports untouched). The module-change path (e.g. `fractions.gcd` → `math.gcd`) is unchanged, and five tests were added for the new forms plus negative and idempotency cases.

Once this lands, downstream qualified-form workarounds in `rewrite-migrate-python` become redundant and can be deleted — `ReplaceAvroSchemaParse` and the qualified-form halves of `ReplaceCgiParseQs`/`ReplaceCgiParseQsl`.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/python/recipes/change_import.py
+++ b/rewrite-python/rewrite/src/rewrite/python/recipes/change_import.py
@@ -170,13 +170,23 @@ class ChangeImport(Recipe):
                             if alias is not None:
                                 self.has_old_import = True
                                 self.old_alias = alias if alias != "" else None
-                        if old_name and not self.has_direct_module_import and stmt.from_ is None:
-                            for imp in stmt.names:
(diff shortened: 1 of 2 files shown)


timtebeek:
`ChangeType` ought to cover this already; weird how that didn't pop up; possibly because we also and separately expose that `ChangeImport`, whereas folks should prefer `ChangeType`.

PR #8261, opened 2026-07-15
ChangeValue cannot update values inside TOML inline tables

### LINKED ISSUE #8260: ChangeValue cannot update values inside TOML inline tables

- OpenRewrite: 8.87.1

I am using the Gradle plugin in a single-module project.

Given this TOML:

testcontainers-mongo = { group = "org.testcontainers", name = "mongodb" }

Configure ChangeValue with the key path testcontainers-mongo.name.

testcontainers-mongo = { group = "org.testcontainers", name = "testcontainers-mongodb" }

## What did you see instead?

The file remains unchanged:

testcontainers-mongo = { group = "org.testcontainers", name = "mongodb" }

No error or stack trace is produced.

## Are you interested in contributing a fix to OpenRewrite?

If needed (https://github.com/openrewrite/rewrite/pull/8261/changes)

KamilPatora:
## What's changed?

Updated TomlPathMatcher to include inline-table key paths when matching TOML values. Added a regression test covering ChangeValue updates inside inline tables.

## What's your motivation?

ChangeValue could not update a value in an inline table when using a dotted path such as testcontainers-mongo.name. The recipe completed without errors, but the value remained unchanged.

Please review the path construction in TomlPathMatcher and confirm that matching for regular TOML tables remains unchanged.

The only workaround was to edit the inline-table value manually. No alternative recipe-based workaround was available.

## Any additional context

The regression test verifies that the nested value is changed while preserving the inline-table structure and other values.

### Checklist

### DIFF
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/TomlPathMatcher.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/TomlPathMatcher.java
@@ -97,26 +97,17 @@ public class TomlPathMatcher {
                 Toml.KeyValue kv = (Toml.KeyValue) value;
                 TomlKey key = kv.getKey();
                 if (key instanceof Toml.Identifier) {
-                    String keyName = ((Toml.Identifier) key).getName();
-                    Cursor parent = current.getParent();
-                    while (parent != null) {
-                        Object parentValue = parent.getValue();
-                        if (parentValue instanceof Toml.Table) {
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks a lot for flagging and fixing this @KamilPatora ! I'll push up a small change to add some additional test coverage, but other than that I think this looks really well.

PR #8269, opened 2026-07-16
Kotlin recipe DSL: preserve scan/generate phases in K2 synthesis

timtebeek:
The rewrite-kotlin K2 compiler plugin synthesized every imperative `recipe(...)` as a plain `Recipe` overriding only `getVisitor()`, so `scan { }.generate { }` and bare `generate { }` recipes silently emitted nothing and `scan { }.edit { }` ran its edit against an empty (never-scanned) accumulator. Recipes that declare a scan/generate phase are now synthesized as a field-less `ScanningRecipe<Any>` subclass whose four lifecycle methods delegate to new `buildImperative*` runtime helpers, preserving every phase while staying Jackson-roundtrippable. Detection unwraps the `IMPLICIT_COERCION_TO_UNIT` that K2 wraps around a `scan { }.generate { }` statement (whose `Scan.generate` returns non-Unit), which is why only that shape slipped through; edit-only recipes keep the lighter `Recipe` + `getVisitor()` path. Adds `RecipePluginScanningTest` covering scan+generate, scan+edit, bare generate, and the synthesized-class shape.

### DIFF
--- a/rewrite-kotlin/src/main/kotlin/org/openrewrite/RecipeDsl.kt
+++ b/rewrite-kotlin/src/main/kotlin/org/openrewrite/RecipeDsl.kt
@@ -713,3 +713,61 @@ public fun buildImperativeVisitor(
     builder.block()
     return builder.build("", "", emptySet(), null).getVisitor()
 }
+
+private fun buildImperativeRecipe(block: RecipeBuilder.() -> Unit): Recipe {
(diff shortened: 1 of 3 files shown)


timtebeek:
Closing for now as I don't immediately have time to wrap this up; continued in
- https://github.com/openrewrite/rewrite/issues/8657

PR #8273, opened 2026-07-16
Kotlin recipe DSL: attach method type on multi-param selector rewrites

timtebeek:
The `rewrite { } to { }` after-template rendered placeholder types from `IrType.classFqName` only, which drops generic type arguments — so a raw `kotlin.Function1` placeholder was emitted for lambda/selector parameters. Callees whose overload resolution depends on that generic argument (e.g. `Iterable<T>.sumOf`, which dispatches on the selector's return type) then failed to resolve, leaving the replacement `J.MethodInvocation` with a null `JavaType.Method` that fails `assertValidTypes` under default `TypeValidation`. This spells out concrete type arguments on the KotlinTemplate path (with `? extends`/`? super` for variance), falling back to the raw spelling for non-concrete arguments so no already-resolving case regresses. Adds a `sumBy → sumOf` `RewriteTest` that fails before and passes after with full type validation.

### DIFF
--- a/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeIrGenerationExtension.kt
+++ b/rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeIrGenerationExtension.kt
@@ -66,7 +66,10 @@ import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.symbols.IrConstructorSymbol
 import org.jetbrains.kotlin.ir.symbols.IrSimpleFunctionSymbol
 import org.jetbrains.kotlin.ir.symbols.IrValueSymbol
+import org.jetbrains.kotlin.ir.types.IrSimpleType
 import org.jetbrains.kotlin.ir.types.IrType
+import org.jetbrains.kotlin.ir.types.IrTypeArgument
+import org.jetbrains.kotlin.ir.types.IrTypeProjection
 import org.jetbrains.kotlin.ir.types.classFqName
 import org.jetbrains.kotlin.ir.types.defaultType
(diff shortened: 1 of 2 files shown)


rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeIrGenerationExtension.kt
```diff
@@ -1620,9 +1624,27 @@ internal class RecipeIrGenerationExtension : IrGenerationExtension {
...
+        val rendered = args.map { renderTypeArgument(it) ?: return fqn }
+        return "$fqn<${rendered.joinToString(", ")}>"
+    }
+
+    private fun renderTypeArgument(arg: IrTypeArgument): String? {
+        if (arg !is IrTypeProjection) return null
+        val inner = renderPlaceholderType(arg.type, javaTemplate = false) ?: return null
+        return when (arg.variance) {
+            Variance.OUT_VARIANCE -> "? extends $inner"
+            Variance.IN_VARIANCE -> "? super $inner"
```
greg-at-moderne: I don't have a ready example, but my (human) gut feeling tells me this might not be taking into account all cases.

greg-at-moderne (approved):
The comment I left is definitely not a blocker.
This PR is an improvement. Probably some other corner cases will show up in the future. We can deal with them later.

rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/RecipeIrGenerationExtension.kt
timtebeek: Thanks! Minimized just now to what's covered by tests.

PR #4394, opened 2024-08-07
Add GradleDependency trait

shanman190:
## What's changed?
Added an implementation of the `Trait` api for Gradle dependencies

## What's your motivation?
Having the same `Trait` api available for Gradle as was introduced for Maven.

## Anything in particular you'd like reviewers to focus on?
Nothing in particular

## Anyone you would like to review specifically?
Anyone

## Have you considered any alternatives or workarounds?
None

## Any additional context
* https://github.com/openrewrite/rewrite/commit/c24e9920b509a79a9189402095b905f6ccbd1cd4

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #8274, opened 2026-07-16
4400 - gradle toml version catalogs

KamilPatora:
# Summary

Adds Gradle Version Catalog support to the following recipes:

- `ChangeDependency`
- `UpgradeDependencyVersion`
- `UpgradePluginVersion`

The recipes can now recognize and update dependencies and plugins declared in the conventional `gradle/libs.versions.toml` catalog in addition to declarations in Gradle build scripts.

# Implementation

This PR introduces semantic traits for Gradle Version Catalogs instead of embedding TOML-specific logic into individual recipes.

New traits include:

- `GradleVersionCatalog`
- `GradleVersionCatalogDependency`
- `GradleVersionCatalogPlugin`

These traits model version catalog entries as dependencies and plugins, allowing existing Gradle recipes to operate on them using the same high-level concepts they already use for Gradle source files.

The implementation supports:

- Direct versions (`version = "..."`)
- Shared version references (`version.ref`)
- Inline table notation
- Compact string notation (`group:artifact[:version]`)
- `module = "group:artifact"` notation
- Plugin declarations

Shared `version.ref` entries are updated conservatively. A referenced version is modified only when every consumer of that reference is supported by the recipe and resolves to the same replacement version, preventing unintended changes to unrelated catalog entries.

# Supported Scope

This PR supports the conventional Gradle version catalog located at `gradle/libs.versions.toml`.

Supported library declarations include:

- `group` / `name` notation
- `module` notation
- Compact string notation
- Direct versions
- `version.ref`

Supported plugin declarations include:

- Direct versions
- `version.ref`

# Out of Scope

The following are intentionally not supported in this PR:

- Rich version declarations (for example `strictly`, `prefer`, `reject`, and `require`)
- Custom or named version catalogs
- Bundle mutation
- `GradleProject` model synchronization after catalog-only changes

When an unsupported version declaration is encountered during a requested version change, the catalog entry is left unchanged rather than partially updating only its coordinates.

# Tests

The test suite covers:

- Direct version updates
- `version.ref` updates
- Shared `version.ref` coordination
- Inline table notation
- Compact string notation
- `module` notation
- Plugin version updates
- Coordinate relocation
- `overrideManagedVersion` behavior
- Unsupported version declarations remaining unchanged
- Gradle Kotlin DSL projects using version catalogs


timtebeek:
Thanks for kicking this off @KamilPatora ! I'd have perhaps expected more of these changes to land in the Gradle dependency trait we added a while back, but I'm not sure how far we've gotten with adoption there:
- https://github.com/openrewrite/rewrite/pull/4394

My aim with the trait would be to limit which recipes even need to know about Toml specifically; ideally that's all handled  by the trait and recipes just concern themselves with the question of bumping a version based on what's encountered. Perhaps @shanman190 has additional thoughts here as well. It's definitely a larger topic, but worth exploring to reduce complexity across recipes.

KamilPatora:
Thanks for the feedback! I’ve only been working with TOML files for about a month, so I’m still getting a feel for how this should fit into the existing GradleDependency trait. Any ideas or guidance would be super welcome.

For now, I kept the TOML stuff in separate catalog traits since it’s a different AST model from build scripts.
Does that seem like the right direction @timtebeek?

timtebeek:
Apologies for the slow response here! In the case of Gradle and Toml I don't have all the answers, and with other work I've not found the time to dive in yet.

KamilPatora:
Hey @timtebeek. Done some improvements here and there. I think this looks better than before... lol

KamilPatora:
Fyi for reviewers. Im gonna split up some of this code to another PR, 3k added lines is too much imo

timtebeek:
That's much appreciated! It's at times a little overwhelming to start a review on a large PR, but appreciate the work you're doing, also in splitting off smaller increments like https://github.com/openrewrite/rewrite/pull/8339 already.

KamilPatora:
Moved creating traits here: https://github.com/openrewrite/rewrite/pull/8376
Still a lot of lines, but most of them are tests really
@timtebeek

PR #8277, opened 2026-07-17
8276 - Incorrect AST values for multiline strings

KamilPatora:
## What's changed?

Fixed multiline TOML string values in `rewrite-toml`.

Triple-quoted strings now correctly remove all three delimiter characters when building the AST value. This applies both when parsing TOML and when using the `ChangeValue` recipe.

## What's your motivation?

previously the rendered TOML was correct, but `Toml.Literal.getValue()` incorrectly included two quote characters at both ends of multiline strings.

not really
## Any additional context

The triple-quote checks must run before the single-quote checks because triple-quoted strings also start with a single quote character.

### DIFF
--- a/rewrite-toml/src/main/java/org/openrewrite/toml/ChangeValue.java
+++ b/rewrite-toml/src/main/java/org/openrewrite/toml/ChangeValue.java
@@ -96,19 +96,6 @@ public class ChangeValue extends Recipe {
                     );
                 }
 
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks a lot for yet another fix to the core TOML support @KamilPatora ! I've added some slight polish and will merge as soon as the tests pass. Great stuff!

Issue #8293, opened 2026-07-20
ChangeType throws NullPointerException on a ClassDeclaration with null type (non-Java / Python sources)


mtthwcmpbll:
### What version of OpenRewrite are you using?

Latest `rewrite-java` (reproduced on `main`).

### How are you running OpenRewrite?

Via the Moderne platform, running `org.openrewrite.java.ChangeType` against Python sources (rewrite-python LSTs) — e.g. the PEP 585 `typing.List` → `builtins.list` renames.

### What is the smallest, simplest way to reproduce the problem?

Run `ChangeType` over a source file that contains a class declaration whose `getType()` is `null`. This occurs with non-Java LSTs (e.g. rewrite-python) where an unattributed class — such as a `@dataclass` — has no resolved `JavaType.FullyQualified`.

`ChangeType.ChangeClassDefinition.visit(Tree, ExecutionContext)` iterates every class in the compilation unit and unconditionally dereferences the type:

```java
for (J.ClassDeclaration declaration : cu.getClasses()) {
    // Check the class name instead of source path, as it might differ
    String fqn = declaration.getType().getFullyQualifiedName(); // NPE when getType() == null
    if (fqn.equals(originalType.getFullyQualifiedName())) {
```

Because this runs on *every* class in any file the recipe touches — regardless of whether the class relates to the type being renamed — a single unattributed class aborts the whole source file's run.

### What did you expect to see?

`ChangeType` skips class declarations with no attributed type and completes without error.

### What happened instead?

```
java.lang.NullPointerException: Cannot invoke "org.openrewrite.java.tree.JavaType$FullyQualified.getFullyQualifiedName()" because the return value of "org.openrewrite.java.tree.J$ClassDeclaration.getType()" is null
	at org.openrewrite.java.ChangeType$ChangeClassDefinition.visit(ChangeType.java:752)
	at org.openrewrite.java.ChangeType$ChangeClassDefinition.visit(ChangeType.java:735)
	at org.openrewrite.TreeVisitor.visitNonNull(TreeVisitor.java:167)
	at org.openrewrite.java.ChangeType$JavaChangeTypeVisitor.visit(ChangeType.java:170)
	...
```

### Proposed fix

Guard the null type and `continue` — a class with no attributed FQN can't be the type being renamed. (The sibling dereference in `visitClassDeclaration` is already null-safe via `TypeUtils.isOfClassType(...)`.)

```java
JavaType.FullyQualified declarationType = declaration.getType();
[...]

PR #8295, opened 2026-07-20
Python: attribute types for decorated class/function declarations

timtebeek:
- Decorated Python declarations (e.g. `@dataclass` classes, decorated functions) were losing their attributed type because rewrite-python looks up ty's types by byte offset, and CPython's `ast` places a decorated declaration's position at the `class`/`def` keyword while ruff (ty's backend) starts the node range at the first decorator's `@`, so the lookup missed. This left `J.ClassDeclaration.getType()` null for `@dataclass` classes, which surfaced as a NullPointerException in `ChangeType` (#8293). The fix realigns `_lookup_type_id` to the first decorator's `@` when a declaration is decorated, restoring types for single, stacked, and whitespaced decorators as well as decorated functions. Adds `TestDecoratedDeclarationType` regression tests, verified to fail without the fix and pass with it.

### DIFF
--- a/rewrite-python/rewrite/tests/python/test_type_attribution.py
+++ b/rewrite-python/rewrite/tests/python/test_type_attribution.py
@@ -1481,6 +1481,94 @@ x = Multi()
             _cleanup_mapping(mapping, tmpdir, client)
 
 
+@requires_ty_types_cli
+class TestDecoratedDeclarationType:
+    """Regression tests for type attribution of *decorated* class/function declarations.
+
+    CPython's ``ast`` places a decorated class/function's ``lineno``/``col_offset``
+    at the ``class``/``def`` keyword, while ruff (ty's backend) starts the
+    statement's range at the first decorator's ``@``. The byte-offset lookup must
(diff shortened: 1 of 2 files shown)


rewrite-python/rewrite/tests/python/test_type_attribution.py
```diff
@@ -1481,6 +1481,94 @@ class Multi(Base, Mixin):
...
+
+    def test_custom_decorated_class_declaration_has_type(self):
+        source = '''def deco(c):
+    return c
+
+
+@deco
+class Foo:
+    pass
+'''
```
sambsnyd, line 1530: The strings in the text blocks breaking with the alignment of the surrounding code makes my eye twitch.

sambsnyd (approved):

rewrite-python/rewrite/tests/python/test_type_attribution.py
timtebeek: I'm going to agree with you there, even if it was just copying an existing pattern; looking at an alternative in
- https://github.com/openrewrite/rewrite/pull/8299

PR #8312, opened 2026-07-22
Represent oversize source files as Quarks in the RPC project parsers

jkschneider:
## What
Large source files can drive the RPC language engines' `parseProject` to multi-GB memory use or OOM, because every discovered file is parsed regardless of size. This caps each RPC project parser at **1 MB** (matching the JVM `JavaScriptParser` default): files over the cap are diverted off the parse path at discovery and represented as a content-less `org.openrewrite.quark.Quark` instead of being parsed or silently dropped.

## How
Each engine (JavaScript/TypeScript, Python, Go, C#) diverts oversize source files during discovery and emits a response item tagged `org.openrewrite.quark.Quark` with only the relative path — no engine-side object, no content. The Java client (`*RewriteRpc.parseProject`) short-circuits that tag and builds the `Quark` locally from the path + `FileAttributes`, so there's no `GetObject` round-trip and no content over the wire. The file is still represented in the LST, just as a stub.

For C# (MSBuild/Roslyn — no file walk) the size gate sits at the per-`Document` loop before the Roslyn tree is built; the `ParseSolution` response item gains a `sourcePath` field for the Quark case.

## Testing
- **rewrite-javascript**: new unit test asserting a >1 MB file lands in the new `unparseableFiles` bucket while a small file stays parseable; existing parse-project discovery tests still pass; `tsc` + Java compile clean.
- **Python / Go / C#**: `py_compile`, `go build`, `dotnet build`, and all four Java clients compile.
- **End-to-end**: building a JS project containing a 20 MB `.js` data file now records it as a `Quark` (LST manifest `data.js,Quark`) with the neighboring small file still parsed to a full AST — previously the engine OOM'd on that file.

### DIFF
--- a/rewrite-csharp/csharp/OpenRewrite/CSharp/SolutionParser.cs
+++ b/rewrite-csharp/csharp/OpenRewrite/CSharp/SolutionParser.cs
@@ -290,10 +290,19 @@ public class SolutionParser
     /// is true, each successfully parsed file is printed and compared against the original
     /// source; mismatches are also returned as <see cref="ParseError"/>.
     /// </summary>
+    // Files larger than this are recorded as Quarks rather than parsed into a
+    // Roslyn tree/LST. Matches the 1 MB cap in the other RPC engines.
+    private const long MaxParseableSizeBytes = 1024 * 1024;
+
+    // Relative paths of source files skipped by the most recent ParseProject call
+    // because they exceed MaxParseableSizeBytes; the RPC layer emits these as Quarks.
+    public readonly List<string> LastOversizePaths = new();
+
     public List<SourceFile> ParseProject(
         Solution solution, string projectPath, string rootDir,
         bool requirePrintEqualsInput = true)
     {
+        LastOversizePaths.Clear();
         var projectName = Path.GetFileNameWithoutExtension(projectPath);
         Log.Debug("ParseProject: starting {ProjectName}", projectName);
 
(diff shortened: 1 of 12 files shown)


rewrite-csharp/csharp/OpenRewrite/CSharp/SolutionParser.cs
```diff
@@ -290,10 +290,19 @@ public async Task<Solution> LoadAsync(string path, CancellationToken ct = defaul
     /// is true, each successfully parsed file is printed and compared against the original
     /// source; mismatches are also returned as <see cref="ParseError"/>.
     /// </summary>
+    // Files larger than this are recorded as Quarks rather than parsed into a
+    // Roslyn tree/LST. Matches the 1 MB cap in the other RPC engines.
+    private const long MaxParseableSizeBytes = 1024 * 1024;
```
shanman190, line 295: Just noting that in CLI, the max size is 10MB.

jkschneider: Yes, that CLI cap is for resource files.

jkschneider: In a real, source-heavy repo the biggest legitimately hand-authored file is ~33 KB, so the cap has ~30× of slack before it could clip anything a human wrote.

PR #8315, opened 2026-07-23
Accept Jupyter notebooks (.ipynb) in JsonParser

timtebeek:
Jupyter notebook `.ipynb` files are strict JSON, so `JsonParser` can parse and losslessly round-trip them; this adds `.ipynb` to its accepted extensions alongside `.json` and `Pipfile.lock`. This exposes notebooks to JSON recipes for envelope/metadata work (e.g. stripping outputs, bumping `language_info`, editing `metadata`); it does not parse the Python inside code cells, which remain opaque JSON strings. Adds tests asserting `.ipynb` acceptance and verifying a realistic notebook parses and prints byte-identically.


timtebeek:
This was exploratory in finding out how ipynb files are handled for the moment: No immediate need for them to be picked up as json for now; we can continue to parse them as plain text.

Issue #1061, opened 2021-09-25
FinalClass should not add 'final' modifier to abstract class


murdos:
> Illegal combination of modifiers: 'abstract' and 'final'

```kotlin
    @Test
    fun abstractClass() = assertUnchanged(
        before = """
            public abstract class A {
                
                public static void foo() {
                }
                
                private A() {
                }
            }
        """
    )
```

PR #8318, opened 2026-07-23
Java 25: recover the method invocation type when an argument type is unresolvable

timtebeek:
On JDK 25, javac's error recovery leaves a method invocation's select as a plain `ErrorType` when one of its arguments fails to resolve, even though the callee symbol resolved cleanly — so `rewrite-java-25` left the invocation's `getMethodType()` null (java-21 bound it). This surfaced as type-driven `MethodMatcher`s silently not firing and `RemoveUnusedImports` dropping still-used static imports, yielding non-compiling output (observed migrating rewrite-testing-frameworks to java-25, openrewrite/rewrite-testing-frameworks#1061).

### Root cause

Two JDK 24 javac changes ended the recovery shape the older parsers rely on:
- [JDK-8328536](https://bugs.openjdk.org/browse/JDK-8328536): `Attr.checkIdInternal` no longer early-returns on an erroneous expected type; it now fully instantiates the method type and stores it on the tree wrapped as a plain `ErrorType`, while the symbol stays a cleanly resolved `MethodSymbol`. Previously the select's type was left null and our `should-stop.ifError=GENERATE` pipeline had `postAttr` fill it with `Type$UnknownType` — the shape the java-11/17/21 mappings bind through their `UnknownType` branches.
- [JDK-8339296](https://bugs.openjdk.org/browse/JDK-8339296): removed `Type$UnknownType` entirely.

The existing `AttrRecover$RecoveryErrorType` reflection cannot rescue this case: javac only attaches it when resolution itself fails (arity/return-context mismatches), never for a resolved callee with an erroneous argument.

### Fix

In `ReloadableJava25TypeMapping.methodInvocationType`, when the select is a plain `ErrorType` but the callee is a cleanly-resolved `MethodSymbol`, recurse with the invocation type javac computed and kept in the error's original type (a resolved `Type.MethodType`, e.g. `(Object)Object` for `requireNonNull` — the same shape a non-erroneous call produces), falling back to the symbol's declared type otherwise.

### Tests

Three tests added to the shared TCK (`JavaParserTypeMappingTest`):
- the regression scenario, gated `@MinimumJava11` since JDK 8's javac predates this error recovery and never bound it; java-11/17/21 pass it via their pre-existing `UnknownType` branches
- a `@MinimumJava25` test pinning the resolved `java.lang.Object` parameter/return shape produced by the new fallback
- an overloaded-callee case (`String.valueOf(<unresolvable>)`) documenting the deliberate trade-off: with an unresolvable argument javac's recovery commits to an arbitrary candidate among the overloads, so we bind that candidate rather than leaving the method type null, and only name and declaring type are pinned

Verified locally: `JavaParserTypeMappingTest` green on the java-8/11/17/21/25 `compatibilityTest` suites, and the full `:rewrite-java-25:compatibilityTest` and `:rewrite-java-21:compatibilityTest` suites pass (648 tests each).

### Out of scope

On JDK 24 runtimes the java-21 parser hits the same plain-`ErrorType` recovery and still returns a null method type (its `UnknownType` branch is unreachable there since JDK-8339296 removed the class). This PR is deliberately scoped to the java-25 parser; porting the fallback to `ReloadableJava21TypeMapping` is left for a follow-up.

### DIFF
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -425,6 +425,15 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
             } catch (Exception e) {
                 // ignore
             }
+
+            // On JDK 24+ an unresolvable argument leaves the select as a plain ErrorType even though
+            // the callee resolved; recover the invocation type javac computed (kept in the error's
+            // original type), falling back to the symbol's declared type, to keep the JavaType.Method.
+            if (symbol instanceof Symbol.MethodSymbol && symbol.kind != Kinds.Kind.ERR &&
+                symbol.type != null && !(symbol.type instanceof Type.ErrorType) && !isUnknownType(symbol.type)) {
+                Type originalType = selectType.getOriginalType();
+                return methodInvocationType(originalType instanceof Type.MethodType ? originalType : symbol.type, symbol);
+            }
         }
 
         if (selectType == null || selectType instanceof Type.ErrorType || symbol == null || symbol.kind == Kinds.Kind.ERR ||
@@ -703,7 +712,7 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
     }
 
     /**
-     * Check for the `UnknownType` which existed up until JDK 22; starting with JDK 23 only the `ErrorType` is used.
+     * Check for the `UnknownType` which existed up until JDK 23; starting with JDK 24 only the `ErrorType` is used.
      * Uses reflection to avoid compile-time dependency on the class, which was removed in JDK 24.
      */
     public static boolean isUnknownType(@Nullable Type type) {
--- a/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
+++ b/rewrite-java-21/src/main/java/org/openrewrite/java/isolated/ReloadableJava21TypeMapping.java
@@ -425,15 +425,6 @@ class ReloadableJava21TypeMapping implements JavaTypeMapping<Tree> {
             } catch (Exception e) {
                 // ignore
             }
-
-            // On JDK 24+ an unresolvable argument leaves the select as a plain ErrorType even though
-            // the callee resolved; recover the invocation type javac computed (kept in the error's
-            // original type), falling back to the symbol's declared type, to keep the JavaType.Method.
-            if (symbol instanceof Symbol.MethodSymbol && symbol.kind != Kinds.Kind.ERR &&
-                symbol.type != null && !(symbol.type instanceof Type.ErrorType) && !isUnknownType(symbol.type)) {
-                Type originalType = selectType.getOriginalType();
-                return methodInvocationType(originalType instanceof Type.MethodType ? originalType : symbol.type, symbol);
-            }
         }
 
         if (selectType == null || selectType instanceof Type.ErrorType || symbol == null || symbol.kind == Kinds.Kind.ERR ||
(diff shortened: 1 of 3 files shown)


timtebeek:
**Review summary** (deep review: JDK source archaeology, javac probes on zulu 25.0.3/26, and a full TCK matrix run across the 8/11/17/21/25 parser modules)

### Root cause — why only the Java 25 parser

Two JDK 24 javac changes, verified against `jdk-21-ga`…`jdk-25-ga` sources:

- [JDK-8328536](https://bugs.openjdk.org/browse/JDK-8328536): through JDK 23, `Attr.checkIdInternal` early-returned on an erroneous expected type, leaving the select's type null; our `should-stop.ifError=GENERATE` pipeline then had `postAttr` fill it with `Type$UnknownType`. JDK 24 instead fully instantiates the method type and stores it wrapped as a **plain `ErrorType`**, while the symbol remains a cleanly resolved `MethodSymbol`.
- [JDK-8339296](https://bugs.openjdk.org/browse/JDK-8339296): removed `Type$UnknownType` entirely.

The existing `AttrRecover$RecoveryErrorType` reflection can never rescue this case: it is only attached when resolution itself fails (arity/return-context mismatches), never for a resolved callee with an erroneous argument. So the new fallback is the only recovery path on JDK 25 — the right fix in the right place. Guard ordering is safe: genuine `RecoveryErrorType` symbols are `kind=ERR` and are excluded.

### Verified

- `:rewrite-java-{8,11,17,21,25}:compatibilityTest --tests JavaParserTypeMappingTest`: all green (8 skips via `@MinimumJava11`); full `:rewrite-java-25:compatibilityTest` = 646 tests, 0 failures.
- Note the 11/17/21 parsers pass the new test through their **pre-existing `UnknownType` branch** (returnType `Unknown`, no parameter types), not this fallback — the assertions are loose enough to accept both shapes.
- Refuted non-issues (checked, no action needed): the theoretical ForAll↔ErrorType recursion cycle is unreachable (all `new ForAll(` sites in javac 25 wrap a `MethodType`), no new CCE at the `(Symbol.MethodSymbol)` cast, no `assert returnType != null` trip via the fallback, and omitting `isUnknownType` from the 25 file is correct.

### Points to consider

1. **Prefer `selectType.getOriginalType()` over `symbol.type`.** On JDK 25/26 the plain `ErrorType` empirically keeps the *resolved* invocation type in `getOriginalType()` (`(java.lang.Object)java.lang.Object` for `requireNonNull`), whereas `symbol.type` gives the unsubstituted `<T>(T)T`. The current fallback therefore produces a declaration-shaped Method whose signature is byte-identical to `methodDeclarationType`'s cache key, so the invocation aliases the declaration's cached instance and violates `methodInvocationType`'s "resolved types" javadoc. Recursing with `getOriginalType()` when it is a `Type.MethodType` (falling back to `symbol.type` otherwise) yields resolved types and mirrors the dispatch already used in `type(Type, Symbol)`.
2. **Phantom-overload risk.** For an overloaded callee, javac's recovery commits the symbol to an arbitrary pick: probed `String.valueOf(Unknown.x())` on 25.0.3 attributes as `valueOf(boolean)` with concrete `parameterTypes=[boolean]`, where the parser previously abstained with null. Given incomplete classpaths are the normal OpenRewrite condition, `MethodMatcher`-based recipes can now match/transform calls that actually bind to a different overload (constructors too). Probably the right trade-off — null attribution had its own fallout (e.g. `RemoveUnusedImports` dropping used static imports) — but worth a TCK test pinning the overloaded-callee behavior.
3. **JDK 24 runtimes still unfixed.** JDK 21–24 select the Java 21 parser, and on JDK 24 it hits the same plain-`ErrorType` select and returns null (#6740 only avoided the `NoClassDefFoundError`; the `isUnknownType` branch is unreachable there). Porting this same block to `ReloadableJava21TypeMapping` is provably inert on 21/22 runtimes (select is `UnknownType` there) and would close the gap; no CI leg exercises 21-on-24 today. JDK 23 is unaffected (`UnknownType` still exists — the `isUnknownType` javadoc saying "starting with JDK 23" is off by one; the switch shipped in JDK 24).
4. Cosmetic: the `@Issue` annotation points at rewrite-testing-frameworks **PR** #1061 rather than an issue.

sambsnyd (approved):

PR #8319, opened 2026-07-23
Route addToRootPom=true pins to the reactor root in UpgradeTransitiveDependencyVersion

steve-aom-elliott:
## Summary

`UpgradeTransitiveDependencyVersion(..., addToRootPom=true)` silently dropped every pin in a multi-module reactor whenever the vulnerable transitive was held only by a child module. Two things combined to hide the write:

1. `UpgradeTransitiveDependencyVersion.visitDocument` returned early on the reactor root because the root itself has no matching transitive of its own (`matchingDependencies.isEmpty()`).
2. `AddManagedDependency.visitDocument`'s write gate (`!Boolean.TRUE.equals(addToRootPom) || acc.rootPoms.contains(document)`) refused to write on the child pom because only reactor-root poms are in `acc.rootPoms`.

Result: the recipe produced no changes at all, even though a child module clearly held the transitive.

## Change

Give `UpgradeTransitiveDependencyVersion` a real scanner: for every pom that holds a matching transitive, walk up `parentPomIsProjectPom()` to find the reactor root's GAV and record the coordinate against it in a new `AddManagedDependency.Scanned#reactorRootTransitives` map.

In the visitor phase, branch on `addToRootPom`:
- `true` — fire only on reactor roots (`!parentPomIsProjectPom()`) and pin every coordinate the scanner collected against that root's GAV.
- `null`/`false` — existing behavior unchanged (per-pom matching with parent-preference dedup).

`AddManagedDependency.visitDocument`'s guard already routes correctly when the root pom is visited (root is in `acc.rootPoms`), so no changes needed there.

## Behavior change

Only the `addToRootPom=true` path is affected. The `null`/`false` path — the common case, including everything `UpgradeTransitiveDependencyVersionTest` covered before — is unchanged.

Callers that were relying on `addToRootPom=true` silently no-op-ing in reactors will now see the recipe produce edits.

## Known scope limitation

If a child has its own competing `<dependencyManagement>` entry for the same coordinate, Maven's inheritance rule means the child's local entry shadows the parent's — the root-scoped pin won't take effect for that specific module until the child entry is removed. This PR does not touch child poms. Filed as follow-up.

## Test plan

- [x] New `addsManagedDependencyToReactorRootWhenAddToRootPomAndOnlyChildHoldsTheTransitive` in `UpgradeTransitiveDependencyVersionTest` covers the reactor-root write.
- [x] Full `UpgradeTransitiveDependencyVersionTest` and `AddManagedDependencyTest` suites pass.
- [x] Full `rewrite-maven:test` module passes.

Draft — soliciting feedback on the accumulator shape and the child-cleanup boundary before finalizing.

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/AddManagedDependency.java
@@ -29,8 +29,11 @@ import org.openrewrite.semver.VersionComparator;
 import org.openrewrite.xml.tree.Xml;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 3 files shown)


steve-aom-elliott:
Superseded by other changes from Shannon

PR #8323, opened 2026-07-24
Python: route imports through a PythonImportService so Java ChangeType stops corrupting files

mtthwcmpbll:
- Fixes the `Tree not found` failure running `UpgradeToPython313`/`314` on Python repos ([customer-requests#2858](https://github.com/moderneinc/customer-requests/issues/2858)): the upgrade recipes delegate ~30 PEP 585 `ChangeType`s to Java's `ChangeType`, which ran Java's `AddImport` against Python LSTs — emitting invalid `import a.b.C` welded onto the leading docstring and crashing the RPC server.

A new `PythonImportService` (wired into `Py.CompilationUnit.service()`, matching the Kotlin/Go precedent) routes add/remove-import to the Python side over RPC; supporting changes add a `visitorOptions` `RewriteRpc.visit` overload, `ImportService.removeImportVisitor`, service-based `maybeRemoveImport`, docstring-safe Python `add_import`/`remove_import`, hub-local facade routing for built-in visitors (was `No child owns visitor 'AutoFormatVisitor'`), a `replace_if_changed` fix for 38 `with_*` methods that called an unimported `replace()`, and `ConcurrentHashMap` object maps in `RewriteRpc` to close the concurrent-visit race behind `Tree not found`.

Non-Python languages are unaffected (all resolve to the same `RemoveImport`/`AddImport` as before; `rewrite-core`, `rewrite-java`, and Kotlin suites pass), and the shared `ConcurrentHashMap` fix also hardens JS/C#/Go against the same race. Verified end-to-end on `jd/tenacity` and `psf/requests` via Moderne CLI 4.4.1: 0 errors, valid imports, no docstring clobbering.


rewrite-core/src/test/java/org/openrewrite/rpc/RewriteRpcTest.java
```diff
@@ -239,6 +239,18 @@ void evictDropsTreeFromBothPeers() {
...
     }
 
+    /**
+     * The id-keyed object maps must stay concurrent, since {@code GetObject.Handler} reads them from
+     * its own thread pool while in-flight visits write them and a {@code HashMap} resize race dropped
+     * a present key as "Tree not found" (customer-requests#2858) -- an invariant rather than a
+     * behavioral test because the resize window is too small to hit deterministically.
+     */
+    @Test
+    void sharedObjectMapsAreThreadSafe() {
```
shanman190: I don't find this test particularly useful.

rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
```diff
@@ -241,6 +241,12 @@ private void addImport(JavaType.FullyQualified owningClass) {
                 j = ((TypedTree) tree).withType(updateType(((TypedTree) tree).getType()));
             } else if (tree instanceof JavaSourceFile) {
                 JavaSourceFile sf = (JavaSourceFile) tree;
+                // Languages that keep imports among the statements (Python) expose none through
+                // getImports(), so hand removal to their import service; on a genuinely import-less
+                // Java file there is nothing to remove and this is a no-op.
+                boolean removeThroughImportService = targetType instanceof JavaType.FullyQualified &&
```
shanman190: `sf.getImports().isEmpty()` as the proxy for "statement-based-import language" works today, and I confirmed it's harmless for Java — `RemoveImport` is a genuine no-op on an import-less file (empty `flatMap`, and the only structural branch is guarded by `c != cu`). But the comment right above it already names the real condition, and this breaks quietly the first time a statement-import language returns a non-empty `getImports()`.

Since `ImportService` is already the dispatch point, could this be a capability on the service instead — something like `usesStatementBasedImports()` defaulting to `false`, overridden in `PythonImportService`? Then the branch states the thing it means, and Python is the only place that has to know about Python.

rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -2008,6 +2029,47 @@ def _hub_release(obj_id: str) -> None:
         _hub_send_next[bundle] = checkpoint
 
 
+def _hub_is_builtin_visitor(visitor_name: Optional[str]) -> bool:
+    """True for a visitor the server implements itself rather than one a recipe bundle owns."""
+    return bool(visitor_name) and visitor_name in _get_visitor_registry()
+
+
+def _hub_local_visit(visitor_items: List[dict], params: dict) -> List[dict]:
```
shanman190, line 2037: This is the piece I'd most like a test on. It's the third writer of `_hub_tree` — previously only `_hub_acquire` (from Java) and `_hub_pull_child_edit` (from a child) wrote it — and the correctness of the hub/child split now rests on an invariant that isn't stated anywhere: it updates `_hub_tree` and `local_objects` but deliberately *not* `_hub_served`, which is what makes the next `_hub_serve_child` diff carry the local edit down to the child. That's right as written, and it's exactly the kind of thing that gets "tidied" into a bug later. Worth a comment saying why `_hub_served` is untouched.

The new tests in `tests/rpc/test_facade.py` stub `local_visit` out entirely, so they cover routing but never execute this function. Given a mistake here surfaces as the same class of `Tree not found` / stale-tree failure the PR is fixing, a direct test — acquire, run a real `AddImport`, assert `_hub_tree` advanced and a subsequent `_hub_serve_child` emits the delta — would be worth more than the routing tests.

rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -2008,6 +2029,47 @@ def _hub_release(obj_id: str) -> None:
...
+        before = tree
+        visitor = _instantiate_visitor(item['visitor'], ctx, item.get('visitorOptions'))
+        after = visitor.visit(before, ctx, cursor)
+        results.append({
+            'modified': after is not before,
+            'deleted': after is None,
+            'hasNewMessages': False,
+            'searchResultIds': [],
+        })
+        if after is None:
```
shanman190, line 2061: On `after is None` this breaks out and reports `'deleted': True`, but leaves the stale tree in `_hub_tree`/`local_objects`, so the next `_hub_serve_child` would serve a file the visitor deleted.

Unreachable for the three visitors in the registry today, and the Java side guards against null on top of that. Flagging only because the registry is the extension point — the next visitor added there inherits the bug silently. `_hub_release(tree_id)` on that branch would close it.

shanman190 (changes requested):
The Python-side work here looks right to me and I'd take it as-is — `PythonImportService` following the Kotlin/Go service precedent, the docstring/blank-line fixes in `add_import`/`remove_import`, and the `replace_if_changed` repair are all clearly good.

My question is on the `rewrite-core` hunk, because I don't think the stated root cause survives the scoping.

`RewriteRpcProcessManager` holds `private final ThreadLocal<R> rpc`, and its javadoc says it allows "a single Rewrite RPC subprocess per thread." `PythonRewriteRpc.get()` resolves through it, so each worker thread owns its own `RewriteRpc`, its own subprocess, and its own `localObjects`/`remoteObjects` — those maps never cross a thread boundary between workers. The ManagedBlocker comment in `send()` points the same way: it's guarding against FJP compensation minting *extra* per-thread instances, which only makes sense because the state is per-thread.

That leaves two threads per `RewriteRpc` — the owning thread and the jsonrpc reader/dispatcher that completes futures and services inbound `GetObject`/`Visit`. As far as I can trace, those strictly alternate: the owner is parked in `send()`'s `getNow`/`sleep` poll loop for the entire window in which the peer can call back, and the nested hub → child → Java flows keep Java blocked throughout. A `HashMap` resize race needs simultaneous structural modification, and I don't see where that can happen.

If that's right, then `ConcurrentHashMap` is still defensible — there's no explicit happens-before edge between the owner's `localObjects.put` before `send()` and the dispatcher's read, so visibility hardening is a real argument — but it's a different argument than the one in the commit message, and it doesn't explain the failure it's credited with fixing.

What makes me want to resolve this before merge rather than after: the commit message says "Regression test reproduces the DELETE on a plain HashMap," but `sharedObjectMapsAreThreadSafe` asserts `isInstanceOf(ConcurrentMap.class)` and its javadoc says the window is "too small to hit deterministically." Both can't be true. If the resize race can't be reproduced *and* the per-thread scoping says it can't occur, then the actual cause of the `Tree not found` failure is still out there and this change may have only moved the timing.

Two ways I'd be happy:

- Land the reproduction. If there's a path where two threads share one `RewriteRpc` that I've missed, that's worth having in the test suite permanently — it'd change how several of us reason about this class.
- Or drop the resize-race rationale, keep the maps concurrent as visibility hardening against the owner/dispatcher alternation, say so in the javadoc, and split the `Tree not found` investigation into its own issue.

Happy to be wrong on this — I got the threading backwards once already while reading it.

---

**Architecture note (not blocking, just worth writing down somewhere durable):** the facade running visitors itself is a new shape, and the split isn't deferred-vs-immediate — it's which process initiated the call. Python-initiated `maybe_add_import` (e.g. `recipes/change_import.py:193`) drains through the visitor's own `_after_visit` in the child and never reaches the facade. Only Java-initiated service dispatch does, both deferred (`maybeAddImport`) and immediate (`maybeRemoveBlankLines` calling `normalizeFormatVisitor(...).visit(...)` inline).

The consequence is that there are now two implementations of the same import insertion selected by who asked: the child's `rewrite` for Python-initiated, the hub's for Java-initiated. Same file, potentially different versions if a bundle pins something else. Not a problem today — it was a hard failure before this PR — but much cheaper to note now than to rediscover from a bug report.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -76,12 +77,16 @@ public class RewriteRpc {
...
+     * multiple in-flight {@link #visit} calls write them, and a {@code HashMap} resize racing a
+     * {@code get} would drop a present key, which the peer surfaces as "Tree not found".
      */
     @VisibleForTesting
-    final Map<String, Object> remoteObjects = new HashMap<>();
+    final Map<String, Object> remoteObjects = new ConcurrentHashMap<>();
 
     @VisibleForTesting
-    final Map<String, Object> localObjects = new HashMap<>();
+    final Map<String, Object> localObjects = new ConcurrentHashMap<>();
```
shanman190: The javadoc says `GetObject.Handler` reads these "while multiple in-flight `visit` calls write them." Given `RewriteRpcProcessManager`'s `ThreadLocal`, multiple in-flight visits are on different `RewriteRpc` instances with different maps — so the concurrent writer described here doesn't exist, and the only genuine cross-thread pair is this instance's owner and its jsonrpc dispatcher. Worth rewording to whichever model you land on, since this javadoc is the thing the next person will reason from.

PR #8325, opened 2026-07-24
Make Kotlin's statement/expression wrappers first-class visited nodes

timtebeek:
`ChangePackage` (and other `java.*` recipes/templates) threw `class K$StatementExpression cannot be cast to class J$MethodInvocation` on Kotlin sources, as surfaced by rewrite-testing-frameworks' MockWebServer migration over a `return if (…) … else …` in a `Dispatcher` override.

The root cause: Kotlin's statement/expression wrapper types had an `acceptKotlin` that bypassed `visitStatement`/`visitExpression` and descended straight into the wrapped node, so JavaTemplate's scope-based replacement could never target the wrapper itself and `apply()` returned it unchanged — which callers then cast to `J.MethodInvocation`.

Both Kotlin wrappers are now first-class visited nodes, each routed through the hook matching the role it actually plays:

- **`K.StatementExpression`** wraps a `Statement` so it can occupy an `Expression` slot (`convertToExpression` wraps only when `j instanceof Statement && !(j instanceof Expression)`). New `KotlinVisitor#visitStatementExpression` routes through `visitExpression`; the wrapped statement continues to route through `visitStatement` on its own. `JavaTemplateJavaExtension#visitExpression` handles `STATEMENT_PREFIX` coordinates when the node is also a `Statement` and parses the template as an expression, which is the more accurate treatment for a node in an expression slot.
- **`K.ExpressionStatement`** is the mirror image — it wraps an `Expression` so it can occupy a `Statement` slot — and had the identical defect. It is far more common: a bare `a + b`, `a > b`, or `a` in any block is wrapped, and `convertToStatement` is called from nine sites in `KotlinTreeParserVisitor`. Confirmed by repro: `JavaTemplate.apply(cursor, exprStmt.getCoordinates().replace(), exprStmt)` returned the untouched wrapper, and the caller's cast threw `K$ExpressionStatement cannot be cast to J$MethodInvocation`. New `KotlinVisitor#visitExpressionStatement` routes through `visitStatement` only; routing through `visitExpression` as well would present the wrapper *and* its child to type-matching recipes as two indistinguishable nodes, since the wrapper delegates `getType()`/prefix/markers to the child.

Both get the matching narrowing override in `KotlinIsoVisitor`, which also picks up a missing `visitAnnotatedExpression` override (pre-existing gap, unrelated).

Regression tests `KotlinTemplateTest.wrapControlFlowReturnExpressionWithBuild` and `KotlinTemplateTest.replaceExpressionStatementWithTemplate` each reproduce their exact `ClassCastException`; both were verified to fail without the corresponding fix. Full `rewrite-kotlin` and `rewrite-gradle` suites pass.

Follow-ups, not in this PR:

- #8327 — `JavaTemplate.apply()` silently no-ops when a `REPLACEMENT` scope is never matched, which is what makes this bug family present as a `ClassCastException` far from its cause.
- `G.ExpressionStatement` is still transparent in the same way, and `rewrite-gradle` parses build scripts as Groovy, so it sits on a hotter path than either Kotlin wrapper. Worth its own PR gated on the `rewrite-groovy` and `rewrite-gradle` suites.

### DIFF
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
@@ -118,6 +118,11 @@ public class KotlinIsoVisitor<P> extends KotlinVisitor<P> {
     }
 
     @Override
+    public K.StatementExpression visitStatementExpression(K.StatementExpression statementExpression, P p) {
+        return (K.StatementExpression) super.visitStatementExpression(statementExpression, p);
+    }
+
+    @Override
     public K.TypeAlias visitTypeAlias(K.TypeAlias typeAlias, P p) {
         return (K.TypeAlias) super.visitTypeAlias(typeAlias, p);
     }
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
@@ -63,6 +63,11 @@ public class KotlinIsoVisitor<P> extends KotlinVisitor<P> {
     }
 
     @Override
+    public K.ExpressionStatement visitExpressionStatement(K.ExpressionStatement expressionStatement, P p) {
+        return (K.ExpressionStatement) super.visitExpressionStatement(expressionStatement, p);
+    }
+
+    @Override
     public K.FunctionType visitFunctionType(K.FunctionType functionType, P p) {
         return (K.FunctionType) super.visitFunctionType(functionType, p);
     }
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
@@ -28,6 +28,11 @@ public class KotlinIsoVisitor<P> extends KotlinVisitor<P> {
     }
 
     @Override
+    public K.AnnotatedExpression visitAnnotatedExpression(K.AnnotatedExpression annotatedExpression, P p) {
+        return (K.AnnotatedExpression) super.visitAnnotatedExpression(annotatedExpression, p);
+    }
+
+    @Override
     public K.AnnotationType visitAnnotationType(K.AnnotationType annotationType, P p) {
         return (K.AnnotationType) super.visitAnnotationType(annotationType, p);
     }
--- a/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinVisitor.java
+++ b/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinVisitor.java
@@ -274,17 +274,11 @@ public class KotlinVisitor<P> extends JavaVisitor<P> {
 
     public J visitStatementExpression(K.StatementExpression statementExpression, P p) {
         K.StatementExpression s = statementExpression;
-        Statement temp = (Statement) visitStatement(s, p);
+        Expression temp = (Expression) visitExpression(s, p);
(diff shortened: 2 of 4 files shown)


timtebeek:
## Why `K.StatementExpression` was transparent to visitors

Some archaeology on where this came from, since it explains the fix. The transparent wrapper has been that way since the class was born in the old `openrewrite/rewrite-kotlin` repo.

- The commit that first added the class (`80876c574c`, *"Added K.StatementExpression to wrap java statements that are expressions in Kotlin"*) already had the exact inline `acceptKotlin` we imported — `v.visit(getStatement())` with the unwrap/re-wrap logic.
- `KotlinVisitor` in that repo **never** had a `visitStatementExpression` (0 occurrences on its final `main`). So the import (`5b5b43d311`) copied it faithfully; the gap predates the merge.

**Why it was built that way** is clear from the originating PR, [openrewrite/rewrite-kotlin#174](https://github.com/openrewrite/rewrite-kotlin/pull/174) — *"Automatically detect and convert Statements to StatementExpression where Expressions are required"* (merged 2023-07-11):

> - Added optional conversion to StatementExpressions in `convert` and `convertAll`.
> - Enabled Statement conversion on all existing calls of `convert`/`convertAll` that expect expressions.
> - Replaced all `visitElement`(s) that cast to an Expression with `convert`.

So `K.StatementExpression` was conceived purely as a **parser-side type adapter**: when the parser needs to put a `J.If`/`J.Switch`/etc. into an `Expression` slot, `convert(...)` wraps it so the LST type-checks. Its whole job was to *disappear* — hence the class doc "Has no state or behavior of its own aside from the Expression it wraps," and hence no visit hook: visitors were meant to see straight through to the wrapped statement.

That transparency was fine for the parser, but was never reconciled with the **recipe** side, where scope/coordinate-based rewriting (JavaTemplate) needs to target the wrapper node by identity. Every *other* `K` statement-expression type (`K.Return`, `K.Unary`, `K.SpreadArgument`, `K.StringTemplate`, …) already routes through `visitStatement`/`visitExpression`; `K.StatementExpression` (and its sibling `K.ExpressionStatement`) were the exceptions. This PR keeps the parse-time behavior (still transparent for plain traversal) while making it a first-class, targetable node — closing a gap that's existed since the wrapper was introduced in mid-2023.

rewrite-kotlin/src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
```diff
@@ -117,6 +117,11 @@ public K.SpreadArgument visitSpreadArgument(K.SpreadArgument spreadArgument, P p
         return (K.SpreadArgument) super.visitSpreadArgument(spreadArgument, p);
     }
 
+    @Override
+    public K.StatementExpression visitStatementExpression(K.StatementExpression statementExpression, P p) {
+        return (K.StatementExpression) super.visitStatementExpression(statementExpression, p);
+    }
```
sambsnyd, line 133: Do we also do this on the groovy side? It has similar constructs

timtebeek: Yes it looks like we do the same on the Groovy side. It did require KotlinTemplate here though to even see the issue, and we've not seen the same for Groovy just yet, so even if we were to tackle it there, I'd propose we do that separately, and only when needed. Unless you feel strongly about changing that here & now.

I'm mostly looking to unblock a fix in rewrite-testing-frameworks specifically for Kotlin, which only needed this change.

sambsnyd: I figure we should be consistent so that subtle differences in behavior don't surprise us and cause a repeat of this investigation but for groovy or scala or whatever later

sambsnyd (approved):

PR #8329, opened 2026-07-26
Add Gradle recipe `RemoveEmptyBuildscriptBlock`

sullis:
Adds a new recipe `org.openrewrite.gradle.RemoveEmptyBuildscriptBlock` that removes `buildscript` blocks which contribute nothing to the build, and wires it into `org.openrewrite.gradle.GradleBestPractices`.

## Behavior

- Applies to `build.gradle(.kts)` and `settings.gradle(.kts)` (Groovy and Kotlin DSL).
- Removes a `buildscript` block whose closure body is empty, or contains only other empty blocks (e.g. an empty `dependencies` or `repositories` block).
- Leaves the block alone when it contains a comment anywhere inside, so no comment is silently deleted.
- Removal is done at the statement-list level rather than by returning `null` from the method invocation visit, so the following statement inherits the removed block's prefix — otherwise removing the first statement in a file leaves its leading blank lines behind.

## Tests

`RemoveEmptyBuildscriptBlockTest` covers the empty block, a block containing only empty nested blocks, the Kotlin DSL, a block between other statements, and the negative cases (real dependencies, a comment in the block, a comment in a nested block). `GradleBestPracticesTest` gains a case verifying the recipe runs as part of the composite recipe.


sullis:
LGTM

PR #8330, opened 2026-07-26
Python: fix match-statement parse desync that corrupts trees over RPC

shanman190:
Two spots in the Python `match`-case parser left source unconsumed, drifting the source cursor so every later node parsed against misaligned text and came out with `None` children. `RpcSendQueue` then crashes at `el.element._id` on the `None` child, failing the tree transfer (`handle_get_object` returns a degenerate `[END_OF_OBJECT]`). Because `match`/`case` is Python 3.10+ and absent from older codebases, the defect only surfaces on modernized code.

## Fixes (both in `_parser_visitor.py`)

- **Guards on value/capture patterns** (`case 1 if x:`, `case Enum.X if x:`, `case v if x:`): the guard source (`if <cond>`) was consumed only when the pattern wrapped as `py.MatchCase`. Value and capture patterns are returned bare, so the guard text stayed in the buffer. Consume it unconditionally and wrap the bare pattern in `py.MatchCase(VALUE/CAPTURE)` so the guard has a home. The existing `py.MatchCase` codec (Python and Java) and printer already handle these kinds — no model or codec change needed.
- **Parenthesized GROUP as a class keyword value** (`case Foo(k=(A() | B())):`): keyword patterns used `__convert` (which does not consume parens) while positional patterns used `__convert_match_pattern`. Use the latter for both, consistent with positional handling.

## Verification

- New regression tests (`tests/rpc/test_match_guard_desync.py`) cover both shapes end-to-end (parse → `RpcSendQueue` → `RpcReceiveQueue` → print), asserting no `None`-element padding and a clean round-trip.
- Serialize→deserialize→print round-trip run over 20,635 real files (numpy, home-assistant, ansible, botocore): **3 crashing files → 0**, no regressions, existing `match`/pattern suite green.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/python/_parser_visitor.py
+++ b/rewrite-python/rewrite/src/rewrite/python/_parser_visitor.py
@@ -1156,9 +1156,32 @@ class ParserVisitor(ast.NodeVisitor):
 
         # Use __convert_match_pattern to handle parentheses (GROUP patterns)
         pattern = self.__convert_match_pattern(node.pattern)
-        if isinstance(pattern, py.MatchCase) and node.guard:
+        if node.guard:
+            # The guard source (`if <cond>`) must be consumed whenever a guard is present, or the
+            # cursor drifts and every subsequent node is parsed against misaligned text. Value and
+            # capture patterns are returned bare (not wrapped in py.MatchCase), so wrap them here to
+            # give the guard a home — the printer emits LITERAL/VALUE/CAPTURE as the inner element.
             guard = self.__pad_left(self.__source_before('if'), self.__convert(node.guard))
-            pattern = pattern.padding.replace(guard=guard)
+            if isinstance(pattern, py.MatchCase):
+                pattern = pattern.padding.replace(guard=guard)
+            else:
+                kind = (py.MatchCase.Pattern.Kind.CAPTURE if isinstance(node.pattern, ast.MatchAs)
+                        else py.MatchCase.Pattern.Kind.VALUE)
+                pattern = py.MatchCase(
+                    random_id(),
+                    Space.EMPTY,
+                    Markers.EMPTY,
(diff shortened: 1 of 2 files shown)


rewrite-python/rewrite/src/rewrite/python/_parser_visitor.py
```diff
@@ -1156,9 +1156,32 @@ def visit_match_case(self, node):
 
         # Use __convert_match_pattern to handle parentheses (GROUP patterns)
         pattern = self.__convert_match_pattern(node.pattern)
-        if isinstance(pattern, py.MatchCase) and node.guard:
+        if node.guard:
+            # The guard source (`if <cond>`) must be consumed whenever a guard is present, or the
+            # cursor drifts and every subsequent node is parsed against misaligned text. Value and
+            # capture patterns are returned bare (not wrapped in py.MatchCase), so wrap them here to
+            # give the guard a home — the printer emits LITERAL/VALUE/CAPTURE as the inner element.
```
shanman190:
```suggestion
```

rewrite-python/rewrite/src/rewrite/python/_parser_visitor.py
```diff
@@ -1403,7 +1426,10 @@ def visit_MatchClass(self, node):
                             Markers.EMPTY,
                             cast(j.Identifier, self.__convert_name(kwd)),
                             _EMPTY_LIST,
-                            self.__pad_left(self.__source_before('='), self.__convert(node.kwd_patterns[i])),
+                            # __convert_match_pattern (not __convert) so a parenthesized GROUP value
+                            # like `k=(A() | B())` has its parens consumed — matching positional
+                            # patterns above; __convert alone leaves the parens and drifts the cursor.
```
shanman190:
```suggestion
```

PR #8335, opened 2026-07-27
Restore class loader aware `JavaParser.runtimeClasspath()`

### LINKED ISSUE #8334: Refaster recipes silently no-op under rewrite-maven-plugin since 8.80.0: JavaParser.runtimeClasspath() switched from ClassGraph to java.class.path

Broken since rewrite `8.80.0` (still present on `main` as of 2026-07-27).

Observed through `org.openrewrite.maven:rewrite-maven-plugin`:
- `6.36.0` (rewrite `8.79.0`) — works
- `6.37.0` (rewrite `8.80.0`) — broken
- `6.44.0` (latest) — still broken

`rewrite-maven-plugin`, recipes supplied via `rewrite.recipeArtifactCoordinates`.

Reproducer: https://github.com/jtama/openrewrite-refactoring-as-code/tree/devquest (originally reported in the OpenRewrite Slack #questions channel).

```bash
mvn -f toxic-library/pom.xml          clean install
mvn -f toxic-library-remover/pom.xml  clean install

cd toxic-library-user
mvn -U org.openrewrite.maven:rewrite-maven-plugin:6.36.0:run \
  -Drewrite.recipeArtifactCoordinates=com.github.jtama:toxic-library-remover:1.0.1-SNAPSHOT \
  -Drewrite.activeRecipes=io.github.jtama.openrewrite.RemovesThatToxicDependency
# then repeat with 6.37.0 / 6.44.0 on a pristine copy and diff the results
```

The Refaster-generated recipe `RemoveFooBarUtilsIsEmptyRecipes` rewrites `FooBarUtils.isEmpty(x)` into `x == null || x.isEmpty()`, so that the accompanying `org.openrewrite.maven.RemoveDependency` can safely drop `toxic-library` from the pom.

### What did you see instead?

The Refaster recipe **silently does nothing** — no error, no warning, it simply never appears in the "Changes have been made to ... by:" output. `RemoveDependency` still removes the dependency from the pom, so the result is a project whose pom no longer declares `toxic-library` while the sources still call `FooBarUtils.isEmpty(...)` and import `com.github.jtama.toxic.FooBarUtils`. **The output does not compile.**

```diff
  import com.github.jtama.toxic.FooBarUtils;   // still here, dependency is gone
- System.out.println(hostel.getName() == null || hostel.getName().isEmpty());   // 8.79.0
+ System.out.println(FooBarUtils.isEmpty(hostel.getName()));                    // 8.80.0+
```

Residual references to the removed library after the run:

| plugin | rewrite core | files still referencing `com.github.jtama.toxic` |
|---|---|---|
| 6.36.0 | 8.79.0 | 1 (pre-existing, unrelated) |
| **6.36.0 source, only `rewrite.version` bumped** | **8.80.0** | **5** |
| 6.37.0 | 8.80.0 | 7 |
| 6.40.0 | 8.83.0 | 5 |
| 6.44.0 | latest | 5 |

Rebuilding `rewrite-maven-plugin` v6.36.0 with *only* `<rewrite.version>` changed from `8.79.0` to `8.80.0` reproduces the regression, which rules out the plugin itself.

### Root cause

- `839bfb7f9` ("Fix being unable to correctly load recipes whose superclasses come from other jars", #7338) changed `RuntimeClasspathCache` in `rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java`:

```diff
-import io.github.classgraph.ClassGraph;
-                    runtimeClasspath = paths = new ClassGraph()
-                            .disableNestedJarScanning()
-                            .getClasspathURIs().stream()
-                            .filter(uri -> "file".equals(uri.getScheme()))
-                            .map(Paths::get)
-                            .collect(toList());
+                    String cp = System.getProperty("java.class.path");
+                    runtimeClasspath = paths = cp == null ? Collections.emptyList() :
+                            Arrays.stream(cp.split(System.getProperty("path.separator")))
+                                    .map(Paths::get)
+                                    .filter(Files::exists)
+                                    .collect(toList());
```

ClassGraph walks the **classloader hierarchy**; `java.class.path` only holds the JVM's *initial* classpath. Inside a Maven JVM that is a single jar:

```
$ mvn org.apache.maven.plugins:maven-help-plugin:3.5.1:system | grep '^java.class.path'
java.class.path=/opt/homebrew/Cellar/maven/3.9.16/libexec/boot/plexus-classworlds-2.11.0.jar
```

Everything Maven loads — the plugin, its dependencies, and recipe artifacts from `recipeArtifactCoordinates` — lives in a plexus `ClassRealm` (`URLClassLoader`), which is invisible to that property. So `JavaParser.runtimeClasspath()` now returns effectively nothing under Maven.

That matters because rewrite-templating emits exactly this for every Refaster `@BeforeTemplate`:

```java
before = JavaTemplate.builder("com.github.jtama.toxic.FooBarUtils.isEmpty(#{value:any(java.lang.String)})")
        .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
        .build();
```

With an empty classpath the template stub fails type attribution, `matcher.find()` never matches, and the recipe no-ops silently.

Two details that make this nastier than a normal regression:

1. **It fails silently and produces non-compiling output.** Nothing is logged. Combined with a `RemoveDependency` in the same recipe list, the user gets a broken tree rather than an error.
2. **Unit tests cannot catch it.** I ran the reproducer's own `RewriteTest` against both `8.79.0` and `8.80.0` — green on both, because under Surefire the recipe classes *are* on `java.class.path`. It only manifests when recipes are loaded through an isolated classloader, i.e. via the build plugins.

In the reproducer, `RemoveFooBarUtilsIsEmpty` is the only Refaster recipe present and it is precisely the only one that breaks; hand-written recipes in the same run (`UseObjectsCompare`, `ExtractInterface`, `RemoveFooBarUtilsStringFormatted`) are unaffected, since they match against the parsed LST via `MethodMatcher`/`UsesType` rather than a runtime-compiled template.

This should affect any Refaster-generated recipe run through `rewrite-maven-plugin` since 6.37.0. I have not verified `rewrite-gradle-plugin`, but it uses an isolated classloader for recipe artifacts too, so it is likely affected as well.

### Suggested fix

- Restore classloader-aware discovery in `RuntimeClasspathCache` — e.g. walk `URLClassLoader`s up from `JavaParser.class.getClassLoader()` (and Maven's `ClassRealm`), falling back to `java.class.path`, rather than relying on `java.class.path` alone. Whatever #7338 needed can presumably be satisfied without dropping the classloader walk.

It would also be worth surfacing a warning when a `JavaTemplate` stub fails to type-attribute, so this class of failure is loud instead of silent.

timtebeek:
- Fixes #8334

### Problem

- Since #7338 (`8.80.0`) `JavaParser.runtimeClasspath()` derived the classpath solely from the `java.class.path` system property. That property only holds the JVM's *initial* classpath. Inside a Maven JVM that is a single jar:

```
java.class.path=.../plexus-classworlds-2.11.0.jar
```

Everything Maven loads — the plugin, its dependencies, and the recipe artifacts `rewrite-maven-plugin` resolves from `rewrite.recipeArtifactCoordinates` and merges into its `ClassRealm` — lives in a `URLClassLoader` that is invisible to that property.

rewrite-templating emits this for every Refaster `@BeforeTemplate`:

```java
JavaTemplate.builder("com.acme.FooBarUtils.isEmpty(#{value:any(java.lang.String)})")
        .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
        .build();
```

With an empty classpath the stub fails type attribution, `matcher.find()` never matches, and the recipe **silently no-ops**. Paired with a `RemoveDependency` in the same recipe list, the user gets a tree that no longer declares the dependency but still calls into it — output that does not compile, with nothing logged.

### Fix

`RuntimeClasspathCache` now walks the class loader hierarchy of both the thread context class loader and the class loader that loaded `JavaParser`, collecting `file:` URLs from every `URLClassLoader` (plexus' `ClassRealm` and Gradle's loaders both qualify), in addition to `java.class.path` — which is still needed because the application class loader is not a `URLClassLoader` on Java 9+. This restores the pre-`8.80.0` reach of the ClassGraph based discovery without reintroducing the dependency.

Entries are cached per class loader rather than in a single static field, since the context class loader varies across recipe runs.

### Test

`JavaParserTest#runtimeClasspathIncludesIsolatedClassLoaderEntries` puts a jar behind a `URLClassLoader` that is not on `java.class.path` and asserts it shows up in `runtimeClasspath()` and `dependenciesFromClasspath(...)`. Verified red before the change, green after; the rest of `:rewrite-java-test:test` stays green.

Not addressed here: the issue's secondary suggestion to warn when a `JavaTemplate` stub fails to type-attribute, so this class of failure is loud rather than silent.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
@@ -32,6 +32,8 @@ import org.openrewrite.test.RewriteTest;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -182,6 +184,27 @@ class JavaParserTest implements RewriteTest {
         ctx.setParserClasspathDownloadTarget(updatedTemp.toFile());
         ctx.getParserClasspathDownloadTarget();
         assertThat(updatedTemp.toFile().exists()).isTrue();
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite/issues/8334")
+    @Test
+    void runtimeClasspathIncludesIsolatedClassLoaderEntries(@TempDir Path temp) throws Exception {
+        // Build tools load recipe artifacts into a class loader of their own, which is never reflected in `java.class.path`
+        Path recipeJar = Files.createFile(temp.resolve("recipe-artifact-1.0.jar"));
+        ClassLoader originalTccl = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(
+          new URLClassLoader(new URL[]{recipeJar.toUri().toURL()}, originalTccl));
+        try {
+            assertThat(JavaParser.runtimeClasspath())
+              .contains(recipeJar)
+              .containsAll(Stream.of(System.getProperty("java.class.path").split(System.getProperty("path.separator")))
+                .map(Paths::get)
+                .filter(Files::exists)
+                .toList());
+            assertThat(JavaParser.dependenciesFromClasspath("recipe-artifact")).containsExactly(recipeJar);
+        } finally {
+            Thread.currentThread().setContextClassLoader(originalTccl);
+        }
     }
 
     @Issue("https://github.com/openrewrite/rewrite/issues/3222")
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/JavaParserTest.java
@@ -189,7 +189,7 @@ class JavaParserTest implements RewriteTest {
     @Issue("https://github.com/openrewrite/rewrite/issues/8334")
     @Test
     void runtimeClasspathIncludesIsolatedClassLoaderEntries(@TempDir Path temp) throws Exception {
-        // Build tools load recipe artifacts into a class loader of their own, which is never reflected in `java.class.path`
+        // Build tools load recipe artifacts into their own class loader, never reflected in `java.class.path`
         Path recipeJar = Files.createFile(temp.resolve("recipe-artifact-1.0.jar"));
         ClassLoader originalTccl = Thread.currentThread().getContextClassLoader();
         Thread.currentThread().setContextClassLoader(
--- a/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java
@@ -31,6 +31,8 @@ import org.openrewrite.style.NamedStyles;
 import java.io.ByteArrayInputStream;
 import java.lang.reflect.Method;
 import java.net.URI;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -44,6 +46,7 @@ import java.util.stream.Stream;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
+import static java.util.Collections.unmodifiableList;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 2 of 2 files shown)


jkschneider:
Deferring to @sambsnyd 's judgment here.

PR #8337, opened 2026-07-27
Prefer PATH over well-known install directories in GoExecutor.find()

timtebeek:
## Problem

`GoExecutor.find()` resolves `go` by trying fixed absolute paths — `$GOROOT/bin/go`, `/usr/local/go/bin/go`, `/opt/homebrew/bin/go`, `/usr/local/bin/go`, `/usr/bin/go` — and only falls back to `which go` if none exist. On a host with more than one Go install, that picks whichever release happens to sit in a fixed location instead of the one a child process would actually get from `PATH`.

That matters because the resolved path is not just used to run `go`. `GoRewriteRpc.ensureGoRoot` derives `GOROOT` from it and injects it into the `rewrite-go-rpc` child process, and the RPC server's installer then runs a bare `exec.Command("go", "build", ...)` resolved from `PATH`. When the two disagree the toolchain refuses:

```
compile: version "go1.24.13" does not match go tool version "go1.25.12"
```

This is not hypothetical. It silently emptied the Go recipe catalog on docs.moderne.io: GitHub's ubuntu image symlinks its default Go into `/usr/bin`, `actions/setup-go` prepends a newer toolchain to `PATH`, so `find()` returned the 1.24.13 binary while the RPC server built with 1.25.12. Every recipe-module install failed, `GoRecipeLoader` logged a warning and returned an empty list, and the docs job — which replaces the catalog wholesale — deleted all 187 `recipes-go` pages across two green runs before anyone noticed.

## Fix

Reorder to: explicit `GOROOT` → `PATH` → well-known directories.

* `GOROOT` stays first because child processes inherit it, so it is the one setting that keeps parent and child consistent by construction.
* `PATH` now beats the fixed locations, matching what a bare `go` in a child process resolves to.
* The well-known directories remain as a tail fallback, which is what they are actually for — GUI-launched processes (IDE plugins started from Finder or the dock) that have no useful `PATH`.

Scanning `PATH` directly also lets the `which`/`where` subprocess go away, and makes the resolution order a pure function of `(GOROOT, PATH, isExecutable)` so it can be tested.

## Testing

New `GoExecutorTest` covers the ordering with an injected executable-lookup predicate. `pathBeatsWellKnownInstallLocations` reproduces the exact GitHub Actions layout and fails against the previous ordering — I verified that by temporarily restoring the old precedence:

```
GoExecutorTest > pathBeatsWellKnownInstallLocations() FAILED
5 tests completed, 1 failed
```

`:rewrite-go:test --tests "org.openrewrite.golang.internal.*"` and `:rewrite-go:license` pass.

## Risk

Behavior only changes on hosts where `go` resolves differently through `PATH` than through the fixed locations — precisely the case that is broken today. Single-install hosts resolve to the same binary either way. Dropping `which` is not a behavior change: `which` scans `PATH` and does not expand shell aliases or functions.

### DIFF
--- a/rewrite-go/src/main/java/org/openrewrite/golang/internal/GoExecutor.java
+++ b/rewrite-go/src/main/java/org/openrewrite/golang/internal/GoExecutor.java
@@ -112,12 +112,8 @@ public final class GoExecutor {
     }
 
     /**
-     * {@code PATH} is searched ahead of the well-known install directories because callers hand the
-     * resolved binary's {@code GOROOT} to child processes that go on to invoke a bare {@code go} of
-     * their own. Preferring a fixed location such as {@code /usr/bin/go} pairs that child's {@code go}
-     * with a {@code GOROOT} from whichever release happens to sit there, and the toolchain rejects the
-     * mismatch with {@code compile: version "goX" does not match go tool version "goY"}. An explicit
-     * {@code GOROOT} still wins, since child processes inherit it.
+     * {@code PATH} is searched ahead of the well-known install directories so the resolved binary matches the
+     * bare {@code go} that child processes invoke; otherwise the toolchain rejects the {@code GOROOT} mismatch.
      */
     @Nullable String find(@Nullable String goRoot, @Nullable String pathEnv, Predicate<Path> isExecutable) {
         String exe = IS_WINDOWS ? name + ".exe" : name;
--- a/rewrite-go/src/main/java/org/openrewrite/golang/internal/GoExecutor.java
+++ b/rewrite-go/src/main/java/org/openrewrite/golang/internal/GoExecutor.java
@@ -20,6 +20,7 @@ import lombok.Value;
 import org.jspecify.annotations.Nullable;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


rewrite-go/src/main/java/org/openrewrite/golang/internal/GoExecutor.java
```diff
@@ -103,15 +105,32 @@ public RunResult run(Path workDir, String executablePath, Map<String, String> en
...
+    }
 
+    /**
+     * {@code PATH} is searched ahead of the well-known install directories because callers hand the
+     * resolved binary's {@code GOROOT} to child processes that go on to invoke a bare {@code go} of
+     * their own. Preferring a fixed location such as {@code /usr/bin/go} pairs that child's {@code go}
+     * with a {@code GOROOT} from whichever release happens to sit there, and the toolchain rejects the
+     * mismatch with {@code compile: version "goX" does not match go tool version "goY"}. An explicit
+     * {@code GOROOT} still wins, since child processes inherit it.
+     */
```
greg-at-moderne, line 117: This comment is either not needed or too verbose.

greg-at-moderne (approved):

PR #8341, opened 2026-07-29
Python: keep expression/statement wrappers type-correct after visits

### LINKED ISSUE #8322: ClassCastException: Py.Await wrapped in Py.StatementExpression cannot be cast to Statement on RPC receive
## Summary

A `Py.Await` node reaches Java's RPC receiver wrapped in a `Py.StatementExpression`, which is typed to hold a `Statement`. `Py.Await` is an `Expression`, not a `Statement`, so deserialization throws `ClassCastException` and the recipe errors out on the file.

Reproduced against `agronholm/anyio@master` while running `org.openrewrite.python.migrate.UpgradeToPython313`. The error surfaces under `org.openrewrite.python.migrate.ReplaceReTemplate`, but the defect is in the `Py` tree construction / RPC round trip, not in that recipe.

## Stack trace

```
java.lang.ClassCastException: class org.openrewrite.python.tree.Py$Await cannot be cast to
class org.openrewrite.java.tree.Statement (org.openrewrite.python.tree.Py$Await and
org.openrewrite.java.tree.Statement are in unnamed module of loader 'app')
  at org.openrewrite.rpc.RpcReceiveQueue.receive(RpcReceiveQueue.java:137)
  at org.openrewrite.python.internal.rpc.PythonReceiver.visitStatementExpression(PythonReceiver.java:147)
  at org.openrewrite.python.internal.rpc.PythonReceiver.visitStatementExpression(PythonReceiver.java:39)
  at org.openrewrite.python.tree.Py$StatementExpression.acceptPython(Py.java:725)
  at org.openrewrite.python.tree.Py.accept(Py.java:54)
  at org.openrewrite.TreeVisitor.visit(TreeVisitor.java:242)
  at org.openrewrite.python.internal.rpc.PythonReceiver.visit(PythonReceiver.java:45)
  at org.openrewrite.python.internal.rpc.PythonReceiver$PythonReceiverDelegate.visit(PythonReceiver.java:371)
  at org.openrewrite.TreeVisitor.visitNonNull(TreeVisitor.java:167)
  at org.openrewrite.java.internal.rpc.JavaReceiver.lambda$visitRightPadded$195(JavaReceiver.java:661)
  at org.openrewrite.rpc.RpcReceiveQueue.receive(RpcReceiveQueue.java:137)
  at org.openrewrite.java.internal.rpc.JavaReceiver.visitRightPadded(JavaReceiver.java:658)
  at org.openrewrite.java.internal.rpc.JavaReceiver.lambda$visitBlock$26(JavaReceiver.java:117)
  at org.openrewrite.rpc.RpcReceiveQueue.receive(RpcReceiveQueue.java:137)
  at org.openrewrite.rpc.RpcReceiveQueue.receiveList(RpcReceiveQueue.java:183)
  ...
```

## Analysis

The two wrapper types have mirrored contracts:

```java
// Py.java
final class StatementExpression implements Py, Expression, Statement {
    Statement statement;      // <-- holds a Statement
}

final class ExpressionStatement implements Py, Expression, Statement {
    Expression expression;    // <-- holds an Expression
}

final class Await implements Py, Expression { ... }   // Expression only
```

So `StatementExpression(Await)` violates the declared type — an expression used as a statement should be wrapped in `Py.ExpressionStatement`, not `Py.StatementExpression`.

Python's dataclasses don't enforce annotations at runtime, so the Python side constructs the invalid node silently. It only fails when the tree crosses into Java's typed receiver, where `PythonReceiver.visitStatementExpression` does the `Statement` cast.

Worth noting why this stays hidden for most code: many `J` nodes implement **both** `Expression` and `Statement` (`J.MethodInvocation`, `J.Assignment`, …), so wrapping those in a `StatementExpression` casts fine by accident. It only breaks for expression-only node types like `Py.Await`.

## Reproduction

```bash
mod config recipes pip install openrewrite-migrate-python   # 0.10.1

mkdir repos && cd repos
git clone --depth 1 https://github.com/agronholm/anyio.git

MODERNE_WRAPPER_VERSION=4.4.1 mod build .
MODERNE_WRAPPER_VERSION=4.4.1 mod run . --recipe org.openrewrite.python.migrate.UpgradeToPython313
```

2 errors of this kind, both attributed to `ReplaceReTemplate`:

- `tests/test_debugging.py`
- `tests/test_taskgroups.py`

(The same run produces 82 unrelated errors from a separate `AddImport`-on-Python problem — ignore those for this issue.)

## Not yet pinned down

I could not isolate the exact source construct. Bare `await <name>` statements appear in `tests/streams/test_memory.py` and `src/anyio/_backends/_asyncio.py` too, and neither of those files fails, so `await <name>` alone is not sufficient. Neither failing file uses `re` at all, so `ReplaceReTemplate`'s preconditions (`uses_method("*..* template(..)")` / `uses_type("re")`) don't obviously explain the selection either. Narrowing to a minimal snippet is the next step.

A defensive fix on the Java side (clear error naming the offending node type instead of a raw CCE) would make this class of tree-shape bug much cheaper to diagnose.

## Environment

- Moderne CLI 4.4.1
- openrewrite (Python) 8.87.7
- openrewrite-migrate-python 0.10.1

kdelay:
`Py.ExpressionStatement` declares an `Expression` child and `Py.StatementExpression` a `Statement` child, but `PythonVisitor` rebuilt both wrappers with whatever the child visit returned. Python dataclasses do not enforce the annotation, so a visitor replacing the child with a node of the other kind (e.g. migrating `yield from x` to `await x`, where `Py.Await` is an `Expression` but not a `Statement`) silently produced `StatementExpression(Await)`. The invalid tree only failed later, on the Java side of the RPC round trip, where `PythonReceiver.visitStatementExpression` casts the child to `Statement` and throws `ClassCastException`.

## Fix (`visitor.py`)

- `visit_expression_statement` / `visit_statement_expression` now check the visited child's kind. When it no longer fits the wrapper's declared type but fits the opposite one, the wrapper is swapped (`ExpressionStatement` <-> `StatementExpression`) instead of rebuilding an invalid node.
- Nodes implementing both `Expression` and `Statement` keep their original wrapper. Both wrappers delegate prefix and markers to the child, so no formatting is lost.

## Verification

- New regression tests (`tests/python/test_wrapper_normalization.py`) cover both directions plus the no-op cases; before the fix 3 of 4 fail with the invalid wrappers, after the fix all 4 pass.
- Full `rewrite-python` suite green: 1755 passed, 85 skipped.

- Fixes #8322


timtebeek (approved):
Thanks a lot for jumping in here @kdelay ! Unexpected and much appreciated. The change looks good to me, with a little polish we can have this go out today.

timtebeek:
Found some more gaps that require a closer look before we merge, and I'd like to get some colleagues involved as well.

What made you pick this up here? As I can see you're broadly doing contributions across a wide set of projects.

PR #8343, opened 2026-07-29
Surface RPC subprocess stderr instead of discarding it

timtebeek:
## What's changed

When an RPC subprocess dies, `RewriteRpcProcess.getLivenessCheck()` reported only a bare exit code. The stderr that explains *why* it exited was consumed by the drain thread and discarded whenever no `stderrRedirect` was set — which is the default everywhere except the `rewrite-javascript` integration tests that opt in via `JavaScriptRewriteRpc.builder().log(path)`.

Two complementary changes:

1. **Always retain a bounded (8KB) tail** of stderr and append it to the `IllegalStateException` message. The drain thread is briefly joined first, so output still in flight when the process dies isn't missed. No configuration needed — this works everywhere immediately.
2. **`REWRITE_RPC_STDERR=console`** mirrors the full stream to the parent's stderr. The tail only reaches the caller through `getLivenessCheck()`, so it doesn't help when the subprocess logs heavily before dying, or misbehaves without ever exiting. Unlike the existing programmatic `.log(path)` hook, an env var is settable in CI without a code change in the repo under test. Follows the existing `REWRITE_RPC_RSS_MS` opt-in convention.

## Why

This has made a recurring CI failure impossible to attribute. `openrewrite/rewrite-static-analysis` scheduled runs failed on 2026-07-15, 07-16, 07-19 and [07-28](https://github.com/openrewrite/rewrite-static-analysis/actions/runs/30386440190) — 4 of the last 14 days, only on `schedule`, never on push/PR — with ~9 TypeScript tests each reporting:

```
java.lang.IllegalStateException: RPC process shut down early with exit code 217
    at org.openrewrite.rpc.RewriteRpcProcess.getLivenessCheck(RewriteRpcProcess.java:188)
    at org.openrewrite.javascript.rpc.JavaScriptRewriteRpc$Builder$$Lambda/0x…get(Unknown Source)
```

The exit code varies *within a single run* (1, 217, 127), which points at a subprocess startup/environment failure — but the node process's own error message is thrown away, so there is nothing to diagnose from. It does not reproduce locally.

For the record, the obvious suspect was ruled out: this is **not** an npm publish lag. The resolved `rewrite-javascript-8.88.0-20260728.163231-66` jar pins `@openrewrite/rewrite@8.88.0-20260728-114130`, and that version is present on the registry.

This PR doesn't fix the flake — it makes the next occurrence diagnosable instead of a fifth dead end.

## Testing

Four regression tests in `RewriteRpcProcessTest`, all following the existing forked-JVM pattern in that file:

- `livenessCheckReportsSubprocessStderr` — subprocess writes a diagnostic to stderr and halts with 217; the liveness check must surface both.
- `livenessCheckRetainsTailOfOversizeStderr` — subprocess brackets 64KB of filler between two markers; eviction must drop the head and keep the tail, since a stack trace or fatal error lands at the end.
- `mirrorsSubprocessStderrToConsoleWhenEnvVarSet` — forked JVM with the env var set must reproduce the subprocess's stderr on its own.
- `doesNotMirrorSubprocessStderrByDefault` — negative control; without it the previous test would still pass if mirroring were unconditional.

Each fails without its corresponding change and passes with it. The full `org.openrewrite.rpc.*` package passes, including `shutdownWaitsForStderrDrainToComplete`, which covers the Windows file-handle behaviour the restructured drain loop touches.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpcProcess.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpcProcess.java
@@ -132,7 +132,7 @@ public class RewriteRpcProcess extends Thread {
             // parent-side file handle after process termination, preventing deletion
             // of the log file.  Instead we drain stderr in a daemon thread.
             process = pb.start();
-            stderrDrainThread = drainStderr(process, stderrRedirect, stderrTail);
+            stderrDrainThread = drainStderr(process, stderrRedirect, stderrTail, mirrorStderrToConsole());
         } catch (IOException e) {
             // Record the failure so start() can surface it instead of busy-waiting forever
             // on `process == null`. Throwing here would just kill this thread silently.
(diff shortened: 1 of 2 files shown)


timtebeek:
Not sure we'd want to go this route, so closing for now.

PR #8350, opened 2026-07-29
Add `UpgradePluginsForMaven4` to bump plugins that fail under Maven 4

timtebeek:
## What's changed

`MigrateToMaven4` migrated the POM *model* (model version, phases, expressions, `<subprojects>`) but left plugins pinned at versions that don't run under Maven 4. This adds `org.openrewrite.maven.UpgradePluginsForMaven4` and wires it into `MigrateToMaven4`.

The floors come from `PluginUpgradeStrategy` in Apache's own [`mvnup upgrade --plugins`](https://maven.apache.org/tools/mvnup.html) ([source](https://github.com/apache/maven/blob/maven-4.0.x/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/mvnup/goals/PluginUpgradeStrategy.java)), as of the 4.0.0-rc-6 milestone:

| plugin | range | why the floor |
|---|---|---|
| `maven-compiler-plugin` | `[3.11.0,4.0.0)` | can't find ErrorProne under Maven 4 classloading |
| `maven-surefire`/`failsafe`/`surefire-report-plugin` | `[3.5.2,4.0.0)` | Maven 4 compatibility |
| `maven-clean`/`enforcer`/`jar`/`shade-plugin` | `[3.5.0,4.0.0)` | Maven 4 compatibility / pre-release 4.x API signatures |
| `maven-install`/`deploy-plugin` | `[3.1.4,4.0.0)` | pre-release 4.x API signatures |
| `maven-resources-plugin` | `[3.3.1,4.0.0)` | pre-release 4.x API signatures |
| `maven-remote-resources-plugin` | `[3.0.0,4.0.0)` | Maven 4 compatibility |
| `exec-maven-plugin` | `[3.5.0,4.0.0)` | Maven 4 compatibility |
| `flatten-maven-plugin` | `[1.2.7,2.0.0)` | Maven 4 compatibility |
| `jaxb2-maven-plugin` | `[3.2.0,4.0.0)` | `jaxb-parent:3.0.0` ships XML that Maven 4 rejects |
| `scala-maven-plugin` | `[4.9.5,5.0.0)` | mutates immutable lists returned by the Maven 4 API |
| `gmavenplus-plugin` | `[4.2.0,5.0.0)` | same |
| `quarkus-maven-plugin` (`io.quarkus` + `io.quarkus.platform`) | `3.26.0` exact | Aether API changes |

Plus the coordinate migration mvnup performs: `org.scala-tools:maven-scala-plugin` → `net.alchim31.maven:scala-maven-plugin:4.9.5` (unmaintained since 2011).

## Version selection

Each entry is a bounded set range: **at least the version mvnup requires, any newer release within the same major**. `SetRange` treats the `)` upper bound as strictly exclusive and `LatestRelease.isValid` filters pre-releases, so a future stable `maven-compiler-plugin:4.0.0` won't be selected and the in-flight `4.0.0-beta-*` line was never a candidate.

Note this is deliberately broader than mvnup, which only ever moves a POM up to the floor: a plugin already above the floor is also carried to the latest release in its major line.

`quarkus-maven-plugin` is the one exception, held at an exact floor. Its version is coupled to the Quarkus platform BOM, so tracking latest — even within 3.x — would bump the user's Quarkus release out from under their `quarkus-bom` rather than just making the build Maven 4-compatible. It keeps mvnup's semantics exactly: below 3.26.0 moves up, above is untouched.

Version-less plugins are left alone, matching mvnup (`upgradePluginVersion` skips a plugin with no `<version>` element). Adding versions everywhere is a separate concern already served by `cleanup.ExplicitPluginVersion` in `BestPractices` / `ReproducibleBuilds`.

Both `build/plugins` and `build/pluginManagement/plugins` are covered, as are versions held in a property.

## Known divergence from mvnup

mvnup's `PluginUpgrade` record carries an optional `latestPreRelease` field, used to also upgrade off Maven-4 *pre-release* lines of a plugin (e.g. `maven-compiler-plugin:4.0.0-beta-4`). A range can't express that, so a POM sitting on a 4.x plugin pre-release is left alone here. Narrow enough to be worth the simplicity; happy to revisit if it comes up.

mvnup additionally bumps the `extra-enforcer-rules` **plugin dependency** to 1.4. That needs different machinery than `UpgradePluginVersion`, so it is out of scope here.

## Tests

New `UpgradePluginsForMaven4Test` covers the upgrade, upgrading from the floor, version-less plugins, `pluginManagement`, property-held versions, the scala-tools migration, and both quarkus directions. Because the selected version tracks the latest release, tests capture whichever version was chosen and assert it satisfies the declared range rather than hardcoding it.

Two `MigrateToMaven4Test` fixtures moved off `maven-compiler-plugin` onto `maven-assembly-plugin`, which isn't in the table. Those tests are about the model/phase/property migration and duplicate-plugin dedup, so keeping them deterministic preserves what they actually assert — `noChangesWhenAlreadyMaven4` in particular would otherwise have stopped testing "no changes" at all.

Full `:rewrite-maven:test` is green; `recipes.csv` regenerated via `:rewrite-maven:recipeCsvGenerate`.


timtebeek:
Switched the floors to open-ended set ranges (`[3.11.0,)` etc.) per your comment, so each entry now means "at least this, but take newer." Pushed in 16baeedb91.

One entry I deliberately left as an exact floor: **`quarkus-maven-plugin`** (both `io.quarkus` and `io.quarkus.platform`). Its version is coupled to the Quarkus platform BOM, so tracking latest here would silently bump a user's Quarkus release out from under their `quarkus-bom` rather than just making the build Maven 4-compatible. It keeps mvnup's exact semantics: a POM below 3.26.0 moves up to 3.26.0, one already above is untouched. Say the word if you'd rather it track latest too.

Worth flagging one behavioural consequence: a plugin that was *already* above the floor now also moves to the latest release, where before it was left alone. That's the intent of "allow newer", but it does mean the recipe touches POMs that were already Maven 4-clean.

Test fallout from making the output non-deterministic:
- `UpgradePluginsForMaven4Test` now captures whichever version was selected and asserts it satisfies the declared range, rather than hardcoding it.
- The two `MigrateToMaven4Test` fixtures moved off `maven-compiler-plugin` to `maven-assembly-plugin`, which isn't in the table. Those tests are about the model/phase/property migration and the duplicate-plugin dedup, so they stay deterministic and keep asserting what they mean to; `UpgradePluginsForMaven4Test` owns the upgrade behaviour.

timtebeek:
Bounded every range to its current major in c57bf7d034 — e.g. `[3.11.0,4.0.0)` for `maven-compiler-plugin`, `[4.2.0,5.0.0)` for `gmavenplus-plugin`, `[1.2.7,2.0.0)` for `flatten-maven-plugin`, `[4.9.5,5.0.0)` for `scala-maven-plugin`.

`SetRange` treats the `)` upper bound as exclusive, and `LatestRelease.isValid` already filters pre-releases, so a future stable `maven-compiler-plugin:4.0.0` won't be selected and the in-flight `4.0.0-beta-*` line was never a candidate either.

Test ranges updated to match, so they assert the real selector rather than a looser one. The already-at-the-floor test uses an exclusive lower bound `(3.11.0,4.0.0)` to prove the version actually moved past the floor and stayed inside the major.

PR #8358, opened 2026-07-31
Fix OneDependencyDeclarationPerStatement mangling DependencyHandler.add() calls

sullis:
## What's changed

Three fixes in the Gradle 9 best-practice recipes.

### `OneDependencyDeclarationPerStatement` corrupted `DependencyHandler.add(...)` calls

`add(configuration, notation)` (and the `addProvider` / `addProviderConvertible` variants) take the *configuration name* as their first argument, not a coordinate. The recipe treated every argument as a coordinate and split the call, turning:

```groovy
dependencies {
    add('testImplementation', platform("org.junit:junit-bom:6.1.2"))
}
```

into two broken statements. These method names are now skipped.

### Duplicate LST ids when splitting a multi-coordinate declaration

Every split statement reused the original `J.MethodInvocation`'s id, leaving duplicate ids in the enclosing block that downstream recipes choke on. Each statement after the first now gets a fresh `Tree.randomId()`.

### `UseJavaExtensionBlock` unchanged-detection never fired

`getStatements()` builds a new list on each call, so `mapped == cu.getStatements()` compared against a freshly-allocated list and was always false. The original list is now held in a local and compared against that.

## Testing

Added `leavesDependencyHandlerAddAlone` to `OneDependencyDeclarationPerStatementTest` and `noChangeToDependencyHandlerAddCalls` to `GradleBestPracticesTest`. `./gradlew :rewrite-gradle:test` passes for both classes.

### DIFF
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/OneDependencyDeclarationPerStatement.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/OneDependencyDeclarationPerStatement.java
@@ -95,6 +95,12 @@ public class OneDependencyDeclarationPerStatement extends Recipe {
                     if (m.getSelect() != null) {
                         return s;
                     }
+                    // `DependencyHandler.add(configuration, notation)` and its provider variants take the
+                    // configuration name as their first argument, so their arguments are not all coordinates
+                    String methodName = m.getSimpleName();
+                    if ("add".equals(methodName) || "addProvider".equals(methodName) || "addProviderConvertible".equals(methodName)) {
+                        return s;
+                    }
                     List<Expression> args = m.getArguments();
                     if (args.size() < 2) {
                         return s;
@@ -122,13 +128,15 @@ public class OneDependencyDeclarationPerStatement extends Recipe {
                     List<Statement> split = new ArrayList<>(args.size());
                     for (int i = 0; i < args.size(); i++) {
                         Expression coord = args.get(i).withPrefix(Space.format(" "));
+                        // Every split beyond the first needs its own id; reusing the original invocation's
+                        // id leaves duplicate ids in the enclosing block, which later recipes choke on.
+                        J.MethodInvocation mi = (i == 0 ? m : m.withId(Tree.randomId()))
+                                .withArguments(singletonList(coord));
                         boolean isLast = i == args.size() - 1;
                         if (isLast && wrappedInReturn) {
-                            J.MethodInvocation innerMi = m.withArguments(singletonList(coord));
-                            split.add(((J.Return) s).withPrefix(subsequentPrefix).withExpression(innerMi));
+                            split.add(((J.Return) s).withPrefix(subsequentPrefix).withExpression(mi));
                         } else {
-                            Space prefix = i == 0 ? stmtPrefix : subsequentPrefix;
-                            split.add(m.withPrefix(prefix).withArguments(singletonList(coord)));
+                            split.add(mi.withPrefix(i == 0 ? stmtPrefix : subsequentPrefix));
                         }
                     }
                     return split;
--- a/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/OneDependencyDeclarationPerStatement.java
+++ b/rewrite-gradle/src/main/java/org/openrewrite/gradle/gradle9/OneDependencyDeclarationPerStatement.java
@@ -44,23 +44,17 @@ import static java.util.Collections.singletonList;
 @EqualsAndHashCode(callSuper = false)
 public class OneDependencyDeclarationPerStatement extends Recipe {
 
-    @Override
-    public String getDisplayName() {
-        return "Use one dependency declaration per statement";
-    }
+    final String displayName = "Use one dependency declaration per statement";
 
-    @Override
-    public String getDescription() {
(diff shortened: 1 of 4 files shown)


kdelay (commented):
I went through this locally at head `1bcc0309` (macOS, JDK 21 toolchain, `./gradlew :rewrite-gradle:test`), and reverted each of the three hunks one at a time to see which tests actually pin them. Two hold up; the LST id one is both untested and, as far as I can measure, incomplete.

### 1. The `add` / `addProvider` / `addProviderConvertible` guard — covered

Dropping just that guard fails exactly the two new tests (`OneDependencyDeclarationPerStatementTest.leavesDependencyHandlerAddAlone`, `GradleBestPracticesTest.noChangeToDependencyHandlerAddCalls`). Diagnosis matches the code: `isMultiComponentLiterals` only bails when *every* argument is a string literal, so `add('testImplementation', platform("..."))` slipped past it because the second argument is a `J.MethodInvocation`.

### 2. `UseJavaExtensionBlock` identity check — real, but pinned in the wrong place

The mechanism checks out: `JRightPadded.getElements` allocates a new `ArrayList` per call (`J.java:819-821`, `G.java:170-172`), so `mapped == body.getStatements()` never fired and `visitMethodInvocation` fell through to `autoFormat(...)` for *any* `subprojects`/`allprojects` block, whether or not a compatibility assignment moved.

Reverting only this hunk fails `GradleBestPracticesTest.noChangeToDependencyHandlerAddCalls` — that test's build script happens to use irregular indentation inside `subprojects`, so the stray `autoFormat` shows up as a diff. All 18 tests in `UseJavaExtensionBlockTest` (including the `NoChange` nested class) still pass. So the coverage is incidental and lives in a test named after the `add()` bug. A `UseJavaExtensionBlockTest$NoChange` case with an oddly indented `subprojects { plugins.withId('java') { ... } }` and no `sourceCompatibility`/`targetCompatibility` would pin it where someone would look for it.

### 3. The fresh-id hunk — nothing pins it, and it only reaches the top node

Two separate findings.

**Nothing fails without it.** With only that hunk reverted (both other fixes in place), the entire `:rewrite-gradle:test` module run is green (~14 min). No existing assertion looks at ids.

**It refreshes the invocation but not its children.** I walked the after-tree and collected duplicate ids for a three-coordinate declaration:

```java
// afterRecipe on buildGradle(...) for: implementation 'a:b:1.0', 'c:d:2.0', 'e:f:3.0'
Set<UUID> seen = new HashSet<>();
List<String> duplicates = new ArrayList<>();
new GroovyIsoVisitor<Integer>() {
    @Override
    public @Nullable J visit(@Nullable Tree tree, Integer p) {
        if (tree instanceof J && !seen.add(tree.getId())) {
            duplicates.add(tree.getClass().getSimpleName() + " -> " + tree);
        }
        return super.visit(tree, p);
    }
}.visit(cu, 0);
assertThat(duplicates).isEmpty();
```

- hunk reverted: 4 duplicates — 2 `J.MethodInvocation` plus 2 `J.Identifier`
- this PR's head: 2 duplicates — `["Identifier -> implementation", "Identifier -> implementation"]`

`m.withId(Tree.randomId())` copies the invocation node only; the `implementation` name identifier (and anything else under `m` that is carried over rather than replaced) is still shared across all splits. If duplicate ids in the enclosing block are the actual problem, this leaves the same class of problem behind.

Swapping the whole subtree instead makes it 0 — same idiom `JavaTemplateParser` uses for exactly this (`JavaTemplateParser.java:391`):

```java
J.MethodInvocation mi = (i == 0 ? m : (J.MethodInvocation) new RandomizeIdVisitor<Integer>().visit(m, 0))
        .withArguments(singletonList(coord));
```

With that plus the probe above, `gradle9.*` and `GradleBestPracticesTest` are green: 79 tests (78 repo tests + the probe), 0 failures.

**One thing I could not reproduce.** I tried to turn "downstream recipes choke on it" into a failing test and did not manage to. Chaining `OneDependencyDeclarationPerStatement` then `DependencyUseMapNotation` over a two-coordinate declaration produced the expected map-notation output both with and without the hunk. I also tried a `RemoveDependency("c", "d", null)` chain, but that is inconclusive: the control run without the split recipe (three already-separate `implementation` lines) does not remove the dependency either in this test setup, so nothing follows from it. Do you have a concrete recipe that fails on the duplicates? Either way an id-uniqueness assertion is worth having, since right now the hunk can be deleted without any test noticing.

### Optional: the guard could be shape-based rather than a name list

`add`, `addProvider` and `addProviderConvertible` share one shape — a leading string literal that is a configuration name, so it cannot contain `:`. Keying on that covers those three plus any project-local helper with the same signature:

```java
List<Expression> args = m.getArguments();
if (args.size() < 2) {
    return s;
}
// A leading string literal with no ':' is a configuration name, not a coordinate
if (args.get(0) instanceof J.Literal &&
        ((J.Literal) args.get(0)).getValue() instanceof String &&
        !((String) ((J.Literal) args.get(0)).getValue()).contains(":")) {
    return s;
}
```

I ran this in place of the name check and `gradle9.*` plus `GradleBestPracticesTest` stay green, including both of your new tests. It also overlaps with `isMultiComponentLiterals`, which already applies the no-colon-first-literal rule to the all-literals case. Your call whether that is in scope here — the name list is not wrong, just narrower.

timtebeek (approved):
Thanks again @sullis ! I've applied a slight bit of polish and will merge soon after.

PR #8364, opened 2026-08-03
Native Node.js lock file regeneration (npm, pnpm, yarn, bun)

jkschneider:
Replaces the package manager shell-out for lock regeneration with a native Java implementation covering `package-lock.json` (v2/v3), `pnpm-lock.yaml` (v6/v9), `yarn.lock` (classic and Berry), and `bun.lock`. Shelling out fails at fleet scale for reasons unrelated to the edit (no package manager on the runner, registry credentials not inherited into the scratch directory), and when it fails the edited `package.json` and the stale lock go out of sync. Design notes in `doc/adr/0011-native-node-lock-regeneration.md`.

The engine resolves the new dependency set and edits the lock file in place. An edit that provably doesn't change the resolved closure is patched directly. Everything else resolves the whole closure seeded by the existing lock — a locked version that still satisfies its range is kept, which is what an incremental install does — and the resolved graph is diffed against the lock, so only entries that actually changed get rewritten. Anything the engine can't reproduce byte-for-byte fails loudly with a structured per-package failure instead of writing a lock. Failures land in a `NodeLockRegenerationFailures` data table.

Registry access goes through the run's `HttpSender`, with credentials discovered from `.npmrc`/`.yarnrc.yml`/pnpm config. npm range semantics live in a new node-semver implementation under `org.openrewrite.semver`, kept separate from the Maven version chain. Yarn Berry checksums are reproduced from the tarball bytes rather than copied.

Fixtures are recorded from real package manager output (npm 11.6.2, pnpm 11.2.2, yarn 1.22.22 and 4.5.3, bun 1.3.10) and replayed offline against a stub HTTP sender. They live in [openrewrite/rewrite-javascript-lock-fixtures](https://github.com/openrewrite/rewrite-javascript-lock-fixtures) (~960 files); the build fetches that repository's latest main onto the test classpath, keeping them out of this diff. The whole-closure fixtures are recorded in two phases: install the pre-edit manifest to get the before lock, then install the edited manifest in the same directory, so the goldens are what an incremental install actually writes. A sweep over a corpus of real repositories produced zero wrong locks; shapes the engine doesn't support defer with structured failures.


jkschneider:
The `NodeVersion`/`NodeComparator`/`NodeRange`/`NodeSemver` classes in `org.openrewrite.semver` are slated for removal: #8366 catalogs what the lock engine needs from the existing `VersionComparator` family, and once that lands this branch rebases onto it and deletes the parallel classes.

jkschneider:
Still needs some work to remove a lot of the manual extraneous *Patcher implementations.

PR #8370, opened 2026-08-03
Emit listing-weight GetMarketplace rows instead of full descriptors

bmuschko:
## Problem

`GetMarketplace` shipped a full, recursive `RecipeDescriptor` for every recipe, but the host only uses lightweight `RecipeListing` metadata. The nested `recipeList` was serialized, transmitted, and held in memory only to be counted into an integer and then discarded — ~82× more nodes than listings (tens of MB vs ~717 KB for an 846-recipe marketplace), hitting the non-JVM engines (Go/Python/JS/C#) hardest. The full descriptor is already fetched lazily via `PrepareRecipe`, so the eager copy was pure redundancy.

## Solution

Emit listing-weight rows — `name`, `displayName`, `description`, `estimatedEffortPerOccurrence`, `options`, `dataTables`, and `recipeCount` (the collapsed transitive `recipeList`) — and stop shipping the descriptor. `describe()`/`prepare()` stay lazy over `PrepareRecipe`.

Backward compatible and additive: `descriptor` is now nullable, `toMarketplace` prefers the lightweight fields and falls back to `fromDescriptor` when `name == null`, and unknown fields are ignored so old/new peers interoperate (host must be at least as new as the engine). Applied in `rewrite-core` plus the `GetMarketplace` handlers for `rewrite-go`, `rewrite-python`, `rewrite-javascript`, and `rewrite-csharp`.

Verified with unit tests (emit/consume, round-trip, descriptor fallback, per-engine transitive count) and cross-process Python→Java and JS→Java round-trips.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
@@ -17,24 +17,58 @@ package org.openrewrite.rpc.request;
 
 import lombok.Value;
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.NlsRewrite;
 import org.openrewrite.config.CategoryDescriptor;
+import org.openrewrite.config.DataTableDescriptor;
+import org.openrewrite.config.OptionDescriptor;
 import org.openrewrite.config.RecipeDescriptor;
 import org.openrewrite.marketplace.RecipeBundle;
 import org.openrewrite.marketplace.RecipeBundleResolver;
 import org.openrewrite.marketplace.RecipeListing;
 import org.openrewrite.marketplace.RecipeMarketplace;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import static java.util.Collections.emptyList;
 import static java.util.Collections.emptySet;
 
 public class GetMarketplaceResponse extends ArrayList<GetMarketplaceResponse.Row> {
     @Value
     public static class Row {
-        RecipeDescriptor descriptor;
+        /**
+         * Listing-weight fields. Emitters populate these and leave {@link #descriptor} null; the full
+         * descriptor is fetched lazily via {@code PrepareRecipe}.
+         */
+        @Nullable String name;
+
+        @Nullable @NlsRewrite.DisplayName String displayName;
+
+        @Nullable @NlsRewrite.Description String description;
+
+        @Nullable Duration estimatedEffortPerOccurrence;
+
+        @Nullable List<OptionDescriptor> options;
+
+        @Nullable List<DataTableDescriptor> dataTables;
+
+        /**
+         * The count of this recipe plus all of its transitive sub-recipes. Emitters must compute it
+         * as {@code 1 + all transitive recipeList entries}; the host uses it as a sort key (see
+         * {@link RecipeListing#getRecipeCount()}).
+         */
+        int recipeCount;
+
+        /**
+         * The full, recursive recipe descriptor. Retained for backward compatibility with peers that
+         * still emit it, but nullable and no longer emitted by up-to-date engines — prefer the
+         * lightweight fields above. Will be removed once nothing emits it.
+         */
+        @Nullable RecipeDescriptor descriptor;
+
         List<List<CategoryDescriptor>> categoryPaths;
 
         /**
@@ -57,25 +91,50 @@ public class GetMarketplaceResponse extends ArrayList<GetMarketplaceResponse.Row
                     !recipe.getPackageName().equals(bundle.getPackageName())) {
                 continue;
             }
+            RecipeListing listing = toListing(recipe, bundle);
             for (List<CategoryDescriptor> categoryPath : recipe.getCategoryPaths()) {
-                marketplace.install(RecipeListing.fromDescriptor(recipe.getDescriptor(), bundle), categoryPath);
+                marketplace.install(listing, categoryPath);
             }
         }
         return marketplace;
     }
 
-    public static GetMarketplaceResponse fromMarketplace(RecipeMarketplace marketplace, List<RecipeBundleResolver> resolvers) {
+    /**
+     * The {@link RecipeListing} a row represents, attributed to {@code bundle}. Prefers the
+     * lightweight fields; a null {@link Row#name} means an older peer sent only {@link Row#descriptor},
+     * so fall back to deriving the listing from it.
+     */
+    private static RecipeListing toListing(Row row, RecipeBundle bundle) {
+        return row.getName() != null ?
+                new RecipeListing(null, row.getName(), row.getDisplayName(), row.getDescription(),
+                        row.getEstimatedEffortPerOccurrence(),
+                        row.getOptions() != null ? row.getOptions() : emptyList(),
+                        row.getDataTables() != null ? row.getDataTables() : emptyList(),
+                        row.getRecipeCount(), bundle) :
+                RecipeListing.fromDescriptor(row.getDescriptor(), bundle);
+    }
+
+    public static GetMarketplaceResponse fromMarketplace(RecipeMarketplace marketplace) {
         Map<String, Row> rowByRecipeId = new LinkedHashMap<>();
         for (RecipeMarketplace.Category category : marketplace.getCategories()) {
-            fromCategory(resolvers, rowByRecipeId, category, new ArrayList<>());
+            fromCategory(rowByRecipeId, category, new ArrayList<>());
         }
         GetMarketplaceResponse response = new GetMarketplaceResponse();
         response.addAll(rowByRecipeId.values());
         return response;
     }
 
-    private static void fromCategory(List<RecipeBundleResolver> resolvers,
-                                     Map<String, Row> rowByRecipeId,
+    /**
+     * @deprecated Use {@link #fromMarketplace(RecipeMarketplace)}. The {@code resolvers} argument is
+     * no longer needed: rows are emitted listing-weight straight off each {@link RecipeListing}, so no
+     * recipe is resolved or described here.
+     */
+    @Deprecated
+    public static GetMarketplaceResponse fromMarketplace(RecipeMarketplace marketplace, List<RecipeBundleResolver> resolvers) {
+        return fromMarketplace(marketplace);
+    }
+
+    private static void fromCategory(Map<String, Row> rowByRecipeId,
                                      RecipeMarketplace.Category category,
                                      List<CategoryDescriptor> parentCategory) {
         List<CategoryDescriptor> categoryPath = new ArrayList<>(parentCategory);
@@ -83,11 +142,13 @@ public class GetMarketplaceResponse extends ArrayList<GetMarketplaceResponse.Row
                 category.getDescription(), emptySet(), false, 0, false));
         for (RecipeListing recipe : category.getRecipes()) {
             rowByRecipeId.computeIfAbsent(recipe.getName(), recipeId ->
-                    new Row(recipe.describe(resolvers), new ArrayList<>(),
-                            recipe.getBundle().getPackageName())).categoryPaths.add(categoryPath);
+                    new Row(recipe.getName(), recipe.getDisplayName(), recipe.getDescription(),
+                            recipe.getEstimatedEffortPerOccurrence(), recipe.getOptions(),
+                            recipe.getDataTables(), recipe.getRecipeCount(), null,
+                            new ArrayList<>(), recipe.getBundle().getPackageName())).categoryPaths.add(categoryPath);
         }
         for (RecipeMarketplace.Category child : category.getCategories()) {
-            fromCategory(resolvers, rowByRecipeId, child, categoryPath);
+            fromCategory(rowByRecipeId, child, categoryPath);
         }
     }
 }
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
@@ -21,7 +21,6 @@ import org.openrewrite.NlsRewrite;
 import org.openrewrite.config.CategoryDescriptor;
 import org.openrewrite.config.DataTableDescriptor;
 import org.openrewrite.config.OptionDescriptor;
-import org.openrewrite.config.RecipeDescriptor;
 import org.openrewrite.marketplace.RecipeBundle;
 import org.openrewrite.marketplace.RecipeBundleResolver;
 import org.openrewrite.marketplace.RecipeListing;
@@ -40,8 +39,8 @@ public class GetMarketplaceResponse extends ArrayList<GetMarketplaceResponse.Row
     @Value
     public static class Row {
         /**
-         * Listing-weight fields. Emitters populate these and leave {@link #descriptor} null; the full
-         * descriptor is fetched lazily via {@code PrepareRecipe}.
+         * Listing-weight fields the host builds a {@link RecipeListing} from. The full descriptor is
+         * fetched lazily per recipe via {@code PrepareRecipe} when detail or execution is needed.
          */
         @Nullable String name;
 
@@ -61,13 +60,6 @@ public class GetMarketplaceResponse extends ArrayList<GetMarketplaceResponse.Row
          * {@link RecipeListing#getRecipeCount()}).
          */
         int recipeCount;
-
-        /**
-         * The full, recursive recipe descriptor. Retained for backward compatibility with peers that
-         * still emit it, but nullable and no longer emitted by up-to-date engines — prefer the
-         * lightweight fields above. Will be removed once nothing emits it.
-         */
-        @Nullable RecipeDescriptor descriptor;
 
         List<List<CategoryDescriptor>> categoryPaths;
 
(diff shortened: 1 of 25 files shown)


rewrite-go/cmd/rpc/main.go
```diff
@@ -1473,11 +1487,16 @@ func (s *server) handleGetMarketplace(params json.RawMessage) (any, *rpcError) {
...
 			packageName = &origin
 		}
 		rows = append(rows, marketplaceRow{
-			Descriptor:    marketplaceDescriptorFromRecipe(desc),
+			Name:          listing.Name,
+			DisplayName:   listing.DisplayName,
+			Description:   listing.Description,
+			Options:       marketplaceOptions(listing.Options),
+			DataTables:    marketplaceDataTables(listing.DataTables),
+			RecipeCount:   listing.RecipeCount,
```
shanman190, line 1499: `EstimatedEffortPerOccurrence` is declared on `marketplaceRow` and populated on `recipe.RecipeListing` by `ToListing`, but this row literal never sets it - so Go always emits null. C#, JS and Python all populate it.

Java reads a bare number as seconds and also accepts an ISO-8601 string, so either wire shape works. Emitting `listing.EstimatedEffortPerOccurrence.Seconds()` as a number is the closest match to what Java itself writes.

rewrite-go/cmd/rpc/main.go
```diff
@@ -1408,8 +1408,19 @@ func parseOrderedColumns(raw json.RawMessage) ([]recipe.ColumnValue, error) {
 	return cols, nil
 }
 
+// marketplaceRow carries listing-weight metadata only: the host builds a RecipeListing from these
+// fields without materializing recipes. The full recursive descriptor is deliberately not shipped —
+// the host fetches it lazily per recipe via the separate PrepareRecipe RPC. RecipeCount collapses
+// the transitive recipeList the host would otherwise only count.
 type marketplaceRow struct {
-	Descriptor    marketplaceDescriptor   `json:"descriptor"`
+	Name                         string                  `json:"name"`
```
shanman190, line 1416: gofmt: the mid-struct comment above `RecipeCount` splits the tag-alignment run, so these six fields end up over-padded relative to what gofmt wants. `gofmt -l ./cmd ./pkg` flags this file.

Heads up that `main.go` was already gofmt-dirty on `main` (an unrelated `pendingDependencyTypes` line), so CI may not fail on it - worth running `gofmt -w` over both.

rewrite-javascript/rewrite/src/rpc/request/get-marketplace.ts
```diff
@@ -65,7 +96,13 @@ export class GetMarketplace {
                                 existing.categoryPaths.push(currentPath);
                             } else {
                                 rowByRecipeId.set(recipe.name, {
-                                    descriptor: recipe,
+                                    name: recipe.name,
+                                    displayName: recipe.displayName,
+                                    description: recipe.description,
+                                    estimatedEffortPerOccurrence: recipe.estimatedEffortPerOccurrence,
```
shanman190, line 102: Unit mismatch with the Java peer, in both directions.

The intent is unambiguously minutes: Java's `Recipe.getEstimatedEffortPerOccurrence()` defaults to `Duration.ofMinutes(5)` (`Recipe.java:213-215`), JS defaults to `5` typed `Minutes` (`recipe.ts:66`), Python to `5` (`recipe.py:224`). The same 5 minutes in all three.

But Java's Jackson `Duration` deserializer reads a bare number as **seconds**. Probed against a replica of `JsonMessageFormatter`'s mapper:

```
Java WRITES PT5M as: {"estimatedEffortPerOccurrence":300.000000000}
Java READS 5        as: PT5S
Java READS 300      as: PT5M
Java READS "PT5M"   as: PT5M
```

Nothing converts on the JS path - `recipe.ts:98`, this line, and `rowToListing` (line 53) all pass the raw number through - so a default JS recipe puts `5` on the wire and the host records 5 seconds.

The reverse is real too, not just theoretical: `rewrite-rpc.ts:322-324` makes JS a GetMarketplace *client*, so a Java engine emitting `300.000000000` is read here as 300 minutes.

Convert at the wire boundary in both directions - seconds as a number, or an ISO-8601 string like the C# side emits (Java accepts either on read).

Worth being explicit that this is pre-existing, not something this PR breaks: the same `5` previously rode inside `Row.descriptor.estimatedEffortPerOccurrence` and `fromDescriptor` copied it into the listing with the same 60x error. Flagging it because this PR promotes the field to a named element of the contract, which makes it the natural moment to pin the unit. `markers.ts:150` shows the codebase already knows how to be explicit where it matters - `estimatedEffortPerOccurrenceMillis`, tested with `300000`.

rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -1440,17 +1440,26 @@ def _collect_marketplace_rows(
...
             else:
+                # Serve the RecipeListing the marketplace already holds: recipeCount was computed
+                # once at install time, so listing never walks the descriptor tree. The host fetches
+                # the full descriptor lazily via PrepareRecipe.
                 rows.append({
-                    'descriptor': _recipe_descriptor_to_dict(recipe_desc),
+                    'name': listing.name,
+                    'displayName': listing.display_name,
+                    'description': listing.description,
+                    'estimatedEffortPerOccurrence': listing.estimated_effort_per_occurrence,
```
shanman190, line 1457: Same unit mismatch as the JS emitter, and pre-existing for the same reason.

`estimated_effort_per_occurrence: int` is minutes (`recipe.py:106`, default `5` at `recipe.py:224`, mirroring Java's `Duration.ofMinutes(5)`), but Java reads a bare number as seconds - so a 5-minute recipe arrives on the host as `PT5S`. Emit seconds, or an ISO-8601 string like the C# side does; Java accepts either on read.

rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -1440,17 +1440,26 @@ def _collect_marketplace_rows(
...
+                    'options': _options_to_dicts(listing),
+                    'dataTables': listing.data_tables,
```
shanman190, line 1459: Raising this because the C# half of this PR fixes the identical hazard and this half does not.

`DataTableDescriptorDto`/`ColumnDescriptorDto` are new here, and `ColumnDescriptorDto` deliberately emits `Type = ""` with the comment *"the wire field (which the Java peer's ColumnDescriptor treats as non-null) is present"*, plus a test asserting it. That reading is right: `ColumnDescriptor.type` has no `@Nullable` and is `@EqualsAndHashCode.Include`.

Python's `DataTable.descriptor()` (`data_table.py:330-352`) emits columns as `{name, displayName, description}` with no `type`, and this line passes those dicts through untouched. Jackson does not enforce non-null on a plain `String` creator param, so it does not throw - the host just ends up with `ColumnDescriptor.type == null` feeding `hashCode()`.

Pre-existing (the same dicts previously rode inside the nested descriptor), and arguably a follow-up. But the fix is the same one-line default the C# side just took, and leaving the two halves inconsistent right after documenting the invariant seems worse than closing it here.

The enclosing `DataTableDescriptor` is fine either way - its `instanceName` and `group` are both `@Nullable`, so their absence is harmless.

rewrite-csharp/csharp/OpenRewrite/Core/RecipeMarketplace.cs
```diff
@@ -181,21 +181,25 @@ public class Category
...
+        public IReadOnlyDictionary<string, (RecipeListing Listing, Recipe? Recipe)> Recipes => _recipes;
 
         internal void Install(Recipe recipe, ReadOnlySpan<CategoryDescriptor> categoryPath)
         {
             if (categoryPath.IsEmpty)
             {
-                _recipes[recipe.GetDescriptor()] = recipe;
+                // Derive the listing once at install; retain the instance for PrepareRecipe.
+                var descriptor = recipe.GetDescriptor();
+                _recipes[descriptor.Name] = (RecipeListing.From(descriptor), recipe);
```
shanman190, line 202: This flips duplicate-name handling from first-wins to last-wins. Previously the dictionary was keyed by the `RecipeDescriptor` record, so two recipes sharing a name produced two entries and `FindRecipe`'s enumeration returned the first; now the second install overwrites the first.

First-wins is the contract in every other implementation:
- Python uses `setdefault` and documents why ("First-wins must match the host's name-keyed RecipeListing and RecipeAttribution"), guarded by `test_marketplace_first_wins.py`
- Java's `Category.findRecipe` returns the first list entry, and `getAllRecipes`' `TreeSet` keeps the first
- The JS emitter keeps the first row's fields and merges later category paths into it

`TryAdd` restores parity here, and the same applies to the descriptor overload just below. Worth adding a C# counterpart to `test_marketplace_first_wins.py` so this does not drift again.

rewrite-csharp/csharp/OpenRewrite/Core/RecipeMarketplace.cs
```diff
@@ -181,21 +181,25 @@ public class Category
     {
         public CategoryDescriptor Descriptor { get; }
         private readonly List<Category> _categories = [];
-        private readonly Dictionary<RecipeDescriptor, Recipe?> _recipes = new();
+        // Keyed by recipe name, holding the listing-weight view and (when installed from a class) the
+        // live recipe instance PrepareRecipe uses to build the full tree.
+        private readonly Dictionary<string, (RecipeListing Listing, Recipe? Recipe)> _recipes = new();
```
shanman190, line 186: Shape comment - applies across all four engines rather than just this line.

Each engine now holds a `RecipeListing`, which is the goal, but in four different containers, each with recipe origin in a parallel side map that gets re-joined by name at emit time:

| | container | origin | dedup |
|---|---|---|---|
| Java | `List<RecipeListing>` | **on** the listing (`bundle`) | first-wins |
| C# | `Dictionary<string, (RecipeListing, Recipe?)>` | `_recipeOrigin` | last-wins |
| JS | `Map<RecipeListing, RecipeConstructor?>` | `recipeOrigin` | keyed by object identity, no dedup |
| Python | `Dict[str, (RecipeListing, Type[Recipe]?)]` | `_attribution` | first-wins |
| Go | `Registration{Listing, Constructor, Descriptor*}` | `s.recipeOrigin` | first-wins |

Java's listing carries its own origin, which is what lets its emitter be a straight projection of the listing. Putting `packageName` on each engine's `RecipeListing` at install time and settling on one name-keyed map would delete four side maps and four join sites, and would make first-wins a single decision instead of four independent ones - which is exactly what the C# and JS rows above currently are.

rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
```diff
@@ -57,37 +83,61 @@ public RecipeMarketplace toMarketplace(RecipeBundle bundle) {
...
         }
         return marketplace;
     }
 
-    public static GetMarketplaceResponse fromMarketplace(RecipeMarketplace marketplace, List<RecipeBundleResolver> resolvers) {
+    /**
+     * The {@link RecipeListing} a row represents, attributed to {@code bundle}, built from its
+     * listing-weight fields.
+     */
+    private static RecipeListing toListing(Row row, RecipeBundle bundle) {
```
shanman190, line 98: The `@Nullable` on `Row.name`/`options`/`dataTables` and this `!= null ? ... : emptyList()` coalescing only make sense with the `fromDescriptor` fallback the PR description mentions, which is not in the diff. Every up-to-date engine always emits all three (Go's `Name string` is a non-pointer, Python emits `listing.name`, C# emits a non-nullable `RecipeListing.Name`, JS emits `readonly name: string`), so there is no reachable null.

Drop the annotations and the coalescing so the types state the actual contract. `Row.recipeCount` as a primitive `int` has the same shape of problem - it silently defaults to 0 where Java's own `fromDescriptor` floors it at 1 - though again, no current engine omits it.

rewrite-core/src/main/java/org/openrewrite/rpc/request/GetMarketplaceResponse.java
```diff
@@ -57,37 +83,61 @@ public RecipeMarketplace toMarketplace(RecipeBundle bundle) {
...
     }
 
-    private static void fromCategory(List<RecipeBundleResolver> resolvers,
-                                     Map<String, Row> rowByRecipeId,
+    /**
+     * @deprecated Use {@link #fromMarketplace(RecipeMarketplace)}. The {@code resolvers} argument is
+     * no longer needed: rows are emitted listing-weight straight off each {@link RecipeListing}, so no
+     * recipe is resolved or described here.
+     */
+    @Deprecated
```
shanman190, line 121: We can probably just remove this as the RPC stuff is internal. It has no remaining in-repo caller (`RewriteRpc` is updated in this PR), and the package is `@Incubating(since = "8.66.0")`, so the shim is churn plus a retained `RecipeBundleResolver` import.

rewrite-csharp/csharp/OpenRewrite/CSharp/Rpc/RewriteRpcServer.cs
```diff
@@ -2176,7 +2187,20 @@ public class EvictRequest
 
 public class GetMarketplaceResponseRow
 {
-    public RecipeDescriptorDto Descriptor { get; set; } = null!;
+    // Listing-weight fields (Name through RecipeCount) carry everything the host needs to list the
+    // marketplace without materializing recipes. Up-to-date engines populate these and leave
+    // Descriptor null; the host fetches the full descriptor lazily per recipe via PrepareRecipe.
```
shanman190, line 2192: Stale: `Descriptor` no longer exists on `GetMarketplaceResponseRow`, so "leave Descriptor null" has nothing left to refer to.

rewrite-python/rewrite/src/rewrite/rpc/server.py
```diff
@@ -1495,6 +1504,22 @@ def _delegate_descriptor(name: str) -> dict:
     }
 
 
+def _options_to_dicts(descriptor) -> List[dict]:
```
shanman190, line 1507: Both call sites now pass a `RecipeListing`, so the parameter name and the docstring ("a descriptor's own ... options") are one rename behind.

rewrite-python/src/integTest/java/org/openrewrite/python/PythonMarketplaceIntegTest.java
```diff
@@ -0,0 +1,70 @@
...
+        RecipeBundle bundle = new RecipeBundle("pip", "openrewrite", null, null, null);
+        RecipeMarketplace marketplace = PythonRewriteRpc.getOrStart().getMarketplace(bundle);
+
+        assertThat(marketplace).isNotNull();
+        assertThat(marketplace.getAllRecipes()).isNotEmpty();
+        // Every listing was built from a lightweight row: name and the transitive recipe count
+        // survive the round-trip, proving the host didn't need the full descriptor.
+        for (RecipeListing listing : marketplace.getAllRecipes()) {
+            assertThat(listing.getName()).isNotBlank();
+            assertThat(listing.getRecipeCount()).isGreaterThanOrEqualTo(1);
```
shanman190, line 67: `>= 1` catches an omitted `recipeCount` but not a wrong one - a direct-children-only count would pass this for every composite. Asserting a known composite's exact count would make this the cross-process guard the class javadoc claims it is.

This is also the natural place to assert `estimatedEffortPerOccurrence` survives the round trip with the right magnitude, given the unit mismatch flagged on the Python emitter.

shanman190 (commented):
Reviewed against the requirement that the RPC marketplaces carry the same shape as the Java `RecipeMarketplace`: a tree of lightweight `RecipeListing`s that can be prepared into full `Recipe` instances.

The shape lands in all four engines. Five things to fix before merge, plus a shape divergence worth settling now. Inline comments carry the detail; summarizing here.

### `estimatedEffortPerOccurrence` means four different things across the four RPC languages

Java's contract, probed against a replica of `JsonMessageFormatter`'s mapper:

```
Java WRITES PT5M as: {"estimatedEffortPerOccurrence":300.000000000}
Java READS 5        as: PT5S
Java READS 300      as: PT5M
Java READS "PT5M"   as: PT5M
```

A JSON number of **seconds**, with ISO-8601 strings tolerated on read. Measured against that:

| engine | emits | verdict |
|---|---|---|
| C# | `"PT5M"` (ISO-8601) | correct |
| JS | bare number meaning **minutes** | 60x low, both directions |
| Python | bare number meaning **minutes** | 60x low |
| Go | nothing | field declared, never set |

The JS/Python unit error is pre-existing - the same number rode inside the nested descriptor - but this PR promotes the field to a named element of a documented contract, so this is the moment to pin it.

### C# flipped first-wins to last-wins

See the inline comment on `RecipeMarketplace.cs`. First-wins is the contract in every other implementation, and `test_marketplace_first_wins.py` guards it on the Python side; C# has no counterpart.

### The two halves of the PR disagree about `ColumnDescriptor.type`

The new C# `ColumnDescriptorDto` deliberately emits `Type = ""` because `ColumnDescriptor.type` has no `@Nullable` and is `@EqualsAndHashCode.Include` - correct reading, with a test asserting it. Python's `DataTable.descriptor()` omits `type` entirely and this PR passes those dicts through untouched, so the host ends up with a null `type` feeding `hashCode()`. Pre-existing, but the fix is the same one-line default the C# side just took, and it seems worse to document the invariant on one side while leaving it violated on the other.

### gofmt

`gofmt -l ./cmd ./pkg` flags `cmd/rpc/main.go` on the new `marketplaceRow` struct. It was already dirty on `main` from an unrelated `pendingDependencyTypes` line, so CI probably will not catch the new one.

### The PR description no longer matches the diff

It says `descriptor` is now nullable, that `toMarketplace` falls back to `fromDescriptor` when `name == null`, and that unknown fields are ignored. None of that is in the diff - `Row.descriptor` is deleted outright - and the host's mapper leaves `FAIL_ON_UNKNOWN_PROPERTIES` at Jackson's default, so unknown fields throw rather than being ignored (`RewriteRpcProcess.java:229` uses the `Module...` constructor). Backward compatibility is not a requirement here, so no code needs to change; just correct the description so the next reader does not go hunting for a fallback that is not there.

### Not anchorable to the diff

`rewrite-python/rewrite/src/rewrite/rpc/server.py` around line 1847: `# Look up the recipe - returns (RecipeDescriptor, Type[Recipe]) tuple` - it returns a listing now.

### Where my evidence stops

Not run: the C# and JS suites. Those findings rest on reading types and call sites, not on `dotnet build` or `npm test` - worth confirming CI is green on both before merging.

Not verified: the description's sizing claim (~82x more nodes, tens of MB vs ~717 KB for 846 recipes). I took it at face value; the direction is obviously right, but nothing here checks the magnitude.

PR #8386, opened 2026-08-04
Run nested ScanningRecipe scanners in the TypeScript scheduler

### LINKED ISSUE #8385: TypeScript scheduler never runs a nested ScanningRecipe's scanner

`@openrewrite/rewrite` 8.87.7, and still on 8.89.0-20260804-200316.

The TypeScript recipe scheduler in `rewrite-javascript/rewrite/src/run.ts` (`scheduleRun` / `scheduleRunStreaming`), as exercised by `RecipeSpec.rewriteRun` and by the Node CLI.

A `ScanningRecipe` nested inside a plain composite `Recipe` never has its scanner run. Its accumulator stays at `initialValue(...)` for the whole run, so the edit phase silently produces no changes. Setting `spec.recipe` to the `ScanningRecipe` directly makes the same test pass.

```ts
class Wrapper extends Recipe {
    readonly name = "test.wrapper";
    readonly displayName = "Wrapper";
    readonly description = "Wraps a scanning recipe.";

    async recipeList(): Promise<Recipe[]> {
        return [new SomeScanningRecipe()];
    }
}

test('scanning recipe nested in a plain composite', async () => {
    const spec = new RecipeSpec();
    spec.recipe = new Wrapper();
    // a source whose change depends on what the scan phase accumulated
    await spec.rewriteRun(typescript(before, after));
});
```

There are two independent causes, and both have to be present for a composite to work:

**1. The scan phase stops traversing the recipe list at the first non-scanning recipe.**

`recurseRecipeList` (src/run.ts:51) treats an `undefined` callback result as "stop traversing", including the rest of the sibling list:

```ts
async function recurseRecipeList<T>(recipe: Recipe, initial: T, fn: (recipe: Recipe, t: T) => Promise<T | undefined>): Promise<T | undefined> {
    let t: T | undefined = await fn(recipe, initial);
    for (const subRecipe of await recipe.recipeList()) {
        if (t === undefined) {
            return undefined;
        }
        t = await recurseRecipeList(subRecipe, t, fn);
    }
    return t;
}
```

The scan-phase callback (src/run.ts:114) returns `undefined` for every recipe that is not itself a `ScanningRecipe`:

```ts
await recurseRecipeList(recipe, b, async (recipe, b2) => {
    if (recipe instanceof ScanningRecipe) {
        return (await recipe.scanner(recipe.accumulator(cursor, ctx))).visit(b2, ctx, cursor)
    }
});
```

So when the root is a plain composite, `fn(root, b)` returns `undefined` and traversal aborts before the first child. Even when the root *is* a `ScanningRecipe`, the first plain sibling in its `recipeList` kills scanning for everything after it.

For contrast, the phase-2 `generate` callback (src/run.ts:124) always returns `generated`, so that traversal is complete. The edit-phase callbacks (src/run.ts:135/144/157) genuinely want the `undefined` short-circuit — that is how a deleted file stops being visited — so the two phases want different things from the same helper.

**2. `recipeList()` is re-invoked per phase and per source file, so sub-recipe instances are not stable.**

`recurseRecipeList` and `hasScanningRecipe` call `recipe.recipeList()` on every invocation: once per file during scanning, again during generate, and again per file during editing. Recipes overwhelmingly construct their sub-recipes inline (`return [new SomeScanningRecipe()]`), so each call returns *new* instances. `ScanningRecipe.accumulator` keys the accumulator off a per-instance `Symbol`, so the edit phase reads a fresh `initialValue(...)` even when the scan phase ran perfectly. `RpcRecipe.recipeList()` additionally re-issues a `prepareRecipe` RPC call for every child on every one of those invocations.

Parity with the Java `RecipeRunCycle`, which is the reference behavior:

- `scanSources` walks the whole recipe list regardless of whether an individual recipe is a `ScanningRecipe`, and discards whatever the scanner's visitor returns (it keeps `after = source`, and asserts the scanner did not mutate the tree — [RecipeRunCycle.java#L174-L199](https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java#L174-L199)). A scanner returning nothing must not halt the run.
- `Recipe.getRecipeList()` in Java is built once and cached, so a nested scanning recipe is the same object in the scan, generate, and edit phases.

### What did you see instead?

The nested `ScanningRecipe` never scans, its accumulator stays at `initialValue(...)`, and the composite reports no changes.

### What is the full stack trace of any errors you encountered?

None — that is the main hazard here. There is no error and no warning; a `RecipeSpec` test of the composite quietly reports no changes, which reads as "the recipe correctly made no change".

### Are you interested in contributing a fix to OpenRewrite?

Yes.

### Additional context

Composites are the normal way these recipes ship. `rewrite-angular` already nests `ScanningRecipe`s inside composites (`AddPrimengProvider` inside `UpgradeToPrimeNG18`, `MigratePFluidToWrapper` inside `UpgradeComponentsTo18`), and all of them no-op under `scheduleRun`.

timtebeek:
- Fixes #8385

A `ScanningRecipe` nested anywhere inside a plain composite `Recipe` never had its scanner run under `scheduleRun`, so its accumulator stayed at `initialValue(...)` and the edit phase silently produced no changes. Two independent causes, both in `rewrite-javascript/rewrite/src/run.ts`:

**The scan phase aborted the recipe list.** `recurseRecipeList` treats an `undefined` callback result as "stop traversing", and the scan callback returned `undefined` for every recipe that was not itself a `ScanningRecipe`. A plain composite root therefore aborted before its first child, and a plain sibling aborted everything after it. Scanning must not modify the tree and its visit result is discarded anyway, so the scanner now runs for its side effects and the source file is always passed through. This matches [`RecipeRunCycle.scanSources`](https://github.com/openrewrite/rewrite/blob/main/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java#L174-L199), which keeps `after = source` regardless of what the scanner returns and asserts the scanner did not mutate the tree.

The `undefined` short-circuit is left intact for the edit phase, where it is how a deleted file stops being visited.

**Sub-recipe instances were not stable across phases.** `recipeList()` was re-invoked once per file while scanning, again while generating, and again per file while editing. Recipes overwhelmingly construct their sub-recipes inline (`return [new SomeScanningRecipe()]`), so each call returned new instances, and `ScanningRecipe.accumulator` keys off a per-instance `Symbol` — the edit phase read a fresh `initialValue(...)` even when the scan phase had run correctly. Each recipe's list is now resolved once per run through a `WeakMap`, mirroring Java's cached `getRecipeList()`. This also stops `RpcRecipe.recipeList()` from re-issuing a `prepareRecipe` RPC call per child per file per phase.

While threading that through, `recursiveOnComplete` was firing its recursive calls without awaiting them, so nested recipes' `onComplete` could still be in flight after `scheduleRun` returned. Awaited now.

### Tests

New `test/run.test.ts`, 9 tests:

- `ScanningRecipe` one level under a plain composite, and three levels deep
- `ScanningRecipe` positioned after a plain sibling under a `ScanningRecipe` root
- `ScanningRecipe` instantiated inside `recipeList()` rather than held in a field
- a cross-file scan: the accumulator sees every file before any file is edited
- a scanner whose visitor returns `undefined` does not halt the run
- nested `onComplete` finishes before the run returns
- two edit-phase deletion tests asserting the short-circuit still yields `after === undefined`

Reverting `src/run.ts` to `main` fails 7 of the 9; the two that still pass are exactly the deletion short-circuit tests, i.e. the behavior this change leaves alone.

### DIFF
--- a/rewrite-javascript/rewrite/src/run.ts
+++ b/rewrite-javascript/rewrite/src/run.ts
@@ -49,12 +49,12 @@ export class Result {
 type RecipeLists = WeakMap<Recipe, Recipe[]>;
 
 async function subRecipes(recipe: Recipe, recipeLists: RecipeLists): Promise<Recipe[]> {
-    let subRecipes = recipeLists.get(recipe);
-    if (subRecipes === undefined) {
-        subRecipes = await recipe.recipeList();
-        recipeLists.set(recipe, subRecipes);
+    let resolved = recipeLists.get(recipe);
+    if (resolved === undefined) {
+        resolved = await recipe.recipeList();
+        recipeLists.set(recipe, resolved);
     }
-    return subRecipes;
+    return resolved;
 }
 
 async function hasScanningRecipe(recipe: Recipe, recipeLists: RecipeLists): Promise<boolean> {
@@ -115,17 +115,16 @@ export async function* scheduleRunStreaming(
     const cursor = rootCursor();
     const recipeLists: RecipeLists = new WeakMap();
     const isScanning = await hasScanningRecipe(recipe, recipeLists);
+    const knownTotal = Array.isArray(before) ? before.length : -1; // -1 = unknown total
 
     if (isScanning) {
         // For scanning recipes, pull files from the generator and scan them immediately.
         // Files are stored for the later edit phase.
         const files: SourceFile[] = [];
-        const iterable = Array.isArray(before) ? before : before;
-        const knownTotal = Array.isArray(before) ? before.length : -1; // -1 = unknown total
 
         // Phase 1: Pull files from generator and scan each immediately
         let scanCount = 0;
-        for await (const b of iterable) {
+        for await (const b of before) {
             files.push(b);
             scanCount++;
             onProgress?.('scanning', scanCount, knownTotal, b.sourcePath);
@@ -172,10 +171,8 @@ export async function* scheduleRunStreaming(
         }
     } else {
         // For non-scanning recipes, process files immediately as they come in
-        const iterable = Array.isArray(before) ? before : before;
-        const knownTotal = Array.isArray(before) ? before.length : -1; // -1 = unknown total
         let processCount = 0;
-        for await (const b of iterable) {
+        for await (const b of before) {
             processCount++;
             onProgress?.('processing', processCount, knownTotal, b.sourcePath);
             const editedB = await recurseRecipeList(recipe, b, recipeLists, async (recipe, b2) => (await recipe.editor()).visit(b2, ctx, cursor));
--- a/rewrite-javascript/rewrite/src/run.ts
+++ b/rewrite-javascript/rewrite/src/run.ts
@@ -40,28 +40,47 @@ export class Result {
     }
 }
 
-async function hasScanningRecipe(recipe: Recipe): Promise<boolean> {
+/**
+ * Recipes commonly instantiate their sub-recipes inside `recipeList()`, so calling it more
+ * than once yields different instances. A scanning sub-recipe would then hold a different
+ * accumulator in every phase and for every source file. Resolving each recipe list once per
+ * run keeps sub-recipe identity, and therefore accumulator identity, stable.
+ */
+type RecipeLists = WeakMap<Recipe, Recipe[]>;
+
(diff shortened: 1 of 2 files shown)


knutwannheden (approved):
👍

PR #8392, opened 2026-08-05
Never CHANGE a ref-deduplicated RPC slot; re-ADD the new value

knutwannheden:
## Motivation

A ref-deduplicated (`asRef`) slot — JavaType attribution being the main case — resolves on the receiving peer to an instance in a persistent ref cache. That instance is aliased by every slot that was delivered by ref, across trees and across source files, and the cache outlives per-file eviction checkpoints. When a recipe replaced only slot A's type with a new instance of the same class, the sender emitted a CHANGE diff and the receiver applied it to the cached before instance *in place* (`JavaTypeReceiver`/`unsafeSet` and its peers' equivalents). Every alias of that instance — slot B in the same file, and later files resolving ref-only ADDs — silently changed too. The sender's CHANGE branch also never registered the new value in its refs map, so a type shared by N slots was re-walked and resent N times, and receiver-side identity dedup degraded into N distinct instances. Finally, the CHANGE walk had no cycle handling: diffing two distinct-but-isomorphic cyclic type graphs — the normal shape when the same type is materialized twice with different reference identity — recursed infinitely and crashed with `StackOverflowError` (C# needed sender-side cycle detection for exactly this crash in #7148; the Java path was never exercised because of the flyweight bug below).

This settles the semantics of a diff on a ref slot: **a ref-deduplicated slot never emits CHANGE**. When the instance changes, the sender re-adds the new value — a full ADD under a fresh ref id the first time, ref-only ADDs for every repeat. ADD in place of CHANGE is already a legal, handled transition in every existing receiver, so this is not a wire-protocol change and updated senders interoperate with older receivers. Cycles terminate structurally on this path because `add()` registers each node in the refs table before walking its fields, so a back edge resolves to a single ref-only ADD. The alternatives — receiver-side copy-on-change, or making CHANGE carry a ref id — would require lock-step changes to all five receivers (the latter a wire-format change), would still need sender-side cycle detection added to the diff walk, and would still leave the N-times resend in place.

Two enablers ride along: `Reference.asRef` now allocates a fresh instance per call in Java and C#, which both shared a per-thread flyweight that made the after/before getters overwrite each other so every ref-slot diff collapsed to NO_CHANGE; and the Python sender's CHANGE-path ref re-keying (which no receiver mirrored, leaving stale objects behind later ref-only ADDs) is gone because the branch it lived in is gone.

## Summary

- Java `RpcSendQueue`: `send()` routes a changed `Reference` slot to `add()`; `sendList()` does the same for `asRef` list items; `Reference.asRef` allocates per call
- TypeScript `queue.ts`: same two routing changes keyed on `isRef`
- Python `send_queue.py`: `_send_as_ref` re-adds instead of CHANGE + re-key; `send_list` as_ref items re-add
- Go `send_queue.go`: `Send`/`sendList` same routing keyed on `IsRef`/`asRef`
- C# `RpcSendQueue.cs`: same routing; `Reference.AsRef` drops its `[ThreadStatic]` flyweight
- Regression tests in all five languages, including an end-to-end `JavaSender`/`JavaReceiver` round trip proving that changing one of two slots sharing a `JavaType` instance leaves the other slot untouched on the receiver, and a cyclic-graph round trip proving the diff terminates and the receiver closes the cycle by identity

## Test plan

- [x] New `JavaReceiverTest.changingOneSlotOfSharedTypeLeavesOtherSlotsUntouched` fails before the fix (unchanged slot came back as `java.util.ArrayList`) and passes after
- [x] New `JavaReceiverTest.diffOfDistinctCyclicTypeGraphsRoundTripsWithCycleIntact` — the same diff under the previous CHANGE semantics was observed to throw `StackOverflowError` after 10,000+ emitted messages
- [x] New protocol-level tests pin ADD/ref emission in each peer: `RpcSendQueueTest` (Java), `queue.test.ts` (TS), `test_send_queue.py` (Python), `send_queue_ref_test.go` (Go), `RpcSendQueueRefTest` (C#) — each observed failing before its sender was fixed
- [x] `./gradlew :rewrite-core:test :rewrite-java:test`
- [x] `npm test` in `rewrite-javascript/rewrite` (1915 passed)
- [x] `pytest tests/rpc/` in `rewrite-python/rewrite` (131 passed)
- [x] `go test ./...` in `rewrite-go`
- [x] `dotnet test` for the touched C# Rpc tests


knutwannheden:
Wire-overhead measurements for the semantics change (200 changed type slots, real `JavaSender`/`JavaReceiver` round trip through paired queues; message count is the metric that tracks serialization and transport cost):

| Scenario | Old (CHANGE delta) | New (re-ADD) |
|---|---|---|
| 200 slots sharing one `JavaType`, all flipped to one new shared instance (the `ChangeType` shape) | 4213 messages, 201 inline values, receiver corrupts aliases | **2228 messages, 9 inline values** — one full ADD plus 199 ref-only ADDs |
| 200 slots with 200 distinct types, each flipped to a distinct new type | 4213 messages, 201 inline values | 5213 messages, 1601 inline values |

The first row is the dominant real-world pattern and gets ~47% cheaper because repeats of the same transition collapse to single-message ref-only ADDs. The second row is the worst case (+24% messages: an ADD inlines each type's scalar fields where a delta emitted NO_CHANGE markers) and requires a recipe materializing hundreds of distinct new type instances in one file — type mappers intern types, so real workloads look like the first row. Initial full-tree sends and unchanged-slot diffs (the overwhelming majority of traffic) take exactly the same code path as before.

Memory: the sender registers one refs-map entry per distinct new value (~tens of bytes, evicted by the existing per-source-file rollback, and the very thing that enables the dedup above); the receiver materializes one new instance per distinct new value, which any aliasing-correct scheme must do — when one slot changes and another does not, the receiver has to end up with two distinct objects.

PR #8394, opened 2026-08-05
`PythonImportService`: fall back to the serving RPC connection when the JVM is the spawned server

knutwannheden:
## Motivation

- PR #8323 added `PythonImportService`, which lets `org.openrewrite.java.ChangeType` (and any `maybeAddImport`/`maybeRemoveImport` caller) retire and add Python imports by dispatching the Python side's own `AddImport`/`RemoveImport` visitors over RPC. Its dispatch obtained the connection via `PythonRewriteRpc.get()`, which is only non-null when the JVM spawned the Python peer (JVM-as-host, e.g. Moderne CLI). In the inverted topology — Python (or JavaScript) spawning `org.openrewrite.maven.rpc.JavaRewriteRpc` as a subprocess, as the pytest harness in #8390 does — the handle is always null and the import edit silently no-oped: `ChangeType(typing.List -> list)` rewrote the annotations but left the now-unused `from typing import List` behind.

The peer that sent the in-flight `Visit` request both owns the tree being visited and implements the import visitors, so the serving connection is exactly the right place to dispatch to. This PR makes that connection discoverable from visitor code.

## Examples

A visitor executing on behalf of a remote peer can reach that peer's own visitors:

```java
// Inside code running under a Visit/BatchVisit request handler:
RewriteRpc rpc = RewriteRpc.current(); // the connection the in-flight request arrived on
rpc.visit(tree, "org.openrewrite.python.RemoveImport", options, p, cursor);
```

`PythonImportService` uses it as a fallback:

```java
RewriteRpc rpc = PythonRewriteRpc.get();
if (rpc == null) {
    rpc = RewriteRpc.current();
}
```

## Summary

- `RewriteRpc.current()`: a request-scoped thread-local holding the `RewriteRpc` instance the currently handled request arrived on, or null outside request handling. `Visit.Handler` and `BatchVisit.Handler` scope it around visitor execution via `withCurrent(...)`, which saves and restores the previous value in a finally block.
- `PythonImportService.RpcImportVisitor.dispatch` falls back to `RewriteRpc.current()` when `PythonRewriteRpc.get()` is null, making Java-side Python import edits work in the Python- and JS-hosted topologies.
- Python's `AddImport` visitor no-ops for a bare builtin name (`hasattr(builtins, module)`): with the add path actually running, `ChangeType` retargeting a type to a builtin such as `list` would otherwise emit a nonsensical `import list`. The guard lives in the visitor rather than `maybe_add_import` because the RPC visitor registry constructs `AddImport` directly.

## Test plan

- [x] `RewriteRpcTest.nestedVisitBackToRequestOriginator`: a visitor served on one peer discovers the connection via `RewriteRpc.current()` and dispatches a follow-up visit back to the requesting peer, which modifies the tree — also proving the nested bidirectional exchange works in the in-process piped setup.
- [x] `PythonImportServiceRpcTest`: two in-process `RewriteRpc` peers where the host plays the Python process; a served visitor's `maybeRemoveImport("typing.List")` dispatches back to the host with options `{module: typing, name: List}`, received by a test stand-in registered under `org.openrewrite.python.RemoveImport`.
- [x] `test_add_import.py::test_builtin_name_is_not_imported` for the builtin guard.
- [x] Full `:rewrite-core:test` and `:rewrite-python:test` suites, plus the pure-Python suites (1627 passed).
- [x] End-to-end with the #8390 harness: `pytest tests/rpc/test_java_recipes.py::TestChangeType` passes — annotations rewritten, `from typing import List` removed, no spurious import. The corresponding xfail in #8390 can be dropped once both are merged.

### DIFF
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
@@ -268,20 +268,17 @@ public class RewriteRpc {
     }
 
     /**
-     * The {@link RewriteRpc} over which the request currently being handled on this thread
-     * arrived, or {@code null} when the thread is not handling an RPC request. Visitors
-     * instantiated on behalf of a remote peer use this to dispatch follow-up visits back to
-     * that peer — the owner of the tree being visited — when this process is the spawned
-     * server and so holds no process-manager handle to the peer (e.g. Python- or JS-hosted
-     * recipe runs delegating to Java recipes).
+     * The instance the request currently being handled on this thread arrived on, or
+     * {@code null} outside request handling. Lets a visitor dispatch follow-up visits back
+     * to the requesting peer when this process is the spawned server and has no other
+     * handle to it.
      */
     public static @Nullable RewriteRpc current() {
         return CURRENT.get();
     }
 
     /**
-     * Runs {@code work} with this instance discoverable via {@link #current()}. Request
-     * handlers wrap the execution of remotely requested work in this.
+     * Runs {@code work} with this instance discoverable via {@link #current()}.
      */
     public <T> T withCurrent(Callable<T> work) throws Exception {
         RewriteRpc previous = CURRENT.get();
--- a/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
+++ b/rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
@@ -41,6 +41,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.*;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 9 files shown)


rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
```diff
@@ -264,6 +267,36 @@ protected Boolean handle(Evict request) {
         jsonRpc.bind();
     }
 
+    /**
+     * The {@link RewriteRpc} over which the request currently being handled on this thread
```
jkschneider: Please trim doc essays.

rewrite-core/src/test/java/org/openrewrite/rpc/RewriteRpcTest.java
```diff
@@ -259,6 +259,32 @@ public Tree preVisit(Tree tree, ExecutionContext ctx) {
         );
     }
 
+    /**
+     * A visitor running inside a Visit request handler can discover the {@link RewriteRpc} the
```
jkschneider: Please trim doc essays.

rewrite-core/src/main/java/org/openrewrite/rpc/RewriteRpc.java
knutwannheden: Trimmed in e7a5f24977: current() is down to two sentences, withCurrent() to one.

rewrite-core/src/test/java/org/openrewrite/rpc/RewriteRpcTest.java
knutwannheden: Removed the test javadoc (and the helper class doc) in e7a5f24977 — the test name and DispatchBackToOriginator's body carry it.

PR #8396, opened 2026-08-05
Send changed codec-less list elements inline over RPC

knutwannheden:
## Motivation

`RpcSendQueue.sendList`'s CHANGE branch emitted an `RpcObjectData` with a null value for list elements that have neither an `onChange` callback nor an `RpcCodec`, unlike `send()` which inlines the value in that case. `Markers.rpcSend` sends the marker list with no `onChange`, and markers like `GitProvenance`, `BuildTool`, and `JavaVersion` have no codec — so a marker that was modified while keeping its id (e.g. `gitProvenance.withBranch("other")`) diffed as CHANGE but its new field values never crossed the wire, and the receiver silently kept the stale marker. The same sender bug existed in the TypeScript, Python, C#, and Go implementations of the protocol.

## Summary

- `rewrite-core` `RpcSendQueue.sendList`: inline the element value on CHANGE when `onChange == null` and no codec exists, mirroring `send()`; the receivers on the Java, TypeScript, and Python sides already apply an inline value on CHANGE, so no receiver changes were needed there
- `rewrite-javascript` `src/rpc/queue.ts`: same fix in `sendList`
- `rewrite-python` `src/rewrite/rpc/send_queue.py`: same fix in `send_list`; it now also passes the element's codec to `_do_change`, which previously received none — a codec'd list element sent without an `on_change` callback emitted no property messages while the receiver's codec consumed them, a latent queue desync
- `rewrite-csharp` `RpcSendQueue`: same fix in `SendList`; additionally `RpcReceiveQueue`'s CHANGE branch deserializes typed inline payloads via `DeserializeInline` (mirroring its ADD path) so wire-format JSON payloads resolve to the right marker type, and `Markers.RpcReceive` passes a null `onChange` so the queue's built-in `IRpcCodec` dispatch and inline-value fallback apply — its previous callback returned the stale before-marker for codec-less types
- `rewrite-go` `pkg/rpc`: same fix in `sendList`; the codec-less `GenericMarker` inlining that `add` already did is extracted into `inlineValue` and applied on all three emit paths (`Send`, `sendList`, `add`), and the receiver's ADD-only `GenericMarker` hydration is extracted into `hydrateGenericMarker` and applied on CHANGE as well

## Test plan

- [x] New wire-format test `RpcSendQueueTest.changedListElementWithoutCodecIsInlined` and paired-queue round-trip `RpcReceiveQueueTest.changedMarkerWithSameIdRoundTrips` (both observed failing with the stale value before the fix)
- [x] New round-trip tests for the TypeScript (`test/rpc/queue.test.ts`), Python (`tests/rpc/test_opaque_roundtrip.py`), C# (`OpenRewrite.Tests/Rpc/ChangedMarkerRoundTripTest.cs`), and Go (`pkg/rpc/marker_codec_test.go`) queues, likewise each observed failing first
- [x] `./gradlew :rewrite-core:test` and the JVM-side RPC round-trip tests in `rewrite-java`, `rewrite-go`, and `rewrite-python`
- [x] TypeScript `npm run typecheck` plus the `test/rpc`, `test/json`, and `test/yaml` vitest suites
- [x] Python `pytest tests/rpc/` (128 passed)
- [x] Full C# `dotnet test` suite (1954 passed), including the `RpcFixture`-based Java↔C# integration tests run against the generated RPC test-server classpath
- [x] Go `go test ./...` in `rewrite-go` (the `pkg/parser` test-build failure is pre-existing and unrelated)


jkschneider:
@copilot resolve the merge conflicts in this pull request

PR #8397, opened 2026-08-05
FindTypes: mark parameterized type usages in non-Java trees

knutwannheden:
## Motivation

`FindTypes` missed some type usages in Python LSTs. For `def foo(items: List[str]) -> List[int]` (with `from typing import List`), `FindTypes("typing.List")` marked the return hint but not the parameter annotation. `visitIdentifier` skipped any identifier whose parent is a `J.ParameterizedType`, expecting `visitTypeName` to mark the enclosing node instead — but `visitTypeName` only sees a parameterized type when a `JavaVisitor` parent routes it there. In Python the parent is `Py.TypeHint`, which visits it as a plain expression, so the usage was never marked. (`J.ArrayType` does not have this gap: `JavaVisitor.visitArrayType` routes its element type through `visitTypeName` itself.)

An extra twist: ty resolves the aliased `typing` generics per PEP 585, so the parameterized type `List[str]` is attributed as builtin `list<String>` while the `List` identifier inside carries `typing.List` — the identifier is the only node attributed with the searched type, so marking the parameterized node alone cannot fix this case.

## Summary

- `visitIdentifier` now skips an identifier inside a parameterized type only when the enclosing parameterized type is itself the match (and is therefore marked as a whole); otherwise the identifier is marked, which covers the aliased-generic case in Python.
- New `visitParameterizedType` override marks a matching parameterized type in every language, upholding the invariant the skip relies on. In Java trees this is idempotent with the existing `visitTypeName` marking (`SearchResult.found` dedups the marker and `found()` guards the data table row). In Python it covers non-aliased generics, e.g. a user-defined `Box[str]`.

## Test plan

- New `FindTypesTest` in `rewrite-python` (RPC-based, disabled in CI like the existing `ChangeTypeTest`): `findsParameterizedTypeUsages` asserts both `List` usages are marked; `findsUserDefinedGenericUsages` asserts a user-defined generic's parameterized type is marked as a whole.
- `./gradlew :rewrite-java-test:test --tests "org.openrewrite.java.search.*"` — all existing search tests (including `FindTypesTest`) pass unchanged.

### DIFF
--- a/rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
@@ -172,11 +172,18 @@ public class FindTypes extends Recipe {
 
         @Override
         public J visitIdentifier(J.Identifier ident, ExecutionContext ctx) {
+            Object parent = getCursor().getParentOrThrow().getValue();
+            // An identifier is only subsumed by an enclosing parameterized type when that type is itself the
+            // match (see `visitParameterizedType`). The identifier can be the sole matching node, e.g. in
+            // Python `items: List[str]` the parameterized type is attributed as builtin `list` while the
+            // `List` identifier is attributed as `typing.List`.
+            boolean matchingParameterizedTypeParent = parent instanceof J.ParameterizedType &&
+                    parameterizedTypeMatches((J.ParameterizedType) parent);
             if (ident.getType() != null &&
                     getCursor().firstEnclosing(J.Import.class) == null &&
                     getCursor().firstEnclosing(J.FieldAccess.class) == null &&
-                    !(getCursor().getParentOrThrow().getValue() instanceof J.ParameterizedType) &&
-                    !(getCursor().getParentOrThrow().getValue() instanceof J.ArrayType)) {
+                    !matchingParameterizedTypeParent &&
+                    !(parent instanceof J.ArrayType)) {
                 JavaType.FullyQualified type = TypeUtils.asFullyQualified(ident.getType());
                 if (typeMatches(Boolean.TRUE.equals(checkAssignability), fullyQualifiedType, type) &&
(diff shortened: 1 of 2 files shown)


rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
```diff
@@ -172,11 +172,18 @@ public JavaSourceFileVisitor(TypeNameMatcher fullyQualifiedType) {
 
         @Override
         public J visitIdentifier(J.Identifier ident, ExecutionContext ctx) {
+            Object parent = getCursor().getParentOrThrow().getValue();
+            // An identifier is only subsumed by an enclosing parameterized type when that type is itself the
```
jkschneider: Please trim doc essays.

rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
```diff
@@ -186,6 +193,24 @@ public J visitIdentifier(J.Identifier ident, ExecutionContext ctx) {
             return super.visitIdentifier(ident, ctx);
         }
 
+        @Override
+        public J visitParameterizedType(J.ParameterizedType type, ExecutionContext ctx) {
+            // In Java trees a matching parameterized type is also marked via `visitTypeName`, invoked by the
```
jkschneider: Please trim doc essays.

rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
knutwannheden: Trimmed to two lines in 303f43b.

rewrite-java/src/main/java/org/openrewrite/java/search/FindTypes.java
knutwannheden: Trimmed to two lines in 303f43b.

PR #8399, opened 2026-08-05
Make the import ChangeType adds agree with how a nested reference is written

### LINKED ISSUE #8398: ChangeType: nested target types get an import that disagrees with how the reference is written

Latest `main` (reproduced at `a2b8834d6f`).

Directly, via `RewriteTest` in `rewrite-java-test`.

```java
// target:   package b; public class Outer { public enum Nested { A } }
// original: package a; public enum Old { A }
rewriteRun(
  spec -> spec.recipe(new ChangeType("a.Old", "b.Outer$Nested", null)),
  java(
    """
      import a.Old;

      class Test {
          Old value = Old.A;
      }
      """,
    """
      import b.Outer;
      import b.Outer.Nested;

      class Test {
          Outer.Nested value = Outer.Nested.A;
      }
      """));
```

## What did you expect to see? What did you see instead?

The reference is written through the outer class (`Outer.Nested`) while imports are added for
**both** the outer class and the nested type. `import b.Outer.Nested;` is left behind with nothing
resolving through it:

- `RemoveUnusedImports` keeps it — the type genuinely is referenced, so it is not unused by that
  recipe's definition;
- `ShortenFullyQualifiedTypeReferences` does not touch `Outer.Nested.A` — it shortens
  package-qualified references, and this one is qualified by a type, not a package.

So the import sits there permanently, reading as dead code to every reviewer of the diff.

## The underlying problem: the emitted form is not deterministic

The dangling import is one symptom. `ChangeType` decides the *reference* form in
`visitIdentifier` / `visitFieldAccess`, and decides *which import to add* independently in
`postVisit`. Whether the two agree then depends on whether `AddImport.findTypeReference` happens to
spot a reference, which in turn depends on incidental details of the input. The same logical change
produces four different outputs:

| output | when |
|---|---|
| `Outer.Nested` + `import Outer;` | no member access on the nested type |
| `Outer.Nested` + **both** imports | member access present (the repro above) |
| `Nested` + `import Outer.Nested;` | member access present *and* `AddImport`'s internal shortener fires |
| `Outer.Nested` + **no import at all** | reference was package-qualified — does not compile |

The last row is a separate, harder bug. `updateOuterClassTypes` rebuilds the type tree without
attributing the intermediate `name` identifiers, so neither `AddImport`'s reference check nor
`ShortenFullyQualifiedTypeReferences` can see that the outer class is referenced, and no import is
added at all:

```java
new ChangeType("java.io.File", "java.util.Map$Entry", true)

class Test {          →     class Test {
    java.io.File p;             Map.Entry p;   // no import java.util.Map;
}                           }
```

## Further limitation: wrong import for deeper nesting

For a target nested more than one level (`b.A$B$C`), references are written as `A.B.C` but the
import added is for the *owning* class, `import b.A.B;`, rather than the outermost class `b.A`.

## This repository's own tests assert both shapes

The inconsistency is already baked into the test suite, which is why there is no free fix:

- `ChangeTypeTest.replaceWithNestedType` asserts `import java.util.Map;` + `Map.Entry p;`
- `ChangePackageTest.innerType` asserts `import some.thing.X.Y;` + `Y foo();`

Same situation, opposite expectations.

## Downstream impact

Scanning the recipe ecosystem for `ChangeType` configurations whose target is a nested type turns
up **43 across 11 repositories**. Note the trigger is wider than the documented `Outer$Nested`
form: `JavaType.ShallowClass.build` treats any uppercase-led dotted chain as nested, so
`com.foo.Outer.Nested` counts too, and that is how most downstream configurations are written.

| repository | nested-target `ChangeType` configs |
|---|---|
| rewrite-jackson | 13 (`PropertyNamingStrategies.*Strategy`, `JsonInclude$Include`) |
| rewrite-spring (v1 / v2) | 3 / 3 (`ResponseEntity.BodyBuilder`, `PropertyMapping.Skip`, `DockerConnectionConfiguration$Host`, ...) |
| rewrite-apache | 2 (`SocketConfig.Builder`, `IOReactorConfig.Builder`) |
| rewrite-migrate-java | 2 (`JSONReader.Feature`, `JSONWriter.Feature`) |
| rewrite-android-studio, rewrite-logging-frameworks, rewrite-spring-to-quarkus, rewrite-jackson-v1 | 1 each |

Downstream goldens likewise encode **both** shapes, because both are currently produced:

- expecting the bare nested name — rewrite-jackson `ReplacePropertyNamingStrategyConstantsTest`,
  rewrite-spring-v1 `JaxrsToSpringMvcResponseEntityTest`
- expecting the outer-qualified name — rewrite-migrate-java `UseJavaUtilBase64Test`,
  rewrite-testing-frameworks `UpdateMockWebServerMockResponseTest`, rewrite-apache
  `MigrateApacheHttpCoreNioTest`, rewrite-spring-v2 `ReplaceDeprecatedDockerApiTest`

`rewrite-testing-frameworks`'s `UpdateMockWebServerMockResponseTest.shouldMigrateMockResponseToBuilder`
is worth calling out: one file there contains both an import-bound reference and a
package-qualified one, so any rule that treats those two differently emits **both forms in the same
file**.

Two live recipes are affected by the missing-import variant today: rewrite-apache's
`IOReactorConfig.Builder` migration and rewrite-spring-v1's `ResponseEntity.BodyBuilder` migration.

## Originally reported downstream

- openrewrite/rewrite-cucumber-jvm#47 (item 7c): `cucumber.api.SnippetType` →
`io.cucumber.junit.CucumberOptions$SnippetType` turns
`@CucumberOptions(snippets = SnippetType.CAMELCASE)` into
`@CucumberOptions(snippets = CucumberOptions.SnippetType.CAMELCASE)` alongside a freshly added
`import io.cucumber.junit.CucumberOptions.SnippetType;` that nothing then uses by that name.

timtebeek:
- Fixes #8398

## What's changed

`ChangeType` decided the *reference* form for a nested target type in `visitIdentifier` /
`visitFieldAccess`, and decided *which import to add* independently in `postVisit`. The two did not
have to agree, and whether they did depended on whether `AddImport.findTypeReference` happened to
spot a reference — so the same logical change produced four different outputs (see the issue).

This makes the import follow the reference: a nested target is written through its outer class
chain, so the import added is the **outermost** class and never the nested type itself.

```diff
  new ChangeType("a.Old", "b.Outer$Nested", null)

  import b.Outer;
- import b.Outer.Nested;      // nothing resolves through this

  class Test {
      Outer.Nested value = Outer.Nested.A;
  }
```

Three related defects live in the same code and are fixed here as well:

- **Missing import on package-qualified references.** `updateOuterClassTypes` rebuilt the type tree
  without attributing the intermediate `name` identifiers, so neither `AddImport`'s reference check
  nor `ShortenFullyQualifiedTypeReferences` could see the outer class was referenced, and *no*
  import was added — `Map.Entry p;` on its own does not compile. Names inside an import statement
  are deliberately left unattributed: an import is not a usage, and treating it as one makes the
  outer class look referenced when it isn't (this is what keeps `ChangePackageTest.innerType`
  passing).
- **Wrong import for deeper nesting.** A `b.A$B$C` target is written `A.B.C` but imported the
  owning class `b.A.B` instead of the outermost `b.A`.
- **Aliased imports of a nested target.** An alias binds the nested type directly, so importing
  only the outermost class leaves nothing referencing it and `AddImport`'s `onlyIfReferenced` check
  emits nothing at all. Kotlin's `import a.b.Original as MyAlias` migrated to *no import*, which
  does not compile. The import target now follows how the reference is actually bound.

Two live downstream recipes hit the missing-import variant today: rewrite-apache's
`IOReactorConfig.Builder` migration and rewrite-spring-v1's `ResponseEntity.BodyBuilder` migration.

## Scope of the guarantee

Narrower than "always emits `Outer.Nested`": when a file already has a single-type import of the
nested type, that import is rewritten in place rather than removed and re-added, and the bare
reference survives. The guarantee is that the import which gets *added* always matches the
reference form that gets *written*, which is what closes the issue. `ChangePackageTest.innerType`
covers that case and is unchanged.

## Behaviour change: output is more verbose when the outer class name is taken

Raised by @knutwannheden. If the target's **outermost** class name collides with an existing
import, the reference falls back to fully qualified rather than being written through the outer
class:

```java
import c.Outer;                          import c.Outer;
import a.Old;

class Test {                     →       class Test {
    Outer other;                             Outer other;
    Old value = Old.A;                       b.Outer.Nested value = b.Outer.Nested.A;
}                                        }
```

`main` produces `import b.Outer.Nested;` + `Nested value = Nested.A;` here — shorter, but by
importing the nested type, which is the shape this change moves away from. Both compile, so this is
readability rather than correctness, and it follows from the direction rather than from a bug: if
the outer name cannot be bound, `Outer.Nested` is not available. Covered by
`changeToNestedTypeWhenOuterClassNameIsTaken`.

Worth noting for review: `hasNoConflictingImport` compares `newType.getClassName()`, the dotted
`"Outer.Nested"`, so it never fires for a nested target — the collision is caught one layer down by
`AddImport.checkImportsForType` returning `IMPORT_AMBIGUITY`, which then triggers
`FullyQualifyTypeReference`. The outcome is right, but reached by a different route than the code
reads. Aligning that check is deliberately left out of this PR; see
https://github.com/openrewrite/rewrite/pull/8399#issuecomment-5203527343.

## Tests

All additions — no existing expectation in `ChangeTypeTest` is modified.

`ChangeTypeTest` gains coverage asserting the import and the reference form **together**, which no
existing test did. Cases that differ only in how the input reference is written are grouped into a
single run with one source file per form:

- `changeToNestedType` — the reported repro (flat → nested), plus a file where an unrelated import
  shares the nested type's simple name
- `changeToNestedTypeWhenOuterClassNameIsTaken` — the collision case above
- `changeFromNestedTypeToFlatType`
- `changeFromNestedTypeToNestedType` — `ViaImport` and `ViaOuterClass` source files
- `changeToDeeplyNestedType` — `ViaImport` and `ViaPackageQualifiedName` source files
- `replaceFullyQualifiedReferenceWithNestedType` — pins the previously missing import

`changeImportAliasToNestedType` is added to `rewrite-kotlin`'s `ChangeTypeTest`, since Java has no
import aliases. It fails without the alias fix above.

Four of these fail against `origin/main`, which is what pins the defects:
`changeToNestedType`, `changeFromNestedTypeToNestedType`, `changeToDeeplyNestedType`,
`replaceFullyQualifiedReferenceWithNestedType`.

## Downstream impact

Measured by publishing this branch to the local Maven repository and diffing full downstream test
suites against a baseline run of the same suite — many downstream failures are offline
dependency-resolution noise, so only a diff is meaningful.

| repository | new failures |
|---|---|
| rewrite-apache | 0 |
| rewrite-migrate-java | 0 |
| rewrite-testing-frameworks | 0 |
| rewrite-jackson | 3 |
| rewrite-spring-v1 | 3 |

The six are all goldens that currently assert the bare nested name and would move to the
outer-qualified one:

- rewrite-jackson — `ReplacePropertyNamingStrategyConstantsTest` x2,
  `CodehausToFasterXMLTest.onlyJsonSerializeInclusion`
- rewrite-spring-v1 — `JaxrsToSpringMvcResponseEntityTest` x2,
  `AuthorizeHttpRequestsTest.noArgAuthorizeRequestsWithVars`

rewrite-spring-v2 could not be measured; it does not compile against any core version right now
(`MigrateListenableFuture.java:42`, `Preconditions.check(UsesType, ListenableToCompletableFuture)`),
pre-existing and unrelated to this change.

These measurements predate the alias fix and the collision test. The alias fix only changes
behaviour when an import alias is present, and no scanned downstream repository combines an aliased
import with a nested-type `ChangeType` target, so the numbers should still hold — but they were not
re-measured against the final commit.

The trade-off against the alternative direction (preferring the bare nested name, which is what a
hand migration usually writes) is laid out in
https://github.com/openrewrite/rewrite/issues/8398#issuecomment-5192256690 — it costs more
downstream churn and needs a same-file consistency problem solved first. **Draft** pending a call
on which shape we want as the default.

## Verification

`rewrite-java-test`, `rewrite-java`, `rewrite-kotlin` and `rewrite-groovy` green locally; CI green
on the pre-condensing commit (https://github.com/openrewrite/rewrite/actions/runs/31035918079) and
re-running for the current one.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeTypeTest.java
@@ -66,6 +66,38 @@ class ChangeTypeTest implements RewriteTest {
       }
       """;
 
+    @Language("java")
+    String flatOld = """
+      package a;
+      public enum Old { A }
+      """;
+
+    @Language("java")
+    String nestedTargetOuter = """
+      package b;
+      public class Outer {
+          public enum Nested { A }
+      }
+      """;
+
+    @Language("java")
+    String nestedOriginalOuter = """
+      package a;
+      public class Outer {
+          public enum Nested { A }
+      }
+      """;
+
+    @Language("java")
+    String deeplyNestedTarget = """
+      package b;
+      public class A {
+          public static class B {
+              public enum C { A }
+          }
+      }
+      """;
+
     @DocumentExample
     @Test
     void doNotAddJavaLangWrapperImports() {
@@ -332,6 +364,235 @@ class ChangeTypeTest implements RewriteTest {
 
               class Test {
                   Map.Entry p;
+              }
+              """
+          )
+        );
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Test
+    void replaceFullyQualifiedReferenceWithNestedType() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("java.io.File", "java.util.Map$Entry", true)),
+          java(
+            """
+              class Test {
+                  java.io.File p;
+              }
+              """,
+            """
+              import java.util.Map;
+
+              class Test {
+                  Map.Entry p;
+              }
+              """
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite-cucumber-jvm/issues/47")
+    @Test
+    void changeToNestedType() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Old", "b.Outer$Nested", null)),
+          java(nestedTargetOuter, SourceSpec::skip),
+          java(flatOld, SourceSpec::skip),
+          java(
+            """
+              package c;
+              public class Nested {}
+              """,
+            SourceSpec::skip
+          ),
+          java(
+            """
+              import a.Old;
+
+              class Test {
+                  Old value = Old.A;
+              }
+              """,
+            """
+              import b.Outer;
+
+              class Test {
+                  Outer.Nested value = Outer.Nested.A;
+              }
+              """
+          ),
+          java(
+            """
+              import a.Old;
+              import c.Nested;
+
+              class UnrelatedSameSimpleName {
+                  Nested other;
+                  Old value = Old.A;
+              }
+              """,
+            """
+              import b.Outer;
+              import c.Nested;
+
+              class UnrelatedSameSimpleName {
+                  Nested other;
+                  Outer.Nested value = Outer.Nested.A;
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void changeToNestedTypeWhenOuterClassNameIsTaken() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Old", "b.Outer$Nested", null)),
+          java(nestedTargetOuter, SourceSpec::skip),
+          java(flatOld, SourceSpec::skip),
+          java(
+            """
+              package c;
+              public class Outer {}
+              """,
+            SourceSpec::skip
+          ),
+          java(
+            """
+              import a.Old;
+              import c.Outer;
+
+              class Test {
+                  Outer other;
+                  Old value = Old.A;
+              }
+              """,
+            """
+              import c.Outer;
+
+              class Test {
+                  Outer other;
+                  b.Outer.Nested value = b.Outer.Nested.A;
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void changeFromNestedTypeToFlatType() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Outer$Nested", "b.New", null)),
+          java(nestedOriginalOuter, SourceSpec::skip),
+          java(
+            """
+              package b;
+              public enum New { A }
+              """,
+            SourceSpec::skip
+          ),
+          java(
+            """
+              import a.Outer.Nested;
+
+              class Test {
+                  Nested value = Nested.A;
+              }
+              """,
+            """
+              import b.New;
+
+              class Test {
+                  New value = New.A;
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void changeFromNestedTypeToNestedType() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Outer$Nested", "b.Other$Inner", null)),
+          java(nestedOriginalOuter, SourceSpec::skip),
+          java(
+            """
+              package b;
+              public class Other {
+                  public enum Inner { A }
+              }
+              """,
+            SourceSpec::skip
+          ),
+          java(
+            """
+              import a.Outer.Nested;
+
+              class ViaImport {
+                  Nested value = Nested.A;
+              }
+              """,
+            """
+              import b.Other;
+
+              class ViaImport {
+                  Other.Inner value = Other.Inner.A;
+              }
+              """
+          ),
+          java(
+            """
+              import a.Outer;
+
+              class ViaOuterClass {
+                  Outer.Nested value = Outer.Nested.A;
+              }
+              """,
+            """
+              import b.Other;
+
+              class ViaOuterClass {
+                  Other.Inner value = Other.Inner.A;
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void changeToDeeplyNestedType() {
+        rewriteRun(
+          spec -> spec.recipe(new ChangeType("a.Old", "b.A$B$C", null)),
+          java(deeplyNestedTarget, SourceSpec::skip),
+          java(flatOld, SourceSpec::skip),
+          java(
+            """
+              import a.Old;
+
+              class ViaImport {
+                  Old value = Old.A;
+              }
+              """,
+            """
+              import b.A;
+
+              class ViaImport {
+                  A.B.C value = A.B.C.A;
+              }
+              """
+          ),
+          java(
+            """
+              class ViaPackageQualifiedName {
+                  a.Old value = a.Old.A;
+              }
+              """,
+            """
+              import b.A;
+
+              class ViaPackageQualifiedName {
+                  A.B.C value = A.B.C.A;
               }
               """
           )
--- a/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
+++ b/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java
@@ -267,17 +267,11 @@ public class ChangeType extends Recipe {
                 }
 
                 JavaType.FullyQualified fullyQualifiedTarget = TypeUtils.asFullyQualified(targetType);
-                if (fullyQualifiedTarget != null) {
-                    JavaType.FullyQualified owningClass = fullyQualifiedTarget.getOwningClass();
-                    if (!topLevelClassnames.contains(getTopLevelClassName(fullyQualifiedTarget).getFullyQualifiedName())) {
-                        if (hasNoConflictingImport(sf)) {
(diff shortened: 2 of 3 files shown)


knutwannheden:
One behavior change worth calling out that isn't in the description: when the target's **outermost** class name collides with an existing import, output gets more verbose.

With `import c.Outer;` already present and `ChangeType("a.Old", "b.Outer$Nested", null)`:

```java
// main
import b.Outer.Nested;
import c.Outer;
Nested value = Nested.A;

// this PR
import c.Outer;
b.Outer.Nested value = b.Outer.Nested.A;
```

Both compile, so this is readability rather than correctness — but it's a real regression in a case the new tests don't cover.

The mechanism: [`hasNoConflictingImport`](https://github.com/openrewrite/rewrite/blob/7b8eb8a6215ea2acd2f0b52454a0ac0fe546d05d/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java#L707-L713) compares against `newType.getClassName()`, which for a nested target is the dotted `"Outer.Nested"`, while [the new call site](https://github.com/openrewrite/rewrite/blob/7b8eb8a6215ea2acd2f0b52454a0ac0fe546d05d/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java#L269-L276) now imports the top-level `Outer`. So the check sees no conflict, `AddImport` declines the collision anyway, and the reference falls back to fully qualified. The mismatch predates this PR (the old `addImport(owningClass)` had it too), but switching to the top-level class makes it reachable in a new case.

`changeToNestedType` covers a collision on the *nested* simple name (`c.Nested`), which works fine — the outer-name collision is the untested one. Worth adding as a test either way, and possibly worth having the conflict check use the simple name of the class actually being imported.

timtebeek:
Good catch, and reproduced exactly as you describe. Added `changeToNestedTypeWhenOuterClassNameIsTaken`
pinning it:

```java
import a.Old;                            import c.Outer;
import c.Outer;

class Test {                     →       class Test {
    Outer other;                             Outer other;
    Old value = Old.A;                       b.Outer.Nested value = b.Outer.Nested.A;
}                                        }
```

against `main`:

```java
import b.Outer.Nested;
import c.Outer;

class Test {
    Outer other;
    Nested value = Nested.A;
}
```

So it is more verbose, as you said, and both compile.

On the mechanism — your reading is right, and it is worth separating the two halves. The
`hasNoConflictingImport` mismatch is real: it compares `newType.getClassName()`, the dotted
`"Outer.Nested"`, so it never fires for a nested target and the collision is instead caught one
layer down, where `AddImport.checkImportsForType` returns `IMPORT_AMBIGUITY` and
`FullyQualifyTypeReference` rewrites the reference. The *outcome* is correct, just reached by a
different route than the code suggests.

I would rather not change the conflict check to use the simple name of the class actually being
imported in this PR, though. Once it detects the collision, the enclosing block skips `addImport`
entirely — which yields the same fully qualified reference, only without the ambiguity path having
run — so it is a refactor of how we get there rather than a change in output. It also touches the
non-nested path, where `getClassName()` and the simple name coincide today but the check is shared
with the `visitIdentifier` fallback. Happy to do it as a follow-up if you want the two aligned.

The verbosity itself follows from the direction under discussion rather than from a bug: if the
outer name cannot be bound, `Outer.Nested` is not available, so fully qualifying is the only option
left that is consistent with "the import added matches the reference written". `main` gets the
shorter output precisely by importing the nested type, which is the shape this PR moves away from.
If we land the alternative direction instead (#8398), this case regains `Nested` for free.

timtebeek:
Was your intention just to see the conflict case documented as a limitation? Or did you intend to see changes to the recipe too? Not immediately clear from the above, and not sure what change I'd even opt to make if there are conflicts.

PR #8406, opened 2026-08-05
Go: distinguish integer widths instead of collapsing all to int

bmuschko:
## Problem

Every Go integer type (`int`, `int8` through `int64`, `uint` through `uint64`, `uintptr`) was attributed as the same LST type, a `JavaType.Primitive` with keyword `int`, so recipe authors could not distinguish an `int` from an `int64` through the type system, which makes type-narrowing recipes unsafe. For example, a recipe rewriting `fmt.Sprintf("%d", n)` to `strconv.Itoa(n)` cannot be written correctly, because `Itoa` requires an `int` and the recipe has no way to know whether `n` is an `int` or an `int64`, so it produces non-compiling output when `n` is an `int64`. The same limitation blocked rewriting `strconv.ParseInt(s, 10, 0)`, which returns `int64`, to `strconv.Atoi(s)`, which returns `int`.

## Fix

Each Go integer width now maps to a distinct LST type, following the approach the C# and Kotlin ports already use. Widths that Java's primitive set can express map to the matching primitive keyword (`int64` to `long`, `int16` to `short`, `int8` to `byte`, `int` and `int32` to `int`), and the unsigned widths, which have no Java primitive, map to a synthetic named type keyed by their Go name (`uint`, `uint64`, `uintptr`, and so on), the same modeling that Go `map` and `chan` already use. This keeps the attribution safe to serialize to the Java side, since every primitive keyword is one that Java understands and the unsigned types travel as ordinary named types. `float32` now maps to `float` rather than `double` for the same reason. While auditing the mapper for types that fell through to `J.Unknown`, two related gaps were closed as well: Go 1.22+ type aliases, including the predeclared `any`, are now unwrapped to their target type, and `unsafe.Pointer` maps to a named type. The matcher helpers `IsInt`, `IsNumeric`, and the method-pattern type resolver were updated to match.

### DIFF
--- a/rewrite-go/pkg/matcher/type_utils.go
+++ b/rewrite-go/pkg/matcher/type_utils.go
@@ -124,11 +124,9 @@ func IsNumeric(t java.JavaType) bool {
 			return true
 		}
 	}
-	fqn := GetFullyQualifiedName(t)
-	switch fqn {
-	case "int", "int8", "int16", "int32", "int64",
-		"uint", "uint8", "uint16", "uint32", "uint64", "uintptr",
-		"float32", "float64", "byte", "rune":
+	// Go's unsigned integers have no Java primitive; they map to named types.
+	switch GetFullyQualifiedName(t) {
+	case "uint", "uint8", "uint16", "uint32", "uint64", "uintptr":
 		return true
 	}
 	return false
@@ -137,16 +135,13 @@ func IsNumeric(t java.JavaType) bool {
 func IsInt(t java.JavaType) bool {
 	if p, ok := t.(*java.JavaTypePrimitive); ok {
 		switch p.Keyword {
-		case "int", "int8", "int16", "int32", "int64",
-			"uint", "uint8", "uint16", "uint32", "uint64", "uintptr",
-			"byte", "rune":
+		case "int", "long", "short", "byte":
 			return true
 		}
 	}
+	// Go's unsigned integers have no Java primitive; they map to named types.
 	switch GetFullyQualifiedName(t) {
-	case "int", "int8", "int16", "int32", "int64",
-		"uint", "uint8", "uint16", "uint32", "uint64", "uintptr",
-		"byte", "rune":
+	case "uint", "uint8", "uint16", "uint32", "uint64", "uintptr":
 		return true
 	}
 	return false
--- a/rewrite-go/pkg/matcher/matcher_test.go
+++ b/rewrite-go/pkg/matcher/matcher_test.go
@@ -92,6 +92,85 @@ func TestIsString(t *testing.T) {
 	}
 }
 
+func TestIsInt(t *testing.T) {
+	// Signed widths map to primitive keywords; unsigned widths to named types.
+	intLike := []java.JavaType{
+		&java.JavaTypePrimitive{Keyword: "int"},   // int, int32
+		&java.JavaTypePrimitive{Keyword: "long"},  // int64
+		&java.JavaTypePrimitive{Keyword: "short"}, // int16
+		&java.JavaTypePrimitive{Keyword: "byte"},  // int8, byte
+		&java.JavaTypeClass{FullyQualifiedName: "uint"},
+		&java.JavaTypeClass{FullyQualifiedName: "uint8"},
+		&java.JavaTypeClass{FullyQualifiedName: "uint16"},
+		&java.JavaTypeClass{FullyQualifiedName: "uint32"},
+		&java.JavaTypeClass{FullyQualifiedName: "uint64"},
+		&java.JavaTypeClass{FullyQualifiedName: "uintptr"},
+	}
+	for _, typ := range intLike {
+		if !IsInt(typ) {
+			t.Errorf("IsInt(%q) = false, want true", GetFullyQualifiedName(typ))
+		}
+	}
+	notInt := []java.JavaType{
+		&java.JavaTypePrimitive{Keyword: "double"},
+		&java.JavaTypePrimitive{Keyword: "String"},
+		&java.JavaTypePrimitive{Keyword: "boolean"},
+		nil,
+	}
+	for _, typ := range notInt {
+		if IsInt(typ) {
+			t.Errorf("IsInt(%q) = true, want false", GetFullyQualifiedName(typ))
+		}
+	}
+}
+
+func TestIsNumeric(t *testing.T) {
+	numeric := []java.JavaType{
+		// Signed widths, floats, and rune map to primitive keywords.
+		&java.JavaTypePrimitive{Keyword: "int"},    // int, int32
+		&java.JavaTypePrimitive{Keyword: "long"},   // int64
+		&java.JavaTypePrimitive{Keyword: "short"},  // int16
+		&java.JavaTypePrimitive{Keyword: "byte"},   // int8, byte
+		&java.JavaTypePrimitive{Keyword: "float"},  // float32
+		&java.JavaTypePrimitive{Keyword: "double"}, // float64
+		&java.JavaTypePrimitive{Keyword: "char"},   // rune
+		// Unsigned widths have no Java primitive, so they are named types.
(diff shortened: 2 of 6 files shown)


rewrite-go/pkg/matcher/type_utils.go
```diff
@@ -124,11 +124,9 @@ func IsNumeric(t java.JavaType) bool {
...
 		}
 	}
-	fqn := GetFullyQualifiedName(t)
-	switch fqn {
-	case "int", "int8", "int16", "int32", "int64",
-		"uint", "uint8", "uint16", "uint32", "uint64", "uintptr",
-		"float32", "float64", "byte", "rune":
+	// Go's unsigned integers have no Java primitive; they map to named types.
+	switch GetFullyQualifiedName(t) {
+	case "uint", "uint8", "uint16", "uint32", "uint64", "uintptr":
```
jkschneider, line 129: I'm not sure if it's better for them to be primitives + a marker? Does a go person think of these as primitive?

rewrite-go/pkg/matcher/type_utils.go
```diff
@@ -124,11 +124,9 @@ func IsNumeric(t java.JavaType) bool {
 			return true
 		}
 	}
-	fqn := GetFullyQualifiedName(t)
-	switch fqn {
-	case "int", "int8", "int16", "int32", "int64",
-		"uint", "uint8", "uint16", "uint32", "uint64", "uintptr",
-		"float32", "float64", "byte", "rune":
```
jkschneider, line 131: did rune get dropped, etc?

bmuschko: `rune` wasn't dropped: the `case "rune"` in the removed FQN switch was unreachable dead code, because a `rune` value is attributed as the `char` primitive (`rune` being Go's Unicode-code-point type, [an alias for `int32`](https://go.dev/ref/spec#Numeric_types)), and `GetFullyQualifiedName` returns a primitive's keyword (`char`), never the Go spelling `rune`, so nothing ever carried FQN `"rune"`; on top of that, the primitive-keyword switch runs first and returns `true` for `char` before the FQN switch is even reached.

The line that actually catches `rune` is the [`char` case in `IsNumeric`](https://github.com/openrewrite/rewrite/blob/c7915434951d1886804c391916f790f8d513b17f/rewrite-go/pkg/matcher/type_utils.go#L123), reached via the [`rune → char` mapping in `mapBasic`](https://github.com/openrewrite/rewrite/blob/c7915434951d1886804c391916f790f8d513b17f/rewrite-go/pkg/parser/type_mapper.go#L118-L119), both pre-existing and untouched by this PR. So the entire old FQN switch was actually dead before this change, since the previous mapper collapsed every integer to the `int` primitive and no type ever had FQN `"uint"`/`"rune"`/etc.

This PR only makes the unsigned FQN cases live for the first time (they now map to named types), keeps exactly those, drops the forever-dead rest, and adds a [test asserting `char` (rune) stays numeric](https://github.com/openrewrite/rewrite/blob/c7915434951d1886804c391916f790f8d513b17f/rewrite-go/pkg/matcher/matcher_test.go#L136) so it can't silently regress.

rewrite-go/pkg/matcher/type_utils.go
bmuschko: > Does a go person think of these as primitive?

Yes, a Go developer thinks of `uint64` as the same tier as `int64`; in Go's model these are all basic types ([`*types.Basic`](https://pkg.go.dev/go/types#BasicKind), where `Int64` and `Uint64` are sibling `BasicKind` constants), so representing `int64` as a `Primitive` but `uint64` as a `Class` is a real fidelity asymmetry worth acknowledging.

The reason that `JavaType.Primitive` isn't a general scalar-type concept but a fixed Java enum (the eight primitives plus `String`/`void`/`null`), so the practical question is whether a given width fits that enum, and the unsigned widths structurally can't, since Java has no unsigned or 64-bit-unsigned type. That's why the widths Java can name stay primitives (`int64`→`long`, `int16`→`short`, and so on) while the unsigned ones become synthetic named types, which is the same split [C#](https://github.com/openrewrite/rewrite/blob/0f17b3ccb5d67c9a815b8339d6bdeb7a34f2e322/rewrite-csharp/csharp/OpenRewrite/CSharp/CSharpTypeMapping.cs#L338-L350) and [Kotlin](https://github.com/openrewrite/rewrite/blob/0f17b3ccb5d67c9a815b8339d6bdeb7a34f2e322/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/tree/KotlinTypeUtils.java#L162-L175) already use for their unsigned and native-int types (each returns a primitive for the widths Java can name and `null` for the rest, which then falls through to a `JavaType.Class`).

> I'm not sure if it's better for them to be primitives + a marker?

A marker can't carry the distinction to where recipes actually query types: markers attach to LST nodes, but a method's parameter and return types and a variable's type are bare `JavaType` (and `JavaType.Primitive` is an enum with no per-instance slot), so a marker on the syntax node never reaches, say, an external method's resolved parameter types. To distinguish these through the type system the width has to live in the type itself, which is why a synthetic `JavaType.Class` (the same modeling as `map`/`chan`, and the choice [C#](https://github.com/openrewrite/rewrite/blob/0f17b3ccb5d67c9a815b8339d6bdeb7a34f2e322/rewrite-csharp/csharp/OpenRewrite/CSharp/CSharpTypeMapping.cs#L338-L350) and [Kotlin](https://github.com/openrewrite/rewrite/blob/0f17b3ccb5d67c9a815b8339d6bdeb7a34f2e322/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/tree/KotlinTypeUtils.java#L162-L175) already made) is the pragmatic answer.

jkschneider: I'm pretty sure we extend the Primitive type enum in other RPC languages.

bmuschko: I can't find one that extends the `Primitive` enum. Every RPC language uses the identical 12-constant Java set (`Boolean, Byte, Char, Double, Float, Int, Long, Short, Void, String, None, Null`):

| Port | Primitive definition | Extra kinds beyond Java? |
|---|---|---|
| JVM (rewrite-java) | [`enum Primitive`](https://github.com/openrewrite/rewrite/blob/637a0812bec3eb9c9435fd60976beeba0559d627/rewrite-java/src/main/java/org/openrewrite/java/tree/JavaType.java#L1163-L1175) | none — the 12 |
| C# | [`enum PrimitiveKind`](https://github.com/openrewrite/rewrite/blob/637a0812bec3eb9c9435fd60976beeba0559d627/rewrite-csharp/csharp/OpenRewrite/Java/JavaType.cs#L466-L480) | none — identical 12 (no `Decimal`/`UInt`/`ULong`/`SByte`) |
| Python | [`class Primitive(Enum)`](https://github.com/openrewrite/rewrite/blob/637a0812bec3eb9c9435fd60976beeba0559d627/rewrite-python/rewrite/src/rewrite/java/support_types.py#L367-L385) | none — identical 12 (`None_` is just a Python-keyword-safe spelling of `None`) |
| JavaScript | [`class Primitive`](https://github.com/openrewrite/rewrite/blob/637a0812bec3eb9c9435fd60976beeba0559d627/rewrite-javascript/rewrite/src/java/type.ts#L146-L160) | none — the constant Java calls `Long` is *renamed* `BigInt`, but it still carries keyword `'long'`; not a new kind |

The closest thing to an "extension" anywhere seems to be cosmetic. The JavaScript port [names its `Long` primitive `BigInt`](https://github.com/openrewrite/rewrite/blob/637a0812bec3eb9c9435fd60976beeba0559d627/rewrite-javascript/rewrite/src/java/type.ts#L159) (`static readonly BigInt = new Primitive('long')`), but it still carries the Java keyword `'long'`, so it's just an alias name for an existing primitive rather than a new kind.

If you have a specific spot in mind where a RPC language adds a primitive kind, could you point me at it?

PR #8412, opened 2026-08-06
Add `destinationFilename` option to `CopyFile`

steve-aom-elliott:
## What

Adds a new optional `destinationFilename` parameter to `org.openrewrite.CopyFile`. When set, the copy is written with that filename in the resolved destination folder instead of keeping the source file's name.

## Why

Today `copyTo` is always interpreted as a directory — the source filename is unconditionally appended. So `fileMatcher: **/application.yml` + `copyTo: test.yml` produces `.../test.yml/application.yml` (a folder named `test.yml`), which is surprising and offers no way to rename during a copy.

## Shape

- `destinationFilename` is nullable / optional. When unset, behavior is unchanged.
- Only allowed with `fileMatcher`. Combining it with `folder` is rejected in `validate()` — otherwise every matched file in a subtree would collide onto the same destination name.
- `MoveFile.computeNewSourcePath` (shared with `CopyFile`) gains an overload that threads the optional filename through and `.normalize()`s the result, so `copyTo: "."` cleanly means "same folder" without leaving a stray `./` segment.

## Example

```yaml
name: com.example.CopyApplicationYaml
recipeList:
  - org.openrewrite.CopyFile:
      fileMatcher: "**/application.yml"
      copyTo: "."
      destinationFilename: "test.yml"
```

Copies `src/main/resources/application.yml` → `src/main/resources/test.yml`, leaving the original in place.


timtebeek (approved):

timtebeek:
Looking at this again we seem to be racking up some mutually exclusive options; does it make sense to have a separate `CopyFolder` alongside `CopyFile`? That way it's clearer which recipe and options to use when.

rewrite-core/src/main/java/org/openrewrite/CopyFile.java
```diff
@@ -54,9 +54,18 @@ public class CopyFile extends ScanningRecipe<CopyFile.Accumulator> {
             example = "../yamls/")
     String copyTo;
 
+    @Nullable
```
jkschneider, line 57: I don't love the overlap between this and copyTo. What other designs do we have available?

steve-aom-elliott:
Going to accomplish this a different way and drop the CopyFile recipe for the time being instead.

PR #8414, opened 2026-08-06
MavenPomDownloader authentication fixes

sambsnyd:
`MavenPomDownloader` now treats a <server> with only <httpHeaders> (bearer token/PAT, no username+password) as authenticated, so those headers are actually sent on the retry after a 4xx.
`normalizeRepository` now retries its reachability probe with authentication before writing a host off as unreachable, instead of giving up on the anonymous failure. 

Incidentally fixed up a couple tests that were slow/disabled

### DIFF
--- a/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
+++ b/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
@@ -1049,19 +1049,17 @@ public class MavenPomDownloader {
     /**
      * Probes {@code url} to establish whether the repository host answers at all. Anonymous first, as every
      * other request is, but a host that refuses unauthenticated requests without answering them — a connection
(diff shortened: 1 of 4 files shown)


rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java
```diff
@@ -1039,6 +1032,38 @@ MavenRepository normalizeRepository(MavenRepository repository) throws Throwable
...
+                uri.replaceFirst("[hH][tT][tT][pP]://", "https://") :
+                uri;
+        return httpsUri.endsWith("/") ? httpsUri : httpsUri + "/";
+    }
+
+    /**
+     * Probes {@code url} to establish whether the repository host answers at all. Anonymous first, as every
+     * other request is, but a host that refuses unauthenticated requests without answering them — a connection
+     * reset or a dropped TLS handshake rather than a 401 — must not be written off as unreachable until tried with credentials. A repository declared to be behind authentication is often exactly the
+     */
```
timtebeek, line 1055: Unfinished sentence here

timtebeek (approved):

Issue #105, opened 2020-12-08
Parser failure for annotated parameter


jkschneider:
```java
public enum AppMessageCriterion {
    public Matcher<Install> getMatcher(Manager manager, FeatureFlags featureFlagManager, List<Install> installs,
      @Nullable List<String> arguments) {
    }
}
```

> 15:26:10.225 [main] ERROR org.openrewrite.java.Java11ParserVisitor - Failed to convert JCVariableDecl for the following cursor stack:
> 15:26:10.226 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- BEGIN PATH ---
> 15:26:10.226 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCCompilationUnit(sourceFile = OMITTED)
> 15:26:10.234 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCClassDecl(name = AppMessageCriterion, line = 55)
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCMethodDecl(line = 153)
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- END PATH ---
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - Failed to convert JCMethodDecl for the following cursor stack:
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- BEGIN PATH ---
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCCompilationUnit(sourceFile =  OMITTED)
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCClassDecl(name = AppMessageCriterion, line = 55)
> 15:26:10.240 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- END PATH ---
> 15:26:10.241 [main] ERROR org.openrewrite.java.Java11ParserVisitor - Failed to convert JCClassDecl for the following cursor stack:
> 15:26:10.241 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- BEGIN PATH ---
> 15:26:10.241 [main] ERROR org.openrewrite.java.Java11ParserVisitor - JCCompilationUnit(sourceFile = OMITTED)
> 15:26:10.241 [main] ERROR org.openrewrite.java.Java11ParserVisitor - --- END PATH ---
[...]

Issue #21, opened 2020-08-15
Kotlin support


changgengli:
Is there any plan to add support to kotlin method usage search and refactor?

jkschneider:
@changgengli Yes there is! To be honest, we plan to do Groovy first to quickly get Gradle remediation (because dependency updates are so important). But now that Kotlin has a new compiler AST, it is more approachable than it was a few years ago.

sambsnyd:
Will be tracking adding kotlin support on our longer-term roadmap, here: https://github.com/openrewrite/rewrite-roadmap/issues/12

sschuberth:
>  To be honest, we plan to do Groovy first to quickly get Gradle remediation

Who needs Groovy if you can do Kotlin DSL with Gradle? 😉

timtebeek:
> Will be tracking adding kotlin support on our longer-term roadmap, here: https://github.com/openrewrite/rewrite-roadmap/issues/12

That link has either gone bad, or links to a non-public repository. Also could not find any [mention of Kotlin on the project board](https://github.com/orgs/openrewrite/projects/4/views/10?filterQuery=kotlin). Is there another place where progress is tracked? Just curious, no pressure. :)

Sineaggi:
@sambsnyd Can we get this issue reopened? There is currently no kotlin support in rewrite, nor any public issue tracking it.

siblznbitz:
Hey @Sineaggi , better late than never but this has shipped :)

Issue #12, opened 2020-07-30
Bug: Mistakenly adds the 'new' keyword


nimakarimipour:
At instantiation of an inner classes, specifically when we have a `.` expression with the `new` keyword (e.g.,`this.new B()`), Java8Parser adds the `new` keyword at the beginning of the expression.

Consider the following as the original code,
```java
class A {
  void bar(B b, C c) {}
  void foo() {
    bar(this.new B(), this.new C());
  }
  class B { }
  class C { }
}
```

And now just by parsing the above code without applying any change, I get the following output,
```java
class A {
  void bar(B b, C c) {}
  void foo() {
    bar(newthis.new B(), newthis.new C());
  }
  class B { }
  class C { }
}
```

Here is the code I used to produce the above output,

```java
JavaParser parser =
            Java8Parser.builder()
                    .relaxedClassTypeMatching(true)
                    .logCompilationWarningsAndErrors(false)
                    .build();


    String uri = "PATH_TO_FILE";
    ArrayList<Path> p = new ArrayList<>();
    p.add(Paths.get(uri));
    ArrayList<J.CompilationUnit> trees = (ArrayList<J.CompilationUnit>) parser.parse(p);
    J.CompilationUnit tree = trees.get(0);
    System.out.println(tree.print());
```

jkschneider:
Thanks @nimakarimipour. Hadn't ever seen this case!

PR #8422, opened 2026-08-07
Mirror the npm, PyPI and NuGet sidecars to the Code Genome Project

jkschneider:
## Why

`rewrite-javascript`, `rewrite-python` and `rewrite-csharp` each publish a **native sidecar** that the Java jar fetches at runtime from a version-pin resource — `npx --package=@openrewrite/rewrite@<v>`, `pip install openrewrite==<v>`, `dotnet tool install OpenRewrite.CSharp.Tool --version <v>`. Those go only to npmjs, PyPI and nuget.org. The Maven artifacts have dual-published to the Code Genome Project for a while, so a customer resolving recipes from CGP still reached out to three other hosts to actually *run* them.

This mirrors the sidecars too. Part of a set:

- **openrewrite/gh-automation#105 — must merge first** (adds the composite action and the reusable-workflow inputs this PR uses)
- moderneinc/codegenomeproject#159 (the serving half)

## How

Uploads go through gh-automation's `publish-cgp-packages` action, **inert unless AWS credentials are present** — so forks and PRs never touch S3.

- **PyPI and NuGet** hang off the reusable workflows' new `cgp_pypi_artifacts` / `cgp_nuget_artifacts` inputs. The AWS env this repo already forwards is workflow-level, so no secret changes are needed here.
- **npm** goes through `npm-publish.yml` instead, because that is where `$DIST_TAG` is decided — CGP's dist-tag pointer must agree with what `npm publish --tag` was given, so both read the same variable. The step runs **unconditionally after** the npm publish, including when npm skipped a duplicate, so versions already on npmjs still backfill into CGP.

No version handoff was needed: the action derives name and version from the artifact itself, so the existing `published-snapshot-version` → `-PnpmPublishVersion` plumbing is untouched.

## Two fixes along the way

**1. Release candidates claimed npm's `latest` tag.** `publish.yml` fires on `v<x.y.z>-rc.<n>` as well as `v<x.y.z>`, but the dist-tag step sent both to `latest` — so `8.86.0-rc.1` became `@openrewrite/rewrite@latest`, ahead of the actual latest release. An `-rc.` tag now goes to `next` alongside the snapshots. This matters beyond npm hygiene now that CGP's pointer is sourced from the same variable and would have faithfully mirrored the mistake.

**2. `csharpPack` reached `publish` only inside `if (hasProperty("nugetApiKey"))`.** A composite action can only upload what something else produced, so a build with CGP credentials but no NuGet key would have found `dist/` empty and silently mirrored nothing. It now packs — but does **not** push — when the NuGet key is absent and `AWS_ACCESS_KEY_ID` is set, mirroring the gate `RewriteCgpPublishPlugin` already uses. A local build with neither credential never invokes `dotnet pack`, so nothing changes for developers.

## Verification

`actionlint` clean; both build scripts verified to still configure (`./gradlew :rewrite-csharp:help`). Derived S3 keys were cross-checked against the codegenomeproject backfill scripts for a real artifact per ecosystem.

- **Not verified:** no upload has run against S3 yet. First real exercise is the next `ci` on main after gh-automation#105 lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,12 @@ jobs:
       java_version: |
         25
         21
+      # Mirror the sidecar packages `publish` just built into the Code Genome Project package
+      # repositories, alongside PyPI and nuget.org. Name and version are read from the wheel /
+      # .nupkg themselves, so no version handoff is involved. The npm sidecar is mirrored from
+      # npm-publish.yml instead, where the dist-tag is decided.
+      cgp_pypi_artifacts: rewrite-python/rewrite/dist/*
+      cgp_nuget_artifacts: rewrite-csharp/csharp/dist/*.nupkg
     secrets:
       gradle_enterprise_access_key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       sonatype_username: ${{ secrets.SONATYPE_USERNAME }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,9 +26,11 @@ jobs:
         25
         21
       # Mirror the sidecar packages `publish` just built into the Code Genome Project package
-      # repositories, alongside PyPI and nuget.org. Name and version are read from the wheel /
-      # .nupkg themselves, so no version handoff is involved. The npm sidecar is mirrored from
-      # npm-publish.yml instead, where the dist-tag is decided.
+      # repositories, alongside npmjs, PyPI and nuget.org. Name and version are read from the
+      # tarball / wheel / .nupkg themselves, so no version handoff is involved. Each of these is
+      # produced by a packaging task `publish` depends on unconditionally, so the mirror keeps
+      # working if any of the three public registry pushes is retired.
+      cgp_npm_artifacts: rewrite-javascript/build/distributions/openrewrite-rewrite-*.tgz
       cgp_pypi_artifacts: rewrite-python/rewrite/dist/*
       cgp_nuget_artifacts: rewrite-csharp/csharp/dist/*.nupkg
     secrets:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,14 +25,6 @@ jobs:
       java_version: |
         25
         21
-      # Mirror the sidecar packages `publish` just built into the Code Genome Project package
-      # repositories, alongside npmjs, PyPI and nuget.org. Name and version are read from the
-      # tarball / wheel / .nupkg themselves, so no version handoff is involved. Each of these is
-      # produced by a packaging task `publish` depends on unconditionally, so the mirror keeps
-      # working if any of the three public registry pushes is retired.
-      cgp_npm_artifacts: rewrite-javascript/build/distributions/openrewrite-rewrite-*.tgz
-      cgp_pypi_artifacts: rewrite-python/rewrite/dist/*
-      cgp_nuget_artifacts: rewrite-csharp/csharp/dist/*.nupkg
     secrets:
       gradle_enterprise_access_key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       sonatype_username: ${{ secrets.SONATYPE_USERNAME }}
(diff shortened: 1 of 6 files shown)


jkschneider:
**The red ✗ here is expected and not a defect — it clears when openrewrite/gh-automation#105 merges.**

This PR passes a new input (`cgp_pypi_artifacts` / `cgp_nuget_artifacts`) to `gh-automation/.github/workflows/ci-gradle.yml@main`. That input only exists on #105's branch, so Actions rejects the workflow before it starts — hence `startup_failure` with no job log rather than an ordinary test failure.

Confirmed directly: `cgp_nuget_artifacts` currently has 0 occurrences in `ci-gradle.yml` on `gh-automation@main`. (GitHub exposes no check annotation for a startup failure, so that is the definitive evidence.)

Merge #105 first, then re-run these checks.

jkschneider:
Pushed a follow-up that fixes a design flaw in how the CGP mirror was wired up.

**The flaw.** The CGP upload was a parasite on the public registry push — in every ecosystem the distributable existed only as a *side effect* of pushing it to a public registry:

- `rewrite-python/build.gradle.kts` — `publish` depended on `pythonPublish`, i.e. the twine upload. `dist/` was populated only by the task whose job is pushing to PyPI.
- `rewrite-csharp/build.gradle.kts` — `csharpPack` reached `publish` only via `csharpPublish` (the nuget.org push), plus an `else if (!System.getenv("AWS_ACCESS_KEY_ID").isNullOrEmpty())` arm added as a workaround.
- `rewrite-javascript/build.gradle.kts` — `npmPack` was not wired into `publish` at all; only `npm-publish.yml` invoked it, immediately before `npm publish`.

Given the stated direction — dual-publish to CGP now, stop publishing to the public registries later — that made CGP depend on exactly the thing that is going to be deleted. Drop a public push and the mirror silently uploads nothing (an empty artifact match warns, it does not fail).

**The fix.** Separate *building* the distributable from *pushing* it, so CGP and the public registry become two independent consumers of one build output:

```
publish ├─ pythonBuild / csharpPack / npmPack   always
        ├─ pythonPublish                        if pypiToken
        └─ csharpPublish                        if nugetApiKey
```

Retiring a public registry is then a deletion — drop its push and its secret — and CGP is untouched.

The npm mirror also moves out of `npm-publish.yml` into `ci.yml`/`publish.yml`, via gh-automation's new `cgp_npm_artifacts` input (openrewrite/gh-automation#105), which derives the dist-tag from the trigger instead of from the npmjs publish step. `npm-publish.yml` is now purely an npmjs concern; a comment records that it exists only for npm's OIDC Trusted Publishing filename binding and can be deleted wholesale when npmjs publishing is retired. The PR/branch guard and the `-rc.`→`next` fix in it are untouched.

**Evidence, not assertion.** `./gradlew publish --dry-run` with `JAVA_HOME=25.0.3-zulu` (`SKIPPED` is `--dry-run`'s marker for "would have run", so presence == in the task graph):

*Before, with neither credential present:*
```
:rewrite-csharp:   (nothing — csharpPack absent entirely; the mirror would have found an empty dist/)
:rewrite-python:pythonBuild SKIPPED      <- only because pythonPublish drags it in
:rewrite-python:pythonPublish SKIPPED
```

*After, without the registry credential:*
```
:rewrite-python:pythonBuild SKIPPED      (no setupPypirc, no pythonPublish)
:rewrite-csharp:csharpPack SKIPPED       (no csharpPublish)
:rewrite-javascript:npmPack SKIPPED
```

*After, with the credential (`-PpypiToken=…` / `-PnugetApiKey=…`):* `pythonPublish` and `csharpPublish` reappear.

`recordPublishedSnapshotVersion` ordering still holds — `npmPack` (#6) → `jar` (#12) → `publish` (#20) → `recordPublishedSnapshotVersion` (#21) — and its `check(embedded == recordedVersion)` compares the freshly built jar against `datedSnapshotVersion`, neither of which this change touches. `datedSnapshotVersion` here is derived from the git commit timestamp, so the tarball `ci` packs and the one `npm-publish.yml` packs carry the identical version.

`:rewrite-python:help`, `:rewrite-csharp:help` and `:rewrite-javascript:help` all BUILD SUCCESSFUL; `actionlint` clean on `ci.yml`, `publish.yml` and `npm-publish.yml`.

The `startup_failure` on this PR is still expected until gh-automation#105 merges — `cgp_npm_artifacts` is one more input that only exists on that branch.

jkschneider:
**Follow-up: the credential gates are now blank-aware.** The decoupling commit gated each public push on its own registry credential, so that retiring a registry is deleting a step and a secret. Deleting the secret would not actually have stopped the push: `publish-gradle.yml` sets `ORG_GRADLE_PROJECT_pypiToken` / `ORG_GRADLE_PROJECT_nugetApiKey` **unconditionally** at workflow `env:` level, so an absent secret still renders as an empty string, Gradle defines the property, and `hasProperty()` is true for `""`. The push would have run with an empty credential and failed at twine / `dotnet nuget push`. Both gates now treat blank as absent (the idiom `recipes-csharp` already used), and `csharpPublish`'s `doFirst` guard goes through the same helper so guard and gate cannot disagree. No-op today — secrets are present, both pushes run as before. Proven with `publish -m` (`SKIPPED` = in the task graph): blank -> `pythonBuild`, `csharpPack` only; present -> also `setupPypirc`, `pythonPublish`, `csharpPublish`.

PR #8435, opened 2026-08-10
Python: make `TyTypesClient` notice when `ty-types` dies

knutwannheden:
## Motivation

`TyTypesClient` had no story for a ty-types process that dies mid-parse. It captured stderr but discarded it, it kept reporting a dead client as available, and a client the circuit breaker killed could never be revived.

The stderr handling is the thread to pull. In `--serve` mode ty-types writes to stderr from exactly one place — `eprintln!("Error reading stdin: {e}")`, immediately before the loop ends — plus whatever the default panic hook emits. There is no `tracing_subscriber`, no `env_logger`, no `panic::set_hook`, and no `catch_unwind`, so a healthy serve session produces zero bytes on stderr. That makes the one genuinely valuable payload — a panic message — the only thing the reader ever sees, and it went to `logger.debug`, where no deployment would surface it. Retaining the tail and logging it at WARNING when ty exits non-zero turns it into a diagnostic someone can act on.

Once stderr lands somewhere useful, acting on the EOF sentinel becomes worthwhile, and that pulls in the rest of the "ty died" behaviour.

## Summary

- The stderr reader thread buffers the last 50 lines in a `deque` rather than logging each line at DEBUG. `_note_exit` joins it with a 1s bound before reading — ty is already reaped by then, so stderr is at EOF — and logs the tail at WARNING on a non-zero exit.
- The EOF sentinel now reaps the process and clears `_initialized`, so `is_available` stops claiming a dead client is usable. Previously callers kept issuing requests against a dead process, each of which could only time out.
- `_drain_stdout` enqueues only JSON objects. A bare `null` was previously indistinguishable from the EOF sentinel, and a JSON array reached `response.get('id')` and raised `AttributeError` out of `_send_request` into the parse, dropping that file from the results.
- `initialize` restarts whenever there is no live process, rather than gating on `_initialized`. It polls for liveness because only `_kill` and `shutdown` detach `_process`: a process that exited on its own is still attached, so a null check alone reaches neither the breaker-killed client nor the self-exited one. Restarting stays explicit — nothing auto-restarts, because re-`initialize` re-indexes the whole project.
- The fake server no longer writes 512 bytes of stderr per request. That modelled a chattiness the real binary does not have; it now writes to stderr only on the "die" path.

## Test plan

- [x] Four new tests, each failing on `main` for the reason it names: `test_process_death_marks_the_client_unavailable` (asserts `is_available` and the WARNING carrying ty's stderr), `test_initialize_revives_a_killed_client`, `test_initialize_revives_a_client_whose_process_died`, and `test_junk_stdout_lines_are_skipped` (which on `main` also errors in teardown with the `AttributeError`).
- [x] `pytest tests/python/test_ty_client.py` — 9 passed.
- [x] `pytest tests/python` — 1635 passed, 10 skipped.
- [x] `ruff check --select E501,F401` clean on both changed files.

## Not in scope

The tests still skip on Windows. Un-skipping them buys nothing today: `ci.yml` delegates to `openrewrite/gh-automation`'s `ci-gradle.yml`, which is `runs-on: ubuntu-latest` with no matrix, so the Python tests never run on Windows at all. The real remedy is a Windows job, which belongs in that repo rather than here.

A per-request deadline in ty-types itself would be the deeper fix — `run_session` is strictly serial with no timeout of any kind, so the client's timeout is the only bound in the system. With one, the reader thread, the queue, and the breaker could all go away.


rewrite-python/rewrite/src/rewrite/python/ty_client.py
```diff
@@ -109,28 +110,33 @@ def _start_process(self) -> None:
                 "ty-types is not installed. Ensure the ty-types binary is on PATH."
             )
 
+        # ty writes to stderr only as it dies, so the diagnostics are worth
+        # keeping; a file, unlike a pipe, never blocks the writer.
+        stderr_fd, stderr_path = tempfile.mkstemp(prefix='ty-types-', suffix='.log')
```
jkschneider: Are we sure we want to create a log file even in a temp dir?

PR #8440, opened 2026-08-10
C#: type attribution for qualified names, arrays, members, and attribute values; marker and RPC fixes; MaybeAddUsing

macsux:
Type attribution and marker fixes in the C# SDK, found while porting the ~80 [WpfAnalyzers](https://github.com/DotNetAnalyzers/WpfAnalyzers) diagnostics to OpenRewrite recipes. Each item is an independent fix; they are separate commits and can be reviewed in order.

### Type attribution

- **Qualified type names carried no type.** `VisitQualifiedName` built the `FieldAccess` and its terminal `Identifier` with `null`, so `System.Windows.DependencyProperty x = ...` was unattributed while the simple form was fine. Affects any recipe type-checking a qualified name.
- **Array types carried no type.** `J.ArrayType` was constructed with `type: null`, so `typeof(int[])` and every array-typed declaration were unattributed. Inner ranks of a jagged type stay null — the syntax nests ranks in the opposite order from the symbol.
- **Property attribute lists were dropped.** `VisitPropertyDeclaration` never read `node.AttributeLists`, so a property's attributes were absorbed into the following modifier's prefix whitespace: no `Annotation`, no `AnnotatedStatement`, invisible to recipes.
- **Declared members and methods are now populated** on mapped types (previously always `null`, so "does type T declare member X?" was unanswerable — including for types from referenced assemblies). Declared-only, matching Java; inherited members remain reachable via `Supertype`/`Interfaces`. Compiler-generated members (backing fields, accessors, static ctors) are filtered out. Population is queued and drained iteratively after the shell is cached, so the transitive walk cannot recurse unboundedly.
- **Event symbols are attributed** as `JavaType.Variable`, consistent with how `PopulateMembers` already listed them.
- **Attribute values are now in the type model.** `Annotations` is populated on `Class`, `Method` and `Variable` from `AttributeData`, building `JavaType.Annotation` with `ElementValue`s — porting Java's `listAnnotations`/`annotationType`/`annotationElementValue`. Named arguments resolve `Element` to the attribute's property; positional arguments resolve to the constructor parameter's matching property, falling back to the constructor (Java's `getElement()` contract is non-null). Mixed `object[]` and nested-array arguments are skipped rather than mapped wrongly.

### RPC

- **Annotation identity keys now match Java's signature.** The C# `TypeSignature` returned `@Type` while `DefaultJavaTypeSignatureBuilder.annotationSignature` produces `@Type(elem=value,...)`; repeated same-typed annotations (e.g. several `[TemplatePart]` on one control) collapsed in the sender's delta cache. The `Values` list key now includes the value for the same reason. Sender-local — no wire-format change.
- **`J.Literal.Value` arrived as a raw `JsonElement`**, so `Literal { Value: string }` patterns matched in-process but never over RPC. Now materialized to the scalar type.
- **`RpcSendQueue.RegisterJavaTypeName` was not thread-safe** — a `static readonly Dictionary` mutated from parallel threads, producing sporadic `IndexOutOfRangeException`. Now a `ConcurrentDictionary`.

### Markers

Three paths silently discarded markers, which is invisible for a search recipe because the marker *is* the output:

- **`Markup.AddMarker` no-opped on `Cs.ExpressionStatement`**, which delegates `Prefix`/`Markers` to its wrapped expression but defined neither wither. It now has both, and the reflection path throws rather than returning the tree unchanged. Safe: a contract test asserts every concrete type exposes the wither.
- **`SearchResult.Found` had the same fallback.** `ParseError` gained `WithMarkers` (the only `Tree` implementation missing it), the contract test was widened from `J` to all of `Tree`, and the fallback now throws.
- **`CSharpPrinter` dropped markers on a class name** by appending `SimpleName` instead of visiting the `Identifier`. Output is byte-identical for unmarked code.

`DirectiveBoundaryInjector` carried a private copy of the same silent-fallback helper; it now uses `J.SetMarkers`.

### New API

- **`CSharpVisitor.MaybeAddUsing`** and the `AddUsing` after-visitor — the C# counterpart of `maybeAddImport`/`AddImport`, which had no equivalent. Idempotent, adds only when the type is referenced by simple name, dedups against existing plain and `global` usings, declines when the simple name is already bound to a different type (CS0104), and matches the file's existing placement and sort convention (compilation-unit vs block-scoped namespace, `System`-first or ordinal, insertion after a file header).

### Testing

`RewriteTest` compiled each C# source into its own compilation, so a type declared in one file was unattributed in another and multi-file `ScanningRecipe` tests were silently meaningless. All C# sources in a run now share one compilation, matching how a project is really parsed.

Coverage added throughout, including an RPC round-trip test that sends a type carrying two same-typed value-bearing annotations through a live Java peer and asserts the values survive in both directions. Suite: 2362 passing.

### DIFF
--- a/rewrite-csharp/csharp/OpenRewrite/CSharp/CSharpTypeMapping.cs
+++ b/rewrite-csharp/csharp/OpenRewrite/CSharp/CSharpTypeMapping.cs
@@ -238,12 +238,101 @@ internal class CSharpTypeMapping
             ? symbol.Interfaces.Select(i => (JavaType.FullyQualified)MapType(i)!).ToList()
             : null;
 
-        // For now, skip members and methods to avoid excessive traversal
-        // These can be populated lazily or in a future enhancement
         cls.UnsafeSet(flags, kind, fqn, typeParameters, supertype, owningClass,
             null, interfaces, null, null);
 
+        // Members/methods are populated only once the shell above is in the cache, so a member
+        // whose signature refers back to the declaring type resolves to that shell instead of
+        // recursing forever. Population itself is queued rather than nested (see DrainMemberQueue)
+        // so that the transitive member -> signature type -> its members walk stays iterative.
+        _memberQueue.Enqueue((symbol, cls));
+        DrainMemberQueue();
+
         return cls;
+    }
+
+    /// <summary>
+    /// Types whose <see cref="JavaType.Class.Members"/>/<see cref="JavaType.Class.Methods"/>
+    /// still need filling in. Populating a type's members maps the member signature types, which
+    /// in turn queues those types; draining iteratively keeps the recursion depth bounded by the
+    /// supertype/interface/type-argument graph rather than by the member graph.
+    /// </summary>
+    private readonly Queue<(INamedTypeSymbol Symbol, JavaType.Class Class)> _memberQueue = new();
+
+    private bool _draining;
+
+    private void DrainMemberQueue()
+    {
+        if (_draining) return;
+        _draining = true;
+        try
+        {
+            while (_memberQueue.Count > 0)
+            {
+                var (symbol, cls) = _memberQueue.Dequeue();
+                PopulateMembers(symbol, cls);
+            }
+        }
+        finally
+        {
+            _draining = false;
+        }
+    }
+
+    /// <summary>
+    /// Maps the members a type <em>declares</em> (inherited members stay reachable through
+    /// <see cref="JavaType.Class.Supertype"/>/<see cref="JavaType.Class.Interfaces"/>, matching
+    /// the Java type mapping). Compiler-generated members (auto-property backing fields, property
+    /// and event accessors, static initializers, and anything with a mangled name) are skipped.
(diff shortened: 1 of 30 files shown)


rewrite-csharp/csharp/OpenRewrite/CSharp/CSharpTypeMapping.cs
```diff
@@ -238,14 +238,103 @@ private JavaType MapNamedType(INamedTypeSymbol symbol)
...
+        // whose signature refers back to the declaring type resolves to that shell instead of
+        // recursing forever. Population itself is queued rather than nested (see DrainMemberQueue)
+        // so that the transitive member -> signature type -> its members walk stays iterative.
+        _memberQueue.Enqueue((symbol, cls));
+        DrainMemberQueue();
+
         return cls;
     }
 
+    /// <summary>
```
jkschneider, line 251: Let's make an effort to cut the doc essays, especially when the behavior is obvious.

PR #8441, opened 2026-08-10
Add regression for parent-managed property override upgrade

### LINKED ISSUE #7436: UpgradeDependencyVersion does not update an override property when the dependency is BOM-managed

`rewrite-maven` from current `main` (observed against `7.x` line).

## What is the problem?

`UpgradeDependencyVersion` does not update a user-pom property override when the targeted dependency is BOM-managed AND the override property is lower than the BOM's managed value. The dependency is left silently pinned below what the BOM provides.

The relevant gate in [`UpgradeDependencyVersion.java`](https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/UpgradeDependencyVersion.java#L393) is:

```java
if (dm != null && isProperty(dm.getRequested().getVersion()) && dm.getRequestedBom() == null) {
    // ... ChangePropertyValue path
}
```

The `dm.getRequestedBom() == null` clause excludes the case where dependency management came from an imported/inherited BOM, even when the user pom redeclares the property in its own `<properties>` block. The recipe falls through to the "add an explicit `<version>`" branch — which is then frequently stripped by `RemoveRedundantDependencyVersions` downstream — and the override property is left orphaned at the old value.

## Reproducer

A Spring Boot Kotlin project pinning `<kotlin.version>` below the version managed by the Spring Boot BOM:

```xml
<project>
  <groupId>com.example</groupId>
  <artifactId>kotlin-app</artifactId>
  <version>1.0-SNAPSHOT</version>
  <parent>
    <version>4.0.5</version>
    <relativePath/>
  </parent>
  <properties>
    <!-- Lower than the 2.2.21 the BOM manages -->
    <kotlin.version>2.1.21</kotlin.version>
  </properties>
  <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib</artifactId>
    </dependency>
  </dependencies>
</project>
```

Running:

```yaml
- org.openrewrite.java.dependencies.UpgradeDependencyVersion:
    groupId: org.jetbrains.kotlin
    artifactId: "*"
    newVersion: 2.2.x
    overrideManagedVersion: true
```

`<kotlin.version>` updated to `2.2.21` (or similar 2.2.x patch). All consumers — `kotlin-stdlib`, `kotlin-reflect`, `kotlin-maven-plugin` (whose pluginManagement entry in `spring-boot-dependencies` also resolves through `${kotlin.version}`) — flow to the new version.

## What did you see instead?

Either:

1. An explicit `<version>2.2.21</version>` is attached to each declared dependency and `<kotlin.version>2.1.21</kotlin.version>` is left orphaned (still overrides the BOM for any consumer that resolves through `${kotlin.version}` — e.g. `kotlin-maven-plugin`, transitive deps); or
2. The added `<version>` is stripped by `RemoveRedundantDependencyVersions` (because the BOM resolves to the same value), again leaving the orphan property in place.

Either outcome means the property still overrides the BOM at the lower value for any consumer the recipe doesn't directly touch.

## Suggested direction

Widen the property-update path so that BOM-managed dependencies whose version flows through a `${...}` placeholder in the user pom can also be upgraded by updating the property. A safety check is needed to avoid bumping a property that is shared by other consumers outside the scope of the upgrade — e.g. only update if every consumer of the property is in the recipe's match scope, or fall back to the existing "detach with explicit version" behavior when not safe.

## Workarounds today

- Add a `ChangePropertyValue` step targeting the specific property (downgrade hazard if the hardcoded value ever lags behind the BOM).
- Add a `RemoveProperty` step (only safe when no other consumers exist).
- Use `UpgradePluginVersion` for any plugins that carry an explicit `<version>${prop}</version>` reference — it follows the property and bumps it via `ChangePropertyValue`. Doesn't help dependencies (no plugin tag) or BOM-managed plugins with no explicit `<version>`.

kalayciburak:
## What's changed?

Adds a regression test for `UpgradeDependencyVersion` when a child POM redeclares a **parent-managed** version property below the managed value and `overrideManagedVersion` is `true`.

In that case the recipe should bump the local property (for example `flyway.version` under a Spring Boot parent) rather than leaving the override orphaned or relying on a redundant explicit `<version>` that `RemoveRedundantDependencyVersions` would strip.

## Why?

Extra coverage for the parent-property override path. This does **not** close #7436 — that issue was about BOM-defined versions, which can't be overridden the same way with local properties.

## Test plan

- [x] `./gradlew :rewrite-maven:test --tests org.openrewrite.maven.UpgradeDependencyVersionTest.upgradesExistingParentOverridePropertyInSamePom`

### DIFF
--- a/rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
+++ b/rewrite-maven/src/test/java/org/openrewrite/maven/UpgradeDependencyVersionTest.java
@@ -1278,6 +1278,60 @@ class UpgradeDependencyVersionTest implements RewriteTest {
         );
     }
 
+    @Test
+    void upgradesExistingParentOverridePropertyInSamePom() {
+        // Child pom redeclares a parent-managed version property below the managed value.
+        // overrideManagedVersion=true should bump that local property rather than leaving it orphaned
+        // (or adding a redundant explicit <version> that RemoveRedundantDependencyVersions would strip).
+        rewriteRun(
+          spec -> spec.recipe(new UpgradeDependencyVersion("org.flywaydb", "flyway-core", "10.15.0", "", true, null)),
+          pomXml(
+            """
+              <project>
+                  <parent>
+                      <groupId>org.springframework.boot</groupId>
+                      <artifactId>spring-boot-dependencies</artifactId>
+                      <version>3.3.0</version>
+                  </parent>
+                  <groupId>com.mycompany</groupId>
+                  <artifactId>my-child</artifactId>
+                  <version>1</version>
+                  <properties>
+                      <flyway.version>10.10.0</flyway.version>
+                  </properties>
+                  <dependencies>
+                      <dependency>
+                          <groupId>org.flywaydb</groupId>
+                          <artifactId>flyway-core</artifactId>
+                      </dependency>
+                  </dependencies>
+              </project>
+              """,
+            """
+              <project>
+                  <parent>
+                      <groupId>org.springframework.boot</groupId>
+                      <artifactId>spring-boot-dependencies</artifactId>
+                      <version>3.3.0</version>
+                  </parent>
+                  <groupId>com.mycompany</groupId>
+                  <artifactId>my-child</artifactId>
+                  <version>1</version>
+                  <properties>
+                      <flyway.version>10.15.0</flyway.version>
+                  </properties>
+                  <dependencies>
+                      <dependency>
+                          <groupId>org.flywaydb</groupId>
+                          <artifactId>flyway-core</artifactId>
+                      </dependency>
+                  </dependencies>
+              </project>
+              """
+          )
+        );
+    }
+
     @Issue("https://github.com/openrewrite/rewrite/issues/4193")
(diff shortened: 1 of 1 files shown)


steve-aom-elliott:
I think I may have missed slightly on the original reproducer as it doesn't actually hit the guard mentioned in the issue. When I dug back into it, I realised my mistake in that BOM-defined versions can't be overridden locally with your own properties, only parent ones.

Given the original reproducer was wrong, this doesn't really cover the scenario initially described by it, but is nonetheless a small increase in coverage

steve-aom-elliott:
@kalayciburak I would ask that you please rephrase the PR title and description if possible, given this, as the tests don't relate to BOM-defined versions, only parent-defined versions, and would not close the other issue, but simply add further coverage.

kalayciburak:
yeah makes sense — retitled/described as parent-managed property coverage only (not BOM, not closing #7436). renamed the test method too.

steve-aom-elliott (approved):

Issue #8424, opened 2026-08-07
Go switch: recipe edits crash the run via nil RPC baselines


bmuschko:
### What happened

When a Go recipe edits code inside or near a `switch` statement, the run can crash instead of producing the refactored source. Two distinct failures show up:

1. **Printer panic** on a type switch. Editing inside a `switch v := x.(type)` guard makes the run panic while printing, with a stack of `VisitSwitch → VisitTypeCast → VisitControlParentheses` (a nil dereference in `VisitControlParentheses`).
2. **RPC errors** (`END_OF_OBJECT`, `coerceToStatementRP`) on files that contain no type switch at all. These have been observed on real projects (for example gin `context_test.go` and gorilla `compression.go`).

In both cases the recipe never completes for the affected file.

### Root cause

Both failures are the same class of defect: a receiver deserializes a node with a hard-coded `nil` baseline. On the diff/patch path the JVM ships the changed parent as `CHANGE` but its untouched children as `NO_CHANGE`. Against a `nil` baseline the receive queue materializes a fresh, empty node, so every `NO_CHANGE` child resolves to its zero value. Downstream code then dereferences that zero value.

This is the same problem already handled for `if`/`for` bodies by `receiveBlockBody` in `rewrite-go/pkg/rpc/java_receiver.go`. A few receiver slots were missed.

### Status

**Fixed (PR to follow):**

- **Switch selector baseline** (`JavaReceiver.VisitSwitch`). The selector was received against `nil`; it now reconstructs a `ControlParentheses` baseline from the existing `Tag`, so a changed guard keeps its unchanged `(type)` wrapper (`TypeCast.Clazz`) instead of collapsing to nil. This resolves symptom 1.
- **Switch selector sender** (`JavaSender.VisitSwitch`). The synthetic `ControlParentheses` minted a fresh UUID for its ID and markers on every send, so a no-op selector was re-sent as a change each cycle and output was non-deterministic. IDs are now derived from the switch, and the trailing space before `{` is carried through instead of being dropped.

Regression tests cover the selector round trip on the diff path, sender determinism, and whitespace fidelity.

**Remaining:**

[...]

PR #8442, opened 2026-08-10
Fix nil pointer panic when printing Go type switches

bmuschko:
## Problem

Running a recipe on Go code that contains a type switch can crash the run instead of producing the refactored source. When a recipe edits inside a type-switch guard, the Go printer panics:

```go
switch v := x.(type) {
case int:
    return v
}
```

```
PANIC in Print: runtime error: invalid memory address or nil pointer dereference
  printer.(*GoPrinter).VisitControlParentheses  go_printer.go:974
  printer.(*GoPrinter).VisitTypeCast            go_printer.go:967
  printer.(*GoPrinter).VisitAssignment          go_printer.go:234
  printer.(*GoPrinter).VisitSwitch              go_printer.go:585
```

## Fix

Java models a switch selector as a `ControlParentheses`, but Go's `java.Switch` stores the tag directly, so `JavaReceiver.VisitSwitch` passed a `nil` baseline. On the diff path the JVM sends the changed selector as `CHANGE` and the untouched `(type)` wrapper (`TypeCast.Clazz`) as `NO_CHANGE`. Against a `nil` baseline that child collapses to `nil`, and the printer dereferences it.

- `JavaReceiver.VisitSwitch` reconstructs the `ControlParentheses` baseline from the existing `Tag`, so a changed selector's unchanged children keep their values. This mirrors `receiveBlockBody` for `if`/`for` bodies.
- `JavaSender.VisitSwitch` derives the synthetic wrapper's IDs from the switch instead of minting a fresh UUID per send, and carries the space before `{`.

- Part of #8424. The `END_OF_OBJECT` and `coerceToStatementRP` errors reported there are a separate baseline slot and remain open.

### DIFF
--- a/rewrite-go/pkg/rpc/java_receiver.go
+++ b/rewrite-go/pkg/rpc/java_receiver.go
@@ -529,8 +529,19 @@ func (r *JavaReceiver) VisitSwitch(sw *java.Switch, p any) java.J {
 	q := p.(*ReceiveQueue)
 	c := *sw // shallow copy to avoid mutating remoteObjects baseline
 	sw = &c
-	// selector - Java sends ControlParentheses, extract inner Expression for Tag
-	if cpResult := q.Receive(nil, func(v any) any { return r.Visit(v.(java.Tree), q) }); cpResult != nil {
+	// Java ships the selector as a ControlParentheses; reconstruct its baseline from the
+	// existing Tag so a changed selector's unchanged children keep their values,
+	// mirroring receiveBlockBody.
+	var selectorBaseline any
+	if sw.Tag != nil {
+		selectorBaseline = &java.ControlParentheses{
+			Tree: java.RightPadded[java.Expression]{
+				Element: sw.Tag.Element,
+				After:   sw.Tag.After,
+			},
+		}
+	}
+	if cpResult := q.Receive(selectorBaseline, func(v any) any { return r.Visit(v.(java.Tree), q) }); cpResult != nil {
 		if cp, ok := cpResult.(*java.ControlParentheses); ok {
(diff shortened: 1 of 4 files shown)


rewrite-go/pkg/rpc/java_receiver.go
```diff
@@ -529,8 +529,19 @@ func (r *JavaReceiver) VisitSwitch(sw *java.Switch, p any) java.J {
...
 	c := *sw // shallow copy to avoid mutating remoteObjects baseline
 	sw = &c
-	// selector - Java sends ControlParentheses, extract inner Expression for Tag
-	if cpResult := q.Receive(nil, func(v any) any { return r.Visit(v.(java.Tree), q) }); cpResult != nil {
+	// Java ships the selector as a ControlParentheses; reconstruct its baseline from the
+	// existing Tag so a changed selector's unchanged children keep their values,
+	// mirroring receiveBlockBody.
+	var selectorBaseline any
+	if sw.Tag != nil {
+		selectorBaseline = &java.ControlParentheses{
```
jkschneider, line 537: This creating a synthetic node thing in the receiver seems incorrect to me. I think the sender should send a ControlParentheses and the receiver should naturally receiver a ControlParentheses.

jkschneider (changes requested):

rewrite-go/pkg/rpc/java_receiver.go
bmuschko: It looks like https://github.com/openrewrite/rewrite/pull/8451 already superseded my PR in the meantime. I am closing my PR.

PR #8443, opened 2026-08-10
Preserve comments on composite literals with a trailing comma

bmuschko:
## Problem

When a recipe runs against Go source, a composite literal whose last element is followed by a trailing comma and a comment loses that comment, and the closing brace collapses onto the previous line.

Given this input:

```go
var x = &Context{
	queryCache: url.Values{"a": []string{"b"}},
	Request:    &http.Request{URL: validURL}, // valid request
}
```

the recipe writes back:

```go
var x = &Context{
	queryCache: url.Values{"a": []string{"b"}},
	Request:    &http.Request{URL: validURL}, }
```

The comment is gone and the `}` no longer sits on its own line. This is silent: the run reports success, so users cannot tell the change was corrupted. On real repositories (for example `gin-gonic/gin`'s `context_test.go`) it surfaces as unexpected diffs, and in related forms as internal errors while writing the patch:

```
expected CHANGE with positions, got END_OF_OBJECT (value=<nil>, valueType=<nil>)
coerceToStatementRP: element does not implement java.Statement (rp=<nil> elem=<nil> nil=true)
```

## Cause

A trailing comma is recorded with a `TrailingComma` marker whose `before` and `after` spaces hold the surrounding whitespace and any comment. Its RPC codec carried only each space's whitespace string, so comments were not sent across the wire.

## Fix

The `TrailingComma` codec now sends and receives the full `before` and `after` spaces, comments included, on both the Java and Go sides.

### DIFF
--- a/rewrite-go/src/main/java/org/openrewrite/golang/marker/TrailingComma.java
+++ b/rewrite-go/src/main/java/org/openrewrite/golang/marker/TrailingComma.java
@@ -17,8 +17,7 @@ package org.openrewrite.golang.marker;
 
 import lombok.Value;
 import lombok.With;
-import org.openrewrite.java.internal.rpc.JavaReceiver;
-import org.openrewrite.java.internal.rpc.JavaSender;
+import org.openrewrite.java.internal.rpc.JavaSpaceRpcCodec;
 import org.openrewrite.java.tree.Space;
 import org.openrewrite.marker.Marker;
 import org.openrewrite.rpc.RpcCodec;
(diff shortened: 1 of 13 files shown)


rewrite-go/src/main/java/org/openrewrite/golang/marker/TrailingComma.java
```diff
@@ -35,15 +37,15 @@ public class TrailingComma implements Marker, RpcCodec<TrailingComma> {
     @Override
     public void rpcSend(TrailingComma after, RpcSendQueue q) {
         q.getAndSend(after, Marker::getId);
-        q.getAndSend(after, t -> t.getBefore().getWhitespace());
-        q.getAndSend(after, t -> t.getAfter().getWhitespace());
+        q.getAndSend(after, TrailingComma::getBefore, space -> new JavaSender().visitSpace(space, q));
```
jkschneider: This looks pretty suspect, creating a new JavaSender() here.

bmuschko: Good call, Jonathan. I pulled the space serialization into shared `JavaSpaceRpcCodec.sendSpace`/`receiveSpace` helpers and pointed the visitors and all eight `Space`-carrying markers at them, so nothing constructs a throwaway `JavaSender`/`JavaReceiver` just to reach the queue anymore. To be upfront, I measured it and this is a readability/maintainability change rather than a perf win: the JIT was already eliding that allocation, so the real value is having one source of truth for the space codec and dropping the pattern you flagged. (Pushed in dd65e020.)

jkschneider: Huh? How do we need a new JavaSender in one case, and refactoring a method out in another?

bmuschko: I’d rather ask than guess again. Your first comment flagged `new JavaSender()` as suspect, so I pulled the space serialization into a static helper. But your follow-up reads like that extraction is itself the problem, since it special-cases `Space` and leaves `new JavaSender()` in place everywhere else.

So, in the first comment, which were you getting at:

1. we shouldn’t be constructing a `JavaSender` here at all, or
2. it’s fine as the shared convention, and singling out `Space` is the inconsistency?

bmuschko: @jkschneider Would you mind responding to my comment?

bmuschko:
This has already been address by https://github.com/openrewrite/rewrite/pull/8581.

PR #8453, opened 2026-08-11
Python: add space to negation unary operator

greg-at-moderne:
## What's changed?

Adding a safety mechanism to Python printer to add an artificial (not stored in LST) whitespace after the negation operator.

## What's your motivation?

Without the fix, Java recipes when executed against Python code might produce invalid code, e.g.
```python
while notx:
  x = check()
```
instead of
```python
while not x:
  x = check()
```

- In Java, the negation operator (`!`) doesn't require a space after it, i.e. you can write `!x`. Therefore the recipes don't bother with adding any space.
- In Python, the operator is `not` and requires a space in practice, otherwise it becomes `notx`, should be `not x`.

I am not fully sure, but I figured a safety mechanism in the printer might actually be helpful. It violates the idempotency expectation.

Alternative would be to have special code in all recipes adding negation (typically in rewrite-static-analysis) to have special handling for Python and add space there.

### DIFF
--- a/rewrite-python/rewrite/src/rewrite/python/printer.py
+++ b/rewrite-python/rewrite/src/rewrite/python/printer.py
@@ -1849,6 +1849,8 @@ class PythonJavaPrinter:
 
         if unary.operator == Unary.Type.Not:
             p.append("not")
+            if not unary.expression.prefix.whitespace:
+                p.append(" ")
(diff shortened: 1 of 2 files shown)


greg-at-moderne:
The actual solution involved #8540 instead.

PR #8458, opened 2026-08-11
Go: add cursor helpers for function-body and init-wrapped-if checks

bmuschko:
## What

Adds two reusable cursor predicates to `rewrite-go` `pkg/visitor`:

- `IsFunctionBodyBlock(c *Cursor) bool` — reports whether the block at the cursor is a function's body rather than a nested block such as a loop or `if` body.
- `IsInitWrappedIf(c *Cursor) bool` — reports whether the `If` at the cursor is the inner statement of a `golang.StatementWithInit` (an `if init; cond` form).

## Why

This is the follow-up from https://github.com/moderneinc/recipes-go/pull/53#discussion_r3759564329, where these helpers were flagged as generic enough to belong upstream rather than duplicated in a downstream recipe repo. Hosting them in the `visitor` package lets recipes across languages share them. The package already imports both `java` and `golang`, so no new import cycle is introduced.

## Not upstreamed

- **`BaseIndent`** — not needed. `rewrite-go` already exposes the identical logic as `Space.Indent()` in `pkg/tree/java/space.go`, so callers should use that instead of a duplicate.
- **`IsErrNotNil`** — stays local to the recipe repo, as it is recipe-specific rather than generic LST navigation.

### DIFF
--- a/rewrite-go/pkg/visitor/cursor_helpers_test.go
+++ b/rewrite-go/pkg/visitor/cursor_helpers_test.go
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2026 the original author or authors.
+ *
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package visitor
+
+import (
+	"testing"
+
+	"github.com/openrewrite/rewrite/rewrite-go/pkg/tree/golang"
+	"github.com/openrewrite/rewrite/rewrite-go/pkg/tree/java"
+)
+
+func TestIsFunctionBodyBlock(t *testing.T) {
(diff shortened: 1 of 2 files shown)


rewrite-go/pkg/visitor/cursor_helpers_test.go
```diff
@@ -0,0 +1,77 @@
...
+package visitor
+
+import (
+	"testing"
+
+	"github.com/openrewrite/rewrite/rewrite-go/pkg/tree/golang"
+	"github.com/openrewrite/rewrite/rewrite-go/pkg/tree/java"
+)
+
+func TestIsFunctionBodyBlock(t *testing.T) {
```
greg-at-moderne, line 65: I am wondering whether we should have more "end-to-end" tests - i.e. a test which would involve parsing a piece of code and assert on count of matches or similar.
As is, the test is rather trivial as it follows the implementation internals.

greg-at-moderne (commented):
Not sure what is meant by:

> Hosting them in the visitor package lets recipes across languages share them.

in the PR description. But feel free to ignore this comment.

bmuschko:
> Hosting them in the visitor package lets recipes across languages share them.

What I was trying to say is that by putting these helpers in the shared `pkg/visitor` package instead of inside a specific recipe package, any recipe can reuse them rather than each copying its own version (which is what happened in recipes-go).

rewrite-go/pkg/visitor/cursor_helpers_test.go
bmuschko: Makes sense. Changed.

greg-at-moderne (approved):

PR #8460, opened 2026-08-11
Don't bump a shared version property to a version its siblings lack

timtebeek:
`UpgradeDependencyVersion` wrote the new version into a shared `<properties>` entry without checking that the version exists for the *other* artifacts resolving through that property, leaving those siblings unresolvable.

The trigger observed in the wild is `UpgradeJackson_2_3`, which upgrades `com.fasterxml.jackson.core:jackson-annotations` to `2.21`. That version exists — `jackson-annotations` uses bare minor versions — but the rest of Jackson 2.x uses three-part versions, so every sibling sharing `${jackson.version}` was rewritten to a version that does not exist. A Moderne flagship run on ALL/Open Source/FINOS (changeset 20260811034058-h2uII, "Spring Boot 4.0 best practices") produced 102 WARNING error markers of the form "Unable to download POM: com.fasterxml.jackson.core:jackson-core:2.21", across `finos/legend-sdlc`, `finos/legend-depot` and `finos/legend-shared`. This change should remove those markers; the recipe's chosen version stays as it is, so please verify against a later flagship run on that org.

## Behaviour

The scan phase now records, per property, the dependency coordinates whose version resolves through it (across every POM in the project, including `dependencyManagement` and plugin dependencies). Before a property is written, every consumer is checked for the new version; a consumer that is a module of the build is skipped, and metadata that cannot be resolved is not treated as evidence that a version is missing, so a download failure never becomes a hard error or a silent skip.

When the version is not published for all consumers, the targeted dependency is **decoupled**: an explicit `<version>` is written on the dependency (or its `dependencyManagement` entry) and the property is left alone. That was the preferred option in the report, and it is also what the Gradle recipe already does for a shared version variable — `rewrite-gradle`'s `UpgradeDependencyVersionTest#doesNotCorruptSharedGradleProperty` asserts exactly this outcome on the same Jackson 2.21 case. The alternatives (skip with a `Markup.warn`, or only update when all consumers agree) leave the requested upgrade undone, and would have left Maven and Gradle behaving differently on identical input.

Single-consumer property updates are unchanged — that path still rewrites the property, now after confirming the version exists for the one artifact that uses it, which also keeps the decision stable across recipe cycles once a dependency has been decoupled.

## Gradle

Already handled, no change needed: `safeUpdatedVersion` in `org.openrewrite.gradle.UpgradeDependencyVersion` requires the agreed version to be published for every untargeted neighbour sharing the variable, covering `ext` blocks, Kotlin `val`s and `gradle.properties`. `:rewrite-gradle:test --tests "*UpgradeDependencyVersionTest"` is green.

Every write of a `<version>` tag now goes through one seam (`changeVersionTagValue`), so plugin `<dependencies>` and `<annotationProcessorPaths>` entries — which reach the same `ChangePropertyValue` path — are covered too, and are recorded as property consumers.

## Tests

Four tests in `rewrite-maven`'s `UpgradeDependencyVersionTest`, all failing before this change except the third:
- `doNotUpgradeSharedPropertyWhenNewVersionMissingForOtherConsumer` — the reported shape; `jackson-core` used to end up at a nonexistent `2.21`.
- `doNotUpgradeSharedManagedPropertyWhenNewVersionMissingForOtherConsumer` — same through `dependencyManagement`.
- `upgradeSharedPropertyWhenNewVersionExistsForOtherConsumer` — the property is still updated when the version resolves for both.
- `doNotUpgradeSharedParentPropertyWhenNewVersionMissingForOtherModule` — property in a local parent, consumers in two modules.
- `doNotUpgradeSharedPluginDependencyPropertyWhenNewVersionMissingForOtherConsumer` — a plugin dependency sharing the property with a regular dependency.
- `upgradeSharedPropertyNameDeclaredSeparatelyPerModule` — same property name declared independently in two unrelated modules must not block each other.

- Full `:rewrite-maven:test` (1440 tests) is green. Rebased onto #8455.


timtebeek:
Closing this in favour of #8656, which carries the full analysis, the reproduction and the reasoning behind the decouple-over-decline decision, so someone else can pick it up without reconstructing any of it.

Not deleting `tim/taipei-v2` — the implementation and its six tests are referenced from the issue as a starting point. Verified against `main` on 2026-08-26: merges clean, all six shared-property tests pass, and `UpgradeDependencyVersionTest` shows no regressions (`deriveFromNexus`, `deriveFromNexusUpgrade` and `badManagedVersion` fail identically on unmodified `main` in the same environment). On the real `finos/legend-depot` `dependencyManagement` block it leaves `${jackson.version}` alone and decouples `jackson-annotations` to an explicit `2.21`.

The one open question noted on the issue: the existence check currently treats an unresolvable metadata download as evidence the version *does* exist, so a network-degraded run quietly reverts to the old behaviour.

Issue #8465, opened 2026-08-11
Gradle ChangeDependency leaves an unversioned dependency when the new coordinates are unmanaged


timtebeek:
### What version of OpenRewrite are you using?

Latest `main` (560b40cc63).

### What is the smallest, simplest way to reproduce the problem?

The Gradle `org.openrewrite.gradle.ChangeDependency` leaves a renamed dependency without a version whenever the declaration had no version of its own, even when a `newVersion` is supplied. This is currently asserted by `ChangeDependencyTest#doNotPinWhenNotVersioned`:

```groovy
plugins {
  id 'java'
  id 'org.springframework.boot' version '2.6.1'
  id 'io.spring.dependency-management' version '1.0.11.RELEASE'
}
dependencies {
    runtimeOnly 'mysql:mysql-connector-java'   // -> 'com.mysql:mysql-connector-j'
}
```

Spring Boot 2.6.1 manages `mysql:mysql-connector-java` but not `com.mysql:mysql-connector-j`, so the resulting build script does not resolve. `overrideManagedVersion` is the documented escape hatch, and its own option description says "No check is done on the NEW dependency to verify if it is managed, it relies on whether the OLD dependency had a managed version."

- The Maven side behaves the other way round: it writes an explicit version when it can tell that the new coordinates are unmanaged (`ChangeDependencyGroupIdAndArtifactIdTest#managedToUnmanagedExternalizedDepMgmt`), and after #8462 it declines to rename when it has no version to write.

### What did you expect to see?

The Gradle recipe not producing a build script that cannot resolve, ideally matching the Maven decision.

### What did you see instead?

An unversioned dependency on coordinates nothing manages.

The obstacle is that the `GradleProject` model does not record which versions come from dependency management, so "is the new GA managed?" cannot be answered the way `ResolvedPom#getManagedVersion` answers it for Maven. `RemoveBomManagedDirectDependencies` works around this by downloading each `platform(...)` POM and querying it, which covers platforms but not the `io.spring.dependency-management` plugin.

### Are you interested in contributing a fix?

- Not immediately — filing to track the Maven/Gradle asymmetry uncovered while fixing #8462.

Issue #8464, opened 2026-08-11
Renaming coordinates without a newVersion carries the old version over to the new artifact


timtebeek:
### What version of OpenRewrite are you using?

Latest `main` (reproduced at 560b40cc63).

### What is the smallest, simplest way to reproduce the problem?

`ChangeDependencyGroupIdAndArtifactId` / `ChangeManagedDependencyGroupIdAndArtifactId` without a `newVersion` renames the coordinates of a `<dependencyManagement>` entry but keeps the version that belonged to the *old* coordinates:

```java
rewriteRun(
  spec -> spec.recipe(new ChangeDependencyGroupIdAndArtifactId(
    "javax.servlet", "javax.servlet-api", "jakarta.servlet", "jakarta.servlet-api", null, null)),
  pomXml(
    """
      <project>
          <groupId>com.mycompany.app</groupId>
          <artifactId>my-app</artifactId>
          <version>1</version>
          <dependencyManagement>
              <dependencies>
                  <dependency>
                      <groupId>javax.servlet</groupId>
                      <artifactId>javax.servlet-api</artifactId>
                      <version>4.0.1</version>
                  </dependency>
              </dependencies>
          </dependencyManagement>
          <dependencies>
              <dependency>
                  <groupId>javax.servlet</groupId>
                  <artifactId>javax.servlet-api</artifactId>
              </dependency>
          </dependencies>
      </project>
      """
  )
);
```

### What did you expect to see?

Either a version that exists for the new coordinates, or no change.

### What did you see instead?

`jakarta.servlet:jakarta.servlet-api:4.0.1`, which does not exist:

```
<!--~~(jakarta.servlet:jakarta.servlet-api:4.0.1 failed. Unable to download POM: jakarta.servlet:jakarta.servlet-api:4.0.1. Tried repositories:
https://repo.maven.apache.org/maven2: HTTP 404)~~>-->
```

- Version lines rarely carry over across a groupId change (`javax.*` → `jakarta.*` is the common case), so carrying the old version over is usually wrong. Same applies to a `<dependency>` with an explicit `<version>`. Related to #8462, which covers the case where the version is dropped entirely rather than carried over.

### Are you interested in contributing a fix?

Not immediately.

Issue #8463, opened 2026-08-11
Renaming a parent's dependencyManagement leaves child modules with a stale MavenResolutionResult


timtebeek:
### What version of OpenRewrite are you using?

Latest `main` (reproduced at 560b40cc63).

### What is the smallest, simplest way to reproduce the problem?

In a multi-module build, a child module declares a dependency without a `<version>` and the parent POM manages it. `ChangeDependencyGroupIdAndArtifactId` renames the dependency in the child and the `<dependencyManagement>` entry in the parent — deliberately not writing an explicit version into the child (see `providedDependencyManagedByLocalParentDoesNotGetExplicitVersion`, `parentAndBomManagedToJustParentManaged`).

That output is correct on disk, but the child document's `MavenResolutionResult` still holds the *parent's parse-time* model, in which the old coordinates are managed and the new ones are not. Any recipe that runs afterwards in the same run and calls `maybeUpdateModel()` on the child then re-resolves against that stale parent and fails:

```java
rewriteRun(
  spec -> spec.recipes(
    new ChangeDependencyGroupIdAndArtifactId("com.squareup.okhttp3", "mockwebserver", null, "mockwebserver3", "5.x", null),
    new RemoveDependency("org.awaitility", "awaitility", null)),   // any recipe that updates the model
  // parent POM: <dependencyManagement> for com.squareup.okhttp3:mockwebserver
  // child POM:  <dependency> com.squareup.okhttp3:mockwebserver with no <version>
);
```

### What did you expect to see?

No marker: after the run, the parent manages `mockwebserver3` and the child resolves.

### What did you see instead?

```xml
<!--~~(No version provided for direct dependency com.squareup.okhttp3:mockwebserver3:compile)~~>--><dependency>
    <groupId>com.squareup.okhttp3</groupId>
    <artifactId>mockwebserver3</artifactId>
</dependency>
```

Two consequences:

1. Error markers on every child module, on POMs that are actually fine. Observed in a Moderne SaaS run of "Spring Boot 4.0 best practices": 56 markers across `spring-cloud/spring-cloud-sleuth` (30) and `spring-projects-experimental/spring-cloud-sleuth-otel` (26), all from `UpgradeOkHttpMockWebServer -> ChangeDependency`. The full `spring-cloud-sleuth` POM chain (root → `tests` → `tests/brave` → module) reproduces it exactly.
[...]

PR #8466, opened 2026-08-11
ChangeDependency: don't rename a managed dependency into unmanaged coordinates without a version

### LINKED ISSUE #8462: ChangeDependency renames a managed dependency into unmanaged coordinates, leaving it without a version

Latest `main` (reproduced at 560b40cc63).

`ChangeDependency` / `org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId` renames a dependency that has no `<version>` of its own — its version comes from an imported BOM — into coordinates that the BOM does not manage. Because no `newVersion` is given, the recipe never considers the version at all, and the resulting POM does not resolve:

```java
rewriteRun(
  spec -> spec.recipe(new ChangeDependencyGroupIdAndArtifactId(
    "com.squareup.okhttp3", "mockwebserver", null, "mockwebserver3", null, null)),
  pomXml(
    """
      <project>
          <artifactId>my-app</artifactId>
          <version>1</version>
          <dependencyManagement>
              <dependencies>
                  <dependency>
                      <artifactId>spring-boot-dependencies</artifactId>
                      <version>2.7.18</version>
                      <type>pom</type>
                      <scope>import</scope>
                  </dependency>
              </dependencies>
          </dependencyManagement>
          <dependencies>
              <dependency>
                  <groupId>com.squareup.okhttp3</groupId>
                  <artifactId>mockwebserver</artifactId>
              </dependency>
          </dependencies>
      </project>
      """
  )
);
```

Either an explicit `<version>` on the renamed dependency, or no change — anything that leaves the POM resolvable.

### What did you see instead?

```xml
<!--~~(No version provided for direct dependency com.squareup.okhttp3:mockwebserver3:compile)~~>--><dependency>
    <groupId>com.squareup.okhttp3</groupId>
    <artifactId>mockwebserver3</artifactId>
</dependency>
```

Spring Boot 2.7 manages `com.squareup.okhttp3:mockwebserver`, but nothing manages `mockwebserver3`, so the rename orphans the version.

### Are you interested in contributing a fix?

Yes — PR incoming.

### Context

Seen in Moderne SaaS runs on repositories under `finos` (84 error markers with this signature across `legend-sdlc`, `legend-depot` and `legend-shared`) driven by Jakarta and Jackson migration recipes that call `ChangeDependency` without a `newVersion`, for `jakarta.servlet:jakarta.servlet-api`, `jakarta.ws.rs:jakarta.ws.rs-api`, `jakarta.validation:jakarta.validation-api`, `jakarta.inject:jakarta.inject-api`, `io.swagger.core.v3:swagger-annotations` and `tools.jackson.core:jackson-{core,databind}`.

Note that the same marker text also shows up for a *different* root cause, tracked separately: a child module whose parent's `<dependencyManagement>` is renamed keeps a stale `MavenResolutionResult`, so any later recipe that calls `maybeUpdateModel()` on that module re-resolves against the stale parent and reports this error even though the POMs on disk are consistent.

timtebeek:
- Fixes #8462.

### What's changed

When a `<dependency>` has no `<version>` of its own, its version comes from dependency management. Renaming its coordinates only produces a resolvable POM if the *new* coordinates end up managed as well. `ChangeDependencyGroupIdAndArtifactId` reasons about that when a `newVersion` is supplied — it writes an explicit `<version>` when it can tell the new coordinates are unmanaged — but with no `newVersion` the whole version block is skipped, so the rename silently orphans the version:

```
<!--~~(No version provided for direct dependency com.squareup.okhttp3:mockwebserver3:compile)~~>--><dependency>
    <groupId>com.squareup.okhttp3</groupId>
    <artifactId>mockwebserver3</artifactId>
</dependency>
```

For an unversioned dependency and no `newVersion`, the recipe now checks whether the new coordinates will have a managed version:

* already managed (very common for `javax.*` → `jakarta.*` under one BOM) — rename, unchanged;
* managed by dependency management that this recipe also renames — either up the project's parent chain, or in a BOM module belonging to the same repository — rename, and defer the model update to the POM that owns that dependency management, as the parent case already did;
* otherwise — leave the dependency alone and attach a warning saying why.

- The last case is the fix. Nothing can invent a version here: carrying the old artifact's version over is what produces `jakarta.servlet:jakarta.servlet-api:4.0.1` (#8464), and picking the latest release is a version bump nobody asked for. Not writing a POM that fails to resolve, and saying so, is the honest outcome — the file stays buildable and the marker is actionable. The alternative of adding a `<dependencyManagement>` entry instead of a `<version>` has the same missing-version problem, one level up.

To spot BOM modules belonging to the same repository, the scanner (which previously no-opped without a `newVersion`) now records which POMs declare dependency management matching the old coordinates; that set is matched against `ResolvedManagedDependency#getBomGav`. Without it, a repository with its own `*-dependencies` BOM module — the `spring-cloud-sleuth` layout — would have had working renames turned into no-ops.

Nothing changes for a dependency with an explicit `<version>`, for new coordinates that are already managed, or for any path where `newVersion` is set; `overrideManagedVersion` is untouched. `:rewrite-maven:test` is green (1438 tests).

### Which of the reported markers this removes

- Removes the `No version provided for direct dependency ...` markers whose cause is a version-less `ChangeDependency`: the `finos` set (#8462) — `jakarta.servlet:jakarta.servlet-api`, `jakarta.ws.rs:jakarta.ws.rs-api`, `jakarta.validation:jakarta.validation-api`, `jakarta.inject:jakarta.inject-api`, `io.swagger.core.v3:swagger-annotations`, `tools.jackson.core:jackson-{core,databind}`.

- Does **not** remove the `com.squareup.okhttp3:mockwebserver3` markers on `spring-cloud-sleuth` / `spring-cloud-sleuth-otel`. Those come from `UpgradeOkHttpMockWebServer`, which does pass `newVersion: 5.x`, and they have a different cause: I replayed the real POM chain (root → `tests` → `tests/brave` → module) and the recipe's own output is correct — the root's `<dependencyManagement>` becomes `mockwebserver3` and `<mockwebserver.version>` is bumped to 5.x, so the POMs on disk resolve. The marker appears because the child module's `MavenResolutionResult` still holds the parent's parse-time model, so the *next* recipe in the run that calls `maybeUpdateModel()` on that module re-resolves against a parent that still manages `mockwebserver`. Filed as #8463, together with the second `finos/legend-shared` symptom (a duplicate unversioned `jakarta.ws.rs-api` added next to the renamed original), which is that same stale model being read by `AddDependency`.

Also filed while here, not addressed:

- * #8464 — renaming coordinates without a `newVersion` carries the old version over to the new artifact, e.g. `jakarta.servlet:jakarta.servlet-api:4.0.1`.
- * #8465 — the Gradle path shares the defect shape but currently asserts it (`ChangeDependencyTest#doNotPinWhenNotVersioned`), and the `GradleProject` model does not record dependency-management-provided versions, so the same decision cannot be made there yet.

I could not reproduce the deletion of `<javax.jax-rs.version>` from `<properties>` while a `<dependencyManagement>` entry still referenced it, so it is not covered by any of the above.


timtebeek:
Don't quite have the time to follow through on this one; we'll stick with the open issue for now.

PR #8472, opened 2026-08-12
Emit J.Empty for a Groovy loop body that is an empty statement

### LINKED ISSUE #8459: Groovy parser leaves J.ForLoop.body null for a for loop with an empty statement body
## Summary

`GroovyParserVisitor` maps a classic `for` loop whose body is an empty statement (`for (...);`) to a `J.ForLoop` whose `body` right-padding holds a `null` element. `J.ForLoop#getBody()` dereferences that field unconditionally, so any visitor that touches the body throws `NullPointerException`. The Java parser emits `J.Empty` for this shape, which is why it has never surfaced there.

## Evidence

Seen in a Moderne SaaS flagship run of *Common static analysis issues* (2026-08-11, dev, org "Netflix + Spring", changeset `20260811034304-hV7qa`) as an ERROR marker on `github.com/Netflix/metacat`, `metacat-thrift/src/test/groovy/com/netflix/metacat/thrift/AbstractThriftServerTest.groovy`:

```
NullPointerException: Cannot invoke "org.openrewrite.java.tree.JRightPadded.getElement()" because "this.body" is null
  org.openrewrite.java.tree.J$ForLoop.getBody(J.java:2431)
  org.openrewrite.staticanalysis.NeedBraces$NeedBracesVisitor.visitForLoop(NeedBraces.java:256)
```

The trigger is line 43 of that file:

```groovy
for (; port < 1024; port = rand.nextInt(65536));
```

## Reproducer

```java
String src =
  "class A {\n" +
  "    void m(java.util.Random rand) {\n" +
  "        int port = 0\n" +
  "        for (; port < 1024; port = rand.nextInt(65536));\n" +
  "    }\n" +
  "}\n";
SourceFile cu = GroovyParser.builder().build()
  .parse(new InMemoryExecutionContext(Throwable::printStackTrace), src)
  .findFirst().orElseThrow();
new NeedBraces().getVisitor().visit(cu, new InMemoryExecutionContext());
```

throws the NPE above, verbatim.

## Root cause

`GroovyParserVisitor.visitForLoop` (around line 2400) builds the body as:

```java
JRightPadded.build(doVisit(forLoop.getLoopBlock()))
```

For an `EmptyStatement` loop block `doVisit` returns `null`, giving a `JRightPadded` whose element is null. `JavaVisitor#visitRightPadded` returns `null` when the visited element is null, so the first `super.visitForLoop(...)` collapses the whole right-padding to null — after which `getBody()` NPEs rather than returning null.

Independent confirmation that the parse is wrong and not merely unusual: `rewrite-test` rejects it outright with *"Source file was parsed into an LST that contains non-whitespace characters in its whitespace. This is indicative of a bug in the parser."*

## Suggested fix

Emit a `J.Empty` body for an `EmptyStatement` loop block, matching what the Java parser produces, and make sure the trailing `;` is consumed so the whitespace validation passes. `visitWhileLoop` and the other loop forms are worth checking for the same shape.

timtebeek:
- Fixes #8459

`GroovyParserVisitor` built loop bodies as `JRightPadded.build(doVisit(forLoop.getLoopBlock()))`. When the body is an empty statement (`for (...);`), `doVisit` returns `null`, so the padding held a `null` element and `J.ForLoop#getBody()` threw a `NullPointerException` for any visitor that touched it:

```
NullPointerException: Cannot invoke "org.openrewrite.java.tree.JRightPadded.getElement()" because "this.body" is null
  org.openrewrite.java.tree.J$ForLoop.getBody(J.java:2431)
  org.openrewrite.staticanalysis.NeedBraces$NeedBracesVisitor.visitForLoop(NeedBraces.java:256)
```

The trailing `;` was also never consumed, so the parse failed `rewrite-test`'s "non-whitespace characters in its whitespace" check.

Loop bodies now go through a `visitLoopBody` helper that emits a `J.Empty` for an `EmptyStatement` — matching what the Java parser produces — carrying a `Semicolon` marker so the `;` round-trips. Since Groovy prints the semicolon from that marker rather than from `printStatementTerminator`, putting it on the `J.Empty` itself (not on the padding) also means a recipe that replaces the body with a block doesn't leave a stray `;` behind.

Applied to all four loop forms, all of which had the same shape: classic `for`, `for`-each, `while` and `do`/`while`.

### DIFF
--- a/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
+++ b/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParserVisitor.java
@@ -2374,9 +2374,13 @@ public class GroovyParserVisitor {
             return labels.get(0).withStatement(condenseLabels(labels.subList(1, labels.size()), s));
         }
 
-        private Statement visitLoopBody(org.codehaus.groovy.ast.stmt.Statement loopBlock) {
-            if (!(loopBlock instanceof EmptyStatement)) {
-                return doVisit(loopBlock);
+        /**
+         * Visits the body of a loop or {@code if}, mapping a body that is just {@code ;} to a {@link J.Empty} rather
+         * than the {@code null} that visiting an {@link EmptyStatement} would otherwise produce.
+         */
+        private Statement visitBody(org.codehaus.groovy.ast.stmt.Statement body) {
+            if (!(body instanceof EmptyStatement)) {
+                return doVisit(body);
             }
             Space prefix = whitespace();
             Markers markers = Markers.EMPTY;
@@ -2410,7 +2414,7 @@ public class GroovyParserVisitor {
                     return new J.ForLoop(randomId(), prefix, Markers.EMPTY,
                             new J.ForLoop.Control(randomId(), controlFmt,
                                     Markers.EMPTY, init, condition, update),
-                            JRightPadded.build(visitLoopBody(forLoop.getLoopBlock())));
+                            JRightPadded.build(visitBody(forLoop.getLoopBlock())));
                 } else {
                     Parameter param = forLoop.getVariable();
                     Space paramFmt = whitespace();
@@ -2440,7 +2444,7 @@ public class GroovyParserVisitor {
 
                     return new J.ForEachLoop(randomId(), prefix, forEachMarkers,
                             new J.ForEachLoop.Control(randomId(), controlFmt, Markers.EMPTY, variable, iterable),
-                            JRightPadded.build(visitLoopBody(forLoop.getLoopBlock())));
+                            JRightPadded.build(visitBody(forLoop.getLoopBlock())));
                 }
             }));
         }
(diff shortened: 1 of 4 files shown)


greg-at-moderne (approved):

timtebeek:
Verified downstream against `NeedBraces` (the recipe from the original marker): with `rewrite-groovy` 8.89.0-SNAPSHOT from mavenLocal, `for (; port < 1024; port = rand.nextInt(65536));` now transforms cleanly to `for (...) {\n}` — no NPE, and no stray semicolon left behind, since the `Semicolon` marker lives on the `J.Empty` that the recipe replaces.

Also extended to `visitIfElse`, which was on the same `doVisit` path: `if (true);` produced a null then-part and dropped the `;`.

Confirmed the resulting shape matches the Java parser exactly — `javac` emits `J.Empty` for all five of `for`, `for`-each, `while`, `do`/`while` and `if`, and Groovy now does too.

timtebeek:
Superseded by
- https://github.com/openrewrite/rewrite/pull/8473

PR #8476, opened 2026-08-12
Add `deleteEmptyParents` option to JSON `DeleteKey`

### LINKED ISSUE #1175: DeleteKey removed workflow_dispatch in `.github/workflows/ci.yml`
Configured `DeleteKey` with `$.jobs.build.strategy.fail-fast`, and it appears to have removed `workflow_dispatch`. But it did remove `fail-fast` as expected.

Not having `workflow_dispatch` removed. Also note how it shifted `{}` next to `main`, which is interesting.

### Example diff
```diff
       - "*"
   pull_request:
     branches:
-      - main
-  workflow_dispatch: {}
+      - main {}
   schedule:
     - cron: "0 18 * * *"
 
 jobs:
   build:
     strategy:
-      fail-fast: false
       matrix:
         java: ["11"]
         os: ["ubuntu-latest"]
```

#### Recipes in example diff:
- org.openrewrite.yaml.DeleteKey

### References:
- Recipe ID: `yG2Zw`
- Recipe Name: "Delete key"
- Repository: openrewrite/rewrite-jhipster
- Created at Thu Nov 04 2021 01:34:46 GMT-0500 (Central Daylight Time)

Jenson3210:
## What's changed?

`org.openrewrite.json.DeleteKey` gains an optional `deleteEmptyParents` option (default `false`, so existing behavior is unchanged). When enabled, objects and arrays that become empty as a result of deleting the key are deleted too, applied recursively so that a whole chain of containers which only held the deleted key is removed:

```json
// DeleteKey("$.engines.node", true)
{                              {
  "name": "my-package",   ->     "name": "my-package"
  "engines": {                 }
    "node": ">=20"
  }
}
```

Recursion falls out of the visit order — children are visited before their parent, so an emptied object is removed from its parent, which may then empty in turn. The visitor tracks the ids of the containers *it* emptied, so objects and arrays that were already empty in the source are left alone.

Two extras came out of this:

- `Json.Array` values are handled as well, so an array element emptied by the deletion is removed, and an array left empty by that is removed from its own parent.
- The whitespace preceding a closing `}`/`]` is preserved when the deleted key was the *last* member. That space is stored as the trailing (`after`) space of the last member's right padding, and it was being deleted along with the member, leaving `{\n  "name": "my-package"}`. `JsonRightPadded.withElements` attempts to carry it over, but its `counter == before.size()` guard counts into the new (shorter) list while comparing against the old size, so it never fires on a removal. The fix here is local to `DeleteKey` rather than in `JsonRightPadded`, to avoid changing behavior for every JSON recipe. This applies regardless of the new option, so plain `DeleteKey` no longer mangles that formatting either.

- Naming of the option (`deleteEmptyParents`) — happy to rename.
- Whether the trailing-space fix belongs in `JsonRightPadded.withElements` instead, as a general fix for all JSON recipes that remove list elements.
- Removing an emptied element from an array changes that array's length; that seemed like the intent of "clean up what's now empty", but it's the one judgement call here.

## Checklist

- [x] I've added unit tests to cover both the happy and unhappy paths
- [x] `./gradlew :rewrite-json:test` passes

- Fixes #1175 — its previously `@Disabled` test `DeleteKeyTest#deleteNestedKeyRemovingUnusedKeysRecursively` is now enabled and passing.


Jenson3210:
Updated: kept the single-argument constructor as `@Deprecated` delegating to a new `@JsonCreator` two-argument constructor, matching the convention used by e.g. `yaml.CopyValue` and `maven.ChangeParentPom`, so downstream modules binding against `new DeleteKey(keyPath)` keep compiling. Added tests for the deprecated constructor and for the declarative (Jackson) path with and without the option.

jkschneider:
Thanks for taking this on, and sorry to do this after the merge — I've reverted it in #8494.

The behavior is something we want, but as its own recipe rather than an option on `DeleteKey`. `DeleteKey` should stay focused on deleting the key it was pointed at; cleaning up containers that are left empty isn't specific to it. Any recipe that removes JSON members can leave an empty object or array behind, and a `RemoveEmptyKeys` recipe composes with all of them — including on JSON that already had empty containers in it before any recipe ran. Where you want both, they can be combined in a declarative recipe.

So the preferred approach here would be a new `RemoveEmptyKeys` recipe in `rewrite-json`. Would you be up for a follow-up PR along those lines?

Two things from this PR worth carrying forward:

- The trailing-whitespace fix (preserving the space before a closing `}`/`]` when the last member is removed) is a real bug and independent of all of this. To your own question in the description — yes, that belongs in `JsonRightPadded.withElements`, where the `counter == before.size()` guard is wrong on removal, so every JSON recipe benefits. Happy to take that as a standalone PR.
- #1175 stays open; `RemoveEmptyKeys` plus `DeleteKey` should close it, and the `@Disabled` test there is a good starting point.

PR #8487, opened 2026-08-13
Size static wildcard imports by their static members in `Autodetect`

timtebeek:
`Autodetect` sized a static wildcard import by counting variables whose *type* was the imported type, whereas `RemoveUnusedImports` counts members whose *owner* is that type. A file that also declares a variable of the imported enum type — such as a `Measure measure` test parameter — inflated the detected `nameCountToUseStarImport` past the number of names the wildcard actually folds, so `RemoveUnusedImports` unfolded a wildcard that `OrderImports` would fold straight back.

Because `minimumFoldedStaticImports` is a minimum across the files that currently hold a wildcard, unfolding one file raised the threshold for the next, so a repository like [moderneinc/rewrite-devcenter](https://github.com/moderneinc/rewrite-devcenter) produced a fresh no-op import change on every nightly run — verified by iterating `mod build` / `mod run` over a clone.

The count is now the set of distinct static members of the imported type that the file actually uses: variables matched by `getOwner()` and filtered on `Flag.Static` (instance fields would inflate the count the same way), plus statically imported methods, which were previously ignored entirely and pinned the threshold to 3 for any file using `import static ...Assertions.*`.

Not addressed here: `minimumFoldedImports` / `minimumFoldedStaticImports` stay at `Integer.MAX_VALUE` when a repository has no wildcard imports at all, so `Math.max(..., 5)` / `Math.max(..., 3)` act as an infinite ceiling rather than a floor; changing that would start introducing wildcard imports into repositories that have none.

### DIFF
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/style/AutodetectTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/style/AutodetectTest.java
@@ -664,6 +664,43 @@ class AutodetectTest implements RewriteTest {
     }
 
     @Test
+    void staticStarImportCountsFoldedMembersNotVariablesOfThatType() {
+        var cus = jp().parse(
+          """
+            package org.openrewrite.test;
+
+            public class Outer {
+                public enum Measure {
+                    One, Two, Three, Four, Five, Six
+                }
+            }
+            """,
+          """
+            package org.openrewrite.test;
+
+            import java.util.List;
+
+            import static org.openrewrite.test.Outer.Measure.*;
+
+            class OuterTest {
+                List<Object> used() {
+                    return List.of(One, Two, Three, Four, Five);
+                }
+
+                void parameterized(Outer.Measure measure) {
+                }
+            }
+            """
+        );
+
+        var detector = Autodetect.detector();
+        cus.forEach(detector::sample);
+        var importLayout = detector.build().getStyle(ImportLayoutStyle.class);
+
+        assertThat(importLayout.getNameCountToUseStarImport()).isEqualTo(5);
+    }
+
+    @Test
     void detectMethodArgs() {
         var cus = jp().parse(
           """
--- a/rewrite-java-test/src/test/java/org/openrewrite/java/style/AutodetectTest.java
+++ b/rewrite-java-test/src/test/java/org/openrewrite/java/style/AutodetectTest.java
@@ -671,7 +671,9 @@ class AutodetectTest implements RewriteTest {
 
             public class Outer {
                 public enum Measure {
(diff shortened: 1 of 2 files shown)


timtebeek:
### Expected downstream impact: minimal

`Autodetect` is not reachable from `rewrite-test`, so **no recipe test expectations change**. `rewrite-test/src/main/java` has zero references to it, `RecipeSpec` has no style-injection path, and none of `rewrite-migrate-java`, `rewrite-spring`, `rewrite-testing-frameworks`, `rewrite-apache`, `rewrite-jackson` or `rewrite-static-analysis` reference `Autodetect` at all. It is invoked only by consumers that parse real projects — `rewrite-maven-plugin` (`AbstractRewriteBaseRunMojo`), `rewrite-gradle-plugin` (`DefaultProjectParser`) and the Moderne CLI.

For actual recipe runs, `nameCountToUseStarImport = max(min-over-wildcard-files, 3)`, so the floor of 3 absorbs the change unless *every* static wildcard in a repository folds more than three members. Measuring the same detector before and after:

| repo shape | before | after |
| --- | --- | --- |
| enum wildcard + a variable of that type | 6 | 5 |
| method-holder wildcard, 1 member used | 3 | 3 |
| method-holder wildcard, 5 members used | 3 | 5 |
| mixed, at least one file folding <= 3 members | 3 | 3 |

Minimum distinct static members actually used, per repository:

| repository | static wildcards | min | before | after |
| --- | --- | --- | --- | --- |
| openrewrite/rewrite | 128 | 1 | 3 | 3 |
| rewrite-migrate-java | 24 | <= 3 | 3 | 3 |
| rewrite-spring | 15-16 | <= 3 | 3 | 3 |
| rewrite-testing-frameworks | 13 | <= 3 | 3 | 3 |

All are dominated by `org.openrewrite.java.Assertions.*`, `java.util.Collections.*` and `java.util.stream.Collectors.*`, which the old code scored as 0 (no variable has type `Assertions`) and the new code scores as 1-3. Both floor to 3, so `OrderImports` and `RemoveUnusedImports` should produce identical output on all of them. `classCountToUseStarImport` is untouched.

The one shape that does change is a repository whose static wildcards *all* fold more than three members: the threshold rises, producing a one-time diff (less folding, some existing wildcards unfolded) and then stabilising. That is the intended correction — it is what stops [moderneinc/rewrite-devcenter](https://github.com/moderneinc/rewrite-devcenter) emitting a fresh no-op import change every night.

`rewrite-kotlin`'s `Autodetect` carries a verbatim copy of the same count feeding `javaStaticsAndEnumsToUseStarImport`; tracking that separately.

MBoegers (approved):

PR #8499, opened 2026-08-13
Give marketplace categories real display names

jkschneider:
Categories with no descriptor fall back to a capitalized package segment, so the
marketplace was rendering `Gradle9`, `Gradle8` and similar next to their recipes.

Adds descriptors across `rewrite-gradle`, `rewrite-groovy`, `rewrite-hcl`,
`rewrite-javascript` and `rewrite-maven`, and regenerates the `recipes.csv` for
the two modules whose output changed.

One thing worth a maintainer's eye: after `recipeCsvGenerate`, the nested
categories pick up their descriptions (`Gradle 8`, `Gradle 9`, `Gradle Plugins`,
`Maven Plugins`, `Maven Utilities` all resolve), but the module-root ones —
`org.openrewrite.gradle`, `.maven`, `.groovy`, `.hcl`, `.javascript` — do not,
and still render as a bare capitalized segment. `rewrite-docker` and
`rewrite-kotlin` bind at that same level from a structurally identical
descriptor, and renaming the file to `category.yml` does not change the outcome,
so this looks like something in how the root category is resolved rather than a
mistake in these entries. They are left in as correct metadata.

`rewrite-javascript`'s CSV is unchanged here — `npmInstall` could not run in this
environment, so that module's CSV was not regenerated.

https://claude.ai/code/session_019r1a149dSe3VMXxtbPEPXs


jkschneider:
Correcting the caveat in the description above: **the module-root categories do bind.** `org.openrewrite.gradle`, `.maven`, `.groovy`, `.hcl` and `.javascript` all resolve correctly, and there is nothing wrong with how the root category is looked up.

The observation was an artifact of how I tested it. `recipeCsvGenerate` *merges into* an existing `recipes.csv` rather than replacing it, so regenerating on top of a CSV produced before the descriptors existed leaves the old undescribed rows in place. Every check I ran — including renaming `groovy-categories.yml` to `category.yml` to see if the file name mattered — was done against an already-present CSV, so all of them reported the stale result.

Deleting `recipes.csv` first and regenerating gives the expected output:

```
Gradle             Recipes to search and transform [Gradle](https://gradle.org/) ...
Gradle 8           Recipes to migrate to [Gradle 8](https://docs.gradle.org/8.0/...
Gradle 9           Recipes to migrate to [Gradle 9](https://docs.gradle.org/9.0.0/...
Gradle Plugins     Recipes to add, remove, and upgrade [Gradle plugins](https://...
```

The consequence is that the CSVs merged in this PR are themselves incomplete — the 84 `rewrite-gradle` rows and the `rewrite-maven` rows carry the category *names* but not the root category's description. A follow-up PR regenerates them from scratch. Apologies for the noise in the original description.

Issue #4852, opened 2025-01-06
Support updating Gradle programmatic Version Catalogs


shanman190:
## What problem are you trying to solve?
Enable updating of version catalogs defined within `settings.gradle` scripts.

## Describe the solution you'd like
I think either of the following would be a reasonable solution:
* `UpgradeDependencyVersion` and `UpgradePluginVersion` also handle programmatic version catalog updates
* New recipes to perform these updates

## Have you considered any alternatives or workarounds?
None exist right now without a user crafting a custom recipe.

## Additional context
Originally asked for [here](https://github.com/openrewrite/rewrite/issues/4400#issuecomment-2421723868).

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
Always. :)

knutwannheden:
Without having given it much thought I would have gravitated towards extending the existing recipe. I also wonder if we should extend the GradleProject marker with some kind of flag that the dependencies are managed externally in TOML.

shanman190:
Yeah, I gravitate toward updating the existing recipes as well rather than creating new ones.

As far as the flag for external dependencies in TOML, I would think that's more closely aligned to #4400 rather than this issue in particular since here the versions are declared directly within the `settings.gradle` file as defined here:
https://docs.gradle.org/current/userguide/version_catalogs.html#sec:programmatic-catalog-versions

climategadgets:
Are you gentlemen working on this issue? Gradle version catalog is a de-facto standard now, complex projects will not exist without it, and the fact that OpenRewrite doesn't update the version catalog makes it a non-starter for this purpose, regrettably. And the blast radius is huge - any recipe that changes dependencies is affected.

timtebeek:
Took a look — confirming this is still relevant (no PR currently references it). Sketch of what it would take to land, building on the consensus above to extend the existing recipes rather than add new ones:

**1. New trait** in `rewrite-gradle/src/main/java/org/openrewrite/gradle/trait/` to recognize the `dependencyResolutionManagement { versionCatalogs { create("libs") { library/version/plugin(...) } } }` DSL nesting. This primitive doesn't exist yet and would be reused by both recipes below.

[...]

PR #8503, opened 2026-08-14
Support programmed Gradle Version Catalogs version upgrades

BoykoAlex:
- Partial fix for #4852. (plugin version upgrade is not included)

## What's changed?
Adds Gradle Version Catalog support (`dependencyResolutionManagement { versionCatalogs { ... } } }`, both Groovy and Kotlin DSL) to `UpgradeDependencyVersion`, so it can bump versions declared in a catalog instead of only versions declared directly in `build.gradle`/`build.gradle.kts` dependency blocks.

Scope for this PR is intentionally limited to `library(...)` and `version(...)` declarations. Catalog `plugin(...)` and `bundle(...)` entries, and `from(...)` (TOML-file-backed catalogs), are not supported yet.

This is implemented via three new traits:
- `GradleVersionCatalog` — represents a single named catalog (`libs { ... }` / `create("libs") { ... }`), exposing its libraries and version declarations, and applying version changes to them.
- `VersionCatalogLibrary` — a single `library(...)` declaration, with its inline version, `versionRef(...)`, or `withoutVersion()` form.
- `VersionCatalogVersion` — a single `version(alias, value)` declaration.

Several libraries commonly share one `version(...)` declaration via `versionRef(...)` (e.g. a dozen `spring-boot-starter-*` artifacts all pointing at one `springBoot` version). `UpgradeDependencyVersion` now handles this correctly:
- If every library sharing a reference is targeted together, the shared `version(...)` declaration itself is bumped once.
- If only some are targeted, the targeted library is detached to its own inline literal, leaving the shared reference and untargeted siblings alone.
- If, across separate recipe runs, the detached members happen to converge back on the same version, they're re-attached to the shared reference.
- A library whose version can't be resolved to a literal at all -- `withoutVersion()`, a rich `.version { strictly(...) } }` constraint, or an interpolated `"${...}"` string -- is treated consistently: left unchanged rather than crashing or being forced to a literal, and it doesn't interfere with reconciliation of any group it isn't genuinely part of.

A new `GradleVersionCatalogVersionReferences` marker snapshots the original versionRef sharing structure the first time a catalog is touched, so the detach/re-attach algorithm above can tell whether two separately-requested bumps actually target the same underlying declaration.

Also adds a few small reusable helpers to `GradleTraitMatcher` (`isTopLevelStatement`, `asChainedInvocation`, `literalArgument`) used by the new traits.

## What's your motivation?
Gradle Version Catalogs are the recommended way to declare dependency versions in modern Gradle projects, but `UpgradeDependencyVersion` previously had no awareness of them, so catalog-based projects couldn't be upgraded automatically.

- The versionRef detach/re-attach algorithm in `GradleVersionCatalog` (`withVersion`, `withOriginalVersionReferencesMarker`, `reconciledAfterDetaching`) — this is the trickiest part of the change and is deliberately kept separate from `UpgradeDependencyVersion`'s existing `safeUpdatedVersion` logic for Gradle variables, since the two cases have different sharing semantics.
- Whether the memoization in `GradleVersionCatalog` (AST walk cached per-instance, discarded on any mutation) is an acceptable tradeoff.

Considered forcing `withoutVersion()` libraries onto the target version whenever one is specified, but decided against special-casing it: it's just one of several shapes (alongside rich `.version { ... }` constraints and interpolated versions) where a library's version can't be resolved to a literal, and all of them are now handled the same way -- left unchanged rather than crashing.

### Checklist
- [x] I've run `./gradlew build` locally


BoykoAlex:
What are the chances of getting this in for the next or next after next 8.90.x patch release? Or were you thinking of 8.91.0?

kdvolder:
We are somewhat  eager to get this merged sooner rather than later if possible.

Note: Alex and I worked on this together, but as Alex is on PTO right now (staritng today)... I will be keeping an eye on this and follow up as needed.

Alex mentioned to me he thinks @timtebeek who we usually work with is probably on PTO at the moment. So I am hoping someone else (maybe @sambsnyd , @knutwannheden or @shanman190) might have time and expertise to review this Gradle-related PR and help us get it to the finish line?

Your time and effort are much appreciated.

shanman190:
This is mostly thinking aloud here, but what I kind of would imagine is the VersionCatalog trait understands how to find and update the information in settings.gradle(.kts) -- foundation of what's here in the PR today. In the scan phase of a recipe that's interested in doing something with the VersionCatalog, we would grab the catalog and add it to the accumulator object(s) to then enable ifs usage in later recipe phases.

So in particular, it kind of feels like the `GradleVersionCatalogVersionReferences` feels unnecessary and not really what would be conveyed in a marker. Then the additional traits _might_ be a bit of an early optimization in practice at the moment?

kdvolder:
>  In the scan phase of a recipe that's interested in doing something with the VersionCatalog, we would grab the catalog and add it to the accumulator object(s) to then enable ifs usage in later recipe phases.

Not sure I understand the suggestion, but the main source of complexity is that the recipe is trying respect the original 'sharing' structure (i.e. when several entries in the catalog reference the same symbolic version). 

For example, suppose we have two artifacts called 'bar' and 'foo' which reference 'barFooVersion' which is bound to version 1.0.0.

Ignoring syntax here but conceptual structure:
```
bar -> barfooVersion
foo -> barfooVersion

barfooVersion = 1.0
```

Supposing we have **two** recipes applying a version upgrade, one changes 'bar' to version 2.0.0
Second recipe changes 'foo' to version '2.0.0'. 

The end  result we want is that the 'sharing structure' is preserved when the new version assignment is 'compatible' with the original sharing structure (i.e. if both bar and foo **eventually** land on the same version , then we can continue to use the symbolic 'barfooVersion' to set that version.

```
bar -> barfooVersion
foo -> barfooVersion

barfooVersion = 2.0   // <- change this instead the version of 'bar' and 'foo' individually
```

To implement this we need to 'remember' somehow what the initial 'sharing structure' was so we can always strive to match it 'as good as possible' given the target version assignment. 

We want this even when there are two (or more) separate recipes making changes individually.

So in our current example, after the first recipe gets applied, but before the second... the assignment is temporarily not consistent with the original sharing and so the 'symbolic link' gets temporarly broken. After first recipe changes 'bar' to version 2.0.0 we get this:


```
bar -> 2.0.0 // changed this, inline because not compatible with other current variable assignements
foo -> barfooVersion

barfooVersion = 1.0   // <- unchanged, still used in assigning foo version
```
Wen the second recipe executes and also changes 'foo' to version 2.0... the recipe can then use the marker to determine that now assignment to both bar and foo have become 'compatible' with the original sharing structure again and we can 'snap back' to using the symbolic barfooVersion.

We thought that using a marker to serve as this kind of 'memory of the original sharing structure' was a good way to do this. But really any mechanism that lets you 'make a memory note' about what the original sharing sturcture looked like that can be carried through multiple recipes that execute one after another would work. So if you think using a scanning recipe as the mechanic to 'remember what was there at the start' is better we can probably do that. However... would that work if there is more than one recipe that needs access to the memory state? 

> Then the additional traits might be a bit of an early optimization in practice at the moment?

No idea, but using traits was kind of the plan suggested by Tim and it seemed like a good idea to us (even if only to abstract away the different handling between kotlin and groovy syntax, it probably seems like a good way to do that). And the management of the 'initial sharing state' marker felt like a natural addition.

kdvolder:
Thinking about it some more. I think you are suggesting that instead of using a marker for the 'initial sharing structure memory' we instead 'grab' the initial sharing structure in the scanning phase of scanning recipe.

Honestly, I though this might be a problem because a scanning accumulator state specifically belongs to a single recipe instance. So it won't allow to share state between two instances of recipe (i.e. two upgraded ependency recipes changing dependencies independently, like the 'bar' and 'foo' version updates in my example above.

This is the reason why we 'rejected' that option when we were hashing out our implementation/design. 

However... I realize now that isn't *really* a problem because both the 'bar' and 'foo' version update recipes can independently grab and store the initial state, so this state doesn't really need to be shared. This is fine as long as they both agree, which I suppose would be the case since the scanning phase happens before any recipe did anything at all, so this means both recipes would both be 'taking a picture' of the same initial state independently and keeping that picture in their own accumulator. The state would not technically be 'shared' but it would still be exactly the same.

kdvolder:
I am working on some cleanups, I'd put this into 'draft' but seems like I don't have 'permission' so this comment will have to do. Don't review until I am done cleaning. I'll post something when ready.

kdvolder:
I did a thorough re-review and restructured the code a bit. I was focusing mostly on trying to clarify the high-level structure. The idea/intention really is that the `GradleVersionCatalog` trait is a complex bit of magic but with a very simple and clean "user facing api". Essentially it gives you two simple methods to read and update versions in a catalog, and that is all there is to it from the user's POV.

Internally it encapsulates an algorithm that tries to match the original sharing structure (how several artifacts all use the same symbolic version name). And it does this regardless of whether these changes happen all together in a single recipe invocation or in separate invocations. The marker makes this possible. But I consider the marker to be just an "implementation detail". The algorithm needs some way to remember what the original sharing structure looks like, so that it can always try to represent the catalog using those symbolic references as much as possible.

What I have done for the latest round of cleanups is to restructure the code and put as much as possible of the "complex machinery" inside `GradleVersionCatalog` as nested private classes and methods. I have also tried to clean up some of the javadoc comments. They are now more focused on explaining the simple "user facing api" and its overall intent.

@shanman190 raised some concerns, which I think come down to there maybe being "too many traits". While I didn't get rid of them, these "helper traits" are now encapsulated inside `GradleVersionCatalog`, which is the only real "user-facing API".

I experimented a bit with collapsing the "implementation helper traits" into something more like helper methods. But to me it felt like this didn't really reduce the complexity very much, while it made the code less readable. (Traits are kind of a nice mechanism to attach a readable api to an AST node, and I think it helps rather than hinders code quality in that sense, even if it might be seen as "overkill" for something as simple as recognizing and parsing a `version(...)` call and extracting its arguments.)

Anyhow, as those things are now strictly encapsulated as "implementation private apis", it shouldn't be too hard to flatten them out into a collection of helper methods if that's what you want.

Note: this kind of "encapsulated magic" was always kind of the idea, but I think because of having all those "little helper traits" scattered around as separate classes, and not clearly labeled as "implementation details", this intention kind of got lost in the weeds. Hopefully it is now much better.

Let me know what you think. Hope we can merge this soon-ish, but do let us know if you want more changes or have more questions.

Kris and Alex

jkschneider:
Thank you for this work. Your commit is carried into #8864 with authorship intact, where the settings DSL and TOML forms of a version catalog are generalized into a single `VersionCatalog` trait. Closing in favor of that PR.

PR #8505, opened 2026-08-14
Send recipe identity, not live recipe graphs, in `RecipesThatMadeChanges`

knutwannheden:
## The bug

`RecipesThatMadeChanges` holds `Collection<List<Recipe>>` — live `Recipe` instances — and had no RPC codec, so `RpcSendQueue` put the raw object into `RpcObjectData.value` and handed the whole graph to Jackson. The marker rides on every file a recipe edited, so this went out on every `GetObject` for a changed file.

Abridged from a reference chain recovered under the Moderne CLI:

```
RecipesThatMadeChanges[recipes] → RecipePath[0]
 → RpcRecipe[editPreconditionVisitor] → RewriteRpc$6[cursor] → Cursor[messages]
 → UpgradePythonVersion$Accumulator[projects] → ProjectState[manifest]
 → Toml$Document[values] → … → SourceSetStore$3[in] → BinaryInput[segment]
```

Printing one Python file was shipping the recipe stack, a precondition visitor, that visitor's `Cursor`, the cursor's message map, a scanning recipe's accumulator, and through it an unrelated file's LST. It bottomed out in `jdk.internal.foreign.NativeMemorySegmentImpl`, which Jackson cannot serialize, so every run that changed Python files spent ~90s in three `Print` calls that each died on a 30s timeout — and still reported SUCCEEDED.

## The fix

A stack travels as identity: one `RecipeThatMadeChanges` per frame, carrying name, display name, rendered instance name, configured option values and estimated effort. It extends `Recipe`, so a received marker still satisfies the marker's `Collection<List<Recipe>>` API, and `Result.timeSavings`, `getRecipeDescriptorsThatMadeChanges()` and rendered diffs keep working untouched. `of()` passes an existing frame through, so forwarding a received marker doesn't degrade it again on the next hop.

`RecipesThatMadeChanges` implements `RpcCodec`. Java, C#, Go, TypeScript and Python each model the frame and carry it without interpreting it, so no peer needs a `Recipe` concept. Java's `rpcSend` is the canonical field order; each peer codec points back to it.

Two things worth knowing:

- The frame carries `@AbstractRecipe`. It subclasses `Recipe` only to satisfy the marker's existing API, and without the annotation classpath scanning instantiates one with no fields set, putting a `null`-named descriptor into every catalog built from that classpath.
- Option *types* don't travel, only values. A received frame's `RecipeDescriptor` therefore never equals a live recipe's, so a recipe that also ran locally can be listed twice.

## Measured

The property that matters is that the marker costs what the recipe stack costs, not what the recipes happen to be holding. Serializing the same two-frame stack through the formatter `GetObject` actually uses, varying only the size of a scanning accumulator reachable from a precondition visitor's cursor:

| accumulator | before | after |
|---|---|---|
| 0 entries | 1,191 B | 845 B |
| 100 | 6,280 B | 845 B |
| 1,000 | 53,080 B | 845 B |
| 10,000 | 530,080 B | 845 B |

Flat, and that is per changed file. `payloadIsBoundedByStackDepthNotRunState` pins it. Note this is with an accumulator that serializes cleanly — the reported one reached an unrelated file's LST and bottomed out in something Jackson could not serialize at all, which is where the timeouts came from rather than the size.

For a plain stack of three local recipes with no run state to reach, the payload goes 1,537 B → 1,309 B. The win in the ordinary case is small; the point is the tail.

Separately, an earlier build of this branch — the same marker slimmed via a Jackson module rather than a codec — running `UpgradeToPython313` under the Moderne CLI on stock 8.89.1: errored `Print` calls went from 3 to 0, total `Print` time from 91,253 ms to 396 ms, and the slowest single `Print` from 30,009 ms to 33 ms, with `fix.patch` byte-identical and all 57 `recipesThatMadeChanges` entries in `changes.json` unchanged. Those numbers are from the Jackson-module build, not this codec.

## Tests

`RecipesThatMadeChangesTest` covers the payload shape and its bound, a round trip of every field, a configured `FindAndReplace` keeping its rendered instance name and typed option values, stack order, forwarding, adding a stack against a real `before`, and an untouched marker coming back as the same instance. One test asserts the frame is not enumerated as a recipe, and fails without `@AbstractRecipe`.

Each peer runtime gets a two-hop round trip — hop 1 exercises its receiver against a Java-shaped stream, hop 2 its own sender against what that receiver produced — both asserting the stream actually carries a recipe name, so neither can pass by degenerating to `NO_CHANGE`. Writing those found two real defects, both fixed here: Go registered codec factories but not `RegisterValueType`, so the sender stamped no `valueType` and the receiver produced `nil`; and `rewrite-python`'s `_get_primitive_value` returned `None` for a plain `dict`, silently dropping any mapping-valued field on send.

Green: `:rewrite-core:test`, `:rewrite-maven:test`, `licenseFormat`; `go test ./pkg/rpc/`; `vitest run` (1932); `pytest` (2070); `dotnet test` Core+Rpc (439, alongside 7 `CsprojParseTests`/`JavaTypeAnnotationRealRpcTest` failures that predate this branch and reproduce on a stashed baseline).


rewrite-core/src/main/java/org/openrewrite/rpc/RpcRecipe.java
```diff
@@ -35,6 +36,7 @@
 
 @RequiredArgsConstructor
 public class RpcRecipe extends ScanningRecipe<Integer> {
+    @JsonIgnore
```
jkschneider: this is weird. it's like all data on RpcRecipe shouldn't be serialized. it seems like in whatever context it is attempting to be serialized it just shouldn't instead.

rewrite-core/src/main/java/org/openrewrite/marker/RecipeIdentity.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private final String name;
+    private final @Nullable String displayName;
+    private final @Nullable String instanceName;
+    private final @Nullable Map<String, Object> options;
+    private final @Nullable Duration estimatedEffortPerOccurrence;
+
+    public RecipeIdentity(String name, @Nullable String displayName, @Nullable String instanceName,
+                          @Nullable Map<String, Object> options, @Nullable Duration estimatedEffortPerOccurrence) {
+        this.name = name;
```
jkschneider: Use `@RequiredArgsConstructor`

rewrite-core/src/main/java/org/openrewrite/marker/RecipeIdentity.java
```diff
@@ -0,0 +1,150 @@
...
+import static java.util.Collections.emptyList;
+
+/**
+ * One frame of a {@link RecipesThatMadeChanges} stack, reduced to what a consumer of the marker
+ * reads: how the recipe is named, how it was configured, and what it is worth. Standing in for the
+ * recipe itself keeps run state — cursors, scanning accumulators, and the source files they reach —
+ * off the wire. Reading anything beyond the fields below yields a {@link Recipe} default.
+ */
+@AbstractRecipe
+public class RecipeIdentity extends Recipe implements RpcCodec<RecipeIdentity> {
```
jkschneider: Should probably call it RecipeThatMadeChanges.

jkschneider (changes requested):

jkschneider (approved):

jkschneider (approved):

PR #8529, opened 2026-08-17
remove explicit `Unit` return type

barbulescu:
## What's changed?
Add Kotlin recipe to remove explicit `Unit` return type

## What's your motivation?
Learn how to implement recipes using Kotlin DSL.

I tried to use declarative approach, but did not work. Claude said that it could be caused by missing K2 compiler support but I had not time to check this. 

@timtebeek 

### Checklist
- [x] I've run `./gradlew :rewrite-kotlin:build` locally, and committed any resulting changes to `recipes.csv`
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch


timtebeek:
Hey thanks! My immediate thoughts went to this recipe which seems related, if perhaps opposite from what we see here:
- https://github.com/openrewrite/rewrite-testing-frameworks/blob/9665ede603fd0d9d56040945623f85727284c124/src/main/java/org/openrewrite/java/testing/cleanup/KotlinTestMethodsShouldReturnUnit.java#L35-L54
- https://github.com/openrewrite/rewrite-testing-frameworks/blob/9665ede603fd0d9d56040945623f85727284c124/src/test/java/org/openrewrite/java/testing/cleanup/KotlinTestMethodsShouldReturnUnitTest.java#L42-L78

Any thoughts to that? Is there overlap, or chance of conflict?

barbulescu:
It looked like a safe, easy recipe - I was so wrong.

At least I learned something today.

timtebeek:
Hah, no worries; Half the challenge in everything we do is remembering all the different edge cases where things might break. At times makes your head hurt, but at least we've learned something today. Thanks for helping out!

Issue #8270, opened 2026-07-16
Support Kotlin 2.4+ in rewrite-kotlin (recipe DSL compiler plugin pinned to 2.3.20)


timtebeek:
### What version of OpenRewrite are you using?

Latest (`rewrite-kotlin` currently built against Kotlin `2.3.20`).

### How are you running OpenRewrite?

As a recipe-authoring consumer via the [kotlin-recipe-starter](https://github.com/moderneinc/kotlin-recipe-starter), which puts `rewrite-kotlin` on `kotlinCompilerPluginClasspath` to use the Kotlin recipe DSL.

### Problem

`rewrite-kotlin` is pinned to the Kotlin `2.3.x` compiler, so consumers cannot move their own `kotlin("jvm")` Gradle plugin to Kotlin `2.4.x`. When they do, compilation of their recipe module fails during `:compileKotlin`:

```
e: java.lang.ClassCastException:
   org.jetbrains.kotlin.fir.extensions.FirExtensionRegistrarAdapter$Companion
   cannot be cast to org.jetbrains.kotlin.extensions.ProjectExtensionDescriptor
> Task :compileKotlin FAILED
   > Internal compiler error. See log for more details
```

- Reproduced by the Dependabot bump in moderneinc/kotlin-recipe-starter#2 (Kotlin `2.3.20` → `2.4.10`); CI: https://github.com/moderneinc/kotlin-recipe-starter/actions/runs/29453236383/job/87480488714

### Root cause

The Kotlin recipe DSL is a K2 compiler plugin shipped inside `rewrite-kotlin`
(`RecipeCompilerPluginRegistrar` → `FirExtensionRegistrarAdapter.registerExtension(...)`), compiled against `kotlin-compiler-embeddable:2.3.20` (`rewrite-kotlin/build.gradle.kts`, `val kotlinVersion = "2.3.20"`). The `@ExperimentalCompilerApi` FIR extension-registration internals are not binary-stable across Kotlin minor versions, so a plugin built for 2.3 cannot be loaded by the 2.4 compiler. This hard-couples a consumer's `kotlin("jvm")` version to the Kotlin version `rewrite-kotlin` was built against.

### Requested change

Support Kotlin `2.4+` in `rewrite-kotlin`. Expected work:

1. Bump `kotlinVersion` (`kotlin-compiler-embeddable`/`reflect`/`stdlib` + `kotlin-bom`) in `rewrite-kotlin/build.gradle.kts` to `2.4.x`.
[...]

PR #8542, opened 2026-08-18
migrate to Kotlin 2.4.10

barbulescu:
## What's changed?
- Add support for Kotlin 2.4.10 as described in https://github.com/openrewrite/rewrite/issues/8270
Tested locally with `kotlin-recipe-starter` and I could not see any issues.

## What's your motivation?
Add support for Kotlin 2.4.10

@timtebeek 

### Checklist
- [x] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv`
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch


timtebeek:
Thanks so much! Great to see this passing already; I was expecting this to be more of a pain, but I guess the `rewrite-kotlin/src/main/kotlin/org/openrewrite/kotlin/recipe/internal/Ir24ParameterCompat.kt` makes this easier with deferred changes. We'll need some validation of trying different Kotlin versions before we merge I reckon, but good to have this set in motion. 🙏🏻

timtebeek:
@sambsnyd do you expect any impact on the work you're doing? Since I know you've been looking at Kotlin/Android.

sambsnyd:
@timtebeek should be fine to upgrade... newer kotlin parsers can generally still parse old code so there isn't much backwards compatibility risk for minor versions

sambsnyd (approved):

timtebeek:
Thanks @barbulescu! I ran a regression sweep across Kotlin versions locally. CI is green and the parser work looks right; the thing I'd like us to decide on before merging is the consumer compatibility window.

## Downstream consumer matrix

Built this branch to mavenLocal (`8.91.0-SNAPSHOT`) and compiled the [kotlin-recipe-starter](https://github.com/openrewrite/kotlin-recipe-starter) sources against it, varying the consumer's `kotlin("jvm")` version. Baseline is the released `8.90.2` (compiler 2.3.20).

**Recipe-DSL consumers** (rewrite-kotlin on `kotlinCompilerPluginClasspath`):

| consumer `kotlin("jvm")` | 8.90.2 (before) | this PR (after) |
|---|---|---|
| 2.0.21 | ❌ | ❌ |
| 2.1.21 | ❌ | ❌ |
| 2.2.21 | ❌ | ❌ |
| 2.3.20 | ✅ | ❌ |
| 2.4.10 | ❌ | ✅ |

Every failure is the same error, in both directions:

```
e: java.lang.ClassCastException: class org.jetbrains.kotlin.fir.extensions.FirExtensionRegistrarAdapter$Companion
   cannot be cast to class org.jetbrains.kotlin.extensions.ExtensionPointDescriptor
	at org.openrewrite.kotlin.recipe.internal.RecipeCompilerPluginRegistrar.registerExtensions(RecipeCompilerPluginRegistrar.kt:56)
```

That is #8270 in mirror image: the DSL compiler plugin works with *exactly one* Kotlin version, and this PR moves that window from 2.3.20 to 2.4.10 rather than widening it. Everyone currently on 2.3.20 — including the starter template's own pin — breaks on upgrade.

**Library-only consumers** (rewrite-kotlin as an ordinary dependency, no compiler plugin) are governed by Kotlin's metadata read-ahead rule instead, so the floor moves from 2.1 to 2.3:

| consumer `kotlin("jvm")` | 8.90.2 | this PR |
|---|---|---|
| 2.0.21 | ❌ | ❌ |
| 2.1.21 | ✅ | ❌ |
| 2.2.21 | ✅ | ❌ |
| 2.3.20 | ✅ | ✅ |
| 2.4.10 | ✅ | ✅ |

(`Module was compiled with an incompatible version of Kotlin ... binary version` on the rewrite-kotlin jar.) Java-only consumers are unaffected.

None of this is a defect in the patch — it's inherent to the compiler-plugin coupling — but it does mean the bump is a breaking change for DSL authors, so it wants a release note and a matching starter bump at minimum.

## Parser language levels

Swept `KotlinParser.KotlinLanguageLevel` over the parser/type suites (`tree.*`, `KotlinTypeMappingTest`, `KotlinTypeGoatTest` — 639 tests):

| level | result |
|---|---|
| KOTLIN_2_0 / 2_1 / 2_2 | pass |
| KOTLIN_2_3 | 1 failure |
| KOTLIN_2_4 | 1 failure |

Both are `LambdaTest.suspendLambda`. At levels 2_3/2_4 the whole enclosing function

```kotlin
fun method ( ) {
    val lambda : suspend ( ) -> Int = suspend { 1 }
}
```

degrades to `J.Unknown` — silently: `parseErrors=0` and no diagnostic even with `logCompilationWarningsAndErrors(true)`. The default stays `KOTLIN_2_2`, so this only bites callers who opt into the two new enum values this PR adds, but it's worth knowing they aren't fully functional yet.

## Smaller notes

- `KotlinParser.java`: `registerInProject(extensionsStorage, environment.getProject(), ignored -> null)` — that third argument is Kotlin's `errorMessage: (Any) -> String`, and it is invoked only from the `catch (AbstractMethodError)` path to build `IllegalStateException(message, cause)`. Returning `null` discards the diagnostic in exactly the compiler-version-mismatch scenario above; something like `ext -> "Failed to register " + ext` costs nothing.
- Redundant imports: plain `FrontendConfigurationKeysKt` and `ExtensionRegistrationUtilsKt` alongside the static imports of their members.
- `KotlinTypeGoat.kt` drops `internal` from `interface Intersection` rather than suppressing the newly-promoted `EXPOSED_TYPE_PARAMETER_BOUND` error, which removes the only `internal` type from the goat fixture. `@Suppress("EXPOSED_TYPE_PARAMETER_BOUND")` would keep that coverage.
- Unrelated re-indent at `KotlinParser.java:145-146`.

barbulescu:
Fixed reported small issues but not sure what do about the compatibility issues.

timtebeek:
Thanks! I'm meanwhile working through small improvement to make mismatches easier to spot
- https://github.com/openrewrite/rewrite/pull/8554
- https://github.com/openrewrite/rewrite/pull/8558

And in parallel exploring whether we can avoid the regression on the lambda case.
- https://github.com/openrewrite/rewrite/pull/8557

timtebeek:
## Downstream recipes against projects on different Kotlin versions

Ran `org.openrewrite.staticanalysis.CommonStaticAnalysis` (rewrite-static-analysis 2.41.1, 74 sub-recipes) over a Kotlin corpus, modelling "a project on Kotlin X" as *that version's stdlib on the parse classpath, parsed at the matching language level*. Corpus is 5 files: ordinary Kotlin plus `suspend { }` lambdas (member, local and top-level), and a modern-syntax file using `when` guard conditions, multi-dollar string interpolation, `data object`, `@JvmInline value class`, sealed interfaces and non-local returns.

| project Kotlin | language level | parse errors | `J.Unknown` | missing types | files changed | patch sha |
|---|---|---|---|---|---|---|
| 2.0.21 | KOTLIN_2_0 | 0 | 0 | 6 | 1 | `972012ed6699` |
| 2.1.21 | KOTLIN_2_1 | 0 | 0 | 6 | 1 | `972012ed6699` |
| 2.2.21 | KOTLIN_2_2 | 0 | 0 | 6 | 1 | `972012ed6699` |
| 2.3.20 | KOTLIN_2_3 | 0 | 0 | 6 | 1 | `972012ed6699` |
| 2.4.10 | KOTLIN_2_4 | 0 | 0 | 6 | 1 | `972012ed6699` |

Byte-identical patches across all five, and byte-identical to released 8.90.2 (which parses the 2.3/2.4 rows at its highest available level, `KOTLIN_2_2`). Repeating every row with an empty parse classpath changes nothing. The modern-syntax file parses clean at every level too. So: no regression in recipe behaviour from this PR.

## One real problem, but it predates this PR

End-to-end through the OpenRewrite Gradle plugin (7.38.0, `rewriteDryRun`, same recipe), varying the consumer's `kotlin("jvm")`:

| consumer `kotlin("jvm")` | rewrite 8.90.2 | this PR |
|---|---|---|
| 2.0.21 | n/a | n/a |
| 2.1.21 | ✅ 1 file changed | ✅ 1 file changed |
| 2.2.21 | ✅ 1 file changed | ✅ 1 file changed |
| 2.3.20 | ✅ 1 file changed | ✅ 1 file changed |
| 2.4.10 | ⚠️ 3/3 Kotlin files fail to parse, 0 changes | ⚠️ 3/3 Kotlin files fail to parse, 0 changes |

(2.0.21 is not measurable here — KGP 2.0.21 itself dies with `IllegalArgumentException: 25.0.3` on my JDK, before rewrite runs.)

When the Kotlin Gradle Plugin is at 2.4.10, every Kotlin file — including `build.gradle.kts` and `settings.gradle.kts` — fails to parse with

```
java.lang.RuntimeException: Neither id nor name are specified
  org.jetbrains.kotlin.com.intellij.ide.plugins.IdeaPluginDescriptorImpl.<init>(IdeaPluginDescriptorImpl.kt:50)
  org.jetbrains.kotlin.com.intellij.ide.plugins.PluginDescriptorLoader.loadDescriptorFromStream(PluginDescriptorLoader.kt:189)
  ...
  org.jetbrains.kotlin.com.intellij.core.CoreApplicationEnvironment.registerExtensionPointAndExtensions(CoreApplicationEnvironment.java:244)
```

and **the build still reports success** — the only surface is a `WARN` that says "run with `--info`", and even at `--info` you get the filename but no cause; the stack trace is `DEBUG`-only. A user sees `BUILD SUCCESSFUL` and an empty result.

Two things worth separating:

- It reproduces identically on released 8.90.2, so **this PR neither causes nor worsens it** — it's driven by the KGP version on the Gradle build classpath, not by the rewrite version. It is not a merge blocker.
- But it is squarely in this PR's blast radius in spirit: the users most likely to adopt a Kotlin-2.4.10 rewrite are the ones already on KGP 2.4.10, and for them recipe runs through the Gradle plugin currently no-op silently. Probably worth an issue on rewrite-gradle-plugin (I found no existing one), covering both the classloader conflict and the fact that a wholesale parse failure is reported as success.

Harness for all of the above is in `.context/recipe-matrix` and `.context/recipe-inproc` in my workspace if anyone wants to re-run it.

timtebeek:
Thanks @barbulescu ! (And Sam for additional review). Looks like we're now good to merge and adopt downstream. The unknown guard I'd added in https://github.com/openrewrite/rewrite/pull/8557 might help turn up any issues.

PR #8561, opened 2026-08-19
Yaml: only treat `%` at the start of a line as a directive

### LINKED ISSUE #8560: YamlParser not print idempotent when a comment before the first document contains `%`

Reproduced on `rewrite-yaml` 8.72.0 through 8.88.4 (latest). Not present in 8.70.0.

### What is the smallest example that reproduces the issue?

A comment that appears **before the first (implicit) document start** and contains a `%`, followed by a blank line, is not printed back idempotently:

```yaml
# 100%

key: value
```

is printed as:

```yaml

# 100%
key: value
```

The blank line that followed the comment is moved to the very start of the file. `Parser.requirePrintEqualsInput` then rejects the source and the file becomes a `ParseError`:

```
java.lang.IllegalStateException: example.yml is not print idempotent.
@@ -1,2 +1,3 @@
+
 # 100%
-
 key: value
	at org.openrewrite.Parser.requirePrintEqualsInput(Parser.java:52)
```

As a `YamlParserTest`:

```java
@Test
void commentContainingPercentBeforeFirstDocument() {
    rewriteRun(
      yaml(
        """
          # 100%

          key: value
          """
      )
    );
}
```

Variants:

| input | idempotent |
| --- | --- |
| `# 100%\n\nkey: value\n` | ❌ |
| `# 100%\nkey: value\n` (no blank line) | ✅ (coincidentally) |
| `# 100%\n\n- a\n` (sequence) | ❌ |
| `#%\n# other\n\nkey: value\n` | ❌ (comment *and* blank line relocate) |
| `key: value\n# 100%\n\nk2: v\n` (not leading trivia) | ✅ |
| `%YAML 1.2\n---\nkey: value\n` (real directive) | ✅ |
| `key: value # 50%\n` (trailing comment) | ✅ |

### What is the full stack trace of any errors you encountered?

N/A — no exception is thrown during parsing; the parse simply is not idempotent, so the file is converted to a `ParseError`.

### Root cause

Introduced by #6529 ("Support for directives in Yaml", first released in v8.72.0).

`YamlParser.parseDirectivesFromPrefix` ([YamlParser.java#L1151-L1208](https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java#L1151)) scans the document prefix for `%` **anywhere in the text**:

```java
if (!inDirective) {
    if (c == '%') {
        inDirective = true;
        i++;
    } else {
        currentPrefix.append(c);
        i++;
    }
}
```

It does not require the `%` to be at the start of a line, and it does not skip `#` comments. For the prefix `# 100%\n\n` it therefore produces a bogus `Yaml.Directive(prefix = "# 100", value = "", suffix = "\n")` and returns the leftover `"\n"` as the document prefix. On print, `Yaml.Document` emits its prefix before its directives, so the trailing blank line is emitted first and the comment second — hence the relocation.

The sibling method `preScanDirectives` ([L1224](https://github.com/openrewrite/rewrite/blob/main/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java#L1224)) already guards this correctly:

```java
if (c == '%' && (i == 0 || source.charAt(i - 1) == '\n' || source.charAt(i - 1) == '\r')) {
```

Because `preScanDirectives` correctly finds no directives here, parsing falls back to `parseDirectives`/`parseDirectivesFromPrefix`, which does not guard. Applying the same "must be at the start of a line" check (and skipping comment lines) in `parseDirectivesFromPrefix` looks like the fix.

### Impact

Any YAML file whose leading comment block contains a `%` loses print idempotency and drops out of the LST entirely. Two common ways to hit this:

- a URL-encoded character in a link, e.g. a comment pointing at `.../My%20Project/...`
- prose such as `# 100% coverage`

Header comment blocks referencing documentation links are a widespread convention in CI pipeline definitions, so files generated from a shared template can fail en masse — one affected template comment can take out every repository derived from it.

- Closes https://github.com/moderneinc/customer-requests/issues/3028

timtebeek:
- Fixes #8560

### Problem

A comment before the first (implicit) document containing a `%` was parsed as a directive:

```yaml
# 100%

key: value
```

printed back as

```yaml

# 100%
key: value
```

- `Parser.requirePrintEqualsInput` then rejects the file and it becomes a `ParseError` — so any YAML whose header comment block contains a percent-encoded URL or prose like `# 100% coverage` drops out of the LST entirely. Introduced by #6529 (v8.72.0).

### Fix

`YamlParser.parseDirectivesFromPrefix` scanned for `%` anywhere in the document prefix. It now applies the same "must be at the start of a line" guard that its sibling `preScanDirectives` already had.

While testing this I found a second, related bug: `preScanDirectives` treated a leading comment line as document content and skipped past everything up to the next `---`, so a directive following a comment was never pre-scanned. It then fell back to reconstructing the directive from SnakeYAML's `DocumentStartEvent`, whose `DumperOptions.Version` enum has no `1.2` — so

```yaml
# comment
%YAML 1.2
---
key: value
```

round-tripped as `%YAML 1.1`. `preScanDirectives` now accumulates a comment line into the pending directive prefix instead, keeping the directive's exact source text.

### Tests

Four cases added to `DirectiveTest`, covering the variants from the issue (comment before a mapping, before a sequence, a comment starting with `%`) plus the directive-after-comment case above. All four fail on `main`.

### DIFF
--- a/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
+++ b/rewrite-yaml/src/main/java/org/openrewrite/yaml/YamlParser.java
@@ -1162,7 +1162,7 @@ public class YamlParser implements org.openrewrite.Parser {
             char c = input.charAt(i);
 
             if (!inDirective) {
-                if (c == '%') {
+                if (c == '%' && (i == 0 || input.charAt(i - 1) == '\n' || input.charAt(i - 1) == '\r')) {
                     inDirective = true;
                     i++;
                 } else {
@@ -1247,6 +1247,12 @@ public class YamlParser implements org.openrewrite.Parser {
                     inDirective = false;
                 } else {
                     value.append(c);
+                    i++;
+                }
+            } else if (c == '#') {
+                // A comment line may precede directives, so treat it as prefix rather than as content
+                while (i < source.length() && source.charAt(i) != '\n' && source.charAt(i) != '\r') {
+                    prefix.append(source.charAt(i));
                     i++;
                 }
             } else if (c == '-' && i + 2 < source.length() &&
--- a/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/DirectiveTest.java
+++ b/rewrite-yaml/src/test/java/org/openrewrite/yaml/tree/DirectiveTest.java
@@ -137,6 +137,68 @@ class DirectiveTest implements RewriteTest {
     }
 
     @Test
+    void percentInLeadingCommentIsNotADirective() {
+        rewriteRun(
+          yaml(
+            """
+              # 100%
+
+              key: value
+              """,
+            spec -> spec.afterRecipe(y -> assertThat(y.getDocuments().getFirst().getDirectives()).isEmpty())
+          )
+        );
+    }
+
+    @Test
+    void percentInLeadingCommentBeforeSequence() {
+        rewriteRun(
+          yaml(
+            """
+              # 100%
+
+              - a
+              """,
+            spec -> spec.afterRecipe(y -> assertThat(y.getDocuments().getFirst().getDirectives()).isEmpty())
+          )
+        );
+    }
+
+    @Test
+    void percentAtStartOfLeadingComment() {
+        rewriteRun(
+          yaml(
+            """
+              #%
+              # other
+
+              key: value
+              """,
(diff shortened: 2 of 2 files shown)


timtebeek:
Given the small size of the change, the readability, and test coverage, I'm fairly confident we can see this through and rely on asynchronous review if needed.

PR #8566, opened 2026-08-19
JavaScript: resolve transitive deps from npm v1 lock files

bmuschko:
`org.openrewrite.javascript.search.DependencyInsight` did not report transitive dependencies from a committed `package-lock.json` when "Include transitive dependencies" was selected.

The npm lock file parser only understood the modern `packages` map (lockfileVersion 2/3). A lockfileVersion 1 file (npm 5/6) has no `packages` map, only a nested `dependencies` tree, so the parser produced an empty resolved-dependency graph. Direct dependencies were still matched by name, but every transitive match was silently lost.

The parser now converts a legacy v1 `dependencies` tree into the same internal shape, so transitive dependencies resolve from v1 lock files just as they do from v2/v3.

### DIFF
--- a/rewrite-javascript/rewrite/src/javascript/node-resolution-result.ts
+++ b/rewrite-javascript/rewrite/src/javascript/node-resolution-result.ts
@@ -60,13 +60,65 @@ export interface PackageLockEntry {
 }
 
 /**
- * Parsed package-lock.json content structure (npm lockfile v3 format).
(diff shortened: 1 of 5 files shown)


sambsnyd (approved):
I don't know NPM dependency resolution semantics deeply enough to comment on whether this is correct, but it looks reasonable to me

jkschneider (approved):

Issue #8568, opened 2026-08-20
Most language `Assertions` ignore `TypeValidation` entirely, including `unknown()`


timtebeek:
## Background

`TypeValidation` reaches a test only through `SourceSpec.validateSource`, which `RewriteTest` invokes for the before-source and for the after-recipe result (`rewrite-test/src/main/java/org/openrewrite/test/RewriteTest.java:350` and `:498`). Each language's `Assertions` chooses what to pass, and most pass `SourceSpec.ValidateSource.noop` — either explicitly, or implicitly via the 5-arg `SourceSpec` constructor that defaults to it (`rewrite-test/src/main/java/org/openrewrite/test/SourceSpec.java:79`).

The result is that in those modules **no** `TypeValidation` invariant is checked at all — not `unknown()`, not `erroneous()`, not the type-metadata flags — even though every one of their compilation units implements `JavaSourceFile`, which is the only thing `org.openrewrite.java.Assertions.validateTypes` requires.

| Module | `validateSource` | `unknown()` honored |
|---|---|---|
| rewrite-java | `Assertions::validateTypes` | yes |
- | rewrite-kotlin | `Assertions::validateTypes` | yes, since #8557 |
| rewrite-groovy, rewrite-gradle, rewrite-scala, rewrite-python, rewrite-javascript, rewrite-csharp, rewrite-ruby, rewrite-go | `ValidateSource.noop` | **no** |

- This is not theoretical. `ScalaParserVisitor` wraps a whole file in a `J.Unknown` when no statement converts (`rewrite-scala/src/main/java/org/openrewrite/scala/ScalaParserVisitor.java:103-120`), the TypeScript parser falls back to `visitUnknown` in a dozen places, and C#/Ruby/Python all carry `visitUnknown` in their visitors. #8557 is the cautionary tale: a `J.Unknown` swallowing an entire top-level Kotlin declaration passed CI unnoticed until the check was added.

Corroborating signal: `typeValidationOptions(...)` appears in **zero** tests across groovy, scala, python, javascript, csharp, ruby and go. Nothing has ever needed to opt out, because nothing is on.

### Out of scope: rewrite-maven and rewrite-docker

Both were checked and deliberately excluded — their trees cannot hold what the remaining flags check:

[...]

PR #8570, opened 2026-08-20
Scala: fix comment-only files printing their comment twice

timtebeek:
A Scala file containing nothing but whitespace and comments produced both a `J.Unknown` holding the entire source and an EOF space holding the same text, so the comment was printed twice and the file failed to parse. `ScalaASTConverter.getRemainingSource` returned `""` whenever dotty produced an empty tree, which masked the duplication for `.scala` sources — `.sbt` sources parse to the `object __SbtScript__` wrapper, so their tree is non-empty and the duplicate was emitted.

Such a file now maps to a compilation unit with an empty statement list and the comments in EOF rather than a whole-file `J.Unknown` (which `rewrite-scala/CLAUDE.md` forbids), and EOF space is built with `ScalaSpace.format` so trailing comments become real `Comment` elements. Whitespace-only and empty files, which previously printed as empty and failed to parse, round-trip too.

Added `CommentOnlyFileTest` covering `//`- and `/* */`-only files on both the derived `.sbt` path and an explicit `.scala` path, with and without a trailing newline, whitespace-only/empty files, and comments leading or trailing a real declaration; 10 of its 14 cases fail without this fix. `./gradlew :rewrite-scala:test` is green (950 tests).

- Note for openrewrite/rewrite#8568 task A3 (enabling `TypeValidation.unknown()` for Scala): comment-only files no longer hit the fallback, but 51 whole-file `J.Unknown`s remain for bare top-level expressions (`42`, `"hello"`, `(a + b) * c`) because `ScalaASTConverter` keeps only conversions that are a `Statement`; wrapping those in the existing `S.ExpressionStatement` looks like the fix, and is left out of scope here.

### DIFF
--- a/rewrite-scala/src/main/java/org/openrewrite/scala/ScalaParserVisitor.java
+++ b/rewrite-scala/src/main/java/org/openrewrite/scala/ScalaParserVisitor.java
@@ -100,29 +100,33 @@ public class ScalaParserVisitor {
             }
         }
 
-        // If we didn't get any statements and have source content, create an Unknown node
-        // But skip if we already have a package declaration (to avoid duplication)
-        if (statements.isEmpty() && !source.trim().isEmpty() && packageDecl == null) {
-            J.Unknown.Source unknownSource = new J.Unknown.Source(
-                randomId(),
-                EMPTY,
-                Markers.EMPTY,
-                source
-            );
-
-            J.Unknown unknown = new J.Unknown(
-                randomId(),
-                EMPTY,
-                Markers.EMPTY,
-                unknownSource
-            );
-
-            statements.add(unknown);
-        }
-
-        // Get remaining source for EOF
         String remainingSource = converter.getRemainingSource(parseResult, source, result.getLastCursorPosition());
-        Space eof = remainingSource.isEmpty() ? EMPTY : Space.build(remainingSource, Collections.emptyList());
+        Space eof = remainingSource.isEmpty() ? EMPTY : ScalaSpace.format(remainingSource);
+
+        if (statements.isEmpty() && packageDecl == null) {
+            Space whole = ScalaSpace.format(source);
+            if (isWhitespaceAndComments(whole)) {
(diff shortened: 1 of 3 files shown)


rewrite-scala/src/main/java/org/openrewrite/scala/ScalaParserVisitor.java
```diff
@@ -100,29 +100,33 @@ public S.CompilationUnit visitCompilationUnit(ScalaParseResult parseResult) {
             }
         }
 
-        // If we didn't get any statements and have source content, create an Unknown node
-        // But skip if we already have a package declaration (to avoid duplication)
```
greg-at-moderne, line 104: Based on this line alone, I'd suggest adding a test case, where the whole source is a package declaration and then some whitespace/comments. Probably already handled, but worth checking.

greg-at-moderne (approved):

rewrite-scala/src/main/java/org/openrewrite/scala/ScalaParserVisitor.java
timtebeek: Added `packageDeclarationFollowedByComments` (5 cases): `package foo` followed by a line comment, a block comment, two comments, a comment with no trailing newline, and trailing whitespace only.

It was not fully handled — 4 of the 5 fail against `main`. The package path skips the `J.Unknown` fallback, so the round-trip was already correct, but EOF was built with `Space.build(text, emptyList())`, which stuffs the comment into whitespace: `getEof().getComments()` came back empty. The switch to `ScalaSpace.format` in this PR makes them real `Comment` elements, so recipes can see them.

greg-at-moderne: OK. Cool. Glad that I mentioned this then.

greg-at-moderne (approved):

PR #8573, opened 2026-08-20
Go: perform type validation in tests

greg-at-moderne:
### Checklist
- [ ] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv`
- [ ] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch

### DIFF
--- a/rewrite-go/src/main/java/org/openrewrite/golang/Assertions.java
+++ b/rewrite-go/src/main/java/org/openrewrite/golang/Assertions.java
@@ -16,6 +16,7 @@
 package org.openrewrite.golang;
 
 import org.jspecify.annotations.Nullable;
+import org.openrewrite.ParseExceptionResult;
 import org.openrewrite.SourceFile;
 import org.openrewrite.Tree;
 import org.openrewrite.golang.marker.GoProject;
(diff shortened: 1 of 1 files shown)


rewrite-go/src/main/java/org/openrewrite/golang/Assertions.java
```diff
@@ -24,23 +25,68 @@
...
+import java.util.List;
 import java.util.function.Consumer;
 
+import static java.util.stream.Collectors.joining;
+
 public final class Assertions {
     private Assertions() {
     }
 
+    public static SourceFile validateTypes(SourceFile source, TypeValidation typeValidation) {
```
timtebeek, line 44: As discussed; whilst already valuable, for now this is a misnomer: we only validate no unknowns/erroneous elements are added, whereas we likely want to validate no missing/nil types are present.

`TypeValidation` itself is misnamed as well, as it's been adding more and more types of validations we could opt into, like the addition that no non-whitespace characters should be present in any prefix, as another guard on the parser we could add here.

